PR #11996, opened 2019-01-11
[WFLY-11587] Fix Weld integration with legacy security

bstansberry:
1) Correct SecuritySubsystemRootResourceDefinition so it actually registers the org.wildfly.legacy-security.server-security-manager capability
2) Correct the module import as ServerSecurityManager is not from the security subsystem module, it's from WildFly Core's org.jboss.as.core.security.

https://issues.jboss.org/browse/WFLY-11587

PR #12010, opened 2019-01-16
[WFLY-11588] eliminates overrides of SimpleResourceDefinition.registerCapabilities

jmesnil:
Configure capabilities using
org.jboss.as.controller.SimpleResourceDefinition.Parameters#addCapabilities
instead of overriding SimpleResourceDefinition.registerCapabilities
(unless a specific treatment is required).

When registerCapabilities is overridden, always call
super.registerCapabilities()

JIRA: https://issues.jboss.org/browse/WFLY-11588


jmesnil:
This may conflict with #11996 (which also fixes a call to registerCapability). I'll rebase once #11996 is merged.

Note that there is a pending PR for WildFly Core at https://github.com/wildfly/wildfly-core/pull/3648 but there is no dependencies between the PRs and they can be merged independently.

PR #12014, opened 2019-01-18
[WFLY-11641] Wildly layers

jfdenise:
Fix for https://issues.jboss.org/browse/WFLY-11641.
Incorporates upgrade of galleon: https://issues.jboss.org/browse/WFLY-11642

This PR defines a set of layers for servlet and full.

Servlet configurations are fully defined with layers.
Full configurations are not impacted by layers.

Subsystems that are layered have been updated to advertise the set of modules they are injecting in Deployment units.


microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -43,4 +58,14 @@ protected SubsystemDefinition() {
     public Collection<AttributeDefinition> getAttributes() {
         return Collections.emptyList();
     }
+
+    @Override
+    public void registerAdditionalRuntimePackages(final ManagementResourceRegistration resourceRegistration) {
```
jmesnil: The same thing should be done for other MicroProfile subsystem (metrics, config, health).
@jfdenise Were you planning to do those in this PR or should I open another one?

jmesnil (changes requested):
Before this PR is merged, could you open WFLY issue related to the use of Galleon layers please (and another one for the Galleon upgrades)?

This PR touches many different subsystems and we need to be able to track down where/when these changes were done.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
jfdenise: I did it at some point, but this PR doesn't reflect it. Adding them in the same PR.

ee/src/main/java/org/jboss/as/ee/subsystem/EeSubsystemRootResource.java
```diff
@@ -42,6 +43,13 @@
  */
 public class EeSubsystemRootResource extends SimpleResourceDefinition {
 
+    public static String WILDFLY_NAMING = "org.wildfly.naming";
+    public static String JBOSS_INVOCATION = "org.jboss.invocation";
+    public static String JSON_API = "javax.json.api";
+    public static String JSON_BIND_API = "javax.json.bind.api";
+    public static String ECLIPSE_YASSON = "org.eclipse.yasson";
+    public static String GLASSFISH_EL = "org.glassfish.javax.el";
```
bstansberry: All of these need to be final.

bstansberry (dismissed):

ee/src/main/java/org/jboss/as/ee/subsystem/EeSubsystemRootResource.java
jfdenise: Thank-you Brian.

bstansberry:
@jfdenise Needs conflict resolved.

The change in https://github.com/wildfly/wildfly/commit/b9e23ef7ed4202424826dd26331abe9345f12953#diff-77ef0e1dc4b191bc42f01885f9dc8eff needs to be applied to infinispan-local-hibernate.xml

bstansberry:
Other than needing the conflict resolved, this LGTM. I checked all the additional package things and they look good.  I did a diff of the server built with this vs both legacy feature packs and galleon without this PR and that looks good. The layers seem reasonable.

The JIRA #s should be added to all the commit messages though.

jfdenise:
@bstansberry , I squashed the commits, only 2 commits: galleon upgrade and layers.

bstansberry (approved):

bstansberry:
retest this please

PR #12017, opened 2019-01-19
Pull request11476 base

bstansberry:
Non-RFE commits from PR 11476. Just to see if these on their own pass CI.


bstansberry:
Nope; they don't.

PR #12019, opened 2019-01-21
WFLY-11616 Add privileged block to Principal retrieval when SM is pre…

manovotn:
…sent.

JIRA - https://issues.jboss.org/browse/WFLY-11616

When injecting `Principal` with SM enabled, there was a permission check failure. Specification (at least CDI one) doesn't talk at all about any required permissions so I simply added a privileged block to handle this.

With this change in place, `CDIBuiltinInjectionTestCase` passes with `-Dsecurity.manager=true`.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/builtin/CDIBuiltinInjectionTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/builtin/CDIBuiltinInjectionTestCase.java
@@ -31,12 +31,10 @@ import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.as.test.integration.common.HttpRequest;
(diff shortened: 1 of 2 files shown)


manovotn:
@bstansberry @nziakova please review

bstansberry:
@darranl FYI.

@manovotn This seems to cause problems with SecurityAuditingTestCase; failures related to that in 4 of the 5 CI jobs. No idea why not 5 of 5.

I had a look at the EJB 3.2 spec and the EE Security spec as well and saw nothing about permission checks for similar use cases, i.e. EJBContext.getCallerPrincipal, SecurityContext.getCallerPrincipal. So +1 to having this just work; it doesn't seem like something meant to be locked down from deployment code.

manovotn:
@bstansberry Checked the test. I don't see any `@Inject` going on there, it seems to be pure EJB approach. Also, locally I am unable to reproduce this - the tests passes for me.

manovotn:
@bstansberry can you trigger retest please? Just so that we can be sure this wasn't a temporary glitch, I lack privileges to do so.

manovotn:
There we go, all green @bstansberry

bstansberry (approved):

kabir:
Retest this please

PR #12020, opened 2019-01-21
[WFLY-11617] minor fix for pooled connection factory transaction type default value and validation, update ha-policy description.

soul2zimate:
https://issues.jboss.org/browse/WFLY-11617
Based on LocalDescriptions.properties: 

> pooled-connection-factory.transaction=The type of transaction supported by this pooled connection factory (can be LOCAL, NONE or XA, default is XA).

Correct pooled connection factory transaction type default value and value validation, also update some valid values for ha-policy in description.


bstansberry:
@ehsavoie please review

ehsavoie (changes requested):
I think PooledConnectionFactoryAdd#getTxSupport should be updated to reflect this change.

soul2zimate:
I have also updated PooledConnectionFactoryAdd.java

ehsavoie (approved):

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/ConnectionFactoryAttributes.java
```diff
@@ -573,7 +574,8 @@
                 .build();
 
         SimpleAttributeDefinition TRANSACTION = SimpleAttributeDefinitionBuilder.create("transaction", STRING)
-                .setDefaultValue(new ModelNode().set("transaction"))
+                .setDefaultValue(new ModelNode().set(CommonAttributes.XA))
```
kabir: @ehsavoie @soul2zimate 
The old value 'transaction' does not seem to exist in any of the new allowed values.
CommonAttributes:
```
String LOCAL = "local";
String NONE = "none";
String XA = "xa";
```
I only see this used by PooledConnectionFactoryAdd, so I think that resource will need transformation of whichever of these new values should map to the old one.

Do you agree?

kabir (dismissed):
I think transformers are needed

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/ConnectionFactoryAttributes.java
ehsavoie: No a transformer is not required as we are narrowing the values but I think a corrector is needed to ensure retro compatibility (@soul2zimate )

kabir (dismissed):
Speaking to Emmanuel, the old one supports the same values. So we are effectively narrowing down what is set in the new version. In the old version, something not falling within the allowed values would default effectively to XA. So we should correct any 'strange' values so that parsing of old configs work

ehsavoie:
:+1:  ok for me

bstansberry:
The corrector is the right thing, but it should log a WARN if it changes the value. If it does that means the user was using an invalid value and they should correct whatever is setting that (xml file, cli script etc.)  I'd like to be able to remove this in WF 19 or later so the WARN is similar to a deprecation notice -- it tells users they are doing something we will not allow in the future.

soul2zimate:
I rebased to also log a WARN if it changes the value inside corrector.

bstansberry (approved):

bstansberry:
@kabir I think your review is out of date.

PR #12022, opened 2019-01-22
[WFLY-9450] NPE on jaxrs subsystem of FAILED deployment

baranowb:
Issue: https://issues.jboss.org/browse/WFLY-9450

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java
@@ -184,7 +184,9 @@ public class DeploymentRestResourcesDefintion extends SimpleResourceDefinition {
                     UndertowService.deploymentServiceName(server, host, contextPath));
             final UndertowDeploymentService deploymentService = (UndertowDeploymentService) controller.getService();
             try {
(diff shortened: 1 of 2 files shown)


jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java
```diff
@@ -184,7 +184,9 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
                     UndertowService.deploymentServiceName(server, host, contextPath));
             final UndertowDeploymentService deploymentService = (UndertowDeploymentService) controller.getService();
             try {
-
+                if(deploymentService.getDeployment() == null) {
+                    return;
```
bstansberry: Just returning here bypasses the catch block. Which is good in that writing to the server log, with a stack trace, is not good in this situation. OTOH it also bypasses the context.addResponseWarning in that catch block, so the user doesn't get the expected result but also gets no feedback as to why.

I can live with that though; I expect there are many other similar cases.

bstansberry (dismissed):

kabir:
Closing as https://github.com/wildfly/wildfly/pull/12731 has been merged and contains the fixes from here

PR #12024, opened 2019-01-22
WFLY-11626 Reduce unnecessary per-request array allocation/copies when parsing jsessionid

pferraro:
https://issues.jboss.org/browse/WFLY-11626

### DIFF
--- a/web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
+++ b/web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
@@ -1,0 +1,102 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.web.session;
+
+import java.nio.CharBuffer;
+import java.util.Arrays;
+import java.util.LinkedList;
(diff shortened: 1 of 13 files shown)


web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A {@link CharSequence} composed of other character sequences.
+ * All methods delegate to one or more of the underlying character sequences, using relative indexes.
+ * @author Paul Ferraro
+ */
+public class CompositeCharSequence implements CharSequence {
```
bstansberry: This can be package protected.

web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
```diff
@@ -0,0 +1,94 @@
...
+        return (result.size() == 1) ? result.get(0) : new CompositeCharSequence(result);
+    }
+
+    @Override
+    public String toString() {
+        CharBuffer buffer = CharBuffer.allocate(this.length());
+        for (CharSequence sequence : this.sequences) {
+            buffer.put(CharBuffer.wrap(sequence));
+        }
+        return String.valueOf(buffer.array());
```
bstansberry: Is there value in caching this result, at the cost of a pointer per object, to save repeated char[] allocations at L88?  I see a lot of CharSequence.toString() calls in here but I'm not sure in how many the CharSequence is one of these.

web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
pferraro: Good point.  It's a generally reusable class, however, this isn't the appropriate module for such a public class.  I'll make this package protected and submit a PR to add this to wildfly-common (seems to be an appropriate dumping ground) and drop this class once a version containing it is integrated.

web-common/src/main/java/org/jboss/as/web/session/CompositeCharSequence.java
```diff
@@ -0,0 +1,94 @@
...
+            if (relativeIndex < sequence.length()) {
+                return sequence.charAt(relativeIndex);
+            }
+            relativeIndex -= sequence.length();
+        }
+        throw new IndexOutOfBoundsException();
+    }
+
+    @Override
+    public CharSequence subSequence(int start, int end) {
```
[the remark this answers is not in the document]
pferraro: Correct - this method is not actually called from any code within WildFly (see org.wildfly.extension.undertow.session.CodedSessionConfig), hence the terse implementation.
Ultimately, this method can be optimized (in the manner to which you've alluded).  The ArrayList<CharSequence> allocation is only *really* necessary if the subsequence spans multiple CharSequences.  I might as well fix that now.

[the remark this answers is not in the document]
pferraro: OK - I've optimized the method for subsequences that fall within a single CharSequence.

bstansberry (approved):

kabir:
Retest this please

PR #12027, opened 2019-01-23
[WFLY-10212]: Deploying of jms bridge requires a lot of mandatory parameters.

ehsavoie:
* Defining default values for:
  - quality-of-service
  - failure-retry-interval
  - max-retries
  - max-batch-size
  - max-batch-time
  - module

Jira: https://issues.jboss.org/browse/WFLY-10212

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeDefinition.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeDefinition.java
@@ -37,6 +37,7 @@ import static org.wildfly.extension.messaging.activemq.MessagingExtension.MESSAG
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.EnumSet;
 
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.AttributeMarshallers;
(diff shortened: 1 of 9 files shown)


messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeDefinition.java
```diff
@@ -69,8 +70,8 @@
     private static final String SOURCE_CREDENTIAL_REFERENCE_NAME = "source-" + CredentialReference.CREDENTIAL_REFERENCE;
     private static final String TARGET_CREDENTIAL_REFERENCE_NAME = "target-" + CredentialReference.CREDENTIAL_REFERENCE;
 
-    public static final SimpleAttributeDefinition MODULE = create("module", STRING)
-            .setRequired(false)
+    public static final SimpleAttributeDefinition MODULE = create("module", STRING, true)
+            .setDefaultValue(new ModelNode("org.apache.activemq.artemis:main"))
```
jmesnil: `module` must not have a default value.

The module that it references MUST depend on the 2 JMS providers (source and targets).
Using `org.apache.activemq.artemis` is a convenient default value for our integration (where we test the JMS bridge from an embedded Artemis broker to the same one) but it's not correct from a feature point of view.

jmesnil (changes requested):
`module` must not have a default value.
Other changes are fine.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeDefinition.java
ehsavoie: ok, i understand your point of view

jmesnil (approved):

PR #12029, opened 2019-01-24
[WFLY-11606]: Upgrade Artemis to 2.6.4

ehsavoie:
* Upgrading Artemis version
* Fixing error message ids

Jira: https://issues.jboss.org/browse/WFLY-11606

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -266,7 +266,7 @@
         <version.jsoup>1.8.3</version.jsoup>
         <version.net.jcip>1.0</version.net.jcip>
         <version.net.shibboleth.utilities.java-support>7.1.1</version.net.shibboleth.utilities.java-support>
-        <version.org.apache.activemq.artemis>2.6.3.jbossorg-00014</version.org.apache.activemq.artemis>
+        <version.org.apache.activemq.artemis>2.6.4</version.org.apache.activemq.artemis>
         <version.org.apache.avro>1.7.6</version.org.apache.avro>
         <version.org.apache.cxf>3.2.7</version.org.apache.cxf>
         <version.org.apache.cxf.xjcplugins>3.2.3</version.org.apache.cxf.xjcplugins>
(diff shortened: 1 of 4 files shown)


ehsavoie:
@jmesnil could you please review ?

kabir:
@mnovak1 Is this one ok with you?

jmesnil (approved):

mnovak1:
@kabir We should wait for AMQ QE team to give us ack that Artemis 2.6.4 broker is without regressions. We ran integration tests for Artemis and they passed. So we're waiting for AMQ team atm.

bstansberry:
@dosoudil FYI. Given the above I don't know if this will get in 16.0.0.Beta1 or not. It's just a micro, I wanted to point it out.

kabir:
Adding the hold label so we don't accidentally merge until we hear from AMQ QE

kabir:
@ehsavoie before merging this we need to make sure that the dependency tree is the same, or adjust for any exclusions. We've had some problems with artemis upgrades in the past and only noticing in the product branch which has some stricter checks

CC @psakar @dosoudil

pom.xml
```diff
@@ -266,7 +266,7 @@
         <version.jsoup>1.8.3</version.jsoup>
         <version.net.jcip>1.0</version.net.jcip>
         <version.net.shibboleth.utilities.java-support>7.1.1</version.net.shibboleth.utilities.java-support>
-        <version.org.apache.activemq.artemis>2.6.3.jbossorg-00014</version.org.apache.activemq.artemis>
+        <version.org.apache.activemq.artemis>2.6.4</version.org.apache.activemq.artemis>
```
bstansberry: Does an upgrade of org.apache.qpid:proton-j come with this?

ehsavoie: no, there is no dependency change (no netty, nor proton-j)

ehsavoie:
This should be replaced by an upgrade to Apache Artemis 2.7.0 to be released soon

PR #12036, opened 2019-01-30
[WFLY-11594] Upgrade MicroProfile Metrics to 1.1.3 + tests

tterem:
* MicroProfile Metrics upgrade to 1.1.3
* tests for related issues

Jira: https://issues.jboss.org/browse/WFLY-11594
Tests for: https://issues.jboss.org/browse/WFWIP-151, https://issues.jboss.org/browse/WFWIP-152, https://issues.jboss.org/browse/WFLY-11458, https://issues.jboss.org/browse/WFLY-11457

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
@@ -41,7 +41,7 @@ import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.wildfly.test.integration.microprofile.metrics.application.resource.ResourceMeteredTimed;
-import org.wildfly.test.integration.microprofile.metrics.application.resource.TestApplication;
+import org.wildfly.test.integration.microprofile.metrics.TestApplication;
 
 
 /**
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.wildfly.test.integration.microprofile.metrics.application;
+
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getJSONMetrics;
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getMetricSubValueFromJSONOutput;
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getMetricValueFromPrometheusOutput;
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getPrometheusMetrics;
+
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.test.integration.common.HttpRequest;
+import org.jboss.as.test.shared.TimeoutUtil;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.test.integration.microprofile.metrics.application.resource.ResourceMeteredTimed;
+import org.wildfly.test.integration.microprofile.metrics.application.resource.TestApplication;
+
+
+/**
+ * Regression tests for SmallRye Metrics issues:
+ * https://github.com/smallrye/smallrye-metrics/issues/39
+ *   Difference between Prometheus and JSON format - Metered annotation
+ *
+ * https://github.com/smallrye/smallrye-metrics/issues/40
+ *   Difference between Prometheus and JSON format - Timed annotation
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class MicroProfileMetricsDifferentFormatsValueTestCase {
+
+    /**
+     * Prometheus metrics
+     */
+    private String prometheus;
+
+    /**
+     * Json metrics
+     */
+    private String json;
+
+    /**
+     * Prepare deployment
+     */
+    @Deployment
+    public static Archive<?> deploy() {
+        WebArchive war = ShrinkWrap.create(WebArchive.class, MicroProfileMetricsDifferentFormatsValueTestCase.class.getSimpleName() + ".war")
+                .addClasses(TestApplication.class)
+                .addClass(ResourceMeteredTimed.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        return war;
+    }
+
+    /**
+     * ManagementClient provides management port and hostname
+     */
+    @ContainerResource
+    ManagementClient managementClient;
+
+    /**
+     * Check Metered annotation
+     */
+    @Test
+    public void testMetered(@ArquillianResource URL url) throws Exception {
+        // requests
+        performCallsForOneEndPoint(url, "metered");
+
+        // asserts
+        commonAsserts("metered");
+        diffAssert("metered", "count", 1, "metered_total", 0.0);
+    }
+
+    /**
+     * Check Timed annotation
+     */
+    @Test
+    public void testTimed(@ArquillianResource URL url) throws Exception {
+        // requests
+        performCallsForOneEndPoint(url, "timed");
+
+        // asserts for same values as metered
+        diffAssert("timed", "count", 1, "timed_seconds_count", 0.0);
+        commonAsserts("timed");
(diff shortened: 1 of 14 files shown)


marekkopecky:
@jmesnil Can you please check this PR and consider to approve it?

@kabir Can you please allow team-city testing for this PR?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsCounterMultipleInvocationsTestCase.java
```diff
@@ -0,0 +1,98 @@
...
+            String content = EntityUtils.toString(resp.getEntity());
+            assertTrue("'Hello World!' message is expected, but was: " + content, content.contains("Hello World!"));
+         }
+      }
+
+      String jsonString = getJSONMetrics(managementClient, "application/helloCounter", true);
+      JSONObject jsonObject = new JSONObject(jsonString);
+      int helloCounter = jsonObject.getInt("helloCounter");
+
+      Assert.assertTrue("Expected value of 'helloCounter' is 5, but was " + helloCounter, helloCounter == 5);
```
jmesnil: should use a assertEquals

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsHistogramMultipleInvocationsTestCase.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+         }
+      }
+
+      json = getJSONMetrics(managementClient, "application/helloHistogram", true);
+
+      checkValue("min", 1.0);
+      checkValue("max", 5.0);
+      checkValue("mean", 3.0);
+      checkValue("stddev", 1.4142135623730951);
```
jmesnil: could you add a comment to explain how the value has been computed?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
```diff
@@ -49,18 +50,28 @@
...
         try (CloseableHttpClient client = HttpClients.createDefault()) {
 
             CloseableHttpResponse resp = client.execute(new HttpOptions(endpointURL));
-
             assertEquals(406, resp.getStatusLine().getStatusCode());
             String content = EntityUtils.toString(resp.getEntity());
             resp.close();
-            assertTrue("'No exporter found for method OPTIONS and media type' message is expected, but was: " + content,
-                    content.contains("No exporter found for method OPTIONS and media type"));
+            assertTrue("'OPTIONS method is only allowed with application/json media type.' message is expected, but was: " + content,
```
jmesnil: Having the test depending on the text is a bit brittle. Checking that the HTTP request returns 406 should be enough

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsRequestMethodsTestCase.java
```diff
@@ -83,7 +87,11 @@ public void checkPost() throws Exception {
 
         try (CloseableHttpClient client = HttpClients.createDefault()) {
             CloseableHttpResponse resp = client.execute(new HttpPost(endpointURL));
-            assertEquals(406, resp.getStatusLine().getStatusCode());
+            assertEquals(405, resp.getStatusLine().getStatusCode());
+            String content = EntityUtils.toString(resp.getEntity());
+            resp.close();
+            assertTrue("'"+ EXPECTED_RESPONSE_MESSAGE + "' message is expected, but was: " + content,
```
jmesnil: same as above. Checking HTTP status code should be enough

jmesnil:
this is ok to test

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
```diff
@@ -0,0 +1,170 @@
...
+        diffAssert(measuredName, "oneMinRate", 1, measuredName + "_one_min_rate_per_second", 0.1 * TimeoutUtil.adjust(1));
+        diffAssert(measuredName, "fiveMinRate", 1, measuredName + "_five_min_rate_per_second", 0.1 * TimeoutUtil.adjust(1));
+        diffAssert(measuredName, "fifteenMinRate", 1, measuredName + "_fifteen_min_rate_per_second", 0.1 * TimeoutUtil.adjust(1));
+    }
+
+    /**
+     * Assert maximal diff for json and prometheus values:
+     * Assert.true(application:{prometheusName} - ({jsonName}->{jsonSubName} * jsonMultiplier) <= {maximalDiff})
+     */
+    private void diffAssert(String jsonName, String jsonSubName, int jsonMultiplier, String prometheusName, double maximalDiff) {
```
jmesnil: I don't think that these differences from JSON and Prometheus output should be used to test metrics.
There is no guarantee that quick successive calls to the JSON & Prometheus output returns "identical" figures.

I'd prefer that assertions are tested separately on Prometheus and JSON output independently based on different in application invocations (that can be controlled).

marekkopecky: @jmesnil:

> There is no guarantee that quick successive calls to the JSON & Prometheus output returns "identical" figures.

There will be never identical, because [json metrics are gotten some time after Prometheus metrics](https://github.com/wildfly/wildfly/pull/12036/files#diff-bcef9e5450c8c46226465018472184c6R170).

But quick successive calls won't be never so quick, because of [this Thread.sleep(100)](https://github.com/wildfly/wildfly/pull/12036/files#diff-bcef9e5450c8c46226465018472184c6R156)

These asserts are not assertEquals, asserts have maximalDiff, so these asserts should pass. These asserts should be OK on weaker machines thanks to TimeoutUtil.

> I'd prefer that assertions are tested separately on Prometheus and JSON output independently based on different in application invocations (that can be controlled).

We still can't say the output value precisely as well in this approach. Test depends on env, [10 requests from the test](https://github.com/wildfly/wildfly/pull/12036/files#diff-bcef9e5450c8c46226465018472184c6R160) takes different time on different machines. Anyway yes, this approach is also an option. But I prefer to compare json and prometheus values, because these values should be similar (because prometheus and json requests should take similar time amout on one machine). This test approach also check the original [#39](https://github.com/smallrye/smallrye-metrics/issues/39) and [#40](https://github.com/smallrye/smallrye-metrics/issues/40) bugs better. WDYT?

kabir:
@tterem I don't know if you saw Jeff's comments, please fix :)

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
tterem: Well, the test was already depending on the text before this change and I updated it to check https://issues.jboss.org/browse/WFWIP-151 so I think the text is important here.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsRequestMethodsTestCase.java
tterem: Same here. These tests check https://issues.jboss.org/browse/WFWIP-152. Are you sure you only want to check HTTP status code?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsCounterMultipleInvocationsTestCase.java
tterem: Thanks, you are right. I will fix this.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsHistogramMultipleInvocationsTestCase.java
tterem: Ok, I will.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
bstansberry: +1 to just the 406. We should not be asserting strings provided by components that have no requirement to keep the same string.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsDifferentFormatsValueTestCase.java
bstansberry: I don't like this at all. This kind of detail should be tested in Smallrye.  I can understand some degree of testing of smallrye details within the WF testsuite, but this is class is too low level.

bstansberry (changes requested):
If a bug can be regression tested in the component's testsuite, that's where it should be done. Otherwise our testsuite will run forever and we'll constantly be unproductively chasing intermittent issues. I don't mind too much some degree of checking for things in the WF testsuite but the JSON vs Prometheus comparison one is too much for me.  The other new tests here are commented as checking for detail bugs also but they have more of a nature of a basic behavior test, i.e. make a # of invocations and check the basic response. I don't mind that.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
rsvoboda: Those string assertions were added by me 3 months ago and nobody objected.
I expect similar string checks are performed at many TestCases.

I think there should be separate issue to revisit all the MP components related checks.
For now I would not drop the check.

rsvoboda:
> I don't like this at all. This kind of detail should be tested in Smallrye. I can understand some degree of testing of smallrye details within the WF testsuite, but this is class is too low level.

This is the main TS used for EAP, it's used as part of many workflows - CVEs / one-offs / CP / new platforms testing

SmallRye TS are not executed as part of regular testing, neither in CP stream.
There are cases when component upgrades pulled in CP stream broke something and it was not noticed because tests were in component TS. Luckily that component TS was executed after tag, so we ended up with just unnecessary re-spin and not with regression.

We need safety checks in main TS to prevent regressions, look at it please as backward compatibility checks again ever-changing SmallRye landscape.

bstansberry (approved):

bstansberry:
Thanks, @tterem @marekkopecky and @rsvoboda for the hard work on identifying these issues, getting fixes and writing tests. Where tests should live is often a tricky problem and I'm sure we'll continue to have discussions about it. But that discussion shouldn't obscure the fine effort that you have made here. My apologies for kicking off a bigger discussion of how to test smallrye in a way that sounded overly negative about the work here.

PR #12038, opened 2019-02-04
[WFLY-11668]: Core queue creation is missing a routing type attribute.

ehsavoie:
* Adding a routing-type attribute to allow for core-queue creation.

Jira: https://issues.jboss.org/browse/WFLY-11668


messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/QueueDefinition.java
```diff
@@ -174,4 +184,8 @@ static boolean forwardToRuntimeQueue(OperationContext context, ModelNode operati
             return true;
         }
     }
+
+   private enum RoutingTpe {
```
jmesnil: typo

jmesnil (changes requested):
Other than the typo, it looks good.

PR #12044, opened 2019-02-06
[WFLY-11678] Reset EE concurrency managed threads to use a null TCCL …

jamezp:
…after execution is finished.

https://issues.jboss.org/browse/WFLY-11678

My only concern here is if this is if the class loader looks like it will be set to `null` before the `TransactionSetupProvider.afterProxyMethod()` method is invoked. I'm not sure if that will be an issue or not.

@mmusgrov Will [this](https://github.com/wildfly/wildfly/blob/d28f05ff7dd1e317d3377ecce30c0f63549b1e9a/transactions/src/main/java/org/jboss/as/txn/ee/concurrency/TransactionSetupProviderImpl.java#L62-L71) if the TCCL is set to `null`?

```
    public void afterProxyMethod(TransactionHandle transactionHandle, String transactionExecutionProperty) {
        final Transaction transaction = ((TransactionHandleImpl) transactionHandle).getTransaction();
        if (transaction != null) {
            try {
                transactionManager.resume(transaction);
            } catch (Throwable e) {
                TransactionLogger.ROOT_LOGGER.debug("failed to resume transaction",e);
            }
        }
    }
```

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/concurrent/handle/ClassLoaderContextHandleFactory.java
+++ b/ee/src/main/java/org/jboss/as/ee/concurrent/handle/ClassLoaderContextHandleFactory.java
@@ -79,9 +79,8 @@ public class ClassLoaderContextHandleFactory implements ContextHandleFactory {
 
         @Override
         public ResetContextHandle setup() throws IllegalStateException {
-            final ClassLoaderResetContextHandle resetContextHandle = new ClassLoaderResetContextHandle(WildFlySecurityManager.getCurrentContextClassLoaderPrivileged());
             WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(classLoader);
-            return resetContextHandle;
+            return ClassLoaderResetContextHandle.INSTANCE;
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


mmusgrov:
@jamezp Well I am not at all familiar with this class (TransactionSetupProviderImpl) but reading the code it does not rely on the class loader. What it appears to do is in the beforeProxyMethod it suspends the current transaction and returns a wrapper around it. The wrapper is subsequently passed back to the class via the afterProxyMethod which resumes the transaction.

Therefore, bearing in mind my caveat of not being familiar with this code, I do not see any issue with clearing the classloader. But I am only answering this question in relation to the reference class (TransactionSetupProviderImpl).

jamezp:
@mmusgrov Perfect thanks! That was pretty much the assessment I had too. I don't think it's going to cause an issue either.

bstansberry:
@emmartins WDYT?

emmartins:
https://github.com/emmartins/wildfly/commit/7c3035563bbf42692fab86806120aa8d50694fc5
Here is the problem with this solution, by removing the comment in this commit's test case it then fails with NPE. I will setup a profiler and dig more on the leak.

jamezp:
@emmartins It's because the `Thread.currentThread()` returns `null`. I'll fix that as it is a bug in the PR.

jamezp:
I'm placing a fixme label on this. What we really need to do is ensure we only set the TCCL on the managed thread itself. There is a possibility we're just executing this in the wrong place. Further investigation needs to be done which I will do today.

jamezp:
I'm closing this PR as it doesn't look correct. The `ContextHandle.reset()` can be invoked at times I didn't expect it to in the `ContextService` itself. This could lead to other issues and I think we need to take a much deeper look at this. We really only want this running on the `ManagedThread` itself.

PR #12049, opened 2019-02-08
Update subsystem transformer and mixed domain tests to use EAP 7.2

kabir:
https://issues.jboss.org/browse/WFLY-11688
https://issues.jboss.org/browse/WFLY-11699
https://issues.jboss.org/browse/WFLY-11700

Needs https://github.com/wildfly/wildfly-core/pull/3664


rhusar (approved):
Looks good, thx!

rhusar:
PS: I would say next time just use EAP_7_2_0 for the name straightaway rather than temporary names, so we don't have to make changes across the source tree; the only change needed would be changing the container from WF to EAP.

kabir:
Yeah, my approach made sense at the time (my thinking was it was a signal to update maven deps pulled in via the LegacyKernelServicesInitializer) but I've learnt the hard way now.

rhusar:
I guess the advantage would have been should things go very awry with WF 16 as a target, it could be disabled while more EAP 7.2 as a target tests be added... but that would be unlikely.

kabir:
Retest this please

PR #12063, opened 2019-02-12
WFLY-11703 remote-cache-container resource is missing transaction configuration

pferraro:
https://issues.jboss.org/browse/WFLY-11703


pferraro:
Static initializer seems to be called multiple times. Closing for investigation.

PR #12066, opened 2019-02-13
[WFLY-11710] Expose metrics from runtime resources

jmesnil:
* Install the MetricsRegistrationService at VERIFY stage (instead of
  RUNTIME) so that any resources created during  RUNTIME (such as
  datasources statistics) will be registered in WildFly management tree.

JIRA: https://issues.jboss.org/browse/WFLY-11710

### DIFF
--- a/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricsRegistrationService.java
+++ b/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricsRegistrationService.java
@@ -30,6 +30,8 @@ import static org.wildfly.extension.microprofile.metrics._private.MicroProfileMe
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -42,8 +44,10 @@ import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.server.Services;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
+import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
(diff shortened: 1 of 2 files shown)


microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemAdd.java
```diff
@@ -67,8 +69,15 @@ public void execute(DeploymentProcessorTarget processorTarget) {
...
-        MetricsRegistrationService.install(context, exposedSubsystems, prefix);
-        MetricsContextService.install(context, securityEnabled);
+        context.addStep(new OperationStepHandler() {
+            @Override
+            public void execute(OperationContext operationContext, ModelNode modelNode) {
+                MetricsContextService.install(context, securityEnabled);
+                MetricsRegistrationService.install(context, exposedSubsystems, prefix);
+
+            }
+        }, VERIFY);
```
bstansberry: Perhaps a comment saying that no one but you is allowed to write code that does this?

j/k

I don't have any better ideas!

bstansberry (approved):

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemAdd.java
jmesnil: @bstansberry that's ugly and very hacky :(
`VERIFY` is a special stage and I agree that I should not have to plug my services at that stage to be sure that I'm the "last" runtime service to start...

Another approach would be to use a StabilityMonitor to wait for "some" service stability to ensure all runtime resources are created and only after that install the MetricsRegistrationService. (which should equally ugly and hacky...)

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricsRegistrationService.java
```diff
@@ -70,22 +75,50 @@ static void install(OperationContext context, List<String> exposedSubsystems, St
...
+        } finally {
+            context.complete();
+        }
+
+    }
+
+    private void doStart(StartContext context) throws StartException {
+        final ServiceContainer serviceContainer = context.getController().getServiceContainer();
+        try {
+            serviceContainer.awaitStability();
```
bstansberry: In theory you might get a false positive re stability, since some OSHs may not have made their MSC changes when this task runs. So it's stable for a millisecond and then it's destabilized again.

When the OperationContext awaits stability it knows all the Stage.RUNTIMES OSHs have completed execution.

bstansberry: If this proves troublesome we can go back to the Stage.VERIFY approach.

PR #12073, opened 2019-02-13
[WFLY-11714] [WFLY-1175] Layers testing improvments

bstansberry:
https://issues.jboss.org/browse/WFLY-11714
https://issues.jboss.org/browse/WFLY-11715

Make testsuite/layers individually runnable in a manner consistent with other testsuites.

Use thin servers for testsuite/layers

### DIFF
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -44,6 +44,7 @@
                         <configuration>
                             <install-dir>${servlet.layers.install.dir}/test-standalone-reference</install-dir>
                             <plugin-options>
+                                <jboss-maven-dist/>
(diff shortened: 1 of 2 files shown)


jamezp (approved):

bstansberry:
Doing this requires changes in the base test from core to pass the maven.repo.local sys prop setting into the child process the test spawns to run the servers. Otherwise you get unstable results on CI because the maven repo the thin server is using isn't the correct one.

bstansberry:
I'm merging this because with the test now ignored the fact thin servers work incorrectly doesn't matter. But creating thin servers fixes the disk size problem that is disrupting CI.

PR #12075, opened 2019-02-14
[WFLY-11718] Ignore MicroProfileMetricsHistogramMultipleInvocationsTestCase

jmesnil:
JIRA: https://issues.jboss.org/browse/WFLY-11718


testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsHistogramMultipleInvocationsTestCase.java
```diff
@@ -53,6 +54,7 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
+@Ignore("WFLY-11718")
```
bstansberry: Is there a JIRA for fixing it?  11718 is specifically for ignoring it.

It's fine to have this pointing to the "ignore it" jira as long as the "fix it" jira will be linked to the "ignore it"; i.e. some sort of trail.

And none of it matters if it will be fixed soon. :)

bstansberry: I've had my coffee now and decided to just tweak the JIRA title.

PR #12078, opened 2019-02-15
[WFLY-11730] Do not record Galleon provisioned state and log provisioning time

aloubyansky:
https://issues.jboss.org/browse/WFLY-11730

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -77,6 +77,8 @@
                         <phase>prepare-package</phase>
                         <configuration>
                             <install-dir>${project.build.directory}/${project.build.finalName}</install-dir>
+                            <record-state>false</record-state>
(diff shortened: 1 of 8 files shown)


bstansberry:
@jfdenise Does this have any negative effect on your demos?

@yersan FYI, in case this affects eap-ization.

yersan:
Thanks for the FYI @bstansberry , I was waiting for this issue; it will affect positively to our product branches since it will resolve https://issues.jboss.org/browse/JBEAP-16213 as not needed at all

jfdenise:
@bstansberry, no impact. We provision the server using Galleon CLI, in this case the .galleon directory is generated. This directory is interesting to get information on what has been installed.

PR #12083, opened 2019-02-18
WFLY-11719 Redo how we test mod_cluster transformers to improve test …

rhusar:
…coverage

Jira
https://issues.jboss.org/browse/WFLY-11719


mod_cluster/extension/src/test/resources/org/wildfly/extension/mod_cluster/subsystem-transform-5_0_0.xml
```diff
@@ -0,0 +1,83 @@
...
+            <custom-load-metric capacity="${modcluster.custom-load-metric.capacity:1.1}"
+                                class="SomeFakeLoadMetricClass1"
+                                module="org.wildfly.extension.mod_cluster"
+                                weight="${modcluster.custom-load-metric.weight:5}"
+            />
+            <custom-load-metric class="SomeFakeLoadMetricClass2"
+                                capacity="${modcluster.custom-load-metric.capacity:1.1}"/>
+            <custom-load-metric class="SomeFakeLoadMetricClass3"
+                                weight="${modcluster.custom-load-metric.weight:5}"/>
+        </dynamic-load-provider>
```
kabir: @rhusar Any reason this does not have <ssl>? I see it in both what appear to be newer and older versions?

rhusar: The problem is that the legacy `ssl` configuration is mutually exclusive with `ssl-context` configuration, so it can only define one. While the newer versions support multiple configurations, we can test both within one file. Introducing a new profile to test this seems unnecessary since its tested in the newer profiles.

PR #12086, opened 2019-02-18
[WFLY-11406] Use new ClassFactory API to fix JDK12 issues.

ropalka:
https://issues.jboss.org/browse/WFLY-11406


ropalka:
Please review @bstansberry

ropalka:
This is not sufficient proposal pull request. It needs to be reworked.

bstansberry:
@ropalka I believe a new jboss-invocation minor is necessary. Ability to pass the a ClassFactory impl into the ProxyFactory. I suppose via ProxyConfiguration.

I say minor because that's an API change.

PR #12092, opened 2019-02-20
Documentation Updates

jamezp:
https://issues.jboss.org/browse/WFLY-11689

@bstansberry Please let me know if you'd like a JIRA for any or all of these. I'll be happy to file them and update the commit messages.

The first commit just updates the documentation version from 15 to 16.

The second commit removes the default profile and adds a new profile to allow the generated site to automatically be copied to the `wildfly.github.io` repository.

The last commit adds a README to describe how to generate and create the docs.wildfly.org web site.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -1,0 +1,83 @@
+= Building WildFly Documentation
+:toc:
+:toclevels: 2
+:icons: font
+:source-highlighter: coderay
+
+== Overview
+
+The `docs` module is not in the module listing of the parent POM by default. Therefore to build you must either activate
+the docs profile or be in the `docs` directory to build the documentation. This documentation assumes you're in the
+`docs` directory.
+
+You can activate the docs profile from the root directory with either `-Pdocs` or `-Ddocs=true`.
+
+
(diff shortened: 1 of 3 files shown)


docs/README.adoc
```diff
@@ -0,0 +1,80 @@
+= Building WildFly Documentation
+:toc:
+:toclevels: 2
+:icons: font
+:source-highlighter: coderay
+
+== Overview
+
+The `docs` module is not in the module listing of the parent POM. Therefore to build documentation you must be in the
+`docs` directory.
```
bstansberry: You can also be in the source root and activate the docs profile in the parent pom:

mvn install -pl docs -Ddocs=true

bstansberry:
Can you stick https://issues.jboss.org/browse/WFLY-11689 on the first commit? Turns out there was a source change for that and your commit is it.

docs/README.adoc
jamezp: I missed that profile, thanks!

bstansberry (approved):

PR #12097, opened 2019-02-20
WFLY-11693 Enable jgroups protocol tranformation testing previously f…

rhusar:
…ailing due to incorrect handling of model overrides by the test framework

Jira - just tucking onto https://issues.jboss.org/browse/WFLY-11693

Requires core upgrade with https://github.com/wildfly/wildfly-core/pull/3684


bstansberry:
retest this please

PR #12098, opened 2019-02-20
Try pretending new classes are in old jar

bstansberry:
This is a workaround to try and avoid failures seen on https://github.com/wildfly/wildfly-core/pull/3685 until we can do something like https://issues.jboss.org/browse/WFCORE-4346

The 7.1 tests load a legacy release of undertow in the child first classloader used by the legacy controller. But the current version is in the parent. PredicatorValidator calls to PredicateParser will eventually result in a ServiceLoader.load call for a Service that has a new impl in newer undertow releases. The META-INF/services file lists that new impl and is seen by the ServiceLoader load call. But the impl, being from the parent CL, does not implement the interface from the child CL. So things blow up.

Changing ChildFirstClassLoader to allow suppression of loading the META-INF/services file from the parent would fix this. This is a quicker workaround that simply avoids testing attributes that use PredicateValidator.


bstansberry:
retest this please

PR #12099, opened 2019-02-21
WFLY-11678: managed threads creation with factory's classloader

emmartins:
JIRA: https://issues.jboss.org/browse/WFLY-11678

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/concurrent/service/ElytronManagedThreadFactory.java
+++ b/ee/src/main/java/org/jboss/as/ee/concurrent/service/ElytronManagedThreadFactory.java
@@ -41,6 +41,8 @@ import java.security.PrivilegedAction;
  */
 public class ElytronManagedThreadFactory extends ManagedThreadFactoryImpl {
 
+    private static final ClassLoader CLASSLOADER = ElytronManagedThreadFactory.class.getClassLoader();
+
     public ElytronManagedThreadFactory(String name, ContextServiceImpl contextService, int priority) {
         super(name, contextService, priority);
     }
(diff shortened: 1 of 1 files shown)


emmartins:
@jamezp wdyt?

jamezp (changes requested):
This doesn't look right to me. We don't want to set the TCCL to the CL that created the thread. The simplest solution here is to just set the TCCL to `null` in the `run()` method.

```
@Override
public void run() {
    try {
        if (securityIdentity != null) {
            securityIdentity.runAs(super::run);
        } else {
            super.run();
        }
    } finally {
        setContextClassLoader(null);
    }
}
```

That said I think we need to look at all the `SetupContextHandle`'s and the `ResetContextHandle`'s. There could be, I have not looked, some of them that use `ThreadLocal`'s that we need to look at. It might make sense to execute these handles in the `ElytronManagedThread.run()`. Again though we'd need to spend some time looking at that.

Either way this doesn't look right to me. I think setting the TCCL to `null` is the correct thing to do.

emmartins:
The real issue is that executor service threads stay alive on run(), if they were just used once for a single task there would be no leak at all...

emmartins:
@jamezp can you please point to which CL setup doesn't look right to you?

jamezp:
@emmartins Both places where the TCCL is set don't look correct to. Also we're already in the thread itself so we should just set the threads context not through the `Thread.currentThread()`.

I don't necessarily think the handlers are wrong BTW. We just need to determine we're not leaking other things on these threads is all. Setting the context class loader to `null` on the current thread seems okay as the handlers will take care of the rest during setup.

jamezp:
retest this please

ee/src/main/java/org/jboss/as/ee/concurrent/service/ElytronManagedThreadFactory.java
```diff
@@ -41,8 +41,14 @@
...
 public class ElytronManagedThreadFactory extends ManagedThreadFactoryImpl {
 
+    /**
+     * the CL in context when instance was created, to be inherited by created threads
+     */
+    private final ClassLoader fccl;
+
     public ElytronManagedThreadFactory(String name, ContextServiceImpl contextService, int priority) {
         super(name, contextService, priority);
+        fccl = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
```
jamezp: It probably makes sense to just use the `org.jboss.as.ee` module class loader here instead of hoping we get the right one.

emmartins:
retest this please

jamezp:
This unfortunately is being blocked by https://issues.jboss.org/browse/WFLY-11769. If we merge this as is the `SecurityIdentity` used will be from the `SecurityDomain` associated with the `org.jboss.as.ee` module.

emmartins:
Please unblock cause executor service tasks are also wrapped with security
info, so when executed the task is ok. Some sort of leak is possible tho,
but not related with this issue.

jamezp (approved):

PR #12100, opened 2019-02-21
[WFLY-11754] Remove camel case from package names.

darranl:
https://issues.jboss.org/browse/WFLY-11754


kabir (approved):
@darranl My main concern was if this could change something which looks like public API, but apart from `org.jboss.as.jpa.processor.secondLevelCache` (which sounds like something internal) all packages are test packages. So I think we're good - feel free to merge after retesting has completed

bstansberry (approved):

PR #12107, opened 2019-02-22
[WFLY-11764] Expose UndertowOptions.PRESERVE_PATH_ON_FORWARD in Under…

spyrkob:
…tow settings

‼️ Alternative PR in https://github.com/wildfly/wildfly/pull/12633 ‼️ - dicussions are happening which approach to go for

JIRA: https://issues.jboss.org/browse/WFLY-11764


spyrkob:
I'm not sure if the undertow xsd version needs to be bumped as well for this change.

@fl4via could you review please?

bstansberry:
This would need transformation handling so having this in a domain mode config where legacy slaves are used would be handled properly.

bstansberry:
This doesn't seem like it is about a listener resource. It seems like it controls the behavior of the servlet container, which has its own resource.

bstansberry:
Is this a legitimately configurable behavior (i.e. multiple correct settings), or is this just a case of we had a bug, we fixed the bug, but some people may have been relying on the buggy behavior so we want a flag to let those people opt out of the fix?

If so just using the system property and logging a WARN if anyone sets it to the value that indicates wanting the buggy behavior seems appropriate. WARN should state the setting may be removed in a future release.

spyrkob:
@bstansberry yes I think it's more of an opt-out then actual configuration option. I'll rewrite this to use system property tomorrow morning. Thanks.

fl4via:
@bstansberry the behavior is not buggy. It is just that it is different from previous server, and although the spec doesn't enforce one way or another, some users were relying on the behavior of previous server. However, if we change that behavior out of the blue to be compatible to previous server, we may end up in the same situation with users relying on the way Undertow has behaved so far. That's why the property was created, to allow toggling between two different, acceptable, behaviors.
As per @bmaxwell, this property may be removed in the future if we decide to stick to the new behavior, the one that mimicks previous server.

bmaxwell:
@bstansberry  / @fl4via - yeah so I'm not a struts expert, but I know a little JSF, so based on JBEAP-16397 it looks like struts actions are like JSF where mappings where you would go to page.faces which is page.xhtml but routed through the Servlet which does the magic.  So it looks like in EAP 6 user went to page.action which does the magic on page.jsp , and either the customer app or the 3rd party thing they were using needs to know the path that was navigated to , and in EAP 7 we are returning the resolved for lack of better word path, when it needs the other.  So if we change this again in the future we should deprecate it and warn people using that system property and also have a response for Struts and other use cases like this as to what they should be using if getServletPath / getRequestURI is not correct.

spyrkob:
Just to make sure I get it right this time :) - it sounds like this should be a management option rather than system property after all (since both behaviours are acceptable)?

bstansberry:
@spyrkob 

It's up to @fl4via whether she wants to support the current (without this fix) behavior in the future or whether she intends to remove it and standardize on the "EAP 6 way".

@fl4via if you know you'll support it, or you're not sure, use a management attribute.  If you know you're going to move to the "EAP 6 way" only, then the system property is fine.

Don't use a system property for the "I'm not sure yet" case. If we're not sure, that means that today both are valid options and that means we should have a management attribute. We don't use a system property for a valid configuration option. We use those for workarounds.

You could even add a management attribute and deprecate it! That in effect exposes a proper config option but says it may go away.

This is all a bit legalistic I know.

fl4via:
@bstansberry sorry for missing this comment. I think it should be an attribute, following the line of thinking that this is not a bug and might be something that users will want to configure.

fl4via:
@spyrkob since we are on a tight schedule for the code freeze, we decided to move forward with the system property. I'll wait for the PR to be rebased before reviewing. Thanks a lot for your work on this!

undertow/src/main/java/org/wildfly/extension/undertow/ServletContainerAdd.java
```diff
@@ -92,7 +92,8 @@ public void installRuntimeServices(OperationContext context, ModelNode model, St
         final ModelNode fileCacheTtlNode = ServletContainerDefinition.FILE_CACHE_TIME_TO_LIVE.resolveModelAttribute(context, model);
         final Integer fileCacheTimeToLive = fileCacheTtlNode.isDefined()  ? fileCacheTtlNode.asInt() : null;
         final int defaultCookieVersion = ServletContainerDefinition.DEFAULT_COOKIE_VERSION.resolveModelAttribute(context, model).asInt();
-        final boolean preservePathOnForward = Boolean.parseBoolean(System.getProperty("io.undertow.servlet.dispatch.preserve_path_of_forward", "false"));
+        final ModelNode preservePathOnForwardNode = ServletContainerDefinition.PRESERVE_PATH_ON_FORWARD.resolveModelAttribute(context, model);
+        final boolean preservePathOnForward = preservePathOnForwardNode.isDefined()? preservePathOnForwardNode.asBoolean():true;
```
fl4via: Please change the default value to false. As I'm putting in the analysis document, I'm choosing the default behavior of the property added to JBoss EAP to keep backwards compatibility with EAP previous versions

fl4via (changes requested):
We need to either change the default behavior in WF or the one in JBoss EAP, I think we must keep the default behavior in JBoss EAP, even if that means to change it back to previous default in WildFly, as both differ right now.

undertow/src/main/java/org/wildfly/extension/undertow/ServletContainerAdd.java
jstourac: Just note: this will lead in default behaviour changes between WildFly 15 (old behaviour), 16, 17 (new behaviour) and 18 (old behaviour again). But I agree with Flavia, that we want to keep old behaviour as default in upcoming version of JBoss EAP to stay consistent there.

spyrkob: Thanks @fl4via. I pushed the change and also removed some redundant code form transformer test. Let me know if there's anything else I can do

undertow/src/test/java/org/wildfly/extension/undertow/UndertowTransformersTestCase.java
```diff
@@ -182,8 +187,17 @@ public void testRejectTransformersEAP_7_2_0() throws Exception {
         final PathAddress subsystemAddress = PathAddress.pathAddress(UndertowExtension.SUBSYSTEM_PATH);
         final PathAddress serverAddress = subsystemAddress.append(UndertowExtension.SERVER_PATH);
         final PathAddress hostAddress = serverAddress.append(UndertowExtension.HOST_PATH);
+
+
+        PathAddress servletContainer = subsystemAddress.append(UndertowExtension.PATH_SERVLET_CONTAINER);
+        PathAddress byteBufferPath = subsystemAddress.append(UndertowExtension.BYTE_BUFFER_POOL_PATH);
```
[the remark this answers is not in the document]
kabir: It is used further down

kabir:
@jstourac @fl4via @spyrkob I have updated the PR to address various points made

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/servlet/preservepath/PreservePathTestCase.java
```diff
@@ -95,33 +107,56 @@ public static WebArchive deployment() {
...
+
+      }
    }
 
-   private void setPreservePathOnForward(String s) throws Exception {
-      ModelNode setPreservePathOp = createOpNode("system-property=io.undertow.servlet.dispatch.preserve_path_of_forward", ModelDescriptionConstants.ADD);
-      setPreservePathOp.get("value").set(s);
+   private void setPreservePathOnForward(Boolean preservePathOnForward) throws Exception {
+      String opName = preservePathOnForward == null ? UNDEFINE_ATTRIBUTE_OPERATION : WRITE_ATTRIBUTE_OPERATION;
+      ModelNode setPreservePathOp = createOpNode("subsystem=undertow/servlet-container=default", WRITE_ATTRIBUTE_OPERATION);
```
jstourac: Instead of `WRITE_ATTRIBUTE_OPERATION`, should not be used `opName` here?

undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd
```diff
@@ -271,6 +271,7 @@
         <xs:attribute name="file-cache-metadata-size" type="xs:integer" use="optional" default="100"/>
         <xs:attribute name="file-cache-time-to-live" type="xs:integer" use="optional"/>
         <xs:attribute name="default-cookie-version" type="xs:integer"  use="optional"/>
+        <xs:attribute name="preserve-path-on-forward" type="xs:boolean" default="true"/>
```
jstourac: default is 'false' now

jstourac (changes requested):
Thank you for your update, I've put two comments that probably require code changes.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/servlet/preservepath/PreservePathTestCase.java
kabir: Good catch, I will fix

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/servlet/preservepath/PreservePathFilter.java
```diff
@@ -44,7 +44,7 @@ public void doFilter(ServletRequest servletRequest,
       file.createNewFile();
       try(FileWriter fw = new FileWriter(file)) {
          fw.write("servletPath: " + request.getServletPath());
+         fw.write("\nrequestUrl: " + request.getRequestURL().toString());
```
jstourac: I noticed that this change affects also `getRequestURI` method. It is mentioned also in the analysis. We should probably extend test coverage also for this method too.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/servlet/preservepath/PreservePathTestCase.java
```diff
@@ -95,33 +107,66 @@ public static WebArchive deployment() {
...
 
-      Thread.sleep(100);
-      FileReader fr = new FileReader(new File(tempDir + "/output.txt"));
-      Assert.assertEquals("servletPath: /preserve-path.jsp", new BufferedReader(fr).readLine());
+      long end = System.currentTimeMillis() + TIMEOUT;
+      File file = new File(tempDir + "/output.txt");
+      while (!file.exists() && System.currentTimeMillis() < end) {
+         Thread.sleep(100);
+      }
+      Assert.assertTrue(file + " was not created within the timeout", file.exists());
```
jstourac: Based on execution on my own machine, there seems to be not enough time for server to complete write into 'output.txt' file before its content is checked - when I enforce longer sleep here (5sec), the tests pass successfully for me.

jstourac: @spyrkob, I've created following PR against this branch, which looks like helps this test to be more stable https://github.com/spyrkob/wildfly/pull/1. Feel free to incorporate into your branch.

spyrkob: Thanks @jstourac, merged it in

spyrkob:
retest this please

jamezp:
This has passed the tests once and @kabir has confirmed the changes are okay so I've added the ready-to-merge label.

kabir:
@jamezp @bstansberry I am adding the feature label to remind you to close out a the rfe paperwork. It should be good to go

PR #12109, opened 2019-02-22
[WFLY-11768] Upgrade libthrift from 0.11.0 to 0.12.0.

jamezp:
https://issues.jboss.org/browse/WFLY-11768

This requires approval from @jmesnil or @jpkrohling. For this reason I've placed a hold on the PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -281,7 +281,7 @@
         <version.org.apache.openjpa>2.4.2</version.org.apache.openjpa>
         <version.org.apache.qpid.proton>0.27.3</version.org.apache.qpid.proton>
         <version.org.apache.santuario>2.1.2</version.org.apache.santuario>
-        <version.org.apache.thrift>0.11.0</version.org.apache.thrift>
+        <version.org.apache.thrift>0.12.0</version.org.apache.thrift>
         <version.org.apache.velocity>2.0</version.org.apache.velocity>
         <version.org.apache.ws.security>2.2.2</version.org.apache.ws.security>
         <version.org.apache.ws.xmlschema>2.2.1</version.org.apache.ws.xmlschema>


jmesnil:
[0.12.0 Release Notes](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310800&version=12339174)

jpkrohling:
I would feel more comfortable with a positive confirmation from QE that it indeed works. We had a similar change upstream merged a couple of weeks ago:

https://github.com/jaegertracing/jaeger-client-java/pull/587

If possible, I'd rather wait for a release and bump both the Thrift and Jaeger's version.

jstourac:
There is a bunch of tests included in WildFly testsuite that cover MP OpenTracing feature integration with WildFly [1,2]. Looking in the test job execution linked in this PR, I can see that those tests have been executed and passed. Thus the MP OpenTracing seems to be working with this PR.

Still, I guess that it is safer to wait for a version bump of `jaegertracing` dependencies that include this new version of `thrift` already. @jpkrohling for when is planned release of  jaegertracing/jaeger[core|client|thrift]? If we decided to update `jaegertracing` deps, we would have to do that quite soon as there is a freeze for WildFly this Wednesday, if I am not mistaken.

----
1. https://github.com/wildfly/wildfly/tree/master/testsuite/integration/microprofile-tck/opentracing
2. https://github.com/wildfly/wildfly/tree/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/microprofile/opentracing

jpkrohling:
True, I think we do have an e2e test ensuring a span could be reported and is available in the backend. This would be sufficient to ensure that everything is still working.

I'm not sure we have a plan for a release yet, but I think we could perform one this week. 

cc @pavolloffay

jamezp:
Closing this as it likely needs to be integrated differently.

PR #12115, opened 2019-02-26
WFLY-9529: setup ee naming context and platform bindings for ear modu…

emmartins:
…les + add ee concurrency context setup action to weld setup actions

Issue: https://issues.jboss.org/browse/WFLY-9529


emmartins:
please do not merge

emmartins:
@stuartwdouglas do you foresee any issue in what this introduces, i.e having a naming context setup for ear modules too, similar to war modules but with just the "ee platform" bindings common to all naming contexts? I believe this is the simplest solution to fix the handling of CDI app initialised/destroyed events, by non ejb CDI components in ear modules. The alternative would be to intercept around the event handling and find out the naming context selector for the module of handler class, and set it up, which I couldn't find a way to do (yet I am not very familiar with Weld integration). The linked JIRA is a good showcase of what's wrong right now, the handling of such events is atm not properly setup wrt naming and ee concurrency.  

This PR also adds the setup action of EE Concurrency to CDI, but I believe that's a peaceful change.

emmartins:
never mind I found an issue with module names clashing

PR #12124, opened 2019-02-27
Fixing test failures with sec mgr enabled on JDK11 & JDK12

ropalka:
https://issues.jboss.org/browse/WFLY-11790
https://issues.jboss.org/browse/WFLY-11791

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/naming/connector/JMXConnectorTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/naming/connector/JMXConnectorTestCase.java
@@ -40,7 +40,6 @@ import java.net.SocketPermission;
 import java.rmi.registry.LocateRegistry;
 import java.util.Properties;
 
-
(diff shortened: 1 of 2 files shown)


bstansberry:
[~ropalka] Is there a JDK bug here?  It seems strange that a call to javax.naming.spi.NamingManager.getURLContext() asking for the context for a scheme where the JDK provides the context factory would  fail with 'Permission check failed (permission"("java.lang.RuntimePermission" "accessClassInPackage.com.sun.jndi.url.rmi")"' . That's a pretty obscure permission to need to be able to invoke something the JVM provides.

Arguably the call to NamingManager.getURLContext() would be in a privileged block but that feels wrong too.

This PR is fine; these questions are more about whether this is a workaround or just what any app would have to do.

ropalka:
Good question @bstansberry but I have no answer for you because I am not security expert.
Maybe @dmlloyd or @darranl (our security guys) might shed some light on it?

ropalka:
@Darranl @dmlloyd stack traces are available in comments of WFLY-11790 & WFLY-11791

dmlloyd:
I'm not sure; this would have to be researched.

PR #12131, opened 2019-03-01
[WFLY-11787] Test checks if the identity sets by HttpServletRequest.login() is propagated into the secured EJB.

dcihak:
Upstream JIRA issue: https://issues.jboss.org/browse/WFLY-11787

Test configures Elytron to use Identity Propagation.
Test deploys application with the secured servlet and checks if the identity sets by HttpServletRequest.login() is propagated into the secured EJB.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/security/identity/propagation/IdentityPropagationServerSetupTask.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/security/identity/propagation/IdentityPropagationServerSetupTask.java
@@ -1,0 +1,191 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.test.integration.web.security.identity.propagation;
+
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.clustering.controller.Operations;
+import org.jboss.as.controller.client.helpers.ClientConstants;
+import org.jboss.as.test.integration.security.common.CoreUtils;
+import org.jboss.as.test.shared.SnapshotRestoreSetupTask;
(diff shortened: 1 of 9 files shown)


dcihak:
Retest this please

dcihak:
@spyrkob Could you please review this PR?

dcihak:
Can we merge this?

bstansberry:
@darranl or @fjuma Please review.

spyrkob:
I'm not sure, but I think the HelloBean will be called locally rather than remotely in the test, so the fix won't actually get tested.

fjuma:
@dcihak Were you able to verify that the test fails before the identity propagation changes (i.e., before Elytron 1.8.0.Final) and passes after?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/security/identity/propagation/IdentityPropagationServerSetupTask.java
```diff
@@ -0,0 +1,191 @@
...
+        ElytronDomainSetup elytronDomainSetup = new ElytronDomainSetup(getUsersFile(), getGroupsFile(), getSecurityDomainName());
+        elytronDomainSetup.setup(managementClient, containerId);
+
+        // /subsystem=elytron/http-authentication-factory=auth-test:add(http-server-mechanism-factory=global,security-domain=auth-test,mechanism-configurations=[{mechanism-name=BASIC}])
+        // /subsystem=undertow/application-security-domain=auth-test:add(http-authentication-factory=auth-test)
+        ServletElytronDomainSetup servletElytronDomainSetup = new ServletElytronDomainSetup(getSecurityDomainName());
+        servletElytronDomainSetup.setup(managementClient, containerId);
+
+        // /subsystem=elytron/sasl-authentication-factory=auth-test:add(sasl-server-factory=configured,security-domain=auth-test,mechanism-configurations=[{mechanism-name=BASIC}])
+        ModelNode addSaslAuthentication = createOpNode("subsystem=elytron/sasl-authentication-factory=" + getSecurityDomainName(), ADD);
```
darranl: Really I think it would be nice to get all of these operations into org.wildfly.test.security.common.AbstractElytronSetupTask to create ConfigurableElement instances so the same resources can be created for further test cases.

That is probably a task outside this PR as it looks like we also need to update AbstractElytronSetupTask to also extend SnapshotRestoreSetupTask or at least make use of that in the tests so the setup tasks can avoid deleting resources.

darranl:
@dcihak Did you have a chance to look into the question from @fjuma - some of these recent bug reports were quite difficult to reproduce in the first place so it would be good to confirm if the test case really does test the underlying issue that was fixed.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/security/identity/propagation/deployment/Hello.java
```diff
@@ -0,0 +1,6 @@
+package org.jboss.as.test.integration.web.security.identity.propagation.deployment;
```
darranl: Missing copyright header

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/security/identity/propagation/deployment/HelloBean.java
```diff
@@ -0,0 +1,17 @@
+package org.jboss.as.test.integration.web.security.identity.propagation.deployment;
```
darranl: Missing copyright header

darranl (changes requested):
Also a couple of missing copyright headers but the previous comments still remain, there are some questions re if this really tests the scenario that was fixed.

darranl:
I am going to close this one as it has had no updates since the original questions were asked, please feel free to either re-open or send in a new PR with the details requested.

PR #12140, opened 2019-03-05
[WFLY-11811] explicitly passing txn default timeout to TxControl during server startup

ochaloup:
https://issues.jboss.org/browse/WFLY-11811

Ensuring the {{TxControl.setDefaultTimeout(...)}} is called during the server startup.

@zhfeng , @tomjenkinson : can you please review?

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java
+++ b/transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java
@@ -24,6 +24,7 @@ package org.jboss.as.txn.service;
 
 import com.arjuna.ats.arjuna.common.CoordinatorEnvironmentBean;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
+import com.arjuna.ats.arjuna.coordinator.TxControl;
 import com.arjuna.ats.arjuna.tools.osb.mbean.ObjStoreBrowser;
 import com.arjuna.ats.jta.common.JTAEnvironmentBean;
 import com.arjuna.orbportability.internal.utils.PostInitLoader;
@@ -86,6 +87,8 @@ public final class ArjunaTransactionManagerService implements Service<com.arjuna
         coordinatorEnvironmentBean.setEnableStatistics(coordinatorEnableStatistics);
         coordinatorEnvironmentBean.setDefaultTimeout(coordinatorDefaultTimeout);
         coordinatorEnvironmentBean.setTransactionStatusManagerEnable(transactionStatusManagerEnable);
+
+        TxControl.setDefaultTimeout(coordinatorDefaultTimeout);
 
         // Object Store Browser bean
         Map<String, String> objStoreBrowserTypes = new HashMap<String, String>();
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
@@ -521,6 +521,8 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
             throws OperationFailedException {
             int timeout = resolvedValue.asInt();
 
+            // TxControl allow timeout to be set to 0 with meaning of no timeout at all
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout(timeout);
             TxControl.setDefaultTimeout(timeout);
             if (timeout == 0) {
                 ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();
@@ -536,6 +538,7 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
                                              final String attributeName, final ModelNode valueToRestore,
                                              final ModelNode valueToRevert, final Void handback)
             throws OperationFailedException {
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout(valueToRestore.asInt());
             TxControl.setDefaultTimeout(valueToRestore.asInt());
             ContextTransactionManager.setGlobalDefaultTransactionTimeout(valueToRestore.asInt());
         }


transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -521,12 +521,13 @@ protected boolean applyUpdateToRuntime(final OperationContext context, final Mod
             throws OperationFailedException {
             int timeout = resolvedValue.asInt();
 
-            TxControl.setDefaultTimeout(timeout);
             if (timeout == 0) {
                 ModelNode model = context.readResource(PathAddress.EMPTY_ADDRESS).getModel();
                 timeout = MAXIMUM_TIMEOUT.resolveModelAttribute(context, model).asInt();
                 TransactionLogger.ROOT_LOGGER.timeoutValueIsSetToMaximum(timeout);
             }
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout(timeout);
```
zhfeng: These lines should be moved to before checking the timeout equal 0, as TxControl allow 0 with the timeout which means no timeout at all.

zhfeng (changes requested):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
ochaloup: ah, correct. thank you!

zhfeng:
@ochaloup it looks good to me as we could ignore the JDK11 failures I think they are unrelated.

transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java
```diff
@@ -87,6 +88,8 @@ public synchronized void start(final StartContext context) throws StartException
         coordinatorEnvironmentBean.setDefaultTimeout(coordinatorDefaultTimeout);
         coordinatorEnvironmentBean.setTransactionStatusManagerEnable(transactionStatusManagerEnable);
 
+        TxControl.setDefaultTimeout(coordinatorDefaultTimeout);
```
tomjenkinson: I am wondering if it makes sense to move this above the setTSMEnable in order to be aligned with the changes in TransactionSubsystemRootResourceDefinition.java. I am also wondering if the fact the two changes are not atomic could be a problem? I am wondering if the TxControl.setDefaultTimeout should be done within the coordinatorEnvironmentBean.setDefaultTimeout under some synchronization block? WDYT?

ochaloup: Ok, sure moving the place sounds logic.

But... as I'm now looking into this again and thinking what you wrote I think it's a bit more difficult. 

For me the synchronization block is not needed here. The worst what can happen (and I already argument that I think it won't happen as new transaction can't (...should not) be started before the WildFly container is ready.

By me the trouble is that the information stored in the `TxControl` could be potentially inconsistent with those which are defined in the beans. If Narayana is started with configuration from system properties or descriptor then it's ensured that before the `TxControl#<clinit>` is called all the beans are initialized with information from those properties. But for the WildFly the sequence is a bit different.

1. WFLY `CoreEnvironmentService` is started
2. Narayana `CoreEnvironmentBean` is initialized (using xml descriptor and properties)
3. `CoreEnvironmentService` uses setters of the core bean
4. WFLY `ArjunaRecoveryManagerService` is started 
5. Narayana `RecoveryEnvironmentBean` is initialized
6. `ArjunaRecoveryManagerService` uses setters of the recovery bean
7. WFLY `ArjunaTransactionManagerService` is started
8. Narayana `CoordinatorEnvironmentBean` is initialized
9. `ArjunaTransactionManagerService` uses setters of the coordinator bean
10. only *!* now is time when `TxControl` can be initialized

If happens the `TxControl` class is initialized somewhere in the middle of the service startup the `TxControl` obtains and stores information not updated from the model. In fact direct use of the `TxControl`, as this PR provides, could be dangerous in the `ArjunaTransactionManagerService`.

The user issue discussed earlier (regarding of this PR) could be that (just in theory) they use Spring and Spring starts touching the transaction manager sometimes before the WFLY is set up. That means the `TxControl` would load data and would not be influenced by changes done in some WFLY service class.
I can simulate it in debugger, I'm not sure if it's a real issue and if the whole WFLY-11811 is an issue. We don't know what was the user setup to hit the inconsistency so it's hard to say.
Anyway, the `TxControl` obtains many information from `CoreEnvironmentBean` and `CoordinatorEnvironmentBean` and it should be `<clinit>` just after the beans are setup by WFLY services. 
Which is I think difficult to arrange.

tomjenkinson: Hi Ondra, I am not sure what you mean by dangerous. Do you mean with the patch as you have it now it is dangerous or that it is dangerous if Spring is doing something before WFLY starts?

I note that TxControl and CoordinatorEnvironmentBean are both in the same Narayana package. Perhaps when the CoordinatorEnvironmentBean reference that corresponds to the one from arjPropertyManager.getCoordinatorEnvironmentBean() (remember there may be more than one of these beans) is modified it calls back to TxControl to let it know about the updated value?

Something like in the ctor of CEB it works out if it is the default one and stashes that info:
```
void TxControl::setDefaultTimeout () {
  if (!init) {
    isDefault = BeanPopulator.getDefaultInstance(CoordinatorEnvironmentBean.class)
    init = true;
  }
  this.defaultTimeout = defaultTimeout;
  if (isDefault) {
    TxControl.setDefaultTimeout(this.defaultTimeout);
  }
}
```

PS regarding your sequences I imagine that number 8 _doesn't_ happen as the instance of CoordinatorEnvironmentBean is already initialized (arjPropertyManager.getCoordinatorEnvironmentBean())

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -521,6 +521,8 @@ protected boolean applyUpdateToRuntime(final OperationContext context, final Mod
             throws OperationFailedException {
             int timeout = resolvedValue.asInt();
 
+            // TxControl allow timeout to be set to 0 with meaning of no timeout at all
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout(timeout);
```
tomjenkinson: Note that there is a revertUpdateToRuntime operation too

ochaloup: yes, correct. thank you!

transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java
tomjenkinson: @ochaloup I think ^^ is still the last topic to resolve? Is the design safe or do you prefer an observer in the internals of Narayana or are they effectively identical. I am concerned as you raised a question of danger and I am not clear on whether it is dangerous or not with the patch as it is right now? I do agree that nothing should be creating transactions before the ATMS is started as that does not make sense.

ochaloup: yes, agree. Just I'm sorry I haven't got time to response on your previous comment yet.

First, sorry for the ambiguousness I meant both "dangers".

* If the Spring "decides" to touch Narayana API then it's dangerous without and still with the patch. If the `TxControl` is initialized before the WildFly transaction services are started then it can lead to errors. Unfortunately I'm currently not sure how the application of Spring-WildFly looks like. I do think the customer was using something like that as he reported EAP in some version but log errors come from Spring.
 Still, if Spring initiate Narayana classes (`<clinit>` is called) then `TxControl` is setup and the Wfly txn services do no change in that configuration.
(It could be it's not possible - aka. there can't be any valid code invoked before WildFly is fully started. )

*
> PS regarding your sequences I imagine that number 8 doesn't happen as the instance of CoordinatorEnvironmentBean is already initialized (arjPropertyManager.getCoordinatorEnvironmentBean())

 Agree, that the number `8` could not happen if the `CoordinatorEnvBean` is already initialized. What is the most important is when `TxControl` is initialized.

*
> Something like in the ctor of CEB it works out if it is the default one and stashes that info

 I think it's feasible just I'm not sure if it's practical. I mean in sense that you would need to provide such handling for any of the attribute defined in the `TxControl`: https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java#L213
For example the reason why I didn't change the order here (https://github.com/ochaloup/wildfly/blob/5ab90ca05d1594cd2b8d7238cc69b0396753bbd4/transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java#L89) is that when I would do

```
 final CoordinatorEnvironmentBean coordinatorEnvironmentBean = rjPropertyManager.getCoordinatorEnvironmentBean();
        coordinatorEnvironmentBean.setEnableStatistics(coordinatorEnableStatistics);
        coordinatorEnvironmentBean.setDefaultTimeout(coordinatorDefaultTimeout);
        TxControl.setDefaultTimeout(coordinatorDefaultTimeout);
coordinatorEnvironmentBean.setTransactionStatusManagerEnable(transactionStatusManagerEnable);
```

then the `setTransactionStatusManagerEnable` would have no reason to be called (aka. `TxController` is intialize before the `setTransactionStatusManagerEnable` is invoked thus the `TxControl` value is not redefined based on the WFLY configuration)

ochaloup: @tomjenkinson my understanding is that `TxControl` was designed as a "cache" for configuration when Java SE is used. But I don't understand why there are so many configuration options which does not have any `setter`. If I take the `TxControl#onePhase` for example. It's set up once at the start of the Narayana and then the processing in the `BasicAction` uses that value no matter if the `CoordinatorEnvironmentBean` is reconfigured. Is it some defence shield for the important settings would not be changed during runtime?

I'm still unsure how to "fix" the behaviour, if we need to fix it. For example we can change the `CEB` to change value of `TxControl` whenever the `setter` of the appropriate value (e.g. the `timeout`) is called.


I don't understand [your proposal](https://github.com/wildfly/wildfly/pull/12140#discussion_r263830525) where you store the `init` variable. I think we don't have a trouble the `TxControl#setDefaultTimeout` to be called several times. I think the issue is that to be called and when it's called. Up to that it's about all the properties in the [TxControl class](https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java#L213).

tomjenkinson: My proposal was just a way for TxControl to obtain a reference to the correct default `CEB` (don't forget there can be more than one instance but there is a specific one that arjPropertyManager.getCEB() will use)

I think your PR looks good if we can be sure the transaction cannot be created before the ATMS is started so we should merge it. We can tackle the observer side of things if it becomes apparent that some use case exists where a TX can be created before the ATMS starts - sounds good? If that is the case though we should create a thread on our community space to discuss it

ochaloup: @tomjenkinson I was debugging this a bit more and I think we can go with this fix as I'm presenting it here. And as you say when there will be some apparent issue we can discuss that on the forum.

On the doubt above on a new txn is not created before the container is started then as transaction manager is in dependency to other subsystem then the txn is needed to be started first for the other subsystem can provide that functionality.
The `TxControl` then can be accessed by the recovery manager but the recovery is started after the txn services are set-up. There is no clash in this as well.

If you are ok with this PR as it is then please approve this PR in the review.

tomjenkinson: +1

tomjenkinson (approved):

ochaloup:
@bstansberry can you please check this PR if it's feasible for merge?

PR #12144, opened 2019-03-06
[WFLY-11814] PrometheusCollector is not thread-safe

jmesnil:
synchronize the methods used to add/remove metric family samples.

JIRA: https://issues.jboss.org/browse/WFLY-11814


emmartins (commented):
This still not thread safe since it would be possible to add and remove at same time on a single ArrayList

emmartins:
I submitted https://github.com/wildfly/wildfly/pull/12135 a few days ago wrt this issue

kabir:
@emmartins @jmesnil Should this be closed? I merged #12135 already

jmesnil:
I think this one can be closed as Eduardo already fixed it.

PR #12145, opened 2019-03-06
[WFLY-11816] Bump the Undertow subsystem model version and schema to …

jamezp:
…version 9.

https://issues.jboss.org/browse/WFLY-11816

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/Namespace.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/Namespace.java
@@ -43,12 +43,13 @@ enum Namespace {
     UNDERTOW_5_0("urn:jboss:domain:undertow:5.0"),
     UNDERTOW_6_0("urn:jboss:domain:undertow:6.0"),
     UNDERTOW_7_0("urn:jboss:domain:undertow:7.0"),
-    UNDERTOW_8_0("urn:jboss:domain:undertow:8.0");
+    UNDERTOW_8_0("urn:jboss:domain:undertow:8.0"),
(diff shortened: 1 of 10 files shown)


kabir:
@jamezp adding the hold label until some actual changes come in. Then I suggest whoever does something to the undertow subsystem includes this commit as the first in their PR

jamezp:
@kabir There will be changes soon for [WFLY-11031](https://issues.jboss.org/browse/WFLY-11031). I could include this commit in that PR if that's better though.

kabir:
@jamezp IMO the version bump commit shouldn't really come in on its own. @darranl taught me that :) And I think there have been some cases of subsystems being bumped and the changes not making the cutoff.  Keeping the PR open, in order to have somewhere for this commit to live is fine though. It will be useful for you or whoever gets to actually changing the subsystem first.

ehsavoie:
@jamezp @kabir  I also have some RFE that will require this (which is where this commit is coming from ;) )

jamezp:
@ehsavoie feel free to cherry pick this then if you need it :) Or I guess a `--ff-only` merge would work too so that if we both submit it all should work well.

darranl:
If there are multiple users working off this bump please avoid cherry picking as it will mean the same commits will end up with different SHAs.

If instead you git merge and multiple PRs are submitted containing the commits in this PR git will handle that automatically even automatically closing this PR as the first one is merged.

darranl:
For anyone sending in a PR including the Undertow subsystem bump the SHA of that commit should be '4550c6afe917f496047ee649604ffd1136f6ae29'

jamezp:
I'm closing this as it's included in #12264.

PR #12147, opened 2019-03-07
[WFLY-11769] do not identity wrap runnable when creating shared manag…

emmartins:
…ed threads

Issue: https://issues.jboss.org/browse/WFLY-11769

Regarding this solution, wrapping the runnable instead of the thread seems better and safer to me, (exit() cleans thread -> runnable ref), and we should only do it when setup context handle is not null, which is when it's creation of a thread for an app. If handle is null then it's an executor factory and for such app shared threads the task wrapping done in executor should be used instead (and that's already being done). 

Please note that I am also going back here for the class loader leak solution, adopting the set CL as null (after creation) practice common everywhere in server code.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/concurrent/ElytronManagedThreadFactory.java
+++ b/ee/src/main/java/org/jboss/as/ee/concurrent/ElytronManagedThreadFactory.java
@@ -20,13 +20,12 @@
  * 2110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-package org.jboss.as.ee.concurrent.service;
+package org.jboss.as.ee.concurrent;
 
 import org.glassfish.enterprise.concurrent.AbstractManagedThread;
 import org.glassfish.enterprise.concurrent.ContextServiceImpl;
 import org.glassfish.enterprise.concurrent.ManagedThreadFactoryImpl;
 import org.glassfish.enterprise.concurrent.spi.ContextHandle;
-import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.auth.server.SecurityIdentity;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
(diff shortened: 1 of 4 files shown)


kabir:
@emmartins the security manager job fails, you are probably missing some privileged blocks

emmartins:
retest this please

emmartins:
retest this please

ee/src/main/java/org/jboss/as/ee/concurrent/ElytronManagedThreadFactory.java
```diff
@@ -41,51 +40,25 @@
...
-                );
-            } else {
-                return new ElytronManagedThread(r, contextHandleForSetup, identity);
-            }
-        } finally {
-            WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(tccl);
+    protected AbstractManagedThread createThread(Runnable r, final ContextHandle contextHandleForSetup) {
+        if (contextHandleForSetup != null) {
+            // app thread, do identity wrap
+            r = SecurityIdentityUtils.doIdentityWrap(r);
```
jamezp: This doesn't look right to me. The `Runnable` will already be wrapped in the `ManagedExecutorServiceImpl`. I don't think we need it here as it seems it could lead to the same identity issue we already see.

jamezp (changes requested):

ee/src/main/java/org/jboss/as/ee/concurrent/ElytronManagedThreadFactory.java
emmartins: If handle is not null it’s not a new thread request from executor, it’s a new thread request from app, which may obtain and use the threadfactory directly.

jamezp: Do we know for sure that the wrapped runnable itself will exit after it's complete? I'd have to dig heavily through the code to find out. I'm only looking for confirmation that we know that `Runnable` passed in will exit :)

emmartins: We decided long ago to not make such control, apps are responsible for their tasks. Anyway don't think that's relevant for the issue, threads created directly by apps are not shared, and will run with this identity while alive, while threads created by executors and shared by apps won't have an initial identity set (which should solve any related leaks too), each task submitted by apps to executor will have their own identity wrap done at executor.

jamezp:
@fjuma Could you please have a look at this as well? It seems reasonable to me.

fjuma:
+1, this looks ok to me, too.

jamezp (approved):

PR #11918, opened 2018-12-04
WFLY-11404 Add journal-file-open-timeout attribute to messaging server

TomasHofman:
https://issues.jboss.org/browse/WFLY-11404

Adds journal-file-open-timeout attribute to activemq messaging server. The parameter has already been defined on activemq side, this will allow it to be configured from Wildfly configuration.

Please make sure your PR meets the following requirements:
- [x] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue(s)
- [x] Pull Request contains description of the issue(s)
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Attached commits represent units of work and are properly formatted

For bigger changes, major and minor component upgrades make sure your PR also meets following requirements:
- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [x] Tests were added to cover changes

For new features ensure as well:
- [x] Analysis was done
- [x] Test Plan has been done
- [ ] Tests were verified in advance

PR #12150, opened 2019-03-11
[WFLY-11828]: Bump messaging schema version to 7.0

ehsavoie:
Bumping version to 7.0

Jira: https://issues.jboss.org/browse/WFLY-11828

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
@@ -92,7 +92,14 @@ import org.wildfly.extension.messaging.activemq.jms.ExternalPooledConnectionFact
  * Domain extension that integrates Apache ActiveMQ 6.
  *
  * <dl>
- * <dt><strong>Current</strong> - WildFly 16</dt>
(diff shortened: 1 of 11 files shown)


kabir:
@ehsavoie adding the hold label for the same reason as mentioned on https://github.com/wildfly/wildfly/pull/12145. It is fine to keep this in a PR so it is 'ready' but the commit should be pulled into a PR that changes the model.

darranl:
As multiple engineers could submit PRs making use of this version bump the version bump commit needs to retain it's SHA which is 'c25e0614dd376be56e862774f5b128f959b58d18'

jamezp (approved):

jamezp:
This needs to get in as #11918 changed the model and we somehow missed the version bump there.

darranl:
The flow on these subsystem bumps where multiple engineers work on the bumped version needs to be: -

1. Create a topic branch for the bump and share it somewhere (PR with Hold label is fine)
2. All engineers fork the first branch and add their changes but NEVER rebase.

If a sync with upstream is required use git merge to avoid breaking the history.

ehsavoie:
@darranl i agree, except that a feature that changed the model was merged without the proper bump in it.

PR #12153, opened 2019-03-11
Revert "[WFLY-11352] WildFly registers multiple distinct drivers for current MySQL driver jar"

maeste:
….s.Driver has different name : Revert "[WFLY-11352] WildFly registers multiple distinct drivers for current MySQL driver jar"

This reverts commit 90bb406c

Jira: https://issues.jboss.org/browse/JBEAP-16515


bstansberry:
For the idx==0 case, can both the simplified name and the complex name be registered? So the WFLY-11352 enhancement is preserved but the regression is fixed?

I think just reverting is fine though. Users going from WF 16 to WF 17 will lose the WFLY-11352 enhancement but it's more important to correct a regression. I can imagine registering under two names may cause problems, plus it's just more complexity.

jamezp:
This may be resolved in #12160. I just want to link these two PR's for now.

bstansberry (approved):

bstansberry:
retest this please

PR #12167, opened 2019-03-19
WFLY-11866 Exception can not be passed-by-reference

wfink:
https://issues.jboss.org/browse/WFLY-11866 is a regression of WFLY-2949.

Exceptions are not passed-by-reference if enabled, this cause issues if the Exception is not marked as Serializable.

- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/remote/LocalEjbReceiver.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/remote/LocalEjbReceiver.java
@@ -204,7 +204,7 @@ public class LocalEjbReceiver extends EJBReceiver {
                             result = view.invoke(interceptorContext);
                         } catch (Exception e) {
                             // WFLY-4331 - clone the exception of an async task
-                            receiverContext.resultReady(new CloningExceptionProducer(resultCloner, e));
+                            receiverContext.resultReady(new CloningExceptionProducer(resultCloner, e, allowPassByReference));
                             return;
                         }
                         // if the result is null, there is no cloning needed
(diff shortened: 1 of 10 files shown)


bstansberry:

tadamski:
@bstansberry Apparently there was some problem with exception cloning (commit 2a5561b39855b77411de2f29056f293d56405eef) so IMO we should find out what it was before merging it. @stuartwdouglas Do you remember why the change above was needed?

jaikiran:
> @bstansberry Apparently there was some problem with exception cloning (commit 2a5561b) 
@tadamski I'm just guessing here - looking at Stuart's commit there, the previous issue seems to be that the exception cloning was being passed a "target type" of the invoked method's return type instead of the `Exception` class and I think that's what Stuart fixed.

Having looked at @wfink's commit here, to me, the changes seem fine. However, given how often this part of code is runinng into issue and also looking at the linked JIRA, how (relatively) easy it's to reproduce this, I think adding a testcase for this issue would help prevent regressions.

tmiyargi:
@jaikiran test in https://github.com/wildfly/wildfly/pull/12191

tadamski (approved):

bstansberry:
@tadamski @wfink @tmiyargi 

Both the commits here were approved so once CI is done if results look good I'll merge it.

tadamski:
@bstansberry thanks!

PR #12168, opened 2019-03-19
[WFLY-11869] JSF Session / View Beans @Destroy not invoked after GC

tmiyargi:
Issue: https://issues.jboss.org/browse/WFLY-11869

View and Session scoped beans skip PreDestroy method if a gc was performed before the bean is destroyed.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
@@ -1,0 +1,84 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2019, Red Hat Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.test.integration.jsf.managedbean.gc;
+
+import java.io.Serializable;
+import java.util.logging.Logger;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.faces.bean.ManagedBean;
+import javax.faces.bean.ManagedProperty;
+import javax.faces.bean.SessionScoped;
+import javax.faces.context.ExternalContext;
+import javax.faces.context.FacesContext;
+
+@ManagedBean
+@SessionScoped
+public class CountBean implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    @ManagedProperty("#{initBean}")
+    private InitBean initBean;
+
+    private static final Logger LOG = Logger.getLogger(CountBean.class.getName());
+
+    private int count;
+
+    @PostConstruct
+    public void init() {
+        LOG.info("CountBean#Initializing counter with @PostConstruct ...");
+        count = initBean.getInit();
+    }
+
+    @PreDestroy
+    public void destroy() {
+        LOG.info("Destroyed CountBean");
+        TestResultsBean.setPreDestroySessionScoped(true);
+    }
+
+    public String invalidateGC() {
+
+        LOG.info("Running Garbage Collect and Invalidating Session");
+
+        System.gc();
+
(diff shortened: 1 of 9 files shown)


bstansberry:
@tmiyargi Do you know whether the problem fixed by https://github.com/wildfly/wildfly/commit/6b28dd31e4fe40addebbb975f70dc10e00c2d59a#diff-67f677c8e97e147da66398f585bfe13f will reappear with this fix?

@fl4via @fjuma fyi.

tmiyargi:
@bstansberry I had attached a dump from before and after the fix, I see the same objects in both so the fix did not make things worse.

tmiyargi:
@bstansberry So I do not see any memory leaks before and after applying the fix, what else is needed to get this PR merged? Any extra check I should make?

bstansberry:
@tmiyargi This needs an approval from @fl4via or @fjuma.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
```diff
@@ -0,0 +1,65 @@
+package org.jboss.as.test.integration.jsf.managedbean.gc;
```
fjuma: Just minor, this file and a few others are missing copyright headers.

tmiyargi: Added headers

fjuma:
@tmiyargi The change seems ok as long as it doesn't re-introduce that memory leak. Is it possible to get in touch with someone from QE (perhaps @honza-kasik?) to see if they can do a sanity check prior to merging this?

honza-kasik:
@fjuma We do not have any automation implemented for this specific issue. Since this issue targets CP stream, I suggest to contact someone who handles it, maybe they can do something.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
```diff
@@ -0,0 +1,86 @@
...
+        LOG.info("Destroyed CountBean");
+        LOG.info("====================");
+        TestResultsBean.setPreDestroySessionScoped(true);
+    }
+
+    public String invalidateGC() {
+
+        LOG.info("Running Garbage Collect and Invalidating Session");
+
+        System.gc();
```
tomjenkinson: I am not certain System.gc() is guaranteed to do anything synchronously: https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#gc()

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
```diff
@@ -0,0 +1,86 @@
...
+    private int count;
+
+    @PostConstruct
+    public void init() {
+        LOG.info("CountBean#Initializing counter with @PostConstruct ...");
+        count = initBean.getInit();
+    }
+
+    @PreDestroy
+    public void destroy() { // add a breakpoint here to check if this method is invoked
```
tomjenkinson: In order that the test runs successfully is a debugger required or is this some artifact of the test development?

stuartwdouglas:
So here is the question, you are claiming that the bean is no longer referenced, and GC cleans it up, and as a result destroy(bean) is never called.
If this is the case then what is going to call destroy(bean) if nothing is holding a reference to it? Is JSF using weak references internally? And if so that is a bug in JSF that we need to fix, but as it stands I think the current fix has a high likelyhood of re-introducing the JSF memory leak.

bstansberry:
@stuartwdouglas AIUI the issue is that WebInjectionContainer currently has essentially a Map<Object, WeakReference<ManagedReference>>. The key being the bean instance. And there isn't any strong ref to the ManagedReference value (unless for some reason the ManagedReferenceFactory caches one, which is doubtful). So when the gc runs the value is cleared even though the bean still exists. And that means the destroyInstance call does nothing.

Was the idea in https://github.com/wildfly/wildfly/commit/6b28dd31e4fe40addebbb975f70dc10e00c2d59a#diff-67f677c8e97e147da66398f585bfe13f to have the *key* be weak, but by mistake it was the *value* that was configured to be weak?  Perhaps not though, as that wouldn't work either, since the ManagedReference value will hold a strong ref to the bean, preventing gc of the key.

stuartwdouglas:
Ok, I see the issue. I don't really see how this could have every worked. Its likely that the other part of that commit fixed the actual memory leak, and the weak reference was never needed.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
bstansberry: System.gc may not do anything, as its use can be disabled via an arg to java, but if does do something it will be done before the call returns.

We don't disable calls to gc in our testsuite.  If we ever did (not expected) these tests would lose some value as checks for regressions of this gc issue, but they'd still be providing additional coverage that the @PreDestroy callback gets invoked.

bstansberry:
Thanks, Stuart. This LGTM.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/managedbean/gc/CountBean.java
tmiyargi: I've removed the comments. No extra set up is needed.

codylerum:
This would explain some infrequent issues I've been having with View Scoped beans bring removed with no @PreDestroy being called. I'll monitor this after it is fixed in the next Wildfly release to see if it cleans up.

PR #11802, opened 2018-11-02
[WFLY-11186][WFLY-11187] Declare a capability for weld subsystem

yersan:
- Remove the org.jboss.as.weld dependency from modules which are not using it
- Add capbility hard requirements for subsystems that depends on Weld

Notice the following: weld dependencies of xts are being treated in https://github.com/wildfly/wildfly/pull/11800, so I don't remove them here

Jira issue: https://issues.jboss.org/browse/WFLY-11186
Includes:  https://issues.jboss.org/browse/WFLY-11187

PR #12173, opened 2019-03-20
[WFLY-9529] proper EE integration for cdi app context lifecycle event…

emmartins:
…s handling, by non EE components

Issue: https://issues.jboss.org/browse/WFLY-9529

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/deployment/TransactionJndiBindingProcessor.java
+++ b/transactions/src/main/java/org/jboss/as/txn/deployment/TransactionJndiBindingProcessor.java
@@ -62,26 +62,26 @@ public class TransactionJndiBindingProcessor implements DeploymentUnitProcessor 
     @Override
     public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
         final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+        if(DeploymentTypeMarker.isType(DeploymentType.EAR, deploymentUnit)) {
+            return;
+        }
         final EEModuleDescription moduleDescription = deploymentUnit.getAttachment(Attachments.EE_MODULE_DESCRIPTION);
-
         if(moduleDescription == null) {
             return;
         }
-
         final ServiceTarget serviceTarget = phaseContext.getServiceTarget();
-        //if this is a war we need to bind to the modules comp namespace
-        if(DeploymentTypeMarker.isType(DeploymentType.WAR,deploymentUnit)) {
-            final ServiceName moduleContextServiceName = ContextNames.contextServiceNameOfModule(moduleDescription.getApplicationName(),moduleDescription.getModuleName());
-            bindServices(deploymentUnit, serviceTarget, moduleContextServiceName);
-        }
-
-        for(ComponentDescription component : moduleDescription.getComponentDescriptions()) {
-            if(component.getNamingMode() == ComponentNamingMode.CREATE) {
-                final ServiceName compContextServiceName = ContextNames.contextServiceNameOfComponent(moduleDescription.getApplicationName(),moduleDescription.getModuleName(),component.getComponentName());
-                bindServices(deploymentUnit, serviceTarget, compContextServiceName);
+        // bind to module
+        final ServiceName moduleContextServiceName = ContextNames.contextServiceNameOfModule(moduleDescription.getApplicationName(),moduleDescription.getModuleName());
+        bindServices(deploymentUnit, serviceTarget, moduleContextServiceName);
+        if (!DeploymentTypeMarker.isType(DeploymentType.WAR, deploymentUnit)) {
+            // bind to each component
+            for (ComponentDescription component : moduleDescription.getComponentDescriptions()) {
+                if (component.getNamingMode() == ComponentNamingMode.CREATE) {
+                    final ServiceName compContextServiceName = ContextNames.contextServiceNameOfComponent(moduleDescription.getApplicationName(), moduleDescription.getModuleName(), component.getComponentName());
+                    bindServices(deploymentUnit, serviceTarget, compContextServiceName);
+                }
             }
         }
-
     }
 
     /**
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/WeldSubsystemAdd.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/WeldSubsystemAdd.java
@@ -42,6 +42,7 @@ import org.jboss.as.weld.deployment.processors.BeanArchiveProcessor;
 import org.jboss.as.weld.deployment.processors.BeanDefiningAnnotationProcessor;
 import org.jboss.as.weld.deployment.processors.BeansXmlProcessor;
 import org.jboss.as.weld.deployment.processors.DevelopmentModeProcessor;
+import org.jboss.as.weld.deployment.processors.EarApplicationScopedObserverMethodProcessor;
 import org.jboss.as.weld.deployment.processors.ExternalBeanArchiveProcessor;
 import org.jboss.as.weld.deployment.processors.WebIntegrationProcessor;
 import org.jboss.as.weld.deployment.processors.WeldBeanManagerServiceProcessor;
@@ -109,6 +110,8 @@ class WeldSubsystemAdd extends AbstractBoottimeAddStepHandler {
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_BEAN_ARCHIVE, new BeanArchiveProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_EXTERNAL_BEAN_ARCHIVE, new ExternalBeanArchiveProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_PORTABLE_EXTENSIONS, new WeldPortableExtensionProcessor());
+                // TODO add processor priority to Phase
+                processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, 0x0F10, new EarApplicationScopedObserverMethodProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_COMPONENT_INTEGRATION, new WeldComponentIntegrationProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.INSTALL, Phase.INSTALL_WELD_DEPLOYMENT, new WeldDeploymentProcessor(checkJtsEnabled(context)));
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.INSTALL, Phase.INSTALL_WELD_BEAN_MANAGER, new WeldBeanManagerServiceProcessor());
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
@@ -1,0 +1,167 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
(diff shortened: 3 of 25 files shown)


emmartins:
ok this is it I think... 

1) The observer methods for the CDI app context lifecycle events are "processed" by a new CDI portable extension, which wraps them with proper setup of JNDI and EE Concurrency Context for the specific deployment "module" bringing the class, which is then used around the invocation of the app event handlers.
2) The context setup is only done for EE modules, which means classes included in EAR libs are out of  scope, the decision for this comes from random references in specs hinting inclusion of these in EE modules, and of course such CDI beans are to be considered pseudo components
3) Including such event handlers in EE components, such as EJB and Servlets, which was previously already supported, continues working fine, cause the EE component interceptors are used "later" in the call tree
4) Included also are integration tests for all "supported" use cases, i.e packaging in a ear module, standalone war and (ejb) jar and event handlers in EE component and non EE component (pure CDI bean)
5) There is a bug with the invocation of BeforeDestroyed and Destroyed event handlers not in web apps, this is due to a undeploy race condition that most times clears the CDI bean modules before these are processed to fire such events. Bug report TODO

emmartins:
@bstansberry this is quite different that I had in mind when we discussed, but I was super optimistic when considering JNDI was all needed. As users have shown in the JIRA they are really expecting other EE techs to be usable, specially EE concurrency, thus the need to do a specific EE context setup around invocation of such event handlers. Please note this could be better optimised making such beans to be considered EE components but specs do not consider those CDI beans EE components atm and going in that direction now could bring side effects, think it's preferable to go for this one more "setup actions" model and step by step add the setup of context for more techs, if makes sense and user ask for it.

weld/subsystem/src/main/java/org/jboss/as/weld/WeldSubsystemAdd.java
```diff
@@ -109,6 +110,8 @@ protected void execute(DeploymentProcessorTarget processorTarget) {
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_BEAN_ARCHIVE, new BeanArchiveProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_EXTERNAL_BEAN_ARCHIVE, new ExternalBeanArchiveProcessor());
                 processorTarget.addDeploymentProcessor(WeldExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_WELD_PORTABLE_EXTENSIONS, new WeldPortableExtensionProcessor());
+                // TODO add processor priority to Phase
```
emmartins: Almost forgot, please this note that if this PR makes sense and passes review+tests we need to first add an entry to Phase for this new weld deployment processor, and update here

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentProcessor.java
```diff
@@ -324,12 +325,16 @@ public void undeploy(DeploymentUnit context) {
         }
     }
 
-    static List<SetupAction> getSetupActions(DeploymentUnit deploymentUnit) {
+    public static List<SetupAction> getSetupActions(DeploymentUnit deploymentUnit) {
```
bstansberry: I haven't digested the important part of this yet, but AFAICT this visibility change is not needed.

emmartins: correct, at some point had the new processor in different package and missed this was not needed after all cleanup, fixed now

jamezp:
I've not looked in detail at this PR, but these failures seem related.

emmartins:
retest this please

emmartins:
@bstansberry so this is it, now there is something I am not sure, and that's the scope of deployments we allow such events to have the EE context/tech available to use. I mean on this PR it's any deployment module that has an EEModuleDescriptor, which pretty much translates to any module but EAR libs, and for that reason the added integration test changes wrt replacing default JMS factory (arquillian deployment gets entitled to use such stuff too thus it has default EE bindings), is that too much I'm not really sure! Please note that from this wide coverage we can reduce it to as much as just cover official EE modules (war, ejb, rar)...

transactions/src/main/java/org/jboss/as/txn/deployment/TransactionJndiBindingProcessor.java
```diff
@@ -57,26 +57,26 @@
...
-        }
-
-        for(ComponentDescription component : moduleDescription.getComponentDescriptions()) {
-            if(component.getNamingMode() == ComponentNamingMode.CREATE) {
-                final ServiceName compContextServiceName = ContextNames.contextServiceNameOfComponent(moduleDescription.getApplicationName(),moduleDescription.getModuleName(),component.getComponentName());
-                bindServices(deploymentUnit, serviceTarget, compContextServiceName);
+        // bind to module
+        final ServiceName moduleContextServiceName = ContextNames.contextServiceNameOfModule(moduleDescription.getApplicationName(),moduleDescription.getModuleName());
+        bindServices(deploymentUnit, serviceTarget, moduleContextServiceName);
+        if (!DeploymentTypeMarker.isType(DeploymentType.WAR, deploymentUnit)) {
```
bstansberry: This DUP didn't previously ignore non-WAR deployments.

Is this guard just an optimization because for a war component.getNamingMode() == ComponentNamingMode.CREATE will always be false?

bstansberry:
retest this please

bstansberry:
@emmartins The tests are failing now.

Without investigating at all my bet is it relates to this week's EE concurrency change. :)

emmartins:
Ok, will check it, thanks.

emmartins:
@bstansberry turns out the issue was introduced by 
FYI below was the forced pushed diff, IMHO deprecated classes should continue to work as expected till removal

`
diff --git a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
index 57407d7f76..6fc4045931 100644
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
@@ -22,19 +22,20 @@
 
 package org.jboss.as.weld.deployment.processors;
 
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.ee.structure.DeploymentType;
 import org.jboss.as.ee.structure.DeploymentTypeMarker;
-import org.jboss.as.ee.weld.WeldDeploymentMarker;
+
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.server.deployment.SetupAction;
+import org.jboss.as.weld.WeldCapability;
 import org.jboss.as.weld.deployment.BeanDeploymentArchiveImpl;
 import org.jboss.as.weld.deployment.BeanDeploymentModule;
 import org.jboss.as.weld.deployment.WeldAttachments;
-import org.jboss.as.weld.deployment.WeldPortableExtensions;
 import org.jboss.as.weld.logging.WeldLogger;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -52,6 +53,8 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Set;
 
+import static org.jboss.as.weld.Capabilities.WELD_CAPABILITY_NAME;
+
 /**
  * Processor that registers a CDI portable extension for EAR deployments, which adds support for EE facilities, to CDI app context lifecycle event handlers.
  *
@@ -62,8 +65,16 @@ public class EarApplicationScopedObserverMethodProcessor implements DeploymentUn
     @Override
     public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
-        if (DeploymentTypeMarker.isType(DeploymentType.EAR, deploymentUnit) && WeldDeploymentMarker.isPartOfWeldDeployment(deploymentUnit)) {
-            WeldPortableExtensions.getPortableExtensions(deploymentUnit).registerExtensionInstance(new PortableExtension(deploymentUnit), deploymentUnit);
+        if (!DeploymentTypeMarker.isType(DeploymentType.EAR, deploymentUnit)) {
+            // ear deployment only processor
+            return;
+        }
+        final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+            final WeldCapability api = support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class).get();
+            if (api.isPartOfWeldDeployment(deploymentUnit)) {
+                api.registerExtensionInstance(new PortableExtension(deploymentUnit), deploymentUnit);
+            }
         }
     }
`

bstansberry:
@emmartins Do you have any more details on how #11802 broke this? If so, please file a JIRA, priority Critical (which I may bump to Blocker).  

I assume the fix you made was related to https://github.com/wildfly/wildfly/pull/12173/files#diff-1c9f40193628011327e879218da8b956R72 -- using the capability instead of direct access.

We don't want the deprecated code to not work properly as we can't assume code outside of WildFly will have moved to using the capability.

I didn't comment on the PR saying to use the capability in the first place because I assumed a subsystem doesn't need to use a cap to access its internal details.

emmartins:
@bstansberry seems both `WeldDeploymentMarker.isPartOfWeldDeployment(deploymentUnit)` and `WeldPortableExtensions.getPortableExtensions(deploymentUnit).registerExtensionInstance(new PortableExtension(deploymentUnit), deploymentUnit)` are not working as expected. Will fill JIRA tomorrow.

emmartins:
@bstansberry jira created --> https://issues.jboss.org/browse/WFLY-12058

transactions/src/main/java/org/jboss/as/txn/deployment/TransactionJndiBindingProcessor.java
emmartins: Aligning with the introduced abstract processor, honestly didn't dig if it's just an optimisation or a guard to possible misusage/extension where a war module may have a component that was lazily marked as having a comp naming context. This should probably should be further reworked to move logic present in way too many processors to the EE module and component description objects

yersan:
@emmartins 

Yes, the deprecated ones are not working as they did before, those classes maintain an internal state. The internal state is no longer used because #11802 removed any uses of these classes deprecated. 

e.g, that means, we cannot use https://github.com/wildfly/wildfly/blob/master/ee/src/main/java/org/jboss/as/ee/weld/WeldDeploymentMarker.java#L53 anymore because nobody is calling https://github.com/wildfly/wildfly/blob/master/ee/src/main/java/org/jboss/as/ee/weld/WeldDeploymentMarker.java#L43

I'm working on a fix in https://issues.jboss.org/browse/WFLY-12058.

If you use in your code the non-deprecated versions it should work though, could you confirm?

emmartins:
Yes it works, this PR uses the capability api since yesterday’s force push, but please note that the serious concerns are the usage of the deprecated classes by 3rd parties, WFLY-12058 may even become a blocker cause CDI portable extensions are popular among the community.

bstansberry (approved):

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/EarApplicationScopedObserverMethodProcessor.java
```diff
@@ -0,0 +1,167 @@
...
+public class EarApplicationScopedObserverMethodProcessor implements DeploymentUnitProcessor {
+
+    @Override
+    public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+        final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+        if (!DeploymentTypeMarker.isType(DeploymentType.EAR, deploymentUnit)) {
+            // ear deployment only processor
+            return;
+        }
+        final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
```
yersan: Just a really minor detail, it's harmless but it is unnecessary to use the capability support here to access to the Weld capability API. We should use the capability service support if we are outside of Weld subsystem, but it is not necessary if you are in the Weld subsystem itself. Direct access to non-deprecated WeldPortableExtensions and WeldDeploymentMarker must work fine.

yersan: Regarding my last comment, if finally it is decided to do not use the capability here, please take into account the changes included in https://github.com/wildfly/wildfly/pull/12270, we should keep using the deprecated WeldPortableExtensions for internal uses in Weld Subsystem

PR #12178, opened 2019-03-21
[WFLY-11876]: Upgrade Artemis to 2.7.0

ehsavoie:
* Updating to Apache Artemis 2.8.0
* Taking new Artemis native packaging into account
* Taking advantage of ARTEMIS-2163.
* Aligning qpid j-proton version with Artemis

Jira: https://issues.jboss.org/browse/WFLY-11876
        https://issues.jboss.org/browse/WFLY-11727

PR #12179, opened 2019-03-21
[WFLY-6684]: JMS Bridge should display statistics about messages that have been processed.

ehsavoie:
Displaying 2 new statistics on JMSBridge:
 - number of processed messages
 - number of aborted messages

Jira: https://issues.jboss.org/browse/WFLY-6684
JBEAP: https://issues.jboss.org/browse/JBEAP-16579

This PR depends on https://github.com/wildfly/wildfly/pull/12178

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
@@ -558,6 +558,27 @@ CLI:
 }
 ----
 
+[[jms-bridge-statistics]]
+=== Statistics of a JMS Bridge
+
+Currently two statistics are available on a JMS bridge: the number of processed messages and the number of aborted/rolled back messages.
+Those are available with the following command :
+
+[source, ruby]
+----
+/subsystem=messaging/jms-bridge=myBridge:read-attribute(name=message-count)
+{
+    "outcome" => "success",
+    "result" => 0L
+}
+
+/subsystem=messaging/jms-bridge=myBridge:read-attribute(name=aborted-message-count)
+{
+    "outcome" => "success",
+    "result" => 0L
+}
+----
+
 == Component Reference
 
 The messaging-activemq subsystem is provided by the Artemis project. For
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
@@ -23,13 +23,13 @@
 package org.wildfly.extension.messaging.activemq.jms.bridge;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.START;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STOP;
 import static org.wildfly.extension.messaging.activemq.CommonAttributes.NAME;
 import static org.wildfly.extension.messaging.activemq.CommonAttributes.PAUSED;
 import static org.wildfly.extension.messaging.activemq.CommonAttributes.STARTED;
+import static org.wildfly.extension.messaging.activemq.jms.bridge.JMSBridgeDefinition.ABORTED_MESSAGE_COUNT;
 import static org.wildfly.extension.messaging.activemq.jms.bridge.JMSBridgeDefinition.PAUSE;
 import static org.wildfly.extension.messaging.activemq.jms.bridge.JMSBridgeDefinition.RESUME;
 
@@ -37,8 +37,6 @@ import org.apache.activemq.artemis.jms.bridge.JMSBridge;
 import org.jboss.as.controller.AbstractRuntimeOnlyHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
-import org.jboss.as.controller.PathAddress;
-import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.logging.ControllerLogger;
 import org.jboss.as.controller.operations.validation.ParametersValidator;
 import org.jboss.as.controller.operations.validation.StringLengthValidator;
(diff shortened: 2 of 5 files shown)


ehsavoie:
retest this please

jamezp:
This could use a rebase since some of the other PR's here have been merged.

jamezp:
I'm not sure why all the CI jobs either didn't run or report back, but retest this please

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
```diff
@@ -558,6 +558,41 @@ CLI:
...
+    "outcome" => "success",
+    "result" => [{
+        "address" => [
+            ("subsystem" => "messaging"),
+            ("jms-bridge" => "myBridge")
+        ],
+        "outcome" => "success",
+        "result" => 0L
+    }]
+}
```
bstansberry: This should not be the structure of the response. I believe it is simply:

```
{
  "outcome" => "success",
  "result" => 0L
}
```

What you show above is what you'd see if the request had been for jms-bridge=*, not for a specific bridge

/subsystem=messaging/jms-bridge=*:read-attribute(name=message-count)

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
```diff
@@ -558,6 +558,41 @@ CLI:
...
```
bstansberry: Same comment as for previous example.

messaging-activemq/src/main/resources/org/wildfly/extension/messaging/activemq/LocalDescriptions.properties
```diff
@@ -506,13 +506,15 @@ http-connector=Used by a remote client to define how it connects to a server ove
 http-listener=The Undertow's http-listener that handles HTTP upgrade requests.
 in-vm-acceptor=Defines a way in which in-VM connections can be made to the ActiveMQ server.
 in-vm-connector=Used by an in-VM client to define how it connects to a server.
+jms-bridge.aborted-message-count=The number of messages aborted/rollback.
```
bstansberry: s/rollback/rolled back/g

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
+                }
+                break;
+            case START:
+                try {
+                    // we do not start the bridge directly but call startBridge() instead
+                    // to ensure the class loader will be able to load any external resources
+                    JMSBridgeService service = (JMSBridgeService) bridgeService.getService();
+                    service.startBridge();
+                } catch (Exception e) {
+                    context.getFailureDescription().set(e.getLocalizedMessage());
```
bstansberry: This applies to all these cases, and is an existing problem...

These all involve failed manipulation of the server runtime state but this handling means nothing about them will appear in the server log.  Can these just throw on any RuntimeException, wrap any checked exception in RuntimeException and throw on, and count on the throwable handling in AbstractOperationContext.executeStep (e.g. currently https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L1032) to log and prepare a response to the client?  Even the failure-description it prepares at L1037 is likely to be more meaningful than e.getLocalizedMessage() which typically doesn't include the exception class and often isn't understandable without that.

bstansberry (changes requested):
Please see my previous comments.

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/JMSBridgeTest.java
```diff
@@ -136,6 +148,15 @@ public void sendAndReceiveMessage() throws Exception {
                 connection.close();
             }
         }
+    }
 
+    private long readMetrics() throws IOException {
```
honza-kasik: Can you also please add a smoke for second metrics attribute? (aborted message count)

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/JMSBridgeTest.java
```diff
@@ -125,6 +136,7 @@ public void sendAndReceiveMessage() throws Exception {
             assertTrue(receivedMessage instanceof TextMessage);
             assertEquals(text, ((TextMessage) receivedMessage).getText());
             assertNotNull("did not get header set by the JMS bridge", receivedMessage.getStringProperty(ActiveMQJMSConstants.AMQ_MESSAGING_BRIDGE_MESSAGE_ID_LIST));
+            assertEquals("bridge metrics are not correct", 5000L,readMetrics());
```
honza-kasik: Can you please edit the message slightly so it would be more explaining? It is obvious, that when assert fails, something is not correct/expected...

honza-kasik (changes requested):
@ehsavoie Thanks for adding tests too. I made a review and I left two comments there. IMHO adding a smoke for a second attribute would be great if you have already test for one.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
  * @author Jeff Mesnil (c) 2012 Red Hat Inc.
  */
 public class JMSBridgeHandler extends AbstractRuntimeOnlyHandler {
 
-    public static final JMSBridgeHandler INSTANCE = new JMSBridgeHandler();
+    public static final JMSBridgeHandler INSTANCE = new JMSBridgeHandler(false);
+    public static final JMSBridgeHandler READ_ONLY_INSTANCE = new JMSBridgeHandler(true);
 
     private ParametersValidator readAttributeValidator = new ParametersValidator();
+    private boolean readOnly;
```
jamezp: Not really a merge blocker, but this could be `final`.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
+                }
+                break;
+            case START:
+                try {
+                    // we do not start the bridge directly but call startBridge() instead
+                    // to ensure the class loader will be able to load any external resources
+                    JMSBridgeService service = (JMSBridgeService) bridgeService.getService();
+                    service.startBridge();
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
jamezp: I think an `OperationFailedException` might be better here. With a `RuntimeException` the stack trace will always be logged which is maybe what we're after here. Previously though this just set a failure description.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
+                }   break;
+            case STOP:
+                try {
+                    bridge.stop();
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
jamezp: Same comment as above.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
+                }   break;
+            case PAUSE:
+                try {
+                    bridge.pause();
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
jamezp: Same comment as above.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
```diff
@@ -47,101 +45,110 @@
...
+                }   break;
+            case RESUME:
+                try {
+                    bridge.resume();
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
jamezp: Same comment as above.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeHandler.java
bstansberry: @jamezp An OFE should only be used to indicate a caller mistake, not a server fault.

Only setting a failure description means a server fault occurred but there would be nothing in the server log to indicate that.

This handling is good, other than perhaps catching and throwing on any RuntimeException and only wrapping checked exceptions. But that's a nit.

bstansberry: An OFE would be appropriate if the bridge was already started and that's what caused startBridge to throw an exception. Looking at the code though if that happened there'd be no exception, just an error msg in the server log.

bstansberry: @jamezp BTW the reason this handling changed vs previously in response to my review comment on this file from 10 days ago. :)

jamezp: Perfect thanks for the explanation!

bstansberry (approved):

bstansberry (approved):

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/JMSBridgeTest.java
```diff
@@ -125,6 +136,8 @@ public void sendAndReceiveMessage() throws Exception {
             assertTrue(receivedMessage instanceof TextMessage);
             assertEquals(text, ((TextMessage) receivedMessage).getText());
             assertNotNull("did not get header set by the JMS bridge", receivedMessage.getStringProperty(ActiveMQJMSConstants.AMQ_MESSAGING_BRIDGE_MESSAGE_ID_LIST));
+             assertEquals("Message count bridge metric is not correct", 5000L, readMetric("message-count"));
```
bstansberry: @ehsavoie Why is this 5000L and not 1L?

bstansberry (commented):
I've withdrawn my approval because I don't understand an assert in the test addition.

bstansberry (changes requested):
See above.

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/JMSBridgeTest.java
ehsavoie: Because a NPE was thrown when getting the value which is 1L. Test fixed

bstansberry (approved):

bstansberry:
@ehsavoie SM failure at https://ci.wildfly.org/viewLog.html?buildId=155604&buildTypeId=WF_PullRequest_LinuxSm is related.

PR #12186, opened 2019-03-27
WFLY-11715, The layers testsuite should use thin servers, un-ignore the layers test

jfdenise:
This is the last step of : https://issues.jboss.org/browse/WFLY-11715
Un-ignoring the layers test.


kabir (approved):

jfdenise:
@kabir I realise that the path to m2 local repository is not passed to test, I need to double check something before this change can be merged.

jfdenise:
@kabir , that is fine, I just added a note in layers/pom.xml to say from where the maven.local.repo system property comes from.

PR #12187, opened 2019-03-27
[WFLY-11911] Remove READ_TIMEOUT from channel configuration options

tmiyargi:
Issue: https://issues.jboss.org/browse/WFLY-11911

READ_TIMEOUT cannot be set as channel option as sample configuration is missleading.

The only configuration options allowed on the channel by the ultimate consumer of the options org.jboss.remoting3.remote.RemoteReadListener are
RemotingOptions.TRANSMIT_WINDOW_SIZE,
RemotingOptions.MAX_OUTBOUND_MESSAGES
RemotingOptions.RECEIVE_WINDOW_SIZE
RemotingOptions.MAX_INBOUND_MESSAGES
RemotingOptions.MAX_OUTBOUND_MESSAGE_SIZE
RemotingOptions.MAX_INBOUND_MESSAGE_SIZE.

### DIFF
--- a/ejb3/src/main/resources/subsystem-templates/ejb3.xml
+++ b/ejb3/src/main/resources/subsystem-templates/ejb3.xml
@@ -32,7 +32,6 @@
        </timer-service>
        <remote connector-ref="http-remoting-connector" thread-pool-name="default">
            <channel-creation-options>
-               <option name="READ_TIMEOUT" value="${prop.remoting-connector.read.timeout:20}" type="xnio"/>
                <option name="MAX_OUTBOUND_MESSAGES" value="1234" type="remoting"/>
            </channel-creation-options>
        </remote>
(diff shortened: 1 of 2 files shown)


bstansberry (changes requested):
The equivalent change needs to be made in ejb3/src/main/resources/subsystem-templates/ejb3.xml

That file drives what ends up in build-legacy and dist-legacy, which we're still maintaining as an aid in checking the the build output from our Galleon-based builds is doing what we expect.

bstansberry:
@chengfang @tadamski FYI. This change is correct; i.e. other options besides the ones mentioned in the JIRA have no effect so we shouldn't put them in our standard config.

@tmiyargi Can you update the JIRA description with more info why only the listed ones are meaningful? It's not obvious and that will save people researching. AIUI it's because org.jboss.remoting3.remote.RemoteReadListener, the ultimate consumer of the options, only processes the ones you listed.

bstansberry (approved):

tadamski (approved):

PR #12198, opened 2019-04-02
WFLY-9477: avoid registering ROOT.war as default web module more than…

michpetrov:
… once

Issue: [WFLY-9477](https://issues.jboss.org/browse/WFLY-9477)

`default-web-module` is an optional attribute with default value "ROOT.war" but because the module names have to be unique per server it essentially becomes required. Since you can control which host to deploy to in other ways the simplest solution would be to remove the default value. However some users may depend on the default value so I propose to register "ROOT.war" once and then ignore it if it's used by other hosts.


bstansberry:
@fl4via Please review.

fl4via (approved):
This looks good to me

fl4via:
Retest this please

fl4via:
@bstansberry done

PR #12199, opened 2019-04-02
[WFLY-11932] Use HTTPS for repository and pluginRepository URLs

bstansberry:
https://issues.jboss.org/browse/WFLY-11932

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,14 @@
         <!-- Version suffix that is appended to directories. Default is the maven GAV version but this can be edited to use a short form version -->
         <server.output.dir.version>${project.version}</server.output.dir.version>
 
+        <!-- Protocol to use for communication with remote maven repositories.
+             You can set to 'http' if you are using a maven proxy and 'https' 
+             interferes with that. Use 'https' for builds that will be released
+             to non-snapshot public maven repos -->
+        <maven.repository.protocol>https</maven.repository.protocol>
+        <!-- The full remote maven repo URL; can be overridden via -D for special use cases -->
+        <maven.repository.url>${maven.repository.protocol}://repository.jboss.org/nexus/content/groups/public/</maven.repository.url>
+
         <!-- Surefire args -->
         <surefire.extra.args></surefire.extra.args>
         <surefire.jpda.args></surefire.jpda.args>
@@ -6711,7 +6719,7 @@
             </snapshots>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Repository Group</name>
-            <url>https://repository.jboss.org/nexus/content/groups/public/</url>
+            <url>${maven.repository.url}</url>
             <layout>default</layout>
         </repository>
     </repositories>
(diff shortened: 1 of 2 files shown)


jamezp:
Should we change all the links to the licenses to be https as well? For example https://github.com/wildfly/wildfly/blob/7169e62f33e77aeb6fdbcc22031204c1d754026e/feature-pack/src/license/full-feature-pack-licenses.xml#L191.

Maybe in the [`Example_subsystem.adoc`](https://github.com/wildfly/wildfly/blob/c6a417fe18f7219a8c7ac89900662692c5ccfaef/docs/src/main/asciidoc/_extending-wildfly/Example_subsystem.adoc#L20) too.

bstansberry:
@jamezp Sure, I'll change the Example_subsystem; it's somewhat related; might as well give good advice.

@psakar @dosoudil Are files like https://github.com/wildfly/wildfly/blob/7169e62f33e77aeb6fdbcc22031204c1d754026e/feature-pack/src/license/full-feature-pack-licenses.xml relevant to WFLY-11932? The point here is when WildFly is built we want it to use HTTPS to pull down things that end up in our dist. I forget whether the license plugin pulls anything down.

bstansberry:
@jamezp I went ahead and updated the [full|servlet]-feature-pack-licenses.xml files, regardless of my question above (which I'm still curious about.) For all the URLs, https works so might as well use that.

psakar:
@bstansberry licenses are not downloaded during build, they must be already present in the feature pack's folder src/main/resources/content/docs/licenses (eg https://github.com/wildfly/wildfly/tree/master/feature-pack/src/main/resources/content/docs/licenses) 
List of approved licenses (their urls) should be updated first (or procedure checking those urls)

bstansberry:
@psakar Thanks. I'll drop that last commit then as it's not relevant to the purpose of the JIRA.

bstansberry:
@psakar WDYT about the new commit I added here? It's your idea of allowing -D to control this. I made the prop just control the protocol vs the whole URL, figuring that would be easier to use, but the entire URL could be configurable if there's a benefit.

pom.xml
```diff
@@ -173,6 +173,12 @@
         <!-- Version suffix that is appended to directories. Default is the maven GAV version but this can be edited to use a short form version -->
         <server.output.dir.version>${project.version}</server.output.dir.version>
 
+        <!-- Protocol to use for communication with remote maven repositories.
+             You can set to 'http' if you are using a maven proxy and 'https' 
+             interferes with that. Use 'https' for builds that will be released
+             to non-snapshot public maven repos -->
+        <maven.repository.protocol>https</maven.repository.protocol>
```
psakar: I personally think that ability to control whole url is better (gives you chance to use different repo), OTOH for the purpose of being able to use http protocol to be able to use maven proxy this is perfectly valid

bstansberry: I'll add a second property, which uses this one for the protocol part of its default value.

bstansberry: @psakar Done

jamezp (approved):

PR #12200, opened 2019-04-02
[WFLY-11934] Upgrade wildfly-arquillian from 2.1.1.Final to 2.2.0.Fin…

jamezp:
…al. This also requires a shrinkwrap and arquillian-core upgrade to stay in sync with what wildfly-arquillian is using.

https://issues.jboss.org/browse/WFLY-11934

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -312,7 +312,7 @@
         <version.org.javassist>3.23.1-GA</version.org.javassist>
         <version.org.jberet>1.3.3.Final</version.org.jberet>
         <version.org.jboss.activemq.artemis.integration>1.0.2</version.org.jboss.activemq.artemis.integration>
-        <version.org.jboss.arquillian.core>1.1.13.Final</version.org.jboss.arquillian.core>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -333,7 +333,7 @@
         <version.org.jboss.security.jboss-negotiation>3.0.5.Final</version.org.jboss.security.jboss-negotiation>
         <version.org.jboss.security.jbossxacml>2.0.8.Final</version.org.jboss.security.jbossxacml>
         <!-- only needed here until wildfly-arquillian has this version properly synced with arquillian itslef  -->
-        <version.org.jboss.shrinkwrap.descriptors>2.0.0-alpha-10</version.org.jboss.shrinkwrap.descriptors>
+        <version.org.jboss.shrinkwrap.descriptors>2.0.0</version.org.jboss.shrinkwrap.descriptors>
```
bstansberry: What!  Straight to a Final?  How can we trust that it's ready when it moves that quickly!

bstansberry (approved):

PR #12208, opened 2019-04-04
WFLY-11946/WFLY-10340 Remove unneeded dependencies from OpenJPA module

scottmarlow:
https://issues.jboss.org/browse/WFLY-10340 (Enable OpenJPA integration tests for Java 10)
https://issues.jboss.org/browse/WFLY-11946 (Remove unneeded dependencies from OpenJPA module)

As per http://openjpa.apache.org/build-and-runtime-dependencies.html, we don't need the org.apache.commons.lang dependency, as openjpa-all.jar includes that and other dependencies.  

Also updated doc accordingly (current doc is at https://docs.wildfly.org/16/Developer_Guide.html#migrating-from-openjpa).

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
@@ -515,10 +515,10 @@ version="1.0"> +
 [[migrating-from-openjpa]]
 == Migrating from OpenJPA
 
-You need to copy the OpenJPA jars (e.g. openjpa-all.jar serp.jar) into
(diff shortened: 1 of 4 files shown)


bstansberry (approved):

scottmarlow:
@tomjenkinson no problem since we currently use a modular class loader, that provides an isolated classloader for the OpenJPA module.  

If you are migrating a large application to WildFly, rewriting all of your database interaction code to use Hibernate, might not be the first thing that you want to do, while you are evaluating WildFly, instead you can delay that by setting up OpenJPA (or EclipseLink), just like other application servers allow.  If it was really important to you as a user, to not use open-jpa-all.jar, you could include the individual jars instead somehow (e.g. either inline or via other modules that you add or use).

PR #11203, opened 2018-05-11
WFLY-10358 Resource annotations should create bindings even if they a…

stuartwdouglas:
https://issues.jboss.org/browse/WFLY-10358

PR #12217, opened 2019-04-10
WFLY-11870 Do not process resource annotations on abstract classes or …

TomasHofman:
…interfaces that do not have concrete subclasses

Issue: https://issues.jboss.org/browse/WFLY-11870
Downstream issue: https://issues.jboss.org/browse/JBEAP-16576
Related to: https://github.com/wildfly/wildfly/pull/11203

This issue comes from a customer case, they are unable to deploy an application to EAP 7.2, which was deployable on EAP 7.1.

The point is that their deployment contains *abstract* class with @EJB annotated field referencing an interface:

```
public abstract class AbstractReferencingBean { // doesn't have concrete subclass
    @EJB
    private ReferencedBeanInterface referencedBean; // doesn't have implementing class
}
```

Neither the referencing abstract class nor the referenced interface have concrete subclasses in this deployment, so theoretically they should be ignored? The deployment fails because a binding processor can't find any bean that would satisfy this injection point.

In the PR, I'm trying to prevent the resource annotations from being processed if an annotated class is an abstract class or an interface and there's no concrete subclass in the deployment.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
@@ -47,6 +47,9 @@ import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
+import org.jboss.as.server.deployment.annotation.CompositeIndex;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
 import org.jboss.modules.Module;
 import org.jboss.msc.service.CircularDependencyException;
 import org.jboss.msc.service.DuplicateServiceException;
@@ -61,6 +64,7 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 
+import java.lang.reflect.Modifier;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -158,8 +162,9 @@ public class ModuleJndiBindingProcessor implements DeploymentUnitProcessor {
         //were only intended to be installed when running as an app client
         boolean appClient = DeploymentTypeMarker.isType(DeploymentType.APPLICATION_CLIENT, deploymentUnit) || this.appclient;
 
-
         if (!MetadataCompleteMarker.isMetadataComplete(phaseContext.getDeploymentUnit()) && !appClient) {
+            final CompositeIndex index = deploymentUnit.getAttachment(org.jboss.as.server.deployment.Attachments.COMPOSITE_ANNOTATION_INDEX);
+
             for (EEModuleClassDescription config : eeModuleDescription.getClassDescriptions()) {
                 if (handledClasses.contains(config.getClassName())) {
                     continue;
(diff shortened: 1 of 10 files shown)


TomasHofman:
@stuartwdouglas as this is related to https://github.com/wildfly/wildfly/pull/11203, could you review?

TomasHofman:
I'm temporarily closing this PR, before I verify that it's passing the TCK.

TomasHofman:
Passes TCK 8 on my machine.

ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -304,6 +313,20 @@ public static boolean equals(Object one, Object two) {
...
+    }
+
+    private static boolean hasConcreteSubclass(CompositeIndex index, ClassInfo classInfo) {
+        final Set<ClassInfo> subclasses;
+        if (Modifier.isInterface(classInfo.flags())) {
+            subclasses = index.getAllKnownImplementors(classInfo.name());
+        } else {
+            subclasses = index.getAllKnownSubclasses(classInfo.name());
+        }
+        return subclasses.stream().anyMatch(ModuleJndiBindingProcessor::isConcreteClass);
```
bstansberry: Please use a for loop. The class metadata memory cost of a method reference vs a for loop is too high for us to be using this code pattern in server code.

ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -158,6 +162,7 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
         //were only intended to be installed when running as an app client
         boolean appClient = DeploymentTypeMarker.isType(DeploymentType.APPLICATION_CLIENT, deploymentUnit) || this.appclient;
 
+        final CompositeIndex index = deploymentUnit.getAttachment(org.jboss.as.server.deployment.Attachments.COMPOSITE_ANNOTATION_INDEX);
```
bstansberry: This should be inside the if (!MetadatdCompleteMarker... block.

ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -167,6 +172,10 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
                 if(config.isInvalid()) {
                     continue;
                 }
+                ClassInfo classInfo = index.getClassByName(DotName.createSimple(config.getClassName()));
```
bstansberry: To ease future maintenance this needs a nice comment before it explaining why this will work; i.e. an Index contains info on all classes in a jar, and the CompositeIndex in the 'index' var aggregates all the Index objects for the jars visible to the deployment unit. So you can scan it for relationships between classes and not miss out on any.

ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -167,6 +172,10 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
+                if (!isConcreteClass(classInfo) && !hasConcreteSubclass(index, classInfo)) {
+                    continue;
+                }
                 final Set<BindingConfiguration> classLevelBindings = new HashSet<>(config.getBindingConfigurations());
```
bstansberry: I recommend moving L175-178 below this line and wrapping it in an  'if (!classLevelBindings.isEmpty())' check. The l175-178 scan is not cheap so it should be avoided whenever possible.

bstansberry (changes requested):
Please see comments above.

Thanks for checking this using the TCK!

TomasHofman:
I updated according to your comments, thanks for input!

ivassile:
@bstansberry We'll include this fix in EAP 7.2.2 if merged. Can you review the changes? Thanks!

bstansberry (approved):

PR #11086, opened 2018-04-03
[WFLY-10146] CMTTxInterceptor cleanups, part 1

dmlloyd:
https://issues.jboss.org/browse/WFLY-10146

PR #12225, opened 2019-04-12
[WFLY-10351] BMT interceptor timeout reset adjustment

ochaloup:
https://issues.jboss.org/browse/WFLY-10351

Adding correct transaction timeout handling for propagate the reset of the timeout to the WFTC and not to Narayana. Adding tests on fact that the method [setTransactionTimeout()](https://docs.oracle.com/javaee/7/api/javax/transaction/UserTransaction.html#setTransactionTimeout-int-) sets the timeout value for thread being active.

@dmlloyd would you be so kind and check if this follows the rationale the issue was created with and how the [WFLY-10146](https://issues.jboss.org/browse/WFLY-10146) was treated. Thanks.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/tx/BMTInterceptor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/tx/BMTInterceptor.java
@@ -24,12 +24,11 @@ package org.jboss.as.ejb3.tx;
 import javax.ejb.EJBException;
 import javax.transaction.SystemException;
 import javax.transaction.Transaction;
-import javax.transaction.TransactionManager;
 
 import org.jboss.as.ejb3.component.EJBComponent;
 import org.jboss.invocation.Interceptor;
 import org.jboss.invocation.InterceptorContext;
-import org.jboss.tm.TransactionTimeoutConfiguration;
+import org.wildfly.transaction.client.ContextTransactionManager;
 
 /**
  * Suspend an incoming tx.
@@ -50,8 +49,8 @@ public abstract class BMTInterceptor implements Interceptor {
 
     @Override
     public Object processInvocation(final InterceptorContext context) throws Exception {
-        TransactionManager tm = component.getTransactionManager();
-        int oldTimeout = getCurrentTransactionTimeout(component);
+        final ContextTransactionManager tm = ContextTransactionManager.getInstance();
+        final int oldTimeout = tm.getTransactionTimeout();
         try {
             Transaction oldTx = tm.suspend();
             try {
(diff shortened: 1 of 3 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/tx/BMTInterceptor.java
```diff
@@ -60,7 +59,7 @@ public Object processInvocation(final InterceptorContext context) throws Excepti
                 if (oldTx != null) tm.resume(oldTx);
             }
         } finally {
-            tm.setTransactionTimeout(oldTimeout == -1 ? 0 : oldTimeout);
+            tm.setTransactionTimeout(oldTimeout == ContextTransactionManager.getGlobalDefaultTransactionTimeout() ? 0 : oldTimeout);
```
bstansberry: The WFLY-10146 change included the following comment before this line, which I'd like to see here as well as it helps clarify what's otherwise pretty confusing:

// See also https://issues.jboss.org/browse/WFTC-44

The part of the WFLY-10146 change adding that comment:

https://github.com/wildfly/wildfly/pull/11086/files#diff-e05009bda935bba48b4ee7ca633ba1ecR170

The rest of the changes in this class all LGTM; they are just changing to standard ContextTransactionManager use. I think this line seems fine too (other than needing the comment) but checking with David or @chengfang or @tadamski makes sense.

dmlloyd: I agree with Brian, adding the comment would be a good idea.  Otherwise IIRC this seems correct.

ochaloup:
@bstansberry @dmlloyd thanks. I added the comment `// See also https://issues.jboss.org/browse/WFTC-44` as proposed.

bstansberry (approved):

ejb3/src/main/java/org/jboss/as/ejb3/tx/BMTInterceptor.java
kabir: @ochaloup Please re-add the comment :)

ochaloup: @kabir just I'm now not sure what you are talking about. I added the comment `// See also https://issues.jboss.org/browse/WFTC-44` above the change of the `tm.setTransactionTimeout(...)`. I can see the change is still at the place if I take a look to the PR "Files changes". Could be that only the github does not show this as "outdated" as I did not do any change on the highlighted line but above it.

kabir: @ochaloup sorry, you are right

PR #12226, opened 2019-04-12
[WFLY-11985] Upgrade Commons Codec to 1.11

ivassile:
https://issues.jboss.org/browse/WFLY-11985

Upgrading Commons Codec to 1.11

Release notes: https://commons.apache.org/proper/commons-codec/changes-report.html#a1.11
Source diff:  https://github.com/apache/commons-codec/compare/1.10...commons-codec-1.11
The pom only has test dependencies, so this should not drive a requirement for upgrading other dependents.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -237,7 +237,7 @@
         <version.com.sun.xml.fastinfoset>1.2.13</version.com.sun.xml.fastinfoset>
         <version.commons-beanutils>1.9.3</version.commons-beanutils>
         <version.commons-cli>1.3.1</version.commons-cli>
-        <version.commons-codec>1.10</version.commons-codec>
+        <version.commons-codec>1.11</version.commons-codec>
         <version.commons-collections>3.2.2</version.commons-collections>
         <version.commons-digester>1.8.1</version.commons-digester>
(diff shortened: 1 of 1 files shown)


bstansberry:
@jimma @ronsigal @fjuma @yrodiere Any objections?

Primary uses are CXF, RESTEasy, Lucene, MyFaces, com.sun.jsf-impl.

See the description for release notes and source diff.  This is a minor version upgrade.

The pom only has test dependencies.

yrodiere:
@bstansberry Lucene does not depend on commons-codec as far as I can tell. In fact it doesn't depend on anything.

Could you please point me what made you think otherwise? I've been wrong before.

The only way this could impact Hibernate Search is through the optional Apache Tika dependency, or the Elasticsearch integration, but as far as I know none of them is included in WildFly.

jimma:
@bstansberry   This upgrade is OK for CXF.

fjuma:
This upgrade is ok for JSF.

bstansberry:
@yrodiere My apologies!  I was grepping our module.xml files looking for uses and wasn't careful enough.  The lucene module has the completely irrelevant... 

artifact name="${org.apache.lucene:lucene-backward-codecs}"

PR #12237, opened 2019-04-18
WFLY-12006 EJB3 IIOP attributes enable-by-default and use-qualified-name should be required

soul2zimate:
https://issues.jboss.org/browse/WFLY-12006

Update attributes enable-by-default and use-qualified-name to required.
Because enable-by-default and use-qualified-name must be required during parsing at https://github.com/wildfly/wildfly/blob/16.0.0.Final/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3Subsystem12Parser.java#L227

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3IIOPResourceDefinition.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3IIOPResourceDefinition.java
@@ -54,15 +54,17 @@ public class EJB3IIOPResourceDefinition extends SimpleResourceDefinition {
 
 
     static final SimpleAttributeDefinition USE_QUALIFIED_NAME =
-            new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.USE_QUALIFIED_NAME, ModelType.BOOLEAN, true)
+            new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.USE_QUALIFIED_NAME, ModelType.BOOLEAN)
                     .setAllowExpression(true)
                     .setFlags(AttributeAccess.Flag.RESTART_NONE)
+                    .setRequired(true)
                     .build();
 
     static final SimpleAttributeDefinition ENABLE_BY_DEFAULT =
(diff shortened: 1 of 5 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3IIOPResourceDefinition.java
```diff
@@ -57,12 +57,14 @@
             new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.USE_QUALIFIED_NAME, ModelType.BOOLEAN, true)
                     .setAllowExpression(true)
                     .setFlags(AttributeAccess.Flag.RESTART_NONE)
+                    .setRequired(true)
```
bstansberry: Please remove the 'true' from L57 as that says the opposite of this setRequired(true).

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3IIOPResourceDefinition.java
```diff
@@ -57,12 +57,14 @@
...
                     .build();
 
     static final SimpleAttributeDefinition ENABLE_BY_DEFAULT =
             new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.ENABLE_BY_DEFAULT, ModelType.BOOLEAN, true)
                     .setAllowExpression(true)
                     .setFlags(AttributeAccess.Flag.RESTART_NONE)
+                    .setRequired(true)
```
bstansberry: Please remove the 'true' from L64 as that says the opposite of this setRequired(true).

bstansberry (changes requested):
EJB3IIOPAdd can be cleaned up a bit too; in performBoottime when it reads these values it should assign to 'boolean' variables, not 'Boolean'.

Technically the EJB subsystem management API version should be bumped for this, but, meh. I think it was never possible to effectively not define these if the resource was added.

soul2zimate:
Thanks, I have updated as per review request. I also removed opposite optional value "true" in three other classes.

bstansberry (approved):

PR #12261, opened 2019-05-07
WFLY-12052 Don't evict SFSBs from memory that fail to passivate eagerly.

pferraro:
https://issues.jboss.org/browse/WFLY-12052


bstansberry (approved):

bmaxwell:
I think we can close this, as the change WFLY-12064 already merged along with user application throwing an exception in PrePassivate achieves the same result without having to hack it in wildfly

PR #12263, opened 2019-05-07
[WFLY-12050] Don't try to get metrics from a UndertowDeploymentServic…

bstansberry:
…e whose ServiceController is not UP

https://issues.jboss.org/browse/WFLY-12050


bstansberry:
retest this please

pferraro (approved):

philippejulien:
There is another condition that needs to have the same change applied on line 134 otherwise the fix won't work.

bstansberry:
@philippejulien Thanks. I added that.

I meant to put a 'hold' label on this as it really needs a test added.

PR #12264, opened 2019-05-08
[WFLY-11031] Introduce a console-access-log for web access logging

jamezp:
https://issues.jboss.org/browse/WFLY-11031

**Note:** This PR currently requires a SNAPSHOT build of WildFly Core until Beta5 is release.

@fl4via Can you please have a look at this as I went a slight different route than what we discussed previously. I added a `setting=console-access-log` resource instead of keeping everything under the `setting=access-log` resource. I felt it made the resource too complicated.

There had been some email discussions with using the standard `pattern` attribute for determining the exchange attributes to include in the JSON structured output. While I realize these are well known patterns I felt it just didn't fit right when we need the ability for defined keys in structured output. I went with a complex `attributes` management model attribute with known exchange attribute types.

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow.adoc
@@ -249,6 +249,52 @@ be matched
 serve up requests that do not match anything.
 |=======================================================================
 
+==== Console Access Logging
+
+Each host allows for access logging to the console which writes structured data in JSON format. This only writes to
+`stdout` and is a single line of JSON structured data.
+
+The `attributes` management model attribute is used to determine which exchange attributes should be logged. This is
+similar to the `pattern` used for traditional access logging. The main difference being since the data is structured
+the ability to use defined keys is essential.
+
+A `metadata` attribute also exists which allows extra metadata to be added to the output. The value of the attribute is
+a set of arbitrary key/value pairs. The values can include management model expressions, which must be resolvable when
+the console access log service is started. The value is resolved once per start or reload of the server.
+
+
+===== CLI Examples
+
+.add-console-access-logging.cli
+----
+/subsystem=undertow/server=default-server/host=default-host/setting=console-access-log:add
+----
+
+.complex-add-console-access-logging.cli
+----
+/subsystem=undertow/server=default-server/host=default-host/setting=console-access-log:add(metadata={"@version"="1", "qualifiedHostName"=${jboss.qualified.host.name:unknown}}, attributes={bytes-sent={}, date-time={key="@timestamp", date-format="yyyy-MM-dd'T'HH:mm:ssSSS"}, remote-host={}, request-line={}, response-header={key-prefix="responseHeader", names=["Content-Type"]}, response-code={}, remote-user={}})
+----
+
+[source,json]
+----
+{
+    "eventSource":"web-access",
+    "hostName":"default-host",
+    "@version":"1",
+    "qualifiedHostName":"localhost.localdomain",
+    "bytesSent":1504,
+    "@timestamp":"2019-05-02T11:57:37123",
+    "remoteHost":"127.0.0.1",
+    "remoteUser":null,
+    "requestLine":"GET / HTTP/2.0",
+    "responseCode":200,
+    "responseHeaderContent-Type":"text/html"
+}
+----
+
+NOTE: The above JSON is formatted only for readability. The output will be on a single line.
+
+
 [[servlet-container-configuration]]
 == Servlet container configuration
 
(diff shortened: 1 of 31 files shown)


jmesnil:
@jamezp #12266 contains the Core 9.0.0.Beta5 upgrade that is required by this PR.

undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
```diff
@@ -108,6 +108,8 @@ undertow.ajp-listener.remove=Removes an AJP listener
 undertow.access-log=The access log configuration for this virtual server.
 undertow.access-log.add=Add access log configuration to this virtual server.
 undertow.access-log.remove=Remove access log configuration from this virtual server.
+undertow.access-log.log-to-file=If the access log messages should be written to the file. Note this attribute is \
+  mutually exclusive with the use-server-log attribute.
```
jstourac: Is this really used? I don't see it as an attribute in `access-log` element.

jamezp: Nope, good catch. I'll get that removed thanks!

jstourac (approved):

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow.adoc
```diff
@@ -249,6 +249,49 @@ be matched
...
+==== Console Access Logging
+
+Each host allows for access logging to the console which writes structured
+data in JSON format. This only writes to `stdout` and is a single line of
+JSON structured data.
+
+The `attributes` management model attribute is used to determine which exchange attributes should be logged. This is
+similar to the `pattern` used for traditional access logging. The main difference being since the data is structured
+the ability to use defined keys is essential.
+
```
bstansberry: There should also be a paragraph like...

The `metadata` management model attribute can be used to add extra metadata to each log record. The value of the attribute is an arbitrary set of key/value pairs. The values can include management model expressions, which must be resolvable when the console access log service is started. The value is resolved once per start or reload of the server.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/access/log/ConsoleAccessLogTestCase.java
```diff
@@ -0,0 +1,432 @@
...
+                CloseableHttpResponse response = httpClient.execute(request)
+        ) {
+            Assert.assertEquals("Failed to access " + uri, HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+        }
+    }
+
+    private Collection<JsonObject> findLines() throws InterruptedException {
+        // Note this could be a potential spot for a race in the test validation. The console-access-log is
+        // asynchronous so we need to wait to ensure it's fully written to the console.
+        TimeUnit.SECONDS.sleep(TimeoutUtil.adjust(1));
```
fl4via: @jamezp I'm okay with having this, but lets keep an eye to see if we don't have random failures on this test when run on slower machines. If that happens, we will have to switch to an assertion in loop: if fail to find expected lines-> try to find lines again, unless timeout of, say, 30 s or 60 s has passed.

fl4via (approved):

fl4via (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/access/log/ConsoleAccessLogTestCase.java
jamezp: +1. I do not generally like things like this, but there seemed to be no other way :)

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow.adoc
```diff
@@ -249,6 +249,52 @@ be matched
...
+Each host allows for access logging to the console which writes structured data in JSON format. This only writes to
+`stdout` and is a single line of JSON structured data.
+
+The `attributes` management model attribute is used to determine which exchange attributes should be logged. This is
+similar to the `pattern` used for traditional access logging. The main difference being since the data is structured
+the ability to use defined keys is essential.
+
+A `metadata` attribute also exists which allows extra metadata to be added to the output. The value of the attribute is
+an arbitrary key/value pair. The values can include management model expressions, which must be resolvable when the
```
bstansberry: I'm not going to block merging for this as I want it in but the value is a *set* of arbitrary key/value pairs, not a single pair. The CLI example is a set.

jamezp: Good catch. I went ahead and fixed that.

jamezp:
It looks like I should not have pushed the last change I did. I didn't notice #12300 and only looked at master-ignore by mistake. This one may need to be manually close and I apologize for that. Once #12300 is merged I'll make a new PR for the text update on the documentation.

bstansberry:
It's fine; I'll merge this. I'm not merging #12300; it's just for CI.  But the only diff here vs what I tested is in the asciidoc.

PR #12270, opened 2019-05-10
[WFLY-12058] Deprecated WeldDeploymentMarker and WeldPortableExtensions functionality not working as expected

yersan:
The changes in this PR correct https://github.com/wildfly/wildfly/pull/11802 to enable again the behavior of WeldDeploymentMarker and WeldPortableExtensions, which were deprecated in that PR.

I haven't found a simple way to get the deprecated classes working together with the new versions, which were moved to a _private package in a different module. 

The proposed solution is to keep accessing to the old versions from the new structures (_private/WeldDeploymentMarker.java and WeldCapabilityImpl.java) and when we are ready to delete the old ones, move the behavior of WeldDeploymentMarker and WeldPortableExtensions to the  _private package.

Jira issue: https://issues.jboss.org/browse/WFLY-12058


emmartins:
@yersan so with this PR the test I added in JIRA passes, and besides that a 3rd party deployment processor with similar code registering a CDI portable extension, yet not in weld subsystem, would work too?

yersan:
Yes @emmartins , it must work if the 3rd party deployment processor is using the deprecated classes. 

If the 3rd party deployment processor wants to get rid of the deprecated files, something they should do, they have to use the capability support to get access to the API exposed by the deprecated ones.

emmartins:
@yersan sounds good to me, thanks

bstansberry (approved):

PR #12276, opened 2019-05-11
[WFLY-12067] Reorganize pom properties to clearly segregate props sol…

bstansberry:
…y related to plugin dependency versions.

No property values are changed.

https://issues.jboss.org/browse/WFLY-12067


darranl:
For a change like this might be good for you to merge you own PR at the point of tagging, seems the kind of thing to trigger a merge conflict (Especially as GitHub conflict identification is broken)

darranl (approved):
Adding an approval, as in my comment I think an ideal change to merge at the point of tagging.

PR #12278, opened 2019-05-13
[WFLY-11848] Tests if JNDI bindings is correctly built in case there is no appName.

dcihak:
JIRA issue: [WFLY-11848](https://issues.jboss.org/browse/WFLY-11848)
Tests if JNDI bindings is correctly built in case there is no appName.


jamezp:
I'll leave this open for another opinion, but I don't think it's a good idea to read the log file to determine if this is working or not. If for whatever reason the log file is changed or turned off this test breaks. Plus the log file could be huge resulting in multiple lines be read into an array.

bstansberry:
I vote for manual verification.

darranl:
@dcihak It looks like the discussion didn't support the approach taken to read the log file, can this PR be closed an re-opened if an alternative test can be devised?

bstansberry:
Closing in favor of #13180, which adapts this to use the utilities we have for this kind of testing.

PR #12283, opened 2019-05-14
[WFLY-11959] AbstractSimpleApplicationClientTestCase pollutes the build/target/ installation

lukasburda:
JIRA: https://issues.jboss.org/browse/WFLY-11959
- Change distribution which basic tests used and polluted to one which is created especially for them

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -206,7 +206,7 @@
 
                <!--
                    A build of target/wildfly which is shared by all modules.
-                   Modules and bundles are not copied as they are read-only (see surefire props).
+                   Bundles are not copied as they are read-only (see surefire props).
                -->
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
@@ -231,7 +231,6 @@
                                            <exclude>bin/*.bat</exclude>
                                            <exclude>bin/*.ps1</exclude>
                                            <exclude>docs/</exclude>
-                                           <exclude>modules/</exclude>
                                            <exclude>welcome-content/*.png</exclude>
                                            <exclude>standalone/data</exclude>
                                            <exclude>standalone/log</exclude>
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -34,6 +34,10 @@
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}/..</jbossas.project.dir>
         <surefire.forked.process.timeout>3600</surefire.forked.process.timeout>
+
+        <!-- Used to provide an absolute location for the distribution under test. -->
+        <jboss.dist>${project.build.directory}/wildfly</jboss.dist>
+        <jboss.home>${jbossas.project.dir}/${wildfly.build.output.dir}</jboss.home>
     </properties>
 
     <dependencies>


kabir:
This is ok to test

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/basic/AbstractSimpleApplicationClientTestCase.java
```diff
@@ -85,4 +87,22 @@ public void testAppClientJBossDescriptor() throws Exception {
             wrapper.quit();
         }
     }
+
+    @AfterClass
+    public static void afterClass() {
+        String appClientDir = System.getProperty("jboss.dist") + File.separator + "appclient";
+        deleteDirectory(new File(appClientDir + File.separator + "tmp"));
+        deleteDirectory(new File(appClientDir + File.separator + "data"));
+    }
```
jamezp: I may be misunderstanding the JIRA, but I don't think we want to do this. It's possible there is useful data in here if tests fail.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/util/AppClientWrapper.java
```diff
@@ -201,6 +201,7 @@ private String getAppClientCommand() throws Exception {
                 " org.jboss.as.appclient" +
                 " -Djboss.server.base.dir="+ asDist + "/appclient" +
                 " -Djboss.home.dir="+ asDist +
+                " -Djboss.server.log.dir=" + System.getProperty("jboss.server.log.dir") +
```
jamezp: I don't think we need this property. However we would need the following two:
```
"-Dorg.jboss.boot.log.file=" + asDist + "/appclient/log/appclient.log" +
"-Dlogging.configuration=file:" + asDist + "/appclient/configuration/logging.properties" +
```

testsuite/integration/pom.xml
```diff
@@ -354,6 +354,7 @@
                         <jbossas.ts.dir>${jbossas.ts.dir}</jbossas.ts.dir>
                         <jbossas.project.dir>${jbossas.project.dir}</jbossas.project.dir>
                         <jboss.dist>${jboss.dist}</jboss.dist>
+                        <jboss.server.log.dir>${basedir}/target/server-logs</jboss.server.log.dir>
```
jamezp: Given the above changes we would likely not need this.

jamezp (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/basic/AbstractSimpleApplicationClientTestCase.java
bstansberry: If what the PR is doing is cleaning up data in build/target/wildfly-xxx, that's not what I'm looking for. There should be nothing written there.

Ideally the test doesn't deal with that directory at all.

lukasburda:
Thanks for the review. I will solve that recently.

lukasburda:
retest this please

testsuite/pom.xml
```diff
@@ -231,7 +231,6 @@
                                            <exclude>bin/*.bat</exclude>
                                            <exclude>bin/*.ps1</exclude>
                                            <exclude>docs/</exclude>
-                                           <exclude>modules/</exclude>
```
bstansberry: We can't do this; copying the modules all over puts too much IO strain on the CI servers.

bstansberry (changes requested):
Perhaps the thing to do here is to leave jboss.dist as is and use the jboss.inst property in AppClientWrapper for paths that are meant to be under testsuite/integaration/basic/target/wildfly. Use jboss.inst for those that are meant to be under build/target/wildfly-xxx.

" -Djboss.modules.dir="+ asInst + "/modules" +


" -mp "+ asInst + "/modules" +

darranl:
This PR seems to have stalled since June so I am going to close it for now, please feel free to re-open if you are ready to continue or submit a PR based on the comments so far.

PR #12289, opened 2019-05-14
 WFLY-12020 Refresh Infinispan cache metrics

pferraro:
https://issues.jboss.org/browse/WFLY-12020

Move cache metrics to runtime cache resources, leveraging AttributeTranslations to translate old registration paths to new paths.
Add missing metrics, deprecated obsolete ones.


pferraro:
Because this didn't make it into WF17, it now needs transformers.

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/CacheContainerMetricExecutor.java
```diff
@@ -29,16 +30,15 @@
...
  *
  * @author Paul Ferraro
  */
 public class CacheContainerMetricExecutor implements MetricExecutor<EmbeddedCacheManager> {
 
     @Override
     public ModelNode execute(OperationContext context, Metric<EmbeddedCacheManager> metric) throws OperationFailedException {
-        String containerName = context.getCurrentAddressValue();
-        CacheContainer container = new PassiveServiceSupplier<CacheContainer>(context.getServiceRegistry(true), InfinispanRequirement.CONTAINER.getServiceName(context, containerName)).get();
+        CacheContainer container = new PassiveServiceSupplier<CacheContainer>(context.getServiceRegistry(true), InfinispanRequirement.CONTAINER.getServiceName(context, UnaryCapabilityNameResolver.DEFAULT)).get();
```
bstansberry: context.getServiceRegistry(true)

That 'true' is nasty because it means what should be a read requires the exclusive lock.

I see that 'true' in a number of places in this PR, and its not a new thing. I'm just kind of randomly choosing this one to comment on.

We might need to make an exception to the rules for this PassiveServiceSupplier thing, although that worries me.

pferraro: We use OperationContext.getServiceRegistry(true) because we are installing (and eventually removing) a temporary service that requires the target service so that we can obtain the requisite Supplier<T>.  If there is a better way to do this, I'm all ears.

pferraro: OK - I've updated this PR to use OperationContext.getServiceRegistry(false) where appropriate.

bstansberry:
retest this please

clustering/common/src/main/java/org/jboss/as/clustering/controller/ReadAttributeTranslationHandler.java
```diff
@@ -48,13 +49,24 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
         OperationStepHandler readTranslatedAttributeHandler = new ReadTranslatedAttributeStepHandler(readAttributeHandler, targetAttribute, this.translation.getReadTranslator());
         // If targetOperation applies to the current resource, we can execute in the current step
         if (targetAddress == currentAddress) {
             readTranslatedAttributeHandler.execute(context, targetOperation);
         } else {
+            if (registration.isRuntimeOnly()) {
+                try {
+                    context.readResourceFromRoot(targetAddress, false);
+                } catch (Resource.NoSuchResourceException ignore) {
+                    // If the target runtime resource does not exist return UNDEFINED
```
bstansberry: @pferraro Why?

pferraro: This is a generic operation handler, used for deprecated attributes/metrics whose return value comes from a different (i.e. target) attribute, potentially at a different path address.  If the target attribute resides at a different path address, and the resource at the target address is a runtime resource, we need to handle the case where no resource exists at the target address, otherwise the subsequent OperationContext.addStep(...) would fail to execute.
If you consider the previous behavior of a given deprecated metric, the "no resource exists" condition is equivalent to the case where the target capability's service (from which the metric is obtained) is not started.  In this case, we would normally return a value of UNDEFINED.  When the target resource is a runtime resource, that runtime resource only exists when the target capability's service (from which the metric is obtained) is started - thus the result of the read-attribute translation is consistent with the previous behavior.

bstansberry: OK. The previous behavior sounds a bit off but I think it's the way things are handled in many places (not just clustering) so changing that is out of scope for this PR.

bstansberry (approved):

PR #12295, opened 2019-05-15
WFLY-12071: use management name for deployment address

michpetrov:
Issue: [WFLY-12071](https://issues.jboss.org/browse/WFLY-12071)

Deployments are stored under their management name, not the file name. When the two are different metrics cannot be retrieved.

### DIFF
--- a/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/deployment/DeploymentMetricService.java
+++ b/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/deployment/DeploymentMetricService.java
@@ -8,6 +8,7 @@ import java.util.function.Supplier;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentCompleteServiceProcessor;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.msc.Service;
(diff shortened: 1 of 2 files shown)


jamezp:
These failures definitely seem related. I'm not sure, for compatibility purposes, this will work.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/deployment/DeploymentMetricService.java
```diff
@@ -65,9 +66,9 @@ public void stop(StopContext stopContext) {
 
     private static PathAddress createDeploymentAddressPrefix(DeploymentUnit deploymentUnit) {
         if (deploymentUnit.getParent() == null) {
-            return PathAddress.pathAddress(DEPLOYMENT, deploymentUnit.getName());
+            return PathAddress.pathAddress(DEPLOYMENT, deploymentUnit.getAttachment(Attachments.MANAGEMENT_NAME));
         } else {
-            return createDeploymentAddressPrefix(deploymentUnit.getParent()).append(SUBDEPLOYMENT, deploymentUnit.getName());
+            return createDeploymentAddressPrefix(deploymentUnit.getParent()).append(SUBDEPLOYMENT, deploymentUnit.getAttachment(Attachments.MANAGEMENT_NAME));
```
bstansberry: I'm pretty sure removing the change to this line will fix the problem.

I don't think a subdeployment has this attachment, and in any case a subdeployment never has a management name that is different from deploymentUnit.getName(). A user can't assign a management name to a subdeployment as it's an internal part of the top level deployment, not something independently added to the server config.

bstansberry:
@jamezp The problem he's fixing here is definitely something that needs fixing. The code generates a management model address and if a runtime name is used and it's different from the management name, that will be an invalid address.

Whether an external consumer of metrics would want the runtime name as part of a metric label rather than the management name is a good question. The general reason they'd differ would be if the mgmt name were foo.war_v2 or something. That's a point for @jmesnil but in any case it seems trying to use runtime name doesn't work.

bstansberry:
@michpetrov This needs a test case.

jamezp:
That makes sense @bstansberry and I meant to tag @jmesnil to have him take a look and pressed the button too soon.

bstansberry (changes requested):
Please add a test case.

michpetrov:
@bstansberry to what extent should this be tested? is "is this running" enough or should I copy e.g. [MicroProfileMetricsApplicationTestCase](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsApplicationTestCase.java)?

bstansberry (approved):

PR #12296, opened 2019-05-15
Fix for WFLY-12080, Upgrade galleon-plugins to 4.0.1

jfdenise:
Fix for: https://issues.jboss.org/browse/WFLY-12080
- Upgrade galleon wildfly plugins to 4.0.1.
- Make generate-all-artifacts-list plugin to run in offline mode
- Added needed dependencies not resolved during build.

Galleon changes:
https://github.com/wildfly/galleon-plugins/compare/4.0.0.Final...4.0.1.Final

PR #12299, opened 2019-05-15
Fix for WFLY-12080, Upgrade galleon-plugins to 4.0.1

bstansberry:
This is #12296 but added on top of #12290 and with the conflict between the two resolved.

Fix for: https://issues.jboss.org/browse/WFLY-12080

Upgrade galleon wildfly plugins to 4.0.1.
Make generate-all-artifacts-list plugin to run in offline mode
Added needed dependencies not resolved during build.
Galleon changes:
wildfly/galleon-plugins@4.0.0.Final...4.0.1.Final

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -202,12 +202,35 @@
                                 </goals>
                                 <phase>package</phase>
                                 <configuration>
+                                    <offline>true</offline>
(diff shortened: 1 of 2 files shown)


jamezp (approved):

jamezp:
Assuming CI passes, and I don't see why it won't, this looks good to me.

PR #12301, opened 2019-05-16
[WFLY-12085]: java.net.URISyntaxException: Illegal character in opaque part at index 7: file:C:\Java\jboss\jboss-as\standalone\configuration/logging.properties

ehsavoie:
* Setting the configurationFileRefreshPeriod to -1 to avoid reloading of
logging.properties

Jira: https://issues.jboss.org/browse/WFLY-12085


jamezp:
I placed a hold on this based on a [JIRA comment](https://issues.jboss.org/browse/WFLY-12085?focusedCommentId=13734584&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13734584). If we decide we should likely file a new JIRA.

bstansberry:
@ehsavoie This looks like a right thing to do, but perhaps a different JIRA? As @jamezp says it won't prevent the WFLY-12085 problem.

PR #12303, opened 2019-05-17
WFLY-12086 Hibernate 2lc should be enabled if Hibernate cache properties are enabled and shared cache mode is not set to NONE

scottmarlow:
https://issues.jboss.org/browse/JBEAP-16901
https://issues.jboss.org/browse/WFLY-12086

HibernatePersistenceProviderAdaptor.addProviderDependencies() should be easier to understand now and has been changed to state the precedence of different cache settings (Hibernate cache properties versus SharedCacheMode).  

Also corrected handling of SharedCacheMode.UNSPECIFIED, to explicitly tell Hibernate to use SharedCacheMode.ENABLE_SELECTIVE (Hibernate treats SharedCacheMode.UNSPECIFIED as caching being disabled, which is not what the JPA container expects, so the JPA container changes UNSPECIFIED to SharedCacheMode.ENABLE_SELECTIVE).

### DIFF
--- a/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
+++ b/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
@@ -99,30 +99,73 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
         final Properties properties = pu.getProperties();
         final String sharedCacheMode = properties.getProperty(SHARED_CACHE_MODE);
 
-        if ( Classification.NONE.equals(platform.defaultCacheClassification())) {
+        if (Classification.NONE.equals(platform.defaultCacheClassification())) {
             JPA_LOGGER.tracef("second level cache is not supported in platform, ignoring shared cache mode");
             pu.setSharedCacheMode(SharedCacheMode.NONE);
         }
-        // check if 2lc is explicitly disabled which takes precedence over other settings
-        boolean sharedCacheDisabled = SharedCacheMode.NONE.equals(pu.getSharedCacheMode())
-                ||
-                NONE.equals(sharedCacheMode) ||
-                SharedCacheMode.UNSPECIFIED.equals(pu.getSharedCacheMode()) ||
-                UNSPECIFIED.equals(sharedCacheMode);
-        if (!sharedCacheDisabled &&
-                (null == properties.getProperty(AvailableSettings.USE_SECOND_LEVEL_CACHE) ||
-                        Boolean.parseBoolean(properties.getProperty(AvailableSettings.USE_SECOND_LEVEL_CACHE))))
-                {
+        // precedence order of cache settings (1 overrides other settings and 3 is lowest precedence level):
+        // 1 - SharedCacheMode.NONE
+        // 2 - AvailableSettings.USE_SECOND_LEVEL_CACHE
+        // 2 - AvailableSettings.USE_QUERY_CACHE
+        // 3 - SharedCacheMode.UNSPECIFIED
+        // 3 - SharedCacheMode.ENABLE_SELECTIVE
+        // 3 - SharedCacheMode.DISABLE_SELECTIVE
+
+        // if SharedCacheMode.NONE, set cacheDisabled to true.
+        boolean cacheDisabled = noneCacheMode(pu)
+        // Or if Hibernate cache settings are specified and Hibernate settings indicate cache is disabled, set cacheDisabled to true.
+                || (haveHibernateCachePropertyDefined(pu) && !hibernateCacheEnabled(pu));
+
+        if (!cacheDisabled) {
             HibernateSecondLevelCache.addSecondLevelCacheDependencies(pu.getProperties(), pu.getScopedPersistenceUnitName());
             JPA_LOGGER.tracef("second level cache enabled for %s", pu.getScopedPersistenceUnitName());
+            // for SharedCacheMode.UNSPECIFIED, enable the cache and enable caching for entities marked with Cacheable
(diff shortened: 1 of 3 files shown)


scottmarlow:
linux elytron 2 failures (in org.jboss.as.test.xts.suspend.wsba.BusinessActivitySuspendTestCase) don't seem related to this change.

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
```diff
@@ -99,32 +99,62 @@ public void addProviderDependencies(PersistenceUnitMetadata pu) {
...
-                UNSPECIFIED.equals(sharedCacheMode);
-        if (!sharedCacheDisabled &&
-                (null == properties.getProperty(AvailableSettings.USE_SECOND_LEVEL_CACHE) ||
-                        Boolean.parseBoolean(properties.getProperty(AvailableSettings.USE_SECOND_LEVEL_CACHE))))
-                {
+        // precedence order of cache settings (1 overrides other settings and 3 is lowest precedence level):
+        // 1 - SharedCacheMode.NONE
+        // 2 - AvailableSettings.USE_SECOND_LEVEL_CACHE
+        // 2 - AvailableSettings.USE_QUERY_CACHE
+        // 3 - SharedCacheMode.UNSPECIFIED =
```
bstansberry: Does that '=' belong there?  Trying to review this it made me think L111/112 were somehow continuations of L110 but I don't think that's the case.

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
```diff
@@ -99,32 +99,62 @@ public void addProviderDependencies(PersistenceUnitMetadata pu) {
...
+        // 3 - SharedCacheMode.ENABLE_SELECTIVE
+        // 3 - SharedCacheMode.DISABLE_SELECTIVE
+
+        // if SharedCacheMode.NONE, disable cache.
+        // if Hibernate cache settings are set to false, disable cache
+        boolean cacheDisabled = noneCacheMode(pu) || (haveHibernateCacheProperty(pu) && !hibernateCacheEnabled(pu));
```
bstansberry: The comment says SharedCacheMode.UNSPECIFIED is lower precedence than vailableSettings.USE_SECOND_LEVEL_CACHE/USE_QUERY_CACHE but this logic gives it higher precedence.

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
```diff
@@ -99,32 +99,62 @@ public void addProviderDependencies(PersistenceUnitMetadata pu) {
...
                     pu.getSharedCacheMode().toString());
-            pu.setSharedCacheMode(SharedCacheMode.NONE);  // ensure that Hibernate doesn't try to use the 2lc for UNSPECIFIED
+            pu.setSharedCacheMode(SharedCacheMode.NONE);  // ensure that Hibernate doesn't try to use the 2lc
         }
     }
 
+    private boolean haveHibernateCacheProperty(PersistenceUnitMetadata pu) {
+        return (pu.getProperties().getProperty(AvailableSettings.USE_SECOND_LEVEL_CACHE) != null ||
+                pu.getProperties().getProperty(AvailableSettings.USE_QUERY_CACHE) != null);
+    }
```
bstansberry: I don't think this method is need as hibernateCacheEnabled has to deal with null anyway (and does) and it would return false if both props are null, the same as this method does.

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
scottmarlow: oops, good catch, removing...

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
scottmarlow: noneCacheMode is the precedent level 1, the highest.  You found a bug that I missed, thanks!  Changing from:

`    private boolean noneCacheMode(PersistenceUnitMetadata pu) {
        return SharedCacheMode.NONE.equals(pu.getSharedCacheMode()) ||
                UNSPECIFIED.equals(pu.getProperties().getProperty(SHARED_CACHE_MODE));
    }
`

To:

`
    private boolean noneCacheMode(PersistenceUnitMetadata pu) {
        return SharedCacheMode.NONE.equals(pu.getSharedCacheMode()) ||
                NONE.equals(pu.getProperties().getProperty(SHARED_CACHE_MODE));
    }
`

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/HibernatePersistenceProviderAdaptor.java
scottmarlow: hibernateCacheEnabled is only returning true for cache enabled and false for cache disabled, but we need a 3rd state to know that the application didn't specify the USE_SECOND_LEVEL_CACHE/USE_QUERY_CACHE properties.

I'll update the comment (if Hibernate cache settings are set to false, disable cache), to make this clearer.

bstansberry (approved):

PR #12305, opened 2019-05-20
Fix for WFLY-11990, Separate opentracing from observability layer

jfdenise:
Fix for https://issues.jboss.org/browse/WFLY-11990,
opentracing in its own layer, optional dependency of observability layer.

### DIFF
--- a/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
@@ -3,9 +3,9 @@
     <dependencies>
         <layer name="management"/>
         <layer name="cdi"/>
+        <layer name="open-tracing" optional="true"/>
     </dependencies>
     <feature-group name="health"/>
     <feature-group name="metrics"/>
     <feature spec="subsystem.microprofile-config-smallrye"/>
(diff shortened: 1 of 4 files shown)


galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
```diff
@@ -8,5 +8,4 @@
     <feature-group name="health"/>
     <feature-group name="metrics"/>
     <feature spec="subsystem.microprofile-config-smallrye"/>
-    <feature spec="subsystem.microprofile-opentracing-smallrye"/>
```
bstansberry: @jfdenise Instead of complete separation, I'd prefer having observability having an optional dependency on open-tracing. So OT can be easily excluded.

Observability is a fairly broadly used concept and has been out there for a while now, so I don't think we should redefine it.

bstansberry (changes requested):

galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
jfdenise: @bstansberry , I agree. Will make changes to the PR.

jfdenise:
@bstansberry , PR updated.

bstansberry (approved):

PR #12315, opened 2019-05-24
[WFLY-12125] Integration test for PLINK-793

tmiyargi:
Issue: https://issues.jboss.org/browse/WFLY-12125

Related Issues: https://issues.jboss.org/browse/PLINK-793, https://issues.jboss.org/browse/WFLY-12111

This is blocked by Picketlink update

### DIFF
--- a/testsuite/integration/picketlink/pom.xml
+++ b/testsuite/integration/picketlink/pom.xml
@@ -38,6 +38,11 @@
             <artifactId>picketlink-idm-impl</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+			<groupId>org.jboss.ejb3</groupId>
+			<artifactId>jboss-ejb3-ext-api</artifactId>
+			<scope>test</scope>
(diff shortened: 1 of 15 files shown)


bstansberry:
@tmiyargi perhaps these failures are because the CI runs use -Dipv6 to force IPv6 testing? When tests fail globally like this that's often the problem, as before sending up the PR the dev typically ran the test locally using IPv4 and it passed.

tmiyargi:
@Brian It fails because it depends on Picketlink update, not sure how you want to proceed with this, maybe just close as I'm not sure you will update Picketlink anymore

bstansberry:
@tmiyargi When the update is available, please combine that commit into this PR or this commit into its PR and ping me or James or any of the mergers.

bstansberry:
retest this please

bstansberry:
I'm not sure but I think the Picketlink version WF is now using has this fix so I'm removing the 'hold' label.

PR #12321, opened 2019-05-27
WFLY-11601 Weld vs Undertow bootstrap: Race condition

fanste:
JIRA: https://issues.jboss.org/browse/WFLY-11601

Changes to the WELD subsystem splitted some finalization logic to a new service named WeldStartCompletionService (WFLY-9732 / WFLY-10784). The UndertowDeploymentService only had WeldStartService as a dependency. Therefore some race conditions may occur if undertow's service executed before weld's completion service.

The fix additionally adds the completion service as a dependency to undertow using the attachment WEB_DEPENDENCIES.

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentProcessor.java
@@ -58,6 +58,7 @@ import org.jboss.as.server.deployment.module.ResourceRoot;
 import org.jboss.as.weld.ServiceNames;
 import org.jboss.as.weld.WeldBootstrapService;
 import org.jboss.as.weld._private.WeldDeploymentMarker;
+import org.jboss.as.weld.WeldStartCompletionService;
 import org.jboss.as.weld.WeldStartService;
 import org.jboss.as.weld.deployment.BeanDeploymentArchiveImpl;
 import org.jboss.as.weld.deployment.BeanDeploymentModule;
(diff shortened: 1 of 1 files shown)


jamezp:
this is ok to test

jamezp:
@manovotn could you please review this?

manovotn:
Hmm, tests failed so apparently, this isn't the way to go.
I don't really see why though.

As it stands, `WeldDeploymentProcessor` already creates a dependency from undertow to `WeldStartService` which only happens if given deployment unit is a CDI deployment (as seen [here](https://github.com/wildfly/wildfly/blob/master/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentProcessor.java#L116)).
So adding `WeldStartCompletionService` under the same condition looks logical.

Both failing tests are EARs that verify init order of dependending services but both operate on pure EJB and servlets, no CDI in play.

manovotn:
After some digging, I am not sure this is the solution.

`WeldStartCompletionService` is deliberately delayed (in `WeldDeploymentCleanupProcessor`) to run after all other EE components so that Weld can perform cleanup and optimizations (dropping unused beans and such).

What we are trying here is the exact opposite and I think we are creating a circular dependency between the two services? I cannot say for sure as I was unable to directly debug bootstrap of services (is there a readme somewhere on how to do that with test in TS?).
Maybe @ropalka would know more?

It would be nice if someone more knowledgeable in WFLY's services and processors took a look.

fanste:
The failed test is only included when maven is run with `-DallTests`. Thats why it did not come up while building myself...

I've searched through the code again and I think this could not be fixed in any way, because of a chicken-egg-problem. 
The `WeldStartCompletionService` is installed using a deployment processor that runs in the CLEANUP-Phase. The completion service was introduced for some performance optimizations regarding EE components. Therefore it won't be available when EE components are initialized.

I think the only solution is to revert the PR that introduced the completion service. But I'm totally new to the whole wildfly code, maybe some expert could help us here?

fanste:
@manovotn: We both came up to the same result. So it may not be that wrong 😄

manovotn:
> I think the only solution is to revert the PR that introduced the completion service. But I'm totally new to the whole wildfly code, maybe some expert could help us here?

I am pretty sure that's something we don't want to do.

kabir:
Adding the hold label. @fanste and @manovotn - if you decide not to proceed with this, please close this PR

manovotn:
@kabir please close the PR, I lack the permissions.
We will have to handle this differently.

PR #12338, opened 2019-06-03
WFLY-12161 Expressions not working in XA DS attr share-prepared-state…

TomasHofman:
…ments

https://issues.jboss.org/browse/WFLY-12161

XA DS failed when expression was used in share-prepared-statement attribute. Attribute has `allowExpressions = true` and also the non-XA equivalent works with expressions.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
@@ -111,7 +111,6 @@ import org.jboss.as.connector.util.ModelNodeUtil;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.dmr.ModelNode;
-import org.jboss.jca.common.api.metadata.Defaults;
 import org.jboss.jca.common.api.metadata.common.Capacity;
 import org.jboss.jca.common.api.metadata.common.Extension;
 import org.jboss.jca.common.api.metadata.common.FlushStrategy;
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
```diff
@@ -187,7 +186,7 @@ static ModifiableDataSource from(final OperationContext operationContext, final
         final DsSecurity security = new DsSecurityImpl(username, password,
                 elytronEnabled? authenticationContext: securityDomain, elytronEnabled, credentialSourceSupplier, reauthPlugin);
 
-        final boolean sharePreparedStatements = ModelNodeUtil.getBooleanIfSetOrGetDefault(operationContext, dataSourceNode, SHARE_PREPARED_STATEMENTS);
+        final Boolean sharePreparedStatements = ModelNodeUtil.getBooleanIfSetOrGetDefault(operationContext, dataSourceNode, SHARE_PREPARED_STATEMENTS);
```
bstansberry: This should never be null as SHARED_PREPARED_STATEMENTS has a default value. ModelNodeUtil.getBooleanIfSetOrGetDefault can return null, but not when this value is passed in.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
```diff
@@ -288,8 +287,7 @@ static ModifiableXaDataSource xaFrom(final OperationContext operationContext, fi
         final DsSecurity security = new DsSecurityImpl(username, password,
                 elytronEnabled? authenticationContext: securityDomain, elytronEnabled, credentialSourceSupplier, reauthPlugin);
 
-        final Boolean sharePreparedStatements = dataSourceNode.hasDefined(SHARE_PREPARED_STATEMENTS.getName()) ? dataSourceNode.get(
-                SHARE_PREPARED_STATEMENTS.getName()).asBoolean() : Defaults.SHARE_PREPARED_STATEMENTS;
+        final Boolean sharePreparedStatements = ModelNodeUtil.getBooleanIfSetOrGetDefault(operationContext, dataSourceNode, SHARE_PREPARED_STATEMENTS);
```
bstansberry: This also should never return null.  I assume this line is the actual bug fix; replacing the custom code with code that calls SHARE_PREPARED_STATEMENTS.resolveModelAttribute.

Ideally most/all of these ModelNodeUtil.getXXXIfSetOrGetDefault call should be replaced by AttributeDefinition.resolveModelAttribute(context, modelnode).asXXX() or .asXXXOrNull() if the AD isn't required and has no default value.  That's out of scope for this though. :)

bstansberry (changes requested):
See previous comments.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
TomasHofman: Yeah, I wanted to get rid of IDE inspection warning...

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourceModelNodeUtil.java
TomasHofman: OK, I will use `resolveModelAttribute().asBoolean()` at least for both SHARE_PREPARED_STATEMENTS readings.

TomasHofman:
PR updated, thanks @bstansberry!

TomasHofman:
I added a second commit with a test case. I originally wanted to implement it elsewhere, but that didn't work out...

bstansberry (approved):

PR #12340, opened 2019-06-04
[WFLY-12162] Upgrading Naming Client to 1.0.11.Final.

ropalka:
https://issues.jboss.org/browse/WFLY-12162

Incorporates:
 * [WFNC-55] Naming Client must always set thread context class loader


bstansberry:
@ropalka @jamezp I think we should defer this to 18 as a change in TCCL behavior at this very late date for 17 is risky and AFAICT this is addressing an issue that we've already worked around.  If this thinking is wrong please ping me in Zulip ASAP.  For example if it doesn't get in the Tuesday TCK run my enthusiasm for it would drop a lot.

ropalka:
I agree @bstansberry @jamezp this can be deferred to 18. It is just a proper fix for effective workaround.

bstansberry (approved):

PR #12354, opened 2019-06-08
[WFLY-12051] Documentation and Test Updates for the aggregation of attributes from multiple realms.

darranl:
https://issues.jboss.org/browse/WFLY-12051

This PR presently also contains the commit for https://github.com/wildfly/wildfly/pull/12353 performing initial refactoring on the test case.

:warning: This PR will not build until an updated Wildfly Core with an updated WildFly Elytron is available.  :warning:


testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/AggregateRealmTestCase.java
```diff
@@ -506,176 +614,227 @@ private void assertNotInRole(final String rolePrintResponse, String role) {
...
+                    .build());
+
+            return configurableElements.toArray(new ConfigurableElement[configurableElements.size()]);
         }
+    }
 
-        private String createPropertiesUserWithHashedPassword(String username, String password, String realmName) {
-            return username + "=" + createHashedPassword(username, password, realmName) + "\n";
-        }
+     static class CustomFSAttributes implements ConfigurableElement {
```
marekkopecky: CustomFSAttributes class (implementation of ConfigurableElement) needs to override remove method from ConfigurableElement element. If remove method won't be override, [default remove method will be used](https://github.com/wildfly/wildfly/blob/master/testsuite/shared/src/main/java/org/wildfly/test/security/common/elytron/ConfigurableElement.java#L62) and it leads to IllegalStateException during test cleanup and cleanup won't be finished. Many elements won't be removed and could cause errors of other tests. This could leads to many false-positive errors during the CI runs.

marekkopecky (changes requested):

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/AggregateRealmTestCase.java
darranl: Good catch, thank you.

PR #12358, opened 2019-06-11
WFLY-10872 Confusing error message when writing simple-load-provider …

rhusar:
…attribute when dynamic load provider is present

Jira
https://issues.jboss.org/browse/WFLY-10872

### DIFF
--- a/mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
+++ b/mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import java.util.function.UnaryOperator;
 
 import org.jboss.as.clustering.controller.AttributeTranslation;
+import org.jboss.as.clustering.controller.AttributeValueTranslator;
 import org.jboss.as.clustering.controller.ChildResourceDefinition;
 import org.jboss.as.clustering.controller.CommonUnaryRequirement;
 import org.jboss.as.clustering.controller.ManagementResourceRegistration;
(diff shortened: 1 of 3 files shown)


mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
```diff
@@ -428,6 +429,33 @@ public PathAddress convertToTargetAddress(PathAddress aliasAddress, AliasContext
             return SimpleLoadProviderResourceDefinition.Attribute.FACTOR;
         }
 
+        AttributeValueTranslator SIMPLE_LOAD_FACTOR_DEFINED_VALIDATOR = (context, value) -> {
```
pferraro: Replace with anonymous class.

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
```diff
@@ -428,6 +429,33 @@ public PathAddress convertToTargetAddress(PathAddress aliasAddress, AliasContext
...
+            // Skip if executed on the existing current resource
+            PathAddress currentAddress = context.getCurrentAddress();
+            if (currentAddress.getLastElement().equals(SimpleLoadProviderResourceDefinition.PATH)) {
+                return value;
+            }
+
+            try {
+                // Otherwise, fail if operation executed on the legacy resource if dynamic load provider is defined
+                PathAddress address = PathAddress.pathAddress(DynamicLoadProviderResourceDefinition.PATH);
+                context.readResource(address);
```
pferraro: Use non-recursive read, e.g. context.readResource(address, false)

pferraro (changes requested):

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
rhusar: Lambdas still no good? :)

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
rhusar: Ah, thanks.

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
```diff
@@ -428,6 +429,36 @@ public PathAddress convertToTargetAddress(PathAddress aliasAddress, AliasContext
             return SimpleLoadProviderResourceDefinition.Attribute.FACTOR;
         }
 
+        AttributeValueTranslator SIMPLE_LOAD_FACTOR_DEFINED_VALIDATOR = new AttributeValueTranslator() {
```
pferraro: This property should be final.  Also, use camel case for non-static properties.
SIMPLE_LOAD_PROVIDER_TRANSLATION, however, should probably be both static and final.

pferraro (changes requested):

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationResourceDefinition.java
rhusar: Ok, cannot be static here, so final will suffice.

pferraro: Yes, but it is allowed for SIMPLE_LOAD_PROVIDER_TRANSLATION.

rhusar: Got it now, yes it should, fixed.

darranl:
@pferraro Are you Ok to double check this one?  GitHub is showing changes where you commented.

pferraro:
@darranl yeah - I'll double check and approve when ready.

pferraro (approved):

PR #12363, opened 2019-06-17
[WFLY-11759] Revert UndertowTransformersTestCase workaround

fl4via:
KernelServicesBuilder.excludeResourceFromParent API is no longer needed, the test runs okay

Jira: https://issues.jboss.org/browse/WFLY-11759


bstansberry:
#12367 incorporates this along with another commit to eliminate the need to keep the legacy test version of Undertow in sync with what's in currently in core.

PR #12371, opened 2019-06-18
[WFLY-12202] Add documentation for the `principal-transformer` attribute in aggregate-realm

ashley-abdelsayed98:
https://issues.jboss.org/browse/WFLY-12202

This also contains the changes in #12354

Analysis document: wildfly/wildfly-proposals#209
Depends on wildfly/wildfly-core#3825

### DIFF
--- a/docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
+++ b/docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
@@ -8,7 +8,9 @@ The `aggregate-realm` resource contains the following attributes: -
  * `authentication-realm` - This realm use used to load the credentials of the identity or perform evidence verification.
  * `authorization-realm` - The realm to use to load the identities attributes used for authorization.
  * `authorization-realms` - A list of one or more realms to use to load the identities attributes and aggregate into a single set of attributes.
- 
+ * `principal-transformer` - A principal transformer that can be used to transform the principal after loading the credentials
+ for authentication but before loading the identities attributes for authorization. This attribute is optional.
+
 NOTE: `authorization-realm` and `authorization-realms` are mutually exclusive so exactly one of these must be specified.
 
 == Defining a Simple Aggregation
@@ -31,6 +33,17 @@ The following command is exactly the same except the `authorization-realms` attr
 /subsystem=elytron/aggregate-realm=combined:add(
     authentication-realm=properties-realm, 
     authorization-realms=[jdbc-realm])
+----
+
+The following command will reference a principal transformer defined in the mappers configuration to be used to transform the principal
+after authentication.
+
+[source,options="nowrap"]
+----
+/subsystem=elytron/aggregate-realm=combined:add(
+    authentication-realm=properties-realm,
+    authorization-realms=[jdbc-realm],
+    principal-transformer=custom-transformer)
 ----
 
 == Aggregating Attributes
--- a/docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
+++ b/docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
@@ -1,0 +1,96 @@
+[[aggregate-security-realm]]
+= Aggregate Security Realm
+
+The aggregate security realm allows for two or more security realms to be aggregated into a single security realm allowing one to be used during the authentication steps and one or more to be used to assemble the identity used for authorization decisions and aggregating any associated attributes.
+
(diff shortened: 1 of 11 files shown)


docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
```diff
@@ -33,6 +35,16 @@ The following command is exactly the same except the `authorization-realms` attr
...
 ----
 
+The following command will define a transformer to be used to transform the principal after authentication.
+
+[source,options="nowrap"]
+----
+/subsystem=elytron/aggregate-realm=combined:add(
+    authentication-realm=properties-realm,
+    authorization-realms=[jdbc-realm],
+    principal-transformer=custom-transformer)
```
fjuma: May be good to mention that the configured principal transformer is a reference to a ```principal-transformer``` that has been defined in the ```mappers``` configuration.

docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
```diff
@@ -8,7 +8,9 @@ The `aggregate-realm` resource contains the following attributes: -
  * `authentication-realm` - This realm use used to load the credentials of the identity or perform evidence verification.
  * `authorization-realm` - The realm to use to load the identities attributes used for authorization.
  * `authorization-realms` - A list of one or more realms to use to load the identities attributes and aggregate into a single set of attributes.
- 
+ * `principal-transformer` - A principal transformer that can be used to transform the principal after loading the credentials
```
fjuma: Would be good to mention that this attribute is optional.

docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
ashley-abdelsayed98: Added

docs/src/main/asciidoc/_elytron/components/Aggregate_Security_Realm.adoc
ashley-abdelsayed98: Added

bstansberry:
this is ok to test

testsuite/shared/src/main/java/org/wildfly/test/security/common/elytron/AggregateSecurityRealm.java
```diff
@@ -54,6 +56,12 @@ public ModelNode getAddOperation() {
         if (authorizationRealm != null) {
             addOperation.get("authorization-realm").set(authorizationRealm);
         }
+        if (authorizationRealms != null) {
```
vmarekrh: Is there any reason to support `authentication-realm` and `authorization-realms` attributes at the same time?
Or why isn't here `else` statement?

ashley-abdelsayed98: These are part of the changes from #12354

vmarekrh: https://github.com/wildfly/wildfly/pull/12354/files#diff-9f7fd02d4150fbf86668f5d6e9f5624dR12
> NOTE: `authorization-realm` and `authorization-realms` are mutually exclusive so exactly one of these must be specified.

By this note, in this code missing else statement.

darranl: This is just simple utility code in the testsuite, this code is not intending to be verifying the operation - if a test was written with an invalid permutation it would quickly fail model validation and the test would fail.

vmarekrh: Then it is for a negative test scenario when we expect to fail.
Ok, Thank you

marekkopecky:
@ashley-abdelsayed98 (cc @fjuma ) I prepared tests to this feature in https://github.com/marekkopecky/wildfly/commit/cd6e0cda928d9c8171ae59dade1172c5c06da9ee

Can you please consider to review this commit and cherry-pick this commit to this your PR?

Btw, this my commit also fixes test bug described in https://github.com/wildfly/wildfly/pull/12354#discussion_r316538611, so if this bug will be fixed in PR12354 as well, it could cause some git conflicts here

marekkopecky (approved):

darranl:
Superseded by https://github.com/wildfly/wildfly/pull/12579

PR #12377, opened 2019-06-20
[WFLY-12216] Log WARN if application specifies @RunAsPrincipal and no…

tmiyargi:
…t @RunAs

Issue: https://issues.jboss.org/browse/WFLY-12216

This will log a WAR when EJB has @RunAsPrincipal annotation and @RunAS is not present

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
@@ -21,6 +21,9 @@
  */
 package org.jboss.as.test.integration.ejb.security;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.PropertyPermission;
 import java.util.concurrent.Callable;
 
 import javax.ejb.EJBAccessException;
@@ -44,7 +47,10 @@ import org.jboss.as.test.integration.ejb.security.runasprincipal.transitive.Simp
 import org.jboss.as.test.integration.ejb.security.runasprincipal.transitive.SingletonStartupBean;
 import org.jboss.as.test.integration.ejb.security.runasprincipal.transitive.StatelessSingletonUseBean;
 import org.jboss.as.test.integration.security.common.AbstractSecurityDomainSetup;
+import org.jboss.as.test.shared.TestLogHandlerSetupTask;
+import org.jboss.as.test.shared.integration.ejb.security.PermissionUtils;
 import org.jboss.as.test.shared.integration.ejb.security.Util;
+import org.jboss.as.test.shared.util.LoggingUtil;
 import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
@@ -64,7 +70,7 @@ import static org.jboss.as.test.shared.integration.ejb.security.PermissionUtils.
  * @author Carlo de Wolf, Ondrej Chaloupka
  */
 @RunWith(Arquillian.class)
-@ServerSetup({EjbSecurityDomainSetup.class})
+@ServerSetup({EjbSecurityDomainSetup.class, RunAsPrincipalTestCase.TestLogHandlerSetup.class})
 @Category(CommonCriteria.class)
 public class RunAsPrincipalTestCase  {
 
@@ -104,10 +110,14 @@ public class RunAsPrincipalTestCase  {
                 .addClass(Util.class)
                 .addClass(Entry.class)
                 .addClass(RunAsPrincipalTestCase.class)
+                .addClass(TestLogHandlerSetupTask.class)
+                .addClass(LoggingUtil.class)
                 .addClasses(AbstractSecurityDomainSetup.class, EjbSecurityDomainSetup.class)
                 .addAsWebInfResource(RunAsPrincipalTestCase.class.getPackage(), "jboss-ejb3.xml", "jboss-ejb3.xml")
                 .addAsManifestResource(new StringAsset("Dependencies: org.jboss.as.controller-client,org.jboss.dmr\n"), "MANIFEST.MF")
-                .addAsManifestResource(createPermissionsXmlAsset(new ElytronPermission("getSecurityDomain")), "permissions.xml");
+                .addAsManifestResource(createPermissionsXmlAsset(new ElytronPermission("getSecurityDomain"),
+                        new PropertyPermission("jboss.server.log.dir", "read"),
+                        PermissionUtils.createFilePermission("read", "standalone", "log", TEST_LOG_FILE_NAME)), "permissions.xml");
         war.addPackage(CommonCriteria.class.getPackage());
         return war;
     }
@@ -155,6 +165,11 @@ public class RunAsPrincipalTestCase  {
         Util.switchIdentitySCF("user1", "password1", callable);
     }
 
+    /**
+     * Test when only RunAsPrincipal is defined, it should fail but also warn the user about
+     * configuration problem.
+     * @throws Exception
+     */
     @Test
     public void testRunAsPrincipal() throws Exception {
         WhoAmI bean = lookupCallerRunAsPrincipal();
@@ -163,6 +178,7 @@ public class RunAsPrincipalTestCase  {
             Assert.fail("Expected EJBAccessException and it was get identity: " + actual);
         } catch (EJBAccessException e) {
             // good
+            Assert.assertTrue("@RunAs warning not found", LoggingUtil.hasLogMessage(TEST_LOG_FILE_NAME, LOG_MESSAGE));
         }
     }
 
@@ -217,5 +233,42 @@ public class RunAsPrincipalTestCase  {
             return checkEjbException(ex.getCause());
         }
         return ex;
+    }
+
+
+
+    private static final String TEST_HANDLER_NAME;
+    private static final String TEST_LOG_FILE_NAME;
+    private static final String LOG_MESSAGE;
+
+    static {
+        /*
+         * Make both the test handler name and the test log file specific for this class and execution so that we do not
+         * interfere with other test classes or multiple subsequent executions of this class against the same container
+         */
+        TEST_HANDLER_NAME = "test-" + RunAsPrincipalTestCase.class.getSimpleName();
+        TEST_LOG_FILE_NAME = TEST_HANDLER_NAME + ".log";
+        LOG_MESSAGE = "WFLYEJB0510";
+    }
+
+    public static class TestLogHandlerSetup extends TestLogHandlerSetupTask {
+
+        @Override
+        public Collection<String> getCategories() {
+            return Arrays.asList("org.jboss.as.ejb3.deployment");
+        }
+
+        @Override
+        public String getLevel() {
+            return "WARN";
+        }
+        @Override
+        public String getHandlerName() {
+            return TEST_HANDLER_NAME;
+        }
+        @Override
+        public String getLogFileName() {
+            return TEST_LOG_FILE_NAME;
+        }
     }
 }
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/TestLogHandlerSetupTask.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/TestLogHandlerSetupTask.java
@@ -1,0 +1,128 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.test.shared;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.Collection;
+import java.util.Deque;
+
+import org.jboss.as.arquillian.api.ServerSetupTask;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.client.Operation;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.test.shared.util.LoggingUtil;
+import org.jboss.dmr.ModelNode;
+import org.jboss.logging.Logger;
+
+/**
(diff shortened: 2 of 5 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
```diff
@@ -218,4 +232,29 @@ private Throwable checkEjbException(Throwable ex) {
...
+            try {
+                fileStream = Files.lines(Paths.get(logPath), StandardCharsets.ISO_8859_1);
+            } catch (IOException e4) {
+                throw new RuntimeException("Server logs has not standard Charsets (UTF8 or ISO_8859_1)");
+            }
+        } catch (IOException e) {
+            // server.log file is not created, it is the same as server.log is empty
+        }
+        if(fileStream != null) {
+            return fileStream.filter(lines -> lines.contains("@RunAs annotation is required when using @RunAsPrincipal")).iterator().hasNext();
```
bstansberry: We're starting to have a lot of this log searching. For stuff where we expect something logged, perhaps use something like https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/ReversedLinesFileReader.html to search from the end rather than starting many thousands of lines before?

@jamezp WDYT?

bstansberry (changes requested):
The reverse lines thing is just a thought but for sure the log message i18n needs to be corrected.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
tmiyargi: @bstansberry In this case, the reverse lines I think is not useful since I expect the log line on EJB application deployment time, at the beginning. I was wondering if you already have something in place that can be reused to search for something in the logs.

bstansberry: @tmiyargi The server.log covers the entire surefire execution which in the case of testsuite/integration/basic might mean logging from hundreds of tests and dozens of reloads.

For example, in the server log from one of the CI jobs for this PR, the WARN is logged on line number 20481.

https://ci.wildfly.org/repository/download/WF_PullRequest_LinuxJdk11/155885:id/testsuite/integration/basic/target/wildfly/standalone/log/server.log

I don't know if there are any utilities around this kind of thing. @jamezp might know.

One idea I just had is a server setup thing that adds a new test-class-specific logging handler and then removes it at the end. So the scan can just cover the file from that new handler. That would work for both this kind of case, where you are asserting the presence of something, and for cases where the assertion is something isn't logged. (I particularly hate those kinds of tests, but people write them, since bug fixes must have regression tests. I'd hate those tests less if they were'n't scanning so much.)

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
```diff
@@ -218,4 +232,40 @@ private Throwable checkEjbException(Throwable ex) {
...
+                (jbossHome.endsWith(File.separator) || jbossHome.endsWith("/")) ? "" : File.separator,
+                File.separator, File.separator);
+        logPath = logPath.replace('/', File.separatorChar);
+        BufferedReader fileReader = null;
+        try {
+            fileReader = Files.newBufferedReader(Paths.get(logPath)); // UTF8 is used by default
+        } catch (MalformedInputException e1) {
+            // some windows machines could accept only StandardCharsets.ISO_8859_1 encoding
+            try {
+                fileReader = Files.newBufferedReader(Paths.get(logPath), StandardCharsets.ISO_8859_1);
```
jamezp: I'm not too sure this is really requires as UTF-8 should work. Either way though we need to ensure the `fileReader` is closed at the end.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
```diff
@@ -218,4 +232,40 @@ private Throwable checkEjbException(Throwable ex) {
...
+            }
+        } catch (IOException e) {
+            // server.log file is not created, it is the same as server.log is empty
+        }
+
+        String line = "";
+        boolean found = false;
+        try {
+            while ((line = fileReader.readLine()) != null) {
+                if(line.contains("@RunAs annotation is required when using @RunAsPrincipal")) {
```
jamezp: We need to use the message id, `WFLYEJB00510`. Once we have localization done the output may be in a different language.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
```diff
@@ -218,4 +232,40 @@ private Throwable checkEjbException(Throwable ex) {
         }
         return ex;
     }
+
+    private boolean logContainsWARN() {
+        String jbossHome = System.getProperty("jboss.inst");
+        String logPath = String.format("%s%sstandalone%slog%sserver.log", jbossHome,
+                (jbossHome.endsWith(File.separator) || jbossHome.endsWith("/")) ? "" : File.separator,
+                File.separator, File.separator);
+        logPath = logPath.replace('/', File.separatorChar);
```
jamezp: It might be a good idea to lookup the path via a management operation instead. Something like:
```
final ModelNode address = Operations.createAddress("subsystem", "logging", "periodic-rotating-file-handler", "FILE");
final ModelNode op = Operations.createOperation("resolve-path", address);
final ModelNode result = client.execute(op);
if (!Operations.isSuccessfulOutcome(result)) {
    Assert.fail(Operations.getFailureDescription(result).asString());
}
final Path logPath = Paths.get(Operations.readResult(result).asString());
```

This would also allow the change to the `testsuite/integration/pom.xml` unnecessary.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
```diff
@@ -96,6 +103,7 @@
 
     @Deployment
     public static Archive<?> runAsDeployment() {
+       // jbossHome = System.getProperty("jboss.home");
```
jamezp: This looks like it can be removed :)

jamezp (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/RunAsPrincipalTestCase.java
jamezp: There is no utility that I'm aware of either, but I do like the idea of having some kind of common handler that is added for tests like this. I don't really like tests checking for log messages however sometimes it's the only option. I'd be happy to look at a utility for this if interested. A benefit of a common `ServerSetupTask` is it could take something like the log level and the log category to reduce the I/O.

One thing I find is useful is to use the `json-formatter` so we have some structured data instead. Though I guess if we're just looking for message id's it might not be a big deal.

tmiyargi: So, I created the utility class, let me know if it's fine. Thank you Brian and James.

bstansberry: Thanks @tmiyargi!

bstansberry (approved):

bstansberry:
@tmiyargi The 'Linux with security manager'  failure is related.

bstansberry:
@tmiyargi there's a checkstyle failure in testsuite/shared.

bstansberry (approved):
@tmiyargi @jamezp I think this is ok to merge. But this is an in-vm test, and executing management ops in in-vm tests is painful in a security manager environment.

Inside the server, we guarantee the jboss.server.log.dir property will be set so it is safe for code to read that.  Taking advantage of that, resolving the log file only requires a single PropertyPermission, not a complex set.  #12405 illustrates that. (A FilePermission is also needed to read the file, but that's a separate thing.)

So, any of these are ok with me:

1) Teresa squashes my last commit from #12405 into hers. (Squashing is fine.)
2) Teresa's already gone above and beyond here by adding a nice utility. We merge this, and I file a JIRA to enhance what she's done and update the commit message in #12405 
3) The #12405 commit is a bad idea, we merge this and we're done.

Please @ ping me with your preference.

tmiyargi:
@bstansberry I can squash the changes no problem, when you say "this is an in-vm test" in which case would you use the other method?

bstansberry:
@tmiyargi The other method is appropriate for tests that are annotated with @RunAsClient. With those the test method is not executing in the server VM so it cannot access the server's system properties.

Since @RunAsClient should be used wherever practical that's an important use case.

testsuite/shared/src/main/java/org/jboss/as/test/shared/TestLogHandlerSetupTask.java
```diff
@@ -0,0 +1,121 @@
...
+                .append("periodic-rotating-file-handler", getHandlerName()));
+        addTestLogOp.get("level").set(getLevel());
+        addTestLogOp.get("append").set("true");
+        addTestLogOp.get("suffix").set(".yyyy-MM-dd");
+        ModelNode file = new ModelNode();
+        file.get("relative-to").set("jboss.server.log.dir");
+        file.get("path").set(getLogFileName());
+        addTestLogOp.get("file").set(file);
+        addTestLogOp.get("formatter").set("%-5p [%c] (%t) %s%e%n");
+        applyUpdate(managementClient.getControllerClient(), addTestLogOp, false);
```
jamezp: We can get slightly better performance here if we use a composite operation so we only invoke a single client call. Again this is not important, but anything we can do to speed up tests is good :)

Purely as an example I'd probably do something like the following.

```
import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.ArrayDeque;
import java.util.Collection;
import java.util.Deque;

import org.jboss.as.arquillian.api.ServerSetupTask;
import org.jboss.as.arquillian.container.ManagementClient;
import org.jboss.as.controller.client.ModelControllerClient;
import org.jboss.as.controller.client.Operation;
import org.jboss.as.controller.client.helpers.Operations;
import org.jboss.as.test.shared.util.LoggingUtil;
import org.jboss.logging.Logger;

/**
 * Used to search for strings in logs, it adds a handler so the search is done on a small file
 * @author tmiyar
 *
 */
public abstract class TestLogHandlerSetupTask implements ServerSetupTask {

    private final Logger LOGGER = Logger.getLogger(TestLogHandlerSetupTask.class);
    private final Deque<ModelNode> removeOps = new ArrayDeque<>();

    @Override
    public void setup(ManagementClient managementClient, String containerId) throws Exception {

        final Operations.CompositeOperationBuilder builder = Operations.CompositeOperationBuilder.create();

        //add handler
        final ModelNode handlerAddress = Operations.createAddress(SUBSYSTEM, "logging", "file-handler", getHandlerName());
        ModelNode addTestLogOp = Operations.createAddOperation(handlerAddress);
        addTestLogOp.get("level").set(getLevel());
        addTestLogOp.get("append").set("true");
        ModelNode file = new ModelNode();
        file.get("relative-to").set("jboss.server.log.dir");
        file.get("path").set(getLogFileName());
        addTestLogOp.get("file").set(file);
        addTestLogOp.get("formatter").set("%-5p [%c] (%t) %s%e%n");
        builder.addStep(addTestLogOp);
        removeOps.add(Operations.createRemoveOperation(handlerAddress));

        //add category with new handler
        Collection<String> categories = getCategories();
        if (categories == null || categories.isEmpty()) {
            LOGGER.warn("getCategories() returned empty collection.");
            return;
        }

        for (String category : categories) {
            if (category == null || category.isEmpty()) {
                LOGGER.warn("Empty category name provided.");
                continue;
            }
            final ModelNode loggerAddress = Operations.createAddress(SUBSYSTEM, "logging", "logger", category);
            ModelNode op = Operations.createAddOperation(loggerAddress);
            op.get("level").set(getLevel());
            op.get("handlers").add(getHandlerName());
            builder.addStep(op);
            removeOps.addFirst(Operations.createRemoveOperation(loggerAddress));
        }
        applyUpdate(managementClient.getControllerClient(), builder.build(), false);
    }

    @Override
    public void tearDown(ManagementClient managementClient, String containerId) throws Exception {
        Path logPath = LoggingUtil.getLogPath(managementClient, "file-handler", getHandlerName());

        // Remove the loggers and handler
        final Operations.CompositeOperationBuilder builder = Operations.CompositeOperationBuilder.create();
        ModelNode removeOp;
        while ((removeOp = removeOps.pollFirst()) != null) {
            builder.addStep(removeOp);
        }

        // remove file, note this needs to be done after the operations have been executed as we need to ensure that
        // no FD's are open. This can be an issue on Windows.
        Files.deleteIfExists(logPath);
    }

    public abstract Collection<String> getCategories();
    public abstract String getLevel();
    public abstract String getHandlerName();
    public abstract String getLogFileName();

    protected void applyUpdate(final ModelControllerClient client, ModelNode update, boolean allowFailure) throws IOException {
        applyUpdate(client, Operation.Factory.create(update), allowFailure);
    }

    private void applyUpdate(final ModelControllerClient client, Operation update, boolean allowFailure) throws IOException {
        ModelNode result = client.execute(update);
        if (!Operations.isSuccessfulOutcome(result)) {
            if (allowFailure) {
                LOGGER.tracef("Failed to configure logger: %s", Operations.getFailureDescription(result).asString());
            } else {
                throw new RuntimeException("Failed to configure logger: " + Operations.getFailureDescription(result).asString());
            }
        } else {
            LOGGER.trace(Operations.readResult(result).asString());
        }
    }
}
```

testsuite/shared/src/main/java/org/jboss/as/test/shared/TestLogHandlerSetupTask.java
```diff
@@ -0,0 +1,121 @@
...
+            op.get("handlers").add(getHandlerName());
+            applyUpdate(managementClient.getControllerClient(), op, false);
+        }
+    }
+
+    @Override
+    public void tearDown(ManagementClient managementClient, String containerId) throws Exception {
+        // remove file
+        Path logPath = LoggingUtil.getLogPath(managementClient, "periodic-rotating-file-handler", getHandlerName());
+        Files.deleteIfExists(logPath);
```
jamezp: The actual delete should happen at the end as WIndows can be very picky about files being logged. So the remove would need to happen first.

testsuite/shared/src/main/java/org/jboss/as/test/shared/util/LoggingUtil.java
```diff
@@ -0,0 +1,94 @@
...
+        return Paths.get(System.getProperty("jboss.server.log.dir")).resolve(logFile);
+    }
+
+    public static boolean hasLogMessage(String logFileName, String logMessage) throws Exception {
+
+        BufferedReader fileReader = null;
+        boolean found = false;
+        Path logPath = LoggingUtil.getInServerLogPath(logFileName);
+
+        fileReader = Files.newBufferedReader(logPath);
```
jamezp: We should probably just use a try with resources here. At a minimum we'd need the close in a finally block. Something like
```
    public static boolean hasLogMessage(String logFileName, String logMessage) throws Exception {

        boolean found = false;
        Path logPath = LoggingUtil.getInServerLogPath(logFileName);

        try (BufferedReader fileReader = Files.newBufferedReader(logPath)) {

            String line = "";
            while ((line = fileReader.readLine()) != null) {
                if (line.contains(logMessage)) {
                    found = true;
                    break;
                }
            }
        }

        return found;
    }
```

jamezp (changes requested):
The `1` file should definitely be removed the git and the file deletion should happen after the handler has been removed. Other than that the comments/suggestions are really just minor tweaks. Thanks for this @tmiyargi

jamezp (approved):

PR #12381, opened 2019-06-21
[WFLY-12175] Fix some wrong links in the documentation

yersan:
Fixed incorrect links in the documentation in :

https://docs.wildfly.org/17/Admin_Guide.html#Description_of_the_Management_Model 
https://docs.wildfly.org/17/Admin_Guide.html#management-resources
https://docs.wildfly.org/17/Admin_Guide.html#description-of-an-attribute
https://docs.wildfly.org/17/Admin_Guide.html#Global_operations
https://docs.wildfly.org/17/Extending_WildFly.html#wildfly-managed-resources

Jira issue: https://issues.jboss.org/browse/WFLY-12175

### DIFF
--- a/docs/src/main/asciidoc/_extending-wildfly/Key_Interfaces_and_Classes_Relevant_to_Extension_Developers.adoc
+++ b/docs/src/main/asciidoc/_extending-wildfly/Key_Interfaces_and_Classes_Relevant_to_Extension_Developers.adoc
@@ -67,7 +67,7 @@ interface) relate to the notion of managed resources in the AS.
 
 Each subsystem is responsible for managing one or more management
 resources. The conceptual characteristics of a management resource are
-covered in some detail in the link:Admin_Guide{outfilesuffix}#Management_resources[Admin
+covered in some detail in the link:Admin_Guide{outfilesuffix}#management-resources[Admin
 Guide]; here we'll just summarize the main points. A management resource
 has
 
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/_extending-wildfly/Key_Interfaces_and_Classes_Relevant_to_Extension_Developers.adoc
```diff
@@ -67,9 +67,8 @@ interface) relate to the notion of managed resources in the AS.
 
 Each subsystem is responsible for managing one or more management
 resources. The conceptual characteristics of a management resource are
-covered in some detail in the link:Admin_Guide{outfilesuffix}#Management_resources[Admin
-Guide]; here we'll just summarize the main points. A management resource
-has
+covered in some detail in the <<management-resources,Admin Guide>>;
```
bstansberry: This link works fine for me with the existing doc, but is broken with this change. This "Extending WildFly" doc is a different page from "Admin Guide" so an internal link will not work.

bstansberry (changes requested):
@yersan One change on here breaks something for me; the rest look good.

docs/src/main/asciidoc/_extending-wildfly/Key_Interfaces_and_Classes_Relevant_to_Extension_Developers.adoc
yersan: Thanks for the review @bstansberry , yes, somehow I did not test correctly this link. Now it is fixed. 

Notice that the intention of this link is to jump to "2.3. Management resources" section on "WildFly Admin Guide".  In our current version, the link jumps to "WildFly Admin Guide" but it doesn't move to "2.3. Management resources". It is a very subtle difference.

bstansberry (approved):
Thanks!

PR #12382, opened 2019-06-24
[WFLY-12255 ][WFLY-951] adjustment of the fix for expiry scanner definition

ochaloup:
https://issues.jboss.org/browse/WFLY-12255 (follow-up to https://issues.jboss.org/browse/WFLY-951)

By Narayana functionality the expiry scanner class name can be defined with shortened version of the property name or with property qualified with full class name.

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
+++ b/transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
@@ -110,7 +110,8 @@ public class ArjunaRecoveryManagerService implements Service<RecoveryManagerServ
         recoveryExtensions.add(SubordinateAtomicActionRecoveryModule.class.getName());
 
         final List<String> expiryScanners;
-        if (System.getProperty("RecoveryEnvironmentBean.expiryScannerClassNames") != null) {
+        if (System.getProperty("RecoveryEnvironmentBean.expiryScannerClassNames") != null ||
+                System.getProperty("com.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.expiryScannerClassNames") != null) {
             expiryScanners = recoveryEnvironmentBean.getExpiryScannerClassNames();
         } else {
             expiryScanners = new ArrayList<String>();


ochaloup:
Retesting locally. I run the single test and the whole fleet of clustering tests as well and all passed. I updated the PR with master and pushed to check the CI.

kabir:
@ochaloup I see `"RecoveryEnvironmentBean.expiryScannerClassNames` shows up in the testsuite. Does it make sense to do a similar test for `com.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.expiryScannerClassNames`?

https://issues.jboss.org/browse/WFLY-951 is very old, I think you should create a new issue following up on that one

ochaloup:
@kabir thanks for your points. I created a new jira and changed the commit description.
But I'm not sure about the testsuite. It's a Narayana specific feature which is tested in the Narayana testsuite. There is no test in WildFly testsuite that would touch this functionality. There is only one place where I can see the name and that's under {{iiop}} testsuite module but that configuration value has no influence on any test there.
If you don't mind I would leave this as functionality for Narayana unit testing. But if you think differently, please, let me know and I will create some.

bstansberry:
@chengfang @tadamski Does the failure in https://ci.wildfly.org/viewLog.html?buildId=156858&buildTypeId=WF_PullRequest_LinuxSm indicate a problem in the EJB timer service handling? The stack trace looks like the code trying to persist the timer is failing.

I'm not sure why that only fails intermittently. Started failing on June 10.

https://ci.wildfly.org/project.html?projectId=WF_PullRequest&buildTypeId=&tab=testDetails&testNameId=-2932164008050579664&order=TEST_STATUS_DESC&branch_WF_PullRequest=__all_branches__&itemsCount=50

Not related to this PR.

chengfang:
@bstansberry the failure doesn't look like from ejb timer processing.  It's failure to get timer db connection, which was in turn casued by security permission check. I guess it will pass if running without security manager.  Not sure if it always fails with sm on.  I'll create a separate issue to track it.

ochaloup:
@kabir just by interest, is there something more to provide for this could be merged?

PR #12385, opened 2019-06-24
[17.0.x] WFLY-12232 Fix FD_SOCK socket binding name/stack mismatch in legacy build template.

pferraro:
https://issues.jboss.org/browse/WFLY-12232


jamezp (approved):

kabir:
@jamezp @bstansberry I did not notice this was 17.0.x - hopefully that is not a problem

jamezp:
@kabir Nope, we wanted it in. I only hadn't merged it yet because there were some downstream wants I wanted to merge at the same time is all. Not a big deal at all.

PR #12386, opened 2019-06-24
[WFLY-12236] Turn off the INFO Artemis audit logging.

jamezp:
https://issues.jboss.org/browse/WFLY-12236

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
@@ -86,6 +86,12 @@ import org.wildfly.extension.messaging.activemq.jms.bridge.JMSBridgeDefinition;
 import static org.wildfly.extension.messaging.activemq.CommonAttributes.EXTERNAL_JMS_QUEUE;
 import static org.wildfly.extension.messaging.activemq.CommonAttributes.EXTERNAL_JMS_TOPIC;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.LogManager;
+import java.util.logging.Logger;
+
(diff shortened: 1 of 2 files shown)


kabir:
Please ignore comment - the dep tree analyser job is configured globally to use master as the upstream branch

bstansberry:
@ehsavoie please approve
@jamezp this needs a PR for master once the details are approved

ehsavoie (approved):

bstansberry:
@jamezp we both dislike these kinds of tests but i wonder if we should add a testsuite/integration/basic test to scan the logs for "org.apache.activemq.audit". Trick is to get it to run late enough in the suite that the messaging tests will have had a chance to generate unwanted logs.

A test like that could be dropped in the future; it would just a be a temporary guard until we're comfortable with how this logging is integrated.

jamezp:
@bstansberry I was actually wondering if I should look at a test that would read a log file and check for repeated messages, with some kind of tolerance, so we could find other messages like this sooner.

That said, sure I could write a test for that. I can pick a test it runs on normally and ensure it runs last.

With regard to the master. I thought @ehsavoie was doing fixing this by updating the configs. If I'm wrong there please let me know and I'll create one.

bstansberry:
@jamezp -- We need the same fix for master. Any downstream change must go upstream first.

Some other change can overwrite that when it's ready. TBH though I suspect some variant of this will remain as I'm extremely reluctant to have this functionality enabled without at least a boolean attribute in the subsystem to turn it on. We'll also have to work through how this relates to things like JSON logging to stdout on the cloud where we want a bit of metadata to allow for discrimination between different 'types' of output. IOW WildFly has an overall logging architecture and we don't just ignore that and throw out whatever.

jamezp:
@bstansberry Okay I created a master PR #12395 and added a test to both.

bstansberry (approved):

PR #12390, opened 2019-06-26
WFLY-12242 Default VERIFY_SUSPECT timeout is too high

pferraro:
https://issues.jboss.org/browse/WFLY-12242


rhusar:
👍  Infinispan has lowered this too since 9.3.

PR #12391, opened 2019-06-26
WFLY-9631 "Failed to reinstate timer" warning is shown when creating …

chengfang:
…large number of EJB timers

JIRA: https://issues.jboss.org/browse/WFLY-9631

During timer data store refresh, avoid removing those timers that were just being created in memory and haven't been persisted to db.


chengfang:
@tadamski please review.

tadamski (commented):
@chengfang not sure if I understand correctly: after this fix we there are no more timers that fail to execute because they were removed but we there is still a possibility of duplicate resource warning, right?

chengfang:
@tadamski this fix resolved both symptoms: duplicate resource warning, and missing timers.  So after this fix, there is no more duplicate timer resource warnings, and all scheduled timers expire correctly.

tadamski (approved):

tadamski:
The fix seems ok to me but apparently I misread a comment in Jira. Approving.

PR #12394, opened 2019-06-28
WFLY-12247, WFLY-12248 Alignment of clustering configurations

pferraro:
Align legacy distributable-web subsystem xml template with xml parser description
https://issues.jboss.org/browse/WFLY-12247

Align jgroups stacks in example configurations with default ha configuration
https://issues.jboss.org/browse/WFLY-12248


jamezp (approved):

clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/DistributableWebXMLParser.java
```diff
@@ -47,15 +47,15 @@ public PersistentResourceXMLDescription getParserDescription() {
                 .addAttribute(DistributableWebResourceDefinition.Attribute.DEFAULT_SESSION_MANAGEMENT.getDefinition())
                 .addAttribute(DistributableWebResourceDefinition.Attribute.DEFAULT_SSO_MANAGEMENT.getDefinition())
                 .addChild(new AttributeXMLBuilderOperator()
-                        .addAttributes(SessionManagementResourceDefinition.Attribute.class)
                         .addAttributes(InfinispanSessionManagementResourceDefinition.Attribute.class)
+                        .addAttributes(SessionManagementResourceDefinition.Attribute.class)
```
bstansberry: @pferraro FWIW I think the legacy template was wrong and the parser was right.

tns:infinispan-session-management extends tns:abstract-session-management. It's debatable whose attributes comes first, a base or an extension, but my feeling is it's the base. And it's tns:abstract-session-management that defines the 'name' attribute. If it's attributes come after then the template should be: 

```
<infinispan-session-management cache-container="web" name="default" granularity="SESSION">
```

not

```
<infinispan-session-management name="default" cache-container="web" granularity="SESSION">
```

PR #12416, opened 2019-07-04
[WFLY-12268] ${maven.repository.url} in repository url is not evaluat…

gaol:
…ed when org.jboss:jboss-parent is missing

Issue: https://issues.jboss.org/browse/WFLY-12268

Changes in this PR:
- Remove the property: `maven.repository.url` because it is not useful for evaluate `jboss-parent` as the repository url ([Known Maven issue](https://issues.apache.org/jira/browse/MNG-5611))
- Added a profile: `http` and can be activated using property: `-Dmaven.repository.protocol=http` to switch to use http repository.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,14 +175,6 @@
         <wildfly.build.output.dir>build/target/${server.output.dir.prefix}-${server.output.dir.version}</wildfly.build.output.dir>
         <wildfly.web.build.output.dir>servlet-dist/target/${server.output.dir.prefix}-servlet-${server.output.dir.version}</wildfly.web.build.output.dir>
 
-        <!-- Protocol to use for communication with remote maven repositories.
-             You can set to 'http' if you are using a maven proxy and 'https' 
-             interferes with that. Use 'https' for builds that will be released
-             to non-snapshot public maven repos -->
-        <maven.repository.protocol>https</maven.repository.protocol>
-        <!-- The full remote maven repo URL; can be overridden via -D for special use cases -->
-        <maven.repository.url>${maven.repository.protocol}://repository.jboss.org/nexus/content/groups/public/</maven.repository.url>
-
         <!-- Surefire args -->
         <surefire.extra.args></surefire.extra.args>
         <surefire.jpda.args></surefire.jpda.args>
@@ -7904,7 +7896,7 @@
             </snapshots>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Repository Group</name>
-            <url>${maven.repository.url}</url>
+            <url>https://repository.jboss.org/nexus/content/groups/public/</url>
             <layout>default</layout>
         </repository>
     </repositories>
@@ -7919,7 +7911,7 @@
             </snapshots>
             <id>jboss-public-repository-group</id>
             <name>JBoss Public Repository Group</name>
-            <url>${maven.repository.url}</url>
+            <url>https://repository.jboss.org/nexus/content/groups/public/</url>
         </pluginRepository>
     </pluginRepositories>
 


gaol:
@bstansberry would you please review? Thanks :)

xstefank:
@wildfly/mergers-full can someone review, please? This is blocking downstream CI builds.

kabir:
@wildfly/mergers-full test

pom.xml
```diff
@@ -8622,5 +8620,47 @@
...
+                <property>
+                    <name>maven.repository.protocol</name>
+                    <value>http</value>
+                </property>
+            </activation>
+	    <repositories>
+		<repository>
+		    <id>jboss-public-repository-group</id>
+		    <name>JBoss Public Repository Group</name>
+		    <url>http://repository.jboss.org/nexus/content/groups/public/</url>
```
jamezp: This has to be `https`.

pom.xml
```diff
@@ -8622,5 +8620,47 @@
...
+			<updatePolicy>never</updatePolicy>
+		    </snapshots>
+		</repository>
+	    </repositories>
+
+	    <pluginRepositories>
+		<pluginRepository>
+		    <id>jboss-public-repository-group</id>
+		    <name>JBoss Public Repository Group</name>
+		    <url>http://repository.jboss.org/nexus/content/groups/public/</url>
```
jamezp: Same as above, this must be `https`.

jamezp (changes requested):
Please note the two changes to use https are required. However I've not yet validated the rest of the changes which I need to have a better look at.

pom.xml
```diff
@@ -180,8 +180,6 @@
              interferes with that. Use 'https' for builds that will be released
              to non-snapshot public maven repos -->
         <maven.repository.protocol>https</maven.repository.protocol>
```
jamezp: This should be deleted as well.

pom.xml
```diff
@@ -8622,5 +8620,47 @@
                 </pluginRepository>
             </pluginRepositories>
         </profile>
+
+
+        <profile>
```
jamezp: I see no need for this profile.

jamezp (changes requested):

jamezp:
This is valid and I was bale to replicate this locally. We'll need a core PR as well. We should just remove the properties and use the URL specifically.

pom.xml
gaol: This is needed to activate the `http` profile by setting value `-Dmaven.repository.protocol=http` to use `HTTP` repository instead of `HTTPS` one, I will explain in the below review comment.

pom.xml
gaol: Actually it is not needed in WildFly, but it is needed in EAP side. I think it is better to have it in WildFly as well so that when new EAP branch is created from WildFly, it won't get missed.

We have internal Maven repositories used for EAP CI running within a self-controlled security environment which does not need `HTTPS`, and using `HTTPS` requires self-signed CA certificate configured in each JVM, and there is [known issue in Windows machines in brontes](https://issues.jboss.org/browse/WFCI-44)

gaol:
Created https://issues.jboss.org/browse/WFCORE-4564 for the WildFly core changes, I will fire PR after we have agreement on the `profile` part :)

pom.xml
jamezp: We don't want a way in WildFly to enable insecure downloads. This looks like a workaround to an issue so once the issue is resolved this would go away anyway.

jamezp (changes requested):

gaol:
@jamezp OK, Updated as required, please review again, thanks

gaol:
PR for WildFly Core is: https://github.com/wildfly/wildfly-core/pull/3857

jamezp (approved):

jamezp:
I'm closing this as we do want these properties for proxies. @gaol I'm putting an email together now explaining what we want to do here.

PR #12420, opened 2019-07-08
[WFLY-12184] http transaction provider itegrated with WFTC

ochaloup:
https://issues.jboss.org/browse/WFLY-12184

Enabling http ejb txn provider in the WFTC module.

@fl4via  or @dmlloyd do you think you can take a look?

### DIFF
--- a/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml
@@ -22,7 +22,7 @@
   ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
   -->
 
-<module xmlns="urn:jboss:module:1.5" name="org.wildfly.transaction.client">
+<module xmlns="urn:jboss:module:1.8" name="org.wildfly.transaction.client">
     <exports>
         <exclude path="org/wildfly/transaction/client/_private" />
     </exports>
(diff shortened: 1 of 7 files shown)


darranl:
Is this a feature?  It sounds like adding new functionality that doesn't already exist?

ochaloup:
@darranl the ejb remoting with http was added to the WFLY/EAP in EAP 7.2.0 https://issues.jboss.org/browse/EAP7-723 with preview as https://issues.jboss.org/browse/EAP7-446 but it was not working for server to server call. I was discussing this with @fl4via regarding her EAP7 issue https://issues.jboss.org/browse/EAP7-1236 but what I understood (if I was not wrong) the backloged EAP7 is only for support configuration of `remote-outbound-connection` for standalone.xml. The processing of the ejb over http is expected to be working (based on the prior mentioned two EAP7 issues).
When working on OpenShift integration I found the server to server does not work. The code is prepared in the repositories of the http client but final bit of integration is not working.

Do you think this should be tracked as a new feature? Should I change the WFLY issue to type of feature? Or a new EAP7 issue should be created?

ochaloup:
@darranl @bstansberry @fl4via please, what do you think about this PR. What I can do to proceed with it? Should this be considered as a new functionality and is the RFE needed then? Could this be merged to WFLY?

(I was retesting the test errors and I assume they are not connected to this fix.)

ochaloup:
@mjurc @msve can you please confirm this PR can be merged without RFE?

bstansberry (approved):
@ochaloup If the tests in this PR had been filed as the reproducer for a Bug JIRA, it seems pretty clear we'd treat the JIRA as valid, as there's no reason this should work for a client running outside the server but not work for a client running inside.

ochaloup:
@bstansberry yes, it's as you said - the tests reproduce the reported issue.

PR #12433, opened 2019-07-15
[17.x][WFLY-11786][WFLY-12175][WFLY-12277] Port doc fixes to 17.0.x branch

bstansberry:
https://issues.jboss.org/browse/WFLY-11786
https://issues.jboss.org/browse/WFLY-12175
https://issues.jboss.org/browse/WFLY-12277

Port recent doc fixes to the 17.0.x branch. The basic point being that then the docs can be generated, with '17' as the version, and then that output can be copied to the '17' dir of a checkout  wildfly/wildlfy.github.io in order to create a PR to fix the actual WF 17 docs.


docs/src/main/asciidoc/Getting_Started_Guide.adoc
```diff
@@ -9,6 +9,7 @@ WildFly team;
 :doctype: book
 :icons: font
 :source-highlighter: coderay
+ifdef::env-github[:imagesdir: images/]
```
jamezp: Not really an issue for this PR, but I wonder if we need any of this `ifdef::env-github` stuff as we generate the documentation.

jamezp (approved):

docs/src/main/asciidoc/Getting_Started_Guide.adoc
bstansberry: @jamezp I'll conduct a test using this PR branch...

bstansberry: @jamezp It does what I thought it does. It allows the images to be properly rendered when looking at the adoc pages in wildfly/wildfly github repo.

Compare section 1.4.4 of this:

https://github.com/bstansberry/wildfly/blob/doc_fixes/docs/src/main/asciidoc/Getting_Started_Guide.adoc

with the same from a branch I just created where I removed the ifdef from the page:

https://github.com/bstansberry/wildfly/blob/doc_fixes/docs/src/main/asciidoc/Getting_Started_Guide.adoc

The latter is missing the big graphic showing the WildFly welcome page.

jamezp: Ah that makes sense.

PR #12437, opened 2019-07-17
[WFLY-12294] Upgrade MicroProfile Metrics 2.0.1

jmesnil:
* Update org.eclipse.microprofile-metrics to 2.0.1
* Update smallrye-metrics to io.smallrye-metrics-2.0:1.1.0
  (both Maven A and V have changed)
* Fix integration tests due to specifitions changes in Prometheus
* conversion (and HTTP status errors)

JIRA: https://issues.jboss.org/browse/WFLY-12294

### DIFF
--- a/feature-pack/pom.xml
+++ b/feature-pack/pom.xml
@@ -3020,7 +3020,7 @@
 
         <dependency>
             <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-metrics</artifactId>
+            <artifactId>smallrye-metrics-2.0</artifactId>
             <exclusions>
                 <exclusion>
                     <groupId>*</groupId>
--- a/microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
+++ b/microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
@@ -1,18 +1,18 @@
-classloader.currentLoadedClass.count.displayName: Current Loaded Class Count
-classloader.currentLoadedClass.count.type: counter
-classloader.currentLoadedClass.count.unit: none
-classloader.currentLoadedClass.count.description: Displays the number of classes that are currently loaded in the Java virtual machine.
-classloader.currentLoadedClass.count.mbean: java.lang:type=ClassLoading/LoadedClassCount
-classloader.totalLoadedClass.count.displayName: Total Loaded Class Count
-classloader.totalLoadedClass.count.type: counter
-classloader.totalLoadedClass.count.unit: none
-classloader.totalLoadedClass.count.description: Displays the total number of classes that have been loaded since the Java virtual machine has started execution.
-classloader.totalLoadedClass.count.mbean: java.lang:type=ClassLoading/TotalLoadedClassCount
-classloader.totalUnloadedClass.count.displayName: Total Unloaded Class Count
-classloader.totalUnloadedClass.count.type: counter
-classloader.totalUnloadedClass.count.unit: none
-classloader.totalUnloadedClass.count.description: Displays the total number of classes unloaded since the Java virtual machine has started execution.
-classloader.totalUnloadedClass.count.mbean: java.lang:type=ClassLoading/UnloadedClassCount
+classloader.loadedClasses.count.displayName: Current Loaded Class Count
+classloader.loadedClasses.count.type: gauge
+classloader.loadedClasses.count.unit: none
+classloader.loadedClasses.count.description: Displays the number of classes that are currently loaded in the Java virtual machine.
+classloader.loadedClasses.count.mbean: java.lang:type=ClassLoading/LoadedClassCount
+classloader.loadedClasses.total.displayName: Total Loaded Class Count
+classloader.loadedClasses.total.type: counter
+classloader.loadedClasses.total.unit: none
+classloader.loadedClasses.total.description: Displays the total number of classes that have been loaded since the Java virtual machine has started execution.
+classloader.loadedClasses.total.mbean: java.lang:type=ClassLoading/TotalLoadedClassCount
+classloader.unloadedClasses.total.displayName: Total Unloaded Class Count
+classloader.unloadedClasses.total.type: counter
+classloader.unloadedClasses.total.unit: none
+classloader.unloadedClasses.total.description: Displays the total number of classes unloaded since the Java virtual machine has started execution.
+classloader.unloadedClasses.total.mbean: java.lang:type=ClassLoading/UnloadedClassCount
 cpu.availableProcessors.displayName: Available Processors
 cpu.availableProcessors.type: gauge
 cpu.availableProcessors.unit: none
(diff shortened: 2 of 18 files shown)


bstansberry:
@jmesnil Is getting rid of io.prometheus:simpleclient going to be part of this work? If not is there a WFLY for that?

jmesnil:
@bstansberry I'll use this branch as the basis to remove io.prometheus:simpleclient but it'll be tracked by a separate WFLY issue and PR

I've put this branch in draft mode as I still need to add the mp-metrics TCK test to it but there should be no other changes to the subsystem codebase (provided the TCK tests pass)

jmesnil:
Tests from MicroProfile Metrics TCK (both API & Rest) are now run in WildFly test suite: e.g. [all tests in `micropropfile.metrics` package on Linux JDK11 job](https://ci.wildfly.org/viewLog.html?buildId=158832&buildTypeId=WF_PullRequest_LinuxJdk11&tab=testsInfo&branch_WF_PullRequest=pull%2F12437)

jmesnil:
@bstansberry re the removal of Prometheus client, I've opened https://issues.jboss.org/browse/WFLY-12311 to track it.
I had a quick glance at it and it will affect the name of the metrics (mp-metrics has the awesome idea to prepend all metrics with their scope (`base`, `vendor`, `application`).
That means that all metrics from subsystem would be prepended with `vendor_`.

rsvoboda (commented):
I miss docs update regarding MP Metrics 2.0, there is docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Metrics.adoc

This is breaking change and changes in behavior should be highlighted  
* The `@Counted` annotation no longer includes the monotonic parameter.
* Metrics are no longer identified by their String name, but by a MetricID class. The MetricID is a String name and a set of tags.
* Some of the base metrics (defined by the specification) changed their type from Counter to Gauge.
* Some of the base metrics were renamed. 
* Accumulating counters now end with total.
* In the OpenMetrics output format (formerly known as Prometheus format), the metric scope and name are no longer separated by :, but by _.

(Info taken from Thorntail, relevant for WF too)

Maybe separate document to track MP breaking changes would be good idea.

rsvoboda:
>  That means that all metrics from subsystem would be prepended with vendor_.

So instead of `wildfly_undertow_request_count_total` we would have `vendor_wildfly_undertow_request_count_total` or `vendor_undertow_request_count_total` ?

jmesnil:
@rsvoboda the admin guide only documents the subsystem. It does not document the user API and its changes.

When we release WildFly 18, we could link to https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/changelog.adoc that list the changes introduced by Metrics 2.0.x and https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/appendix.adoc#migration-hint-to-20 that helps migrating existing Metrics 1.x application

pom.xml
```diff
@@ -268,10 +268,11 @@
         <version.io.opentracing.tracerresolver>0.1.5</version.io.opentracing.tracerresolver>
         <version.io.opentracing.servlet>0.1.0</version.io.opentracing.servlet>
         <version.io.reactivex.rxjava>2.2.2</version.io.reactivex.rxjava>
+        <version.io.rest-assured>3.0.0</version.io.rest-assured>
         <version.io.prometheus.simpleclient>0.6.0</version.io.prometheus.simpleclient>
         <version.io.smallrye.smallrye-config>1.3.6</version.io.smallrye.smallrye-config>
         <version.io.smallrye.smallrye-health>1.0.2</version.io.smallrye.smallrye-health>
-        <version.io.smallrye.smallrye-metrics>1.1.3</version.io.smallrye.smallrye-metrics>
+        <version.io.smallrye.smallrye-metrics>1.1.0</version.io.smallrye.smallrye-metrics>
```
jstourac: is this downgrade change intentional?

jmesnil: @jstourac it is not a downgrade. The smallrye-metrics component changes its GAV convention, we upgraded from `io.smallrye:smallrye-metrics:1.1.3` (that covers MP Metrics 1.x) to `io.smallrye:smallrye-metrics-2.0:1.1.0` (that covers MP Metrics 2.x).

I updated the `artifactId` in the pom.xml but did not change the name of the property that sets the version. sorry for the confusion :)

feature-pack/pom.xml
```diff
@@ -3020,7 +3020,7 @@
 
         <dependency>
             <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-metrics</artifactId>
+            <artifactId>smallrye-metrics-2.0</artifactId>
```
kabir: I see you adding a license for this, but should the license for io.smallrye:smallrye-metrics-2.0 also be removed?

microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
```diff
@@ -1,18 +1,18 @@
...
-classloader.totalLoadedClass.count.unit: none
-classloader.totalLoadedClass.count.description: Displays the total number of classes that have been loaded since the Java virtual machine has started execution.
-classloader.totalLoadedClass.count.mbean: java.lang:type=ClassLoading/TotalLoadedClassCount
-classloader.totalUnloadedClass.count.displayName: Total Unloaded Class Count
-classloader.totalUnloadedClass.count.type: counter
-classloader.totalUnloadedClass.count.unit: none
-classloader.totalUnloadedClass.count.description: Displays the total number of classes unloaded since the Java virtual machine has started execution.
-classloader.totalUnloadedClass.count.mbean: java.lang:type=ClassLoading/UnloadedClassCount
+classloader.loadedClasses.count.displayName: Current Loaded Class Count
+classloader.loadedClasses.count.type: gauge
```
kabir: I presume this 'gauge' value is intentional?

pom.xml
```diff
@@ -5121,6 +5129,16 @@
                 <artifactId>microprofile-metrics-api</artifactId>
                 <version>${version.org.eclipse.microprofile.metrics.api}</version>
             </dependency>
+            <dependency>
+                <groupId>org.eclipse.microprofile.metrics</groupId>
+                <artifactId>microprofile-metrics-api-tck</artifactId>
+                <version>${version.org.eclipse.microprofile.metrics.api}</version>
```
kabir: I think this should be `<scope>test</scope>`?

pom.xml
```diff
@@ -5121,6 +5129,16 @@
...
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.microprofile.metrics</groupId>
+                <artifactId>microprofile-metrics-rest-tck</artifactId>
+                <version>${version.org.eclipse.microprofile.metrics.api}</version>
```
kabir: I think this should be `<scope>test</scope>`?

microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
```diff
@@ -83,15 +80,15 @@ memory.usedNonHeap.unit: bytes
 thread.count.mbean: java.lang:type=Threading/ThreadCount
 thread.count.description: Number of currently deployed threads
 thread.count.unit: none
-thread.count.type: counter
+thread.count.type: gauge
```
kabir: Again here and for the rest of the file, I guess the replacement of 'counter' with 'gauge' is intentional?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
```diff
@@ -110,10 +113,10 @@ public void testJvmUptimeMetric() throws Exception {
...
 
     private void checkJvmUptime(JsonObject jvmUptime) {
-        assertEquals(jvmUptime.getString("type"), "gauge");
-        assertEquals(jvmUptime.getString("unit"), "milliseconds");
-        assertEquals(jvmUptime.getString("displayName"), "JVM Uptime");
-        assertEquals(jvmUptime.getString("description"), "Displays the uptime of the Java virtual machine");
+        assertEquals("gauge", jvmUptime.getString("type"));
+        assertEquals("milliseconds", jvmUptime.getString("unit"));
+        assertEquals("JVM Uptime", jvmUptime.getString("displayName"));
+        assertEquals("Displays the uptime of the Java virtual machine", jvmUptime.getString("description"));
```
kabir: Seems to come from base-metrics.properties. Are those meant to be localised? If they are this test might fail in the future. It isn't introduced by this PR though, so it is probably fine

kabir (commented):
Apart from the small stuff I found it looks ok to me

feature-pack/pom.xml
jmesnil: sorry, that's not in the commit message but when I did this upgrade, I realised that the license for `io.smallrye:smallrye-metrics` was missing.

That's why the PR adds the license for `io.smallrye:smallrye-metrics-2.0` and does remove any other one.

microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
jmesnil: that's a change in the spec: https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/required-metrics.adoc#classloading-jvm-stats

microprofile/metrics-smallrye/src/main/resources/io/smallrye/metrics/base-metrics.properties
jmesnil: yes, it is intentional to comply with the spec https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/required-metrics.adoc#thread-jvm-stats

jmesnil: likewise for `thread.daemon.count.type` and `thread.max.count.type` below

pom.xml
jmesnil: you're right, I'll fix it. They are in the `test` scope in testsuite/integration/microprofile-tck/metrics/pom.xml but they should also have this scope in the dependencyManagement.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/metadata/MicroProfileMetricsMetadataTestCase.java
jmesnil: I don't know if they will ever be localized (the consumers of these metrics are other systems such as Prometheus)  In any case, I think that's out of scope of this upgrade.

feature-pack/pom.xml
kabir: @psakar @dosoudil any idea?

PR #12439, opened 2019-07-19
[JBJCA-1389] - NullPointerException raised when calling isWrapperFor(…

adamecp:
Description of issue: When calling Connection.isWrapperFor(...) on a connection which has been closed (i.e. returned to the pool, managed connection disassociated) previously, a NullPointerException is raised. The exception should be "Connection is not associated with a managed connection: ..."

Jira issue: [https://issues.jboss.org/browse/JBJCA-1389](https://issues.jboss.org/browse/JBJCA-1389)


kabir (changes requested):
@adamecp The new test fails

bstansberry:
@maeste My guess is these tests fail because an IJ upgrade they depend on is not in place.  Could you incorporate this commit into whatever branch you use for the upgrade so the test and the upgrade can be processed together?  Thanks.

PR #12441, opened 2019-07-19
[WFLY-12307] Fix the build-legacy module by combining the two profile…

jamezp:
…s of the same name into one.

https://issues.jboss.org/browse/WFLY-12307

Remove the extra profile as only one will be activated since they have the same name. We may also be to a point where we could consider removing the `build-legacy` and `dist-legacy` modules.

### DIFF
--- a/build-legacy/pom.xml
+++ b/build-legacy/pom.xml
@@ -215,18 +215,6 @@
                             </execution>
                         </executions>
                     </plugin>
-                </plugins>
-            </build>
-        </profile>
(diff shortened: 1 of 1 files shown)


kabir:
@jamezp Not sure if I am missing something but I see two profiles with different names activated by different properties

jamezp:
Um, yes you're right. I'll close this. I did have issues building this, but I'll look at why. Really we probably should just use the `legacy-release` profile and property.

PR #12449, opened 2019-07-24
WFLY-12321

fl4via:
Do not merge yet

Jira: https://issues.jboss.org/browse/WFLY-12321

This PR has to incorporate a few more feedbacks and will most likely be broken into more PRs and Jiras.
EAP Jira (temporary): https://issues.jboss.org/browse/JBEAP-17049

### DIFF
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
@@ -79,6 +79,11 @@ public class BeanExpirationScheduler<I, T> implements Scheduler<I> {
     }
 
     @Override
+    public void prepareRescheduling(I id) {
+        cancel(id);
+    }
+
+    @Override
     public void cancel(I id) {
         Future<?> future = this.expirationFutures.remove(id);
         if (future != null) {
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/EagerEvictionScheduler.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/EagerEvictionScheduler.java
@@ -82,6 +82,11 @@ public class EagerEvictionScheduler<I, T> implements Scheduler<I>, BeanGroupEvic
     }
 
     @Override
+    public void prepareRescheduling(I id) {
+        cancel(id);
+    }
+
+    @Override
     public void cancel(I id) {
         Future<?> future = this.evictionFutures.remove(id);
         if (future != null) {
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/ExpirationTracker.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/ExpirationTracker.java
@@ -1,0 +1,254 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.clustering.ejb.infinispan;
+
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Tracks expiration of multiple id objects.
+ *
+ * @author Flavia Rainone
+ */
+class ExpirationTracker<I> {
+
+    // the expiration timeout in milisseconds
+    private final long timeout;
+    // all expiration infos being tracked mapped by id
+    private final Map<I, ExpirationInfo> trackedIds;
+    // the first expiration to occur
+    private ExpirationInfo firstExpiration;
+    // the last expiration to occur
+    private ExpirationInfo lastExpiration;
+    // expiration info: a single node in a double linked list
+    private class ExpirationInfo {
+        ExpirationInfo nextExpiration;
+        ExpirationInfo previousExpiration;
+        volatile long expiration;
+        final I id;
+
+        ExpirationInfo(I id) {
+            this.id = id;
+        }
+    }
+
+    /**
+     * Creates an ExpirationTracker with configured {@code timeout} for expiration
+     * @param timeout the time it takes for an id to expire.
+     */
+    ExpirationTracker(Duration timeout) {
+        this.timeout = timeout.toMillis();
+        trackedIds = new HashMap<>();
+    }
+
+    /**
+     * Tracks expiration of bean identified by {@code id} starting the count down
+     * now.
+     *
+     * @param id the id of the bean that will expire
+     */
+    void trackExpiration(I id) {
+        final ExpirationInfo expiration;
+        synchronized (this) {
+            // if we are already tracking this id, move the id to the lastExpiration
+            // position, as it will be the last id to expire;
+            if (trackedIds.containsKey(id)) {
+                expiration = trackedIds.get(id);
+                if (expiration != lastExpiration) {
+                    // remove the node from the expiration linked list
+                    if (expiration.nextExpiration != null) {
+                        expiration.nextExpiration.previousExpiration = expiration.previousExpiration;
+                    }
+                    if (expiration.previousExpiration != null) {
+                        expiration.previousExpiration.nextExpiration = expiration.nextExpiration;
+                    } else if (firstExpiration == expiration) {
+                        firstExpiration = firstExpiration.nextExpiration;
+                    }
+                    expiration.previousExpiration = null;
+                    expiration.nextExpiration = null;
+                }
+            } else {
+                // if we are not tracking the id at all, add it to the expiration list
+                expiration = new ExpirationInfo(id);
+                trackedIds.put(id, expiration);
+            }
+            // if the list is empty, now add a single expiration node
+            if (lastExpiration == null) {
+                firstExpiration = lastExpiration = expiration;
+            } else if (lastExpiration != expiration) {
+                // if the last expiration to occur is not current node, add it to the
+                // last of the expiration list
+                expiration.previousExpiration = lastExpiration;
+                lastExpiration.nextExpiration = expiration;
+                lastExpiration = expiration;
+            }
+        }
+        // calculate expiration time
+        synchronized(expiration) {
+            expiration.expiration = System.currentTimeMillis() + timeout;
+        }
+    }
+
+    /**
+     * If expiration is being tracked for {@code id}, invalidates it.
+     * This method does not remove the invalidated bean from this tracker and
+     * it is expected that expiration will be reescheduled or canceled in the future.
+     *
+     * @param id the id whose expiration, if existent, will be invalidated
+     */
+    void invalidateExpiration(I id) {
+        final ExpirationInfo expiration;
+        synchronized (this) {
+            // if we are already tracking this id, mark its expiration to -1
+            if (trackedIds.containsKey(id)) {
+                expiration = trackedIds.get(id);
+                synchronized (expiration) {
+                    expiration.expiration = -1;
+                }
+            }
+        }
+    }
+
+    /**
+     * Retry expiration of a failed to expire id. If id is already being
+     * tracked, do nothing; if not, add it to this tracker as the next to
+     * expire id.
+     *
+     * @param id the id of the bean that failed to expire
+     */
+    void retryExpiration(I id) {
+        synchronized (this) {
+            // do nothing if bean is already being tracked
+            if (trackedIds.containsKey(id)) {
+                return;
+            }
+            final ExpirationInfo expirationInfo = new ExpirationInfo(id);
+            // bean is expired already
+            synchronized (expirationInfo) {
+                expirationInfo.expiration = 0;
+            }
+            expirationInfo.nextExpiration = firstExpiration;
+            if (firstExpiration != null) {
+                firstExpiration.previousExpiration = expirationInfo;
+            }
+            firstExpiration = expirationInfo;
+            if (lastExpiration == null) {
+                lastExpiration = expirationInfo;
+            }
+            trackedIds.put(id, expirationInfo);
+        }
+    }
+
+    /**
+     * Stop tracking the expiration of the bean identified by {@code id},
+     * regardless of whether is has already expired or not.
+     *
+     * @param id the id of the bean that will no longer expire
+     */
+    void forget(I id) {
+        synchronized (this) {
+            if (!trackedIds.containsKey(id)) {
+                return;
+            }
+            final ExpirationInfo expirationInfo = trackedIds.remove(id);
+            if (expirationInfo.previousExpiration != null) {
+                expirationInfo.previousExpiration.nextExpiration = expirationInfo.nextExpiration;
+            }
+            if (expirationInfo.nextExpiration != null) {
+                expirationInfo.nextExpiration.previousExpiration = expirationInfo.previousExpiration;
+            }
+            if (firstExpiration == expirationInfo) {
+                firstExpiration = expirationInfo.nextExpiration;
+            }
+            if (lastExpiration == expirationInfo) {
+                lastExpiration = expirationInfo.previousExpiration;
+            }
+        }
+    }
+
+    /**
+     * Returns an id that has expired. Automatically removes the expired id from
+     * this tracker.
+     *
+     * @param currentTimeInMillis the current time in milliseconds (provided to
+     *                            prevent multiple calls to System.currentTimeInMillis())
+     *
+     * @return the id of the expired bean; {@code null} if no id has expired
+     */
+    I getExpiredId(long currentTimeInMillis) {
+        synchronized (this) {
+            // skip any expiration that is invalidated (i.e., expiration is -1)
+            ExpirationInfo expiration = this.firstExpiration;
+            while (expiration != null && readExpiration(expiration) == -1) {
+                expiration = expiration.nextExpiration;
+            }
+            // if first expiration has expired
+            if (expiration != null && readExpiration(expiration) <= currentTimeInMillis) {
+                final I expiredId = expiration.id;
+                if (expiration.previousExpiration != null) {
+                    expiration.previousExpiration.nextExpiration = expiration.nextExpiration;
+                }
+                if (expiration.nextExpiration != null) {
+                    expiration.nextExpiration.previousExpiration = expiration.previousExpiration;
+                }
+                if (firstExpiration == expiration) { // it might not be case if expiration.expiration is -1
+                    firstExpiration = expiration.nextExpiration;
+                }
+                if (lastExpiration == expiration) { // dont set it to null if expiration.previousExpiration is not null
+                    lastExpiration = expiration.previousExpiration;
+                }
+                trackedIds.remove(expiredId);
+                return expiredId;
+            }
+            return null;
+        }
+    }
+
+    private long readExpiration(ExpirationInfo info) {
+        synchronized(info) {
+            return info.expiration;
+        }
+    }
+
+    /**
+     * Returns all ids whose expiration is being tracked.
+     *
+     * @return an  {@code unmodifiable set} containing all beans.
+     */
+    Collection<I> getTrackedIds() {
+        return Collections.unmodifiableSet(trackedIds.keySet());
+    }
+
+    /**
+     * Returns the next expiration time to occur.
+     * @return when will the first expiration occur in milliseconds; returns
+     *         {@code -1} if no ids are being tracked. Notice that the expiration
+     *         may have already occurred or may still occur (i.e., this tracker
+     *         does not compare the expiration with current time; it just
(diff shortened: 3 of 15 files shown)


clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/InfinispanBeanManager.java
```diff
@@ -286,7 +286,6 @@ Node locatePrimaryOwner(I id) {
             InfinispanEjbLogger.ROOT_LOGGER.debugf("Could not find bean %s", id);
             return null;
         }
-        this.schedule(bean);
```
pferraro: There needs to be be some action taken at the point to prevent concurrent expiration, in particular, when the bean is schedule to be expired by a different member than the one currently performing the invocation.

clustering/api/src/main/java/org/wildfly/clustering/dispatcher/CommandDispatcher.java
```diff
@@ -44,6 +44,14 @@
      */
     C getContext();
 
+    /**
+     * Returns true if {@code member} is current node
+     *
+     * @param member member node
+     * @return true if {@code member} is this local node
+     */
+    boolean isLocal(Node member);
```
pferraro: We are removing this, correct?

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
```diff
@@ -38,46 +36,49 @@
...
 public class BeanExpirationScheduler<I, T> implements Scheduler<I> {
-    final Map<I, Future<?>> expirationFutures = new ConcurrentHashMap<>();
-    final Batcher<TransactionBatch> batcher;
-    final BeanRemover<I, T> remover;
-    final ExpirationConfiguration<T> expiration;
+    private final Batcher<TransactionBatch> batcher;
+    private final BeanRemover<I, T> remover;
+    private final ExpirationConfiguration<T> expiration;
+    private final ExpirationTracker<I> expirationTracker;
+    private Future<?> expireTask;
```
pferraro: expireTask should be volatile, no?

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
```diff
@@ -38,46 +36,49 @@
...
-            synchronized (task) {
-                this.expirationFutures.put(id, this.expiration.getExecutor().schedule(task, timeout.toMillis(), TimeUnit.MILLISECONDS));
+        if (expirationTracker != null) {
+            expirationTracker.trackExpiration(id);
+            if (expireTask == null) {
+                InfinispanEjbLogger.ROOT_LOGGER.tracef("Scheduling stateful session bean %s to expire in %s", id, timeout);
+                Runnable task = new ExpirationTask();
+                // Make sure the expiration future map insertion happens before map removal (during task execution).
+                synchronized (this) {
+                    expireTask = this.expiration.getExecutor().schedule(task, timeout.toMillis(), TimeUnit.MILLISECONDS);
```
pferraro: I think you want an additional null check here - since multiple concurrent threads can reach this synchronized block resulting in multiple expire task creation.
e.g.
```java
synchronized (this) {
   if (expireTask == null) {
       expireTask = ...;
   }
}
```

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
```diff
@@ -87,51 +88,59 @@ public void cancel(Locality locality) {
...
     public void close() {
-        for (Future<?> future: this.expirationFutures.values()) {
-            future.cancel(false);
+        final Future<?> task;
+        synchronized (this) {
+            if (expireTask == null) {
+                return;
+            }
+            task = expireTask;
+            expireTask = null;
```
pferraro: Should we be concerned that the expireTask can be reset after this synchronized block completes?  Whose cancels the Future in that case?

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/InfinispanBeanManager.java
```diff
@@ -229,8 +234,13 @@ private void cancel(Bean<I, T> bean) {
...
     }
 
-    void schedule(Bean<I, T> bean) {
+    private void schedule(Bean<I, T> bean) {
         if (this.dispatcher != null) {
+            final Node node = locatePrimaryOwner(bean.getId());
+            if (dispatcher.isLocal(node)) {
+                scheduler.schedule(bean.getId());
+                return;
+            }
```
pferraro:
```java
   if (this.dispatcherFactory.getGroup().isSingleton()) {
       scheduler.cancel(bean.getId());
       return;
   }
```

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/InfinispanBeanManager.java
```diff
@@ -407,13 +417,17 @@ public T acquire() {
 
         @Override
         public boolean release() {
+            schedule(bean);
```
pferraro: I don't think it makes sense to schedule expiration/passivation on release.  The last access time isn't updated until close(), and we already schedule for expiration/passivation then.

ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
```diff
@@ -169,6 +169,8 @@ public void release(V bean) {
                     // However, removing it directly is faster than scheduling it for immediate removal.
                     remove(id);
                 }
+            } else {
+                remove(id);
```
pferraro: This doesn't look right.  This would mean that a @Stateful EJB never retains its state across multiple invocations if it doesn't define a @StatefulTimeout.

pferraro (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
fl4via: I'll see, I'll skip this one

fl4via:
@pferraro I've updated the PR with all your feedback. Also, I removed the commit for "At LocalCommandDispatcher for clustering, executeOnMember should check if this node is equal to member, and not if it is equal to itself". I was thinking it shoudl have its own Jira, dont you agree? Or do you prefer to have the commit at this PR as it was before?

pferraro:
@fl4via 
> I've updated the PR with all your feedback. Also, I removed the commit for "At LocalCommandDispatcher for clustering, executeOnMember should check if this node is equal to member, and not if it is equal to itself". I was thinking it shoudl have its own Jira, dont you agree? Or do you prefer to have the commit at this PR as it was before?

Agreed. I merged it last week.
https://github.com/wildfly/wildfly/pull/12445

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/Scheduler.java
```diff
@@ -35,6 +35,13 @@
      */
     void schedule(I id);
 
+    /**
+     * Invalidates a previously scheduled task for specified bean whose task
+     * will be rescheduled again.
+     * @param id a bean identifier
+     */
+    void prepareRescheduling(I id);
```
pferraro: Why create a new method?  Aren't we just changing the behavior of the existing cancel(...) method?

pferraro (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/dispatcher/bean/CommandDispatcherBean.java
```diff
@@ -34,6 +34,11 @@ public void destroy() {
         this.close();
     }
 
+    @Override
+    public boolean isLocal(Node node) {
```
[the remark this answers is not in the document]
fl4via: Somehow I must have failed to pull everything yesterday... I'll fix it, thanks for pointing this :-)

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/Scheduler.java
fl4via: We are not, because cancel will invoke ExpirationTracker.forget, that will remove all entries from ExpirationTracker. If we keep calling cancel like before the gain is relatively small. I'm trying to make sure we schedule or cancel only once per call, I'm not sure if that's entirely feasible, but I'm trying to at least decrease the amount of times we add items to the tracker, to prevent the cost of invoking currentTimeInMillis

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/InfinispanBeanManager.java
fl4via: that's a great catch, I updated the PR with this

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
fl4via: I agree, I fixed that by checking if expireTask is different from null at the end of the file, before trying to reschedule it and reset the field.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
fl4via: +1

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/InfinispanBeanManager.java
fl4via: I added prepareRescheduling method below, btw, all methods and names are candidates for renaming if the name doesnt sound right, it is hard to get it right sometimes when you're not a native speaker ;-)

clustering/api/src/main/java/org/wildfly/clustering/dispatcher/CommandDispatcher.java
fl4via: +1 this is done

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
fl4via: done

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/Scheduler.java
pferraro: Ah, I see.  I assumed that there was no longer a use case for ExpirationTracker.forget, but there is.  So, cancel(I) is still triggered from cancel(Locaility, Locality) and from SchedulableBean.close() in the event that the bean is invalid (i.e. was removed).

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/PrepareReeschedulingSchedulerCommand.java
```diff
@@ -0,0 +1,48 @@
...
+import org.wildfly.clustering.dispatcher.Command;
+import org.wildfly.clustering.ejb.Bean;
+
+/**
+ * Command that temporarily holds off the sheduling of a session, preventing
+ * the cost of a cancellation. The scheduling must be canceled or rescheduled at
+ * the near future.
+ * @author Flavia Rainone
+ */
+public class PrepareReeschedulingSchedulerCommand<I> implements Command<Void, Scheduler<I>> {
```
pferraro: Spelling fix: PrepareR**e**schedulingSchedulerCommand :)

fl4via: Ooops, thanks :-)

rhusar:
With the latest changes `org.jboss.as.test.clustering.cluster.ejb.stateful.StatefulTimeoutTestCase#timeout` fails consistently.

fl4via:
Retest this please

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/BeanExpirationScheduler.java
```diff
@@ -38,46 +36,57 @@
...
-        if (future != null) {
-            future.cancel(false);
-        }
+        this.expirationTracker.forget(id);
     }
 
     @Override
     public void cancel(Locality locality) {
-        for (I id: this.expirationFutures.keySet()) {
+        for (I id: expirationTracker.getTrackedIds()) {
```
pferraro: Iterating over this set will be prone to ConcurrentModificationExceptions.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/EagerEvictionScheduler.java
```diff
@@ -40,53 +40,66 @@
...
-            future.cancel(false);
-        }
+        expirationTracker.forget(id);
+        groupIdMap.remove(id);
     }
 
     @Override
     public void cancel(Locality locality) {
-        for (I id: this.evictionFutures.keySet()) {
+        for (I id: this.expirationTracker.getTrackedIds()) {
```
pferraro: Iterating over this set will be prone to ConcurrentModificationExceptions.

pferraro (changes requested):

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/ExpirationTracker.java
```diff
@@ -0,0 +1,253 @@
...
+                if (expiration.nextExpiration != null) {
+                    expiration.nextExpiration.previousExpiration = expiration.previousExpiration;
+                }
+                if (expiration.previousExpiration != null) {
+                    expiration.previousExpiration.nextExpiration = expiration.nextExpiration;
+                } else if (firstExpiration == expiration &&
+                        firstExpiration != lastExpiration) {
+                    firstExpiration = firstExpiration.nextExpiration;
+                }
+                expiration.previousExpiration = null;
```
stuartwdouglas: the nextExpriation must also me cleared, or an endless loop can result

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/ExpirationTracker.java
```diff
@@ -0,0 +1,254 @@
...
+                // if we are not tracking the id at all, add it to the expiration list
+                expiration = new ExpirationInfo(id);
+                trackedIds.put(id, expiration);
+            }
+            // if the list is empty, now add a single expiration node
+            if (lastExpiration == null) {
+                firstExpiration = lastExpiration = expiration;
+            } else if (lastExpiration != expiration) {
+                // if the last expiration to occur is not current node, add it to the
+                // last of the expiration list
```
pferraro: This logic makes the assumption that the newly tracked bean will be the last to expire.  While this is true when scheduling a bean for expiration at the end of an invocation, this is generally not the case when a member takes on the responsibility for expiring a given bean, i.e. when triggered from InfinispanBeanManager.schedule(Locality oldLocality, Locality newLocality).

pferraro (changes requested):

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/ExpirationTracker.java
fl4via: Oh, I see. I'll try to think on a solution and I'll assert the perf gains with the benchmark to make sure this PR is really needed.

pferraro: One solution is to replace the expiration tracker with a collection, e.g. io.undertow.util.ConcurrentDirectDeque.  We can then construct the scheduler with a queue implementation appropriate for the use case.  In the clustered case, where the assumption above is violated, we could use a priority queue, where elements are inserted according to their next expiration, rather then just appending to the end.  Insertions/removals would be slower log(N) instead of constant time, but that way we can still optimize for the local case.

fl4via:
@pferraro this commit is ready for review

fl4via:
Retest this please

fl4via:
Retest this please

fl4via:
Retest this please

bstansberry:
@fl4via @pferraro Is this meant for WF 18?

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+        this.dispatcher.close();
+    }
+
+    @Override
+    public void evict(I id) {
+        try {
+            // Cache eviction is a local operation, so we need to broadcast this to the cluster
+            this.dispatcher.executeOnGroup(new EvictCommand<>(id));
```
pferraro: Given that this is now a local-only scheduler, there is no need to dispatch an eviction command.

pferraro (changes requested):

pferraro:
@bstansberry This is a performance optimization for local SFSBs.  Not essential for WF18, but still desirable.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
```diff
@@ -0,0 +1,188 @@
...
+    private volatile Future<?> evictionTask;
+
+    private final CommandDispatcher<BeanGroupEvictor<I>> dispatcher;
+
+    NonClusteredEagerEvictionScheduler(BeanFactory<I, T> factory, BeanGroupEvictor<I> evictor, ScheduledExecutorService executor, Duration idleTimeout, CommandDispatcherFactory dispatcherFactory, String dispatcherName) {
+        this.factory = factory;
+        this.executor = executor;
+        this.idleTimeout = idleTimeout;
+        this.expirationTracker = new ExpirationTracker<>(idleTimeout);
+        this.dispatcher = dispatcherFactory.createCommandDispatcher(dispatcherName + "/eviction", evictor);
```
pferraro: This scheduler no longer has any need to create a CommandDispatcher, no?

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+        this.dispatcher.close();
+    }
+
+    @Override
+    public void evict(I id) {
+        // do nothing, this is a local-only scheduler
+    }
+
+    static class EvictCommand<I> implements Command<Void, BeanGroupEvictor<I>> {
```
pferraro: I don't think this class is referenced any longer.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
```diff
@@ -0,0 +1,188 @@
...
+            } catch (ExecutionException e) {
+                // Ignore
+            }
+        }
+        this.dispatcher.close();
+    }
+
+    @Override
+    public void evict(I id) {
+        // do nothing, this is a local-only scheduler
```
pferraro: This method needs to evict the bean, i.e. `this.evictor.evict(id);`

pferraro (changes requested):

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
fl4via: done

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
fl4via: done

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/NonClusteredEagerEvictionScheduler.java
fl4via: done

fl4via:
@pferraro  all requested changes are done

pferraro (approved):
Thanks Flavia. I think this is ready now.

PR #12450, opened 2019-07-25
[WFLY-5286]: Provide CLI operation which returns used journal type - NIOxASYNCIO

ehsavoie:
Adding a new runtime-journal-type attribute to define the effective journal type being used.

Jira: https://issues.jboss.org/browse/WFLY-5286

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
@@ -1,0 +1,176 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2011, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.test.manualmode.messaging;
+
+import static org.jboss.as.controller.client.helpers.ClientConstants.VALUE;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.UnknownHostException;
+import org.jboss.arquillian.container.test.api.ContainerController;
+
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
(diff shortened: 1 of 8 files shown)


testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+    protected static ContainerController container;
+
+    @Test
+    public void testReadFileJournaltype() throws IOException {
+        if (!container.isStarted(DEFAULT_FULL_JBOSSAS)) {
+            container.start(DEFAULT_FULL_JBOSSAS);
+        }
+        ManagementClient managementClient = createManagementClient();
+        assertEquals("ASYNCIO", readJournalType(managementClient));
+        assertNotNull(readRuntimeJournalType(managementClient));
```
tterem: Is there a reason why we only check this is not null?
What about `assertEquals("ASYNCIO", readRuntimeJournalType(managementClient));` instead ?
Before falling back to NIO, these attributes should be the same, right?

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    public void testReadDatabaseJournaltype() throws IOException {
+        if (!container.isStarted(DEFAULT_FULL_JBOSSAS)) {
+            container.start(DEFAULT_FULL_JBOSSAS);
+        }
+        ManagementClient managementClient = createManagementClient();
+        assertEquals("ASYNCIO", readJournalType(managementClient));
+        assertNotNull(readRuntimeJournalType(managementClient));
```
tterem: Same as above

tterem:
@ehsavoie This PR contains two commits, but only https://github.com/wildfly/wildfly/pull/12450/commits/27ba97dafe6b631c537e22fe3b135a93b6a9fe7a implements new feature. I suggest to create separate PR (and issue) for https://github.com/wildfly/wildfly/pull/12450/commits/21eeae7bcd71d89795a101eb5c3c8cbbdc76e62c.
cc @marekkopecky

bstansberry:
Best practice is that a model/schema version bump is a separate commit so the boilerplate for the bump is separate from the change that forces it, thus making the important part of the change more clear. It is better that the version bump commit should be part of the first PR that forces the bump, as the bump is part of the overall piece of work. We don't want to merge version bumps and then never merge any change that justifies them.

My only rule about the JIRA referenced in the version bump commit is that it must not be for some RFE unrelated to the PR. A separate JIRA for the bump is ok, or using the RFE JIRA that requires the bump is ok. A separate JIRA for the bump is a good strategy if there are going to be several RFEs and you don't know which PR is going to be the one that gets merged first.

marekkopecky (changes requested):
@bstansberry Thank you for your clarification.

> A separate JIRA for the bump is a good strategy if there are going to be several RFEs 

AFAIK, there is at least these two (but probably more):
* https://github.com/wildfly/wildfly-proposals/pull/228
* https://github.com/wildfly/wildfly-proposals/pull/225

This RFE (AFAIK) adds new attribute `runtime-journal-type` only. I'm looking to the diff between messaging-activemq schema 7 and messaging-activemq schema 8. New attribute is not mentioned (@ehsavoie probably ok because this attribute is runtime attribute?), but there are some changes related with credential store and these changes seems unrelated with this RFE (@ehsavoie right? can you link the jira where CS changes are described?), but still part of this PR. But this PR should contains changes related with this RFE only, so I reject this PR for now ...

ehsavoie:
A Jira for the version has been created https://issues.jboss.org/browse/WFLY-12326 . I'll update the commit with it as the XML bump is not required.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
ehsavoie: because we can't presume of the value as on Windows it would always be NIO for example

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/RuntimeJournalTypeManagementTestCase.java
ehsavoie: same as above, we can't presume of the value

marekkopecky:
@ehsavoie So it seems that this PR is currently blocked by https://github.com/wildfly/wildfly/pull/12453 ... right?

ehsavoie:
@marekkopecky yes

tterem (approved):

bstansberry (approved):

PR #12453, opened 2019-07-26
[WFLY-12326]: Messaging subsystem model version bump.

ehsavoie:
* Updating Model Version to 8.0.0

Jira: https://issues.jboss.org/browse/WFLY-12326

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
@@ -98,7 +98,14 @@ import org.wildfly.extension.messaging.activemq.jms.ExternalPooledConnectionFact
  * Domain extension that integrates Apache ActiveMQ 6.
  *
  * <dl>
- * <dt><strong>Current</strong> - WildFly 17</dt>
+ * <dt><strong>Current</strong> - WildFly 18</dt>
+ *   <dd>
+ *     <ul>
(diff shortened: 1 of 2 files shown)


marekkopecky:
Should this PR adds new xsd schema, like `./src/main/resources/schema/wildfly-messaging-activemq_8_0.xsd`?

ehsavoie:
@marekkopecky no  this is just a model version bump, no schema changes are required since only runtime attributes are added

marekkopecky:
@ehsavoie  (cc @bstansberry ): So, I see these two options now based on your previous comment:

* we are going to use 8.0 model version, but 7.0 xml schema (== `<subsystem xmlns="urn:jboss:domain:messaging-activemq:7.0">` in standalone-full.xml). In my PoV, usability of this proposal is not good, this could be confusing.
* we are going to use `xmlns="urn:jboss:domain:messaging-activemq:8.0"` but we expect that this xml still uses 7.0 xsd. So if WF users want to know xsd details about 8.0 configuration, they should check 7.0 xsd. In my PoV, usability of this is not good as well.

Maybe I'm missing something, but TBH this is really not clear to me.

ehsavoie:
@marekkopecky we are using your 1st option. The configuration doesn't change but the server model has changed we we change the server model version not the server configuration version.

bstansberry:
We don't increase xsd versions if there is no change in the xsd. This is the WF policy since the AS 7 days.  There are other subsystems where the numbers aren't the same.

PR #12459, opened 2019-07-30
WFLY-12334 Using hotrod session manager with near cache throws IllegalAccessError

pferraro:
https://issues.jboss.org/browse/WFLY-12334

N.B. The changes here are temporary, until Infinispan 9.4.16.Final is released and integrated.


bstansberry:
@pferraro Is this going to be reverted before WF 18.0.0.Final? If yes, how about Beta?

pferraro:
@bstansberry  It's fine to include this in a beta, but I'll want to restore the current module packaging when I upgrade to Infinispan 9.4.16.Final before WF 18.0.0.Final.

pferraro:
Infinispan 9.4.16.Final was released - so the workaround in this PR will no longer be necessary.

PR #12460, opened 2019-07-30
WFLY-12265 Warning when singleton EJB implements SessionBean interface

TomasHofman:
https://issues.jboss.org/browse/WFLY-12265


TomasHofman:
Looks test case is failing with security manager.

chengfang (changes requested):
I think the warning message should indicate that implementing SessionBean interface is ignored (the action taken by ejb container), and what the application should do to fix it (remove the implement SessionBean clause).

TomasHofman:
@chengfang I updated the message to:

> [EJB3.2 spec, section 4.9.2] Singleton session beans are not allowed to implement 'javax.ejb.SessionBean' interface. This interface on bean '%s' is going to be ignored and should be removed.

chengfang (approved):

TomasHofman:
Rebased.

PR #12461, opened 2019-07-30
WFLY-12165 Expose management metrics for HotRod caches

pferraro:
https://issues.jboss.org/browse/WFLY-12165

Analysis document:
https://github.com/wildfly/wildfly-proposals/pull/230


clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/remote/RemoteCacheContainerOperation.java
```diff
@@ -0,0 +1,83 @@
...
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+/**
+ * @author Paul Ferraro
+ */
+public enum RemoteCacheContainerOperation implements Operation<RemoteCacheManagerMXBean> {
```
bstansberry: This isn't part of the analysis and AFAICT isn't really used.

pferraro: You're right - this was an after-thought after the analysis doc was written.
The operations are also missing registration!  Fixing.

pferraro: Fixed.

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/remote/InvalidationNearCacheResourceDefinition.java
```diff
@@ -67,7 +71,17 @@ public AttributeDefinition getDefinition() {
     }
 
     static void buildTransformation(ModelVersion version, ResourceTransformationDescriptionBuilder parent) {
-        // No transformations yet
+        ResourceTransformationDescriptionBuilder builder = parent.addChildResource(PATH);
+        if (InfinispanModel.VERSION_11_0_0.requiresTransformation(version)) {
+            builder.getAttributeBuilder().setValueConverter(new AttributeConverter.DefaultAttributeConverter() {
```
jamezp: Should we be converting or rejecting the defined attribute?

jamezp (commented):
We should add some transformer tests to this for the new attribute.

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/remote/InvalidationNearCacheResourceDefinition.java
pferraro: The previous behavior was equivalent to a value of FALSE.  Thus, if the value is undefined or FALSE, then we discard.  Otherwise (i.e. it is defined and TRUE), then we reject.  Is this incorrect?

pferraro:
> We should add some transformer tests to this for the new attribute.

The trivial conversion is already covered, but I'll make sure this is covered by the rejection tests.

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/remote/InvalidationNearCacheResourceDefinition.java
pferraro: If it defined, we leave it defined.  We only need to convert the case where it is now undefined (because it was required previously).  Is that incorrect?

jamezp: Ah okay so it's no longer required? If that's the case this seems okay.

pferraro: Sorry.  My mistake. It was *not* required previously, but it had a weird default value, which is better expressed as being undefined.  So, the conversion is: if undefined, set it to the old default value.  If defined, leave as defined.  Does that make sense?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/remote/AbstractHotRodWebFailoverTestCase.java
```diff
@@ -47,4 +67,44 @@ public void beforeTestMethod() {
...
+        Assert.assertEquals(0L, result.get("writes").asLong());
+
+        result = execute(this.client2, readResource);
+        Assert.assertEquals(0L, result.get("hits").asLong());
+        Assert.assertEquals(0L, result.get("writes").asLong());
+
+        result = execute(this.client3, readResource);
+        Assert.assertEquals(0L, result.get("hits").asLong());
+        Assert.assertEquals(0L, result.get("writes").asLong());
+    }
```
tommaso-borgato: @pferraro test sequence is:

response = client.execute(new HttpGet(uri1));
response = client.execute(new HttpGet(uri2));
response = client.execute(new HttpGet(uri3));

lifecycle.stop(NODE_1);

response = client.execute(new HttpGet(uri2));
response = client.execute(new HttpGet(uri2));
response = client.execute(new HttpGet(uri3));
response = client.execute(new HttpGet(uri3));

lifecycle.start(NODE_1);

response = client.execute(new HttpGet(uri2));
response = client.execute(new HttpGet(uri2));
response = client.execute(new HttpGet(uri3));

lifecycle.stop(NODE_2);

response = client.execute(new HttpGet(uri1));
response = client.execute(new HttpGet(uri1));
response = client.execute(new HttpGet(uri3));
response = client.execute(new HttpGet(uri3));

lifecycle.start(NODE_2);

response = client.execute(new HttpGet(uri1));
response = client.execute(new HttpGet(uri1));

hence, NODE_2 never gets hit after restart: that's the reason why it's statistics remain at 0, correct?

tommaso-borgato (approved):
looks in sync with TP

kabir:
Retest this please

kabir:
@tommaso-borgato @pferraro Am I reading it right that if this passes it can be merged?

PR #12392, opened 2019-06-27
[WFLY-12228] Upgrade MicroProfile Health 2.0.1

jmesnil:
* Upgrade smallrye-health to io.smallrye:smallrye-health-2.0:1.0.2
  (artifactId has changed).
* Use a HealthReporter instance instead of SmallryeHealthReporter (that
  is based on CDI-only injection)
* Bump the management model (and XML version) to 2.0
  * add the empty-checks-status to the subsystem management model
* Update admin guide

JIRA: https://issues.jboss.org/browse/WFLY-12228

PR #12465, opened 2019-08-01
[WFLY-12342] Add server probes for readiness health check

jmesnil:
Add 3 healthcheck probes to determine the readiness of the server:

* server-state returns UP when the server is "running"
* boot-errors returns UP if there were no boot-errors
* deployments-status returns UP if all deployments status are OK

JIRA: https://issues.jboss.org/browse/WFLY-12342


jmesnil:
*Do not merge before #12392 is merged*

Only the 2nd commit of this PR should be reviewed.

jmesnil:
Failures are related to the PR (tests assume that there are no other health check probes than those coming from the deployments)

PR #12466, opened 2019-08-01
WFLY-12344 SecurityManager push/pull is expensive also for datasources

maeste:
Upstream jira: https://issues.jboss.org/browse/WFLY-12344


kabir:
@darranl @fjuma WDYT?

darranl:
This is JCA code so should be fine, the class PicketBoxSubjectFactory is in the package 'org.jboss.jca.core.security.picketbox' so I assume Stefano has double checked the creation works without being wrapped with the push / pop/

xstefank:
@kabir @maeste can this be merged, please?

PR #12489, opened 2019-08-09
[WFLY-1160] Provide ability to easily apply certain JBoss module libraries to all deployments running in a server

yersan:
WidlFly counterpart PR of https://github.com/wildfly/wildfly-core/pull/3890

This PR should be still on hold since it requires the core upgrade with the WildlFly-core counterpart merged to compile, however, it is open to get some early feedback about the implementation.

It prepares the resource definition and handling for global-directory.
Notice documentation and wildfly-capability PR is still missing. 

Jira issue: https://issues.jboss.org/browse/WFLY-1160
Analisys doc: https://github.com/wildfly/wildfly-proposals/pull/191


tterem:
Hi @yersan,

I was trying new feature manually and I have few topics regarding usability and user experience for discussion:

**1.** There can only be one global-directory configured, so what is the point of allowing users to set its name? I believe it can be confusing in a scenario when user tries to find out if global-directory is configured but he doesn't use the correct name.

```
[standalone@localhost:9990 /] /subsystem=ee/global-directory=my-global-directory:add(path=lib, relative-to=jboss.home.dir)
{"outcome" => "success"}

[standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:read-resource
{
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0216: Management resource '[
    (\"subsystem\" => \"ee\"),
    (\"global-directory\" => \"my-common-libs\")
]' not found",
    "rolled-back" => true
}

[standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:add(path=lib, relative-to=jboss.home.dir)
{
    "outcome" => "failed",
    "failure-description" => "WFLYEE0121: Global directory my-common-libs cannot be added, only one global directory can be defined.",
    "rolled-back" => true
}

[standalone@localhost:9990 /] 
```

I suggest to either not allow custom name for global-directory or change failure description: 
`Global directory my-common-libs cannot be added, only one global directory can be defined.` -> 
`Global directory my-common-libs cannot be added, because global directory my-global-directory is already defined.`


**2.** In a scenario when one of the classes in global-directory fails (e.g. throws an exception), it would be useful to notify the user which jar file contains this class.



**3.** When global directory contains a corrupted jar file, cli prints this after I try to deploy application:
```
{"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-2" => {"WFLYCTL0412: Required s
ervices that are not installed:" => ["module.resolved.service.\"deployment.deployment.war\".main"],"WFLYCTL0180: Services with missing/unavailabl
e dependencies" => ["jboss.module.service.\"deployment.deployment.war\".main is missing [module.resolved.service.\"deployment.deployment.war\".ma
in]"]}}}
```

However, when I successfully deploy an application (with correct jar in global directory), then switch correct jar for corrupted jar and restart server, cli doesn't print anything at all, but deployment is no longer available. I think behaviour should be more consistent here.

In both cases, I am able to see that jar couldn't be opened in server logs, but wouldn't it be better if cli also informed the user what went wrong?

I also tried a scenario when I had two jars in the global directory - correct jar and corrupted jar. Deployment only needed the correct jar, but it still didn't deploy and printed the message above. Maybe it would be better if it was deployed and just showed a warning that corrupted jar is present in global directory.

**4.** Which jar will be used if global directory has following structure?
a/a.jar
b/a.jar 
Analysis says that 

> The jar resources of this module will be created iterating over all jar files in alphabetical order and size descending. 

However, in my experience, the smaller jar was used. Also, I think the user should be notified if two jars with same name are present in global directory.


**5.** What if user wants to use global directory only for one specific deployment? I think it would be a good idea to enable users to choose if their deployment needs global directory or not (e.g. new parameter to deploy command).


**6.** As for implementing `Nice-to-Have Requirements`, it would enable to use for example Hibernate api or MP api in the jars which could be very useful.


**7.** What should happen if jar contains class that has same name and package as some java ee class? I tried using jax-rs Client in deployment and included my custom javax.ws.rs.client.Client class (not used by deployment) into jar in global directory. App deployed but I got an exception when I tried to access it:

`org.jboss.resteasy.spi.UnhandledException: java.lang.LinkageError: loader constraint violation: loader (instance of org/jboss/modules/ModuleClassLoader) previously initiated loading for a different type with name "javax/ws/rs/client/Client"`

We should decide which class is going to be used in this situation. I believe this scenario can only happen when user decides to override behaviour of some java ee class so his new class should be used. But user should still be notified about this.

What do you think about these?

yersan:
> 1. There can only be one global-directory configured, so what is the point of allowing users to set its name? 

It is a management resource which is not created by default, the users should supply a name to create it. Can we create it by default and do not allow the users to remove it? yes it could be, however at this point I'm unable to see real benefits; the use of a global directory is to resolve special cases, but I don't see it as something by default.

Having it as a management resource, it is a good option because if in the future, if we see a new need that requires more than one global directory, we will be prepared to add more than one as another management resource.

The idea was, if there is a need, we could allow more configurations per global-directory, e. g. Should this global-directory export the services? Should it export the classes? Should it be optional?

Independently, changing the message sounds pretty reasonable to me. I've just changed the code to show this error instead:

'Global directory my-common-libs cannot be added, because global directory my-global-directory is already defined.'

yersan:
> 2. In a scenario when one of the classes in global-directory fails (e.g. throws an exception), it would be useful to notify the user which jar file contains this class.

Would you mind expanding a bit more what do you mean here? I do not understand at all what are you referring with a class throwing an exception. An example case would be great.

If you are referring to show which jar contains the offending class responsible of an error as, e.g. `org.jboss.resteasy.spi.UnhandledException: java.lang.LinkageError: loader constraint violation: loader (instance of org/jboss/modules/ModuleClassLoader) previously initiated loading for a different type with name "javax/ws/rs/client/Client"`, then it is a bit of scope of this RFE. I think it doesn't have a trivial solution. 

It would be nice to be able to get the information like that one but working with loaded and available classes to find incorrect configurations automatically is out of the scope of this RFE.

yersan:
> 3. In both cases, I am able to see that jar couldn't be opened in server logs, but wouldn't it be better if cli also informed the user what went wrong?

I agree, however, at first glance, it seems nontrivial to accomplish it. I would investigate it a bit more. If you agree, I think it could be added as a nice to have, but it shouldn't prevent to get this RFE done for WildFly 18.

yersan:
> 4. Which jar will be used if global directory has the following structure?
a/a.jar
b/a.jar

Yes, you're right, the smaller one is used. I'm going to change it to work as it was decided in the initial analysis, using ascending file size. The idea is to have a deterministic way of loading the files. The limit is two jars with the same file name and size, then only one file is loaded, which one is loaded is undetermined, basically, we are considering both equals.

yersan:
> 5. As for implementing `Nice-to-Have Requirements`, it would enable to use for example Hibernate api or MP api in the jars which could be very useful.

I don't follow you on this, would you mind expanding a bit more what do you mean here?

bstansberry:
@yersan @tterem AIUI the idea is if the CLI user executes a 'reload' or a 'shutdown --restart=true' the CLI should capture and report any errors during boot as part of its response to those commands. If my understanding is correct, that is out of scope for this RFE.

yersan:
> > 4. Which jar will be used if global directory has the following structure?
> > a/a.jar
> > b/a.jar
> 
> Yes, you're right, the smaller one is used. I'm going to change it to work as it was decided in the initial analysis, using ascending file size. The idea is to have a deterministic way of loading the files. The limit is two jars with the same file name and size, then only one file is loaded, which one is loaded is undetermined, basically, we are considering both equals.

Thinking a bit more, I think we should forget about size and just use the full jar path to process the file in a deterministic order. So, in your previous example: global-directory/a/a.jar will be processed before than global-directory/b/a.jar

@tterem Do you agree?

tterem:
> Thinking a bit more, I think we should forget about size and just use the full jar path to process the file in a deterministic order. So, in your previous example: global-directory/a/a.jar will be processed before than global-directory/b/a.jar
> 
> @tterem Do you agree?

Yes.

tterem:
> > 2. In a scenario when one of the classes in global-directory fails (e.g. throws an exception), it would be useful to notify the user which jar file contains this class.
> 
> Would you mind expanding a bit more what do you mean here? I do not understand at all what are you referring with a class throwing an exception. An example case would be great.
> 
> If you are referring to show which jar contains the offending class responsible of an error as, e.g. `org.jboss.resteasy.spi.UnhandledException: java.lang.LinkageError: loader constraint violation: loader (instance of org/jboss/modules/ModuleClassLoader) previously initiated loading for a different type with name "javax/ws/rs/client/Client"`, then it is a bit of scope of this RFE. I think it doesn't have a trivial solution.
> 
> It would be nice to be able to get the information like that one but working with loaded and available classes to find incorrect configurations automatically is out of the scope of this RFE.

I was referring to exceptions and errors in general.
Example:  I create an library with method that contains only 'throw new  MyCustomException()', then this exception is thrown at runtime and  application crashes. How can user find out which library/jar is responsible for it?

This RFE brings library support, but it seems that current PR doesn't provide any logs/mechanisms for investigation and fixing this negative scenario. This RFE needs to keep good developer experience.

tterem:
> I don't follow you on this, would you mind expanding a bit more what do you mean here?

Thinking about user experience, this part of analysis is important for WF users, because users may want to use Hibernate API and MP API. Hibernate API is higly used by users and MP is new popular technology, WDYT?

> We could allow the possibility of adding existing module dependencies to the new module created from the global-directory configuration

yersan:
Question 5)
> > I don't follow you on this, would you mind expanding a bit more what do you mean here?
> 
> Thinking about user experience, this part of analysis is important for WF users, because users may want to use Hibernate API and MP API. Hibernate API is higly used by users and MP is new popular technology, WDYT?

Do you mean to add as a global-directory a new version of Hibernate API as a set of jar files?
e.g, suppose we ship in WildFly Hibernate 5.0 and users want to use Hibernate 6.0, so add all Hibernate 6.0 required jar files in a global-directory?
 
If so, take into account that the intention of global-directory is not to replace existing modules that we ship with the server. There could be other modules in the server related to one specific Hibernate API for example. The intention of a global directory is to supply common user classes to share across all deployments, those classes should not interfere with other existing modules in the server, if so, users should deal with it and disable existing modules and module dependents

yersan:
Question 2)

> > Would you mind expanding a bit more what do you mean here? I do not understand at all what are you referring with a class throwing an exception. An example case would be great.
> > If you are referring to show which jar contains the offending class responsible of an error as, e.g. `org.jboss.resteasy.spi.UnhandledException: java.lang.LinkageError: loader constraint violation: loader (instance of org/jboss/modules/ModuleClassLoader) previously initiated loading for a different type with name "javax/ws/rs/client/Client"`, then it is a bit of scope of this RFE. I think it doesn't have a trivial solution.
> > It would be nice to be able to get the information like that one but working with loaded and available classes to find incorrect configurations automatically is out of the scope of this RFE.
> 
> I was referring to exceptions and errors in general.
> Example: I create an library with method that contains only 'throw new MyCustomException()', then this exception is thrown at runtime and application crashes. How can user find out which library/jar is responsible for it?
> 
> This RFE brings library support, but it seems that current PR doesn't provide any logs/mechanisms for investigation and fixing this negative scenario. This RFE needs to keep good developer experience.

I see this case similar to having errors from classes that comes from existing modules. What do we do if we have invoked a class that throws an NFE and that class is in, for example, in netty-all-4.1.34.Final.jar? You will see a stack trace and you should find out looking at the stack trace which class was invoked, but you won't see that the error was an error from a class which is in  4.1.34.Final.jar

do you consider both similar cases?

tterem:
> Do you mean to add as a global-directory a new version of Hibernate API as a set of jar files?
> e.g, suppose we ship in WildFly Hibernate 5.0 and users want to use Hibernate 6.0, so add all Hibernate 6.0 required jar files in a global-directory?
> 
> If so, take into account that the intention of global-directory is not to replace existing modules that we ship with the server. There could be other modules in the server related to one specific Hibernate API for example. The intention of a global directory is to supply common user classes to share across all deployments, those classes should not interfere with other existing modules in the server, if so, users should deal with it and disable existing modules and module dependents

I meant, what if user wants to use Hibernate in his library? I meant something like this:
`/subsystem=ee/global-directory=my-common-libs:add(path=lib, relative-to=jboss.home.dir, dependencies=[{name=org.hibernate.something,export=true,import=false}])`

tterem:
> I see this case similar to having errors from classes that comes from existing modules. What do we do if we have invoked a class that throws an NFE and that class is in, for example, in netty-all-4.1.34.Final.jar? You will see a stack trace and you should find out looking at the stack trace which class was invoked, but you won't see that the error was an error from a class which is in 4.1.34.Final.jar
> 
> do you consider both similar cases?

It's similar but there is a difference that netty-all is supported and provided by wildfly so users can easily discover where is the issue. On the other hand, they can put many custom libraries into global directory, even more versions of the same library like a/my-custom-library.1.0.jar and b/my-custom-library.2.0.jar. In this case, it could be useful to know which library caused an error.

bstansberry:
There's a lot of good discussion here. I just want to highlight that WildFly already has a facility for adding a module as a dependency of all deployments. We also support adding non-global dependencies on arbitrary modules via the Dependencies property in MANIFEST.MF. This RFE is basically about making it easier for people to create a module so their apps can depend on it. It's not particularly about changing or improving behavior when users add a dependency on a module to their app. That might be a good thing to do, but if it's done for this use case then the other existing use cases need to be considered so we have consistency, and if we change existing behavior we run the risk of breaking things. My preference is that kind of thing be kept out of scope.

marekkopecky:
> We also support adding non-global dependencies on arbitrary modules via the Dependencies property in MANIFEST.MF. 

It is also possible by jboss-deployment-structure.xml file. But in both cases, users needs to list dependencies and users know exactly which dependencies are used. But this RFE is about adding all jars from some folder where more versions of the same artifact can be present (see the discussion of RFE).

marekkopecky:
@yersan Can you please provide answer to question 6 from [this comment](https://github.com/wildfly/wildfly/pull/12489#issuecomment-521720014) as well?

tterem:
@yersan I have found another problem to discuss. Consider following structure of the global directory:
a/my-custom-library-1.jar
	A.class
b/my-custom-library-2.jar
	A.class
	B.class

User has library my-custom-library-1 with class A and new version of this library my-custom-library-2 with changed class A and added new class B, which uses methods of new class A.

Current implementation loads class A from my-custom-library-1.jar and class B from my-custom-library-2.jar. This causes exception when class B tries to use methods from class A that weren't in my-custom-library-1 version of A.

This doesn't seem very user-friendly in a situation when user puts more version of same artifact to the global directory. I think this behaviour could be improved. WDYT?

yersan:
> @yersan I have found another problem to discuss. Consider following structure of the global directory:
> a/my-custom-library-1.jar
> A.class
> b/my-custom-library-2.jar
> A.class
> B.class
> 
> User has library my-custom-library-1 with class A and new version of this library my-custom-library-2 with changed class A and added new class B, which uses methods of new class A.
> 
> Current implementation loads class A from my-custom-library-1.jar and class B from my-custom-library-2.jar. This causes exception when class B tries to use methods from class A that weren't in my-custom-library-1 version of A.
> 
> This doesn't seem very user-friendly in a situation when user puts more version of same artifact to the global directory. I think this behaviour could be improved. WDYT?

Right @tterem The error is expected. Users should pay attention to what they are adding to the common lib directory. Enabling this feature, it is easiest to have classpath pollution. The case you are describing is exposing two similar classes to their deployments, they have to decide what the wants to use.

This case would be similar to, forgetting the possibility of a shared lib directory, distribute both libraries into your, e. g. WEB-INF/lib folder. In that case, I would bet they will be facing similar issues. We should not have different versions of the same class on two different jars available in the classpath

yersan:
Question 6) 

> @yersan Can you please provide answer to question 6 from [this comment](https://github.com/wildfly/wildfly/pull/12489#issuecomment-521720014) as well?

@marekkopecky You will see the same error if the jar containing the offending class is added as a global-module. Currently, global-modules and the module generated from a global-directory are added as the first system dependency, so they can interfere with the server implicit modules if they contain the same classes as any other server dependency. It is similar to the case described by @tterem here: https://github.com/wildfly/wildfly/pull/12489#issuecomment-523892530

The point here is to be able to show an error soon instead of hiding something to your deployment, that could have some benefits, or change it in the future in a different RFE. By now the key point is, global-directories, as well as global-modules, should not be used to try to override server versions.

yersan:
> Question 6)
> 
> > @yersan Can you please provide answer to question 6 from [this comment](https://github.com/wildfly/wildfly/pull/12489#issuecomment-521720014) as well?
> 
> @marekkopecky You will see the same error if the jar containing the offending class is added as a global-module. Currently, global-modules and the module generated from a global-directory are added as the first system dependency, so they can interfere with the server implicit modules if they contain the same classes as any other server dependency. It is similar to the case described by @tterem here: [#12489 (comment)](https://github.com/wildfly/wildfly/pull/12489#issuecomment-523892530)
> 
> The point here is to be able to show an error soon instead of hiding something to your deployment, it has some benefits. If there is a need, we could change it in the future in a different RFE. By now the key point is; global-directories, as well as global-modules, should not be used to try to override server versions.

I would like to comment about https://github.com/wildfly/wildfly/pull/12489#issuecomment-523892530 and point 6) (linkage error). Initially, I thought they were the same, but they are not at all.

Regarding https://github.com/wildfly/wildfly/pull/12489#issuecomment-523892530, we can mitigate it if, instead of having a single module for all the jar files found in the global-directory, we create one module by jar file. Then, the classloader of the module that represents each jar will take care of its classes, allowing b/my-custom-library-2.jar access to its own version of A.class and a/my-custom-library-1.jar accessing to its version. That would require a change in the analysis document because instead of having one single module per global-directory, there would be one module per jar file. That sounds pretty reasonable to me, so I think we should do it. I will require changes in wildfly-core too.

@tterem @marekkopecky Do you agree?

marekkopecky:
> Regarding #12489 (comment), we can mitigate it if, instead of having a single module for all the jar files found in the global-directory, we create one module by jar file.

I adds some-spec.jar and some-impl.jar. Two modules will be created. some-impl.jar needs to have dependency to `some-spec.jar` module. Basically, each module should see all other modules, am I right?

marekkopecky:
> That would require a change in the analysis document because instead of having one single module per global-directory, there would be one module per jar file.

This is a dev decision. Anyway, if you change the implementation in this PR, we check usability of prepared solution once the new implementation will be ready.

yersan:
> > That would require a change in the analysis document because instead of having one single module per global-directory, there would be one module per jar file.
> 
> This is a dev decision. Anyway, if you change the implementation in this PR, we check usability of prepared solution once the new implementation will be ready.

Then we will keep the original strategy of one module per global-directory. That will allow the users to move the jars included in their application libs to a commonplace and they would be treated similarly. If in the future someone needs a module per group of jars, we could enable the uses of multiple global-directory.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,71 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
 if the Module's `META-INF` path should be available to the deployment's
 class loader.
 
+[[global-directories]]
+== Global Directories
```
tterem: I think the title should be Global Directory, since now there can be only one.

yersan: This was changed to Global directory

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ee/globaldirectory/GlobalDirectoryTestCase.java
```diff
@@ -0,0 +1,279 @@
...
+     * 3. Check if global-directory is registered properly and verify its attributes
+     * 4. Restart server
+     * 5. Set logging level to DEBUG
+     * 6. Deploy test application deployment
+     * 7. Call some method from global-directory in deployment
+     * 8. Verify that class from jar which is first in alphabetical order was used
+     * 9. Check if server log file contains all jars in alphabetical order
+     */
+    @Test
+    public void testLoadOrder() throws Exception {
```
yersan: This test is expected to fail. It needs wildfly-core 10.0.0.Beta6

yersan:
retest this please

bstansberry:
retest this please

marekkopecky:
retest this please

marekkopecky:
@yersan Team city checks still fails, see for example this: https://ci.wildfly.org/viewLog.html?buildId=166121&buildTypeId=WF_PullRequest_DependencyTree&tab=buildLog

Can you please try to make a rebase of your branch? It may help ...

yersan:
@marekkopecky @tterem I rebased onto #12585 to make this pass. Additionally, I've fixed the GlobalDirectoryDomainTestCase which was failing on Windows. Now all test should pass, let's see what CI says.

yersan:
@tterem GlobalDirectoryTestCase#testLoadOrder is still unstable in team city. Could you take a look?

tterem:
@yersan I am working on it.

tterem:
Hi @yersan 

What is the correct way to use global directory with Security Manager? 

I was trying to fix the tests that are failing with Security Manager. I am getting AccessControlException when using "getProtectionDomain" and "getClassLoader". I tried it manually using simple app but the result is same.


`org.jboss.resteasy.spi.UnhandledException: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getProtectionDomain")" in code source "(jar:file:/home/tterem/wildfly/build/target/wildfly-18.0.0.Beta1-SNAPSHOT/lib/SingleLibrary-1.0-SNAPSHOT-secmgr4.jar!/ <no signer certificates>)" of "ModuleClassLoader for Module "deployment.external.global-directory.my-common-libs" from Service Module Loader")`

I tried configuring permissions with permissions.xml file and also with cli command ([based on this example](https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.2/html/how_to_configure_server_security/java_security_manager)) but it only applies for code on deployment, not code in global directory libraries. I tried to include permissions.xml in global directory directly as a file and also as part of jar file, but it didn't help.

yersan:
@tterem I've checked the issue with your tests, it looks like JBoss Modules does not make honor to the builder API, so I have filled https://github.com/jboss-modules/jboss-modules/pull/246 to address it there.

@dmlloyd Do you plan to release a new JBoss Module release for WF19? I can work around on this issue on WildFly-core, however, my understanding is we should fix it on JBoss Modules, would you mind giving us your feedback? thanks!

dmlloyd:
I can produce a JBoss Modules release at any time, just let me know when.

yersan:
Ok, thanks for the quick answer @dmlloyd , then as soon as jboss-modules/jboss-modules#246  is accepted we could have the fix on wildfly-core. We'll wait for https://issues.jboss.org/browse/MODULES-395 then

yersan:
Hi again @dmlloyd , since  jboss-modules/jboss-modules#246 was merged in 1.9 branch, could you release jboss modules 1.9.2.Final?

dmlloyd:
I'm working on the release but there's a test regression I'm bisecting now.

dmlloyd:
OK the release is done.

yersan:
JBoss Modules 1.9.2.Final has been added to wildfly-core 11.0.0.Beta3.
Now we need to wait for a wildfly core upgrade in wildfly to finish this PR.

bstansberry:
retest this please

yersan:
Hello @kabir the core-upgrade-needed label is no longer expected here. 
@tterem I've just updated the permissions in the tests and squashed the commits, the CI should pass now.

yersan:
Hello @tterem, the security issues were fixed with the jboss modules upgrade, however, `GlobalDirectoryTestCase.testLoadOrder` is still failing on team city. Could you take a look, thanks?

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
```diff
@@ -0,0 +1,175 @@
...
+        for (AttributeDefinition attribute : ATTRIBUTES) {
+            resourceRegistration.registerReadWriteAttribute(attribute, null, handler);
+        }
+    }
+
+    private static class GlobalDirectoryRemoveHandler extends AbstractRemoveStepHandler {
+
+        @Override
+        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+            context.restartRequired();
```
bstansberry: This seems odd. Other dynamically created modules, e.g. deployments, don't require a full process restart. I'd ask about .setRemoveRestartLevel(OperationEntry.Flag.RESTART_JVM) above as well but I'll just ask here.

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
```diff
@@ -0,0 +1,175 @@
...
+ * The resource definition for global-directory child resource in EE subsystem.
+ *
+ * @author Yeray Borges
+ */
+public class GlobalDirectoryResourceDefinition extends PersistentResourceDefinition {
+
+    static SimpleAttributeDefinition PATH = create(ModelDescriptionConstants.PATH, ModelType.STRING, false)
+            .setAllowExpression(true)
+            .addArbitraryDescriptor(FILESYSTEM_PATH, new ModelNode(true))
+            .setRestartJVM()
```
bstansberry: Why not setRestartAllServices?

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
```diff
@@ -0,0 +1,175 @@
...
+            .build();
+
+    static SimpleAttributeDefinition RELATIVE_TO = create(ModelDescriptionConstants.RELATIVE_TO, ModelType.STRING, true)
+            .setAllowExpression(false)
+            .setRestartJVM()
```
bstansberry: Why not setRestartAllServices?

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
 if the Module's `META-INF` path should be available to the deployment's
 class loader.
 
+[[global-directory]]
+== Global Directory
+Global modules can be used to shared common libraries across all deployed applications, but it could be unpractical if the name of a shared library changes very often or if there are many libraries you want to share. Both cases will require changes in the underlying module.xml that represents this global module.
```
bstansberry: s/shared/share

s/unpractical/impractical

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
...
+
+The EE subsystem allows the configuration of a global directory, which represents a directory tree scanned automatically to include .jar files and resources as a single additional dependency. This module dependency is added as a system dependency on each deployed application. Basically, with a global directory, you will be relying on WildFly to automatize the maintenance and configuration of a JBoss module that represents the jar files and resources of a specific directory.
```
bstansberry: s/automatize/automate

s/JBoss module/JBoss Module module/g

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
...
+----
+  [standalone@localhost:9990 /] /subsystem=ee/global-directory=my-common-libs:add(path=lib, relative-to=jboss.home.dir)
+----
+
+The following attributes are available on the global-directory resource:
+
+* path: The path of the directory to scan. (Mandatory)
+* relative-to: The name of another previously named path, or of one of the standard paths provided by the system. (Optional)
+
+When a `global-directory` is created, the server establishes a JBoss module with one https://jboss-modules.github.io/jboss-modules/manual/#path-resource-loader[Path Resource Loader] created using 'path' and 'relative-to' attributes and one https://jboss-modules.github.io/jboss-modules/manual/#jar-resource-loader[Jar Resource loader] for each jar file included in this directory and subdirectories.
```
bstansberry: s/JBoss module/JBoss Modules module/g

s/and subdirectories/and its subdirectories/g

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
...
+/my-common-libs/A/A/z-lib.jar
+/my-common-libs/A/a-lib.jar
+/my-common-libs/A/b-lib.jar
+/my-common-libs/a-lib.jar
+/my-common-libs/A/B/a-lib.jar
+/my-common-libs/properties-1.properties
+/my-common-libs/A/B/properties-2.properties
+....
+
+The JBoss module generated after scanning this global-directory will be equivalent to the following module.xml:
```
bstansberry: JBoss Modules module

or just 'module'

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
...
+
+    <dependencies>
+        <module name="javaee.api"/>
+    </dependencies>
+</module>
+....
+
+The name of the generated module follows the pattern _deployment.external.global-directory.{global-directory-name}_ and as such, it can be excluded selectively using your deployment-structure.xml.
+
+All resources will be available on the application class loader. For example, you could access to the above property files using the context ClassLoader of your current thread:
```
bstansberry: s/on the application/from the application/g

bstansberry: s/access to/access

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Application_Deployment_Configuration.adoc
```diff
@@ -39,6 +39,75 @@ The optional `meta-inf` attribute, which defaults to `true`, indicates
...
+
+All classes inside of each jar file will also be available, and the order of how the resource-root are created internally will govern the order of the class loading. The jar resources of the generated module will be created iterating over all jar files found in the directory tree. Each directory is scanned alphabetically starting from the root, and on each level, each subdirectory is also explored alphabetically until visiting all the branch. Files found on each level are also added in alphabetical order.
+
+Notice you should know which classes are exposed on each .jar file and avoid conflicts including the same class twice with incompatible binary change. In those cases, classloading errors are likely to occur. Specifically, you should not add classes that interfere with the classes the server already makes available for your application; the goal of a global directory is not to override and replace existing library versions shipped with the server. It is a facility that will allow removing common frameworks you usually add to your application libs to a commonplace to facilitate its maintenance.
```
bstansberry: s/removing/moving
s/commonplace/common place/g
s/its maintenance/maintenance

bstansberry (changes requested):

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
```diff
@@ -0,0 +1,175 @@
...
+    static final SimpleAttributeDefinition[] ATTRIBUTES = new SimpleAttributeDefinition[]{PATH, RELATIVE_TO};
+
+    private static final AbstractAddStepHandler ADD = new GlobalDirectoryAddHandler();
+    private static final AbstractRemoveStepHandler REMOVE = new GlobalDirectoryRemoveHandler();
+
+    public static final GlobalDirectoryResourceDefinition INSTANCE = new GlobalDirectoryResourceDefinition();
+
+    public GlobalDirectoryResourceDefinition() {
+        super(new SimpleResourceDefinition.Parameters(PathElement.pathElement(EESubsystemModel.GLOBAL_DIRECTORY), EeExtension.getResourceDescriptionResolver(EESubsystemModel.GLOBAL_DIRECTORY))
+                .setAddRestartLevel(OperationEntry.Flag.RESTART_JVM)
```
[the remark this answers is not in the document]
yersan: This was left as a mistake. At first glance, I thought the only reliable way to be able to unload the module created by the configuration of a global-directory was to execute a full server restart. 

After your comment, I've been checking how it behaves using a reload and since all the services are removed, the module created is also unloaded. @tterem I've changed the tests to use a reload instead of a full server restart, let's wait for CI.

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
yersan: Fixed, now we are using a ReloadRequiredRemoveStepHandler

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
yersan: Fixed, now using RestartAllServices flag

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
yersan: Fixed, now using RestartAllServices flag

ee/src/main/java/org/jboss/as/ee/subsystem/GlobalDirectoryResourceDefinition.java
```diff
@@ -0,0 +1,168 @@
...
+    static final SimpleAttributeDefinition[] ATTRIBUTES = new SimpleAttributeDefinition[]{PATH, RELATIVE_TO};
+
+    private static final AbstractAddStepHandler ADD = new GlobalDirectoryAddHandler();
+    private static final AbstractRemoveStepHandler REMOVE = new ReloadRequiredRemoveStepHandler();
+
+    public static final GlobalDirectoryResourceDefinition INSTANCE = new GlobalDirectoryResourceDefinition();
+
+    public GlobalDirectoryResourceDefinition() {
+        super(new SimpleResourceDefinition.Parameters(PathElement.pathElement(EESubsystemModel.GLOBAL_DIRECTORY), EeExtension.getResourceDescriptionResolver(EESubsystemModel.GLOBAL_DIRECTORY))
+                .setAddRestartLevel(OperationEntry.Flag.RESTART_ALL_SERVICES)
```
yersan: Ups, my fault, this basically should be removed. I'll fix it after CI result

yersan:
Just rebased to fix conflicts with the EE subsystem model upgrade merged on a different PR.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ee/globaldirectory/GlobalDirectoryBase.java
```diff
@@ -0,0 +1,351 @@
...
+public class GlobalDirectoryBase extends AbstractCliTestBase {
+
+    protected static final String SUBSYSTEM_EE = "ee";
+    protected static final Path GLOBAL_DIRECTORY_PATH = Paths.get(TestSuiteEnvironment.getTmpDir(), "global-directory");
+    protected static final File GLOBAL_DIRECTORY_FILE = GLOBAL_DIRECTORY_PATH.toFile();
+    protected static final String GLOBAL_DIRECTORY_NAME = "global-directory";
+    protected static final Path SECOND_GLOBAL_DIRECTORY_PATH = Paths.get(TestSuiteEnvironment.getTmpDir(),"global-directory-2");
+    protected static final String SECOND_GLOBAL_DIRECTORY_NAME = "global-directory-2";
+
+    protected static final String DUPLICATE_ERROR_GLOBAL_DIRECTORY_CODE = "WFLYEE0121";
```
tterem: After latest rebase [the error code changed to WFLYEE0123 ](https://github.com/wildfly/wildfly/pull/12489/files#diff-c303b6e6cd0ecc32578eb7b9b9869e94R1160)so this line needs to updated as well.

yersan: Fixed

yersan:
restest this please

ee/src/main/java/org/jboss/as/ee/structure/GlobalDirectoryDependencyProcessor.java
```diff
@@ -0,0 +1,102 @@
...
+        final ServiceTarget target = phaseContext.getServiceTarget();
+        final ServiceTarget externalServiceTarget = deploymentUnit.getAttachment(Attachments.EXTERNAL_SERVICE_TARGET);
+
+        final ServiceName allDirReadyServiceName = depUnitServiceName.append("directory-services-ready");
+        final ServiceBuilder<?> allDirReadyServiceBuilder = target.addService(allDirReadyServiceName);
+        final List<Supplier<GlobalDirectoryResourceDefinition.GlobalDirectory>> allDirReadySuppliers = new ArrayList<>();
+
+        List<ServiceName> serviceNames = serviceRegistry.getServiceNames();
+        for (ServiceName serviceName : serviceNames) {
+            if (capabilitySupport.getCapabilityServiceName(EE_GLOBAL_DIRECTORY_CAPABILITY_NAME).isParentOf(serviceName)) {
```
bstansberry: Better is to call capabilitySupport.getCapabilityServiceName(EE_GLOBAL_DIRECTORY_CAPABILITY_NAME) outside the loop and cache the result in a local var.

bstansberry (changes requested):
@yersan Looks good except for one perf change I'd like.

ee/src/main/java/org/jboss/as/ee/structure/GlobalDirectoryDependencyProcessor.java
yersan: Move outside of the loop, thanks

bstansberry (approved):

tterem (approved):

PR #12496, opened 2019-08-13
WFLY-12374: fix transaction attribute alternatives

michpetrov:
Issue: [WFLY-12374](https://issues.jboss.org/browse/WFLY-12374)

This blocks wildfly/wildfly-core#3863 ~~(update pending)~~, [WFCORE-4287](https://issues.jboss.org/browse/WFCORE-4287)

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/ObjectStoreTypeTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/ObjectStoreTypeTestCase.java
@@ -273,8 +273,9 @@ public class ObjectStoreTypeTestCase extends AbstractCliTestBase {
         assertTrue(modeNode != null);
         ModelNode alternatives = modeNode.get("alternatives");
         assertTrue(alternatives != null);
-        assertEquals(2, alternatives.asList().size());
-        for ( int nbAlternative = 0; nbAlternative < 2 ; nbAlternative++ )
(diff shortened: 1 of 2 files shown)


darranl:
@tomjenkinson Could you add some review to this one or nominate someone to review?  Looks like clean CI so should be able to go straight in.

darranl:
@bstansberry / @jmesnil Just catching up on some older PRs in the queue - are you able to provide an answer to the question from @tomjenkinson we may then be able to get this one moving.

bstansberry:
@michpetrov @tomjenkinson @darranl 

It's correct that for resource attributes (what you see with a :read-resource op) aliases are not alternatives.

A small tricky thing is we usually use the same definition objects for both the resource attributes and the parameters for the 'add' operation. And for the 'add' operation 'use-journal-store' and 'use-hornetq-store' are better treated as alternatives, as it's better not to include both in the values for an add op. #13388 is meant to deal with this distinction. The meaningful commit is https://github.com/wildfly/wildfly/pull/13388/commits/4d625473eda4a74da7dd5c16b5ffcbba6c0a3564

TBH I'm not sure that's worth it vs what's here; it adds some complexity for very little gain. The only practical effect of what it does is the /subsystem=transactions:read-operation-description(name=add) output continues to report that the two params are alternatives. And, as a result, the CLI if you are using tab completion to suggest possible parameters will not show one as an option once you've included the other. I doubt people are using CLI tab completion much as an aid in adding the transaction subsystem. Neither this PR or #13388 has any effect on how the server handles the add operation.

bstansberry:
Superseded by #13388

PR #12268, opened 2019-05-10
[WFLY-6143] Server interceptors

tadamski:
https://issues.jboss.org/browse/WFLY-6143

Requires core upgrade:
https://github.com/wildfly/wildfly-core/pull/3772

Please make sure your PR meets the following requirements:
- [ ] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue(s)
- [ ] Pull Request contains description of the issue(s)
- [ ] Pull Request does not include fixes for issues other than the main ticket
- [ ] Attached commits represent units of work and are properly formatted

For bigger changes, major and minor component upgrades make sure your PR also meets following requirements:
- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

For new features ensure as well:
- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

PR #12498, opened 2019-08-14
WFLY-6143 Add interceptor module substitution test case

zhantemirov:
Due to the original high-priority feature (https://github.com/wildfly/wildfly/pull/12268) it was decided to add this manual-mode test case a little bit later as a separate PR. 

Changes:
* Added test case that substitutes an interceptor module with a newer version and checks the changes. 
* AbstractServerInterceptorsSetupTask and all relevant classes have been moved to _shared_ module.

Jira issue: https://issues.jboss.org/browse/WFLY-6143
Analysis doc: wildfly/wildfly-proposals#180


bstansberry:
@zhantemirov There seems to be a problem with the test on Windows.

Also, I see this in the log and that tells me the test is doing something it should not:

Caused by: java.lang.IllegalArgumentException: Could not create C:\buildAgent\work\bb15431f347cd651\testsuite\integration\manualmode\..\..\..\build\target\wildfly-18.0.0.Beta1-SNAPSHOT\modules\interceptor-module\main

It is not acceptable for a test to write in build\target\wildfly-18.0.0.Beta1-SNAPSHOT\.  Modules that tests create should be installed somewhere in the testsuite module's target dir and that location should be on the server's module path. I'm pretty sure this kind of thing must be being done elsewhere so it should be a matter of finding such a case and following the same setup. Sorry I don't have time to find one.

bstansberry:
A hint:

The standard surefire config in testsuite/pom.xml sets the module.path system prop as follows:

<module.path>${jboss.dist}/modules${path.separator}${basedir}/target/modules</module.path>

So I believe you should be able to write modules in ${basedir}/target/modules i.e. testsuite/integration/manualmode/target/modules and the server will pick them up.

bstansberry:
@zhantemirov After looking into how this module creation stuff is being done, using TestModule from org.wildfly.core:wildfly-core-testsuite-shared, I believe correcting that is out of scope for this JIRA. For sure it would require a WFCORE JIRA to fix TestModule.

So, the task here is just to fix the Windows failures, whatever the cause might be.

zhantemirov:
@bstansberry, thank you for the hints. I was looking at it yesterday and now I have a feeling that providing fix for the Windows build will probably require TestModule fix, but I haven't tried another variants yet. Will provide feedback a little bit later.

bstansberry:
@zhantemirov https://github.com/wildfly/wildfly-core/pull/3903 would fix TestModule

zhantemirov:
@bstansberry I've made some research on this Windows issue and here are the results.

The IllegalArgumentException that is shown in the TeamCity logs

`Caused by: java.lang.IllegalArgumentException: Could not create C:\buildAgent\work\bb15431f347cd651\testsuite\integration\manualmode\..\..\..\build\target\wildfly-18.0.0.Final-SNAPSHOT\modules\interceptor-module\main
	at org.jboss.as.test.module.util.TestModule.create(TestModule.java:138)
	at org.jboss.as.test.manualmode.ejb.interceptor.serverside.InterceptorSubstituteTestCase.createUpdatedInterceptorModule(InterceptorSubstituteTestCase.java:175)
	`

is caused by an java.nio.file.FileSystemException while trying to remove (with _file.delete()_ method) existing module folder with files in it before creating a new module folder:

`C:\szhantem\wildfly\testsuite\integration\manualmode\target\modules\interceptor-module\main\server-side-interceptor-substitute.jar: The process cannot access the file because it is being used by another process.`

I tried to use _java.nio.file.Files.delete()_,  _FileUtils.deleteQuietly()_, and even _System.gc()_ - nothing helps due to the fact the file is locked by a process.

It would be great to get some advices/suggestions from you, thanks.

bstansberry:
@zhantemirov I suspect it is the server process itself that is preventing removal. Windows is more picky about that.

A workaround is to catch and ignore the problem, and to use a not-likely-to-be-duplicated name for the module so its presence doesn't disrupt subsequent tests.

I doubt reloading the server after removing any refs to a module from the config would help, but I could be wrong.

zhantemirov:
@bstansberry thanks for the tip, I slightly modified server reloading order - so now it works both on Unix and Windows machines.

zhantemirov:
@bstansberry could you please find some time to review the changes? Thanks in advance.

bstansberry (approved):

PR #12499, opened 2019-08-14
[WFLY-12243]: Monitor resource adapter thread pools.

ehsavoie:
* Adding metrics to `activemq.artemis.client.global.thread.pool` and  `activemq.artemis.client.global.scheduled.thread.pool`.

 Jira: https://issues.jboss.org/browse/WFLY-12243


bstansberry:
@ehsavoie Checkstyle problem

bstansberry:
@ehsavoie Security manager failures.

ehsavoie:
retest this please

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingSubsystemRootResourceDefinition.java
```diff
@@ -19,51 +19,145 @@
...
             .setRequired(false)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
 
-    public static final SimpleAttributeDefinition GLOBAL_CLIENT_SCHEDULED_THREAD_POOL_MAX_SIZE = create("global-client-scheduled-thread-pool-max-size", INT)
+    public static final SimpleAttributeDefinition GLOBAL_CLIENT_THREAD_POOL_ACTIVE_COUNT = create(GLOBAL_CLIENT_PREFIX + org.jboss.as.threads.CommonAttributes.ACTIVE_COUNT, INT)
+            .setAttributeGroup("global-client")
+            .setStorageRuntime()
+            .setUndefinedMetricValue(new ModelNode(0))
```
bstansberry: Please use ModelNode.ZERO here and for all the others.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingSubsystemRootResourceDefinition.java
```diff
@@ -19,51 +19,145 @@
...
             .setXmlName("thread-pool-max-size")
             .setRequired(false)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
 
-    public static final SimpleAttributeDefinition GLOBAL_CLIENT_SCHEDULED_THREAD_POOL_MAX_SIZE = create("global-client-scheduled-thread-pool-max-size", INT)
+    public static final SimpleAttributeDefinition GLOBAL_CLIENT_THREAD_POOL_ACTIVE_COUNT = create(GLOBAL_CLIENT_PREFIX + org.jboss.as.threads.CommonAttributes.ACTIVE_COUNT, INT)
+            .setAttributeGroup("global-client")
+            .setStorageRuntime()
```
bstansberry: This isn't necessary, as MRR.registerMetric results in it recording it as runtime regardless of what the AD says.

bstansberry: Please add AttributeAccess.FLAG.GAUGE_METRIC or COUNTER_METRIC for each of these.

bstansberry (changes requested):

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/MessagingSubsystemRootResourceDefinition.java
ehsavoie: I didn't set it to all metrics as for some I'm not sure if the semantics really match (like LARGEST_THREAD_COUNT and KEEPALIVE_TIME)

bstansberry: Per https://prometheus.io/docs/concepts/metric_types/ a LARGEST_THREAD_COUNT sounds like a counter as it counts and is monotonically increasing unless reset to zero.

A KEEPALIVE_TIME doesn't sound like a metric at all; it sounds like a runtime-attribute.

bstansberry (approved):

PR #12502, opened 2019-08-15
WFLY-12380 Fix org.wildfly.clustering.service.ServiceSupplier usage

pferraro:
https://issues.jboss.org/browse/WFLY-12380

Avoids locking of service registry whenever possible.


bstansberry (approved):
This LGTM.

I was just looking at a HotRod metrics PR that also used 'true' so that will need updating.

pferraro:
> I was just looking at a HotRod metrics PR that also used 'true' so that will need updating.

As will https://github.com/wildfly/wildfly/pull/12289

PR #12508, opened 2019-08-20
[WFLY-12394] Execute portions of the testsuite using a slimmed server instead of the normal dist

bstansberry:
https://issues.jboss.org/browse/WFLY-12394

This would be accompanied by new PR jobs that run with -Dts.layers rather than -DallTests which in turn would trigger slimmed installations and different surefire executions in the various testsuites.

@jfdenise @mjurc @yersan @jmesnil  FYI.


bstansberry:
A couple custom jobs meant to look like the new PR jobs I contemplate:

https://ci.wildfly.org/viewLog.html?buildId=162612&tab=buildResultsDiv&buildTypeId=WF_MasterLinux

https://ci.wildfly.org/viewLog.html?buildId=162600&tab=buildResultsDiv&buildTypeId=WF_MasterWindowsJdk11

I'm not sure if those are running correctly at the moment; will check again later.

rsvoboda:
@msve fyi

testsuite/integration/microprofile-tck/config/pom.xml
```diff
@@ -79,4 +79,103 @@
...
+                                            <inherit-packages>false</inherit-packages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <configurations>
+                                        <config>
+                                            <model>standalone</model>
+                                            <name>standalone.xml</name>
+                                            <layers>
+                                                <layer>core-server</layer>
+                                                <layer>observability</layer>
```
jfdenise: @bstansberry , you are expecting observability to bring web-server through its dependency on cdi or you don't need webserver?
When we have WFLY-12299 in, web-server would have to be added explicitly. Same apply to the other provisioned server.

bstansberry: Yes, I was counting on it being pulled transitively. I'll update that.

bstansberry: Done.

bstansberry:
I've taken this out of draft state and removed the hold label as this is good for review.

There's more I plan to do on this topic but this is a large and useful chunk.

jfdenise (approved):

bstansberry:
retest this please

PR #12514, opened 2019-08-22
WFLY-10057 EJB subsystem configure max threads and core threads independently

chengfang:
This PR adds a `core-threads` attribute to wildfly ejb3 subsystem thread-pool configuration, to support independent configuration of `core-threads` and `max-threads`, proper timeout of non-core threads, and reuse existing threads as much as possible without creating new threads.  The implementation of this feature uses an executor service backed by `org.jboss.threads.EnhancedQueueExecutor`.  This PR requires wildfly-core to be upgraded to incorporate wildfly-core PR:
https://github.com/wildfly/wildfly-core/pull/3880/

JIRA: https://issues.jboss.org/browse/WFLY-10057

Related analysis document proposal:
https://github.com/wildfly/wildfly-proposals/pull/227

Related Issues:  
https://issues.jboss.org/browse/EAP7-488
Fine-grained control on ejb3 thread-pool and reuse idle threads in pool before creating new

https://issues.jboss.org/browse/WFCORE-1446
EJB thread pool keepalive not honored and is not reusing inactive threads

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemXMLPersister.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemXMLPersister.java
@@ -22,6 +22,9 @@
 
 package org.jboss.as.ejb3.subsystem;
 
+import java.util.List;
+import javax.xml.stream.XMLStreamException;
+
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.persistence.SubsystemMarshallingContext;
 import org.jboss.as.remoting.Attribute;
@@ -30,10 +33,6 @@ import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
 import org.jboss.staxmapper.XMLElementWriter;
 import org.jboss.staxmapper.XMLExtendedStreamWriter;
-
-import javax.xml.stream.XMLStreamException;
-
-import java.util.List;
 
 import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.*;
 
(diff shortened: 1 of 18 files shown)


kabir (changes requested):
@chengfang We need some transformer tests for the added attribute

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemXMLPersister.java
```diff
@@ -22,6 +22,11 @@
 
 package org.jboss.as.ejb3.subsystem;
 
+import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.*;
```
jamezp: This is banned by checkstyle. We can't have wildcard imports.

jamezp (dismissed):

chengfang:
> @chengfang We need some transformer tests for the added attribute

@kabir I've added community documentation for this issue to docs/.../EJB3.adoc, and also added transformer tests for the new attribute.

bstansberry:
retest this please

mnovak1:
@chengfang thanks for spotting this, here is the fix to test - could you merge it:
https://github.com/chengfang/wildfly/pull/2

mnovak1:
retest this please

kabir (commented):
@chengfang I've only taken a super-quick look and I think the transformer test looks ok. However, I notice that the non-rejecting transformer test using subsystem-ejb3-transform.xml does not contain a `<thread-pool>` entry. Would it be possible to add that (without the new `<core-threads>` element)? That way we get coverage that the subsystem can work with an older slave.

chengfang:
> the non-rejecting transformer test using subsystem-ejb3-transform.xml does not contain a entry. Would it be possible to add that (without the new element)?

@kabir can you be more specific?  For `subsystem-ejb3-transform.xml`, I merely updated the schema version from `5.0` to `6.0` to use the current version.  Did you mean to add more xml elements and attributes (that are not new) to `subsystem-ejb3-transform.xml` so that the non-rejecting tests will cover more?

kabir:
@chengfang Yes, exactly - it is the 'missing' <thread-pool> that stands out since that is what has the new core-threads element/attribute

chengfang:
> it is the 'missing' `<thread-pool>` that stands out since that is what has the new core-threads element/attribute

@kabir `subsystem-ejb3-transform.xml` already contains a `thread-pool` element, around line 47. Any other entries are missing in `subsystem-ejb3-transform.xml`?

kabir:
@chengfang you are right, sorry for the noise. I will take a better look in the morning

kabir:
This should be good to go once the retest I just triggered on TC has finished

kabir (approved):

PR #12529, opened 2019-08-25
[WFLY-6143] Remove unusable entry from legacy feature pack subsystem …

bstansberry:
…template

Correction to the WFLY-6143 work.


bstansberry:
Closing as this is incorporated as a commit in #12530.

PR #12530, opened 2019-08-25
Feature vs galleon packs

bstansberry:
Cleans up various issues in the legacy feature packs subsystem templates that result in config file differences between legacy builds and current galleon-driven builds.

In cases where the problem is due to a WF 18 change not being applied to the subsystem template, I used the JIRA id of the change that missed this part. There's also a commit related to infinispan where some ordering has just drifted between the two config styles, perhaps in WF 17. No bug there; the change is just to make them consistent to make diffing cleaner.


pferraro (approved):
LGTM.

PR #12531, opened 2019-08-26
Https and datasources galleon layers

jfdenise:
Issues:
https://issues.jboss.org/browse/WFLY-12431
https://issues.jboss.org/browse/WFLY-12407

- Layers added
- Reworked default config and existing layers to benefit from new layers.
- Documented new layers in admin guide.


bstansberry (approved):

jfdenise:
I left h2-datasource to not depend on datasources layer. You need h2 layers to be combined with other layer or a default config to provision them. 
Full default standalone config is not defined with layers. So bringing just the needed features to the default config instead of the whole datasources is safe. Some features are additive (eg: elytron permissions), adding tx layer on top of the default config would duplicate the tx security permissions.

PR #12534, opened 2019-08-26
[WFLY-12384] enforce managed executors use thread factory without a c…

emmartins:
…ontext service

Issue: https://issues.jboss.org/browse/WFLY-12384
Description:  Manage executors, when configured with a thread factory reference, should only use it as the source of config for their own thread factory (atm that's only threads priority), not use it directly. This is due to the executor thread factory must have no context service, otherwise it behaves as if it was apps creating threads directly, leaving deployment resources linked in its runnable during whole thread life.


bstansberry:
@emmartins It's fine to not depend on an invalid thread-factory (i.e. one meant for another purpose), and, for compatibility purposes, to do what you did and use the ref just to find the useful config data (the priority.)

But if the executor resources are not really using the thread factories, those attributes should be deprecated and the executor resources should get new attributes of their own (alternative to thread-factory) and use those to configure themselves.

bstansberry (approved):

PR #12543, opened 2019-08-28
[WFLY-12298] Implement RESTEasy context parameters in Wildfly management model

ronsigal:
https://issues.jboss.org/browse/WFLY-12298 "Allow to change RESTEasy settings via CLI"

Attributes representing RESTEasy context parameters are added to the jaxrs subsystem. They can be set in the wildfly management model and passed as context parameters to RESTEasy.

- [ yes] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [ yes] Pull Request contains link to the JIRA issue(s)
- [ yes] Pull Request contains description of the issue(s)
- [ yes] Pull Request does not include fixes for issues other than the main ticket
- [ yes] Attached commits represent units of work and are properly formatted

- [ yes] Pull Request requires a change to the documentation
- [ no] Documentation have been updated accordingly: TBD
- [ yes] Tests were added to cover changes

- [ yes] Analysis was done
- [ yes] Test Plan has been done
- [ yes] Tests were verified in advance


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsContextParamHandler.java
```diff
@@ -0,0 +1,294 @@
...
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.registry.Resource;
+import org.jboss.as.jaxrs.logging.JaxrsLogger;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+public class JaxrsContextParamHandler {
+
+   // Attributes in alphabetical order
+   public static final String JAX_RS_2_0_REQUEST_MATCHING = "jaxrs.2.0.request.matching";
```
OndrejKotek: There are 21 attributes defined. However, in the analysis it is proposed that all attributes defined in the RESTeasy
[User Guide](https://docs.jboss.org/resteasy/docs/3.8.1.Final/userguide/html/Installation_Configuration.html#configuration_switches) + _resteasy.providers.remove_ will be available (26 attributes). Can you update the analysis with reasons for omitting some of the parameters, or fix this PR, please?

OndrejKotek: I also do not see the _resteasy.providers.remove_ attribute.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
```diff
@@ -0,0 +1,375 @@
...
+
+/**
+ * Tests for setting RESTEasy context parameters using the Wildfly management model.
+ *
+ * @author <a href="rsigal@redhat.com">Ron Sigal</a>
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup(ResteasyAttributeTestCase.AttributeTestCaseDeploymentSetup.class)
+public class ResteasyAttributeTestCase extends ContainerResourceMgmtTestBase {//extends ContainerResourceMgmtTestBase{ //extends ContainerResourceMgmtTestBase {
```
OndrejKotek: Comments can be removed.

OndrejKotek:
@ronsigal Can you also add a test case verifying that existing deployments are not affected by a change of the attributes please?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
OndrejKotek: ```//extends ContainerResourceMgmtTestBase{ //extends ContainerResourceMgmtTestBase {```

bstansberry:
The WildFly management API standard is that multi-word attribute names are deliminated with a '-' not a '.'.  I think we'd need to do quite a bit of testing around attributes with '.' as that char as used in other ways, e.g. with complex attributes (ones with fields inside) the . allows users to specify specific fields rather than the entire attribute.

ronsigal:
I've added a JAXRS section to the Admin Guide. It's subject to some issues that have come up here concerning syntax.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsContextParamHandler.java
```diff
@@ -0,0 +1,294 @@
...
+
+   private static AttributeDefinition getListAttributeDefinition(String name, ModelType modelType) {
+      return new StringListAttributeDefinition.Builder(name)
+            .setRequired(false)
+            .setAllowExpression(false)
+            .setValidator(JavaIdListValidator.INSTANCE)
+            .build();
+   }
+
+   private static AttributeDefinition getMapAttributeDefinition(String name, ModelType modelType) {
```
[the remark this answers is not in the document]
kabir: @ronsigal your example is perfect, and what we use in other places where we have class names. So the key uses '-', for the values we can use '.' for class names. The logging subsystem uses this, and so does jgroups

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsContextParamHandler.java
```diff
@@ -0,0 +1,271 @@
...
+         return false;
+      }
+
+      @Override
+      protected void revertUpdateToRuntime(OperationContext context, ModelNode operation, String attributeName, ModelNode valueToRestore, ModelNode valueToRevert, Void handback) throws OperationFailedException{
+         JaxrsContextParamHandler.setContextParam(attributeName, valueToRestore);
+      }
+   }
+
+   static class ContextParamReadHandler extends ReadAttributeHandler {
```
kabir: Delete this class, it seems to be the root of my observation regarding the defaults. In the call to register the attribute, pass in a null handler and it will add the proper read handler which deals with the defaults for you.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
```diff
@@ -89,39 +72,18 @@ public void initialize(final ExtensionContext context) {
         final SubsystemRegistration subsystem = context.registerSubsystem(SUBSYSTEM_NAME, CURRENT_MODEL_VERSION);
         final ManagementResourceRegistration registration = subsystem.registerSubsystemModel(JaxrsSubsystemDefinition.INSTANCE);
         registration.registerOperationHandler(GenericSubsystemDescribeHandler.DEFINITION, GenericSubsystemDescribeHandler.INSTANCE);
+        JaxrsContextParamHandler.registerAttributes(registration);
```
kabir: Since the parser registers this under `<context-parameters>`, I think it makes sense to put all these new attributes under a separate child resource rather than directly in the subsystem root. That way if we ever end up adding a new set of attributes in this subsystem relating to something else, we have a clear separation between the different things. Also, it will make the transformers (which I have not checked yet simpler).

I suggest using PathElement("configuration", "context-parameters") as the element name (@bstansberry I think the plural form is ok here since it contains attributes rather than child resources?)

jaxrs/src/main/resources/schema/jboss-as-jaxrs_2_0.xsd
```diff
@@ -0,0 +1,49 @@
...
+       </xs:all>
+    </xs:complexType>
+    <xs:complexType name="context-parametersType">
+       <xs:choice minOccurs="0" maxOccurs="unbounded">
+          <xs:element name="context-parameter" type="context-parameterType"/>
+       </xs:choice>
+    </xs:complexType>
+    <xs:complexType name="context-parameterType" mixed="true">
+       <xs:attribute name="name" type="xs:string" use="required"/>
+       <xs:attribute name="value" type="xs:string" use="required"/>
```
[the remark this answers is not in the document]
kabir: @ronsigal I agree it needs to be a known set. If more things are exposed later, we need a new model version and schema versions

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_2_0.java
```diff
@@ -0,0 +1,132 @@
...
+
+      String name = reader.getAttributeValue(0);
+      String value = reader.getAttributeValue(1);
+      if (PATTERN_PROPERTY_LIST.matcher(value).matches() || PATTERN_LIST.matcher(value).matches()) {
+         String[] ss = value.split(",");
+         ModelNode modelNode = new ModelNode();
+         for (String s : ss) {
+            modelNode.add(s.trim());
+         }
+         JaxrsContextParamHandler.getContextParameters().put(name, modelNode);
```
kabir: This should not deal with JaxrsContextParameterHandler directly.

Instead the add handler (see my other point about putting these into a child resource of the subsystem) should be enhanced to deal with all these new parameters, so they get added to the model (they don't at the moment) and populate the context parameters cache in the same way as your write attribute handler.

I was getting quite confused why your subsystem test was working while not being able to see values persisted in the xml show up in the model of a restarted server :)

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsContextParamHandler.java
```diff
@@ -0,0 +1,271 @@
...
+
+   private static final String JAVA_ID = "[a-zA-Z_$][a-zA-Z_$0-9]*(\\.[a-zA-Z_$][a-zA-Z_$0-9]*)*";
+   private static final String JNDI_ID = JAVA_ID + ":" + JAVA_ID + "(/" + JAVA_ID + ")*";
+   private static final Pattern PATTERN_LIST_ELEMENT = Pattern.compile(JAVA_ID + "|" + JNDI_ID);
+   private static final Pattern PATTERN_PROPERTY_LIST_ELEMENT = Pattern.compile(".+:.+(,.+:.+)*");
+   private static final String NO_DEFAULT = "noDefault";
+
+   private static final Map<String, ModelNode> contextParams = new TreeMap<String, ModelNode>();
+
+   private static final Map<String, Boolean> contextParametersBoolean = new HashMap<String, Boolean>();
```
kabir: I noticed when trying to edit your code that you are using 3 spaces for indentation. The WildFly code style is 4 spaces. I've not checked everywhere, but I presume it will apply to all the code you have done due to your IDE settings?

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_2_0.java
```diff
@@ -0,0 +1,132 @@
...
+      ParseUtils.requireNoContent(reader);
+  }
+
+   /**
+    * {@inheritDoc}
+    */
+   @Override
+   public void writeContent(final XMLExtendedStreamWriter streamWriter, final SubsystemMarshallingContext context) throws XMLStreamException {
+       context.startSubsystemElement(JaxrsExtension.NAMESPACE_2_0, false);
+       Map<String, ModelNode> contextParameters = JaxrsContextParamHandler.getContextParameters();
```
kabir: I think we should just use what is in the model when writing this out. I mentioned using a service to store this state rather than a static field, and that service can only be populated in a normal running server.

If someone starts the service with --admin-only, there is no runtime stage for this service to get installed. And we need to be able to do the marshalling when running in admin-only

kabir (changes requested):
@ronsigal Quite a lot going on here :)

Also, you need to add transformers for all the new attributes, and add tests for those.

https://docs.wildfly.org/17/Extending_WildFly.html#Domain_Mode_Subsystem_Transformers explains what transformers are if you are not familiar

ronsigal:
@kabir, @bstansberry, @OndrejKotek, thank you all for putting in the time to make sense of all of this. I will try to do better next time. I'm thinking of applying for a promotion to Junior Developer. ;-)

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsContextParamHandler.java
```diff
@@ -0,0 +1,294 @@
...
+      }
+   }
+
+   ///////////////////////////////////////////////////////////////////////////////
+   ///                          Private functions                              ///
+   ///////////////////////////////////////////////////////////////////////////////
+   private static AttributeDefinition getAttributeDefinition(String name, ModelType modelType, ModelNode defaultValue) {
+      if (NO_DEFAULT.equals(defaultValue.asString())) {
+         return new SimpleAttributeDefinitionBuilder(name, modelType)
+               .setRequired(false)
```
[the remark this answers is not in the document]
hpehl: Thanks @bstansberry for the hint. I've added the `affects-model` to [WFLY-12298](https://issues.jboss.org/browse/WFLY-12298) and filed [HAL-1633](https://issues.jboss.org/browse/HAL-1633) to track the implementation in HAL.

bstansberry:
@ronsigal @kabir Should this be closed as out of date? #12835 sounds like the same thing.

ronsigal:
Right. I kept it around to remind me of all the things I had to fix. My Hall of Shame. ;-)

I'll close it now.

PR #12571, opened 2019-09-02
[WFLY-12304]: Upgrade Apache Artemis from 2.8.1 to 2.10.1

ehsavoie:
* Updating artemis version
* Adding new dependency on jctools-core
* pageSizeBytes is now an int instead of a long.

Jira: https://issues.jboss.org/browse/WFLY-12304

### DIFF
--- a/feature-pack/pom.xml
+++ b/feature-pack/pom.xml
@@ -1473,6 +1473,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jctools</groupId>
+            <artifactId>jctools-core</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jgroups</groupId>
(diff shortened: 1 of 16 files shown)


scottmarlow:
Just started an internal Jakarta EE 8 platform TCK run (JMS tests only).

bstansberry:
@ehsavoie Please create a separate JIRA/PR to separate the management API / xsd change from the upgrade. Treat the API/xsd change as a bug where we accepted LONG even though in practice only INT values were useful.  The Artemis upgrade forces us to fix the bug but the bug is there regardless.

ehsavoie:
https://github.com/wildfly/wildfly/pull/12610 is the PR for the LONG -> INT change

bstansberry (approved):

PR #12572, opened 2019-09-02
[WFLY-12427] Add missing dependencies for smallrye-health

jmesnil:
smallrye-health and its subsystem were missing dependencies to
org.jboss.logging.
Once logging was fixed, it showed that smallrye-health was missing a
dependency to smallrye-config to ensure that the SmallRyeHealthReporter
(that is not used by WildFly) was properly injected its
ConfigProperties.

JIRA: https://issues.jboss.org/browse/WFLY-12427

### DIFF
--- a/feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/health/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/health/main/module.xml
@@ -33,9 +33,12 @@
 
     <dependencies>
         <module name="javax.api"/>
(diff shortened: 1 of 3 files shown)


kabir:
@jmesnil still fails in EarOpenTracingWithWeldProbeTestCase

jmesnil:
Failure is related to the use of Weld development mode. When I switched it off, the test passed.
I'll check if there is a good reason to run this test in this weld mode.

jmesnil:
retest this please

bstansberry (approved):

PR #12562, opened 2019-08-30
WFLY-12460 Upgrade jboss-transaction-api_1.2_spec to jboss-transactio…

tomjenkinson:
…n-api_1.3_spec 2.0.0CR1

Thanks for submitting your Pull Request!

Please make sure your PR meets the following requirements:
- [x] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue(s)
https://issues.jboss.org/browse/WFLY-12460
- [x] Pull Request contains description of the issue(s)
Rev the artifact ID of the transactions API
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Attached commits represent units of work and are properly formatted

For bigger changes, major and minor component upgrades make sure your PR also meets following requirements:
- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

For new features ensure as well:
- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

PR #12584, opened 2019-09-04
Revert "WFLY-12460 Upgrade jboss-transaction-api_1.2_spec to jboss-transactio…"

kabir:
Reverts wildfly/wildfly#12562


bstansberry:
Superceded by #12585

PR #12587, opened 2019-09-04
[WFLY-12485] Galleon layers for embedded infinispan-based web cluster…

bstansberry:
…ing and JPA with a distributed 2nd level cache

https://issues.jboss.org/browse/WFLY-12485

### DIFF
--- a/galleon-pack/src/main/resources/feature_groups/distributable-web.xml
+++ b/galleon-pack/src/main/resources/feature_groups/distributable-web.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="distributable-web" xmlns="urn:jboss:galleon:feature-group:1.0">
+
+    <feature spec="subsystem.distributable-web">
+        <param name="default-session-management" value="default"/>
+        <param name="default-single-sign-on-management" value="default"/>
+        <feature spec="subsystem.distributable-web.infinispan-session-management">
+            <param name="infinispan-session-management" value="default"/>
+            <param name="cache-container" value="web"/>
+            <param name="granularity" value="SESSION"/>
+            <feature spec="subsystem.distributable-web.infinispan-session-management.affinity.primary-owner"/>
+        </feature>
+        <feature spec="subsystem.distributable-web.infinispan-single-sign-on-management">
+            <param name="infinispan-single-sign-on-management" value="default"/>
+            <param name="cache-container" value="web"/>
+            <param name="cache" value="sso"/>
+        </feature>
+        <feature spec="subsystem.distributable-web.routing.infinispan">
+            <param name="cache-container" value="web"/>
+            <param name="cache" value="routing"/>
+        </feature>
+    </feature>
+
+</feature-group-spec>
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -1059,6 +1059,58 @@
                             </configurations>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>jpa-distributed-provisioning</id>
+                        <goals>
+                            <goal>provision</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <install-dir>${layers.install.dir}/jpa-distributed</install-dir>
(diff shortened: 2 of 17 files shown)


bstansberry:
@jfdenise Please review.

bstansberry:
And @pferraro :)

As I mentioned in a chat message, a 'jgroups' layer may make sense.

bstansberry:
I just realized I didn't update the doc page so I killed the CI jobs and will resubmit.

testsuite/layers/pom.xml
```diff
@@ -1058,6 +1058,58 @@
...
                         </configuration>
+                    </execution>
+                    <execution>
+                        <id>jpa-distributed-provisioning</id>
+                        <goals>
+                            <goal>provision</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <install-dir>${layers.install.dir}/jpa</install-dir>
```
jfdenise: Should be jpa-distributable

testsuite/layers/pom.xml
```diff
@@ -1319,6 +1371,58 @@
...
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>web-clustering-provisioning</id>
+                        <goals>
+                            <goal>provision</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <install-dir>${layers.install.dir}/jpa</install-dir>
```
jfdenise: should be web-clustering

jfdenise: You should add the 2 new layers to all-layers-provisioning-full execution (end of this file).

bstansberry:
I added the doc page content.

@pferraro @jfdenise One thing to think about is what we would call a layer that used hotrod based web session caching. The impl details of this PR are quite simple; I think the main thing to think about is stuff like names to ensure we don't paint ourselves into a corner.

galleon-pack/src/main/resources/feature_groups/distributable-web.xml
```diff
@@ -0,0 +1,24 @@
...
+<feature-group-spec name="distributable-web" xmlns="urn:jboss:galleon:feature-group:1.0">
+
+    <feature spec="subsystem.distributable-web">
+        <param name="default-session-management" value="default"/>
+        <param name="default-single-sign-on-management" value="default"/>
+        <feature spec="subsystem.distributable-web.infinispan-session-management">
+            <param name="infinispan-session-management" value="default"/>
+            <param name="cache-container" value="web"/>
+            <param name="granularity" value="SESSION"/>
+            <feature spec="subsystem.distributable-web.infinispan-session-management.affinity.primary-owner"/>
```
pferraro: We probably want a separate non-clustered version of this feature-group-spec configured to use spec="subsystem.distributable-web.infinispan-session-management.affinity.local".
While "primary-owner" will fallback to equivalent behavior, it is likely going to be confusing.

galleon-pack/src/main/resources/feature_groups/distributable-web.xml
```diff
@@ -0,0 +1,24 @@
...
+        </feature>
+        <feature spec="subsystem.distributable-web.infinispan-single-sign-on-management">
+            <param name="infinispan-single-sign-on-management" value="default"/>
+            <param name="cache-container" value="web"/>
+            <param name="cache" value="sso"/>
+        </feature>
+        <feature spec="subsystem.distributable-web.routing.infinispan">
```
pferraro: A non-clustered version of this feature-spec-group would want to use:
<feature spec="subsystem.distributable-web.routing.local"/>

galleon-pack/src/main/resources/feature_groups/distributable-web.xml
bstansberry: @pferraro What would that feature-group be called?

pferraro: distributable-web-local?

bstansberry:
I'm going to close this for now as this is WF 19 stuff and #12599 has a test-only variant of this for use in WF !8 tests. So keep it simple and deal with 18 work first.

PR #12588, opened 2019-09-04
WFLY-12487 Include the original cause of failure, if jsf-injection module load fails

jaikiran:
The commit here introduces a relatively trivial enhancement to the exception that gets thrown while processing a JSF deployment. The details of this enhancement are noted in https://issues.jboss.org/browse/WFLY-12487.

With this change, the `DeploymentUnitProcessingException` will now include the original cause of the jsf-injection module load failure, if and when that happens.


bstansberry:
@fjuma OK? LGTM.

fjuma (approved):

bstansberry:
Thanks @jaikiran!

PR #12589, opened 2019-09-04
WFLY-12480 Switch from javax.persistence:javax.persistence-api to jakarta.persistence:jakarta.persistence-api

scottmarlow:
https://issues.jboss.org/browse/WFLY-12480

### DIFF
--- a/feature-pack/src/license/full-feature-pack-licenses.xml
+++ b/feature-pack/src/license/full-feature-pack-licenses.xml
@@ -664,6 +664,22 @@
       </licenses>
     </dependency>
     <dependency>
+      <groupId>jakarta.persistence</groupId>
+      <artifactId>jakarta.persistence-api</artifactId>
+      <licenses>
+        <license>
+          <name>Eclipse Public License 2.0</name>
+          <url>http://www.eclipse.org/legal/epl-v10.html</url>
+          <distribution>repo</distribution>
+        </license>
+        <license>
+          <name>Eclipse Distribution License, Version 1.0</name>
+          <url>http://repository.jboss.org/licenses/edl-1.0.txt</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
       <groupId>jakarta.security.enterprise</groupId>
       <artifactId>jakarta.security.enterprise-api</artifactId>
       <licenses>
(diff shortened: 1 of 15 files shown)


scottmarlow:
JDK11 "Execution timeout" during build, doesn't seem related to this change.

scottmarlow:
retest this please

feature-pack/src/license/full-feature-pack-licenses.xml
```diff
@@ -700,7 +700,7 @@
       <artifactId>javax.persistence-api</artifactId>
```
bstansberry: This groupId and artifactId needs to be updated too. Also, when you do that please put this dependency in the new correct alphabetical position, as these elements are kept in alphabetical order by maven GA.

pom.xml
```diff
@@ -286,7 +286,7 @@
         <version.javax.json.bind.api>1.0</version.javax.json.bind.api>
         <version.javax.jws.jsr181-api>1.0-MR1</version.javax.jws.jsr181-api>
         <version.javax.mail>1.6.2</version.javax.mail>
-        <version.javax.persistence>2.2</version.javax.persistence>
+        <version.jakarta.persistence>2.2.3</version.jakarta.persistence>
```
bstansberry: Please put this in the new alphabetical position.

bstansberry (changes requested):
Thanks, Scott. Needs a couple tweaks but overall looks good.

feature-pack/src/license/full-feature-pack-licenses.xml
scottmarlow: thanks, changed full-feature-pack-licenses.xml and moved to just before jakarta.security.enterprise.

pom.xml
scottmarlow: Thanks, moved to just before jakarta.security.enterprise

bstansberry (dismissed):

scottmarlow:
just squashed the update (root pom.xml + feature-pack/src/license/full-feature-pack-licenses.xml).

bstansberry (changes requested):
Can you add

<exclude>javax.persistence:persistence-api</exclude>

to the maven-enforcer-plugin ban-bad-dependencies execution in the root pom?

That will result in a failure if any of our deps are pulling that in without us having added an exclude in dependencyManagement. With that I see that org.hibernate:hibernate-core: is pulling in the old artifact. (Not a surprise!) There may be others; my local build failed when it hit that one.

scottmarlow:
@bstansberry under bannedDependencies, I already see <exclude>javax.persistence:persistence-api</exclude> was added in 2011, does that mean we aren't really banning that dependency?  I wonder if something else is interfering with javax.persistence:persistence-api that prevented you from seeing a failure for org.hibernate:hibernate-core.

bstansberry:
@scottmarlow I'm sorry; I typed the wrong thing and wasted your time!

I meant add

javax.persistence:javax.persistence-api

Note the 'javax' in the artifactId part.

So the existing 'javax.persistence:persistence-api' in there I guess is banning a different earlier variant of this API. You persistence guys need to get a bit more persistent about your names!

bstansberry (approved):
@scottmarlow Never mind re: ban-bad-dependencies. Once all the things adding new maven GAs are in we can do a bulk commit to do that. There are some already in and then various PRs otherwise ready and it's harder to do per PR than in one easy change.

scottmarlow:
+1

PR #12586, opened 2019-09-04
[WFLY-12433] Move to jboss-jaxrs-api_2.1_spec to 2.0.0.CR2

asoldano:
https://issues.jboss.org/browse/WFLY-12433

PR #12556, opened 2019-08-29
[WFLY-12087]:Upgrade CXF to 3.3.3

jimma:
https://issues.jboss.org/browse/WFLY-12087

This upgrade will fix security vulnerability issue in org.apache.santuario:xmlsec dependency.

PR #12490, opened 2019-08-11
[WFLY-6144] EJB static client interceptors

tadamski:
https://issues.jboss.org/browse/WFLY-6144
https://issues.jboss.org/browse/EAP7-475

Please make sure your PR meets the following requirements:
- [ ] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue(s)
- [ ] Pull Request contains description of the issue(s)
- [ ] Pull Request does not include fixes for issues other than the main ticket
- [ ] Attached commits represent units of work and are properly formatted

For bigger changes, major and minor component upgrades make sure your PR also meets following requirements:
- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

For new features ensure as well:
- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

PR #12590, opened 2019-09-04
Test Merge PR's

jamezp:
#12490 
#12556 
#12579 
#12582 
#12586


jamezp:
The failing test on the Windows JDK 8 job seems to be a common failure. I've tested the same test on a Windows VM and it passes. It also passed on all other jobs.

PR #12591, opened 2019-09-05
[WFLY-12495] Separate out feature groups related to clustering

bstansberry:
https://issues.jboss.org/browse/WFLY-12495

This will make it easier for our OpenShift images to use these config chunks in OpenShift-specific layers.


pferraro (approved):
LGTM.

PR #12592, opened 2019-09-05
[WFLY-12496] Use correct address for node1 in integration tests

istraka:
https://issues.jboss.org/browse/WFLY-12496

### DIFF
--- a/testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/transaction/nooutbound/TransactionContextRemoteCallTestCase.java
+++ b/testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/transaction/nooutbound/TransactionContextRemoteCallTestCase.java
@@ -25,7 +25,6 @@ package org.jboss.as.test.multinode.transaction.nooutbound;
 import static org.jboss.as.test.shared.integration.ejb.security.PermissionUtils.createPermissionsXmlAsset;
 
 import java.net.SocketPermission;
-import java.net.URL;
 
 import javax.ejb.EJBException;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


bstansberry:
retest this please

This requires approval from @ehsavoie as he's dealing with blocker issues involving these tests and we don't want to conflict.

bstansberry:
I put the 18.x label on this because if it passes CI and doesn't interfere with other work there's no reason not to merge it for 18.

bstansberry:
@istraka The 'Linux with security manager' CI job failures are due to this change. Deployment needs permissions.

ehsavoie (approved):

testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/transaction/nooutbound/TransactionContextRemoteCallTestCase.java
```diff
@@ -85,15 +81,15 @@
     @Test
     @OperateOnDeployment(CLIENT_DEPLOYMENT)
     public void remotePlusHttpMandatoryBean() {
-        ProviderUrlData providerUrl = new ProviderUrlData("remote+http", url.getHost(), serverPort);
+        ProviderUrlData providerUrl = new ProviderUrlData("remote+http", TestSuiteEnvironment.getServerAddressNode1(), serverPort);
```
bstansberry: This test now reads a system property so I think the set of permissions added to CLIENT_DEPLOYMENT in clientDeployment() needs to include a PropertyPermission.

bstansberry (changes requested):

kabir:
@istraka if you agree can you add the missing permission? Or if not, leave a message

kabir:
Superseded by 12779

PR #12594, opened 2019-09-06
[WFLY-12503] For -Delytron and -Dsecurity.manager testsuite runs disable surefire for all mo…

bstansberry:
…dules that don't handle that property specially

https://issues.jboss.org/browse/WFLY-12503


bstansberry:
The -Delytron case doesn't work for some reason. Closing for now.

PR #12595, opened 2019-09-06
Draft migration to jakarta.inject without requiring wilfly-core migration

yersan:
Please ignore, I just created it to test the CI.


yersan:
This wasn't as I expected, @james just in case if you find it useful. The CI results are here: https://ci.wildfly.org/viewLog.html?buildId=166671&tab=queuedBuildOverviewTab

PR #12600, opened 2019-09-06
[WFLY-12392] Removing javaee.api dependency from JSF modules

ashley-abdelsayed98:
https://issues.jboss.org/browse/WFLY-12392

### DIFF
--- a/feature-pack/src/main/resources/modules/system/layers/base/com/sun/jsf-impl/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/com/sun/jsf-impl/main/module.xml
@@ -29,7 +29,16 @@
 
     <dependencies>
         <module name="javax.faces.api"/>
-        <module name="javaee.api"/>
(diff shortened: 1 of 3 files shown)


fjuma (approved):

bstansberry (changes requested):
@fjuma @ashley-abdelsayed98 

I don't think all the 'export="true"' is correct.  The javaee.api module is basically just a kind of wrapper, so *it* depends on the detailed ones and then exports them so other modules like the JSF ones can access those classes.  But now the JSF modules are directly depending on the detailed ones so they get access to the classes directly. These 'export="true"' lines now have the effect that the JSF modules are exposing those classes to whatever other modules depend on the JSF ones. I don't think we want that.

fjuma:
Ah, good catch, @bstansberry.

ashley-abdelsayed98:
thanks @bstansberry, i've removed all the export="true"

bstansberry (approved):

PR #12603, opened 2019-09-06
Move some javax.* to jakarta.*

jamezp:
This includes a temporary commit to add the jakarta.json module to full. This module lives in core which for testing purposes should wait for this to be merged, then reverted once merged into core and core is released. This also pertains to the jakarta.inject dependency as well.

This includes the following as to not cause conflicts with having these separate commits:

- https://issues.jboss.org/browse/WFLY-12499
- https://issues.jboss.org/browse/WFLY-12501
- https://issues.jboss.org/browse/WFLY-12502

### DIFF
--- a/client/shade/pom.xml
+++ b/client/shade/pom.xml
@@ -216,12 +216,12 @@
         <!-- Required by the artemis-jms-client -->
         <dependency>
             <groupId>org.glassfish</groupId>
-            <artifactId>javax.json</artifactId>
+            <artifactId>jakarta.json</artifactId>
         </dependency>
 
         <dependency>
-            <groupId>javax.json</groupId>
-            <artifactId>javax.json-api</artifactId>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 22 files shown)


bstansberry (changes requested):
@jamezp I don't want new modules, definitely not 'jakarta.xxx' ones. These modules provide code in particular packages, and those packages are javax.xxx not 'jakarta.xxx'.  There's no need to change their names just because the source of an included resource has changed. Even org.glassfish.javax->org.glassfish.jakarta would require a fair bit of convincing.

When we need to deal with Jakarta EE 9 and there's a ton of package renaming we will probably need new modules. But if we introduce them now, when we have no strong plan for how we'll deal with EE 9, we'll have added complexity to our current life and may have closed off ways to deal with supporting both EE 8 and EE 9 by burning module names.

jamezp:
@bstansberry Okay no problem. I'll revert that part. I saw that EL created a new module and I should have just asked rather than assuming :)

bstansberry:
@jamezp what is the EL change you are referring to?

jamezp:
Ah I see. It's just the implementation module that changes from `org.glassfish.javax.el` to `org.glassfish.jakarta.el`. Do we want the implementation modules renamed?

jamezp:
The branch for the revert is: https://github.com/wildfly/wildfly/compare/master...jamezp:WFLY-12501-WFLY-12502-revert?expand=1

A PR can be opened for this once https://github.com/wildfly/wildfly-core/pull/3947 is done and a core upgrade is merged.

bstansberry:
@jamezp Can you add a commit adding the no-longer wanted GAs to the maven-enforcer-plugin ban-bad-dependencies execution in the root pom?

That will result in a failure if any of our deps are pulling that in without us having added an exclude in dependencyManagement.

I don't think that needs to be part of the commits we plan to revert as banning the deps in full is valid even though the versions we want will come in via core.

What's here LGTM. I think the org.glassfish.jakarta is ok. Not my favorite but I don't have a good argument and we've already done it for el.

bstansberry (approved):
Ignore my last request; it's easier to do that in a batch once these are all in.

PR #12604, opened 2019-09-06
WFLY-12511 Update community documentation for Jakarta EE 8

scottmarlow:
https://issues.jboss.org/browse/WFLY-12511
First pass of doc changes for Jakarta EE 8

### DIFF
--- a/docs/src/main/asciidoc/Getting_Started_Guide.adoc
+++ b/docs/src/main/asciidoc/Getting_Started_Guide.adoc
@@ -23,16 +23,15 @@ ifdef::basebackend-html[toc::[]]
 
 WildFly {wildflyVersion} is the latest release in a series of JBoss open-source
 application server offerings. WildFly {wildflyVersion} is an exceptionally fast,
-lightweight and powerful implementation of the Java Enterprise Edition {javaee_version}
+lightweight and powerful implementation of the Jakarta
 Platform specifications. The state-of-the-art architecture built on the
 Modular Service Container enables services on-demand when your
-application requires them. The table below lists the Java Enterprise
-Edition {javaee_version} technologies and the technologies available in WildFly {wildflyVersion}
+application requires them. The table below lists the technologies available in WildFly {wildflyVersion}
 server configuration profiles.
 
 [cols=",,,,",options="header"]
 |=======================================================================
-|Java EE {javaee_version} Platform Technology |Java EE {javaee_version} Full Profile |Java EE {javaee_version} Web
+|Jakarta EE Platform Technology |Jakarta EE Full Profile |Jakarta EE Web
 Profile |WildFly {wildflyVersion} Full Profile |WildFly {wildflyVersion} Web Profile
 
 |JSR-356: Java API for Web Socket |X |X |X |X
@@ -54,7 +53,7 @@ Profile |WildFly {wildflyVersion} Full Profile |WildFly {wildflyVersion} Web Pro
 
 |JSR-352: Batch Applications for the Java Platform 1.0 |X |-- |X |--
 
-|JSR-236: Concurrency Utilities for Java EE 1.0 |X |X |X |X
+|JSR-236: Concurrency Utilities for Jakarta EE |X |X |X |X
 
 |JSR-365: Contexts and Dependency Injection for Java 2.0 |X |X |X |X
 
@@ -67,7 +66,7 @@ Profile |WildFly {wildflyVersion} Full Profile |WildFly {wildflyVersion} Web Pro
 
 |JSR-318: Interceptors 1.2 |X |X |X |X
 
-|JSR-322: Java EE Connector Architecture 1.7 |X |-- |X |X
+|JSR-322: Jakarta EE Connector Architecture 1.7 |X |-- |X |X
 
 |JSR-338: Java Persistence 2.2 |X |X |X |X
 
@@ -100,13 +99,13 @@ Profile |WildFly {wildflyVersion} Full Profile |WildFly {wildflyVersion} Web Pro
 
 |JSR-115: Java Authorization Contract for Containers 1.5 |X |-- |X |--
 
-|JSR-88: Java EE Application Deployment 1.2 |Optional |-- |-- |--
+|JSR-88: Jakarta EE Application Deployment 1.2 |Optional |-- |-- |--
 
 |JSR-77: J2EE Management 1.1 |X |  |X | 
 
 |JSR-45: Debugging Support for Other Languages 1.0 |X |X |X |X
 
-|JSR-375: Java EE Security API 1.0 |X |X |X |X
+|JSR-375: Jakarta EE Security API 1.0 |X |X |X |X
 |=======================================================================
 
 Missing ActiveMQ Artemis and JMS?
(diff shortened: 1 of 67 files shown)


docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Concurrency_Utilities_Configuration.adoc
```diff
@@ -1,8 +1,8 @@
 [[EE_Concurrency_Utilities_Configuration]]
 = EE Concurrency Utilities
 
-EE Concurrency Utilities (JSR 236) were introduced with Java EE 7, to
```
scottmarlow: This should either remain as Java EE 7 or perhaps change to "EE Concurrency Utilities were introduced to ease the task of writing multithreaded Jakarta EE applications."  @emmartins what do you prefer?

bstansberry: +1 to just dropping "Java EE 7".

Tangent: IMHO in general this kind of "introduced with XXX" text sprinkled through docs is more maintenance trouble than it is worth. Typically XXX is "WildFly <some_number> but we see here that even Java EE X adds some cost.

docs/src/main/asciidoc/_developer-guide/EJB_invocations_from_a_remote_client_using_JNDI.adoc
```diff
@@ -332,7 +332,7 @@ name are as follows:
 *app-name* : This is the name of the .ear (without the .ear suffix) that
 you have deployed on the server and contains your EJBs.
 
-* Java EE 6 allows you to override the application name, to a name of
+* Jakarta EE 8 allows you to override the application name, to a name of
```
bstansberry: WDYT about just dropping the '8' here and in most of the similar places where you're doing this kind of version correction?

If EE ever evolves such that the doc chunks with these #s included are no longer spec-required etc, there would need to be a doc change regardless.  I doubt having #s in the text in some spots would help much in finding places to correct, so I don't think the #s add much value.

Of course there are places where the # is required, e.g. anything where we described what spec version we support.

bstansberry:
@scottmarlow Please ping me when this is ready to go.

docs/src/main/asciidoc/_developer-guide/How_do_I_migrate_my_application_from_AS7_to_WildFly.adoc
```diff
@@ -16,13 +16,13 @@ The purpose of this guide is to document changes that are needed to successfully
 
 The list of {ProductName} new functionality is extensive, being the most relevant, with respect to server and application migrations:
 
-* Java EE7 - {ProductName} is a certified implementation of Java EE7, meeting both the Web and the Full profiles, and already includes support for the latest iterations of CDI (1.2) and Web Sockets (1.1).
+* Jakarta EE 8 - {ProductName} is a certified implementation of Jakarta EE 8, meeting both the Web and the Full profiles, and already includes support for the latest iterations of CDI (1.2) and Web Sockets (1.1).
```
jamezp: Full is not called "Jakarta EE 8 Full Platform" and the web is still called "Jakarta EE 8 Web Profile". We may need to change the full reference(s) to Platform.

ee/src/main/java/org/jboss/as/ee/concurrent/ManagedScheduledExecutorServiceImpl.java
```diff
@@ -38,7 +38,7 @@
 import static org.jboss.as.ee.concurrent.ControlPointUtils.doWrap;
 
 /**
- * WildFly's extension of Java EE 7 RI {@link org.glassfish.enterprise.concurrent.ManagedScheduledExecutorServiceImpl}.
+ * WildFly's extension of Jakarta EE 7 RI {@link org.glassfish.enterprise.concurrent.ManagedScheduledExecutorServiceImpl}.
```
jamezp: The version here should likely be removed.

jamezp (commented):
I think for the most part this looks okay. I do think we should remove the version number, e.g. 8, where we can or at least change it to use the property `{javaee_version}` which should also likely be renamed.

I also think we should remove the various API versions where we can. I think some of them are wrong anyway. There is a reference to JAX-RS 2.0 and we're at 2.1.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/EE_Concurrency_Utilities_Configuration.adoc
scottmarlow: Updating to:

> EE Concurrency Utilities (JSR 236) were introduced to ease the task of writing multithreaded applications. Instances of these utilities are managed by WildFly, and the related configuration.

docs/src/main/asciidoc/_developer-guide/EJB_invocations_from_a_remote_client_using_JNDI.adoc
scottmarlow: makes sense, thanks.

scottmarlow: just updated locally, will push changes soon, commits should be squashed but will leave separate for now.

docs/src/main/asciidoc/Getting_Started_Guide.adoc
```diff
@@ -54,7 +54,7 @@ Profile |WildFly {wildflyVersion} Full Profile |WildFly {wildflyVersion} Web Pro
 
 |JSR-352: Batch Applications for the Java Platform 1.0 |X |-- |X |--
 
-|JSR-236: Concurrency Utilities for Java EE 1.0 |X |X |X |X
+|JSR-236: Concurrency Utilities for Jakarta EE |X |X |X |X
```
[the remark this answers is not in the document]
scottmarlow: Thanks James, I just removed the others also.

docs/src/main/asciidoc/Getting_Started_Guide.adoc
```diff
@@ -289,26 +289,26 @@ process.
 ===== Standalone Server Configurations
 
 * standalone.xml (_default_)
-** Java Enterprise Edition {javaee_version} web profile certified configuration with
+** Jakarta Enterprise Edition {javaee_version} web profile certified configuration with
```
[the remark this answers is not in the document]
scottmarlow: I will change:
  Jakarta Enterprise Edition
To:
Jakarta EE

Good catch, there are two of these.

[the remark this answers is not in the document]
scottmarlow: Also removing  {javaee_version}

docs/src/main/asciidoc/_developer-guide/How_do_I_migrate_my_application_from_AS7_to_WildFly.adoc
scottmarlow: Good catch, updating to "Full Profile".  Platform could also work

ee/src/main/java/org/jboss/as/ee/concurrent/ManagedScheduledExecutorServiceImpl.java
scottmarlow: I think that we can also remove Jakarta, will change to:
>  WildFly's extension of {@link org.glassfish.enterprise.concurrent.ManagedScheduledExecutorServiceImpl}.

scottmarlow:
I'll change JAX-RS 2.0, to just JAX-RS and push changes.  

I think that we need at least one more pass to check if there are new Jakarta names for things like JAX-RS, which according to https://www.eclipse.org/lists/jaxrs-dev/msg00680.html should be "Jakarta RESTful Web Services" or something like that.  

We should use https://jakarta.ee/legal/acronym_guidelines/ as a guide.

scottmarlow:
Do we really need to change from "EJB" to "Jakarta Enterprise Beans"?  We will discuss this elsewhere and update this pull request after.

scottmarlow:
We will start a separate branch for renaming JAX-RS + EJB + many others, will update this pr to not include the JAX-RS (cleanup) and EJB changes.

scottmarlow:
@bstansberry @jamezp the changes in this pr are now complete.

scottmarlow:
dependency tree error is:

> Canceled (Error while applying patch; cannot find commit 00cf8e2db397753b4b5beca60618f5d4b6a4d345 in the https://github.com/wildfly/wildfly.git repository, possible reason: refs/pull/12604/merge branch was updated and the commit selected for the buil...
> Canceled with comment: Build revision not found 

need to figure out the cause of that

scottmarlow:
This is a common failure that others have seen

bstansberry (approved):

PR #12606, opened 2019-09-08
Make ServiceLoader module dependencies on org.wildfly.clustering.web.…

bstansberry:
…infinispan optional to avoid requiring infinispan subsystem

Currently just for CI test. @pferraro FYI.

When running tests of a Galleon-trimmed server using the web-server and datasources layers, SharedSessionTestCase would fail as InfinispanLegacySessionManagementProviderFactory was getting discovered and would wire in infinispan service deps, but the config did not have the infinispan subsystem.

The only module that has compile-time deps on clustering/web/infinispan is clustering/web/extension so the other deps are for service loading and should be optional.


bstansberry:
Superceded by #12614

PR #12610, opened 2019-09-09
[WFLY-12515]: page-size-bytes should be an INT instead of a LONG

ehsavoie:
* pageSizeBytes is now an int instead of a long.
* Uupdating the XML schema accordingly.

Jira: https://issues.jboss.org/browse/WFLY-12515


kabir:
@ehsavoie The AddressSettingsTestCase failure seems related

bstansberry:
retest this please

PR #12614, opened 2019-09-09
WFLY-12505 Dependency cleanup of clustering modules

pferraro:
Also, WFLY-12506 Document purpose of each clustering module

https://issues.jboss.org/browse/WFLY-12505
https://issues.jboss.org/browse/WFLY-12506

### DIFF
--- a/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/main/module.xml
@@ -42,7 +42,7 @@
         <module name="org.jboss.jandex"/>
         <module name="org.jboss.logging"/>
         <module name="org.jboss.marshalling"/>
-        <module name="org.jgroups"/>
+        <module name="org.jgroups" optional="true"/>
         <module name="org.reactivestreams"/>
         <module name="sun.jdk"/>
     </dependencies>
--- a/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/clustering/spi/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/clustering/spi/main/module.xml
@@ -33,14 +33,12 @@
 
     <dependencies>
         <module name="org.jboss.as.clustering.common"/>
-        <module name="org.jboss.as.controller"/>
-        <module name="org.jboss.modules"/>
-        <module name="org.jboss.msc"/>
         <module name="org.wildfly.clustering.api"/>
-        <module name="org.wildfly.clustering.ejb.infinispan" services="import"/>
-        <module name="org.wildfly.clustering.server" services="import"/>
+        <!-- Enable dynamic installation of clustering services for EJB affinity support -->
+        <module name="org.wildfly.clustering.ejb.infinispan" services="import" optional="true"/>
+        <!-- Enable dynamic installation of clustering services per channel or cache -->
+        <module name="org.wildfly.clustering.server" services="import" optional="true"/>
         <module name="org.wildfly.clustering.service"/>
         <module name="org.wildfly.clustering.singleton"/>
-        <module name="org.wildfly.clustering.web.infinispan" services="import"/>
     </dependencies>
 </module>
--- a/clustering/api/pom.xml
+++ b/clustering/api/pom.xml
@@ -40,6 +40,9 @@
     <packaging>jar</packaging>
 
     <name>WildFly: Clustering public API</name>
+    <description>
+        This module contains a public API for a variety of clustering capabilities for use by both applications and other subsystems.
+    </description>
 
     <dependencies>
         <dependency>
--- a/clustering/common/pom.xml
+++ b/clustering/common/pom.xml
@@ -40,6 +40,10 @@
     <packaging>jar</packaging>
 
     <name>WildFly: Common code for clustering subsystems</name>
+    <description>
+        The code in this module is not explicitly related to clustering, but rather contains resuable code used by clustering modules
+        and any modules that integrate with clustering.
+    </description>
 
     <dependencies>
         <dependency>
--- a/clustering/ee/cache/pom.xml
+++ b/clustering/ee/cache/pom.xml
@@ -60,7 +60,6 @@
         <dependency>
             <groupId>org.jboss.spec.javax.transaction</groupId>
             <artifactId>jboss-transaction-api_1.3_spec</artifactId>
-            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.kohsuke.metainf-services</groupId>
--- a/clustering/ee/cache/pom.xml
+++ b/clustering/ee/cache/pom.xml
@@ -40,6 +40,9 @@
     <packaging>jar</packaging>
 
     <name>WildFly: Common EE implementations for caches</name>
+    <description>
+        This module contains code that is common to the wildlfy-clustering-ee-hotrod and wildlfy-clustering-ee-infinispan modules.
+    </description>
 
     <dependencies>
         <dependency>
--- a/clustering/ee/hotrod/pom.xml
+++ b/clustering/ee/hotrod/pom.xml
@@ -38,6 +38,9 @@
     <packaging>jar</packaging>
 
     <name>WildFly: EE clustering - HotRod service provider</name>
+    <description>
+        This module contains code that is common to the wildlfy-clustering-ejb-hotrod and wildlfy-clustering-web-hotrod modules.
+    </description>
 
     <dependencies>
         <dependency>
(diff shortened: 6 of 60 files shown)


feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/clustering/spi/main/module.xml
```diff
@@ -33,14 +33,12 @@
...
-        <module name="org.jboss.as.controller"/>
-        <module name="org.jboss.modules"/>
-        <module name="org.jboss.msc"/>
         <module name="org.wildfly.clustering.api"/>
-        <module name="org.wildfly.clustering.ejb.infinispan" services="import"/>
-        <module name="org.wildfly.clustering.server" services="import"/>
+        <!-- Enable dynamic installation of clustering services for EJB affinity support -->
+        <module name="org.wildfly.clustering.ejb.infinispan" services="import" optional="true"/>
+        <!-- Enable dynamic installation of clustering services per channel or cache -->
+        <module name="org.wildfly.clustering.server" services="import" optional="true"/>
```
bstansberry: In my testing work I found I needed to ensure Galleon provisioned this if the JGroups subsystem was used, i.e. via

https://github.com/wildfly/wildfly/pull/12508/commits/a23ae9cf642eced338d5149353c422d38cfea2e7#diff-4f073585cd66264ee958d5a02ee1d60bR272

I'm not sure if that will still be necessary given all the other changes here but I suspect it will. I'm rebasing my branch onto this one and will see what happens.

feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/main/module.xml
```diff
@@ -42,7 +42,7 @@
         <module name="org.jboss.jandex"/>
         <module name="org.jboss.logging"/>
         <module name="org.jboss.marshalling"/>
-        <module name="org.jgroups"/>
+        <module name="org.jgroups" optional="true"/>
```
bstansberry: My impress from chat is this might not be safe in a situation where the infinispan subsystem is provisioned for local caching but nothing else in the config results in jgroups being provisioned.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/clustering/spi/main/module.xml
bstansberry: Yep, testsuite/layers LayersTests fails trying to boot a server provisioned with 'jpa-distributed'.

```
22:05:31,821 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-1) MSC000001: Failed to start service org.wildfly.clustering.jgroups.channel-module.ee: org.jboss.msc.service.StartException in service org.wildfly.clustering.jgroups.channel-module.ee: java.lang.IllegalArgumentException: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server
	at org.wildfly.clustering.service.FunctionalService.start(FunctionalService.java:70)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:1738)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1700)
	at org.jboss.msc.service.ServiceControllerImpl$ControllerTask.run(ServiceControllerImpl.java:1558)
	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:1982)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1486)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1377)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalArgumentException: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server
	at org.jboss.as.clustering.controller.ModuleServiceConfigurator.get(ModuleServiceConfigurator.java:80)
	at org.jboss.as.clustering.controller.ModuleServiceConfigurator.get(ModuleServiceConfigurator.java:48)
	at org.wildfly.clustering.service.FunctionalService.start(FunctionalService.java:67)
	... 8 more
Caused by: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server
	at org.jboss.modules.ModuleLoader.loadModule(ModuleLoader.java:297)
	at org.jboss.as.clustering.controller.ModuleServiceConfigurator.get(ModuleServiceConfigurator.java:78)
	... 10 more

22:05:31,934 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "jgroups"),
    ("channel" => "ee")
]) - failure description: {"WFLYCTL0080: Failed services" => {"org.wildfly.clustering.jgroups.channel-module.ee" => "java.lang.IllegalArgumentException: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server
    Caused by: java.lang.IllegalArgumentException: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server
    Caused by: org.jboss.modules.ModuleNotFoundException: org.wildfly.clustering.server"}}
```

bstansberry: That commit I linked above is admittedly a convoluted way to do that. The org.jboss.as.clustering.jgroups module.xml could just have the dep, with a comment about why it's needed

feature-pack/src/main/resources/modules/system/layers/base/org/infinispan/main/module.xml
pferraro: The infinispan subsystem itself ensures that the jgroups module is provisioned.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/clustering/spi/main/module.xml
pferraro: The attempt to load this module is loaded based on the default value of this attribute:
https://github.com/wildfly/wildfly/blob/master/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/ChannelResourceDefinition.java#L112

bstansberry: @pferraro Yep. That's been there for at least 2 years, so to remove it would require some sort of compatibility handling.  If we don't do that we need to be sure the module gets provisioned.

BTW, my PR layered on top of this got good results -- https://github.com/wildfly/wildfly/pull/12508

Both in the standard CI jobs and in the custom jobs that test against Galleon slimmed servers:

https://ci.wildfly.org/viewLog.html?buildId=167028&

https://ci.wildfly.org/viewLog.html?buildId=167029&

But that had a commit to deal with this issue. I think that problem would be better addressed in this PR.  Other than that this LGTM.

pferraro: > That's been there for at least 2 years, so to remove it would require some sort of compatibility handling. If we don't do that we need to be sure the module gets provisioned.

We wouldn't want to remove it - as this attribute configures the module from which channel-based services configure their marshaller.

What I meant to say is that we don't want to hardcode the module to be loaded in the ResourceDefinition constructor - as this value is dynamic.

bstansberry: @pferraro There's no facility for telling Galleon a module is needed based on the value of an attribute. If a user wants to refer to a custom module they'll need to ensure Galleon provisions it by explicitly declaring the package in their provisioning.xml or making it a default package in config or layer they provide. But in this default case we don't want them to have to do that so I think we need to hard code it.

@jfdenise @ehsavoie I don't know if there's a possible future enhancement here; i.e. AttributeDefinition.getArbitraryDescriptors() can look for a "module-name" arbitrary descriptor and if found put something in the feature spec which galleon somehow uses to require a package that matches the configured value.  That couldn't work though if the attribute allowed an expression, which this one does.

pferraro: @bstansberry OK.  So that means it needs to be non-optional in module.xml.

pferraro: Addressed by https://github.com/wildfly/wildfly/pull/12614/commits/19401d7fd1392c123d850e2e4e114db648cde772

bstansberry (approved):

bstansberry:
Thanks, Paul!

PR #12621, opened 2019-09-10
[WFLY-12526] Ensure isUserInRole uses a PrivilegedAction as the deployments ProtectionDomain is not relevant.

darranl:
https://issues.jboss.org/browse/WFLY-12526

Reviewing the call stack in the Jira issue the deployment's ProtectionDomain should be dropped as not revevant for the ongoing call.


darranl:
@fl4via If you could please take a look at this one as it is in the Undertow subsystem.

fl4via (approved):
I agree with your analysis on why we should add this privileged block @darranl

PR #12623, opened 2019-09-10
[WFLY-12530] Add doPrivileged calls to access the AuthConfigFactory.

darranl:
https://issues.jboss.org/browse/WFLY-12530

The stack trace is shown on the Jira issue, the JASPICSecurityContext use of AuthConfigFactory is an internal detail - the deployment is making use of standard servlet APIs so should not require these permissions to be granted.


darranl:
@fl4via If you could please take a look at this one as it is in the Undertow subsystem.

bstansberry (approved):

PR #12626, opened 2019-09-11
[WFLY-12537] Incoming RunAsPrincipal is not being propagated to an unsecured EJB

skyllarr:
https://issues.jboss.org/browse/WFLY-12537

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/InjectPrincipalTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/InjectPrincipalTestCase.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.test.integration.weld.builtinBeans;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.Assert;
+import org.junit.runner.RunWith;
+
+
+/**
+ * See <a href="https://issues.jboss.org/browse/WFLY-11587">WFLY-11587</a>.
+ */
+@RunWith(Arquillian.class)
+public class InjectPrincipalTestCase {
+
+    public static final String ANONYMOUS_PRINCIPAL = "anonymous";
+    public static final String NON_ANONYMOUS_PRINCIPAL = "non-anonymous";
+
+    @Deployment
+    public static Archive<?> createTestArchive() {
+        return ShrinkWrap.create(WebArchive.class).addPackage(InjectPrincipalTestCase.class.getPackage())
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
+                .addAsWebInfResource(InjectPrincipalTestCase.class.getPackage(), "jboss-ejb3.xml", "jboss-ejb3.xml");
+    }
+
+    @Test
+    public void testAnonymousPrincipalInjected(BeanWithInjectedPrincipal beanA, BeanWithPrincipalFromEJBContext beanB) {
+        try {
+            Assert.assertEquals(ANONYMOUS_PRINCIPAL, beanA.getPrincipalName());
+            Assert.assertEquals(ANONYMOUS_PRINCIPAL, beanB.getPrincipalName());
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
+        }
+    }
+
+    @Test
+    public void testNonAnonymousPrincipalInjected(CallerWithIdentity callerWithIdentity) throws Exception {
+        try {
+            // Assert.assertEquals(NON_ANONYMOUS_PRINCIPAL, callerWithIdentity.getCallerPrincipalInjected()); TODO see issue WFLY-12538
+            Assert.assertEquals(NON_ANONYMOUS_PRINCIPAL, callerWithIdentity.getCallerPrincipalFromEJBContext());
+            Assert.assertEquals(NON_ANONYMOUS_PRINCIPAL, callerWithIdentity.getCallerPrincipalFromEJBContextSecuredBean());
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
+        }
+    }
+}
--- a/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
@@ -55,6 +55,7 @@ import javax.transaction.UserTransaction;
 import org.jboss.as.core.security.ServerSecurityManager;
 import org.jboss.as.ee.component.BasicComponent;
 import org.jboss.as.ee.component.ComponentView;
+import org.jboss.as.ee.component.interceptors.InvocationType;
 import org.jboss.as.ejb3.component.allowedmethods.AllowedMethodsInformation;
 import org.jboss.as.ejb3.component.interceptors.ShutDownInterceptorFactory;
 import org.jboss.as.ejb3.component.invocationmetrics.InvocationMetrics;
@@ -647,13 +648,19 @@ public abstract class EJBComponent extends BasicComponent implements ServerActiv
     }
 
     private SecurityIdentity getCallerSecurityIdentity() {
-        if (incomingRunAsIdentity != null) {
-            return incomingRunAsIdentity;
-        } else if (securityRequired) {
-            return securityDomain.getCurrentSecurityIdentity();
+        InvocationType invocationType = CurrentInvocationContext.get().getPrivateData(InvocationType.class);
+        boolean isRemote = invocationType != null && invocationType.equals(InvocationType.REMOTE);
+        if (isRemote) {
+            if (incomingRunAsIdentity != null) {
+                return incomingRunAsIdentity;
+            } else if (securityRequired) {
+                return securityDomain.getCurrentSecurityIdentity();
+            } else {
+                // unsecured EJB
+                return securityDomain.getAnonymousSecurityIdentity();
+            }
         } else {
-            // unsecured EJB
-            return securityDomain.getAnonymousSecurityIdentity();
+            return (incomingRunAsIdentity == null) ? securityDomain.getCurrentSecurityIdentity() : incomingRunAsIdentity;
         }
     }
 
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/BeanWithInjectedPrincipal.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/BeanWithInjectedPrincipal.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.test.integration.weld.builtinBeans;
+
+import java.security.Principal;
+
+import javax.enterprise.context.SessionScoped;
+import javax.inject.Inject;
+
+import java.io.Serializable;
+
+@SessionScoped
+public class BeanWithInjectedPrincipal implements Serializable {
+
+    @Inject
+    Principal principal;
+
+    public String getPrincipalName() {
+        return principal.getName();
+    }
+}
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/BeanWithPrincipalFromEJBContext.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/BeanWithPrincipalFromEJBContext.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.test.integration.weld.builtinBeans;
+
+import javax.ejb.Stateless;
+import javax.ejb.EJBContext;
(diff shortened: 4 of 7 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponentDescription.java
```diff
@@ -1188,22 +1188,19 @@ public String toString() {
...
         // Next interceptor: run-as-principal
-        // Switch users if there's a run-as principal
-        if (runAsPrincipal != null) {
-            interceptorFactories.put(InterceptorOrder.View.RUN_AS_PRINCIPAL, new ImmediateInterceptorFactory(new RunAsPrincipalInterceptor(runAsPrincipal)));
-
+        if (runAsPrincipal == null && !propagateSecurity) {
+            interceptorFactories.put(InterceptorOrder.View.RUN_AS_PRINCIPAL, new ImmediateInterceptorFactory(new RunAsPrincipalInterceptor(RunAsPrincipalInterceptor.ANONYMOUS_PRINCIPAL, false)));
+        } else {
+            // Switch users (set new incoming)
+            interceptorFactories.put(InterceptorOrder.View.RUN_AS_PRINCIPAL, new ImmediateInterceptorFactory(new RunAsPrincipalInterceptor(runAsPrincipal, isSecurityRequired())));
```
Skyllarr: @darranl We have to add RunAsPrincipalInterceptor to beans that do not have RunAsPrincipal annotation and propagate security. This is to be able to propagate incoming runas identity further. It should behave the same as legacy.

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
```diff
@@ -35,16 +39,30 @@
...
     private final String runAsPrincipal;
+    private final boolean securityRequired;
 
-    public RunAsPrincipalInterceptor(final String runAsPrincipal) {
+    public RunAsPrincipalInterceptor(final String runAsPrincipal, boolean securityRequired) {
         this.runAsPrincipal = runAsPrincipal;
+        this.securityRequired = securityRequired;
+    }
+
+    private boolean remotingContextIsSet (){
```
Skyllarr: @darranl  Not sure if I should leave this method here or I should create for example SecurityActions class helper for this package or reuse some existing ones (i.e. make them public)?

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
```diff
@@ -61,9 +79,15 @@ public Object processInvocation(final InterceptorContext context) throws Excepti
         final SecurityIdentity oldIncomingRunAsIdentity = ejbComponent.getIncomingRunAsIdentity();
         SecurityIdentity newIdentity;
         try {
-            // The run-as-principal operation should succeed if the current identity is authorized to
-            // run as a user with the given name or if the caller has sufficient permission
-            if (runAsPrincipal.equals(ANONYMOUS_PRINCIPAL)) {
+            boolean remotingContextIsSet = remotingContextIsSet();
+            boolean isRemoteStatelessSecurityDisabled = !(ejbComponent instanceof StatefulSessionComponent) && remotingContextIsSet && !securityRequired;
```
Skyllarr: make principal anonymous and incoming identity null if  isRemoteStatelessSecurityDisabled

Skyllarr: not sure if this is the best way to identify stateless bean

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/security/RemoteIdentityTestCase.java
```diff
@@ -99,12 +99,4 @@ public void testSwitched() throws Exception {
         assertEquals("user1", targetBean.getPrincipalName("user1", "password1"));
     }
 
-    @Test
```
Skyllarr: @darranl I don't think this test has valid objective. EntryBean is unsecured stateless remote bean, so its principal is anonymous. This bean should not propagate guest identity to the injected SecurityInformation EJB in my opinion. The returned identity should be anonymous and not guest IMO. I do not think this test should be present for Elytron if the test objective is invalid.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/weld/builtinBeans/InjectPrincipalTestCase.java
```diff
@@ -0,0 +1,64 @@
...
+            Assert.assertEquals(ANONYMOUS_PRINCIPAL, beanB.getPrincipalName());
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
+        }
+    }
+
+    @Test
+    public void testNonAnonymousPrincipalInjected(CallerWithIdentity callerWithIdentity) throws Exception {
+        try {
+            // Assert.assertEquals(NON_ANONYMOUS_PRINCIPAL, callerWithIdentity.getCallerPrincipalInjected()); TODO see issue WFLY-12538
```
Skyllarr: @nziakova I added your test cases from issue [WFLY-11604](https://issues.jboss.org/browse/WFLY-11604). I commented out this one because it still needs to be addressed (see https://issues.jboss.org/browse/WFLY-12538)

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
jaikiran: Hello @Skyllarr, you can use a instanceof check against `org.jboss.as.ejb3.component.stateless.StatelessSessionComponent` to be sure if it's a stateless EJB. In its current form, the above code in the PR, even singleton session bean component will pass this check. I don't know if you want that to happen.
Having said that, I'm just curious - is there a reason why this (security) logic is only applicable for stateless beans?

Skyllarr: @jaikiran Hello, thanks for review! I changed it to StatelessSessionComponent. This way the identity will not be propagated if the bean is a stateless unsecured remote bean. I did not find any reference for this, but failing tests lead me this way and it also made sense to me that I should not propagate identity for such a bean.

darranl:
Retest this please

darranl:
@tadamski / @chengfang / @fjuma Looks like this one could use some additional review, I am not overly familiar myself with the EJB integration to answer Diana's questions.

Skyllarr:
Retest this please

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/security/RemoteIdentityTestCase.java
fjuma: @Skyllarr I think it is correct that the injected bean is making use of the identity from the Remoting connection.

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
```diff
@@ -35,16 +39,30 @@
...
         this.runAsPrincipal = runAsPrincipal;
+        this.securityRequired = securityRequired;
+    }
+
+    private boolean remotingContextIsSet (){
+        RemoteConnection remoteConnection;
+        if(WildFlySecurityManager.isChecking()) {
+            remoteConnection = WildFlySecurityManager.doUnchecked((PrivilegedAction<RemoteConnection>) () -> RemotingContext.getRemoteConnection());
+        } else {
+            remoteConnection = RemotingContext.getRemoteConnection();
```
fjuma: The `RemotingContext` class is meant to be used for legacy security, I don't think we should be using it for Elytron.

ejb3/src/main/java/org/jboss/as/ejb3/security/EJBSecurityViewConfigurator.java
```diff
@@ -145,9 +145,7 @@ public void configure(DeploymentPhaseContext context, ComponentConfiguration com
         }
 
         final boolean securityRequired = beanHasMethodLevelSecurityMetadata || ejbComponentDescription.hasBeanLevelSecurityMetadata();
-        if (securityRequired) {
-            ejbComponentDescription.setSecurityRequired(securityRequired);
-        }
+        ejbComponentDescription.setSecurityRequired(securityRequired);
```
fjuma: `setSecurityRequired` should only be called if `securityRequired` is `true` (see https://issues.redhat.com/browse/WFLY-12301).

fjuma:
@Skyllarr I think we should take a step back and see if there is another way to solve this problem. One thing to note is that the meaning of the `incomingRunAsIdentity` in `RunAsPrincipalInterceptor` is slightly different for the Elytron case than it was for legacy security. For Elytron, if there is a `RunAsPrincipal` annotation specified on an EJB, the `RunAsPrincipalInterceptor` will use `securityDomain.createAdHocIdentity(runAsPrincipal)` or `currentIdentity.createRunAsIdentity(runAsPrincipal)` to get the new identity to run as. The interceptor will then call `newIdentity.runAs` to switch users so this new identity will be used for outgoing calls. The purpose of the `incomingRunAsIdentity` variable is only to make sure that the `RunAsPrincipal` annotation doesn't affect the caller's identity (see https://github.com/wildfly-security-incubator/wildfly/pull/11/commits/08b5940ac9928ccfac5335fdcb4bb68781c9530c).

Taking a look at `InjectPrincipalTestCase#testNonAnonymousPrincipalInjected`:
* We have a bean with a `RunAsPrincipal` annotation that has a method that invokes an unsecured bean.
* When `callerWithIdentity.getCallerPrincipalFromEJBContext()` is called, we get to the `RunAsPrincipalInterceptor`, where a `newIdentity` is correctly created using the "non-anonymous" principal from the `RunAsPrincipal` annotation. The "non-anonmyous" identity will be used for outgoing calls.
* The `incomingRunAsIdentity` is just set to the anonymous identity prior to switching users since that was the current identity when `callerWithIdentity.getCallerPrincipalFromEJBContext()` was called.
* When we eventually end up in `EJBComponent.getCallerSecurityIdentity`, `securityDomain.getCurrentSecurityIdentity()` will return the "non-anonymous" user. The problem is that because `beanB` is an unsecured bean, `securityRequired` evaluates to `false` so we just return the anonymous identity (see https://github.com/wildfly/wildfly/blob/2f14ebd211bd64fe55c615e2b2dd7e9f0ad88711/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java#L648-L653).

The tricky part is that we do want the anonymous identity to be returned for the Remoting -> unsecured EJB case (as per https://issues.redhat.com/browse/WFLY-8414). So it seems like we only want the "non-anonymous" identity to be returned for the EJB -> unsecured EJB case. Perhaps you could try to see if there is a way to differentiate between these two cases in `EJBComponent#getCallerPrincipal` so that `securityDomain.getCurrentSecurityIdentity()` gets returned in the latter case.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/security/RemoteIdentityTestCase.java
Skyllarr: @fjuma I see, after reading https://github.com/wildfly/wildfly/pull/12626#issuecomment-568093027 it probably makes sense to keep it here.

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
Skyllarr: Ok it might not be needed

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/security/RemoteIdentityTestCase.java
Skyllarr: @fjuma I see, after reading https://github.com/wildfly/wildfly/pull/12626#issuecomment-568093027 it probably makes sense to keep it here.

ejb3/src/main/java/org/jboss/as/ejb3/security/RunAsPrincipalInterceptor.java
Skyllarr: @fjuma Ok it might not be needed

ejb3/src/main/java/org/jboss/as/ejb3/security/EJBSecurityViewConfigurator.java
Skyllarr: @fjuma Ok reverted

Skyllarr:
> The tricky part is that we do want the anonymous identity to be returned for the Remoting -> unsecured EJB case (as per https://issues.redhat.com/browse/WFLY-8414). 

Is it okay that this would not match the behaviour of legacy? If I understand it correctly legacy would reauthenticate from remoting to unsecured bean and return non-anonymous: 

https://github.com/wildfly/wildfly/blob/master/security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java#L321

> So it seems like we only want the "non-anonymous" identity to be returned for the EJB -> unsecured EJB case. Perhaps you could try to see if there is a way to differentiate between these two cases in `EJBComponent#getCallerPrincipal` so that `securityDomain.getCurrentSecurityIdentity()` gets returned in the latter case.

So it is important to distinguish if the caller was from EJB or not. I do not see a way to lookup this information in the current code. So maybe we could add it to the security identity (something like `boolean isEJB` ) and set it in RunAsPrincipalInterceptor to have the appropriate behaviour.

fjuma:
> > The tricky part is that we do want the anonymous identity to be returned for the Remoting -> unsecured EJB case (as per https://issues.redhat.com/browse/WFLY-8414).
> 
> Is it okay that this would not match the behaviour of legacy? If I understand it correctly legacy would reauthenticate from remoting to unsecured bean and return non-anonymous:
> 
> https://github.com/wildfly/wildfly/blob/master/security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java#L321
> 

For the Remoting -> unsecured EJB case, the anonymous identity was returned when using legacy security. There's a description of why the anonymous identity gets returned here:

https://issues.redhat.com/browse/JBEAP-9744?focusedCommentId=13399451&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13399451

Skyllarr:
> For the Remoting -> unsecured EJB case, the anonymous identity was returned when using legacy security. There's a description of why the anonymous identity gets returned here:
> 
> https://issues.redhat.com/browse/JBEAP-9744?focusedCommentId=13399451&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13399451

@fjuma Yes sorry you're right.

>So it is important to distinguish if the caller was from EJB or not. I do not see a way to lookup this information in the current code. So maybe we could add it to the security identity (something like boolean isEJB ) and set it in RunAsPrincipalInterceptor to have the appropriate behaviour.

@fjuma Do you think it would be good to do it this way?

fjuma:
> > For the Remoting -> unsecured EJB case, the anonymous identity was returned when using legacy security. There's a description of why the anonymous identity gets returned here:
> > https://issues.redhat.com/browse/JBEAP-9744?focusedCommentId=13399451&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13399451
> 
> @fjuma Yes sorry you're right.
> 
> > So it is important to distinguish if the caller was from EJB or not. I do not see a way to lookup this information in the current code. So maybe we could add it to the security identity (something like boolean isEJB ) and set it in RunAsPrincipalInterceptor to have the appropriate behaviour.
> 
> @fjuma Do you think it would be good to do it this way?

I don't think this should be set in `RunAsPrincipalInterceptor` as this interceptor is already correctly making sure the correct identity is used for outgoing calls. I think we should take a closer look at `EJBComponent#getCallerPrincipal` to try to distinguish between the two cases there. It looks like `CurrentInvocationContext.get()` could be used there to get the interceptor context. I wonder if we could possibly make use of `InterceptorContext.getPrivateData()` to help us distinguish between the two cases by either trying to make use of existing data or perhaps by adding something new that could be used?

Skyllarr:
@fjuma I have updated the PR accordingly. Thanks!

fjuma:
Thanks @Skyllarr, this fix looks good to me. I think it would be good to check if the new tests being added here along with the existing test cases now cover the different scenarios (i.e., Remoting -> unsecured EJB, Remoting -> secured EJB, EJB -> unsecured EJB, EJB -> secured EJB) to make sure these are all working properly.

Skyllarr:
> Thanks @Skyllarr, this fix looks good to me. I think it would be good to check if the new tests being added here along with the existing test cases now cover the different scenarios (i.e., Remoting -> unsecured EJB, Remoting -> secured EJB, EJB -> unsecured EJB, EJB -> secured EJB) to make sure these are all working properly.

@fjuma Remoting -> secured or unsecured EJB are covered in `org.jboss.as.test.integration.ejb.remote.security.RemoteIdentityTestCase`. I added test that covers EJB -> secured EJB just now. Thanks

fjuma (approved):

fjuma:
Thanks @Skyllarr.

Skyllarr:
rebased

bstansberry:
Thanks @Skyllarr

PR #12629, opened 2019-09-11
More javax to Jakarta moves

jamezp:
https://issues.jboss.org/browse/WFLY-12439
https://issues.jboss.org/browse/WFLY-12544


jamezp:
I forgot to note the test fix from @fl4via seems correct to me. It does resolve expressions that were previously not resolved if the servlet spec 3.0 or lower was defined, however the RI does this as well.

bstansberry (approved):

PR #12630, opened 2019-09-12
[WFLY-12540] Move from javax.activation:activation:1.1.1 to com.sun.a…

bstansberry:
…ctivation:jakarta.activation:1.2.1

https://issues.jboss.org/browse/WFLY-12540

@asoldano fyi in case there's a reason not to move to com.sun.activation (which is the eclipse-ee4j artifact.) I mention this because the pom's cxf entries have exclusions for com.sun.activation:javax.activation. OTOH this change means we are getting closer to the CXF dep set, which seems good. OTOH maybe there was a reason we stayed on the older javax.activation:activation and you guys know what it is. :)

### DIFF
--- a/bean-validation/pom.xml
+++ b/bean-validation/pom.xml
@@ -124,8 +124,8 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>javax.activation</groupId>
-            <artifactId>activation</artifactId>
+            <groupId>com.sun.activation</groupId>
+            <artifactId>jakarta.activation</artifactId>
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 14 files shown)


asoldano (commented):
@bstansberry this looks correct to me. Let's tag @jimma here too in case he has additional comments / different opinion.

jamezp (commented):
This looks good, but does need one more change before it can be merged. In the `testsuite/mixed-domain/pom.xml` an exclusion is needed on the `org.wildfly.core:wildfly-core-testsuite-shared`.

```
        <dependency>
            <artifactId>wildfly-core-testsuite-shared</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>javax.activation</groupId>
                    <artifactId>activation</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

After that is done this is good to go IMO.

jamezp (approved):

PR #12645, opened 2019-09-15
[WFLY-12542] MP Metrics should require the MP Config capability.

bstansberry:
Also, if services from other capabilities depend on ConfigProviderService being started before they can use MP Config, then have the MP Config capability install a void service they can depend upon.  Have MP Metrics depend on that instead of an internal impl service of the config subsystem.

https://issues.jboss.org/browse/WFLY-12542


bstansberry:
@jmesnil WDYT?

If this is right we should add an entry to wildfly-capabilities. Actually that's needed regardless.

bstansberry:
I'm not sure though whether the capability is actually required; i.e. whether it can be optional and then the service dependency only added if the cap is present. Thinking being that if the metrics modules depend on the mp config API module and it depends on the smallryte-config stuff, then the default ConfigProviderResolver.instance() provider loading may work just fine. Any functionality of the subsystem is gone, but basic stuff like system prop and env var config sources may work just fine.

This is just speculation though.

jmesnil:
@bstansberry Your approach looks fine and I think the capability is actually required.
I need to confirm but your patch likely also fix https://issues.jboss.org/browse/WFLY-12549 that deals exactly with your example: the mp.metrics.tags that is used by smallrye-metrics could come from any ConfigSource, including one specified in the microprofile-config-smallrye subsystem.

jmesnil (approved):

PR #12647, opened 2019-09-15
[WFLY-12560 Convert code to use the ServerReload methods that take Ma…

bstansberry:
…nagementClient. Deprecate the ModelControllerClient variants. Improve ServerReload javadoc.

https://issues.jboss.org/browse/WFLY-12560

Besides being a code cleanup, this is a further attempt to avoid intermittent failures in multinode tests, which we are still seeing.

### DIFF
--- a/testsuite/compat/src/test/java/org/jboss/as/test/compat/jpa/hibernate/transformer/AbstractVerifyHibernate51CompatibilityTestCase.java
+++ b/testsuite/compat/src/test/java/org/jboss/as/test/compat/jpa/hibernate/transformer/AbstractVerifyHibernate51CompatibilityTestCase.java
@@ -490,7 +490,7 @@ public abstract class AbstractVerifyHibernate51CompatibilityTestCase {
         public void tearDown(ManagementClient managementClient, String s) throws Exception {
             ModelNode op = Operations.createRemoveOperation(PROP_ADDR_ENABLETRANSFORMER);
             managementClient.getControllerClient().execute(op);
-            ServerReload.executeReloadAndWaitForCompletion(managementClient.getControllerClient());
+            ServerReload.executeReloadAndWaitForCompletion(managementClient);
         }
     }
 }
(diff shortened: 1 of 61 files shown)


yersan (changes requested):
Additionally to my previous question for review, I still can see the uses of deprecated methods in WildFly I think should be changed with this PR, e.g:
https://github.com/wildfly/wildfly/blob/master/testsuite/compat/src/test/java/org/jboss/as/test/compat/jpa/hibernate/transformer/AbstractVerifyHibernate51CompatibilityTestCase.java#L485

Another detail is we currently have two versions of ServerReload:
https://github.com/wildfly/wildfly/blob/master/testsuite/shared/src/main/java/org/jboss/as/test/shared/ServerReload.java

And the other comming from WildFly core test suite:
https://github.com/wildfly/wildfly-core/blob/master/testsuite/shared/src/main/java/org/jboss/as/test/integration/management/util/ServerReload.java

Do we want to continue keeping them independently? if so, WildFly should not use the ServerReload version comming from WildFly core, currently, we are using it in some tests, and we should apply the improvements to the core one too. Otherwise, we should apply all the changes in the WildFly core version and use that one in WildFly and delete the WildFly one

WDYT?

bstansberry:
@yersan 

Thanks for noticing AbstractVerifyHibernate51CompatibilityTestCase.java -- I'll have a look but I assume I just missed it.

Also, thanks for the inputs re WildFly Core.

I don't want to convert full WildFly to using the core version. The ManagementClient class is an Arquillian class and Arq should not be a dep of core.

OTOH, you're right, there's code in full using the core ServerReload and mostly it's taking a ManagementClient and extracting the ModelControllerClient to make the call.  In this PR I should update those calls to use the full ServerReload passing the ManagementClient.

If the core version of ServerReload can be improved, that would be a separate task. I think the priority of that would be based on whether we see problems related to *necessary* use of core ServerReload. 

In general I'm not a fan of using core testsuite/shared in full WildFly. That's a really vague statement, and there are many cases where it's absolutely appropriate. The problem is testsuite/shared is not providing a proper contract. There's some stuff in there that's definitely designed for use across different code bases, and then other stuff that people have added over time but it doesn't really provide a strong enough contract to be used between two different code bases. Some day we might refactor things or something to separate out the two categories. But anyway I have no basic problem with duplicated test utility code in core and full.

bstansberry:
@yersan I updated the existing commit to fix the things you noted about it. Then I added a 2nd commit cleaning up the use of core's ServerReload.

Thanks for the review!

yersan (commented):
Sure, np, you're welcome @bstansberry 

LGTM, however, I still have one pending to change class on my radar: https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ConnectionFactoryManagementTestCase.java#L118

bstansberry:
@yersan I need to take my radar in to the shop to get it tuned up; your's is working better than mine!

That test is fixed.

yersan (approved):

PR #12651, opened 2019-09-16
[WFLY-12557] Fix CDI injection of health check probes

jmesnil:
* Use BeanManager select() to get references on injected beans *managed*
  by CDI (instead of creating manually unmanaged references
* add missing javax.entreprise.api to
  org.eclipse.microprofile.health.api JBoss module
* remove microprofile-tck/health/tck-suite.xml file added to exclude CDI
  producer test now that all tests pass again

JIRA: https://issues.jboss.org/browse/WFLY-12557


jmesnil:
This issue is not blocker for WildFly 18 but it solves an issue highlighted by MP Metrics 2.1 TCK.
With that fix, we pass all of MP Metrics 2.1 TCK (and @manovotn confirmed it may solve nasty issues where CDI managed beans would conflict with the unmanaged instances created by the subsystem)

jmesnil:
@manovotn could you please review and approve?

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
```diff
@@ -51,84 +45,66 @@
...
-                Unmanaged.UnmanagedInstance<HealthCheck> healthCheckInstance = unmanagedHealthCheck.newInstance();
-                HealthCheck healthCheck = healthCheckInstance.produce().inject().postConstruct().get();
-                instances.add(healthCheckInstance);
-                healthFunction.accept(healthCheck);
-            } catch (Exception e) {
-                throw new RuntimeException("Failed to register health bean", e);
-            }
+        Instance<HealthCheck> healthChecks = bm.createInstance().select(HealthCheck.class, qualifier);
+        for (HealthCheck healthCheck : healthChecks) {
+            healthFunction.accept(healthCheck);
```
manovotn: Can the bean instances obtained here be `@Dependent` beans?
If so then you are responsible for disposing them properly once you no longer need then in order to avoid leaks. CDI cannot control `@Dependent` bean's lifecycle if you retrieve them via `Instance` because normally the bean lifecycle is linked to the bean injecting it, but obviously, that won't work here.
You can destroy it with `Instance.destroy(healthCheck)` for every dependent bean you obtained (you can do that even for proxied beans but I am not sure that's desirable in this case?).

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
```diff
@@ -51,84 +45,66 @@
...
+    private void removeHealthCheck(AnnotationLiteral qualifier, BeanManager bm,
                                    Consumer<HealthCheck> healthFunction) {
-        instances.forEach(healthCheck -> {
-            healthFunction.accept(healthCheck.get());
-            healthCheck.preDestroy().dispose();
-        });
-        instances.clear();
+        Instance<HealthCheck> healthChecks = bm.createInstance().select(HealthCheck.class, qualifier);
+        for (HealthCheck healthCheck : healthChecks) {
+            healthFunction.accept(healthCheck);
```
manovotn: Note that every invocation of `get()` on an `Instance` results in *new* injection which in case of `@Dependent` bean means a whole new bean. So you would be attempting to remove health check on different beans then where you added it.

Best way would be to store the set of beans the first time you got them in `addHealthCheck()` and then in `removeHealthCheck()` you would just iterate over them and finally do `Instance.destroy()` on all of them (since it happens before shutdown, it should be safe)?

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
```diff
@@ -51,84 +45,66 @@
...
-    private List<AnnotatedType<? extends HealthCheck>> livenessDelegates = new ArrayList<>();
-    private Collection<Unmanaged.UnmanagedInstance<HealthCheck>> livenessInstances = new ArrayList<>();
+        static final HealthLiteral INSTANCE = new HealthLiteral();
 
-    private List<AnnotatedType<? extends HealthCheck>> readinessDelegates = new ArrayList<>();
-    private Collection<Unmanaged.UnmanagedInstance<HealthCheck>> readinessInstances = new ArrayList<>();
+        private static final long serialVersionUID = 1L;
 
+    }
+    static final class LivenessLiteral extends AnnotationLiteral<Liveness> implements Liveness {
```
manovotn: Would it be worth MP Health spec issue to add these literals there in the future?
Ideally, along with `@Target` on those qualifiers ;)

manovotn (changes requested):
Added some remarks.

Also, I am seeing test failures in `org.wildfly.test.integration.microprofile.health`.

manovotn:
And one more note - if `SmallRyeHealthReporter` (the original one) gets picked up as bean, then you may want to `veto()` that. From what I understood, it is never going to be used anyway.

If you want to `veto` a bean, you add a PAT (`ProcessAnnotatedType`) observer into CDI extension that looks like this:

```
    public void vetoSrHealthReporter(@Observes ProcessAnnotatedType<SmallRyeHealthReporter> pat) {
       pat.veto();
    }
```

jmesnil:
@manovotn I've addressed all your points, thanks for the review

jmesnil:
damn, a new CI job has been triggered but all jobs had previously passed: https://ci.wildfly.org/project/WF_PullRequest?branch=pull%2F12651

bstansberry:
@jmesnil The Windows JDK 8 job had a metrics tck failure.

bstansberry:
@jmesnil I'm going to defer to Matej on this one.

jmesnil:
@bstansberry I don't think this PR (that relates to MP *Health*) is causing the failure in MP *Metrics*.

I'll check the test but given his name `ConcurrentGaugedClassBeanTest.countedMethodsNotCalledYet` and it's the 1st failure on 1679 runs, I suspect it's a race condition in the test.

jmesnil:
@bstansberry jan confirmed that the failure is due to the test fixture (https://gitter.im/smallrye-io/metrics?at=5d80914436461106bb31c9cc) and will require a change in MP Metrics TCK.

This does not prevent to merge this PR

manovotn (changes requested):
I think it's still wrong.
You are retrieving the instances twice, once in `addHealthChecks` and then in `removeHealthCheck`.
What you want to do is retrieve them in `addHealthChecks`, store all you got in some collection and then in `removeHealthCheck` you just iterate over that collection, do `healthFunction.accept(healthCheck);` and finally perform `instance.destroy(healthCheck)`.

Does that make sense @jmesnil ?

jmesnil:
@manovotn Unless I'm mistaken, the latest updated does what you suggested:

* I have 3 `List` to store the different `HealthChecks`: https://github.com/wildfly/wildfly/blob/6381a5b2182c07b8b980d809e88bbc98b14a32c3/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java#L74-L76

* in ADV, I pass the `List` corresponding to the Qualifier and store the created instance:
https://github.com/wildfly/wildfly/blob/6381a5b2182c07b8b980d809e88bbc98b14a32c3/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java#L97

* Finally, in BeforeShutdow, I use the 3 `List` to destroy the instances:
https://github.com/wildfly/wildfly/blob/6381a5b2182c07b8b980d809e88bbc98b14a32c3/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java#L107-L109
https://github.com/wildfly/wildfly/blob/6381a5b2182c07b8b980d809e88bbc98b14a32c3/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java#L114-L119

manovotn:
Oops, this was PEBKAC from me :)
You did exactly that, sorry, I was just blind.

The only other thing is that you are using different instance of `Instance` to invoke `destroy()`.
According to [javadoc of that method](https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/javax/enterprise/inject/Instance.java#L227-L230), you should be using the same one.
Under the hood, Weld shares the one you obtain via `BM.createInstance()` so in theory you are safe anyway, just letting you know that if you want to strictly stick to CDI spec, it should be different.

manovotn (approved):

jmesnil:
@manovotn thanks for your review, I've amended my commit to use the same instance of `Instance` for both ADV and BS events

bstansberry (approved):
Thanks for the review @manovotn

PR #12659, opened 2019-09-18
[WFLY-12541][WFLY-12566] Licenses cleanup

bstansberry:
https://issues.jboss.org/browse/WFLY-12541
https://issues.jboss.org/browse/WFLY-12566

Drops, at least for now, the additional GPLv2+CE file. Aligns the EPL 2 and GPLv2+CE entries to the standard names and urls.


yersan (commented):
I would like to add just a note to ensure that keeping them is the intention and we do not want to clean them now in this PR. We still are keeping two .txt files with the same content but a different name:

'gnu general public license, version 2 with the classpath exception.txt' and 'gnu general public license v2.0 only, with classpath exception.txt'

If that's the intention, please ignore this.

jamezp (approved):

bstansberry:
@yersan Good question. Yes; prior to WF 18 we had two files with the same content but different names. Keeping those was intentional. Really just to limit the scope of this.

AIUI ideally the 'name' for all entries would be a value in the short form 'Identifier' column at https://spdx.org/licenses and then there'd be the same URL and a single file for each. Cleaning that up would be a good task for a future release. We'd need to confirm first that the tooling that checks the licenses would be happy, and it's late in the WF 18 cycle for that. I think it's ok, as the suggestion to use the short form 'Identifier' came from Petr Sakar and he's the expert. But it's late to do it now.

PR #12662, opened 2019-09-19
[WFLY-12576] Eliminate StabilityMonitor usages to improve reliability.

ropalka:
https://issues.jboss.org/browse/WFLY-12576

Stability monitors will be deprecated in next MSC release due to their unreliability.
This PR should improve our test suite reliability.

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/secondlevelcache/InfinispanCacheDeploymentListener.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/secondlevelcache/InfinispanCacheDeploymentListener.java
@@ -27,17 +27,19 @@ import static org.jboss.as.jpa.messages.JpaLogger.ROOT_LOGGER;
 import java.security.AccessController;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
 import java.util.function.Supplier;
 
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.jboss.as.controller.ServiceNameFactory;
 import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.server.CurrentServiceContainer;
+import org.jboss.msc.service.LifecycleEvent;
+import org.jboss.msc.service.LifecycleListener;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.service.StabilityMonitor;
 import org.jipijapa.cache.spi.Classification;
 import org.jipijapa.cache.spi.Wrapper;
 import org.jipijapa.event.spi.EventListener;
@@ -93,25 +95,19 @@ public class InfinispanCacheDeploymentListener implements EventListener {
                 builder.requires(ServiceNameFactory.parseServiceName(InfinispanCacheRequirement.CONFIGURATION.getName()).append(container, cache));
             }
         }
-
-        ServiceController<?> controller = builder.install();
-
-        // Ensure cache configuration services are started
-        StabilityMonitor monitor = new StabilityMonitor();
-        monitor.addController(controller);
-        try {
-            monitor.awaitStability();
-            // TODO Figure out why the awaitStability() returns prematurely while there are still dependencies starting
-            while (controller.getState() == ServiceController.State.DOWN) {
-                Thread.yield();
-                monitor.awaitStability();
+        final CountDownLatch latch = new CountDownLatch(1);
+        builder.addListener(new LifecycleListener() {
+            @Override
+            public void handleEvent(final ServiceController<?> controller, final LifecycleEvent event) {
+                if (event == LifecycleEvent.UP) {
+                    latch.countDown();
+                    controller.removeListener(this);
+                }
             }
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw e;
-        } finally {
-            monitor.removeController(controller);
-        }
+        });
+        ServiceController<?> controller = builder.install();
+        // Ensure cache configuration services are started
+        latch.await();
 
         return new CacheWrapper(manager.get(), controller);
     }
@@ -155,21 +151,24 @@ public class InfinispanCacheDeploymentListener implements EventListener {
             if (ROOT_LOGGER.isTraceEnabled()) {
                 ROOT_LOGGER.tracef("stop second level cache by removing dependency on service '%s'", this.controller.getName().getCanonicalName());
             }
-            StabilityMonitor monitor = new StabilityMonitor();
-            monitor.addController(this.controller);
-            this.controller.setMode(ServiceController.Mode.REMOVE);
+            final CountDownLatch latch = new CountDownLatch(1);
+            controller.addListener(new LifecycleListener() {
+                @Override
+                public void handleEvent(final ServiceController<?> controller, final LifecycleEvent event) {
+                    if (event == LifecycleEvent.REMOVED) latch.countDown();
+                }
+            });
+            controller.setMode(ServiceController.Mode.REMOVE);
             try {
-                monitor.awaitStability();
-            } catch (InterruptedException e) {
+                latch.await();
+            } catch (InterruptedException ie) {
                 Thread.currentThread().interrupt();
-            } finally {
-                monitor.removeController(this.controller);
             }
         }
     }
 
     private static ServiceContainer currentServiceContainer() {
-        if(System.getSecurityManager() == null) {
+        if (System.getSecurityManager() == null) {
             return CurrentServiceContainer.getServiceContainer();
         }
         return AccessController.doPrivileged(CurrentServiceContainer.GET_ACTION);
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyBindingRegistry.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyBindingRegistry.java
@@ -26,15 +26,17 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.wildfly.extension.messaging.activemq.BinderServiceUtil.installBinderService;
 import static org.wildfly.extension.messaging.activemq.logging.MessagingLogger.ROOT_LOGGER;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.Locale;
 
 import org.apache.activemq.artemis.spi.core.naming.BindingRegistry;
+import org.jboss.msc.service.LifecycleEvent;
+import org.jboss.msc.service.LifecycleListener;
 import org.wildfly.extension.messaging.activemq.logging.MessagingLogger;
 import org.jboss.as.naming.ManagedReferenceFactory;
 import org.jboss.as.naming.deployment.ContextNames;
 import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
-import org.jboss.msc.service.StabilityMonitor;
 
 /**
  * A {@link BindingRegistry} implementation for WildFly.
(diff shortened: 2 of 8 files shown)


security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainResourceDefinition.java
```diff
@@ -226,19 +227,20 @@ protected void executeRuntimeStep(OperationContext context, ModelNode operation)
...
+            controller.addListener(listener);
+            latch.await();
         } catch (final InterruptedException e) {
             Thread.currentThread().interrupt();
             throw SecurityLogger.ROOT_LOGGER.interruptedWaitingForSecurityDomain(controller.getName().getSimpleName());
-        } finally {
-            monitor.removeController(controller);
-        }
-
+        } catch (final Throwable ignored) {}
```
pferraro: Shouldn't the lifecycle listener be removed in a finally block?

webservices/server-integration/src/main/java/org/jboss/as/webservices/publish/EndpointPublisherImpl.java
```diff
@@ -175,22 +178,25 @@ protected void doDeploy(ServiceTarget target, DeploymentUnit unit) {
...
             for (Endpoint ep : endpoints) {
-                final ServiceName serviceName = EndpointService.getServiceName(unit, ep.getShortName());
-                monitor.addController(registry.getRequiredService(serviceName));
-            }
-            try {
-                monitor.awaitStability();
-            } finally {
-                monitor.clear();
+                serviceName = EndpointService.getServiceName(unit, ep.getShortName());
+                registry.getRequiredService(serviceName).addListener(listener);
```
pferraro: Shouldn't this lifecycle listener be removed in a finally block?

pferraro (changes requested):

security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainResourceDefinition.java
ropalka: Of course it must be removed @pferraro - thanks for spotting that out.

webservices/server-integration/src/main/java/org/jboss/as/webservices/publish/EndpointPublisherImpl.java
ropalka: Of course it must be removed @pferraro - thanks for spotting that out.

ropalka:
Fixed @pferraro

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/secondlevelcache/InfinispanCacheDeploymentListener.java
```diff
@@ -92,25 +94,19 @@ public Wrapper startCache(Classification classification, Properties properties)
                 builder.requires(dependencyName);
             }
         }
-
+        final UninterruptibleCountDownLatch latch = new UninterruptibleCountDownLatch(1);
```
pferraro: Is there a particular reason for ignoring interrupts here?  Wouldn't it be more appropriate to propagate the InterruptedException to the caller?

pferraro (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/secondlevelcache/InfinispanCacheDeploymentListener.java
ropalka: No particular reason - fixed.

kabir:
Retest this please

kabir:
Retest this please

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/secondlevelcache/InfinispanCacheDeploymentListener.java
```diff
@@ -151,21 +148,20 @@ public void close() {
...
-            }
+            final UninterruptibleCountDownLatch latch = new UninterruptibleCountDownLatch(1);
+            controller.addListener(new LifecycleListener() {
+                @Override
+                public void handleEvent(final ServiceController<?> controller, final LifecycleEvent event) {
+                    if (event == LifecycleEvent.REMOVED) latch.countDown();
+                }
+            });
+            controller.setMode(ServiceController.Mode.REMOVE);
+            latch.awaitUninterruptibly();
```
pferraro: Why ignore interrupts here?

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyBindingRegistry.java
```diff
@@ -91,16 +90,20 @@ public void unbind(String name) {
...
+        });
         try {
-            monitor.awaitStability();
-            ROOT_LOGGER.unboundJndiName(bindInfo.getAbsoluteJndiName());
-        } catch (InterruptedException e) {
-            ROOT_LOGGER.failedToUnbindJndiName(name, 5, SECONDS.toString().toLowerCase(Locale.US));
+            bindingService.setMode(ServiceController.Mode.REMOVE);
         } finally {
-            monitor.removeController(bindingService);
+            latch.awaitUninterruptibly();
```
pferraro: Why ignore interrupts here?

webservices/server-integration/src/main/java/org/jboss/as/webservices/publish/EndpointPublisherImpl.java
```diff
@@ -175,22 +178,28 @@ protected void doDeploy(ServiceTarget target, DeploymentUnit unit) {
...
             }
+            latch.awaitUninterruptibly();
```
pferraro: Why ignore interrupts here?

pferraro (changes requested):

ropalka:
Fixed @pferraro - interrupts are not ignored anymore.

pferraro (approved):

bstansberry:
retest this please

bstansberry:
retest this please

darranl:
I have just kicked off CI again, this looks to be approved - if we can get a clean run this should likely be merged by now.

darranl:
Some execution timeouts on the CI run - but this is not the first PR I have seen those on today so kicked off the failed jobs again.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyBindingRegistry.java
```diff
@@ -91,16 +93,24 @@ public void unbind(String name) {
...
         try {
-            monitor.awaitStability();
-            ROOT_LOGGER.unboundJndiName(bindInfo.getAbsoluteJndiName());
-        } catch (InterruptedException e) {
-            ROOT_LOGGER.failedToUnbindJndiName(name, 5, SECONDS.toString().toLowerCase(Locale.US));
+            bindingService.setMode(ServiceController.Mode.REMOVE);
         } finally {
-            monitor.removeController(bindingService);
+            try {
+                latch.await(5, SECONDS);
```
bstansberry, line 110: Before this would wait until interrupted; now it has a fairly short 5 sec timeout. 

History shows that originally it was a 5 sec wait but a change you made in Jan 2014 made it indefinite. At this point I'm concerned that we have almost 7 years of bake with it this way, vs changing it back and hoping that not blocking is ok and that 5 secs is still a valid number.

security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainResourceDefinition.java
```diff
@@ -226,19 +227,21 @@ protected void executeRuntimeStep(OperationContext context, ModelNode operation)
...
+            final CountDownLatch latch = new CountDownLatch(1);
+            final LifecycleListener listener = new LifecycleListener() {
+                @Override
+                public void handleEvent(ServiceController<?> controller, LifecycleEvent event) {
+                    latch.countDown();
+                    controller.removeListener(this);
+                }
+            };
+            controller.addListener(listener);
+            latch.await();
```
bstansberry, line 240: AIUI this method will now wait indefinitely. The method javadoc says 100ms, so if that is going to change the javadoc should.

The concern with waiting indefinitely is it could hang the management operation in the management thread and with the exclusive management op lock held. The OperationContext would never time out on its own, as the thread that does that kind of thing is the one blocking. Either the client would have to cancel the op (if it's using remoting) or the user would need to use the management API to find and terminate the hanging op.

It's probably ok. The scenario where this could happen is if this is called in the middle of a composite op where 

step 1 installs the security domain, but not something it depends on
step 2 invokes this
step 3 installs whatever the security domain depends on

An unlikely scenario.

Before that would fail here; now it will hang. Presumably anyone who was doing that saw the failures and fixed their logic.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyBindingRegistry.java
ropalka: I returned it back to previous undefinite value.

security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainResourceDefinition.java
ropalka: I returned it back to previous value.

ropalka:
Sorry for the delay resolving your comments @bstansberry . Please review.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentCleanupProcessor.java
```diff
@@ -74,39 +71,26 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
-            if (controller != null) {
-                serviceControllers.add(controller);
-            }
+        // add dependency on our WeldBootstrapService and WeldStartService to ensure this goes after them
+        ServiceBuilder<?> weldStartCompletionServiceBuilder = serviceTarget.addService(weldStartCompletionServiceName);
+        final Supplier<WeldBootstrapService> bootstrapSupplier = weldStartCompletionServiceBuilder.requires(weldBootstrapServiceName);
+        weldStartCompletionServiceBuilder.requires(weldStartServiceName);
+        // require component start services from top level deployment
+        for (ServiceName componentStartSN : getComponentStartServiceNames(deploymentUnit)) {
+            weldStartCompletionServiceBuilder.requires(componentStartSN);
```
bstansberry, line 80: Why didn't this use service dependencies before instead of this StabilityMonitor thing? It seems like the obvious and far easier thing to have done.

There was something complicated about this service but I don't recall what it was.

bstansberry: @ropalka ^^^

bstansberry: https://issues.redhat.com/browse/WFLY-10784 is why. I don't completely understand but the switch from dependencies was deliberate.

ropalka: Anything depending on StabilityMonitors is broken @bstansberry @manovotn 
because StabilityMonitors might report false positives about achieved stability
(see https://issues.redhat.com/browse/MSC-255 ).
AFAICS WeldDeploymentCleanupProcessor is depending just on EE_MODULE_DESCRIPTOR
start component controllers and these should be always available (are not optional) if component is registered.
Further more I see there was a test case provided for WFLY-10784 and it is passing.

manovotn: Hmm, the switch was made because we needed to make sure our service goes last, even after optional deployments (which was the cause of the problem in the linked JIRA). I think the problem was that when the optional service failed to boot, Weld was stuck waiting on it. The stability monitor was a suggested way to fix that issue which seemed to have worked.
That being said, I know little of MSC, so I just went with the suggestion back then. There could have been a different way to achieve the same thing.

manovotn: If setting the dependencies does the trick of Weld going last, then it's probably fine. I am no MSC expert nor do I have a memory of why are all those services exactly this way. So let's just say that green tests are a good sign and in favor of accepting the change ;-)

bstansberry: Thanks @manovotn. It's true the test is passing, so this seems good. 👍

bstansberry (approved):

PR #12675, opened 2019-09-27
WFLY-12599 Add tests for MP Metrics tags

rsvoboda:
Add tests for MP Metrics tags, both declarative and programmatic approach
https://issues.jboss.org/browse/WFLY-12599

- [x] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue(s)
- [x] Pull Request contains description of the issue(s)
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Attached commits represent units of work and are properly formatted

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsTagsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/application/MicroProfileMetricsTagsTestCase.java
@@ -1,0 +1,97 @@
+package org.wildfly.test.integration.microprofile.metrics.application;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.test.integration.common.HttpRequest;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.test.integration.microprofile.metrics.TestApplication;
+import org.wildfly.test.integration.microprofile.metrics.application.resource.ResourceWithTags;
+
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getMetricValueFromPrometheusOutput;
+import static org.wildfly.test.integration.microprofile.metrics.MetricsHelper.getPrometheusMetrics;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+public class MicroProfileMetricsTagsTestCase {
+
+    @ContainerResource
+    ManagementClient managementClient;
+    @ArquillianResource
+    URL url;
+
+    String greetings = this.getClass().getPackage().getName().replaceAll("\\.", "_")
+            + "_resource_ResourceWithTags_greetings_total";
+    String unknownGreetings = "UnknownGreetings_total";
+
+    @Deployment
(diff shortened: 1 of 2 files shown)


jmesnil:
@rsvoboda afaict, there is nothing specific to WildFly in these tests, shouldn't they be contributed to MP Metrics TCK directly?

jmesnil:
or to smallrye-metrics?

rsvoboda:
My main concern is that none of those TS are executed against WF

jmesnil:
@rsvoboda MicroProfile Metrics TCK is executed against WildFly: for example https://ci.wildfly.org/viewLog.html?buildId=174740&buildTypeId=WF_MasterLinuxJdk11&tab=testsInfo

rsvoboda:
Adding tests to MP TCK is not solving need to have tests coverage present now.
It takes a lot of time to release new version of MP spec and there are no time guaranties.
I didn't notice MP specs micro releases just to increase test coverage in TCK.

rsvoboda:
@msve I think we need your input on this topic.

Both approaches (WF TS vs. MP TCK) have pros and cons.

msve:
I agree with Rostislav's approach, while TCK test have their place, they are too tied to the release cycle of the spec. WF TS seems to me like a better place to include more higher level tests and I think it will be in general more flexible for us.

rsvoboda:
Thanks, I also looked at mp metrics tck tests and they have different approach than what's used in WF TS, they inject registry or bean and do checks on them. No checks on the final output aka from outside of the runtime / from user facing side.

 - https://github.com/eclipse/microprofile-metrics/blob/master/tck/api/src/main/java/org/eclipse/microprofile/metrics/tck/metrics/
 - https://github.com/eclipse/microprofile-metrics/blob/master/tck/api/src/main/java/org/eclipse/microprofile/metrics/tck/tags
 - https://github.com/eclipse/microprofile-metrics/blob/master/tck/api/src/main/java/org/eclipse/microprofile/metrics/tck/metrics/CountedMethodTagBeanTest.java
 - https://github.com/eclipse/microprofile-metrics/blob/master/tck/api/src/main/java/org/eclipse/microprofile/metrics/tck/tags/TagsTest.java

jmesnil:
I won't hold the merge of these tests but I still think that WildFly test suite is not the right place for them.
They should belong to the MicroProfile Metrics TCK to assert that the metrics tag output is properly specified and tested. I can understand that this would mean that the tests would not be actually run before next MP Metrics release though.
However, in that case, I would still prefer to move these tests to smallrye-metrics as this is ultimately the component that is dealing with the metrics output. Besides, that would also benefit the whole smallrye ecosystem that way (including Quarkus)

I'm concerned that we would move spec or components tests into WildFly *integration* test suite when there is nothing specific to WildFly in them.

bstansberry:
@rsvoboda @jmesnil @msve  I don't mind having these temporarily in the WF test suite so long as there is a PR open to put them in a place where they more properly belong, either the spec TCK or smallrye.

We can clean them out once we move the component version that has the tests.

rsvoboda:
In my comment above I mentioned that MP TCK  take a different approach than tests in WF TS

> Thanks, I also looked at mp metrics tck tests and they have different approach than what's used in WF TS, they inject registry or bean and do checks on them. No checks on the final output aka from outside of the runtime / from user facing side.

 SRye tests are not running against WF but usually against `arquillian-weld-embedded` + `weld-core`

Back to Brian's comment
>I don't mind having these temporarily in the WF test suite so long as there is a PR open to put them in a place where they more properly belong, either the spec TCK or smallrye.

I would say that if development team thinks it would make sense to enhance TCK or smallrye they can proceed with it, but I don't think eap qe should go beyond WF repo and duplicate their effort.

darranl:
This PR presently looks stalled, I am going to add it to the agenda for our meeting on Monday.

rhusar:
retest this please

darranl:
@jmesnil We were discussing this one previously but you were away so just wanted to check with you if any final objections before we merge it?  I think it has been open long enough now that if we are not going to merge we should close the PR.

darranl:
I am just kicking off CI again, if we get a clean run I will merge.  We can always revert later but sitting in the PR queue for almost a year isn't helping either way.

@bstansberry FYI if no one has closed this PR by the time CI passes I will add the ready to merge label.

rsvoboda:
@msve do you want to include this PR into WildFly TS? I have no objections to close this tbh, it's stalled for quite some time.
PR was opened for quite a long time, triggered some discussion around the appropriate location of such tests.

jmesnil:
I'm closing this PR as it's been stalled for some time and is no longer is in the correct integration test suite

PR #12676, opened 2019-09-27
WFLY-12601 Remove use of StabilityMonitor in clustering modules

pferraro:
https://issues.jboss.org/browse/WFLY-12601

Removes use of one-time service installation by clustering runtime metrics/operations.
Also deprecates and removes all uses of problematic ActiveServiceSupplier/PassiveServiceSupplier, including within the clustering testsuite.


clustering/server/src/main/java/org/wildfly/clustering/server/singleton/ServiceLifecycle.java
```diff
@@ -57,36 +66,54 @@ public ServiceLifecycle(ServiceController<?> controller) {
...
+            Mode currentMode = this.controller.getMode();
+            Mode targetMode = transition.modeTransitions.get(currentMode);
+            if (targetMode == null) {
+                throw new IllegalStateException(currentMode.name());
             }
 
-            monitor.awaitStability();
+            this.controller.setMode(targetMode);
+
+            latch.await();
```
ropalka: This latch.await() has the same concurrency issues as had code code before. See WFLY-12601, issue (2) description.

ropalka (changes requested):
See issue (2) description in WFLY-12601

clustering/server/src/main/java/org/wildfly/clustering/server/singleton/ServiceLifecycle.java
pferraro: Can you elaborate?  Your comment in WFLY-12601 seems specific obtaining service values, but ServiceLifecycle does not read service values - it merely seeks to start or stop a given service by transitioning it between Mode.NEVER and Mode.ACTIVE.

clustering/service/src/main/java/org/wildfly/clustering/service/CountDownLifecycleListener.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public CountDownLifecycleListener(CountDownLatch latch, Set<LifecycleEvent> targetEvents) {
+        this.targetEvents = targetEvents;
+        this.latch = latch;
+    }
+
+    @Override
+    public void handleEvent(ServiceController<?> controller, LifecycleEvent event) {
+        if (this.targetEvents.contains(event)) {
+            this.latch.countDown();
```
ropalka: Don't forget to call controller.removeListener(this);
after this.latch.countDown(); statement

ropalka (changes requested):

clustering/service/src/main/java/org/wildfly/clustering/service/CountDownLifecycleListener.java
pferraro: That is already handled by the caller (i.e. the one who registers the listener also removes it).

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueCaptorServiceConfigurator.java
```diff
@@ -0,0 +1,73 @@
...
+        return this.captor.getServiceName().append("captor");
+    }
+
+    @Override
+    public ServiceBuilder<?> build(ServiceTarget target) {
+        ServiceBuilder<?> builder = target.addService(this.getServiceName());
+        return this.dependency.register(builder)
+                .setInstance(this)
+                .setInitialMode(ServiceController.Mode.PASSIVE);
+    }
```
ropalka: Excellent
 * our metrics collecting service has PASSIVE mode (to don't interact with observed value providing service)
 * it registers dependency on observed service on line 59 as we talked about
 * when observed service will START / STOP our metrics collecting service will also START / STOP
 * finally our metrics collecting service register / unregister itself with statistics engine in service lifecycle methods

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
```diff
@@ -0,0 +1,74 @@
...
+import org.wildfly.common.function.ExceptionFunction;
+
+/**
+ * Executes a given function against the captured value of a service.
+ * @author Paul Ferraro
+ */
+public class ServiceValueExecutor<T> implements ServiceValueCaptor<T>, FunctionExecutor<T> {
+
+    private final ServiceName name;
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
```
ropalka: This seems too heavyweight to me. I would say this lock is not needed at all.

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
```diff
@@ -0,0 +1,74 @@
...
+    public ServiceName getServiceName() {
+        return this.name;
+    }
+
+    @Override
+    public void accept(T value) {
+        Lock lock = this.lock.writeLock();
+        lock.lock();
+        try {
+            this.value = value;
```
ropalka: If you'll eliminate the lock, accept method will contain just "this.value = value;" statement

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    @Override
+    public <R, E extends Exception> R execute(ExceptionFunction<T, R, E> function) throws E {
+        Lock lock = this.lock.readLock();
+        lock.lock();
+        try {
+            T value = this.value;
+            return (value != null) ? function.apply(value) : null;
```
ropalka: And above two lines will do perfect job for you.
Yes, you could miss some statistics, but this is the case anyway
because there's a race condition between retrieving the executor and invoking it.

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutorRegistry.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void remove(ServiceName name) {
+        this.executors.remove(name);
+    }
+
+    @Override
+    public FunctionExecutor<T> get(ServiceName name) {
+        return this.executors.get(name);
+    }
+}
```
ropalka: And this is our statistics registry facade.

ropalka (approved):
You did excellent job @pferraro with this statistics collecting rewrite.
It satisfies MSC concurrency requirements as we discussed earlier.
My below comment about eliminating RW lock in ServiceValueExecutor is just a hint.

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
pferraro: The primary goal of the read lock is to prevent the backing service from stopping (stopping requires a write lock) while the executor runs.  If you think the RWLock is too heavy weight (we're talking about 1 per backing service, not per metric), we could probably just use a synchronization block.  WDYT?

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
pferraro: See write lock reasoning below.

clustering/common/src/main/java/org/jboss/as/clustering/controller/ServiceValueExecutor.java
ropalka: You better know @pferraro what kind of behaviour of your clustering services you want to achieve.
What I wanted to point out is sometimes synchronization using intrinsic lock or volatile variable might be sufficient comparing to heavyweight RWLock.
If RWLock is your design decision here I fully accept it.

ropalka: Anyway you were right my point was there is little value to allow concurrent threads to execute(...) on a given executor.

ropalka: To be 100% clear my proposal was you might delete ReetrantRWLock and rely just on volatile variable writes and accesses. There is no need to synchronize accept(...) and execute(...) methods. As result you would have less locks involved in your statistics gathering scenarios.
The drawback would be sometimes you could miss some statistics you're not missing ATM when ReentrantRWLock is at place. The question is how acceptable such statistics loss would be for you @pferraro ?

pferraro: The concern is not about missing statistics, as you pointed out, any service starting after a volatile executor read of null would be missed, but, I think, more critically, preventing a service from stopping (i.e. blocking an invocation of ServiceValueExecutor.accept(null)) after a volatile executor read of non-null until execution completes.
I would say that if MP metrics uses a single thread for gathering metrics, then synrchonized blocks are sufficient to get the job done.  RWLocks would only be beneficial if metrics are ever gathered by a multi-threaded process.  It's worth getting this right now, in the event that this becomes a reusable pattern for delivering metrics across the application server.
@jmesnil @bstansberry  Does MP metrics use multiple WF controller threads to gather runtime metrics?

ropalka: In my opinion the answer to question whether MP metrics use single or multiple threads for statistics gathering is irrelevant because it is implementation detail that can change any time.
It is better to assume multiple threads are used.

bstansberry: In any case it's single threaded.

bstansberry:
@pferraro If you'd like further review, please ping someone, otherwise unless @jmesnil wants to review please go ahead and merge.

PR #12699, opened 2019-10-07
Experiment -- don't configure the DEFAULT_MODULE

bstansberry:
Just an experiment.


bstansberry:
Doesn't fail any tests. May not be the right thing to do, but doesn't fail anything.

PR #12701, opened 2019-10-07
WFLY-12615 Upgrade jboss-ejb-client from 4.0.23 to 4.0.25

chengfang:
JIRA: https://issues.jboss.org/browse/WFLY-12615

Diff between jboss-ejb-client 4.0.23 and 4.0.25: https://github.com/wildfly/jboss-ejb-client/compare/4.0.23.Final...4.0.25.Final

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -342,7 +342,7 @@
         <version.org.jboss.activemq.artemis.integration>1.0.2</version.org.jboss.activemq.artemis.integration>
         <version.org.jboss.arquillian.core>1.4.0.Final</version.org.jboss.arquillian.core>
         <version.org.jboss.common.jboss-common-beans>2.0.1.Final</version.org.jboss.common.jboss-common-beans>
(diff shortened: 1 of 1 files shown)


bstansberry:
@chengfang In the past when CI fails like this for xxx-client upgrades it's been because CI uses -Dipv6 and that exposes a problem in IPv6 handling in the upgrade.

Or it could just be that the upgrade broke one thing as it seems the same test is failing everywhere.

chengfang:
@bstansberry this looks like a legit failure (see more info https://issues.jboss.org/browse/JBEAP-17344).  I'm looking into this.

jamezp:
@chengfang What is the status of this? It's holding up an EAP PR as well.

chengfang:
We're working on a more complete fix for [EJBCLIENT-347](https://issues.jboss.org/browse/EJBCLIENT-347) , which was the issue causing the regression.  cc @tadamski

bstansberry:
Superceded by #12816

PR #12706, opened 2019-10-10
[WFLY-12651] ModuleClassLoaderLocator created for every request when using default module

gaol:
Issue: https://issues.jboss.org/browse/WFLY-12651

When `jboss-module-name`s are defined for the login module, no matter default module(`org.picketbox`) is used or not, each `ModuleClassLoaderLocator.get(List<String> modules)` call(on each request) will create a new `ModuleClassLoaderLocator.CombinedClassLoader` instance.

The proposed fix is to introduce a cache in `ModuleClassLoaderLocator` for each specified ModuleClassLoaders and current context ClassLoader combination to avoid to create new instance on each request.

The cache won't be cleaned in the JVM, but there should not be much combinations, so the cache won't be big.

### DIFF
--- a/security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainAdd.java
+++ b/security/subsystem/src/main/java/org/jboss/as/security/SecurityDomainAdd.java
@@ -127,7 +127,6 @@ import org.wildfly.clustering.infinispan.spi.service.TemplateConfigurationServic
  */
 class SecurityDomainAdd extends AbstractAddStepHandler {
 
-    private static final String DEFAULT_MODULE = "org.picketbox";
     private static final String LEGACY_CACHE_NAME = "auth-cache";
 
     static final SecurityDomainAdd INSTANCE = new SecurityDomainAdd();
@@ -254,8 +253,6 @@ class SecurityDomainAdd extends AbstractAddStepHandler {
             ModelNode moduleName = LoginModuleResourceDefinition.MODULE.resolveModelAttribute(context, module);
             if (moduleName.isDefined() && !moduleName.asString().isEmpty()) {
                 mappingInfo.addJBossModuleName(moduleName.asString());
-            } else {
-                mappingInfo.addJBossModuleName(DEFAULT_MODULE);
             }
         }
 
@@ -281,8 +278,6 @@ class SecurityDomainAdd extends AbstractAddStepHandler {
             ModelNode moduleName = LoginModuleResourceDefinition.MODULE.resolveModelAttribute(context, module);
             if (moduleName.isDefined() && !moduleName.asString().isEmpty()) {
                 identityTrustInfo.addJBossModuleName(moduleName.asString());
-            } else {
-                identityTrustInfo.addJBossModuleName(DEFAULT_MODULE);
             }
         }
         applicationPolicy.setIdentityTrustInfo(identityTrustInfo);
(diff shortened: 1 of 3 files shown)


gaol:
@bstansberry Could you please review? Thanks :)

bstansberry:
Thanks @gaol.

This will leak classloaders. If we cache like this, the ModuleClassLoaderLocator object will need to expose a method to clear the cache, and the security subsystem will need to install a new DeploymentUnitProcessor that handles any undeploy by getting the ModuleClassLoaderLocator via ClassLoaderLocatorFactory.get() and invoking that method. (It's fine to be overly aggressive about clearing the cache.)

A more complex datastructure for the cache would be more efficient. A Map<ClassLoader, Map<List<String>, ClassLoader>>.  The key to the outer map is the TCCL, key to the inner is the List<String> passed to get(List<String>). This avoids any object allocations in the normal case of a cache hit.

The get(String key) method could be more efficient too, using Collections.singletonList instead of creating a single element ArrayList.

bstansberry:
Re: the DUP, if there's an existing one where clearing this cache is a valid addition to what the DUP is doing, using that instead of adding a new one is fine too.

gaol:
> A more complex datastructure for the cache would be more efficient. A Map<ClassLoader, Map<List, ClassLoader>>. The key to the outer map is the TCCL, key to the inner is the List passed to get(List). This avoids any object allocations in the normal case of a cache hit.

Are all TCCLs ModuleClassLoader? if they are, we may use `Map<List<String>, ClassLoader>` with just one TCCL's module name calculation before cache hit. Key is the list of all modules including TCCL's module name

gaol:
Updated with the following changes:
* Revert the addition of DEFAULT_MODULE(org.picketbox) in SecurityDomainAdd
* Tries to clear the cache item with current deployment's module identifier in SecurityDependencyProcessor.undeploy
* Updated the cache in ModuleClassLoaderLocator to use TCCL's module identifier as the key

bstansberry:
@gaol Casting to ModuleClassLoader is probably fine but the module name does not form the equals() contract for ModuleClassLoader. I think it's less fragile to just completely clear the cache and not try to be precise. It's not a problem if something gets thrown away that doesn't need to be as it's not a big deal to create another one on the next invocation.

gaol:
@bstansberry Thank you for the suggestions.

Updated the PR with following changes:
* Use TCCL as the key for the outer map of the cache
* Clear the cache aggressively when the deployment is un-deployed

If we agree to just do the part of reverting DEFAULT_MODULE in SecuriyDomainAdd only for the moment as discussed in ML and https://issues.jboss.org/browse/JBEAP-17756, please just re-open your PR: https://github.com/wildfly/wildfly/pull/12699, and I will update this one for the cache only. :)

gaol:
@sguilhen Would you please review as well? Thank you very much! :smile:

bstansberry:
@gaol LGTM. Re whether or not to do the caching, I'm happy to defer to you and @sguilhen -- I'm fine either way. If you want me to reopen my PR, please just let me know.

sguilhen (approved):
Looks good to me too

gaol:
@bstansberry @sguilhen  Thank you all for the review.  :)

In that case, I would prefer to do the caching, it should fix all potential cases that too much `ModuleClassLoaderLocator$CombinedClassLoader` instances crated, it won't have risks of ClassLoader leaks because of aggressive caching clear strategy.  WDYT? @sguilhen :)

sguilhen:
@gaol sounds reasonable to me to include the cache

bstansberry (approved):

PR #12713, opened 2019-10-15
WFLY-12674 wildfly.jpa.skipquerydetach set to true, will now skip clearing of query results

scottmarlow:
https://issues.jboss.org/browse/WFLY-12674

Some more AS5 compatibility changes that we missed in WFLY-3674


scottmarlow:
The eap720.MixedDomain720TestSuite timeout error doesn't look related to this change.

scottmarlow:
The Linux JDK11 failures is related to this change I think, the "Failed to setup Weld contexts: javax.naming.NameNotFoundException: java:comp/BeanManager" sounds like the "java:comp/BeanManager" isn't available early enough.

scottmarlow:
I'm seeing this "Failed to setup Weld contexts: javax.naming.NameNotFoundException: java:comp/BeanManager" locally now.

scottmarlow:
I might need to move the test changes to a different test, as I think its because <property name="jboss.as.jpa.classtransformer" value="true"/> is needed for the current test but bean validators are not available early enough for non-Hibernate persistence providers. :(

scottmarlow:
The Linux jdk8 failure should of been fixed in last pushed commit (teamcity shows org.jboss.as.test.integration.jpa.mockprovider.classtransformer.ClassFileTransformerTestCase test_withSkipQueryDetachDisabled failures which are out of date now).

scottmarlow:
We seem to still get the "javax.naming.NameNotFoundException: java:comp/BeanManager" failure for some of these tests.

scottmarlow:
I updated [EAP7-1365] to reflect that this change is ready I think.  @kabir is anything missing for this pr?

kabir:
@scottmarlow There is no test plan linked

scottmarlow:
@kabir I updated the https://github.com/wildfly/wildfly-proposals/pull/255 to mention the test plan inline, which is to update WF testsuite to ensure that the feature works as expected when enabled and disabled.  Are you expecting a separate test plan document for this change?

The covering test changes are:
[Test with the feature enabled](https://github.com/wildfly/wildfly/pull/12713/commits/d74ec3191332936d6920a19564281ce6bf682c6f#diff-599391f3739caa5d8342e38ccc09416eR79).

[Test with the feature disabled](https://github.com/wildfly/wildfly/pull/12713/commits/d74ec3191332936d6920a19564281ce6bf682c6f#diff-599391f3739caa5d8342e38ccc09416eR109).

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/mockprovider/skipquerydetach/persistence.xml
```diff
@@ -0,0 +1,46 @@
...
+  -->
+
+<persistence xmlns="http://java.sun.com/xml/ns/persistence" version="2.0">
+    <persistence-unit name="queryresultsaremanaged">
+        <provider>org.jboss.as.test.integration.jpa.mockprovider.skipquerydetach.TestPersistenceProvider</provider>
+        <description>Persistence Unit.</description>
+        <jta-data-source>java:jboss/datasources/ExampleDS</jta-data-source>
+        <properties>
+            <property name="jboss.as.jpa.adapterClass" value="org.jboss.as.test.integration.jpa.mockprovider.skipquerydetach.TestAdapter"/>
+            <!-- set skipquerydetach to false so that query results (in non-jta tx) stay associated with the persistence context. -->
```
zhantemirov: A typo? set skipquerydetach to `true`

scottmarlow: good catch @zhantemirov I will correct

scottmarlow: I pushed an update to correct the typo.  Thanks!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/mockprovider/skipquerydetach/SFSB1.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+        Query query =  emQueryResultsAreDetached.createQuery("mock query");
+        if(TestEntityManager.getInvocations().contains("clear")) {
+            return "invalid state, entity manager was cleared (detached) after query was created but before executed," +
+                    "which means that 'wildfly.jpa.skipquerydetach=false' (default) behaviour didn't work";
+        }
+        query.getSingleResult();
+        if(TestEntityManager.getInvocations().contains("close")) {
+            return "invalid state, entity manager was closed before SFSB1 bean call completed, " +
+                    "which means that 'wildfly.jpa.skipquerydetach=true' couldn't work, as the " +
```
zhantemirov: If I understand this correctly, we are testing the case with `wildfly.jpa.skipquerydetach=false`, aren't we? Should it be `false` in that case?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/mockprovider/skipquerydetach/SFSB1.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public String queryWithSkipQueryDetachDisabled() {
+        if(TestEntityManager.getInvocations().contains("close")) {
+            return "invalid state, entity manager has already been previously closed";
+        }
+        if(TestEntityManager.getInvocations().contains("clear")) {
+            return "invalid state, entity manager has already been previously cleared";
+        }
+        Query query =  emQueryResultsAreDetached.createQuery("mock query");
+        if(TestEntityManager.getInvocations().contains("clear")) {
```
zhantemirov: And one more thing.. As we test the same `TestPersistenceProvider` (with the same `adapterClass`) but two different instances (which only differ in `skipquerydetach` property) - is there a reason that we don't have `TestEntityManager.getInvocations().contains("close")` check?

zhantemirov:
Besides the provided feedback - all LGTM, thanks.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/mockprovider/skipquerydetach/SFSB1.java
scottmarlow: On line 92, we do have "close" check, which shouldn't be needed but if there is an unexpected bug in the mock persistence provider EntityManager implementation, anything is possible, so best to check for if close or clear had already been called, which we do (lines 92 + 95).  

Then, after the query is created, we check again if clear has been called on line 99.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/mockprovider/skipquerydetach/SFSB1.java
scottmarlow: oh, right, good catch, line 106 should be updated.  Thanks!

bstansberry (approved):

docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
```diff
@@ -858,6 +858,12 @@ persistence context is closed (e.g. business method ends). Defaults to
 false (entities are cleared after EntityManager invocation) and if set
 to true, the detach is deferred until the context is closed.
 
+|wildfly.jpa.skipquerydetach |Controls whether transaction scoped
```
bstansberry, line 861: s/whether/whether a/g

docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
```diff
@@ -858,6 +858,12 @@ persistence context is closed (e.g. business method ends). Defaults to
...
+persistence context used in non-JTA transaction thread, will detach
```
bstansberry, line 862: s/used in/used in a/g
s/thread,/thread/g

docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
scottmarlow: it may be out of scope, but will see what you think of also making same change nearby in jboss.as.jpa.deferdetach, to be consistent.

docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
scottmarlow: will update jboss.as.jpa.deferdetach also for this change.

scottmarlow:
retest this please

docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
bstansberry: +1

PR #12719, opened 2019-10-16
[WFLY-12686] Updated Microprofile Metrics to 2.2

tobiasstadler:
Issue: https://issues.jboss.org/browse/WFLY-12686 and https://issues.jboss.org/browse/WFLY-12735

PR #12721, opened 2019-10-16
[WFLY-12687] Updated Smallrye Config to 1.3.11

tobiasstadler:
Issue: https://issues.jboss.org/browse/WFLY-12687

### DIFF
--- a/testsuite/integration/microprofile-tck/config/tck-suite.xml
+++ b/testsuite/integration/microprofile-tck/config/tck-suite.xml
@@ -26,6 +26,19 @@
                     <exclude name=".*" />
                 </methods>
             </class>
+            <!-- These test are also exclueded in the SmallRye TCK -->
+            <class name="org.eclipse.microprofile.config.tck.ConfigProviderTest">
+                <methods>
+                    <exclude name="testEnvironmentConfigSource"/>
+                </methods>
+            </class>
+            <class name="org.eclipse.microprofile.config.tck.EmptyValuesTest">
+                <methods>
+                    <exclude name="testEmptyStringPropertyFromConfigFile"/>
+                    <exclude name="testEmptyStringProgrammaticLookup"/>
+                    <exclude name="testEmptyStringValues"/>
+                </methods>
+            </class>
         </classes>
     </test>
 
--- a/pom.xml
+++ b/pom.xml
@@ -274,7 +274,7 @@
         <version.io.opentracing.servlet>0.2.3</version.io.opentracing.servlet>
         <version.io.reactivex.rxjava>2.2.2</version.io.reactivex.rxjava>
         <version.io.rest-assured>3.0.0</version.io.rest-assured>
-        <version.io.smallrye.smallrye-config>1.3.6</version.io.smallrye.smallrye-config>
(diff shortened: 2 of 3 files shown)


rhusar (changes requested):
Tests need to be adapted as well.
```
ConfigProviderTest.testEnvironmentConfigSource
EmptyValuesTest.testEmptyStringProgrammaticLookup
EmptyValuesTest.testEmptyStringPropertyFromConfigFile
```

rhusar:
Also a component upgrade Jira is needed and needs to prefix the commit message.

jmesnil:
@tobiasstadler thanks for the PR.

Your PR looks fine but I'll put it on hold for the time being.
After smallrye-config 1.3.6 was released, changes in the components were made for its use in Quarkus.
They resulted in challenges to the TCK about the `EmptyValuesTest` tests that fail in our WildFly Jobs.

We are in the process of releasing smallrye-config 1.3.10 to run the MP Config TCK with Quarkus. At this point, we will be able to upgrade smallrye-config in WildFly with the same TCK challenges that Quarkus has.

jmesnil (changes requested):
smallrye-config 1.3.11 has been released and this is the version we should upgrade to: https://mvnrepository.com/artifact/io.smallrye/smallrye-config/1.3.11

It challenges some MicroProfile Config TCK tests (as defined in https://github.com/smallrye/smallrye-config/blob/master/testsuite/tck/tck-suite.xml).

We should update https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/config/tck-suite.xml to add the same exclusions.

@tobiasstadler do you want to take care of ^^? thanks

tobiasstadler:
@jmesnil I updated the PR

rhusar (changes requested):
Tests are failing:

org.eclipse.microprofile.metrics.tck
(1)

GlobalTagsTest.customConfigSource
 

org.wildfly.test.integration.microprofile.config.smallrye.app
(2)

MicroProfileConfigTestCase.testGetWithArraySetListDefaultProperties
 

MicroProfileConfigTestCase.testGetWithArraySetListOverriddenProperties
 

TestSuite: org.eclipse.microprofile.config.tck
(1)

ConfigProviderTest.testEnvironmentConfigSource

tobiasstadler:
@rhusar I am not sure what to with GlobalTagsTest.customConfigSource. As per the api spec `ConfigProviderResolver#registerConfig` should throw an IllegalStateException if there is already a Config registered within the Application (with is the case here). I think the test needs to call `releaseConfig` before  `registerConfig`. What do you think?

testsuite/integration/microprofile-tck/config/tck-suite.xml
```diff
@@ -26,6 +26,19 @@
                     <exclude name=".*" />
                 </methods>
             </class>
+            <!-- These test are also exclueded in the SmallRye TCK -->
```
rhusar: s/exclueded/excluded

rhusar (changes requested):

rhusar:
@tobiasstadler Its possible, I would need to have a better look...

jmesnil:
@tobiasstadler same comment than for #12719 

Thanks for your help!

PR #12727, opened 2019-10-18
WFLY-12689 WebNonTxEmCloserAction#dependencies() should return an emp…

msfm:
…ty set instead of null

When a web application contains both MBean (jboss-service.xml) and
persistence-unit (persistence.xml), this causes NullPointerException and
the WAR deployment fails.

JIRA: https://issues.jboss.org/browse/WFLY-12689


bstansberry (approved):
Thanks!

I filed https://github.com/wildfly/wildfly-core/pull/3992 to make this part of the SetupAction contract. This one is the only one in WildFly that returned null.

scottmarlow:
UndertowDeploymentProcessor does check for null, so if the contract changes to not expect null, UndertowDeploymentProcessor can probably also change.

PR #12740, opened 2019-10-24
[WFLY-12707] Explicitly skip the enforcer and checkstyle plugins wher…

jamezp:
…e required instead of attempting to override the property.

https://issues.jboss.org/browse/WFLY-12707

Note we skip the enforcer plugin in all integration tests which seems a bit weird IMO. I started cleaning it up, but it's quite the web of dependencies that need to be cleaned up. If we'd prefer that approach I can take the time to fix it, however I didn't want to spend too much time and went with the simple approach first.


jamezp:
This is a run where the `-Denforcer.skip=false` and `-Dcheckstyle.skip=false` was passed on the command line. https://ci.wildfly.org/viewLog.html?buildId=172635&buildTypeId=WF_MasterLinux

bstansberry (approved):

PR #12749, opened 2019-10-28
[WFLY-12635] Http requests failed with ISPN000299 after redirect and session invalidation

rmartinc:
PR for https://issues.jboss.org/browse/WFLY-12635

@pferraro I'm submitting the PR with the changes we have been commented these days. Three comments:

1. I have just added the ISA to the `invalidate` method (if you want to include this in other methods just let me know, but I think this is what we wanted).
2. The mocked tests have been updated to check the new two changes.
3. I have removed the full app test because if an ISA is thrown a full test makes no sense IMO. The same, if you think it is needed just let me know.

Regards.

### DIFF
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/logging/UndertowClusteringLogger.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/logging/UndertowClusteringLogger.java
@@ -58,4 +58,7 @@ public interface UndertowClusteringLogger extends BasicLogger {
     @Message(id = 8, value = "No distributable single sign-on management provider found for %s; using legacy provider based on static configuration")
     @LogMessage(level = Level.WARN)
     void legacySingleSignOnProviderInUse(String name);
+
+    @Message(id = 9, value = "Invalidation attempted for session %s after the response was committed (e.g. after HttpServletResponse.sendRedirect or sendError)")
+    IllegalStateException batchIsAlreadyClosed(String sessionId);
 }
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
@@ -89,6 +89,12 @@ public class DistributableSession implements io.undertow.server.session.Session 
                     // Don't propagate exceptions at the stage, since response was already committed
                     UndertowClusteringLogger.ROOT_LOGGER.warn(e.getLocalizedMessage(), e);
                 }
+            } else if (this.batch.getState() != Batch.State.CLOSED) {
+                Batcher<Batch> batcher = this.manager.getSessionManager().getBatcher();
+                try (BatchContext context = batcher.resumeBatch(this.batch)) {
+                    // Ensure batch is closed.
+                    this.batch.close();
+                }
             }
         } finally {
             this.closeTask.accept(exchange);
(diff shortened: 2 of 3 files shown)


clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/logging/UndertowClusteringLogger.java
```diff
@@ -58,4 +58,8 @@
     @Message(id = 8, value = "No distributable single sign-on management provider found for %s; using legacy provider based on static configuration")
     @LogMessage(level = Level.WARN)
     void legacySingleSignOnProviderInUse(String name);
+
+    @Message(id = 9, value = "Session %s is being modified when the response is closed (e.g. after HttpServletResponse sendRedirect or sendError). " +
```
pferraro: This should probably read: "Session %s invalidation attempted after the response was committed (e.g. after HttpServletResponse.sendRedirect(...) or sendError(...)."

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
```diff
@@ -89,6 +89,9 @@ public void requestDone(HttpServerExchange exchange) {
                     // Don't propagate exceptions at the stage, since response was already committed
                     UndertowClusteringLogger.ROOT_LOGGER.warn(e.getLocalizedMessage(), e);
                 }
+            } else if (this.batch.getState() != Batch.State.CLOSED) {
```
pferraro: When is this ever the case?

pferraro (changes requested):

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/logging/UndertowClusteringLogger.java
rmartinc: Done!

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
rmartinc: This was the initial problem in this issue. The first request invalidated the session and the second one detected it as invalid and threw an exception, see [this lines](https://github.com/wildfly/wildfly/blob/master/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java#L254). The lock here was not released and the session remained locked. Remember I first added the release in the same `validate` method (before the throw) and you recommended me to move it to the `requestDone` method.

pferraro: Oh right.  Now I remember. :)
However, you'll want to make sure to resume the batch before closing.
e.g.
```java
try (BatchContext context = batcher.resumeBatch(this.batch)) {
    // Ensure batch is closed.
    this.batch.close();
}
```

pferraro (changes requested):

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
rmartinc: Right! Like in the previous if. It's done now. Just submitted the new version.

pferraro (approved):
Nice work!

PR #12709, opened 2019-10-14
WFLY-12662 BouncyCastle update from 1.60 to 1.64

rsvoboda:
BouncyCastle update from 1.60 to 1.64
https://issues.jboss.org/browse/WFLY-12662

Please make sure your PR meets the following requirements:
- [x] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue(s)
- [x] Pull Request contains description of the issue(s)
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Attached commits represent units of work and are properly formatted

PR #12750, opened 2019-10-28
WFLY-12720 BouncyCastle update from 1.60 to 1.64

bstansberry:
Issue: https://issues.jboss.org/browse/WFLY-12720
Upstream: #12709


bstansberry:
I don't see this as necessary for 18.0.1 as the CVE fix in 1.64 is for a CVE introduced in 1.63, and we have no other WF problem reports with 1.60.

PR #12757, opened 2019-10-30
[WFLY-12730]:Move SimpleGroup to a public outter class

jimma:
- https://issues.jboss.org/browse/WFLY-12730
- https://issues.apache.org/jira/browse/CXF-7815 changed to use reflection to access Group's member for java.security.acl APIs removal in JDK 13.
Now the SimpleGroup which created to store the roles from Elytron securityIdentity is a private static class: https://github.com/wildfly/wildfly/blob/18.0.0.Final/webservices/server-integration/src/main/java/org/jboss/as/webservices/util/SubjectUtil.java#L185
It should be change to put it in a public class for CXF or other integration layer to easily access.


jimma:
retest this please.

asoldano (approved):

jamezp (approved):

PR #12760, opened 2019-10-31
[WFLY-12733] EJB Timers: Forcing refresh timers in a database-data-store in a clustered environment

rhuan080:
The EJB Timer has a way to persist timers in a database using datasources. It is described [here|https://docs.jboss.org/author/display/WFLY10/EJB3+Clustered+Database+Timers]. The database-data-store has an attribute called refresh-interval that define a time in milliseconds to refresh the timers reading the timers from database. Look this sample of configuration.

```
                    <data-stores>
                        <file-data-store name="default-file-store" path="timer-service-data" relative-to="jboss.server.data.dir"/>
                        <database-data-store name="clustered-store" datasource-jndi-name="java:/jdbc/MyDataSource" partition="timer" refresh-interval="60000" allow-execution="true"/>
                    </data-stores>
```

This pull request add a method in TimerServiceImpl to force a refresh programmatically.


bstansberry:
this is ok to test

darranl:
I see this one is now pre-checked so I have kicked off CI again.

FYI it looks like the set of CI jobs has changed so an old failure may remain in the list even if current CI passes Ok,

bstansberry:
I believe this is superseded by #13183

PR #12769, opened 2019-11-04
Fix MicroProfile TCKs test maven module names

rhusar:


rhusar:
@bstansberry As I said on the call, just changing the human readable names :-)

bstansberry (approved):

PR #12770, opened 2019-11-04
[WFLY-12750] Register a capability for the EE security subystem.

darranl:
https://issues.jboss.org/browse/WFLY-12750
https://issues.jboss.org/browse/WFLY-12755

### DIFF
--- a/ee-security/src/main/java/org/wildfly/extension/eesecurity/EESecuritySubsystemDefinition.java
+++ b/ee-security/src/main/java/org/wildfly/extension/eesecurity/EESecuritySubsystemDefinition.java
@@ -26,17 +26,28 @@ import java.util.Collections;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.PersistentResourceDefinition;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
+import org.jboss.as.controller.capability.RuntimeCapability;
 
 /**
  * @author Stuart Douglas
  */
 public class EESecuritySubsystemDefinition extends PersistentResourceDefinition {
 
+    static final String EE_SECURITY_CAPABILITY_NAME = "org.wildfly.ee.security";
+    static final String WELD_CAPABILITY_NAME = "org.wildfly.weld";
+
+    static final RuntimeCapability<Void> EE_SECURITY_CAPABILITY =
+            RuntimeCapability.Builder.of(EE_SECURITY_CAPABILITY_NAME)
+                    .setServiceType(Void.class)
+                    .addRequirements(WELD_CAPABILITY_NAME)
+                    .build();
+
     public static final EESecuritySubsystemDefinition INSTANCE = new EESecuritySubsystemDefinition();
 
     private EESecuritySubsystemDefinition() {
         super(new Parameters(EESecurityExtension.SUBSYSTEM_PATH, EESecurityExtension.getResolver())
                 .setAddHandler(EESecuritySubsystemAdd.INSTANCE)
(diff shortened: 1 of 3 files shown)


ee-security/src/main/java/org/wildfly/extension/eesecurity/EESecuritySubsystemDefinition.java
```diff
@@ -26,17 +26,28 @@
...
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
+import org.jboss.as.controller.capability.RuntimeCapability;
 
 /**
  * @author Stuart Douglas
  */
 public class EESecuritySubsystemDefinition extends PersistentResourceDefinition {
 
+    static final String EE_SECURITY_CAPABILITY_NAME = "org.wildfly.ee.security";
+    static final String WELD_CAPABILITY_NAME = "org.wildfly.weld";
```
yersan: Besides of the CI errors related to the missing Weld feature in ee security layer, if Weld cap is required to the EE security, I am wondering if it could better to import the capability name from https://github.com/wildfly/wildfly/blob/master/weld/common/src/main/java/org/jboss/as/weld/Capabilities.java#L30 instead of using a string. An advantage is you can easily find what code is using this capability finding the variable usages, the "disadvantage" is you need to include the weld common maven dep into the security subsystem. That could be in provider scope, it is not needed at runtime and you don't need to include it as module dependency. @darranl Do you see something wrong using that approach?

darranl: IMO having inter module dependencies just to reference a capability name is a bad practice - the whole point of specifying a capability in the management model in this way is we don't know if it exists or not.

We did have discussions at the outset of the introduction of capabilities and agreed to avoid trying to come up with common modules to hold those definitions so the advantage described is not something we are aiming for and the disadvantage described is something we have already agreed to avoid.

Also it is incorrect to say the dependency would be scope PROVIDED only - the subsystem needs to be able to call this code as it is installed meaning it will require a dependency adding to it's module.

Overall unless we are truly using some exposed APIs we need to minimising the inter subsystem dependencies.

Related to this I am also not sure if the WELD capability name is correct, EE Security does not require WELD it required CDI.

However if the present layers mean EE Security can be provisioned without CDI that is a bug as EE Security and it's implementation are entirely CDI based and can not operate at all without it.

yersan: Ok, thanks for clear it up, good to know that those discussions were already taken, then code like this Capabilities.java#L30 is not the recommended way, I did not know was the wrong way :(

> Also it is incorrect to say the dependency would be scope PROVIDED only - the subsystem needs to be able to call this code as it is installed meaning it will require a dependency adding to it's module.

What I was trying to point out is we are depending on a constant that is resolved at compile-time, not at runtime, so we do not need a JBoss Module dependency, however, the maven dependency is required to build the project. We split the server using Galleon which takes pieces already compiled, so, IMO it really does not affect, however, I agree, it adds a requirement that can be avoided.

darranl: Ok I see what you mean about the dependency, however I think that approach could still be risky as you end up with a dependency on a module with plenty of other code that could not be in-lined.

Overall we have no need to use common definitions of capabilities.

In the meantime I am going to see what we need to do regarding this broken layer: -
    https://issues.jboss.org/browse/WFLY-12755

bstansberry (approved):

PR #12776, opened 2019-11-05
WFLY-12759 Extract microprofile-config galleon layer from observability

pferraro:
https://issues.jboss.org/browse/WFLY-12759

Microprofile specs like OpenAPI don't require metrics nor health, suggesting the microprofile layers should be more granular.

### DIFF
--- a/galleon-pack/src/main/resources/layers/standalone/microprofile-config/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/microprofile-config/layer-spec.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="microprofile-config">
+    <dependencies>
+        <layer name="management"/>
+        <layer name="cdi"/>
+    </dependencies>
+    <feature spec="subsystem.microprofile-config-smallrye"/>
+</layer-spec>
--- a/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
@@ -1,11 +1,9 @@
 <?xml version="1.0" ?>
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="observability">
     <dependencies>
-        <layer name="management"/>
-        <layer name="cdi"/>
+        <layer name="microprofile-config"/>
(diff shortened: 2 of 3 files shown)


testsuite/integration/microprofile-tck/config/pom.xml
```diff
@@ -173,7 +173,7 @@
                                             <name>standalone.xml</name>
                                             <layers>
                                                 <layer>datasources-web-server</layer>
-                                                <layer>observability</layer>
+                                                <layer>microprofile-config</layer>
```
bstansberry: I'm not sure I want to do this as running against this combo is part of the test plan for our recommended combos in our images.

pferraro: The intention was to use a minimal profile sufficient to pass the MicroProfile Config TCK.  It also provides an opportunity to validate the microprofile-config layer.
https://github.com/wildfly/wildfly/pull/12774 will use a profile more in line with the test plan you described.
If you still disagree, I will revert this change.

bstansberry: I'd prefer not to change this right now.  Right now #12774 is an empty module.

I don't particularly disagree with the change; I just don't feel like reopening a long discussion right now when I know there will be lots of other opportunities to do it. :)

bstansberry (changes requested):
This will need an entry in the testsuite/layers/pom.xml. All layers should have an entry there as that is the basic boot validation for the layer.

It also needs an update to https://github.com/wildfly/wildfly/blob/master/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc

testsuite/integration/microprofile-tck/config/pom.xml
pferraro: No problem.

pferraro:
@bstansberry Fixed.

bstansberry (approved):

bstansberry:
@jmesnil OK with you?

rhusar (approved):
Thanks for this. Looks good.

rhusar:
Actually, would it make any sense to document this in {{docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc}}? Seems unnecessary as its basically not useful by itself.

bstansberry:
@rhusar MicroProfile Config is usable directly by applications, so any layer at all could possibly be decorated with this one.  

It seems simpler to document them all too.

PR #12783, opened 2019-11-06
[WFLY-12766] Upgrade WildFly Common to 1.5.2.Final

yersan:
Jira issue: https://issues.jboss.org/browse/WFLY-12766


darranl (dismissed):
We should actually look to remove this dependency as it is already defined in WildFly Core.

bstansberry (approved):

PR #12788, opened 2019-11-07
[WFLY-12135] Error deploying web service without http-listener (only https-listener)

iweiss:
Issue: https://issues.jboss.org/browse/WFLY-12135

DON'T MERGE. This depends on https://github.com/jbossws/jbossws-common/pull/19.

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/WebServerService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/WebServerService.java
@@ -50,11 +50,11 @@ class WebServerService implements CommonWebServer, Service<WebServerService> {
         Map<String, UndertowListener> listeners = getListenerMap();
         UndertowListener listener = null;
         for (String p : listeners.keySet()) {
-            if (protocol.toLowerCase().contains(p)) {
+            if (protocol.toLowerCase().contains(p) || listeners.size() == 1) {
                 listener = listeners.get(p);
             }
         }
-        if (listener != null && listener.getProtocol() == HttpListenerService.PROTOCOL && secure) {
+        if (listener != null && listener.getProtocol().equals(HttpListenerService.PROTOCOL) && secure) {
             if (listeners.containsKey(HttpsListenerService.PROTOCOL)) {
                 listener = listeners.get(HttpsListenerService.PROTOCOL);
             } else {
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
@@ -21,6 +21,8 @@
  */
 package org.jboss.as.webservices.deployers;
 
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
+import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
@@ -28,10 +30,14 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.webservices.logging.WSLogger;
 import org.jboss.as.webservices.util.ASHelper;
 import org.jboss.as.webservices.util.WSAttachmentKeys;
+import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.wsf.spi.classloading.ClassLoaderProvider;
 import org.jboss.wsf.spi.deployment.Deployment;
 import org.jboss.wsf.spi.deployment.DeploymentAspect;
+import org.wildfly.extension.undertow.Server;
+import org.wildfly.extension.undertow.UndertowService;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
@@ -68,6 +74,8 @@ public final class AspectDeploymentProcessor implements DeploymentUnitProcessor 
             WSLogger.ROOT_LOGGER.tracef("%s start: %s", aspect, unit.getName());
             ClassLoader origClassLoader = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
             try {
+                dep.setProperty("isHttpsOnly", isHttpsOnly(phaseContext, unit));
+
                 WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(aspect.getLoader());
                 dep.addAttachment(ServiceTarget.class, phaseContext.getServiceTarget());
                 aspect.start(dep);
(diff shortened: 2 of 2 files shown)


bstansberry:
The org.wildfly.undertow capability is private:

https://github.com/wildfly/wildfly-capabilities/blob/master/org/wildfly/undertow/capability.adoc

That's because UndertowService is a really poor contract for use outside the subsystem; it exposes all sorts of stuff, including start/stop, that should not be visible external to the extension module.

The WS subsystem already has code like this, and it has no choice because it uses what the undertow subsystem has historically exposed. So doing more of the same isn't a problem that should prevent merging a fix like this. But @fl4via it would be good to get a cleaner contract sometime during the next couple WF releases.

undertow/src/main/java/org/wildfly/extension/undertow/WebServerService.java
```diff
@@ -50,7 +50,7 @@ public int getPort(final String protocol, final boolean secure) {
         Map<String, UndertowListener> listeners = getListenerMap();
         UndertowListener listener = null;
         for (String p : listeners.keySet()) {
-            if (protocol.toLowerCase().contains(p)) {
+            if (protocol.toLowerCase().contains(p) || listeners.size() == 1) {
```
bstansberry: The CommonWebServer javadoc needs to specify the contract here, as it's not obvious what the expected behavior is.

This '|| listeners.size() == 1' doesn't seem precise enough. What is there is only an AJP listener?  What if there's only HTTPS and 'secure' is false?

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
```diff
@@ -68,6 +78,8 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
             WSLogger.ROOT_LOGGER.tracef("%s start: %s", aspect, unit.getName());
             ClassLoader origClassLoader = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
             try {
+                dep.addAttachment(Boolean.class, isHttpsOnly(phaseContext, unit));
```
bstansberry: @jimma Is 'Boolean' an acceptable type for this? 

Well, you'll decide whether the JBoss WS fix gets merged and whatever is used there is what will be needed here. :)

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
```diff
@@ -112,4 +124,33 @@ private static boolean isWebServiceDeployment(final DeploymentUnit unit) {
...
+    private boolean isHttpsOnly(final DeploymentPhaseContext phaseContext, final DeploymentUnit unit) {
+        CapabilityServiceSupport serviceSupport = unit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+        ServiceName undertowServiceName = serviceSupport.getCapabilityServiceName("org.wildfly.undertow");
+        ServiceController<?> undertowServiceController = phaseContext.getServiceRegistry().getService(undertowServiceName);
+        UndertowService undertowService = (UndertowService) undertowServiceController.getValue();
+
+        Server defaultServer = null;
+
+        for (Server server : undertowService.getServers()) {
+            if (server.getName().equals("default-server")) {
```
bstansberry: We shouldn't use a magic value. UndertowService exposes a getDefaultServer() which seems like it should provide the correct value.

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
```diff
@@ -112,4 +124,33 @@ private static boolean isWebServiceDeployment(final DeploymentUnit unit) {
...
+                defaultServer = server;
+            }
+        }
+
+        if (defaultServer == null) return false;
+
+        List<String> listeners = new ArrayList<>();
+        for (UndertowListener listener : defaultServer.getListeners()) {
+            listeners.add(listener.getProtocol());
+        }
```
bstansberry: Copying the protocols out into a list isn't needed to check if there's just 1 listener whose protocol is 'https'.

bstansberry (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/WebServerService.java
iweiss: `|| listeneres.size() == 1` will fix the `secure == false` when there's only a HTTPS listener issue. Right now it fails.

Is there's only an AJP listener than it would throw an exception, as expected since it's not a HTTP-family listener.

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
iweiss: If not, this could be replace with a Map<Integer, String> to map port to protocol.

darranl:
@iweiss I see a forced push, does that address the comments Brian had?

A problem with the GitHub UI is stale reviews are not cleared so an issue sits in the queue looking like no changes have been made.

darranl:
@iweiss This one has now been stalled since last year but we have a big backlog of PRs in the queue making it hard to see the ones that should be given attention - can we close this one for now and pick it up again once you are ready to address the discussion points so far?

iweiss:
@darranl I think it's fair, but I would like to do a last push on this if you don't mind.

CC @jimma @bstansberry

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/AspectDeploymentProcessor.java
```diff
@@ -68,6 +74,8 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
             WSLogger.ROOT_LOGGER.tracef("%s start: %s", aspect, unit.getName());
             ClassLoader origClassLoader = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
             try {
+                dep.setProperty("isHttpsOnly", isHttpsOnly(phaseContext, unit));
+
```
iweiss, line 78: @bstansberry @darranl As suggested by @jimma, I replaced the boolean attachment with a property. Looks cleaner and better.

iweiss:
https://github.com/jbossws/jbossws-common/pull/19 has been merge.

iweiss:
retest please

iweiss:
jbossws/jbossws-common#19 is now part of WildFly, https://github.com/wildfly/wildfly/pull/13552, so this PR is now ready to be merged.

darranl:
@jimma / @fl4via Are we able to give this one a review so we can see if it can be merged?

darranl:
@iweiss I have just kicked off some of the failed CI jobs again but TBH the failures do look related.

darranl:
FYI the "Linux - Elytron - JDK 8" job is an old job that can be ignored.

fjuma:
@iweiss Is this PR still relevant? There haven't been any updates on it in a long time.

iweiss:
@fjuma no idea. I would say no, since I haven't heard about it in a while, but I'm not really sure.

iweiss:
Agreed

PR #12797, opened 2019-11-13
WFLY-12777 Stateful bean with @StatefulTimeout value -1 should not ti…

chengfang:
…meout or removed

Fixed `org.jboss.as.ejb3.cache.simple.SimpleCache` class to properly handle stateful-timeout value -1 (never timeout).

JIRA: https://issues.jboss.org/browse/WFLY-12777

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/cache/distributable/DistributableCacheFactoryBuilderServiceConfigurator.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/cache/distributable/DistributableCacheFactoryBuilderServiceConfigurator.java
@@ -134,8 +134,13 @@ public class DistributableCacheFactoryBuilderServiceConfigurator<K, V extends Id
             @Override
             public Duration getTimeout() {
                 StatefulTimeoutInfo info = description.getStatefulTimeout();
+
+                // A value of -1 means the bean will never be removed due to timeout
+                if (info == null || info.getValue() < 0) {
+                    return null;
+                }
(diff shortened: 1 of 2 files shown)


scottmarlow (approved):
Looks good to me, the -1 setting is also mentioned in ejb-jar_3.1.xsd:
>         The stateful-timeoutType represents the amount of time
>         a stateful session bean can be idle(not receive any client
>         invocations) before it is eligible for removal by the container.
> 
>         A timeout value of 0 means the bean is immediately eligible for removal.
> 
>         A timeout value of -1 means the bean will never be removed due to timeout.
>

scottmarlow:
@pferraro could you please review this change also.

pferraro (changes requested):
DistributableCache needs to be fixed as well:
https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/cache/distributable/DistributableCacheFactoryBuilderServiceConfigurator.java#L138

pferraro:
Frankly, this particular detail in the spec is a bit illogical...
@StatefulTimeout(value=-24, unit=TimeUnit.HOURS), is invalid, but
@StatefulTimeout(value=-1, unit=TimeUnit.DAYS), while logically equivalent to the above, is valid and means the bean never times out.

iMO, we ought to clean this up in future versions of the EJB spec.  Really, *any* negative value should to be interpreted in the same way.

chengfang:
@pferraro thanks.  I'll update DistributableCache, and loosen the condition to allow for all negative numbers.

chengfang:
I've updated the PR based on the review comments.  Please review again.

pferraro (approved):

PR #12800, opened 2019-11-13
WFLY-12590 Integrate MicroProfile Fault Tolerance 2.0

rhusar:
Jira
https://issues.jboss.org/browse/WFLY-12590

Proposal 
https://github.com/wildfly/wildfly-proposals/pull/250

### DIFF
--- a/feature-pack/src/license/full-feature-pack-licenses.xml
+++ b/feature-pack/src/license/full-feature-pack-licenses.xml
@@ -221,6 +221,28 @@
       </licenses>
     </dependency>
     <dependency>
+      <groupId>com.netflix.archaius</groupId>
+      <artifactId>archaius-core</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>com.netflix.hystrix</groupId>
+      <artifactId>hystrix-core</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
       <groupId>com.squareup.okhttp3</groupId>
       <artifactId>okhttp</artifactId>
       <licenses>
@@ -396,6 +418,17 @@
     <dependency>
       <groupId>commons-collections</groupId>
       <artifactId>commons-collections</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>commons-configuration</groupId>
+      <artifactId>commons-configuration</artifactId>
       <licenses>
         <license>
           <name>Apache License 2.0</name>
@@ -592,6 +625,17 @@
       </licenses>
     </dependency>
     <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxjava</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
       <groupId>io.reactivex.rxjava2</groupId>
       <artifactId>rxjava</artifactId>
       <licenses>
@@ -627,6 +671,17 @@
     <dependency>
       <groupId>io.smallrye.config</groupId>
       <artifactId>smallrye-config-source-file-system</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye</groupId>
+      <artifactId>smallrye-fault-tolerance</artifactId>
       <licenses>
         <license>
           <name>Apache License 2.0</name>
@@ -1836,6 +1891,17 @@
       </licenses>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+      <artifactId>microprofile-fault-tolerance-api</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.microprofile.health</groupId>
       <artifactId>microprofile-health-api</artifactId>
       <licenses>
@@ -1988,6 +2054,20 @@
           <name>GNU General Public License v2.0 only, with Classpath exception</name>
           <url>https://fedoraproject.org/wiki/Licensing/GPL_Classpath_Exception</url>
           <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+      <licenses>
+        <license>
+          <name>Public Domain, per Creative Commons CC0</name>
+          <url>http://creativecommons.org/publicdomain/zero/1.0/</url>
+        </license>
+        <license>
+          <name>BSD-2-Clause</name>
+          <url>https://opensource.org/licenses/BSD-2-Clause</url>
         </license>
       </licenses>
     </dependency>
@@ -3010,6 +3090,28 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>wildfly-microprofile-config-smallrye</artifactId>
+      <licenses>
+        <license>
+          <name>GNU Lesser General Public License v2.1 or later</name>
+          <url>http://www.gnu.org/licenses/old-licenses/lgpl-2.1-standalone.html</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>wildfly-microprofile-fault-tolerance-smallrye-extension</artifactId>
+      <licenses>
+        <license>
+          <name>GNU Lesser General Public License v2.1 or later</name>
+          <url>http://www.gnu.org/licenses/old-licenses/lgpl-2.1-standalone.html</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>wildfly-microprofile-fault-tolerance-smallrye-cdi</artifactId>
       <licenses>
         <license>
           <name>GNU Lesser General Public License v2.1 or later</name>
--- a/feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2019, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="io.smallrye.fault-tolerance">
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${io.smallrye:smallrye-fault-tolerance}"/>
+    </resources>
+
+    <dependencies>
+        <module name="com.netflix.hystrix.core"/>
+        <module name="io.reactivex.rxjava"/>
+        <module name="io.smallrye.config"/>
+        <module name="io.smallrye.metrics"/>
+        <module name="javax.annotation.api"/>
+        <module name="javax.api"/>
+        <module name="javax.enterprise.api"/>
+        <module name="javax.enterprise.concurrent.api"/>
+        <module name="org.eclipse.microprofile.config.api"/>
+        <module name="org.eclipse.microprofile.fault-tolerance.api"/>
+        <module name="org.eclipse.microprofile.metrics.api"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
+        <module name="org.wildfly.microprofile.fault-tolerance-smallrye.cdi"/>
+    </dependencies>
+</module>
--- a/feature-pack/src/main/resources/modules/system/layers/base/org/hdrhistogram/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/org/hdrhistogram/main/module.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2019, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.hdrhistogram">
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.hdrhistogram:HdrHistogram}"/>
+    </resources>
+
+    <dependencies>
+    </dependencies>
+</module>
--- a/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main/module.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2019, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.wildfly.extension.microprofile.fault-tolerance-smallrye">
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.wildfly:wildfly-microprofile-fault-tolerance-smallrye-extension}"/>
+    </resources>
+
+    <dependencies>
+        <module name="com.netflix.archaius.core"/>
+        <module name="com.netflix.hystrix.core"/>
+        <module name="io.smallrye.fault-tolerance"/>
+        <module name="javax.annotation.api"/>
+        <module name="javax.enterprise.api"/>
+        <module name="org.apache.commons.configuration"/>
+        <module name="org.eclipse.microprofile.config.api"/>
+        <module name="org.jboss.as.clustering.common"/>
+        <module name="org.jboss.as.controller"/>
+        <module name="org.jboss.as.server"/>
+        <module name="org.jboss.as.weld.common"/>
+        <module name="org.jboss.jandex"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.modules"/>
+        <module name="org.jboss.staxmapper"/>
+        <module name="org.jboss.weld.api"/>
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
+        <module name="org.wildfly.microprofile.fault-tolerance-smallrye.cdi"/>
+    </dependencies>
+</module>
--- a/microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
+++ b/microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2018 Red Hat, Inc, and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.extension.microprofile.faulttolerance.cdi;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+
+import io.smallrye.faulttolerance.CommandListener;
+import io.smallrye.faulttolerance.config.FaultToleranceOperation;
+import org.jboss.weld.context.RequestContext;
+import org.jboss.weld.context.unbound.Unbound;
+import org.jboss.weld.manager.api.WeldManager;
+
+/**
+ * Activates CDI request context when a Hystrix command that wraps a Fault Tolerance operation is executed.
+ *
+ * @author Martin Kouba
+ * @author Radoslav Husar
+ */
+@ApplicationScoped
+public class RequestContextCommandListener implements CommandListener {
+
+    private final WeldManager weldManager;
+
+    private final RequestContext requestContext;
+
+    private static final ThreadLocal<Boolean> isActivator = new ThreadLocal<>();
+
+    @Inject
+    public RequestContextCommandListener(@Unbound RequestContext requestContext, WeldManager weldManager) {
+        this.requestContext = requestContext;
+        this.weldManager = weldManager;
+    }
+
+    @Override
+    public void beforeExecution(FaultToleranceOperation operation) {
+        // Note that Hystrix commands are executed on a dedicated thread where a CDI request context should not be active
+        // TODO: maybe we should activate the context for any FT operation?
+        if (operation.isAsync() && !weldManager.isContextActive(RequestScoped.class)) {
+            requestContext.activate();
+            isActivator.set(true);
+        }
+    }
+
+    @Override
+    public void afterExecution(FaultToleranceOperation operation) {
+        if (Boolean.TRUE.equals(isActivator.get())) {
+            try {
+                requestContext.invalidate();
+                requestContext.deactivate();
+            } finally {
+                isActivator.remove();
+            }
+        }
+    }
+
+}
--- a/microprofile/fault-tolerance-smallrye/extension/pom.xml
+++ b/microprofile/fault-tolerance-smallrye/extension/pom.xml
@@ -1,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2019, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>19.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>wildfly-microprofile-fault-tolerance-smallrye-extension</artifactId>
+    <name>WildFly: MicroProfile Fault Tolerance - Extension</name>
+
+    <dependencies>
+        <!-- MicroProfile Fault Tolerance and SmallRye dependencies -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-microprofile-fault-tolerance-smallrye-cdi</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+            <artifactId>microprofile-fault-tolerance-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>smallrye-fault-tolerance</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-weld-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.inject</groupId>
+            <artifactId>jakarta.inject-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.netflix.archaius</groupId>
+            <artifactId>archaius-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>commons-configuration</groupId>
+            <artifactId>commons-configuration</artifactId>
+        </dependency>
+        <!-- Subsystem dependencies -->
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-clustering-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-controller</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-processor</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-subsystem-test</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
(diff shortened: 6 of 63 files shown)


microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+                    configInstance.setProperty(key, value);
+                }
+            }
+        } else {
+            MicroProfileFaultToleranceLogger.ROOT_LOGGER.hystrixAlreadyConfigured(deploymentUnit.getName());
+        }
+
+
+        // Weld Extension
+        CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
```
yersan: I think the new microprofile-fault-tolerance subsystem could need to create a capability with a hard requirement on Weld capability. 

If CDI is not a hard requirement to make microprofile-fault-tolerance subsystem work properly, then I think at a minimum, this piece of code getting the Weld Capability should be moved to be processed before L:55. At first glance, it sounds like a possible bug if the error on L:85 is thrown after processing L:55 when we have multiple deployments being deployed at the same time.

microprofile/fault-tolerance-smallrye/extension/src/test/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceModelTest.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.microprofile.faulttolerance;
+
+import static org.junit.runners.Parameterized.*;
```
yersan: Wildcard static imports are not allowed.

yersan (changes requested):

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceLogger.java
```diff
@@ -0,0 +1,54 @@
...
+@MessageLogger(projectCode = "WFLYMPFTEXT", length = 4)
+public interface MicroProfileFaultToleranceLogger extends BasicLogger {
+
+    MicroProfileFaultToleranceLogger ROOT_LOGGER = Logger.getMessageLogger(MicroProfileFaultToleranceLogger.class, MicroProfileFaultToleranceLogger.class.getPackage().getName());
+
+    @LogMessage(level = INFO)
+    @Message(id = 1, value = "Activating MicroProfile Fault Tolerance subsystem.")
+    void activatingSubsystem();
+
+    @Message(id = 2, value = "Deployment %s requires use of %s capability but it is not currently registered.")
```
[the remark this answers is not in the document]
rhusar: Agreed – changed now.

microprofile/fault-tolerance-smallrye/extension/src/test/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceModelTest.java
rhusar: They actually are allowed  in the testsuite - if they weren't the CI runs would fail. We can enable this for testsuite as well if we like...

rhusar: I changed it anyway.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceServiceHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public void removeServices(OperationContext context, ModelNode model) {
+    }
+
+    @Override
+    public void accept(DeploymentProcessorTarget deploymentProcessorTarget) {
+        deploymentProcessorTarget.addDeploymentProcessor(MicroProfileFaultToleranceExtension.SUBSYSTEM_NAME, Phase.DEPENDENCIES, Phase.DEPENDENCIES_MICROPROFILE_OPENTRACING + 1, new MicroProfileFaultToleranceDependenciesProcessor());
+        deploymentProcessorTarget.addDeploymentProcessor(MicroProfileFaultToleranceExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.POST_MODULE_MICROPROFILE_OPENTRACING + 1, new MicroProfileFaultToleranceDeploymentProcessor());
```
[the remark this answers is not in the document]
rhusar: Ooops, thanks, this was just not updated after the core upgrade with the required change, fixed now.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
rhusar: The weld code is similar to MP Config IIRC, if its wrong, its wrong across the board. Also, switched the code blocks - you are right.

microprofile/fault-tolerance-smallrye/extension/src/test/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceModelTest.java
yersan: Ouuh, good to know it, thanks!

microprofile/fault-tolerance-smallrye/extension/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+            <artifactId>wildfly-subsystem-test</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
```
yersan: It looks like these dependencies are not needed. I would say they are harmless there, but if they are unneeded  ....
microprofile-fault-tolerance-api
smallrye-fault-tolerance
weld-spi
weld-api
weld-core-impl

yersan (changes requested):

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main/module.xml
```diff
@@ -0,0 +1,55 @@
...
+        <module name="org.jboss.as.weld.common"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.modules"/>
+        <module name="org.jboss.msc"/>
+        <module name="org.jboss.staxmapper"/>
+        <module name="org.jboss.weld.api"/>
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
+        <module name="org.wildfly.microprofile.fault-tolerance-smallrye.cdi"/>
+    </dependencies>
```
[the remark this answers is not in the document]
kabir: The weld ones are needed to have proper proxy generation. Otherwise you get a warning message that they will be created in the deployment classloader which is problematic if the proxied class has package protecte methods and things like that.

I have added similar eclipse and smallrye dependencies for my MP specs and they are essentially needed to have those on the classpath so it scans for CDI beans to wire up all the MP magic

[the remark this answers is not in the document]
yersan: Cool!, good to know it. I see some of them are being already exported by `io.smallrye.fault-tolerance`, I'm still with doubts about the intention if they are exported why add them here, maybe the export is unnecessary. My point is to not add more than needed since it helps Galleon in reducing the size when we have custom provisioning. 

So, then we can reduce the list to only `org.jboss.msc` and `javax.annotation.api`

microprofile/fault-tolerance-smallrye/extension/pom.xml
rhusar: This is in order to be able to load exception classes to help debugging. Its much more descriptive than CDNF.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceServiceHandler.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public void removeServices(OperationContext context, ModelNode model) {
+    }
+
+    @Override
+    public void accept(DeploymentProcessorTarget deploymentProcessorTarget) {
+        deploymentProcessorTarget.addDeploymentProcessor(MicroProfileFaultToleranceExtension.SUBSYSTEM_NAME, Phase.DEPENDENCIES, Phase.DEPENDENCIES_MICROPROFILE_FAULT_TOLERANCE, new MicroProfileFaultToleranceDependenciesProcessor());
+        deploymentProcessorTarget.addDeploymentProcessor(MicroProfileFaultToleranceExtension.SUBSYSTEM_NAME, Phase.POST_MODULE, Phase.DEPENDENCIES_MICROPROFILE_FAULT_TOLERANCE, new MicroProfileFaultToleranceDeploymentProcessor());
```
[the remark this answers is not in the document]
rhusar: Oops, fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        if (!hystrixConfigured.getAndSet(true)) {
+            // TODO Alternative solution to support hystrix configuration only from the "globally scoped config"
+            // as opposed to deployments providing their configuration.
+            //ClassLoader extensionClassLoader = MicroProfileFaultToleranceExtension.class.getClassLoader();
+            //Config mpConfig = ConfigProvider.getConfig(extensionClassLoader);
+
+            Module module = deploymentUnit.getAttachment(Attachments.MODULE);
+            Config mpConfig = ConfigProvider.getConfig(module.getClassLoader());
```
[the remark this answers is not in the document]
rhusar: I have added a capability to establish proper dependency. Thanks!

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main/module.xml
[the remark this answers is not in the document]
rhusar: I have removed the unneeded ones now. Thanks!

yersan (commented):
@rhusar Thank you for review the pending notes, I cannot approve now since the RFE is still pending for the pre-check status, but as it is now, it looks good to me.

bstansberry:
@yersan Your approval does not require any pre-check. A github approval is a code review approval, i.e. you think this PR is good per your understanding of the requirements, which is good information to have. It's the responsibility of mergers to confirm that other requirements are met before we merge. We don't solely rely on a github approvals for that.

yersan (approved):
Thanks for clearing it up @bstansberry , I assume commits will be squashed before merging, from my understanding it looks good.

microprofile/fault-tolerance-smallrye/extension/src/main/resources/org/wildfly/extension/microprofile/faulttolerance/LocalDescriptions.properties
```diff
@@ -0,0 +1,25 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this software; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+# 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+#
+
+microprofile-fault-tolerance-smallrye=MicroProfile Fault Tolerance extension.
+microprofile-fault-tolerance-smallrye.add=Add MicroProfile Fault Tolerance extension.
+microprofile-fault-tolerance-smallrye.remove=Remove MicroProfile Fault Tolerance extension.
```
pferraro: These descriptions describe the subsystem resource, not the extension (which provides the subsystem).
i.e. replace "extension" with "subsystem".

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceCDIExtension.java
```diff
@@ -0,0 +1,46 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.microprofile.faulttolerance.deployment;
```
pferraro: Should this not reside within org.wildfly.extension.microprofile.faulttolerance.cdi ?

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+            weldCapability = support.getCapabilityRuntimeAPI(Capabilities.WELD_CAPABILITY_NAME, WeldCapability.class);
+        } catch (CapabilityServiceSupport.NoSuchCapabilityException e) {
+            throw MicroProfileFaultToleranceLogger.ROOT_LOGGER.deploymentRequiresCapability(deploymentUnit.getName(), WELD_CAPABILITY_NAME);
+        }
+        if (weldCapability.isPartOfWeldDeployment(deploymentUnit)) {
+            weldCapability.registerExtensionInstance(new MicroProfileFaultToleranceCDIExtension(), deploymentUnit);
+        }
+
+        synchronized (remainingHystrixConfiguringDeployments) {
+            if (remainingHystrixConfiguringDeployments.get() == 0) {
```
pferraro: This is racy.  Any thread reaching this block will compete with a undeploy thread that just set the remainingHystrixConfiguringDeployments counter to 0.  This can result an undeploy thread clearing the configuration that a deploy thread just created.

pferraro (changes requested):

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDependenciesProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public void deploy(DeploymentPhaseContext phaseContext) {
+        DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+
+        ModuleSpecification moduleSpecification = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
+        ModuleLoader moduleLoader = Module.getBootModuleLoader();
+
+        // Fault Tolerance API
+        moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "org.eclipse.microprofile.fault-tolerance.api", false, false, true, false));
```
pferraro: What services does this module have to import?

pferraro: If this is not a weld deployment, we don't need to register these dependencies, right?

pferraro (changes requested):

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="io.smallrye.fault-tolerance">
```
pferraro: This module should be private.

pferraro (changes requested):

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MicroProfileFaultToleranceResourceDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+
+    enum Capability implements org.jboss.as.clustering.controller.Capability {
+        MICROPROFILE_FAULT_TOLERANCE("org.wildfly.microprofile.fault-tolerance"),
+        ;
+        private final RuntimeCapability<Void> definition;
+
+        Capability(String name) {
+            this.definition = RuntimeCapability.Builder.of(name)
+                    .addRequirements("org.wildfly.microprofile.config")
+                    .build();
```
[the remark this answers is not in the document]
pferraro: Actually mp-config already requires weld, so this is unnecessary.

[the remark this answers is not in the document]
rhusar: Right, that's the intention.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+    public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+        DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+
+        // Weld Extension
+        CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+
+        WeldCapability weldCapability;
+        try {
+            weldCapability = support.getCapabilityRuntimeAPI(Capabilities.WELD_CAPABILITY_NAME, WeldCapability.class);
+        } catch (CapabilityServiceSupport.NoSuchCapabilityException e) {
```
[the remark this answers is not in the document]
rhusar: Sorry, I am not following, this line is catching not throwing. The subsequent line thows DeploymentUnitProcessingException.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceCDIExtension.java
rhusar: The idea of `org.wildfly.extension.microprofile.faulttolerance.cdi` is to include the minimum we need to expose to deployments, so we don't have to expose the subsystem to the deployments.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
rhusar: Good point, org.wildfly.extension.microprofile.faulttolerance.deployment.MicroProfileFaultToleranceDeploymentProcessor#undeploy needs to be in the synchronized block as well then.

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
rhusar: I was just trying to be consistent with with config module is doing. But I agree it should be private.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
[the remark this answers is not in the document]
pferraro: Meaning, I wouldn't throw a checked exception (or bother with a localized message) for a condition that cannot occur.  In other words `throw new IllegalStateException(e)` is sufficient.
I'm not entirely sure why CapabilityServiceSupport.getCapabilityRuntimeAPI(...) throws this exception, unless it expects it to be used before the runtime stage (i.e. before requirement constraints are enforced).

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceCDIExtension.java
pferraro: We shouldn't need to expose this module to deployments either.  See below.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
pferraro: You could use: 
```java
if (remainingHystrixConfiguringDeployments.compareAndSet(0, 1)) {
   // create config
} else {
   remainingHystrixConfiguringDeployments.incrementAndGet();
}
```

In which case, you don't need to synchronize anything - however, this might allow deployments to attempt to read the config before it is finished being created.  You could then solve that problem with a latch or barrier.  However, it might just be easier to use synchronized blocks in deploy vs undeploy, in which case you can replace the AtomicInteger with a volatile int.

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
pferraro: microprofile-config needs to do this only because of the way smallrye-config relies on the TCCL.  See https://issues.redhat.com/browse/WFLY-12861

microprofile/fault-tolerance-smallrye/extension/src/main/resources/org/wildfly/extension/microprofile/faulttolerance/LocalDescriptions.properties
rhusar: Fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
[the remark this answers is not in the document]
rhusar: Fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
rhusar: Using synchronized block now.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDependenciesProcessor.java
rhusar: None - fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDependenciesProcessor.java
```diff
@@ -0,0 +1,56 @@
...
+        // Fault Tolerance SmallRye Implementation
+        moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "io.smallrye.fault-tolerance", false, false, true, false));
```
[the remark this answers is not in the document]
rhusar: Thanks for looking into this – fixed now.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDependenciesProcessor.java
rhusar: Fixed.

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
rhusar: Fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+            throw MicroProfileFaultToleranceLogger.ROOT_LOGGER.deploymentRequiresCapability(deploymentUnit.getName(), WELD_CAPABILITY_NAME);
+        }
+        if (weldCapability.isPartOfWeldDeployment(deploymentUnit)) {
+            weldCapability.registerExtensionInstance(new MicroProfileFaultToleranceCDIExtension(), deploymentUnit);
```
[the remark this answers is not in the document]
rhusar: Thanks for looking into this, though the correct extension is this

`weldCapability.registerExtensionInstance(new HystrixExtension(), deploymentUnit);`

Fixed now and removed the dependency from deployments.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceCDIExtension.java
rhusar: I misspoke, its the minimal set to expose to `io.smallrye.fault-tolerance`.

pferraro (changes requested):

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+        CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+
+        WeldCapability weldCapability;
+        try {
+            weldCapability = support.getCapabilityRuntimeAPI(Capabilities.WELD_CAPABILITY_NAME, WeldCapability.class);
+        } catch (CapabilityServiceSupport.NoSuchCapabilityException e) {
+            throw new IllegalStateException();
+        }
+
+        if (weldCapability.isPartOfWeldDeployment(deploymentUnit)) {
```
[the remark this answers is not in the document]
rhusar: True – removed now.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceMarker.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jboss.as.server.deployment.DeploymentUnit;
+import org.jboss.as.server.deployment.annotation.CompositeIndex;
+import org.wildfly.extension.microprofile.faulttolerance.MicroProfileFaultToleranceLogger;
+
+/**
+ * Utility class which marks MP FT deployments.
+ *
+ * @author Radoslav Husar
+ */
+public enum MicroProfileFaultToleranceMarker {
```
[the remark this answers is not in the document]
rhusar: That's the latest fad from IntellJ IDEA to create utility classes as enums – clearly that's bringing more confusion then value since you have to ask. Reworked with private constructor then.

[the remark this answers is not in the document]
pferraro: Enum are indeed useful for implementing the singleton pattern - but to do that you would have a single INSTANCE value, and your methods would be non-static.

mnovak (approved):
This PR passed pre-check of the related RFE thus approving (note I didn't do code review of this PR).

rhusar:
Thanks @mnovak1, great job!

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Activates CDI request context when a Hystrix command that wraps a Fault Tolerance operation is executed.
+ *
+ * @author Martin Kouba
+ * @author Radoslav Husar
+ */
+@ApplicationScoped
+public class RequestContextCommandListener implements CommandListener {
+
+    private final BeanManagerImpl beanManager;
```
pferraro: Why inject the BeanManager implementation class instead of org.jboss.weld.manager.api.WeldManager?  Then you can remove the dependencies on the org.jboss.weld.core and org.jboss.weld.spi modules.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceAnnotation.java
```diff
@@ -0,0 +1,78 @@
...
+    ASYNCHRONOUS("Asynchronous"),
+    ;
+
+    private final DotName dotName;
+
+    MicroProfileFaultToleranceAnnotation(String simpleName) {
+        this.dotName = DotName.createComponentized(DotNames.ORG_ECLIPSE_MICROPROFILE_FAULTTOLERANCE, simpleName);
+    }
+
+    private enum DotNames {
```
pferraro: Why is this an enum?

microprofile/fault-tolerance-smallrye/extension/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+            <artifactId>weld-spi</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-core-impl</artifactId>
+        </dependency>
```
pferraro: Are these really build dependencies?  I don't see where they are referenced.

testsuite/integration/microprofile-tck/fault-tolerance/pom.xml
```diff
@@ -0,0 +1,211 @@
...
+        <ts.elytron.cli>${jbossas.ts.dir}/shared/enable-elytron.cli</ts.elytron.cli>
+        <jboss.dist>${project.build.directory}/wildfly</jboss.dist>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-core-impl</artifactId>
+            <scope>test</scope>
+        </dependency>
```
pferraro: Is this really a dependency for running the TCK?

pferraro (changes requested):

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
```diff
@@ -0,0 +1,49 @@
...
+        <property name="jboss.api" value="private"/>
+    </properties>
+    <resources>
+        <artifact name="${io.smallrye:smallrye-fault-tolerance}"/>
+    </resources>
+    <dependencies>
+        <module name="com.netflix.hystrix.core"/>
+        <module name="io.reactivex.rxjava"/>
+        <module name="io.smallrye.config"/>
+        <module name="io.smallrye.metrics"/>
```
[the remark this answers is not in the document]
bstansberry: Further to this I don't see io.reactivex.rxjava in https://github.com/smallrye/smallrye-fault-tolerance/blob/2.1.3/pom.xml

Not needing two variants of RxJava in WildFly would be better.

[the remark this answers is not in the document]
bstansberry: I missed that hystrix has the rxjava dep. I still don't see why it's needed here and it would be nice to be rid of it altogether in favor of rxjava2, but I can imagine getting rid of it altogether might be a problem.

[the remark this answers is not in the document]
rhusar: @bstansberry Dependencies in maven can have other dependencies – its a dependency brought in by Hystrix. SR just didn't make it explicit which they should since they use it too (in `CompositeObservableCommand` and `HystrixCommandInterceptor`).

```
[INFO] io.smallrye:smallrye-fault-tolerance:jar:2.1.4-SNAPSHOT
[INFO] +- org.eclipse.microprofile.fault-tolerance:microprofile-fault-tolerance-api:jar:2.0.2:compile
[INFO] +- com.netflix.hystrix:hystrix-core:jar:1.5.12:compile
[INFO] |  +- org.slf4j:slf4j-api:jar:1.7.0:compile
[INFO] |  +- com.netflix.archaius:archaius-core:jar:0.4.1:compile
[INFO] |  |  \- commons-configuration:commons-configuration:jar:1.8:compile
[INFO] |  |     +- commons-lang:commons-lang:jar:2.6:compile
[INFO] |  |     \- commons-logging:commons-logging:jar:1.1.1:compile
[INFO] |  +- io.reactivex:rxjava:jar:1.2.0:compile
[INFO] |  \- org.hdrhistogram:HdrHistogram:jar:2.1.9:compile
```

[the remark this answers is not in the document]
rhusar: @bstansberry Its needed by the classes mentioned above^. This will be get ridden off with SR FT .next soonish.

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
rhusar: That's really question for Martin, but doing what you suggested yields

Caused by: java.lang.ClassNotFoundException: org.jboss.weld.manager.api.WeldManager from [Module \"org.wildfly.microprofile.fault-tolerance-smallrye.cdi\" version 19.0.0.Beta1-SNAPSHOT from local module loader @6537cf78 (finder: local module finder @67b6d4ae (roots: /Users/rhusar/git/wildfly/testsuite/integration/microprofile/target/wildfly/modules,/Users/rhusar/git/wildfly/testsuite/integration/microprofile/target/wildfly/modules/system/layers/base))]

rhusar: The `org.jboss.weld.manager.api.WeldManager` is in the weld-spi module....

rhusar: Anyway, I dont see how can those be removed.

feature-pack/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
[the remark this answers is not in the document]
rhusar: @pferraro They are necessary until SR FT is rewritten. Removing that dependency fails with:
```
    Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type MetricRegistry with qualifiers @Default
  at injection point [UnbackedAnnotatedField] @Inject io.smallrye.faulttolerance.metrics.MetricsCollectorFactory.registry
  at io.smallrye.faulttolerance.metrics.MetricsCollectorFactory.registry(MetricsCollectorFactory.java:0)
```

feature-pack/src/main/resources/modules/system/layers/base/org/hdrhistogram/main/module.xml
```diff
@@ -0,0 +1,32 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.hdrhistogram">
```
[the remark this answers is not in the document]
rhusar: Yes, that was the intention - fixed.

microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceAnnotation.java
rhusar: Changed to no-method interface...

microprofile/fault-tolerance-smallrye/extension/pom.xml
rhusar: I have explained that above. Let me remove it so I don't need to explain again.

testsuite/integration/microprofile-tck/fault-tolerance/pom.xml
rhusar: I have explained that above. Let me remove it so I don't need to explain again.

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
pferraro: You can at least remove the dependency on the org.jboss.weld.core module, no?

rhusar: No – that fails with CNF as well.

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
```diff
@@ -0,0 +1,72 @@
...
+public class RequestContextCommandListener implements CommandListener {
+
+    private final WeldManager weldManager;
+
+    private final RequestContext requestContext;
+
+    private static final ThreadLocal<Boolean> isActivator = new ThreadLocal<>();
+
+    @Inject
+    public RequestContextCommandListener(@Unbound RequestContext requestContext, BeanManagerImpl weldManager) {
```
pferraro: Use WeldManager.  Then remove dependency on weld-core-impl.

pferraro (changes requested):

rhusar:
Proactively pinging mergers @jamezp @kabir @bstansberry @pferraro @darranl to look into merging this. Please realize that all of the MP PRs will conflict so a lot of rebasing is ahead so the earlier we start the better :-) but mainly yesterday was my last working day this year and I will be off in the beginning of January as well, and rebasing is the last thing I want to do on vacation.

kabir (changes requested):
You need to add the licenses for all the new dependencies to full-feature-pack-licenses.xml.

rhusar:
@kabir Ok, updated, I thought that task was automated.

darranl:
I think there is a magic group to ping @wildfly/mergers-full but maybe you have to be a marger to ping a merger ;-)

rhusar:
@darranl yes i know but we cannot do that. That group gives us 404 because we are not members of the organization.

pferraro (approved):
This looks good now.

darranl:
@rhusar going to discuss it with the other mergers but I think there could be a way to get that to work, also to make it easier for non-mergers to provide reviews.

feature-pack/src/license/full-feature-pack-licenses.xml
```diff
@@ -1991,6 +2057,20 @@
...
+        <license>
+          <name>Public Domain, per Creative Commons CC0</name>
+          <url>http://creativecommons.org/publicdomain/zero/1.0/</url>
+        </license>
+        <license>
+          <name>BSD-2-Clause</name>
+          <url>https://opensource.org/licenses/BSD-2-Clause</url>
+        </license>
+      </licenses>
+    </dependency>
```
jamezp: These two are not quite correct, but I'll file an issue to fix these later as it's not important until a release is to be done. Seems like something that should not hold up the PR.

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
```diff
@@ -0,0 +1,71 @@
...
+    @Inject
+    public RequestContextCommandListener(@Unbound RequestContext requestContext, WeldManager weldManager) {
+        this.requestContext = requestContext;
+        this.weldManager = weldManager;
+    }
+
+    @Override
+    public void beforeExecution(FaultToleranceOperation operation) {
+        // Note that Hystrix commands are executed on a dedicated thread where a CDI request context should not be active
+        // TODO: maybe we should activate the context for any FT operation?
```
jamezp: Again not holding up the PR for this, but is there a resolution to this TODO?

jamezp (approved):

feature-pack/src/license/full-feature-pack-licenses.xml
jamezp: Just linking the JIRA here https://issues.redhat.com/browse/WFLY-12909

rhusar: WFLY-12909 is resolved now, resoving here as well.

microprofile/fault-tolerance-smallrye/cdi/src/main/java/org/wildfly/extension/microprofile/faulttolerance/cdi/RequestContextCommandListener.java
rhusar: This should be resolved with SR FT 4.x coming in the next release, the TODO is from the ported code.

PR #12804, opened 2019-11-15
[WFLY-12624] Adjust default client mapping to return hostname if possible

rachmatowicz:
This PR does the following:
- adjusts the information that is sent in the default client-mapping for the server to the EJB client in a topology update. Prefer hostnames over IP addresses whenever the server is bound to a hostname rather than a IP address literal.
- this supports EJBCLIENT-349

Issue: https://issues.jboss.org/browse/WFLY-12624

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemotingConnectorClientMappingsEntryProviderService.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemotingConnectorClientMappingsEntryProviderService.java
@@ -22,6 +22,7 @@
 
 package org.jboss.as.ejb3.remote;
 
+import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.AbstractMap;
@@ -49,6 +50,7 @@ import org.wildfly.clustering.service.ServiceSupplierDependency;
 import org.wildfly.clustering.service.SupplierDependency;
 import org.wildfly.clustering.spi.ClusteringCacheRequirement;
 import org.wildfly.clustering.spi.ClusteringRequirement;
+import org.wildfly.common.net.Inet;
 
 /**
  * @author Jaikiran Pai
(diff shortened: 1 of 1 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemotingConnectorClientMappingsEntryProviderService.java
```diff
@@ -91,24 +93,36 @@ public ServiceConfigurator configure(OperationContext context) {
...
             ret.addAll(info.getSocketBinding().getClientMappings());
         } else {
-            // TODO: We use the textual form of IP address as the destination address for now.
-            // This needs to be configurable (i.e. send either host name or the IP address). But
-            // since this is a corner case (i.e. absence of any client-mappings for a socket binding),
-            // this should be OK for now
-            final String destinationAddress = info.getSocketBinding().getAddress().getHostAddress();
+            // for the destination, prefer the hostname over the literal IP address
+            final InetAddress destination = info.getSocketBinding().getAddress();
+            final String destinationName = Inet.toURLString(destination, true);
```
bstansberry: This will wrap an IPv6 address literal with [ ] which is different from the previous code. Will the consumer handle that correctly?

jamezp:
This looks like it needs a rebase as #12801 has been merged.

ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemotingConnectorClientMappingsEntryProviderService.java
bstansberry: @rachmatowicz ^^^

darranl:
+1 on needing a rebase - in general if borrowing a commit from elsewhere you should try and preserve it's SHA so commit don't end up being duplicated.

ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemotingConnectorClientMappingsEntryProviderService.java
rachmatowicz: I'm pretty sure that I did check that aspect, Brian. The host part gets unpacked in EJBClientChannel in the EJB client library. IIRC, I did run the testsuite with IPv6 addresses and the results looked correct.

bstansberry:
retest this please

PR #12678, opened 2019-09-28
[WFLY-12603][WFLY-12604] Run clustering tests against Galleon-slimmed servers

bstansberry:
https://issues.jboss.org/browse/WFLY-12603
https://issues.jboss.org/browse/WFLY-12604

WFLY-12603 is the main goal. WFLY-12604 is a problem the tests revealed.

PR #12810, opened 2019-11-18
Partial fix for WFLY-12751, Move base layers to the full galleon feat…

bstansberry:
…ure pack

This leaves the web-clustering decorator layer behind pending resolution of test issues with it

Partial completion of https://issues.jboss.org/browse/WFLY-12751


bstansberry:
@jfdenise Look ok? I want to have these 'base' layers available for other things I am doing and don't want to block on #12678 so I dropped the web-clustering aspect from your original commit.

jfdenise (approved):

PR #12812, opened 2019-11-19
WFLY-12805 Add missing dependency

soul2zimate:
Issue: https://issues.jboss.org/browse/WFLY-12805

Add missing dependency for weld transaction.

@manovotn Please review.


manovotn:
I am currently on PTO, @mkouba can you please glance at this?

mkouba:
The PR looks good. The only thing I'm concerned about is that there are most probably no automated tests with JTS enabled?

soul2zimate:
I don't see such weld test with JTS enabled. There is only one for JTA.

PR #12814, opened 2019-11-21
[WFLY-12575][WFLY-12697] Upgrade to RESTEasy 3.9.2.Final

asoldano:
https://issues.jboss.org/browse/WFLY-12575
https://issues.jboss.org/browse/WFLY-12697


jamezp:
These failures definitely look related :(

asoldano:
Ah, I noticed similar thing in RESTEasy testsuite, Jackson 2.10 seems to have changed the default format used for duration values. I believe we just need to put up with that and update the test; we could argue the test is checking impl details here

jamezp:
Closing this in favor of #12818 12818

PR #12822, opened 2019-11-25
WFLY-12814 PreDestroy method not called by stateful beans with timeout 0

chengfang:
JIRA: https://issues.jboss.org/browse/WFLY-12814

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
@@ -147,8 +147,13 @@ public class SimpleCache<K, V extends Identifiable<K>> implements Cache<K, V>, P
         Entry<V> entry = this.entries.get(id);
         if ((entry != null) && entry.done()) {
             if (this.timeout != null) {
-                // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
-                this.scheduler.schedule(id, this.timeout.isZero() ? Instant.now() : Instant.now().plus(this.timeout));
+                if (!this.timeout.isZero()) {
+                    this.scheduler.schedule(id, Instant.now().plus(this.timeout));
+                } else {
+                    // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
+                    // However, removing it directly is faster than scheduling it for immediate removal.
+                    remove(id);
+                }
             }
         }
     }
--- a/ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
@@ -147,13 +147,8 @@ public class SimpleCache<K, V extends Identifiable<K>> implements Cache<K, V>, P
         Entry<V> entry = this.entries.get(id);
         if ((entry != null) && entry.done()) {
             if (this.timeout != null) {
-                if (!this.timeout.isZero()) {
-                    this.scheduler.schedule(id, Instant.now().plus(this.timeout));
-                } else {
-                    // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
-                    // However, removing it directly is faster than scheduling it for immediate removal.
-                    remove(id);
-                }
+                // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
+                this.scheduler.schedule(id, this.timeout.isZero() ? Instant.now() : Instant.now().plus(this.timeout));
             }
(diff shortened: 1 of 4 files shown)


chengfang:
@pferraro can you please review?

ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
```diff
@@ -147,13 +147,8 @@ public void release(V bean) {
...
             if (this.timeout != null) {
-                if (!this.timeout.isZero()) {
-                    this.scheduler.schedule(id, Instant.now().plus(this.timeout));
-                } else {
-                    // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
-                    // However, removing it directly is faster than scheduling it for immediate removal.
-                    remove(id);
-                }
+                // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
+                this.scheduler.schedule(id, this.timeout.isZero() ? Instant.now() : Instant.now().plus(this.timeout));
```
pferraro: I don't believe that this change is sufficient to completely fix this bug.
Essentially, the above change merely moves the the call to remove(...) to a separate thread, and thus has the effect of usually delaying the call to Cache.remove(...).

The root cause of the bug (i.e. the reason why the @PreDestroy callback is not being triggered) is actually due to this logic:
https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionComponent.java#L158
The @PreDestroy is skipped because StatefulSessionComponentInstance.isSynchronizationRegistered() still returns true.

If you look at the release logic within StatefulSessionSynchronizationInterceptor:
https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionSynchronizationInterceptor.java#L206
The call to Cache.release(...) precedes the call to StatefulSessionComponentInstance.setSynchronizationRegistered(false).
Thus, even with the above changes, it is still possible for the @PreDestroy listener to be skipped, if the scheduler thread initiates the call to Cache.remove(...) before the interceptor sets the synchronizationRegistered flag.

To completely fix this issue, I would instead suggest that logic within StatefulSessionSynchronizationInterceptor.releaseInstance(...) be modified such that setSynchronizationRegistered(false) is called before the instance is released from the cache.

pferraro (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
chengfang: Thanks, @pferraro for the detailed analysis.  I'll look into this more.

ejb3/src/main/java/org/jboss/as/ejb3/cache/simple/SimpleCache.java
```diff
@@ -147,8 +147,13 @@ public void release(V bean) {
...
             if (this.timeout != null) {
-                // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
-                this.scheduler.schedule(id, this.timeout.isZero() ? Instant.now() : Instant.now().plus(this.timeout));
+                if (!this.timeout.isZero()) {
+                    this.scheduler.schedule(id, Instant.now().plus(this.timeout));
+                } else {
+                    // The EJB specification allows a 0 timeout, which means the bean is immediately eligible for removal.
+                    // However, removing it directly is faster than scheduling it for immediate removal.
+                    remove(id);
+                }
```
pferraro: This change should be reverted.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/stateful/timeout/StatefulTimeoutTestCase.java
```diff
@@ -162,4 +162,21 @@ public void testStatefulBeanWithPassivationNotDiscardedWhileInTransaction() thro
...
+
+    /**
+     * Verifies that a stateful bean with timeout value 0 is eligible fore removal immediately, and its
+     * preDestroy method is invoked.
+     */
+    @Test
+    public void testStatefulTimeout0() throws Exception {
+        Annotated0TimeoutBean timeout0 = lookup(Annotated0TimeoutBean.class);
+        timeout0.doStuff();
+        Thread.sleep(2000);
```
pferraro: This sleep should not be necessary.

pferraro (changes requested):

pferraro (approved):
LGTM.

PR #12824, opened 2019-11-25
[WFLY-12811] Fix injection of MP Config in HealthCheck bean

jmesnil:
* When HealthChecks are called from WildFly management endpoints for
  health, set the TCCL to the deployment's classloader so that resources that
  depend on it (such as MicroProfile Config) can be properly loaded.
* Add a ConfigProperty to MicroProfileHealthTestBase test whose value
  is only set in a ConfigSource inside the deployment to ensure that if
  the TCCL was not correct, the property would not be found and the test
  would fail.

JIRA: https://issues.jboss.org/browse/WFLY-12811

### DIFF
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporter.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporter.java
@@ -22,8 +22,7 @@
 package org.wildfly.extension.microprofile.health;
 
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 import java.util.Map;
 
 import javax.json.Json;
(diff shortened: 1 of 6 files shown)


microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporter.java
```diff
@@ -93,7 +93,16 @@ private final SmallRyeHealth getHealth(String emptyChecksStatus, List<HealthChec
     private HealthCheckResponse.State processChecks(Iterable<HealthCheck> checks, JsonArrayBuilder results, HealthCheckResponse.State status) {
         if (checks != null) {
             for (HealthCheck check : checks) {
-                status = fillCheck(check, results, status);
+                // use the HealthCheck class loader instead of the TCCL (which is the server's ModuleClassLoader
+                // to ensure that any resources that checks the TCCL (such as MP Config) will use the correct one
+                // when the health checks are called.
+                final ClassLoader oldTCCL = Thread.currentThread().getContextClassLoader();
+                try {
+                    Thread.currentThread().setContextClassLoader(check.getClass().getClassLoader());
```
bstansberry: This is kind of an approximation of the deployment's classloader; i.e. the check might be loaded from an external module or some other classloading arrangement like libs in an ear. Perhaps that doesn't matter. And in any case this handles the common case, so +1.

One point is fillCheck is doing a null check on the 'check' variable. Either that's unneeded there or it's needed here.

jmesnil: @bstansberry thanks, I've amended the code so that now, it uses the DeploymenUnit's module class loader to prevent any approximation (that's the same class loader used to register MicroProfile Config's `Config` instance).

I've also removed the null check in `fillCheck` as I checked for `null` when adding the HealthChecks in the various `addXXX` methods of `HealthReporter`

bstansberry (approved):

PR #12827, opened 2019-11-26
Update MicroProfileMetricsCounterMultipleInvocationsTestCase

jmesnil:
Add a CustomCounterMetrics that depends on MicroProfile Config to
configure itself.


jmesnil:
@istraka fiy

bstansberry (approved):

PR #12836, opened 2019-12-06
[WFLY-12758]:  Integrating OpenTracing

ehsavoie:
* Configuring jaeger tracers.
* Need to obtain the configuration service in the TraceInitializer.
* opentracing-interceptors dependency is set to provided.

Jira: https://issues.jboss.org/browse/WFLY-12758
      https://issues.jboss.org/browse/WFLY-12812


ehsavoie:
@jmesnil @pferraro  @zhantemirov Could you please review this one ?

ehsavoie:
Galleon failure is due to the fact that jaeger is not provisioned if no tracer is defined in the subsystem which is now required with

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+        super.performRuntime(context, operation, model);
+        ServiceName serviceName = TRACER_CAPABILITY.getCapabilityServiceName(context.getCurrentAddressValue());
+        ServiceBuilder builder = context.getServiceTarget().addService(serviceName);
+        String outboundSocketBindingName = TracerAttributes.SENDER_BINDING.resolveModelAttribute(context, model).asStringOrNull();
+        Supplier<OutboundSocketBinding> outboundSocketBindingSupplier;
+        if (outboundSocketBindingName != null) {
+            outboundSocketBindingSupplier = builder.requires(OUTBOUND_SOCKET_BINDING_CAPABILITY.getCapabilityServiceName(outboundSocketBindingName));
+        } else {
+            outboundSocketBindingSupplier = new Supplier<OutboundSocketBinding>() {
```
pferraro: You can replace this anonymous class with:
org.wildfly.common.function.Functions.constantSupplier(null);

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -32,42 +32,77 @@
...
     private static final RuntimeCapability<Void> OPENTRACING_CAPABILITY = RuntimeCapability.Builder
             .of(OPENTRACING_CAPABILITY_NAME)
             .addRequirements(WELD_CAPABILITY_NAME)
             .build();
 
+    public static final RuntimeCapability<Void> DEFAULT_TRACER_CAPABILITY = RuntimeCapability.Builder
+            .of(DEFAULT_TRACER_CAPABILITY_NAME, true)
+            .build();
+
+    public static final RuntimeCapability<Void> TRACER_CAPABILITY = RuntimeCapability.Builder
```
pferraro: This duplicates JaegerTracerConfigurationDefinition.TRACER_CAPABILITY.  You can delete this.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -32,42 +32,77 @@
...
+            .setRestartAllServices()
+            .build();
+
     protected SubsystemDefinition() {
-        super( new SimpleResourceDefinition.Parameters(SubsystemExtension.SUBSYSTEM_PATH, SubsystemExtension.getResourceDescriptionResolver(SubsystemExtension.SUBSYSTEM_NAME))
+        super(new SimpleResourceDefinition.Parameters(SubsystemExtension.SUBSYSTEM_PATH, SubsystemExtension.getResourceDescriptionResolver())
                 .setAddHandler(SubsystemAdd.INSTANCE)
-                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
+                .setRemoveHandler(new SubsystemRemoveHandler())
                 .setCapabilities(OPENTRACING_CAPABILITY)
```
pferraro: Should you not also register DEFAULT_TRACER_CAPABILITY?

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -32,42 +32,77 @@
...
     private static final String OPENTRACING_CAPABILITY_NAME = "org.wildfly.microprofile.opentracing";
+    public static final String DEFAULT_TRACER_CAPABILITY_NAME = "org.wildfly.microprofile.opentracing.default-tracer";
 
     private static final RuntimeCapability<Void> OPENTRACING_CAPABILITY = RuntimeCapability.Builder
             .of(OPENTRACING_CAPABILITY_NAME)
             .addRequirements(WELD_CAPABILITY_NAME)
             .build();
 
+    public static final RuntimeCapability<Void> DEFAULT_TRACER_CAPABILITY = RuntimeCapability.Builder
+            .of(DEFAULT_TRACER_CAPABILITY_NAME, true)
```
pferraro: This should probably define a service type of TracerConfiguration.  It should not be dynamic.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -32,42 +32,77 @@
...
-            "org.wildfly.microprofile.opentracing-smallrye",
-            "io.opentracing.contrib.opentracing-interceptors",
-    };
+        "io.smallrye.opentracing",
+        "org.wildfly.microprofile.opentracing-smallrye",
+        "io.opentracing.contrib.opentracing-interceptors",};
+
+    public static final SimpleAttributeDefinition DEFAULT_TRACER = SimpleAttributeDefinitionBuilder
+            .create(TracerConfigurationConstants.DEFAULT_TRACER, ModelType.STRING, true)
+            .setCapabilityReference(TRACER_CAPABILITY_NAME)
```
pferraro: After you additionally register DEFAULT_TRACER_CAPABILITY (see below), you won't be able to use this method.  You'll instead need to use:
setCapabilityReference(TRACER_CAPABILITY_NAME, DEFAULT_TRACER_CAPABILITY)

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -79,4 +114,39 @@ public void registerAdditionalRuntimePackages(final ManagementResourceRegistrati
...
+    private static final class DefaultTracerWriteAttributeHandler extends ReloadRequiredWriteAttributeHandler {
+
+        private DefaultTracerWriteAttributeHandler() {
+            super(DEFAULT_TRACER);
+        }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, AttributeDefinition attributeDefinition,
+                ModelNode newValue, ModelNode oldValue) {
+            super.recordCapabilitiesAndRequirements(context, attributeDefinition, newValue, oldValue);
```
pferraro: I don't think you want to call the super implementation, as you are handling capability registration conditionally.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -79,4 +114,39 @@ public void registerAdditionalRuntimePackages(final ManagementResourceRegistrati
...
+
+    private static final class SubsystemRemoveHandler extends ReloadRequiredRemoveStepHandler {
+
+        private SubsystemRemoveHandler() {
+            super();
+        }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+            super.recordCapabilitiesAndRequirements(context, operation, resource);
```
pferraro: I don't think you want to call the super implementation, as you are handling capability registration conditionally.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemAdd.java
```diff
@@ -19,41 +19,60 @@
...
                         new TracingDeploymentProcessor()
                 );
             }
         }, OperationContext.Stage.RUNTIME);
+        super.performBoottime(context, operation, model);
+    }
+
+    @Override
+    protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+        super.recordCapabilitiesAndRequirements(context, operation, resource);
```
pferraro: I don't think you want to call the super implementation, as you are handling capability registration conditionally.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemAdd.java
```diff
@@ -19,41 +19,60 @@
...
+        ModelNode defaultTracer = DEFAULT_TRACER.resolveModelAttribute(context, operation);
+        if (defaultTracer.isDefined()) {
+            context.registerCapability(RuntimeCapability.Builder.of(DEFAULT_TRACER_CAPABILITY.getDynamicName(defaultTracer.asString()), false).build());
```
pferraro: DEFAULT_TRACER_CAPABILITY shouldn't be dynamic.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -79,4 +114,39 @@ public void registerAdditionalRuntimePackages(final ManagementResourceRegistrati
...
+        private DefaultTracerWriteAttributeHandler() {
+            super(DEFAULT_TRACER);
+        }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, AttributeDefinition attributeDefinition,
+                ModelNode newValue, ModelNode oldValue) {
+            super.recordCapabilitiesAndRequirements(context, attributeDefinition, newValue, oldValue);
+            if (oldValue.isDefined()) {
+                context.deregisterCapability(DEFAULT_TRACER_CAPABILITY.getDynamicName(oldValue.asString()));
```
pferraro: DEFAULT_TRACER_CAPABILITY shouldn't be dynamic.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -79,4 +114,39 @@ public void registerAdditionalRuntimePackages(final ManagementResourceRegistrati
...
+            }
+            if (newValue.isDefined()) {
+                context.registerCapability(RuntimeCapability.Builder.of(DEFAULT_TRACER_CAPABILITY.getDynamicName(newValue.asString()), false).build());
```
pferraro: DEFAULT_TRACER_CAPABILITY shouldn't be dynamic.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
```diff
@@ -79,4 +114,39 @@ public void registerAdditionalRuntimePackages(final ManagementResourceRegistrati
...
+        private SubsystemRemoveHandler() {
+            super();
+        }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource) throws OperationFailedException {
+            super.recordCapabilitiesAndRequirements(context, operation, resource);
+            ModelNode defaultTracer = DEFAULT_TRACER.resolveModelAttribute(context, resource.getModel());
+            if (defaultTracer.isDefined()) {
+                context.deregisterCapability(RuntimeCapability.Builder.of(DEFAULT_TRACER_CAPABILITY.getDynamicName(defaultTracer.asString()), false).build().getName());
```
pferraro: DEFAULT_TRACER_CAPABILITY shouldn't be dynamic.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemExtension.java
```diff
@@ -42,10 +43,17 @@
...
+    private static final ModelVersion CURRENT_MODEL_VERSION = VERSION_2_0_0;
+
+    private static final PersistentResourceXMLParser PARSER = SubsytemParser_2_0.INSTANCE;
+    public static final String NAMESPACE = SubsytemParser_2_0.OPENTRACING_NAMESPACE;
 
     static ResourceDescriptionResolver getResourceDescriptionResolver(final String... keyPrefix) {
-        return getResourceDescriptionResolver(true, keyPrefix);
+        String[] prefixes = new String[keyPrefix.length + 1];
+        prefixes[0] = SUBSYSTEM_NAME;
+        System.arraycopy(keyPrefix, 0, prefixes, 1, keyPrefix.length);
```
pferraro: This seems a bit unnecessary.  Any way to avoid it?

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
```diff
@@ -0,0 +1,110 @@
...
+            RuntimeCapability.Builder.of(OUTBOUND_SOCKET_BINDING_CAPABILITY_NAME,
+                    true, OutboundSocketBinding.class)
+                    .build();
+
+    public static final StringListAttributeDefinition PROPAGATION = StringListAttributeDefinition.Builder.of(TracerConfigurationConstants.PROPAGATION)
+            .setAllowNullElement(false)
+            .setRequired(false)
+            .setAllowedValues("JAEGER", "B3")
+            .setAttributeGroup("codec-configuration")
+            .setRestartAllServices()
```
pferraro: Why must this require a restart of all services?  Same question for all attributes below.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
```diff
@@ -0,0 +1,110 @@
...
+            .build();
+
+    public static final SimpleAttributeDefinition SAMPLER_TYPE = SimpleAttributeDefinitionBuilder.create(TracerConfigurationConstants.SAMPLER_TYPE, ModelType.STRING, true)
+            .setAllowedValues("const", "probabilistic", "ratelimiting", "remote")
```
pferraro: Is there an enum for these values?

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
```diff
@@ -66,8 +78,41 @@ public void deploy(DeploymentPhaseContext deploymentPhaseContext) throws Deploym
                     deploymentPhaseContext.getDeploymentUnit().getName(), WELD_CAPABILITY_NAME
             ));
         }
-        setServiceName(deploymentUnit);
+        Object resolvedTracer = null;
+        try {
+            final Module module = deploymentUnit.getAttachment(Attachments.MODULE);
+            if (module != null) {
```
pferraro: When is this ever null?

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
```diff
@@ -110,10 +168,30 @@ private JBossWebMetaData getJBossWebMetaData(DeploymentUnit deploymentUnit) {
...
     }
 
+    private String getTracerConfiguration(DeploymentPhaseContext deploymentPhaseContext) {
+        DeploymentUnit deploymentUnit = deploymentPhaseContext.getDeploymentUnit();
+        JBossWebMetaData jbossWebMetaData = getJBossWebMetaData(deploymentUnit);
+        if (null == jbossWebMetaData || null == jbossWebMetaData.getContextParams()) {
+            return null;
+        }
+        for (ParamValueMetaData param : jbossWebMetaData.getContextParams()) {
+            if (TracerInitializer.SMALLRYE_OPENTRACING_TRACER_CONFIGURATION.equals(param.getParamName())) {
```
pferraro: Having not read the opentracing spec thoroughly...
Is this really how users should define opentracing configuration?  I would expect this configuration to be defined via MP Config. Configuration via servlet context parameters is very retro.

pferraro (changes requested):
Cursory review.  More to come.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private JaegerTracerConfigurationAddHandler() {
+        super(ATTRIBUTES);
+    }
+
+    @Override
+    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+        super.performRuntime(context, operation, model);
+        ServiceName serviceName = TRACER_CAPABILITY.getCapabilityServiceName(context.getCurrentAddressValue());
+        ServiceBuilder builder = context.getServiceTarget().addService(serviceName);
```
bstansberry: This should should use context.getCapabilityServiceTarget.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        String outboundSocketBindingName = TracerAttributes.SENDER_BINDING.resolveModelAttribute(context, model).asStringOrNull();
+        Supplier<OutboundSocketBinding> outboundSocketBindingSupplier;
+        if (outboundSocketBindingName != null) {
+            outboundSocketBindingSupplier = builder.requires(OUTBOUND_SOCKET_BINDING_CAPABILITY.getCapabilityServiceName(outboundSocketBindingName));
```
bstansberry: Using CapabilityServiceBuilder from CapabilityServiceTarget removes the need for a fake OUTBOUND_SOCKET_BINDING_CAPABILITY just to generate a ServiceName.

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/WildFlyTracerFactory.java
```diff
@@ -0,0 +1,110 @@
...
+import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceName;
+
+/**
+ * Registry of all the tracers configuration registered in WildFly.
+ * @author Emmanuel Hugonnet (c) 2019 Red Hat, Inc.
+ */
+public class WildFlyTracerFactory {
+
+    private static final Map<ServiceName, Configuration> CONFIGURATIONS = new HashMap<>();
```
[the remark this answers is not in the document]
bstansberry: This can be Map<String, Configuration> with null used for the compatibility tracer. That eliminates all the conversion from simple tracer name string -> RuntimeCapability -> ServiceName.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/OpentracingConfigurationService.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public void stop(StopContext context) {
+        consumer.accept(null);
+    }
+
+
+    @SuppressWarnings("unchecked")
+    public static final void installTracerConfigurationService(ServiceBuilder builder, TracerConfiguration config, ServiceName serviceName) {
+        builder.setInstance(new OpentracingConfigurationService(WildFlyTracerFactory.registerTracer(serviceName), config));
+        builder.install();
```
bstansberry: If WildFlyTracerFactory.registerTracer took a simple config name string and not a ServiceName, then the OCS constructor is
```
public OpentracingConfigurationService(TracerConfiguration config, String configName) {
   this.config = config;
   this.consumer = WildFlyTracerFactory.registerTracer(configName)
}
```
and then this method can go away with the instance construction and the install moved to the code that is doing the rest of the manipulation of the ServiceBuilder.

If a getName() method was valid for the TracerConfiguration interface, then you just need 
```
public OpentracingConfigurationService(TracerConfiguration config) { 
```

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationDefinition.java
```diff
@@ -0,0 +1,78 @@
...
+
+
+/**
+ *
+ * @author Emmanuel Hugonnet (c) 2019 Red Hat, Inc.
+ */
+public class JaegerTracerConfigurationDefinition extends SimpleResourceDefinition {
+
+    public static final RuntimeCapability<Void> TRACER_CAPABILITY = RuntimeCapability.Builder
+            .of(TRACER_CAPABILITY_NAME, true, OpentracingConfigurationService.class)
```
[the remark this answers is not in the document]
bstansberry: +1.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/opentracing-smallrye/main/module.xml
```diff
@@ -39,6 +40,8 @@
         <module name="org.wildfly.security.elytron"/>
 
         <module name="javax.enterprise.api" />
+        <module name="io.opentracing.opentracing-api"/>
         <module name="org.wildfly.microprofile.opentracing-smallrye"/>
+        <module name="io.jaegertracing.jaeger" optional="true"/>
```
bstansberry: Please add a comment noting what makes it required, e.g.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/microprofile/opentracing-smallrye/main/module.xml
```diff
@@ -25,6 +25,11 @@
     </resources>
 
     <dependencies>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.as.controller"/>
```
bstansberry: See my comment about WildFlyTraceFactory.CONFIGURATIONS using simple String keys, which would remove the need for RuntimeCapability in this module, which removes the need for this dependency.

You need to add org.jboss.dmr though.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/microprofile/opentracing-smallrye/main/module.xml
```diff
@@ -25,6 +25,11 @@
...
+        <module name="org.jboss.as.server"/>
```
bstansberry: I don't see the use of this.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/microprofile/opentracing-smallrye/main/module.xml
```diff
@@ -25,6 +25,11 @@
...
+        <module name="org.wildfly.security.elytron"/>
```
bstansberry: I don't see the use of this.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemAdd.java
ehsavoie: I need it to be able to have the tracer name in the deployment

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
ehsavoie: No because this default is only set if the default attribute (which is nullable) is set

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemExtension.java
ehsavoie: Looking into it but this is not relevant to this PR :)

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
ehsavoie: Because this is a configuration of a tracer, not a tracer itself thus you need to recreate a Tracer instance to update it which means restarting the deployments that have the tracer already injected.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
ehsavoie: Yes in jaeger's code but I don't want to depends on it

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
ehsavoie: I don't know :) Maybe I can remove this

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
ehsavoie: This is how this was done in the initial integration. Tracer configuration is not really specified in opentracing specs.

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/opentracing-smallrye/main/module.xml
ehsavoie: Well it will be required as it is required by adding the legacy behaviour now.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemExtension.java
ehsavoie: no part of this PR ;)  but fixing that

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/OpentracingConfigurationService.java
```diff
@@ -0,0 +1,62 @@
...
+import org.jboss.msc.service.StopContext;
+
+import org.wildfly.microprofile.opentracing.smallrye.TracerConfiguration;
+import org.wildfly.microprofile.opentracing.smallrye.WildFlyTracerFactory;
+
+/**
+ * Service providing the Tracer configuration.
+ * @author Emmanuel Hugonnet (c) 2019 Red Hat, Inc.
+ */
+public class OpentracingConfigurationService implements Service {
```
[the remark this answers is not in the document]
ehsavoie: I agree

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationDefinition.java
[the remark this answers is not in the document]
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
pferraro: But we would only need to restart those deployments with a dependency on the specific tracer configuration, not all deployments, right?
In that case you can use setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES).

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
pferraro: So, in that case, you'll probably want to create your own Enum.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
pferraro: To align with vendor-specific configuration for other MP specs, I would suggest using vendor specific MP config namespace. e.g. "mp.opentracing.extensions.tracer.configuration" (or similar).

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/OpentracingConfigurationService.java
ehsavoie: ok and I can even remove that Service class

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
bstansberry: AttributeAccess.Flag.RESTART_RESOURCE_SERVICES is something of an anti-pattern as a typical user won't understand that bouncing a service means bouncing the deployment. And bouncing a deployment isn't graceful.

This seems like a possible future enhancement.

zhantemirov:
From QE point of view tests are okay, thank you. Additional test cases will be added by me later.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
ehsavoie: I'd rather not as it doesn't add anything since the value is passed to jaeger directly

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracerInitializer.java
```diff
@@ -42,6 +40,7 @@
 public class TracerInitializer implements ServletContextListener {
```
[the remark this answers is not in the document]
ehsavoie: I agree, I have something working for this

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracerAttributes.java
ehsavoie: ok so keeping my restartAllServices

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
ehsavoie: Except for those that were already defined

microprofile/opentracing-extension/src/main/resources/org/wildfly/extension/microprofile/opentracing/LocalDescriptions.properties
```diff
@@ -1,3 +1,33 @@
...
+microprofile-opentracing-smallrye.tracer.compression=Activate compression of traces.
+microprofile-opentracing-smallrye.tracer.connection-timeout=Tracer connection timeout.
+microprofile-opentracing-smallrye.tracer.endpoint=Tracer endpoint.
+microprofile-opentracing-smallrye.tracer.encoding=Tracer message encoding.
+microprofile-opentracing-smallrye.tracer.message-max-bytes=Tracer message max size.
+microprofile-opentracing-smallrye.tracer.max-request=Tracer max request.
+microprofile-opentracing-smallrye.tracer.read-timeout=Tracer read timeout.
+microprofile-opentracing-smallrye.tracer.sampler=Tracer sampler rate.
+microprofile-opentracing-smallrye.tracer.propagation=List of formats to use for propagating the trace context.
+microprofile-opentracing-smallrye.tracer.sampler-param=The Jaeger sampler parameter.
```
bstansberry: This needs a more complete explanation.

pferraro (changes requested):

microprofile/opentracing-extension/src/main/resources/org/wildfly/extension/microprofile/opentracing/LocalDescriptions.properties
```diff
@@ -1,3 +1,33 @@
...
+microprofile-opentracing-smallrye.jaeger-tracer=Jaeger tracer configuration.
+microprofile-opentracing-smallrye.tracer=Tracer configuration.
+microprofile-opentracing-smallrye.tracer.compression=Activate compression of traces.
+microprofile-opentracing-smallrye.tracer.connection-timeout=Tracer connection timeout.
+microprofile-opentracing-smallrye.tracer.endpoint=Tracer endpoint.
+microprofile-opentracing-smallrye.tracer.encoding=Tracer message encoding.
+microprofile-opentracing-smallrye.tracer.message-max-bytes=Tracer message max size.
+microprofile-opentracing-smallrye.tracer.max-request=Tracer max request.
+microprofile-opentracing-smallrye.tracer.read-timeout=Tracer read timeout.
+microprofile-opentracing-smallrye.tracer.sampler=Tracer sampler rate.
```
bstansberry: rate?

kabir:
Retest this please

feature-pack/src/main/resources/modules/system/layers/base/org/wildfly/microprofile/opentracing-smallrye/main/module.xml
```diff
@@ -25,6 +25,9 @@
     </resources>
 
     <dependencies>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.dmr"/>
```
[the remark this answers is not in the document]
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentDefinition.java
```diff
@@ -0,0 +1,53 @@
...
+            .build();
+
+    private TracingDeploymentDefinition() {
+          super(new Parameters(SubsystemExtension.SUBSYSTEM_PATH, SubsystemExtension.getResourceDescriptionResolver())
+                  .setFeature(false));
+    }
+
+    @Override
+    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+        super.registerAttributes(resourceRegistration);
```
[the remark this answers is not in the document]
ehsavoie: I need the name of the referenced tracer in the deployment processor

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/SubsystemDefinition.java
ehsavoie: This would be a Resource capability instead of one bounded to the value of an attribute

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerTracerConfigurationAddHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        String outboundSocketBindingName = TracerAttributes.SENDER_BINDING.resolveModelAttribute(context, model).asStringOrNull();
+        Supplier<OutboundSocketBinding> outboundSocketBindingSupplier;
+        if (outboundSocketBindingName != null) {
+             outboundSocketBindingSupplier = builder.requiresCapability(OUTBOUND_SOCKET_BINDING_CAPABILITY_NAME, OutboundSocketBinding.class, outboundSocketBindingName);
+        } else {
+            outboundSocketBindingSupplier = () -> null;
+        }
+        TracerConfiguration config = new JaegerTracerConfiguration(context, operation, outboundSocketBindingSupplier);
+        builder.setInstance(Service.newInstance(WildFlyTracerFactory.registerTracer(config.getName()), config));
+        builder.install();
```
pferraro: Services should be installed ON_DEMAND whenever possible.  Is there specific reason for this to be ACTIVE?  It seems to me that it should only start when a deployment needs it.

pferraro (changes requested):

zhantemirov:
Additional test cases have been added to the PR and combined with existing ones. 
From QE point of view the test side looks good, thanks.

darranl:
Just adding the missing_reqs label as community docs need to be ready at the same time as the feature.

darranl:
@ehsavoie Did your latest push address everything in the review from @pferraro - if so I can dismiss the review.

ehsavoie:
@darranl everything except the parser

zhantemirov:
@ehsavoie I can't see tests presented in https://github.com/ehsavoie/wildfly/pull/7/ PR.

ehsavoie:
@zhantemirov sorry, fixed

jamezp:
I've merged #12878 which requires a rebase of this PR.

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracerConfigurationConstants.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static final String DEFAULT_TRACER = "default-tracer";
+    public static final String PROPAGATION = "propagation";
+
+    public static final String SAMPLER_TYPE = "sampler-type";
+    public static final String SAMPLER_PARAM = "sampler-param";
+    public static final String SAMPLER_MANAGER_HOST_PORT = "sampler-manager-host-port";
+
+    public static final String SENDER_AGENT_BINDING = "sender-binding";
+    public static final String SENDER_AGENT_HOST = "sender-host";
```
bstansberry: This is no longer used.

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracerConfigurationConstants.java
```diff
@@ -0,0 +1,52 @@
...
+    public static final String SENDER_AGENT_PORT = "sender-port";
```
bstansberry: This is no longer used.

bstansberry (changes requested):
Hi @ehsavoie -- last round of change requests, I hope. ;)

If these could be in by my Monday morning that would be great as I'd like to get it merged as early as possible Monday.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDependencyProcessor.java
```diff
@@ -30,17 +32,64 @@
...
+        JBossWebMetaData jbossWebMetaData = getJBossWebMetaData(deploymentUnit);
+        if (null == jbossWebMetaData || null == jbossWebMetaData.getContextParams()) {
+            return null;
+        }
+        for (ParamValueMetaData param : jbossWebMetaData.getContextParams()) {
+            if (SMALLRYE_OPENTRACING_TRACER_CONFIGURATION.equals(param.getParamName())) {
+                if (param.getParamValue() != null && !param.getParamValue().startsWith(TRACER_CAPABILITY_NAME)) {
+                    return TRACER_CAPABILITY.getCapabilityServiceName(param.getParamValue());
+                }
+                return ServiceName.parse(param.getParamValue());
```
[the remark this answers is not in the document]
ehsavoie: I agree

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/WildFlyTracerFactory.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public static String getDefaultTracerName() {
+        TracerConfiguration configuration = DEFAULT_CONFIGURATION.get();
+        if (configuration != null) {
+            return configuration.getName();
+        }
+        return null;
+    }
+
+    public static void registerCompatibilityTracer(TracerConfiguration config) {
```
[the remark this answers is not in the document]
ehsavoie: ok

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
```diff
@@ -137,10 +125,51 @@ private String getServiceName(DeploymentUnit deploymentUnit) {
...
+        JBossWebMetaData jbossWebMetaData = getJBossWebMetaData(deploymentUnit);
+        if (null == jbossWebMetaData || null == jbossWebMetaData.getContextParams()) {
+            return null;
+        }
+        for (ParamValueMetaData param : jbossWebMetaData.getContextParams()) {
+            if (SMALLRYE_OPENTRACING_TRACER_CONFIGURATION.equals(param.getParamName())) {
+                if (param.getParamValue() != null && !param.getParamValue().startsWith(TRACER_CAPABILITY_NAME)) {
+                    return TRACER_CAPABILITY.getDynamicName(param.getParamValue());
+                }
+                return param.getParamValue();
```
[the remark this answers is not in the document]
ehsavoie: ok

bstansberry (approved):

PR #12837, opened 2019-12-06
WFLY-12855 Fix UnsupportedOperationException on undeploy

pferraro:
... due to change in AttachmentList backing collection type introduced by WFCORE-3948

https://issues.redhat.com/browse/WFLY-12855


rhusar (approved):

bstansberry:
@pferraro I'm going to merge this but we can revert it once the next WF Core beta is out. I'm going to revert the WFCORE-3948 fix and replace it with a SynchronizedList-backed AttachmentList. The one DUP where it's known we need to deal with concurrent writes while iterating, that DUP will synchronize on the list while iterating. Any other bugs we find with the same problem (only found one so far in 10 years) can be also synchronize while iterating.

I don't want external projects that build on WF to need to do what you did here.

PR #12838, opened 2019-12-07
Run clustering tests against Galleon-slimmed servers

pferraro:
An attempt to fix test failures seen in https://github.com/wildfly/wildfly/pull/12678


pferraro:
miserable failure.  closing.

PR #12842, opened 2019-12-09
WFLY-12588 BeanManager.findBean(...) can return an expired bean

pferraro:
https://issues.redhat.com/browse/WFLY-12588

### DIFF
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/bean/InfinispanBean.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/bean/InfinispanBean.java
@@ -75,7 +75,7 @@ public class InfinispanBean<I, T> implements Bean<I, T> {
 
     @Override
     public boolean isExpired() {
-        return this.entry.isExpired(this.timeout);
+        return this.entry.isExpired(this.timeout) && this.group.isCloseable();
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/bean/InfinispanBean.java
```diff
@@ -75,7 +75,7 @@ public I getGroupId() {
 
     @Override
     public boolean isExpired() {
```
bstansberry: @pferraro Am I correct in that before this PR this method was never used outside a test case?

I ask because I don't understand the implications of L78 but TBH if Bean.isExpired was never used you can make the semantics what you like and it doesn't matter if I understand. :)  (The L78 change doesn't seem wrong, I just don't pretend to deep understanding of the relationship to group.isCloseable vs bean.isExpired.)

The rest of this PR all seems fine.

pferraro: References to Bean.isExpired() were replaced by BeanEntry.isExpired(...) at some point, which meant only having to look up 1 cache entry to determine if a bean is expired.  However, a bean should *not* be considered expired if it is still referenced (i.e. Bean.isCloseable() returns false).  A transaction spanning invocations is a prime example of a bean that is not yet closeable.

bstansberry (approved):

bstansberry:
Thanks @pferraro !

PR #12844, opened 2019-12-10
[WFLY-12839] Prepare tests for allowing addition of HTTP headers to management console responses

tterem:
Jira: https://issues.redhat.com/browse/WFLY-12839

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/console/HttpManagementConstantHeadersTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/console/HttpManagementConstantHeadersTestCase.java
@@ -1,0 +1,328 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/console/HttpManagementConstantHeadersTestCase.java
```diff
@@ -0,0 +1,328 @@
...
+   @Test
+   public void testHeadersOverride() throws Exception {
+      HttpGet get = new HttpGet(managementConsoleUrl.toURI().toString());
+      HttpResponse response = httpClient.execute(get);
+      assertEquals(200, response.getStatusLine().getStatusCode());
+
+      Header[] headerArray = response.getAllHeaders();
+
+      List<Header> headerList = Arrays.stream(headerArray)
+            .filter(header -> !header.getName().equals("Connection") && !header.getName().equals("Date")
```
darranl: Probably not the end of the world as this is in a test but FYI streaming an array that could have been iterated in a for loop is very expensive, a lot of intermediate objects needs to be created and garbage collected to do something that could have been performed with minimal objects.

Don't change the PR for this as I am looking to merge but more a FYI.

darranl (approved):

PR #12845, opened 2019-12-11
WFLY-12863 microprofile-health subsystem should not expose smallrye implementation to deployments

pferraro:
https://issues.redhat.com/browse/WFLY-12863


pferraro:
Ah crap - I accidentally overwrote this branch.

PR #12846, opened 2019-12-12
WFLY-12871 - System Exception (EJBException) should be thrown instea…

panossot:
System Exception (EJBException) should be thrown instead of ApplicationException when rollback=false

https://issues.redhat.com/browse/WFLY-12871

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
@@ -115,6 +115,7 @@ public class CMTTxInterceptor implements Interceptor {
         } catch (HeuristicMixedException | SystemException | HeuristicRollbackException e) {
             throw new EJBException(e);
         }
+
     }
 
     private void endTransaction(final Transaction tx, final Throwable outerEx) {
@@ -127,6 +128,7 @@ public class CMTTxInterceptor implements Interceptor {
 
     public Object processInvocation(InterceptorContext invocation) throws Exception {
         final EJBComponent component = (EJBComponent) invocation.getPrivateData(Component.class);
+
         final ContextTransactionManager tm = ContextTransactionManager.getInstance();
         final int oldTimeout = tm.getTransactionTimeout();
         try {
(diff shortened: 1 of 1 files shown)


panossot:
Retest this please

darranl:
@tadamski / @chengfang Can you please give this PR a quick review?

bstansberry:
@chengfang / @tadamski Please review.

@panossot Please adjust the whitespace -- space by curly brackets, between 'if' and '(', between operands and operators.

ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
```diff
@@ -243,7 +246,13 @@ protected Object invokeInOurTx(InterceptorContext invocation, final EJBComponent
                 } catch (EJBException | RemoteException e) {
                     throw e;
                 } catch (RuntimeException e) {
-                    throw ae != null ? e : new EJBException(e);
+                    if(ae != null && !ae.isRollback()) {
+                        except = e;
```
chengfang: do we need to wrap it as EJBException(e)?

panossot: In case of the tx fail an EJBTransactionRolledbackException will be thrown which is an EJBException. (https://github.com/jboss-set/eap-additional-testsuite/blob/master/modules/src/main/java/org/jboss/additional/testsuite/jdkall/present/ejb/exceptions/ClientExceptionTestCase.java#L50). I think in the other cases it should remain a RuntimeException.

panossot:
Whitespace adjustments are done.

ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
chengfang: if it is `EJBTransactionRolledbackException` and hence `EJBException`, it will already been handled by the `catch (EJBException | RemoteException)` above this clause.  So here it will be some non-EJBExcpetion RuntimeException.

panossot: The RuntimeException at this point is the Exception that is thrown by MyApplicationExceptionNoRollbackRuntimeException (https://github.com/jboss-set/eap-additional-testsuite/blob/master/modules/src/main/java/org/jboss/additional/testsuite/jdkall/present/ejb/exceptions/MyApplicationExceptionNoRollbackRuntimeException.java#L12).
The EJBTransactionRolledbackException is thrown because of the RuntimeException here : https://github.com/jboss-set/eap-additional-testsuite/blob/master/modules/src/main/java/org/jboss/additional/testsuite/jdkall/present/ejb/exceptions/HelloBean.java#L41

chengfang (approved):

PR #12847, opened 2019-12-12
[WFLY-12859]: Acceptor is open after broker starts but before queues are created resulting in QUEUE_DOES_NOT_EXIST message=AMQ229017 (the queue is in the standalone.xml file)

ehsavoie:
* JMS queues and topics are added to the broker configuration before it
starts so they are created before being available in WildFly
* Behaviour is now the same as in Artemis thus the acceptro starts after
all the queues are available

Jira: https://issues.redhat.com/browse/WFLY-12859

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
@@ -185,7 +185,7 @@ class ServerAdd extends AbstractAddStepHandler {
     private static final String ARTEMIS_BROKER_CONFIG_JDBC_LOCK_ACQUISITION_TIMEOUT_MILLIS = "brokerconfig.storeConfiguration.jdbcLockAcquisitionTimeoutMillis";
 
     private ServerAdd() {
-        super(ACTIVEMQ_SERVER_CAPABILITY, ServerDefinition.ATTRIBUTES);
+        super(ServerDefinition.ATTRIBUTES);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
```diff
@@ -456,15 +456,23 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
-                Supplier<ActiveMQServer> serverSupplier = sb.requires(activeMQServiceName);
-                final QueueService queueService = new QueueService(serverSupplier, queueConfiguration, false, false);
-                sb.setInitialMode(Mode.PASSIVE);
-                sb.setInstance(queueService);
-                sb.install();
+            if (model.hasDefined(CommonAttributes.QUEUE)) {
+                Set<String> coreQueues = new HashSet<>();
+                for (Property prop : model.get(CommonAttributes.QUEUE).asPropertyList()) {
+                    coreQueues.add(prop.getName());
+                }
```
bstansberry: You don't need to copy out the names; you can just do 

ModelNode coreQueues = model.get(CommonAttributes.QUEUE);

And then 

if (coreQueues.has(queueConfiguration.getName())) {

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
```diff
@@ -456,15 +456,23 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
+                for (CoreQueueConfiguration queueConfiguration : configuration.getQueueConfigurations()) {
+                    if (coreQueues.contains(queueConfiguration.getName())) {
```
bstansberry: I assume the QueueService for other core queues not defined in the model is pointless.

bstansberry (approved):

PR #12864, opened 2019-12-20
WFLY-12908 MicroProfile ConfigProviderResolver static reference should be set eagerly

pferraro:
https://issues.redhat.com/browse/WFLY-12908

And other miscellaneous optimizations, including:
* Removing redundant no-op service implementation classes
* Optimize Config building DUP, removing iteration over all services in ServiceRegistry.


jmesnil (approved):

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/DirConfigSourceService.java
```diff
@@ -53,22 +54,23 @@
     private final int ordinal;
     private final Supplier<PathManager> pathManager;
 
-    private ConfigSource configSource;
+    private final Map<String, ConfigSource> sources;
```
jmesnil: sorry, I approved the PR as I tried to ack/nack individual commits :)

You are right, that iterating over all services is not a bright idea... 
I'm not sure that a DirConfigSourceService should hold references to *all* config sources. This service should be responsible to create a new `DirConfigSource` when it is started but keeping a references to all `sources` is fishy.
It seems incorrect to me that a `Service` that is used to "serve" a DirConfigSource does not hold the value (either by the deprecated `getValue()` method or with a new `Supplier`).

pferraro: > I'm not sure that a DirConfigSourceService should hold references to all config sources. This service should be responsible to create a new DirConfigSource when it is started but keeping a references to all sources is fishy.

The general idea is that DirConfigSourceService registers its ConfigSource on start and unregisters it on stop.  We can abstract this behind some kind of registration interface if you'd prefer, but in the end, all its doing is a map put on start, and a map remove on stop.

> It seems incorrect to me that a Service that is used to "serve" a DirConfigSource does not hold the value (either by the deprecated getValue() method or with a new Supplier).

There is no such MSC requirement that a Service needs to supply a value.

pferraro: OK - this stuff should be more intuitive now.

jmesnil (approved):

jmesnil:
@pferraro thanks!

PR #12871, opened 2019-12-30
[WFLY-12914] Upgrade Galleon to match WildFly Core

bstansberry:
https://issues.redhat.com/browse/WFLY-12914


bstansberry:
I put 'hold' on this because I'm working another commit to simplify config, but if this passes CI I may just go with this and get it in and use a different issue for the config stuff.

PR #12872, opened 2019-12-31
[WFLY-12915] Clean out unneeded provisioning boilerplate

bstansberry:
https://issues.redhat.com/browse/WFLY-12915

PR #12874, opened 2020-01-03
[WFLY-12826] switch to GAV-based locations for feature-pack dependencies

aloubyansky:
And as a consequence remove redundant transitive feature-pack configs from the provisioning configs
Also related to WFLY-12915 (cleaning up provisioning boilerplate configs).

https://issues.redhat.com/browse/WFLY-12826

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -82,20 +82,6 @@
                             <offline>true</offline>
                             <feature-packs>
                                 <feature-pack>
-                                    <transitive>true</transitive>
-                                    <groupId>org.wildfly.core</groupId>
-                                    <artifactId>wildfly-core-galleon-pack</artifactId>
-                                    <version>${version.org.wildfly.core}</version>
-                                    <inheritConfigs>false</inheritConfigs>
-                                </feature-pack>
-                                <feature-pack>
-                                    <transitive>true</transitive>
-                                    <groupId>${project.groupId}</groupId>
-                                    <artifactId>wildfly-servlet-galleon-pack</artifactId>
-                                    <version>${project.version}</version>
-                                    <inheritConfigs>false</inheritConfigs>
-                                </feature-pack>
-                                <feature-pack>
                                     <groupId>${project.groupId}</groupId>
                                     <artifactId>wildfly-galleon-pack</artifactId>
                                     <version>${project.version}</version>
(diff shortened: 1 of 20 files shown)


aloubyansky:
FYI @jfdenise and @bstansberry

aloubyansky:
BTW, I couldn't see where the following props are used. https://github.com/wildfly/wildfly/blob/master/pom.xml#L208-L209

jfdenise:
They are used there: https://github.com/wildfly/wildfly/blob/master/dist/pom.xml#L224
BTW, the producer should be upgraded to 1.1.1.Final.

aloubyansky:
Ok, thanks.

kabir:
Not related to this PR, but @jfdenise I need to understand this a bit better, since I believe I can fix up https://github.com/wildfly/wildfly-galleon-pack-template a bit along these lines?

aloubyansky:
@kabir right, you can try an equivalent change there.
Changing the schema to `3.1` here https://github.com/wildfly/wildfly-galleon-pack-template/blob/master/galleon-pack/wildfly-feature-pack-build.xml#L20 will generate a `feature-pack.xml` with feature-pack dependencies using GAV format instead of Galleon's FPL.
After that you can remove transitive feature-packs here https://github.com/wildfly/wildfly-galleon-pack-template/blob/master/testsuite/integration/subsystem/pom.xml#L79
Galleon and WFGP versions should of course be updated here https://github.com/wildfly/wildfly-galleon-pack-template/blob/master/pom.xml#L60

bstansberry:
@aloubyansky Does using the 3.1 schema eliminate the need for the 2nd commit from my #12872 PR?

I think I'll try it out and see. :) . To reproduce my problem I just remove any universe or producer artifact from my maven repo and try and build.

aloubyansky:
Yes, it should

aloubyansky:
I just released Galleon 4.2.2.Final that fixes an issue for Kabir's use-case. I am going to update this PR to use it.

bstansberry:
I confirmed this eliminates the problem I saw when doing #12872 :)

bstansberry (approved):

bstansberry:
Thanks @aloubyansky !

PR #12877, opened 2020-01-06
WFLY-4699 Expose EJB3 deployment information at runtime

chengfang:
JIRA:
https://issues.redhat.com/browse/WFLY-4699
https://issues.redhat.com/browse/EAP7-435

Analysis Document for this RFE:
https://github.com/wildfly/wildfly-proposals/pull/271

This PR implements the above RFE to expose ejb deployment info as management runtime resources. These deployment info can come from ejb deployment descriptor and annotations. This PR includes implementation changes, tests, and community docs.


tadamski (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/management/deployments/EjbJarRuntimeResourceTestBase.java
```diff
@@ -72,8 +75,36 @@
...
+    private static final String AFTER_COMPLETION_METHOD = "after-completion-method";
+    private static final String PASSIVATION_CAPABLE = "passivation-capable";
+    private static final String REMOVE_METHODS = "remove-methods";
+    private static final String RETAIN_IF_EXCEPTION = "retain-if-exception";
+    private static final String BEAN_METHOD = "bean-method";
+
+    // SINGLETON specific resource names
+    private static final String DEPENDS_ON = "depends-on";
+    private static final String INIT_ON_STARTUP = "init-on-startup";
+    private static final String CONCURRENCY_MANAGEMENT_TYPE = "concurrency-management-type";
```
tommaso-borgato, line 108: @chengfang I can't find tests related to `Bean statistics (beans in use, access count, idle count, ...)`, can you point me at them?

tommaso-borgato (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/management/deployments/EjbJarRuntimeResourceTestBase.java
chengfang: @tommaso-borgato  these bean stats are existing features (already there before we implement this feature), so they are not covered by this PR.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/AbstractEJBComponentResourceDefinition.java
```diff
@@ -51,11 +51,35 @@
  */
 public abstract class AbstractEJBComponentResourceDefinition extends SimpleResourceDefinition {
 
-    public static final SimpleAttributeDefinition COMPONENT_CLASS_NAME = new SimpleAttributeDefinitionBuilder("component-class-name", ModelType.STRING)
+     static final SimpleAttributeDefinition COMPONENT_CLASS_NAME = new SimpleAttributeDefinitionBuilder("component-class-name", ModelType.STRING)
             .setValidator(new StringLengthValidator(1))
             .setFlags(AttributeAccess.Flag.STORAGE_RUNTIME)
             .build();
 
+     static final SimpleAttributeDefinition JNDI_NAMES = new SimpleAttributeDefinitionBuilder("jndi-names", ModelType.LIST)
```
bstansberry, line 59: @chengfang For all the various LIST attributes you should used one of the ListAttributeDefinition impls. My guess is all of them are essentially List<String>, so use StringListAttributeDefinition. The pattern would be

```
static final SimpleAttributeDefinition JNDI_NAMES = StringListAttributeDefinition.Builder.of("jndi-names")     
```

followed by what you've got now.

This results in the management API description having the correct 'value-type' as discussed in https://docs.wildfly.org/19.1/Admin_Guide.html#description-of-an-attribute

bstansberry (changes requested):
Small change to the AttributeDefintion impls aside, looks good.

bstansberry (approved):

bstansberry:
@chengfang I've merged this but the test failure at https://ci.wildfly.org/viewLog.html?buildId=206002&tab=buildResultsDiv&buildTypeId=WF_PullRequest_LinuxJdk11&branch_WF_PullRequest=pull%2F13307 has happened a couple times in association with this PR and I don't believe otherwise, so that's something to look into.

chengfang:
@bstansberry this intermittent failure was tracked by the following JIRA, which was first reported on 05/Apr/2018. So this test bug exists before my PR, but my PR might have triggered it more often. I'll take a look.

[WFLY-10173](https://issues.redhat.com/browse/WFLY-10173) EjbInvocationStatisticsTestCase fails on Windows: wait-time=0

bstansberry:
@chengfang Thanks. I feel better that it happened before. :) I didn't see why this PR would affect the actual value of the "wait-time" stat which is why I went ahead and merged it.

PR #12878, opened 2020-01-06
[WFLY-12486]:  Memory leak in OpenTracing when deployment is redeploy…

ehsavoie:
…ed multiple times.

**To Be Considered ONLY IF** https://github.com/wildfly/wildfly/pull/12836 doesn't make it in WF19.
This is why it has the 'hold' label

* Removing the use of a ContextListener to close the tracers
* Using deployment processor to create and inject the tracer and properly close it on undeploy.

Jira: https://issues.redhat.com/browse/WFLY-12486


ehsavoie:
**To Be Considered ONLY IF ** https://github.com/wildfly/wildfly/pull/12836 doesn't make it in WF19.

PR #12883, opened 2020-01-07
[WFLY-12473] Add instance id obfuscation in Undertow session route

fl4via:
RFE: https://issues.redhat.com/browse/WFLY-12473
Analysis: https://github.com/wildfly/wildfly-proposals/pull/246

Plus:
https://issues.redhat.com/browse/WFLY-13858
https://issues.redhat.com/browse/WFLY-12932
https://issues.redhat.com/browse/WFLY-12933


undertow/src/test/java/org/wildfly/extension/undertow/UndertowServiceTestCase.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Test
+    public void testUuidInstanceId() throws Exception {
+        final UndertowService undertowService = load("undertow-service-uuid-instance-id-attribute.xml");
+        assertNotNull(UUID.fromString(undertowService.getInstanceId()));
+        assertFalse(undertowService.isStatisticsEnabled());
+        assertEquals("undertow-server", undertowService.getDefaultServer());
+        assertEquals(DEFAULT_SERVLET_CONTAINER, undertowService.getDefaultContainer());
+        assertEquals("virtual-host3", undertowService.getDefaultVirtualHost());
+    }
```
tommaso-borgato, line 139: @fl4via looking at default `standalone-ha.xml` shipped e.g. with EAP 7.3 I see that element `application-security-domains` is not defined by default; in all the files undertow-service-*.xml used in the tests that element is always defined: would it make sense to add one more tests with no `application-security-domains`?

tommaso-borgato (changes requested):

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow.adoc
```diff
@@ -43,6 +41,18 @@ for most use cases and provides reasonable performance settings.
...
     </subsystem>
 ----
+[cols=",",options="header"]
+|=======================================================================
+|Attribute |Description
+|default-server | the default server to use for deployments
+|default-virtual-host | the default virtual host to use for deployments
+|default-servlet-container | the default servlet container to use for deployments
+|instance-id | the id of Undertow. Defaults to "${jboss.node.name}"
+|uuid-instance-id | set this to true to indicate that a UUID should be used as the instance id. Can be used only when instance-id is undefined
```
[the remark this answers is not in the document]
fl4via: @tommaso-borgato  yes, that's the case, as lon as uuid-instance-id is undefined as well. I'll improve the documentation to make this clear and resubmit the PR

undertow/src/test/java/org/wildfly/extension/undertow/UndertowServiceTestCase.java
fl4via: certainly, @tommaso-borgato , I'll make that change

fl4via:
@tommaso-borgato  based on the feedback I got at Undertow team discussions, I'm proposing a small change: to allow to have both instance-id and uuid-instance-id defined. So, it will make clearer the idea from the previous commit that the uuid is generated from ${jboss.node.name}, to keep the final value of instance-id consistent across server restarts, and it allows to have users define a different string to have that uuid based from. If both instance-id and uuid-instance-id are defined, the uuid will be generated from the instance-id value. Please, have a look at the commit I added and let me know if QE is okay with that change on such a short notice.
If I get your okay, I'll squash the proposed change commit into previous commit. If I don't, I'll create a new RFE Jira to have this change incorporated in a future release.

fl4via:
I've just updated the wildfly-proposal document with this proposed change as well: https://github.com/wildfly/wildfly-proposals/pull/246

tommaso-borgato:
@fl4via as agreed, the change is refused as part of this RFE and will be added to a follow-up EAP7

fl4via:
@tommaso-borgato thanks for your feedback. This PR is updated and is now final, ready for review

undertow/src/main/java/org/wildfly/extension/undertow/Constants.java
```diff
@@ -46,6 +46,7 @@
     String HTTP_INVOKER = "http-invoker";
     String LISTENER = "listener";
     String INSTANCE_ID = "instance-id";
+    String UUID_INSTANCE_ID = "uuid-instance-id";
```
pferraro, line 49: IMO, this attribute name is no longer sufficiently descriptive, since, when enabled, the generated instance-id has nothing to do with a UUID.
Perhaps "obfuscate-instance-id"?

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
```diff
@@ -118,13 +120,22 @@ private static void registerTransformersWildFly18(ResourceTransformationDescript
     }
 
     private static void registerTransformersWildFly16(ResourceTransformationDescriptionBuilder subsystemBuilder) {
+        subsystemBuilder.getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, UUID_INSTANCE_ID)
+                .setDiscard(DiscardAttributeChecker.ALWAYS, UUID_INSTANCE_ID)
```
pferraro, line 125: I think we want to use DiscardAttributeChecker.UNDEFINED here (and for other version transformations), since value of true should result in rejection.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemAdd.java
```diff
@@ -102,8 +105,9 @@ protected void performBoottime(OperationContext context, ModelNode operation, Re
         final String defaultSecurityDomain = UndertowRootDefinition.DEFAULT_SECURITY_DOMAIN.resolveModelAttribute(context, model).asString();
 
         final ModelNode instanceIdModel = UndertowRootDefinition.INSTANCE_ID.resolveModelAttribute(context, model);
-        final String instanceId = instanceIdModel.isDefined() ? instanceIdModel.asString() : null;
-
+        final ModelNode uuidInstanceIdModel = UndertowRootDefinition.UUID_INSTANCE_ID.resolveModelAttribute(context, model);
+        final String instanceId = instanceIdModel.isDefined() ? (uuidInstanceIdModel.asBoolean() ?
```
pferraro, line 109: IMO, this logic actually belongs in org.wildfly.extension.undertow.Server#getRoute().

The getRoute() method should obfuscate the entire route (which may also include the server name) when the obfuscation flag is enabled.

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowRootDefinition.java
```diff
@@ -87,6 +87,14 @@
...
                     .setAllowExpression(true)
                     .setDefaultValue(new ModelNode().set(new ValueExpression("${jboss.node.name}")))
+                    .setAlternatives(Constants.UUID_INSTANCE_ID)
+                    .build();
+    protected static final SimpleAttributeDefinition UUID_INSTANCE_ID =
+            new SimpleAttributeDefinitionBuilder(Constants.UUID_INSTANCE_ID, ModelType.BOOLEAN, true)
+                    .setRestartAllServices()
+                    .setAllowExpression(true)
+                    .setDefaultValue(ModelNode.FALSE)
+                    .setAlternatives(Constants.INSTANCE_ID)
```
pferraro, line 97: These 2 instances are not incompatible, thus should not utilize setAlternatives(...).

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemAdd.java
```diff
@@ -102,8 +105,9 @@ protected void performBoottime(OperationContext context, ModelNode operation, Re
...
+                new String(Base64.getUrlEncoder().encode(instanceIdModel.asBytes())) : instanceIdModel.asString()) : null;
```
msfm, line 110: This just generates a base64 encoded string from the `instanceIdModel` value. A base64 encoded string can be easily decoded to the original value (= hostname by default). For example, the original value can decode it by the following command:

```
$ echo -n <generate-encoded-value> | base64 -d
```

So, as this approach can not hide the actual hostname information, I think this is not what the customer wants. 

I would suggest generating a base64 encoded string from a MD5 hash digest of the `instanceIdModel` value. Java's name-based UUID (`UUID.nameUUIDFromBytes()`) is internally generated from MD5 hash digest of the input value. So, I think this approach is the same level of uniqueness to `UUID.nameUUIDFromBytes()`.

Also, `Base64.Encoder` generates an encoded string with a padding (`=` by default). However, `=` is not allowed in V0 Cookie. So, I think it's better to use `Base64.getUrlEncoder().withoutPadding()` instead.

For example:

```
        final String instanceId = instanceIdModel.isDefined() ?
            (uuidInstanceIdModel.asBoolean() ? generateObfuscatedInstanceId(instanceIdModel.asString()) : instanceIdModel.asString()): null;

    ...(snip)...

    private String generateObfuscatedInstanceId(String instanceId) {
        try {
            MessageDigest md = MessageDigest.getInstance("MD5");
            byte[] md5Bytes = md.digest(instanceId.getBytes());
            Base64.Encoder encoder = Base64.getUrlEncoder().withoutPadding(); // = is not allowed in V0 Cookie
            String encodedString = encoder.encodeToString(md5Bytes);
            UndertowLogger.ROOT_LOGGER.infof("Obfuscated value [%s] is generated from [%s] and will be used as instance-id (jvmRoute)", encodedString, instanceId);
            return encodedString;
        } catch (NoSuchAlgorithmException e) {
            throw new IllegalArgumentException(e);
        }
    }
```

darranl: This may take the problem full circle but are they concerned about these values being mapped to known host names?  If so maybe some additional secret data is needed in the hash to prevent comparison against known host names.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowRootDefinition.java
fl4via: pferraro it came as a request from the feature team because they did not want to have this changed this late at the game. However, as I pointed, now that we're renaming, it makes less and less sense to forbid instance-id definition when obfuscated-instance-id is used.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemAdd.java
fl4via: @darranl  thanks for your feedback. I'll ask if that's a concern or not. is that something complex to do, or something super simple? The question here is, if the additional secret data is interesting to the user, should it be a follow up RFE or should we add it to this RFE? Please let me know what you think about this.

pferraro: @darranl is referring to a salt.  I think it is sufficient to use the Undertow server name as the salt, as the route identifier (which is the value we are actually interested in) is generated per Undertow server.  This is sufficient to prevent comparison against known host names.

darranl: +1 provided that server name is not shared publicly it should be sufficient in this case.

Where only public information goes into a hash it is very easy for someone to create their own reference tables in a kind of reverse lookup.  Once some non public information is added it becomes a lot more complex.

The piece I am not familiar with here is the full use of these fields in case my suggestion made it much more complex where we need to be able to map back.

pferraro: The name of an Undertow server resource is effectively arbitrary, and has no intrinsic meaning outside of the subsystem model - hence my suggestion to use it as a salt (since the name of an Undertow server resource is unique per server).

Both the instance-id and the names of Undertow servers stored as clear text in the model, so there should be no need/reason to map back.

darranl: @pferraro that does sound like enough then.

If anyone wanted more than that we would likely need to be looking at stronger approaches, possibly with keys etc...

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
fl4via: I assumed that those checkers were applied in sequence, i.e., fist if attribute is defined, it would be rejected. If not, apply next on the list: discard always, which would imply discard undefined. I'm updating the PR with DiscardAttributeChecker.UNDEFINED

undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
```diff
@@ -27,7 +27,8 @@ undertow.configuration=null
 undertow.default-virtual-host=The default virtual host to use for deployments
 undertow.default-servlet-container=The default servlet container to use for deployments
 undertow.default-server=The default server to use for deployments
-undertow.instance-id=The cluster instance id
+undertow.instance-id=The cluster instance id (defaults to {$jboss.node.name} if undefined)
+undertow.obfuscate-instance-id=Obfuscate the instance-id and server name when routing
```
fl4via, line 31: @pferraro  now, the question that remains is it a problem to call it obfuscate-instance-id when we are actually obfuscating the whole route? If it is, would 'obfuscate-route' attribute name be easy enough for the final user to understand the purpose of the attribute? Not obfuscating the server-name would defeat the usage of it as a salt.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
```diff
@@ -118,13 +120,22 @@ private static void registerTransformersWildFly18(ResourceTransformationDescript
     }
 
     private static void registerTransformersWildFly16(ResourceTransformationDescriptionBuilder subsystemBuilder) {
+        subsystemBuilder.getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED,OBFUSCATE_INSTANCE_ID)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, OBFUSCATE_INSTANCE_ID)
```
pferraro, line 125: Oops - my previous comment was incomplete. :(
We also want to discard if the attribute is defined, but set to the default value. i.e.
```java
.setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, OBFUSCATE_INSTANCE_ID);
```
Same for other versions.

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -169,9 +172,24 @@ public String getDefaultHost() {
...
-        String defaultServerRoute = service.getInstanceId();
-        return this.name.equals(service.getDefaultServer()) ? defaultServerRoute : String.join("-", defaultServerRoute, this.name);
+        final UndertowService service = this.undertowService.getValue();
+        final String defaultServerRoute = service.getInstanceId();
+        final String route = this.name.equals(service.getDefaultServer()) ? defaultServerRoute : String.join("-", defaultServerRoute, this.name);
+        if (service.isObfuscateInstanceId()) {
+            try {
+                final MessageDigest md = MessageDigest.getInstance("MD5");
+                // salt
+                md.update(this.name.getBytes());
```
pferraro, line 182: Given that we don't know the default character encoding of a given host, it's probably best to keep this consistent across all hosts, e.g.
```java
md.update(this.name.getBytes(StandardCharsets.UTF-8));
```

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
fl4via: This broke transformation tests, I see errors such as the one below:
`[ERROR]   UndertowTransformersTestCase.testTransformersEAP_7_1_0:72->testTransformers:284->testTransformers:302->AbstractSubsystemTest.checkSubsystemModelTransformation:206->AbstractSubsystemTest.checkSubsystemModelTransformation:220 : {
 expected:<...-host, instance-id, []server, servlet-cont...> but was:<...-host, instance-id, [obfuscate-instance-id, ]server, servlet-cont...>`

For that reason, I'm reverting this change, @pferraro

undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
fl4via: @pferraro  that makes more sense to me. I'll update the PR with this.

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -169,9 +172,24 @@ public String getDefaultHost() {
     }
 
     public String getRoute() {
-        UndertowService service = this.undertowService.getValue();
-        String defaultServerRoute = service.getInstanceId();
-        return this.name.equals(service.getDefaultServer()) ? defaultServerRoute : String.join("-", defaultServerRoute, this.name);
+        final UndertowService service = this.undertowService.getValue();
+        final String defaultServerRoute = service.getInstanceId();
+        final String route = this.name.equals(service.getDefaultServer()) ? defaultServerRoute : String.join("-", defaultServerRoute, this.name);
```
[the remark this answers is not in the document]
fl4via: ok, noted

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -169,9 +172,24 @@ public String getDefaultHost() {
...
+        if (service.isObfuscateInstanceId()) {
+            try {
+                final MessageDigest md = MessageDigest.getInstance("MD5");
+                // salt
+                md.update(this.name.getBytes());
+                // encode
+                final Base64.Encoder encoder = Base64.getUrlEncoder().withoutPadding(); // = is not allowed in V0 Cookie
+                final String encodedRoute = encoder.encodeToString(md.digest(route.getBytes()));
```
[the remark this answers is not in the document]
fl4via: @pferraro  good catch, I'll update the PR

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
fl4via: @pferraro  Oh, I see. Please disregard my other comment about reverting the PR because of the test failure. I'll add the discard instead :-)

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
fl4via: noted

fl4via:
I'll be updating this PR later to incorporate the more recent feedback

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -169,8 +174,23 @@ public String getDefaultHost() {
     }
 
     public String getRoute() {
-        UndertowService service = this.undertowService.getValue();
-        String defaultServerRoute = service.getInstanceId();
+        final UndertowService service = this.undertowService.getValue();
+        final String defaultServerRoute = service.getInstanceId();
+        if (service.isObfuscateSessionRoute()) {
+            try {
+                final MessageDigest md = MessageDigest.getInstance("SHA-256");
```
pferraro, line 181: Given that we are already salting the hash of the instance-id (usually the host name) using the server name, thus preventing mapping back to known host names, is SHA-256 really necessary here?
This results in route identifiers of 43 characters, as opposed to 22 characters when using MD5.

Perhaps we leave this determination up to the user, i.e. make this configurable?
This would replace obfuscate-session-route="true|false" with session-route-digest-algorithm="...", where undefined means we that do not obfuscate.  WDYT?

undertow/src/test/java/org/wildfly/extension/undertow/ServerServiceTestCase.java
```diff
@@ -0,0 +1,195 @@
+/*
```
tommaso-borgato, line 1: @fl4via I don't understand how this test is relevant in EAP7-1187

tommaso-borgato:
@fl4via I see tests covering EJB invocation; can you point me at those covering regular HTTP invocation? I mean when route is obfuscated ....

undertow/src/test/java/org/wildfly/extension/undertow/ServerServiceTestCase.java
```diff
@@ -0,0 +1,195 @@
...
+
+import org.jboss.as.subsystem.test.KernelServices;
+import org.jboss.as.subsystem.test.KernelServicesBuilder;
+import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceName;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import sun.nio.cs.UTF_8;
```
tommaso-borgato, line 39: @fl4via compilation fails: `import sun.nio.cs.UTF_8;`: is unused import

tommaso-borgato (changes requested):

undertow/src/test/java/org/wildfly/extension/undertow/ServerServiceTestCase.java
pferraro: She probably meant to use java.nio.charset.StandardCharsets.UTF_8.
IDE automation strikes again! :)

undertow/src/test/java/org/wildfly/extension/undertow/ServerServiceTestCase.java
pferraro: This is a unit test covering the new logic in org.wildfly.extension.undertow.Server, which is the core logic of EAP7-1187.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
```diff
@@ -118,13 +120,20 @@ private static void registerTransformersWildFly18(ResourceTransformationDescript
     }
 
     private static void registerTransformersWildFly16(ResourceTransformationDescriptionBuilder subsystemBuilder) {
+        subsystemBuilder.getAttributeBuilder().addRejectCheck(RejectAttributeChecker.DEFINED, OBFUSCATE_SESSION_ROUTE)
+                .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, OBFUSCATE_SESSION_ROUTE)
+                .end();
```
bstansberry, line 125: @fl4via I believe you need a new 'registerTransformersWildFly22 method (and a MODEL_VERSION_WILDFLY_22) and this code goes there.

It then shouldn't be necessary to repeat this logic in the other methods. You handle it in current->MODEL_VERSION_WILDFLY_22 and the chain transformation means its handled for all the earlier ones.

bstansberry (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
bstansberry: A code style nit is that it's nicer to do the setDiscard call before the addRejectCheck call. The order of these doesn't matter technically as the transformation engine does discards before rejects. But maintainers are less likely to know that and putting the code in the order they actually happen helps make it clearer what will happen.

I wouldn't mention this if it weren't for that fact that my comment above means this will be getting touched anyway.

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
pferraro: @fl4via To reiterate, we do not require a 32 byte hash algorithm for this.  MD5 is sufficient for the purpose of obfuscating the session route and is consistent with the requested behavior in the original feature request.
We can discuss making this configurable at a later date.

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
fl4via: Sure, @bstansberry  I'll update the PR

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
fl4via: @pferraro I'm updating the PR with this change

bstansberry (approved):

pferraro (approved):

PR #12884, opened 2020-01-07
[WFLY-11388] User is not informed about incorrect values in exposed-s…

gaol:
…ubsystems for MP Metrics

Issue: https://issues.redhat.com/browse/WFLY-11388

### DIFF
--- a/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
+++ b/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
@@ -22,19 +22,33 @@
 
 package org.wildfly.extension.microprofile.metrics;
 
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.PersistentResourceDefinition;
+import org.jboss.as.controller.ReloadRequiredWriteAttributeHandler;
 import org.jboss.as.controller.ServiceRemoveStepHandler;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.StringListAttributeDefinition;
 import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 import org.jboss.msc.service.ServiceName;
+import org.wildfly.extension.microprofile.metrics._private.MicroProfileMetricsLogger;
+
 
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2018 Red Hat inc.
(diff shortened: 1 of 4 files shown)


gaol:
@jmesnil Would you please review? Thanks

gaol:
retest this please

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
```diff
@@ -90,4 +102,46 @@ protected MicroProfileMetricsSubsystemDefinition() {
...
+                    exposedSubsystems.add(n.asString());
+                }
+                validExposedSubsystems(subsystems, exposedSubsystems);
+            }
+            super.finishModelStage(context, operation, attributeName, newValue, oldValue, model);
+        }
+    }
+
+    static void validExposedSubsystems(Set<String> subsystems, List<String> exposedSubsystems) throws OperationFailedException {
+        List<String> unkownSubsystems = new LinkedList<>();
```
yersan: One suggestion s/unkown/unknown

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
```diff
@@ -90,4 +102,46 @@ protected MicroProfileMetricsSubsystemDefinition() {
...
+    private class ExposedSubsystemWriterHandler extends ReloadRequiredWriteAttributeHandler {
+        private ExposedSubsystemWriterHandler() {
+            super(EXPOSED_SUBSYSTEMS);
+        }
+
+        @Override
+        protected void finishModelStage(final OperationContext context, final ModelNode operation, String attributeName,
+                                        ModelNode newValue, ModelNode oldValue, final Resource model) throws OperationFailedException {
+            if (newValue.isDefined()) {
+                Set<String> subsystems = context.readResourceFromRoot(EMPTY_ADDRESS).getChildrenNames("subsystem");
```
yersan: It would be better to just read the subsystem children and avoid read the model recursively. 

I suggest using `context.readResourceFromRoot(PathAddress.pathAddress(PathElement.pathElement(SUBSYSTEM)), false).getChildrenNames(SUBSYSTEM);` instead

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/metrics/base/MicroProfileMetricsBaseTestCase.java
```diff
@@ -76,4 +86,62 @@ public void testBaseMetrics() throws Exception {
...
+        }
+    }
+
+    @Test
+    public void testExposedSubsystem() throws Exception {
+        final ModelNode address = Operations.createAddress("subsystem", "microprofile-metrics-smallrye");
+        ModelNode op = Operations.createWriteAttributeOperation(address, "exposed-subsystems", new ModelNode().add("jca").add("undertow"));
+        ModelNode result = ManagementOperations.executeOperationRaw(managementClient.getControllerClient(), op);
+        assertEquals("success", result.get("outcome").asString());
+        ServerReload.executeReloadAndWaitForCompletion(managementClient.getControllerClient(), 50000);
```
yersan: These operations are deprecated, you should use the non-deprecated ones.

Additionally, if you really need to use a different timeout, the best option would be to use TimeoutUtil.adjust(...) wrapping your desired timeout value, which would make the test more robust for executions in slow machines.

e.g: `ServerReload.executeReloadAndWaitForCompletion(managementClient, TimeoutUtil.adjust(50000));`

yersan (changes requested):

gaol:
@yersan Thank you for the review, I have updated accordingly.

yersan (approved):
LGTM, thanks @gaol

gaol:
@yersan is this OK to get merged?

yersan:
Hello @gaol , I approved it some time ago, but now it seems it got stale. 

Would you mind rebasing and force push it again to see a clean CI result? Anyway, I bet it should be also reviewed by @jmesnil

gaol:
@yersan Thanks Yeray, :)  I rebased latest master branch. @jmesnil Jeff, would you please also review? Thanks :)

darranl:
Are there any implications for domain mode with this level of validation?

Could we end up in a situation where a subsystem is not present on an older host but the metrics subsystem could still be present listing all subsystems?

yersan (commented):
> Are there any implications for domain mode with this level of validation?

No, there aren't any implications because metrics is not enabled on Domain Mode.

> Could we end up in a situation where a subsystem is not present on an older host but the metrics subsystem could still be present listing all subsystems?

No, we cannot. The reason, above.

bstansberry:
Sorry, I don't want to do this.

PR #12888, opened 2020-01-08
[WFCORE-4788] Upgrade httpclient 4.5.10

jmesnil:
*Temporarily* add the httpclient dependency directly in the
integration test suite and fix the expectation of the
PreservePathTestCase assertion.

Once a release of WildFly Core that contains WFCORE-4788, the 2 added
dependencies will be removed from the pom.xml

JIRA: https://issues.redhat.com/browse/WFCORE-4788


jmesnil:
Once it is merged, https://github.com/wildfly/wildfly-core/pull/4050 can be retested.

bstansberry (approved):

PR #12889, opened 2020-01-08
[WFLY-12945] Refactoring undertow subsystem to use new MSC value API

ropalka:
https://issues.redhat.com/browse/WFLY-12945

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/Host.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/Host.java
@@ -39,7 +39,9 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
 import io.undertow.Handlers;
@@ -63,7 +65,6 @@ import org.jboss.msc.service.Service;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
-import org.jboss.msc.value.InjectedValue;
 import org.wildfly.extension.undertow.deployment.GateHandlerWrapper;
 import org.wildfly.extension.undertow.logging.UndertowLogger;
 import org.wildfly.security.manager.WildFlySecurityManager;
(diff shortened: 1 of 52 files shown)


darranl:
Just placing on hold for now, we have run into some error reporting issues with the new approach that may need looking into first.

ropalka:
Retest this please.

ropalka:
Please remove the hold flag @darranl the conflict have been fixed.

darranl:
@ropalka Sorry didn't see the other comment - FYI the hold label was not in relation to the conflicts it was added as we were running into issues over on WildFly Core where use of the new MSC API was dropping errors starting the server so was leading to failed start ups with no error reporting as to what the underlying problem was.

We were able to manually debug and fix the underlying issue but it was still a concern that the error to diagnose the error was no longer reported.

ropalka:
Could I reproduce it somehow @darranl ? I can't see any reported issue in MSC JIRA.

darranl:
@ropalka I think we can give this another go but for now I think avoid any changes to the JACC related services, those are already in a bit of complicated position as we need to backport some of them.

ropalka:
Are you aware of this PR @bstansberry @kabir ?

undertow/src/main/java/org/wildfly/extension/undertow/HttpListenerService.java
```diff
@@ -54,24 +55,21 @@
...
  */
 public class HttpListenerService extends ListenerService {
     private volatile AcceptingChannel<StreamConnection> server;
 
     private final ChannelUpgradeHandler httpUpgradeHandler = new ChannelUpgradeHandler();
-    protected final InjectedValue<ListenerRegistry> httpListenerRegistry = new InjectedValue<>();
-    /**
-     * @deprecated Replaced by HTTP_UPGRADE_REGISTRY.getCapabilityServiceName()
-     */
-    @Deprecated
```
bstansberry, line 66: This comment and annotation should not have been removed.

undertow/src/main/java/org/wildfly/extension/undertow/Host.java
```diff
@@ -132,13 +139,13 @@ private String getDeployedContextPath(DeploymentInfo deploymentInfo) {
 
     @Override
     public void start(StartContext context) throws StartException {
-        suspendControllerInjectedValue.getValue().registerActivity(suspendListener);
-        if(suspendControllerInjectedValue.getValue().getState() == SuspendController.State.RUNNING) {
+        suspendController.get().registerActivity(suspendListener);
+        if(suspendController.get().getState() == SuspendController.State.RUNNING) {
```
bstansberry, line 143: @jasondlee fyi ^^^ is going to result in conflicts

jasondlee: Noted. I'll keep an eye out so I can merge and resolve as soon as I can.

bstansberry:
@ropalka Can you fix the conflict and we'll see if we can get this done in the next few days? Thanks.

ropalka:
Fixed @bstansberry including your feedback about @Deprecated annotation

darranl:
Sorry @ropalka but looks like a bunch of conflicts again, I think we should try again as we are at the start of WildFly 24 development.

ropalka:
Fixed @darranl

darranl:
Thanks @ropalka now is the time to get this in before any further refactoring of the Undertow subsystem begins for WildFly 24

bstansberry:
Thanks, @ropalka -- sorry this took so long.

PR #12898, opened 2020-01-10
[WFLY-12957]: MP Fault tolerance tests are failing on Ubuntu

ehsavoie:
* Passing the environmùent variables with dots as server args.

Jira: https://issues.redhat.com/browse/WFLY-12957


pferraro:
@rhusar Please review.

rhusar (changes requested):
The idea is correct thanks Emmanuel – just replacing the environmental variable with parameters, *but* we should get rid of the unneeded definitions as well, opened https://github.com/wildfly/wildfly/pull/12915 to replace this PR.

pferraro:
Superseded by https://github.com/wildfly/wildfly/pull/12915

PR #12900, opened 2020-01-11
WFLY-12958 Calling Asynchronous EJB will use the propagated caller tr…

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-12958

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
@@ -133,21 +133,8 @@ public class CMTTxInterceptor implements Interceptor {
         try {
             final MethodIntf methodIntf = MethodIntfHelper.of(invocation);
             final Method method = invocation.getMethod();
-            TransactionAttributeType attr = component.getTransactionAttributeType(methodIntf, method);
+            final TransactionAttributeType attr = component.getTransactionAttributeType(methodIntf, method);
             final int timeoutInSeconds = component.getTransactionTimeout(methodIntf, method);
-
-            final ComponentView view = invocation.getPrivateData(ComponentView.class);
-            if (view != null && view.isAsynchronous(method)) {
-                // adjust for asynchronous methods. The client’s transaction context does not propagate with
-                // an asynchronous method invocation.
-                switch (attr) {
-                    case REQUIRED:
-                    case MANDATORY:
-                    case SUPPORTS:
-                        attr = TransactionAttributeType.REQUIRES_NEW;
-                        break;
-                }
-            }
 
             switch (attr) {
                 case MANDATORY:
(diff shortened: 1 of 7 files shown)


chengfang:
I'll look into test failures (look like real failures caused by the changes).

bstansberry:
Perhaps a new method or two in TransactionPropagationTestCase for REQUIRES and SUPPORTS? It looks like the existing testRemoteInvocation already covers MANDATORY.

chengfang:
will add these tests. I tested these cases with some manual testing.

bstansberry:
@chengfang please @ ping me when you do so I can process this quickly

chengfang:
@bstansberry I just added test for async method with REQUIRED transaction attribute. The existing tests already covers MANDATORY, SUPPORTS.

bstansberry (approved):

ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
```diff
@@ -141,6 +144,15 @@ public Object processInvocation(InterceptorContext invocation) throws Exception
...
                     return notSupported(invocation, component);
                 case REQUIRED:
+                    final ComponentView view = invocation.getPrivateData(ComponentView.class);
+                    if (view != null && view.isAsynchronous(method)) {
+                        // EJB 3.2 4.5.3 Transactions
+                        // The client’s transaction context does not propagate with an asynchronous method invocation. From the
+                        // Bean Provider’s point of view, there is never a transaction context flowing in from the client. This
+                        // means, for example, that the semantics of the REQUIRED transaction attribute on an asynchronous
+                        // method are exactly the same as REQUIRES_NEW.
+                        return requiresNew(invocation, component, timeoutInSeconds);
```
wolfc: How did we get to the point of having a tx here during an async invocation?

testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/transaction/async/ClientBean.java
```diff
@@ -84,10 +84,28 @@ public void callToStatusByTransactionmanager() throws Exception {
...
+    public void callToRequired() throws Exception {
+        final TransactionalRemote remote = getRemote(TransactionalStatusByManager.class);
+        userTransaction.begin();
+
+        // asyncWithRequired() will throw RuntimeException and cause its transaction to rollback.
+        // But it has no bearing on the transaction here, which should be able to commit okay.
+        remote.asyncWithRequired();
+
+        // wait a bit for the transaction in asyncWithRequired method to rollback
+        Thread.sleep(1000);
```
wolfc: You hold the Future, so better use it.

chengfang: `asyncWithRequired` method returns void since it just throws a `RuntimeException` with no need to return a value. We could add a return value and use `get()` instead of `sleep()`.

ejb3/src/main/java/org/jboss/as/ejb3/tx/CMTTxInterceptor.java
chengfang: Not sure why, it could be some missed check during earlier part of invocation chain. I'll look into it.

testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/transaction/async/ClientBean.java
jamezp: Get instead of sleep would definitely be preferred as sleep could be racy.

wolfc: If it did return void, then we have another issue. It should be throwing an `ExecutionException`.
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutionException.html

chengfang: When it returns void, I think it should just be a fire-and-forget on the client side, and whatever happened in the target method is not visible to the client at all.  The `ExecutionException` is available via the `get()` method when the method returns a `Future`.

PR #12903, opened 2020-01-13
WFLY-12516 - fix auth context propagation

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-12516


bstansberry:
@baranowb Please pull the tests from #12993 into your branch so we can have a single PR.

darranl (changes requested):
Just looking at the code but the commits needs WFLY Jira issues to be referenced not JBEAP issues.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
```diff
@@ -352,7 +352,7 @@ public void push(final String securityDomain, String userName, char[] password,
         contexts.push(previous);
         SecurityContext current = establishSecurityContext(securityDomain);
         if (propagate &&  previous != null) {
-            current.setSubjectInfo(getSubjectInfo(previous));
+            propagateSubject(current,previous);
             current.setIncomingRunAs(previous.getOutgoingRunAs());
```
darranl: As a full method has been added handling the propagation from a previous context to a current context shouldn't this second "copy" also be within the "propagateSubject" method?

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
```diff
@@ -365,6 +365,26 @@ public void push(final String securityDomain, String userName, char[] password,
         }
     }
 
+    private void propagateSubject(final SecurityContext target, final SecurityContext source) {
+        final SecurityContextUtil previousUtil = source.getUtil();
+        final SecurityContextUtil currentUtil = target.getUtil();
+        if(target.getSecurityDomain() != null && source.getSecurityDomain() != null && target.getSecurityDomain().equals(source.getSecurityDomain())){
```
darranl: Only a minor comment but shouldn't it be possible to skip the middle && source.getSecurityDomain() or does the final .equals call throw a NPE if null is passed into it?

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
```diff
@@ -365,6 +365,26 @@ public void push(final String securityDomain, String userName, char[] password,
...
+            currentUtil.createSubjectInfo(previousUtil.getUserPrincipal(), previousUtil.getCredential(), previousUtil.getSubject());
+        }
+        if (previousUtil.getRoles() != null) {
+            try {
+                currentUtil.setRoles((RoleGroup) previousUtil.getRoles().clone());
+            } catch (CloneNotSupportedException e) {
+                // should not happen, RoleGroup supports clone.
+                throw new RuntimeException(e);
+            }
+        }
```
darranl: What is the reason this block is needed?  The previous implementation was not manually cloning the RoleGroup and the bug report seems to be in relation to popping the Principal.

I have always considered role propagation in PicketBox across domains to be a bit of an undefined, does it make sense to be cloning the roles across if we both have the same or different domain?

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
```diff
@@ -365,6 +365,26 @@ public void push(final String securityDomain, String userName, char[] password,
...
+        currentUtil.setIdentities(previousUtil.getIdentities(Identity.class));
```
darranl: I don't know how these identities are now used but this doesn't seem to be something we previously propagated.

darranl:
I don't know enough about the inner workings of PicketBox but at the moment the original issue seemed to be about ensuring the push / pop of a principal is handled cleanly where differing domains are in use but a couple of items seem to be propagation of additional types that we did not previously cover.

Overall how we propagate between different SecurityDomains with PicketBox was never clearly defined, that is why with WildFly Elytron we did implement a true domain to domain propagation into the core.

For the bug fix I think we certainly do need to make sure the pops restore us back to a previous state, is there a reason we need to handle the other items during propagation?  I am not sure with those additional items if existing apps could see a change in behaviour?

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: Not all implementations of RoleGroup had it(clone) in PIcketBox. If users have their own impl, we cant be sure they have as well.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: JDoc does not describe what/if of getSecurityDomain. So this is precaution.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: There are different ways of handling incoming identity, did not want to touch it. Fewer changes the better.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: AFAIR - this is due that SubjectInfo is created, not copied. If you copy, you use identities from copy, upon creation, they are empty.

darranl:
@baranowb If you are happy your code changes are justified lets go ahead, I don't know enough about the inner details of PicketBox which by the sounds of it you went into in detail for this issue.

baranowb:
Im looking into test case as it keeps failing localy. give me a sec and I will push changes to this branch.

darranl:
Can we please clean up the commit messages in this PR, generally we should try for a meaningful sentence that if someone reads in five years time it makes some sense without forcing them to open a bunch of Jira issues for context.

Then if we get a clean CI run we can see about merging.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/EJBContextMultipleSDTestCase.java
```diff
@@ -0,0 +1,230 @@
...
+import org.junit.FixMethodOrder;
+import org.junit.runners.MethodSorters;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+/**
+ * @author Francesco Marchioni
+ * JIRA: https://issues.redhat.com/browse/JBEAP-17862
```
fjuma, line 69: Just a minor comment but this should reference the WFLY issue instead of the JBEAP issue.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/EJBContextMultipleSDTestCase.java
```diff
@@ -0,0 +1,230 @@
...
+        ejbClientProperties.put(Context.PROVIDER_URL, REMOTE_HTTP);
+        ejbClientProperties.put(Context.SECURITY_PRINCIPAL, USER);
+        ejbClientProperties.put(Context.SECURITY_CREDENTIALS, PWD);
+
+        Context context = new InitialContext(ejbClientProperties);
+        return context;
+    }
+
+    @Test
+    public void firstTest() throws Exception {
```
fjuma, line 111: Another minor comment but it would be good to give these test methods names that indicate what the test is trying to accomplish.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/ejb/BaseHello.java
```diff
@@ -0,0 +1,68 @@
+package org.jboss.as.test.integration.ejb.security.securitydomain.ejb;
```
fjuma, line 1: Missing copyright header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/ejb/Hello.java
```diff
@@ -0,0 +1,15 @@
+package org.jboss.as.test.integration.ejb.security.securitydomain.ejb;
```
fjuma, line 1: Missing copyright header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/ejb/HelloOneBean.java
```diff
@@ -0,0 +1,25 @@
+package org.jboss.as.test.integration.ejb.security.securitydomain.ejb;
```
fjuma, line 1: Missing copyright header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/ejb/HelloTwoBean.java
```diff
@@ -0,0 +1,25 @@
+package org.jboss.as.test.integration.ejb.security.securitydomain.ejb;
```
fjuma, line 1: Missing copyright header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/securitydomain/ejb/Info.java
```diff
@@ -0,0 +1,56 @@
+package org.jboss.as.test.integration.ejb.security.securitydomain.ejb;
```
fjuma, line 1: Missing copyright header.

fjuma (changes requested):

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
fjuma: @baranowb I might have misunderstood, but if calling `setIdentities` and `setRoles` is only needed if the `SubjectInfo` is being created, then shouldn't these last two blocks move into the `else` block where `createSubjectInfo` is being called?

bstansberry:
#13264 combines this with the needed Picketbox upgrade. It looks like the tests fail with the -Delytron profile:

https://ci.wildfly.org/viewLog.html?buildId=202081&buildTypeId=WF_PullRequest_LinuxElytron

That's not too surprising if they are legacy security tests but perhaps they need to be disabled if Elytron is used?

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: @fjuma which else block?

tomjenkinson: @baranowb - perhaps @fjuma refers to the `else` block started on line 373 which contains line 374 createSubjectInfo, but I can't say for sure, @fjuma is that the one?

sguilhen: That's the one she is referring to. When we discussed this months ago in the Jira I suggested following the original flow when the domains match (the if clause above) and only copy things otherwise. In this case the `setIdentities` and `setRoles` above should have been moved to the else block as @fjuma pointed`. Not sure if @baranowb came across some other issue when doing it but from my perspective the whole thing should be in the else block so that it is only performed when we need to create copies of the subject info.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
sguilhen: One minor comment about this, have you tried not cloning the roles? Did you run into issues if you don't do it? Reason I'm asking is that I'm not sure this is required - any roles established by the previous context should not come into play when creating a new one and performing a new authentication. I might be mistaken and this might be required but I'm just not sure about it.

bstansberry:
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/container/interceptor/security/SwitchIdentityTestCase.java#L96 is an example of how to disable the new test when the elytron profile is used.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: AFAIR, yes, tried with minimal set of cloning up until it worked.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: For some reason git displayed commit that did not have it( previous one ). At this moment I cant be sure, let me test( but it seems it should have been in 'else')

darranl (changes requested):
I have just seen another problem, why is there a binary jar file in this pull request?

https://github.com/wildfly/wildfly/pull/12903/files#diff-120bd7f76b30871c7b0b33d8d2c65aa9

We don't add pre-created jar files the source instead should be a part of the test and the jar created by the build or the test.

security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
baranowb: @fjuma @sguilhen  - yep, tested in 7.3.x ( inclusion of setRoles and setIdentities in 'else') so it makes sense to update this.

darranl:
Superseded by https://github.com/wildfly/wildfly/pull/13271

PR #12905, opened 2020-01-14
[WFLY-12590] Add a fault-tolerance layer test to testsuite/layers

bstansberry:
Follow-on to #12800 to add a test in testsuite/layers.

Also prunes some cruft from the equivalent test for openapi.


bstansberry:
This was implemented as part of #12853

PR #12886, opened 2020-01-07
WFLY-12937, WFLY-12938, WFLY-12939 Fix memory leaks due to orphaned expired cache entries following topology change

pferraro:
https://issues.redhat.com/browse/WFLY-12937
https://issues.redhat.com/browse/WFLY-12938
https://issues.redhat.com/browse/WFLY-12939

PR #12908, opened 2020-01-14
WFLY-12961 Fix intermittent deadlock between LocalScheduler.close() and run()

pferraro:
https://issues.redhat.com/browse/WFLY-12961

### DIFF
--- a/clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/LocalScheduler.java
+++ b/clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/LocalScheduler.java
@@ -26,8 +26,6 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Iterator;
 import java.util.Map;
-import java.util.concurrent.CancellationException;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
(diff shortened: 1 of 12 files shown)


clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/LocalScheduler.java
```diff
@@ -115,15 +115,13 @@ public void accept(Map.Entry<T, Instant> entry) {
     }
 
     @Override
-    public synchronized void close() {
-        WildFlySecurityManager.doPrivilegedWithParameter(this.executor, DefaultExecutorService.SHUTDOWN_ACTION);
-        if (this.future != null) {
+    public void close() {
+        WildFlySecurityManager.doPrivilegedWithParameter(this.executor, DefaultExecutorService.SHUTDOWN_NOW_ACTION);
```
bstansberry: @pferraro So it's ok if any executing run() method gets interrupted and any scheduled but not yet executed run() doesn't get called?

I figure yes, but I'm asking because this seems to be the key change in the PR.

pferraro: Correct.

bstansberry (approved):

bstansberry:
@pferraro This will conflict with #12886. I'll put the ready for merge label on both and leave it to you to merge in the order you prefer.

PR #12916, opened 2020-01-15
WFLY-12909 Fix license names for org.hdrhistogram:HdrHistogram licenses

rhusar:
Jira
https://issues.redhat.com/browse/WFLY-12909

Fixed names, swapped order.

### DIFF
--- a/feature-pack/src/license/full-feature-pack-licenses.xml
+++ b/feature-pack/src/license/full-feature-pack-licenses.xml
@@ -2128,12 +2128,12 @@
       <artifactId>HdrHistogram</artifactId>
       <licenses>
         <license>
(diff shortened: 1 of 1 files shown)


rhusar:
@jamezp  Please review. Maybe its enough to keep just one license if we decipher https://github.com/HdrHistogram/HdrHistogram/blob/master/LICENSE.txt? IANAL...

jamezp (approved):
@rhusar I wasn't too sure either. Both licenses are mentioned in that license so I say we just leave both.

jamezp:
I added the ready-to-merge repository while the tests are still running as this is only a license change.

PR #12896, opened 2020-01-10
WFLY-12954 Skip passivation of entries on stop if passivation store uses purgeOnStartup.

pferraro:
https://issues.redhat.com/browse/WFLY-12954

PR #12768, opened 2019-11-04
[WFLY-11930] Allow Java EE Subsystem to enable/disable descriptor bas…

yersan:
…ed property replacement for permissions descriptors

wildfly counterpart of https://github.com/wildfly/wildfly-core/pull/4002. It requires a wildfly-core upgrade including that PR to make this to pass.

Jira issue: https://issues.jboss.org/browse/WFLY-11930

PR #12918, opened 2020-01-15
Combined PR

darranl:
Combined PR of the following PRs: -

https://github.com/wildfly/wildfly/pull/12908
https://github.com/wildfly/wildfly/pull/12880
https://github.com/wildfly/wildfly/pull/12896
https://github.com/wildfly/wildfly/pull/12768


darranl:
The jobs are playing up a bit but we had a good CI run here: -

https://ci.wildfly.org/viewLog.html?buildId=182216&tab=queuedBuildOverviewTab

The Linux with SecurityManager job was hanging but we had good runs on the individual PRs with that one so don't think there should be an interaction so going ahead and merging.

darranl:
Merged

PR #12927, opened 2020-01-16
[WFLY-12985] Provide a separate feature pack and dist for non-observability MP specs

bstansberry:
https://issues.redhat.com/browse/WFLY-12985


bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
@jamezp I fixed the -Drelease problem. Thanks for pointing it out!

jamezp (approved):
This looks good to me.

bstansberry:
retest this please

bstansberry:
retest this please

ee-galleon-pack/src/main/resources/packages/product.conf/content/modules/system/layers/base/org/jboss/as/product/wildfly-ee/module.xml
```diff
@@ -0,0 +1,38 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.5" name="org.jboss.as.product" slot="wildfly-ee">
```
yersan: Should not this change be also effective for the legacy product module.xml?
https://github.com/wildfly/wildfly/blob/master/dist-legacy/src/distribution/resources/modules/system/layers/base/org/jboss/as/product/wildfly-full/module.xml#L25

We could add and additional verification to ensure modules/system/layers/base/org/jboss/as/product/wildfly-ee/module.xml has the correct one

ee-galleon-pack/pom.xml
```diff
@@ -0,0 +1,4155 @@
...
+
+        <dependency>
+            <groupId>org.eclipse.microprofile.rest.client</groupId>
+            <artifactId>microprofile-rest-client-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api<
```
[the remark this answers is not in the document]
bstansberry: Thanks; it looks like I missed deleting this one. I'll update the relevant commit.

ee-galleon-pack/src/main/resources/packages/product.conf/content/modules/system/layers/base/org/jboss/as/product/wildfly-ee/module.xml
bstansberry: I think it should, yes.

bstansberry: Oh, definitely it should, as the name of the module dir was not matching the module.xml!

yersan (approved):

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
```diff
@@ -115,6 +115,14 @@ WildFly defined layers to assemble a custom configuration. <<galleon-cli-recipes
 
 Basic layers:
 
+* _microprofile-fault-tolerance_: Support for MicroProfile Fault Tolerance.
+* _microprofile-jwt_: Support for MicroProfile JWT.
```
jfdenise: These layers are still documented in the WildFly EE layers, should be removed from EE.

feature-pack/pom.xml
```diff
@@ -332,17 +332,6 @@
             <artifactId>jakarta.security.enterprise</artifactId>
         </dependency>
 
-        <dependency>
```
dosoudil: If these artifacts are removed from feature-pack, they're not runtime artifacts anymore.
Moving them to galleon-pack means that they're build dependencies, not runtime.
Is it correct from the perspective of using WildFly Server?
cc @rgdoliveira

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
bstansberry: Thanks; will fix.

feature-pack/pom.xml
bstansberry: They are still runtime dependencies of any server that uses the 'wildfly' galleon pack. The output of this maven module and the 'servlet-feature-pack' maven module are 'legacy feature packs' and are not used in any WildFly distribution. We just maintain these modules so projects that still want to create custom WF-technology-based servers using the legacy feature packs can use their output. But an official WildFly dist does not use the output of these. (You could do 'mvn install -pl dist -am' and this maven module wouldn't even build.) 

However, our official dists do use some *source* from these maven modules. This is so we can build the official Galleon feature packs and these legacy ones and not have to duplicate the source, which would almost certainly lead to bugs.  The Galleon feature pack poms actually copy over the source resources they need:

https://github.com/wildfly/wildfly/blob/master/galleon-pack/pom.xml#L73
https://github.com/wildfly/wildfly/blob/master/servlet-galleon-pack/pom.xml#L548

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
bstansberry: Fixed.

bstansberry:
retest this please

bstansberry:
retest this please

PR #12940, opened 2020-01-27
[WFLY-12952] Register default readiness probe for deployment

jmesnil:
When a deployment does not provide any MicroProfile readiness probe,
add a default one (which returns UP as long as the deployment is deployed).

This allow to configure the overall server readiness to be DOWN when
the server is started and becomes UP as soon as a deployment is
deployed.

This default probe is disabled when the MicroProfile Health TCK is run.

JIRA: https://issues.redhat.com/browse/WFLY-12952

### DIFF
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
@@ -37,11 +37,14 @@ import javax.enterprise.inject.spi.ProcessAnnotatedType;
 import javax.enterprise.util.AnnotationLiteral;
 
 import io.smallrye.health.SmallRyeHealthReporter;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.health.Health;
 import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
 import org.eclipse.microprofile.health.Liveness;
 import org.eclipse.microprofile.health.Readiness;
-import org.jboss.modules.ModuleClassLoader;
+import org.jboss.modules.Module;
 import org.wildfly.extension.microprofile.health.HealthReporter;
 
 /**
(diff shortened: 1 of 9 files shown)


microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
```diff
@@ -96,12 +101,21 @@ private void afterDeploymentValidation(@Observes final AfterDeploymentValidation
         addHealthChecks(HealthLiteral.INSTANCE, reporter::addHealthCheck, healthChecks);
         addHealthChecks(LivenessLiteral.INSTANCE, reporter::addLivenessCheck, livenessChecks);
         addHealthChecks(ReadinessLiteral.INSTANCE, reporter::addReadinessCheck, readinessChecks);
+        if (readinessChecks.isEmpty()) {
+            Config config = ConfigProvider.getConfig(module.getClassLoader());
+            boolean disableDefaultprocedure = config.getOptionalValue("mp.health.disable-default-procedures", Boolean.class).orElse(false);
```
bstansberry: In the context of other MP specs I've seen a particular pattern of config property keys used for impl-specific config properties (something like mp.<spec>.extension.<therest> -- @pferraro knows).  Is there such a thing for health? If not is it wise to act is if there was?

jmesnil: I'm not sure to understand your comment.

This property `mp.health.disable-default-procedures` is defined by the [spec itself](https://github.com/eclipse/microprofile-health/blob/2.1/spec/src/main/asciidoc/protocol-wireformat.adoc#disabling-default-vendor-procedures)

jmesnil:
I flagged it as `19.x` as this would be great to get that enhancement with WildFly 19 (and MicroProfile 3.3).
While it is not strictly required, this allows to use the `/health/ready` probe in all cases (deployments with or without readiness probe). WildFly can be configured with `empty-readiness-checks-status = DOWN` so that it will not be considered ready until the (first) deployment is deployed. If It contains readiness probes they will be used to determine the readiness of the deployment (and thus the overall readiness of the app server). If it does not contain any readiness probe, this automatic probe will make sure the deployment is considered ready when it is deployed.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
bstansberry: OK; thanks.

bstansberry (approved):

PR #12942, opened 2020-01-27
Fix for WFLY-13017, cloud-server layer to extend jaxrs-server layer

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-13017

Refactor cloud-server to align dependencies of datasources-web-server, jaxrs-server and cloud-server.

### DIFF
--- a/galleon-pack/src/main/resources/layers/standalone/cloud-server/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/cloud-server/layer-spec.xml
@@ -1,13 +1,10 @@
 <?xml version="1.0" ?>
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="cloud-server">
     <dependencies>
-        <layer name="cloud-profile"/>
-        <layer name="core-server"/>
-        <layer name="core-tools"/>
+        <layer name="jaxrs-server"/>
+        <layer name="ee-security" optional="true"/>
+        <layer name="jms-activemq" optional="true"/>
+        <layer name="observability" optional="true"/>
+        <layer name="resource-adapters" optional="true"/>
     </dependencies>
-    <feature-group name="undertow-elytron-security"/>
-    <packages>
-        <!-- Support for installing legacy one-off patches -->
-        <package name="org.jboss.as.patching.cli" optional="true"/>
-    </packages>
 </layer-spec>
--- a/galleon-pack/src/main/resources/layers/standalone/jaxrs-server/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/jaxrs-server/layer-spec.xml
@@ -2,7 +2,9 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="jaxrs-server">
     <dependencies>
         <layer name="datasources-web-server"/>
-        <layer name="jaxrs" />
+        <!-- cloud-server depends on jaxrs-server, jaxrs could be excluded from cloud-server,
+             this is why jaxrs is an optional dependency -->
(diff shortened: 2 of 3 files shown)


galleon-pack/src/main/resources/layers/standalone/jaxrs-server/layer-spec.xml
```diff
@@ -2,7 +2,7 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="jaxrs-server">
     <dependencies>
         <layer name="datasources-web-server"/>
-        <layer name="jaxrs" />
+        <layer name="jaxrs" optional="true"/>
```
bstansberry: A comment here as to why this is optional would be good. Some maintainer some day will wonder and may naively remove the optional.

jfdenise: Ok, added.

jmesnil (approved):
@jfdenise I'm approving the PR, content is fine.

Just a note for future updates, it'd be better to use `monospace` text instead of _emphasized_ text for layers.
This looks more consistent with code listing and make it easier to distinguish "user input" from other emphasized text

galleon-pack/src/main/resources/layers/standalone/cloud-server/layer-spec.xml
```diff
@@ -1,13 +1,10 @@
...
-        <layer name="cloud-profile"/>
-        <layer name="core-server"/>
-        <layer name="core-tools"/>
+        <layer name="jaxrs-server"/>
+        <layer name="ee-security" optional="true"/>
+        <layer name="jms-activemq" optional="true"/>
+        <layer name="observability" optional="true"/>
+        <layer name="resource-adapters" optional="true"/>
     </dependencies>
-    <feature-group name="undertow-elytron-security"/>
```
jmesnil: @jfdenise could you please confirm my understanding?
 
* `core-server`, `core-tools` and `undertow-elytron-security` will now come from `jaxrs-server` > `datasources-web-server`
*  `cloud-profile` is no longer required as all layers listed in it are now listed in `cloud-server`. But shouldn't still be a relation between `cloud-server` and `cloud-profile`?

jfdenise: What kind of relation we could have? cloud-profile dependency is to me an implementation detail. With dependency on jaxrs-server we offer the same level of flexibility.

bstansberry: This is a tangent and not something to address here but I suspect the documentation of cloud-profile is no longer accurate.

cloud-profile, an aggregation of the basic layers (except h2-* layers) and a web-server .

The "basic layers" list is a moving target.

Hmm, maybe my #12927 change to that file (https://github.com/wildfly/wildfly/pull/12927/files#diff-9a6f6b7a6bd30fddb786db3840aac00f) somewhat makes that go away by moving FT/JWT/OpenAPI into a different section.  But that text can break in the future, as soon as we add a new layer.

Perhaps drop it from the doc page?

bstansberry (approved):

galleon-pack/src/main/resources/layers/standalone/cloud-server/layer-spec.xml
bstansberry: @jfdenise ^^^

jfdenise: Yes, +1 to remove it from the doc.

PR #12945, opened 2020-01-28
Clustering jpa2lc tests with no ejb, exclude layers test

jfdenise:
Issues:
https://issues.redhat.com/browse/WFLY-13029
https://issues.redhat.com/browse/WFLY-13030


rhusar (approved):
I think this looks good. Not a fan of the direction where the `testsuite/integration/clustering/pom.xml` has been going recently but what can we do...

bstansberry (approved):

PR #12948, opened 2020-01-29
WFLY-12681 Provide a 'microprofile-platform' Galleon layer

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-12681

Proposal
https://github.com/wildfly/wildfly-proposals/pull/293

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
@@ -115,6 +115,7 @@ WildFly defined layers to assemble a custom configuration. <<galleon-cli-recipes
 
 Basic layers:
 
+* _microprofile-platform_: Support for available MicroProfile specifications.
 * _microprofile-fault-tolerance_: Support for MicroProfile Fault Tolerance.
 * _microprofile-jwt_: Support for MicroProfile JWT.
 * _microprofile-openapi_: Support for MicroProfile OpenAPI.
--- a/galleon-pack/src/main/resources/layers/standalone/microprofile-platform/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/microprofile-platform/layer-spec.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" ?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2020, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="microprofile-platform">
+    <dependencies>
+        <layer name="microprofile-config" optional="true"/>
+        <layer name="microprofile-fault-tolerance" optional="true"/>
+        <layer name="microprofile-health" optional="true"/>
+        <layer name="microprofile-jwt" optional="true"/>
+        <layer name="microprofile-metrics" optional="true"/>
+        <layer name="microprofile-openapi" optional="true"/>
+        <layer name="open-tracing" optional="true"/>
+    </dependencies>
+</layer-spec>
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1344,6 +1344,48 @@
                                 </configuration>
                             </execution>
 
+                            <!-- Provision a server with microprofile-platform support. -->
+                            <execution>
+                                <id>microprofile-platform-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${project.build.directory}/wildfly-microprofile-platform</install-dir>
+                                    <record-state>false</record-state>
+                                    <log-time>${galleon.log.time}</log-time>
+                                    <offline>true</offline>
+                                    <plugin-options>
+                                        <jboss-maven-dist/>
+                                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                                        <optional-packages>passive+</optional-packages>
+                                    </plugin-options>
+                                    <feature-packs>
+                                        <feature-pack>
+                                            <groupId>${testsuite.ee.galleon.pack.groupId}</groupId>
+                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
+                                            <version>${testsuite.ee.galleon.pack.version}</version>
+                                            <inherit-configs>false</inherit-configs>
+                                            <inherit-packages>false</inherit-packages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <configurations>
+                                        <config>
+                                            <model>standalone</model>
+                                            <name>standalone.xml</name>
+                                            <layers>
+                                                <layer>microprofile-platform</layer>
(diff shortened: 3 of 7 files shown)


docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
```diff
@@ -122,6 +122,7 @@ Basic layers:
 * _jaxrs_: Support for JAXRS. Depends on _web-server_ (defined in the WildFly servlet feature-pack).
 * _jms-activemq_: Support for connections to a remote JMS broker.
 * _jpa_: Support for JPA (using the latest WildFly supported Hibernate release).
+* _microprofile_: Support for MicroProfile specification including Config, Fault Tolerance, JWT, Metrics, OpenAPI, OpenTracing.
```
pferraro: I'm not sure I would bother enumerating the individual specs, unless we want to keep this list updated (e.g. health) with every new release.

pferraro (changes requested):

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
rhusar: My thinking was to keep the list updated so people dont assume this includes some specs we dont yet support or ones that might be only incubating. What better wording do you suggest?

darranl: Would including an additional qualifier in the name make sense?  Either platform version  or allow for a second one `microprofile_experimental`

rhusar: Reworded to be more general.

rhusar:
As per @bstansberry 's suggestion, made all the specs optional except for config which is required by all specifications.

rhusar:
For the sake of completeness, I have opened the proposal as well https://github.com/wildfly/wildfly-proposals/pull/293

pferraro (approved):
LGTM

rhusar:
Rebased again..

darranl (approved):

darranl:
No objection from me, do we need to double check the approach on Monday?

Also FYI this needs a rebase again.

rhusar:
@darranl Right, lets discuss on Mondays call.

galleon-pack/src/main/resources/layers/standalone/microprofile/layer-spec.xml
```diff
@@ -0,0 +1,32 @@
...
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="microprofile">
+    <dependencies>
+        <layer name="microprofile-config" optional="true"/>
+        <layer name="microprofile-fault-tolerance" optional="true"/>
+        <layer name="microprofile-jwt" optional="true"/>
+        <layer name="microprofile-metrics" optional="true"/>
+        <layer name="microprofile-openapi" optional="true"/>
+        <layer name="open-tracing" optional="true"/>
```
yersan, line 30: @rhusar Just out of curiosity, why we don't include microprofile-health?

rhusar: @yersan Fixed, thanks. The reason being that 'microprofile-health' layer did not exist in January.

darranl:
Sorry @rhusar but this is conflicting again.

rhusar:
@darranl Thanks, back from PTO, rebased now.

testsuite/layers/pom.xml
```diff
@@ -2266,6 +2302,7 @@
                                                 <layer>logging</layer>
                                                 <layer>management</layer>
                                                 <layer>management-security-realm</layer>
+                                                <layer>microprofile-platform</layer>
                                                 <layer>microprofile-config</layer>
```
tommaso-borgato, line 2306: @rhusar given `microprofile-platform` already includes `microprofile-config`, `microprofile-fault-tolerance`, `microprofile-health`, `microprofile-jwt`, `microprofile-metrics`, `microprofile-openapi`, `open-tracing`: can we remove them from the list? or we keep them to prevent errors in case, in some future, they are removed from `microprofile-platform` or `microprofile-platform` itself is removed?

tommaso-borgato (changes requested):

testsuite/layers/pom.xml
rhusar: @tommaso-borgato We can – but the existing layers are already overlapping (e.g. undertow is listed explicitly even though its a dependency of any of the web server layers, e.g. datasources-web-server). So this is intentional.

testsuite/integration/basic/pom.xml
```diff
@@ -1229,6 +1270,45 @@
...
+                                        <JBOSS_HOME>${project.build.directory}/wildfly-microprofile-platform</JBOSS_HOME>
+                                        <!-- dummy env property for MicroProfile Config testing -->
+                                        <MPCONFIG_TEST_ENV_VAR>integ-basicmpconfig-test-env-var-value</MPCONFIG_TEST_ENV_VAR>
+                                        <TEST_PROP>JMXPropertyEditorsTestCase-WFLY-12899</TEST_PROP>
+                                    </environmentVariables>
+                                    <redirectTestOutputToFile>${testLogToFile}</redirectTestOutputToFile>
+                                    <includes>
+                                        <include>org/wildfly/test/integration/microprofile/**/*TestCase.java</include>
+                                    </includes>
+                                    <excludes>
```
tommaso-borgato, line 1299: @rhusar what drove the decision to exclude these tests? I am asking just for getting some understanding ... (I already asked on the AD, so if you answer there I'll copy the answer here and vice-versa)

tommaso-borgato (changes requested):

testsuite/integration/basic/pom.xml
rhusar: @tommaso-borgato These tests have other dependencies in order to run (e.g. security) while the profile only provisions base server (`jaxrs-server`) with the `microprofile-plaform` layer and of course `jmx-remoting` & `security-manager` which are needed by Arquillian. We might consider adding more layers to support more tests, but I am not sure if that doesn't defeat the purpose of the test.

tommaso-borgato: @rhusar no, that's fine .... it was just for understanding

galleon-pack/src/main/resources/layers/standalone/microprofile-platform/layer-spec.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" ?>
```
tommaso-borgato, line 1: @rhusar please consider reading [this comment](https://docs.google.com/document/d/1FXdEQc1Q8v1GmsTbwb6XDl_GTFrcFYuQPpkAOV6hrUI/edit?disco=AAAAJxVMytU) in order to validate the location where this layer is defined

testsuite/integration/basic/pom.xml
tommaso-borgato: @rhusar .... this layer looks like one of the biggest candidate for adoption, also given we are in the XP 2 context; 

what about adding one or more galleon-maven-plugin & maven-surefire-plugin executions in order to run the excluded tests on wildfly servers taylored with the minimal number of necessary layers? Something analogous will be done for the JSF layer with regard to the EJB layer (@yersan is that correct?) ...

tommaso-borgato (changes requested):

galleon-pack/src/main/resources/layers/standalone/microprofile-platform/layer-spec.xml
rhusar: @tommaso-borgato The comment confirms the location is correct. Resolving.

testsuite/integration/basic/pom.xml
rhusar: @tommaso-borgato We can certainly do that, the price we pay is the extra testsuite execution time which is kind of exploding atm. Do you have a link handy where the JSF+EJB layer tests are discussed by any chance?

tommaso-borgato: @rhusar I am waiting for EAP7-1529 to be completed with JSF+EJB test, that's why I tagged @yersan: here is the TP snippet relative to this: https://gitlab.mw.lab.eng.bos.redhat.com/jbossqe-eap/test-plans/merge_requests/380/diffs#4e1bf40d542063d096fcd61d1fb6f49b4aba74ab_0_177

Probably he can contribute to this discussion ....

tommaso-borgato: @rhusar in addition to that, microprofile tests are fast (about 10 seconds on my laptop)

tommaso-borgato: ^^^ do you have any update on the missing tests? would it be possible, for example, to add security tests on a slimmed server with the minimum amount of necessary layers?

tommaso-borgato (changes requested):

testsuite/integration/basic/pom.xml
rhusar: Actually, it looks like the problem is only with the test setup. I will put this on the backlog, but its not clear what those tests are missing – will need to check with Darran when he comes back.

rhusar:
@bstansberry Needed a rebase again (also I think I botched the previous rebased, fixed that too).

PR #12950, opened 2020-01-30
[WFLY-4989] testcase unignored

moulalis:
https://issues.redhat.com/browse/WFLY-4989

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/jacc/JACCAuthzPropagationTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/jacc/JACCAuthzPropagationTestCase.java
@@ -50,7 +50,6 @@ import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/jacc/JACCAuthzPropagationTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/jacc/JACCAuthzPropagationTestCase.java
@@ -62,7 +62,6 @@ import org.junit.runner.RunWith;
 @RunWith(Arquillian.class)
 @ServerSetup({JACCAuthzPropagationTestCase.SecurityDomainsSetup.class})
 @RunAsClient
-@Ignore("See WFLY-4989")
 public class JACCAuthzPropagationTestCase {
 
     private static final Logger LOGGER = Logger.getLogger(JACCAuthzPropagationTestCase.class);


bstansberry:
this is ok to test

jmesnil:
@moulalis CI jobs failed due to an checkstyle error: 

```
JACCAuthzPropagationTestCase.java:53:8: Unused import - org.junit.Ignore. [UnusedImports]
```

moulalis:
> @moulalis CI jobs failed due to an checkstyle error:
> 
> ```
> JACCAuthzPropagationTestCase.java:53:8: Unused import - org.junit.Ignore. [UnusedImports]
> ```

I have removed unused import

bstansberry:
Hi @moulalis . Thanks for these. Before removing the Ignore annotation though we need to believe that whatever caused it to be applied has been resolved. For the ones you've sent PRs for I believe the problem is the tests were unreliable. Do you know if anything was done to make them reliable?

Also, the failure in the "Galleon Linux - JDK 11" job is because this test won't work with that job because it requires capabilities that the WildFly servers provisioned by that job won't have (i.e. ejb support).  To correct that problem a line above this one would need to be added to exclude this test from the job:

https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/pom.xml#L1005

darranl:
@moulalis Issues such as WFLY-4989 are not just about removing the @ignore annotation - if that was all that was required we would have done it at the time.

In most cases where we add the annotation it would be for one of two reasons: -

Intermittent failures in our CI environment making many PR runs fail.
A change in implementation where the test was heavily dependent on the old implementation rather than the feature being tested.
In both cases the Jira issue is more about taking the time to understand the original problem and provide appropriate fixes either to get the test passing or to get the test stable again.

Any fixes required could either be to the test case itself or they may be required within the implementation of the feature under test.

The failures reported in CI are directly related to the test case that was reactivated so further work does appear required before they can be activated, for this reason I am going to close this PR for now. Please feel free to reopen if you work through the remaining issues.

darranl:
@moulalis FYI if you would like to speak to the WildFly team to discuss the requirements for these or any other open issues you can also join us in Zulip at https://wildfly.zulipchat.com/ or we also have a developer mailing list at https://lists.jboss.org/mailman/listinfo/wildfly-dev which you can join.

moulalis:
> @moulalis Issues such as WFLY-4989 are not just about removing the @ignore annotation - if that was all that was required we would have done it at the time.
> 
> In most cases where we add the annotation it would be for one of two reasons: -
> 
> Intermittent failures in our CI environment making many PR runs fail.
> A change in implementation where the test was heavily dependent on the old implementation rather than the feature being tested.
> In both cases the Jira issue is more about taking the time to understand the original problem and provide appropriate fixes either to get the test passing or to get the test stable again.
> 
> Any fixes required could either be to the test case itself or they may be required within the implementation of the feature under test.
> 
> The failures reported in CI are directly related to the test case that was reactivated so further work does appear required before they can be activated, for this reason I am going to close this PR for now. Please feel free to reopen if you work through the remaining issues.

Hi @darranl ,

Thank you for the information. I wanted to reproduce the same on CI and check before changing anything in code as it was very old.

darranl:
@moulalis if you want to continue to use these for CI runs, let us know - we can add a HOLD label so no one accidentally merges early.

Alternatively may be able to look at if TeamCity access is possible, that way you can kick off custom CI jobs against your topic branches to try out any changes.

moulalis:
> @moulalis if you want to continue to use these for CI runs, let us know - we can add a HOLD label so no one accidentally merges early.
> 
> Alternatively may be able to look at if TeamCity access is possible, that way you can kick off custom CI jobs against your topic branches to try out any changes.

Thank you @darranl I will do the same. I have also asked for TeamCity access.

PR #12952, opened 2020-01-31
[WFLY-13042]: Jaeger tracer sender-binding isn't configurable.

ehsavoie:
* Obtaining the socket-bindings when the tracer is being configured.
* Adding security constraint on the attribute definition.

Jira: https://issues.redhat.com/browse/WFLY-13042

### DIFF
--- a/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/resolver/JaegerTracerConfiguration.java
+++ b/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/resolver/JaegerTracerConfiguration.java
@@ -54,6 +54,7 @@ public class JaegerTracerConfiguration implements TracerConfiguration {
     private final CodecConfiguration codecConfig;
     private final SamplerConfiguration samplerConfig;
     private final ReporterConfiguration reporterConfig;
+    private final Supplier<OutboundSocketBinding> outboundSocketBindingSupplier;
     private final boolean traceId128Bit;
     private final Map<String, String> tracerTags;
     private final ModelNode model;
@@ -61,6 +62,7 @@ public class JaegerTracerConfiguration implements TracerConfiguration {
 
     public JaegerTracerConfiguration(OperationContext context, ModelNode configuration, Supplier<OutboundSocketBinding> outboundSocketBindingSupplier) throws OperationFailedException {
         this.name= TRACER_CAPABILITY.getCapabilityServiceName(context.getCurrentAddressValue()).getCanonicalName();
+        this.outboundSocketBindingSupplier = outboundSocketBindingSupplier;
         model = new ModelNode();
         for(AttributeDefinition att : ATTRIBUTES) {
             ModelNode value = att.resolveModelAttribute(context, configuration);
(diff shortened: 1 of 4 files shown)


ehsavoie:
retest this please

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/resolver/JaegerTracerConfiguration.java
```diff
@@ -54,13 +54,15 @@
     private final CodecConfiguration codecConfig;
     private final SamplerConfiguration samplerConfig;
     private final ReporterConfiguration reporterConfig;
+    private Supplier<OutboundSocketBinding> outboundSocketBindingSupplier = null;
```
jamezp: Not a huge deal, but this should probably be `final`.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/resolver/JaegerTracerConfiguration.java
```diff
@@ -102,6 +100,11 @@ public String getName() {
 
     @Override
     public Tracer createTracer(String serviceName) {
+        Configuration.SenderConfiguration senderConfiguration = reporterConfig.getSenderConfiguration();
+        if(outboundSocketBindingSupplier.get() != null) {
+            senderConfiguration.withAgentHost(outboundSocketBindingSupplier.get().getUnresolvedDestinationAddress())
+                    .withAgentPort(outboundSocketBindingSupplier.get().getDestinationPort());
+        }
```
jamezp: I don't really know how this is used, but we might want to do something more like.
```
final OutboundSocketBinding binding = outboundSocketBindingSupplier.get();
if(outboundSocketBindingSupplier.get() != null) {
    senderConfiguration.withAgentHost(binding.getUnresolvedDestinationAddress())
            .withAgentPort(binding.getDestinationPort());
}
```

I don't know what the overhead of `get()` is in MSC, but it seems like a single `get()` would be better.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/opentracing/SubsystemConfigurationTestCase.java
```diff
@@ -59,6 +65,14 @@
 @ServerSetup(SubsystemConfigurationTestCase.SetupTask.class)
 public class SubsystemConfigurationTestCase {
 
+    private static final String OUTBOUND_BINDING = "jaeger-sender";
+    private static final ModelNode OUTBOUND_BINDING_ADDRESS = PathAddress
+            .pathAddress(SOCKET_BINDING_GROUP, "standard-sockets")
+            .append(REMOTE_DESTINATION_OUTBOUND_SOCKET_BINDING, OUTBOUND_BINDING)
+            .toModelNode();
+    private static final int SENDER_PORT = 6832;
+    private static final String SENDER_HOST = NetworkUtils.formatPossibleIpv6Address(System.getProperty("node0", "localhost"));
```
jamezp: This should be using `TestSuiteEnvironment.getServerAddress()`. That method already formats ipv6 addresses and also checks the two properties `management.address` and `node0`.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/opentracing/application/TracerConfigurationApplication.java
```diff
@@ -23,8 +27,37 @@
...
+                        if ("ThriftUdpTransport".equals(transport.getClass().getSimpleName())) {
+                            Field socketField = transport.getClass().getDeclaredField("socket");
+                            socketField.setAccessible(true);
+                            DatagramSocket socket = (DatagramSocket) socketField.get(transport);
+                            return "sender-binding=" + NetworkUtils.formatIPAddressForURI(socket.getInetAddress()) + ":" + socket.getPort();
+                        }
+                    }
+                }
+            } catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException ex) {
+                ex.printStackTrace();
```
jamezp: Should we not be throwing the error here?

jamezp (changes requested):

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/opentracing/application/TracerConfigurationApplication.java
ehsavoie: yes, originally I wanted to avoid exception with different tracers implementations but this has been covered higher in the code

jamezp (approved):

bstansberry (changes requested):
@ehsavoie This failed checkstyle in testsuite/integration/basic

bstansberry (approved):

PR #12953, opened 2020-01-31
[WFLY-4990] Testcase Unignored and unused import removed

moulalis:
https://issues.redhat.com/browse/WFLY-4990


darranl:
@moulalis Issues such as WFLY-4990 are not just about removing the @Ignore annotation - if that was all that was required we would have done it at the time.

In most cases where we add the annotation it would be for one of two reasons: -

1. Intermittent failures in our CI environment making many PR runs fail.
2. A change in implementation where the test was heavily dependent on the old implementation rather than the feature being tested.

In both cases the Jira issue is more about taking the time to understand the original problem and provide appropriate fixes either to get the test passing or to get the test stable again.

Any fixes required could either be to the test case itself or they may be required within the implementation of the feature under test.

The failures reported in CI are directly related to the test case that was reactivated so further work does appear required before they can be activated, for this reason I am going to close this PR for now.  Please feel free to reopen if you work through the remaining issues.

darranl:
@moulalis FYI if you would like to speak to the WildFly team to discuss the requirements for these or any other open issues you can also join us in Zulip at https://wildfly.zulipchat.com/ or we also have a developer mailing list at https://lists.jboss.org/mailman/listinfo/wildfly-dev which you can join.

PR #12954, opened 2020-01-31
[WFLY-4991] Testcase Unignored and unused import removed

moulalis:
https://issues.redhat.com/browse/WFLY-4991


jmesnil:
duplicates #12954

PR #12956, opened 2020-02-03
Fix for WFLY-13050, metrics should be optional in observability layer

jfdenise:
* Layer spec: Optional dependencies
* Doc: Moved observability aggregation layer to decorator sections. Updated description.

### DIFF
--- a/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
+++ b/galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
@@ -1,8 +1,9 @@
 <?xml version="1.0" ?>
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="observability">
     <dependencies>
+        <!-- required by health subsystem -->
         <layer name="microprofile-config"/>
-        <layer name="microprofile-metrics"/>
+        <layer name="microprofile-metrics" optional="true"/>
         <layer name="open-tracing" optional="true"/>
     </dependencies>
     <feature-group name="health"/>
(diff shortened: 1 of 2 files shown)


galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" ?>
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="observability">
     <dependencies>
-        <layer name="microprofile-config"/>
-        <layer name="microprofile-metrics"/>
+        <layer name="microprofile-config" optional="true"/>
```
bstansberry: I'm reluctant to make config optional. The odds are very high that someone trying to exclude it will generate a broken config.  OpenTracing doesn't require it, but if all you want is OT then just specify that.

bstansberry: Further to my last point; this layer has the health subsystem as a fixed feature group, not an excludable layer and health requires config.

bstansberry (changes requested):

galleon-pack/src/main/resources/layers/standalone/observability/layer-spec.xml
jfdenise: I missed, that I added a comment and removed the optional attribute.

bstansberry (dismissed):

bstansberry:
@jfdenise Can you rebase and move this change to the file that is now in ee-galleon-pack?

jfdenise:
@bstansberry , thank-you for having done it.

PR #12957, opened 2020-02-04
[WFLY-13052] MicroProfile Config TCK ShouldThrowException

jmesnil:
Add an Arquillian DeploymentExceptionTransformer to ensure that a
DeploymentException is properly thrown when the deployment on WildFly
Server is unsuccessful (the actual cause of the deployment failure is
lost but they are not asserted by the tests that uses
@ShouldThrowException.

* Remove corresponding excluded tests from tck-suite.xml

JIRA: https://issues.redhat.com/browse/WFLY-13052


rhusar (approved):
👍  Better than not running the tests at all.

xstefank (approved):
I don't think that any of the specs is actually testing the exact messages of DeploymentExceptions so I think this is a good solution to the problem (not saying that the arquillian adapter issue shouldn't be resolved). The important point is that the deployment failed.

PR #12961, opened 2020-02-06
[WFLY-13032] Add missing host exclusions and update related test case

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13032


bstansberry:
@yersan With #12927 merged this will need some rework. (I think you know all this; I'm just writing it down in case.)

1) https://github.com/wildfly/wildfly/pull/12961/files#diff-a7ad39e72966a67ce48b7ada66d69701 shouldn't include anything about FT, JWT or OpenAPI. (If #12927 left FT in there, please use this PR to correct that.)

2) https://github.com/wildfly/wildfly/blob/master/ee-galleon-pack/src/main/resources/feature_groups/domain-host-excludes.xml and https://github.com/wildfly/wildfly/blob/master/galleon-pack/src/main/resources/feature_groups/domain-host-excludes.xml now both exist. Both need the additional release version stuff, but the ee-galleon-pack variant does not get FT, JWT or OpenAPI. (Again if #12927 left FT in the ee-galleon-pack variant please correct that in this PR.)

Thanks!

bstansberry (approved):

PR #12964, opened 2020-02-07
[WFLY-2102] Sanitizing name resource for env and adding warning log

lvydra:
Sanitizing name prefixed with a '/' and logging warning message

Issue: https://issues.redhat.com/browse/WFLY-2102


bstansberry:
this is ok to test

bstansberry:
@emmartins Please review.

darranl:
@emmartins Are you able to provide the requested review on this PR?  This looks like if approved it should be fairly straightforward to merge to get it out of the queue and into WildFly.

darranl:
FYI also kicked off CI again to double check we are Ok there.

emmartins:
@darranl @bstansberry the patch looks ok, just not sure if the issue should be solved like this, silently (fix the env/ concat to name without warning), or continue to fail with a better error msg. May be worth to re-check specs and how RI handles such names, I'm not currently naming comp leader but can do it, may take some time tho... Thoughts?

emmartins:
Thinking a bit more on it, if opting for this new behaviour, it would be preferable performance wise to address the issue when env/ is added to the name string

lvydra:
@emmartins  Hi, OK, I will modify it accordingly. Thank you

lvydra:
Cc @emmartins

emmartins:
@lvydra @darranl @bstansberry LGTM

bstansberry:
Thanks, @lvydra !

PR #12965, opened 2020-02-07
[WFLY-13023] Upgrade MicroProfile Config 1.4

jmesnil:
* Upgrade smallrye-config 1.6.0
* Update the GAV of smallrye-config artifacts

JIRA: https://issues.redhat.com/browse/WFLY-13023

---

* [MicroProfile Config 1.4 Release Notes](https://github.com/eclipse/microprofile-config/milestone/7?closed=1)
* [smallrye-config 1.6.0 Release Notes](https://github.com/smallrye/smallrye-config/releases/tag/1.6.0)

### DIFF
--- a/feature-pack/src/license/full-feature-pack-licenses.xml
+++ b/feature-pack/src/license/full-feature-pack-licenses.xml
@@ -604,39 +604,6 @@
     </dependency>
     <dependency>
       <groupId>io.smallrye</groupId>
-      <artifactId>smallrye-config</artifactId>
-      <licenses>
-        <license>
-          <name>Apache License 2.0</name>
-          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
-          <distribution>repo</distribution>
-        </license>
-      </licenses>
-    </dependency>
-    <dependency>
-      <groupId>io.smallrye</groupId>
-      <artifactId>smallrye-config-common</artifactId>
-      <licenses>
-        <license>
-          <name>Apache License 2.0</name>
-          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
-          <distribution>repo</distribution>
-        </license>
-      </licenses>
-    </dependency>
-    <dependency>
-      <groupId>io.smallrye.config</groupId>
-      <artifactId>smallrye-config-source-file-system</artifactId>
-      <licenses>
-        <license>
-          <name>Apache License 2.0</name>
-          <url>http://www.apache.org/licenses/LICENSE-2.0</url>
-          <distribution>repo</distribution>
-        </license>
-      </licenses>
-    </dependency>
-    <dependency>
-      <groupId>io.smallrye</groupId>
       <artifactId>smallrye-health</artifactId>
       <licenses>
         <license>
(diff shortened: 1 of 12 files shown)


jmesnil:
Dev analysis is at https://github.com/wildfly/wildfly-proposals/pull/285

jmesnil:
I put this PR on hold, WildFly integration test suite detected a regression in smallrye-config (tracked by https://github.com/smallrye/smallrye-config/issues/236)

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/app/MicroProfileConfigTestCase.java
```diff
@@ -200,9 +200,6 @@ public void testGetIntegerProperties() throws Exception {
             assertTextContainsProperty(text, "integerDefault", -42);
             assertTextContainsProperty(text, SubsystemConfigSourceTask.INTEGER_OVERRIDDEN_PROP_NAME,
                     SubsystemConfigSourceTask.INTEGER_OVERRIDDEN_PROP_VALUE);
-
-            assertTextContainsProperty(text, "intBadValue", 0);
```
marekkopecky: What is the reason for removing this asserts? Is it related with https://github.com/eclipse/microprofile-config/issues/446 ?

marekkopecky (changes requested):

bstansberry:
@jmesnil Once this is ready please file the same change against the 19.0.x branch up move the 19.x label to that PR. Thanks!

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/app/MicroProfileConfigTestCase.java
jmesnil: this assert was incorrect and revealed a bug in smallrye-config.

When the config property had a default value that could not be converted to the expected type, smallrye-config set it to the "default value" of the types (e.g. `0` for `int`).

With https://github.com/smallrye/smallrye-config/issues/211, it has been fixed. The default value will be converted and this throw an exception that would prevent the deployment to be deployed.

feature-pack/src/license/full-feature-pack-licenses.xml
```diff
@@ -603,7 +603,7 @@
       </licenses>
     </dependency>
     <dependency>
-      <groupId>io.smallrye</groupId>
+      <groupId>io.smallrye.config</groupId>
```
bstansberry: Please re-alphabetize. (I think it's important as EAP productization files PRs to fix it any time it's not in order.)

feature-pack/src/license/full-feature-pack-licenses.xml
```diff
@@ -614,7 +614,7 @@
...
```
bstansberry: Please re-alphabetize.

bstansberry (changes requested):

feature-pack/src/license/full-feature-pack-licenses.xml
jmesnil: done

feature-pack/src/license/full-feature-pack-licenses.xml
jmesnil: done

marekkopecky:
@jmesnil I'm looking to new spec, and it contains this section:
```
Cleaning up a ConfigSource

If a ConfigSource implements the java.lang.AutoCloseable interface then the close() method will be called when the underlying Config is being released.
```

I call this code (in both deployment and standalone application):
```java
        ConfigProviderResolver resolver = ConfigProviderResolver.instance();
        ConfigBuilder builder = resolver.getBuilder();
        Config config = builder.addDefaultSources().withSources(new CustomConfigSource()).build();
        resolver.releaseConfig(config);
```

And it seems that close method is not called. Can you please check it?

jmesnil:
> And it seems that close method is not called. Can you please check it?

Yes, that is correct. `close` is not called. David raised the issue in https://github.com/eclipse/microprofile-config/issues/136#issuecomment-535962313 and mention it in [smallrye-config](https://github.com/smallrye/smallrye-config/blob/70912e9e8f34e42ee7a5616849fa598085e8b64a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java#L136).

marekkopecky:
@jmesnil Is the same valid also for Converters? I don't see description of Converters in  https://github.com/eclipse/microprofile-config/issues/136
but spec contains this:
```
Cleaning up a Converter

If a Converter implements the java.lang.AutoCloseable interface then the close() method will be called when the underlying Config is being released.
```

jmesnil:
That's correct, converters have the same issue than config source regarding to when and how they can be properly closed. That's a good point to mention converters in https://github.com/eclipse/microprofile-config/issues/136 too

bstansberry (approved):

PR #12974, opened 2020-02-10
WFLY-13074 Don't start the app context unless the server is running.

soul2zimate:
https://issues.redhat.com/browse/WFLY-13074

analysis is in the downstream 7.2.z issue, thanks.

### DIFF
--- a/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
+++ b/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
@@ -33,6 +33,7 @@ import io.undertow.servlet.api.Deployment;
 import org.jboss.as.clustering.context.DefaultThreadFactory;
 import org.jboss.as.server.suspend.ServerActivity;
 import org.jboss.as.server.suspend.ServerActivityCallback;
+import org.jboss.as.server.suspend.SuspendController.State;
 import org.jboss.logging.Logger;
 import org.jboss.modcluster.container.Connector;
 import org.jboss.modcluster.container.ContainerEventHandler;
@@ -113,10 +114,16 @@ public class UndertowEventHandlerAdapterService implements UndertowEventListener
 
     private synchronized void onStart(Context context) {
         ContainerEventHandler handler = this.configuration.getContainerEventHandler();
+
         handler.add(context);
 
+        State state = this.configuration.getSuspendController().getState();
+
         // TODO break into onDeploymentAdd once implemented in Undertow
-        handler.start(context);
+        if(state == State.RUNNING) {
+            handler.start(context);
+        }
+
         this.contexts.add(context);
     }
 


soul2zimate:
@pferraro  @rhusar  Could you please review this ?

rhusar:
@soul2zimate Will do soon. Quick question, does unsuspend re-enable the context?

soul2zimate:
Yes, because `handler.add(context); ` is always executed in the onStart() method.  When user resumes the server, context is re-enabled in resume() method in https://github.com/wildfly/wildfly/blob/19.0.0.Beta2/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java#L191-L196

soul2zimate:
Could anyone from clustering team review this change ? I need this approved and merged before bring up the one for the downstream issue.

rhusar:
@soul2zimate Sorry for the delay, we were in the meeting last week, I ll test and review this this week.

spyrkob:
@rhusar did you have a chance to review this? We're getting ready to tag CP8 next week and would like to include this if possible.

rhusar (approved):
I am approving this, as this solves whether the events are sent or not. 

However, I wonder if this fix is really effective? Even with this patch, the contexts are shown as enabled in the Undertow LB implementation.

```
[standalone@127.0.0.1:10490 /] /subsystem=undertow/configuration=filter/mod-cluster=load-balancer/balancer=mycluster/node=node1:read-resource(include-runtime,recursive
{
    "outcome" => "success",
    "result" => {
        "aliases" => [
            "default-host",
            "localhost"
        ],
        "cache-connections" => 40,
        "elected" => 0,
        "flush-packets" => false,
        "load" => 66,
        "load-balancing-group" => undefined,
        "max-connections" => 40,
        "open-connections" => 4,
        "ping" => 10,
        "queue-new-requests" => true,
        "read" => 0L,
        "request-queue-size" => 1000,
        "status" => "NODE_UP",
        "timeout" => 0,
        "ttl" => 60L,
        "uri" => "ajp://127.0.0.1:8009/?#",
        "written" => 0L,
        "context" => {
            "/clusterbench-granular" => {
                "requests" => 0,
                "status" => "enabled"
            },
            "/clusterbench-passivating" => {
                "requests" => 0,
                "status" => "enabled"
            },
            "/clusterbench" => {
                "requests" => 0,
                "status" => "enabled"
            },
            "/" => {
                "requests" => 0,
                "status" => "enabled"
            },
            "/wildfly-services" => {
                "requests" => 0,
                "status" => "enabled"
            }
        }
    }
}
```

soul2zimate:
This just makes sure that the context is not started when server is started in suspend mode.

In regard to the attribute value, it is how current [ModClusterService](https://github.com/modcluster/mod_cluster/blob/1.4.1.Final/core/src/main/java/org/jboss/modcluster/ModClusterService.java) implemented, we can enable the context either `start` or `add` a context, but we can only disable context when we `stop` the context. Do you think there is anything need to be changed in mod_cluster ?

rhusar:
Opened https://issues.redhat.com/browse/WFLY-13237 to investigate my previous comment.

bstansberry:
retest this please

PR #12971, opened 2020-02-09
[WFLY-13050] Metrics should be optional in the observability layer

bstansberry:
This is just a conflict resolution of #12956

https://issues.redhat.com/browse/WFLY-13050

PR #12975, opened 2020-02-10
[WFLY-13072] Limit HostExcludesTestCase extension searching to module…

bstansberry:
…s/system/layers

Work around the problem discussed at https://github.com/wildfly/wildfly/pull/12971#issuecomment-583973140

This is probably a better approach anyway.


bstansberry:
Note that this "works around the problem discussed at #12971" in two ways:

1) It limits the module scanning to those in system/layers. I believe the problem I noted with the previous version of this was because some other test was leaving behind a test module in modules/ and that was what was causing the failures. Our standard modules were not triggering MODULES-377

2) I added a catch block for ModuleLoadException. So even if one of our standard modules did trigger MODULES-377, unless it was an Extension module that the test really needs to load, the test will still work despite the exception.

yersan (approved):

PR #13000, opened 2020-02-17
[WFLY-13102] assertion error is expected when ejb is not available

ochaloup:
https://issues.redhat.com/browse/WFLY-13102

Fix for intermittent failure for the testcase where exception is expected but testsuite is run with {{-ea}} and {{AssertionError}} may be coming from the {{EJBClientInvocationContext}} class. It's expected by the test behaviour that EJB fails to invoke the remote EJB.


bstansberry:

chengfang (approved):

PR #12983, opened 2020-02-12
[WFLY-13048] Upgrade MicroProfile Health 2.2

jmesnil:
JIRA: https://issues.redhat.com/browse/WFLY-13048

Diff to previous version integrated in WildFly: https://github.com/eclipse/microprofile-health/compare/2.1...2.2

Signed-off-by: Jeff Mesnil <jmesnil@redhat.com>

PR #13004, opened 2020-02-17
[WFLY-13101] Tests for Microprofile Health 2.2

fabiobrz:
This PR adds a test case for validating `HealthCheckResponse` class, which has been defined as concrete in the latest MicroProfile Health 2.2 release. This is basically the main change introduced by this last version of the spec from the end user point of view.

* Please be aware that this PR depends on https://github.com/wildfly/wildfly/pull/12983 and thus _shouuld not_ be merged before the latter is, as well.
 
* JIRA issue: https://issues.redhat.com/browse/WFLY-13101

This test case was run against the Wildfly feature branch which already has the MicroProfile/SmallRye upgrade, see [1].

The test failed, because there's a mismatch in the spec between a property name and the actual payload JSON schema. An issue to MicroProfile Health project has been filed [2] and the test case has been marked as ignored for this reason.

[1]
https://github.com/jmesnil/wildfly/tree/WFLY-13048_upgrade_microprofile-health_2.2

[2]
https://github.com/eclipse/microprofile-health/issues/243

- [x] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue(s)
- [x] Pull Request contains description of the issue(s)
- [x] Pull Request does not include fixes for issues other than the main ticket
- [x] Attached commits represent units of work and are properly formatted

### DIFF
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/ConcreteHealthCheckResponseTestCase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/ConcreteHealthCheckResponseTestCase.java
@@ -1,0 +1,100 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.test.integration.microprofile.health;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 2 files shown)


bstansberry:
this is ok to test

bstansberry:
@jmesnil If these changes are ok can you pull them into your branch(es) for the upgrade? Your submitting the commits as part of your PR counts as the necessary approval of them.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/ConcreteHealthCheckResponseTestCase.java
```diff
@@ -0,0 +1,100 @@
...
+            HttpUriRequest request = new HttpGet(
+                    String.format("http://%s:%d/health", managementClient.getMgmtAddress(), managementClient.getMgmtPort()));
+
+            try (CloseableHttpResponse response = client.execute(request)) {
+                Assert.assertEquals(
+                        "Health check endpoint call failed",
+                        HttpServletResponse.SC_OK, response.getStatusLine().getStatusCode());
+
+                final String responseContent = EntityUtils.toString(response.getEntity());
+                HealthCheckResponse typedResponse = new ObjectMapper().readValue(
```
jmesnil: -1, the spec does not specify that the JSON schema must map to the `HealthCheckResponse`.

jmesnil (changes requested):
I'm -1 on this test.

The MP Health TCK already uses extensively the `HealthCheckResponse `class.
This test does not bring any additional assertions to what is already tested by the TCK.

fabiobrz:
Hi @jmesnil although the spec doesn't mandate for the `HealthCheckResponse` class to map to the protocol JSON schema, I see that the use case [1] which was referred by the PR [2] that brought the change in is actually about deserialization with - for instance - REST Client. This would help the end user (Wildfly end user, in this very case) to leverage such tools easily. WDYT?
Also @xstefank what are actual plans on `HealthCheckResponse` class in future? Is mapping to the JSON schema going to be a requirement? 

[1]
https://github.com/eclipse/microprofile-health/issues/194

[2]
https://github.com/eclipse/microprofile-health/pull/218

xstefank:
@jmesnil I believe this is an oversight on the spec side. I will raise this in the community. However, this kind of test can be added as an actual new TCK test in my opinion. So I agree that this test is not necessary for wfly test suite.

marekkopecky:
> The MP Health TCK already uses extensively the HealthCheckResponse class.
This test does not bring any additional assertions to what is already tested by the TCK.

@jmesnil It seems that TCK uses JSON-P, not JSON-B/Jackson

jmesnil:
@marekkopecky that's correct. One more reason to move that test in the microprofile-health tck to make sure that https://github.com/eclipse/microprofile-health/issues/243 is properly addressed.

However that does not change my opinion that this test does not belong to WildFly integration test suite.

fabiobrz:
PR can be closed as necessary test coverage will be added to https://github.com/eclipse/microprofile-health, see https://github.com/eclipse/microprofile-health/issues/243
Thanks for your feedback @jmesnil, @xstefank, @marekkopecky and @bstansberry

PR #12882, opened 2020-01-07
[WFLY-12931] Upgrade Undertow schema, improve documentation and tests

fl4via:
Upgrade schema and other small tasks needed for WFLY-12473

Jiras:
https://issues.redhat.com/browse/WFLY-12931
https://issues.redhat.com/browse/WFLY-12932
https://issues.redhat.com/browse/WFLY-12933

PR #13008, opened 2020-02-18
[WFLY-13065] Options for reverse-proxy max-request-time and connection-idle-timeout are specified as seconds

rmartinc:
Fix for https://issues.redhat.com/browse/WFLY-13065. Comments:

* It requires a bump for undertow xml version to 11 (I have cloned the changes previously done in 10).
* The only difference in the XML is that all the default values for the `reverse-proxy-handlerType` have been added.
* Test added to check the `max-request-time` with a very low value and check the connection is aborted (503 returned).
* I have also changed the creation of the `ProxyHandler` to use the builder (to avoid the deprecated warning). Here there are two values (`rewrite-host-header` and `reuse-x-forwarded`) that could also be added to the wildfly configuration (now they are always false). Just in case you want to use the same bump for them.

No hurry with this one. It's a minor thing and the bump complicates it a bit.

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/handlers/ReverseProxyHandler.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/handlers/ReverseProxyHandler.java
@@ -53,6 +53,7 @@ public class ReverseProxyHandler extends Handler {
             .setRequired(false)
             .setAllowExpression(true)
             .setDefaultValue(new ModelNode(30))
+            .setMeasurementUnit(MeasurementUnit.SECONDS)
             .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
             .build();
 
@@ -73,7 +74,7 @@ public class ReverseProxyHandler extends Handler {
             .setRequired(false)
             .setAllowExpression(true)
             .setDefaultValue(new ModelNode(-1))
-            .setMeasurementUnit(MeasurementUnit.SECONDS)
+            .setMeasurementUnit(MeasurementUnit.MILLISECONDS)
             .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
             .build();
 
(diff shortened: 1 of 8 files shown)


yersan:
retest this please

bstansberry:
@rmartinc Thanks! We require that the schema/xsd version bump commits be in a separate commit from the functional change, so we can clearly distinguish the two.

@fl4via already has commits for a version bump in #12882. Please coordinate with her as to how to the version bump.

rmartinc:
@bstansberry OK thanks! I'll contact @fl4via this afternoon to send her the little xsd changes this issue needs. Then I'll redo the PR.

fl4via:
@bstansberry  can this PR be converted as a dependent on #12882 so that the other one can also be merged? It contains improvements on the Undertow subsystem that don't depend on RFEs and are ready to be merged as they are

rmartinc:
@fl4via I still don't see the default values for `reverse-proxy-handlerType` in the XSD of your PR. What's the idea finally? If you add them to the in your bump my PR is reduced to a trivial one. I remember that we chatted about doing that, but I'm not updating this PR because I don't see your changes. :-)

fl4via (approved):

fl4via:
This PR is ready to go, @rmartinc  rebased it and removed the addition of the new schema version because it was already bumped by a previous commit.

undertow/src/main/java/org/wildfly/extension/undertow/handlers/ReverseProxyHandler.java
```diff
@@ -95,8 +96,8 @@
     public static final AttributeDefinition CONNECTION_IDLE_TIMEOUT = new SimpleAttributeDefinitionBuilder(Constants.CONNECTION_IDLE_TIMEOUT, ModelType.INT)
             .setRequired(false)
             .setAllowExpression(true)
-            .setDefaultValue(new ModelNode(60L))
-            .setMeasurementUnit(MeasurementUnit.SECONDS)
+            .setDefaultValue(new ModelNode(60000))
```
bstansberry, line 99: We need a transformer to deal with this one. If this isn't configured a server running the current version will use 60000 and one running a previous version will us 60. The servers should be consistent, so the transformer would explicitly tell the legacy server to use 60000.

Unfortunately "Attribute has a different default value" at https://docs.wildfly.org/19.1/Extending_WildFly.html#common-transformation-use-cases doesn't explicitly illustrate how to do this

rmartinc:
I have added the transformer for the default value in the bump from version 10 to 11. Let's see the tests!

bstansberry (approved):

PR #13012, opened 2020-02-18
WFLY-13122 Include MicroProfile Fault Tolerance subsystem in the defa…

rhusar:
…ult server profiles

Resolves
https://issues.redhat.com/browse/WFLY-13122


darranl:
I think this one had some follow up considerations based on the comments form @bstansberry on the 19.0.x PR so I am going to add it to our agenda for Monday,

rhusar:
Won't do – we have a microprofile server profile which includes this by default already. The documentation explains how to configure the subsystem in an existing server profile if the users chooses to use MP FT.

PR #13014, opened 2020-02-19
[WFLY-13127] Upgrade smallrye openapi to 1.2.0 to bring in https://gi…

bstansberry:
…thub.com/smallrye/smallrye-open-api/issues/248 fix

https://issues.redhat.com/browse/WFLY-13127


bstansberry:
I removed the hold label. A concern I had with this is it moved this lib to a version that has a dependency on smallrye-config 1.5.0, while WF was on 1.4.1. But now WF is on 1.6.1.

bstansberry:
I'll point this on hold until Paul Ferraro determines if it's the right way to go. He can remove the label if it is.

PR #13016, opened 2020-02-19
[WFLY-13128] Add a Galleon layer for batch-jberet

bstansberry:
https://issues.redhat.com/browse/WFLY-13128

@chengfang @jamezp  @jfdenise FYI

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
@@ -123,6 +123,7 @@ Basic layers:
 
 Basic layers:
 
+* _batch-jberet_: Tech preview support for Jakarta Batch. Depends on _cdi_.
 * _bean-validation_: Support for Jakarta Bean Validation. Optionally depends on _cdi_.
 * _cdi_: Support for CDI. Optionally depends on _bean-validation_.
 * _datasources_: Support for datasources.
--- a/ee-galleon-pack/src/main/resources/feature_groups/batch-jberet-elytron.xml
+++ b/ee-galleon-pack/src/main/resources/feature_groups/batch-jberet-elytron.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="batch-jberet-elytron" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature-group name="batch-jberet">
+        <feature spec="subsystem.batch-jberet">
+            <param name="security-domain" value="ApplicationDomain"/>
+        </feature>
+    </feature-group>
+</feature-group-spec>
--- a/ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
+++ b/ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="batch-jberet">
+    <dependencies>
+        <!-- CDI should be optional but to be conservative in the first cut at this, require it
+        <layer name="cdi" optional="true"/>
+        -->
+        <layer name="cdi"/>
+        <layer name="ee"/>
+        <layer name="elytron"/>
+        <layer name="transactions"/>
+    </dependencies>
+    <feature-group name="batch-jberet-elytron"/>
+</layer-spec>
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1150,6 +1150,45 @@
                                     </configurations>
                                 </configuration>
                             </execution>
+                            <!-- Provision a cloud-profile server with batch-jberet as well -->
+                            <execution>
+                                <id>batch-jberet-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${project.build.directory}/wildfly-batch</install-dir>
+                                    <record-state>false</record-state>
+                                    <log-time>${galleon.log.time}</log-time>
+                                    <offline>true</offline>
+                                    <plugin-options>
+                                        <jboss-maven-dist/>
+                                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                                        <optional-packages>passive+</optional-packages>
+                                    </plugin-options>
+                                    <feature-packs>
+                                        <feature-pack>
+                                            <groupId>${project.groupId}</groupId>
+                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
+                                            <version>${testsuite.ee.galleon.pack.version}</version>
+                                            <inherit-configs>false</inherit-configs>
+                                            <inherit-packages>false</inherit-packages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <configurations>
+                                        <config>
+                                            <model>standalone</model>
+                                            <name>standalone.xml</name>
+                                            <layers>
+                                                <layer>cloud-server</layer>
+                                                <layer>h2-default-datasource</layer>
+                                                <layer>batch-jberet</layer>
+                                            </layers>
+                                        </config>
+                                    </configurations>
+                                </configuration>
+                            </execution>
                         </executions>
                     </plugin>
                     <plugin>
@@ -1282,6 +1321,39 @@
                                         <include>org/jboss/as/test/integration/security/loginmodules/Ldap*TestCase.java</include>
                                         <include>org/jboss/as/test/integration/jca/moduledeployment/*TestCase.java</include>
                                     </includes>
+                                </configuration>
+                            </execution>
+                            <!-- Tests against the install with batch-jberet -->
+                            <execution>
+                                <id>batch-jberet-layers.surefire</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <systemPropertyVariables>
+                                        <jboss.install.dir>${basedir}/target/wildfly-batch</jboss.install.dir>
+                                        <jboss.home>${project.build.directory}/wildfly-batch</jboss.home>
+                                        <jboss.home.dir>${project.build.directory}/wildfly-batch</jboss.home.dir>
+                                        <jbossas.dist>${project.build.directory}/wildfly-batch</jbossas.dist>
+                                        <jboss.dist>${project.build.directory}/wildfly-batch</jboss.dist>
+                                        <!-- Override the standard module path that points at the shared module set from dist -->
+                                        <module.path>${project.build.directory}/wildfly-batch/modules${path.separator}${basedir}/target/modules</module.path>
+                                    </systemPropertyVariables>
+                                    <environmentVariables>
+                                        <JBOSS_HOME>${project.build.directory}/wildfly-batch</JBOSS_HOME>
+                                    </environmentVariables>
+                                    <redirectTestOutputToFile>true</redirectTestOutputToFile>
+                                    <includes>
+                                        <include>org/jboss/as/test/integration/batch/**/*TestCase.java</include>
+                                    </includes>
+                                    <excludes>
+                                        <!-- Test app has a manifest dep on org.jboss.remoting3 module -->
+                                        <exclude>org/jboss/as/test/integration/batch/chunk/ChunkPartitionTestCase.java</exclude>
+                                        <exclude>org/jboss/as/test/integration/batch/deployment/JobControlTestCase.java</exclude>
+                                        <!-- Test requires agroal -->
+                                        <exclude>org/jboss/as/test/integration/batch/repository/JdbcRepositoryTestCase.java</exclude>
+                                    </excludes>
                                 </configuration>
                             </execution>
                         </executions>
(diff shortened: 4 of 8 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/deployment/JobXmlVisibilityTestCase.java
```diff
@@ -216,7 +215,7 @@ private static JavaArchive createEjb(final String name) {
         final JavaArchive jar = ShrinkWrap.create(JavaArchive.class, name)
                 .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml")
                 // Add at least one EJB annotated class to make this an EJB JAR
-                .addClasses(CountingItemReader.class, CountingItemWriter.class, BatchExecutionService.class);
+                .addClasses(CountingItemReader.class, CountingItemWriter.class, SimpleEJB.class);
```
bstansberry: @jamezp I realize now that I forget to exclude this test at https://github.com/wildfly/wildfly/pull/13016/files#diff-ec12efbf5a13385ba26a3162a8b61e72R1190 but it still passes. It runs against a server with no EJB and still passes. It's not a surprise the deployment doesn't fail because of no EJB, as having no subsystem means the SimpleEJB is not processed. I just don't know whether the test is still meaningful; i.e. should I exclude it?

jamezp: I'll need to debug it to see what's going on here. I _think_ the point of this was to ensure it was processed as a sub-deployment. Since this is only used in an EAR that is probably why it works. I do need to test to see what happens when the EJB JAR is in a WAR.

That said I'm comfortable doing whatever you think is appropriate here. I will look into the issue I described above, but I think that's separate from what you're doing. TBH I'm fine with removing the `SimpleEJB` all together as I'm not sure what I was thinking when I thought an EJB was required :)

jamezp (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/deployment/JobXmlVisibilityTestCase.java
jamezp: Actually the EJB isn't required at all. I think I used 

> Add at least one EJB annotated class to make this an EJB JAR

 because I was using EJB's for the tests. The spec does not require it to be an EJB at all. It can be any JAR. I think we can remove the comment and the `SimpleEJB`. This test is really just testing that the deployment model can see the correct job XML files.

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
```diff
@@ -123,6 +123,7 @@ Basic layers:
 
 Basic layers:
 
+* _batch-jberet_: Tech preview support for Jakarta Batch. Optionally depends on _cdi_.
```
jfdenise: Seems that you changed to be non optional

testsuite/layers/pom.xml
```diff
@@ -1004,6 +1004,42 @@
...
                     </execution>
                     <!-- wildfly layers -->
+                    <execution>
+                        <id>batch-jberet-provisioning</id>
+                        <goals>
+                            <goal>provision</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <install-dir>${layers.install.dir}/bean-validation</install-dir>
```
jfdenise: Your install-dir is not correct bean-validation ==> batch-jberet.

jfdenise (changes requested):

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
bstansberry: Ah, yeah; thanks!

ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="batch-jberet">
+    <dependencies>
+        <!-- CDI should be optional but to be conservative in the first cut at this, require it
+        <layer name="cdi" optional="true"/>
+        -->
+        <layer name="cdi"/>
+        <layer name="ee"/>
+        <layer name="transactions"/>
```
bstansberry: @jamezp @chengfang @jfdenise 

A bit of a design/best practices question for you...

The batch-jberet subsystem supports a 'jdbc-repository' element which in turn requires that a datasource be configured.

I don't include any datasource related layer here because the 'batch-jberet' feature group is not using any 'jdbc-repository'.  If the user wanted to use that they would need to provision an actual DS (including the driver) so there's no benefit here to having an optional 'datasource' layer dep and always generating a bunch of config and pulling in other modules that are not used and are not even usable without further user effort.

I think I'm right but am pointing it out in case I'm wrong!

Another 'best practices' question is I declare a dep on 'ee' even though it is transitively required via 'transactions'.  But I see now I don't declare 'elytron' even though it is needed and is also pulled via transactions. My feeling is I should list them both. I think it's better practice to list non-obvious deps rather than relying on transitive ones. Particularly 'elytron', where one could imagine a future where someone would not require 'elytron' in 'transactions' and this would break.  I don't think it's necessary to spell out all deps everywhere (e.g. something the depends on jax-rs doesn't need to spell out a webserver dep), but for subtle cases it's probably better to be explicit.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/deployment/JobXmlVisibilityTestCase.java
jamezp: After another look I think we need to leave the test as you have it. In other words what you have is what we need.

ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
jamezp: I agree with this. I don't see the need to the datasource, but Elytron does make sense.

I think EE also makes sense as there are modules from EE that I believe are used in deployment processing.

bstansberry: I added elytron.

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
bstansberry: Fixed

testsuite/layers/pom.xml
bstansberry: Fixed

ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
```diff
@@ -0,0 +1,13 @@
...
+    <dependencies>
+        <!-- CDI should be optional but to be conservative in the first cut at this, require it
+        <layer name="cdi" optional="true"/>
+        -->
+        <layer name="cdi"/>
+        <layer name="ee"/>
+        <layer name="elytron"/>
+        <layer name="transactions"/>
+    </dependencies>
+    <feature-group name="batch-jberet"/>
```
jfdenise: @bstansberry , jberet feature-group doesn't secure jberet with elytron security-domain. In The example CLI script we are defining it is. I am wandering if we should secure jberet in the script: https://github.com/wildfly/wildfly-proposals/pull/294#discussion_r381855034

bstansberry:
There is some question about what the proper configuration should be with respect to security integration. In an integration configured for elytron we need the change here:

https://github.com/wildfly/wildfly/blob/master/feature-pack/src/main/resources/content/docs/examples/enable-elytron.cli#L9

Generally I want our layers oriented toward an elytron-based security configuration, so perhaps adding that should be part of the standard batch-jberet layer. But we need to think a bit about what we should do if the legacy-security layer is involved.

ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
bstansberry: Just to close out this conversation -- the discussion on wildfly-proposals says that adding that bit of config is correct when elytron is used. Which raises the question of how this layer should deal with that. I posted a separate general comment raising that question.

darranl:
@bstansberry FYI this one currently has a merge conflict but should we try and discuss this next week?

I suspect we want any layering changes in ASAP for WildFly 20 to give us time in case we need follow ups.

bstansberry:
@darranl I added the security domain setting as I think any layer should be designed to work properly with elytron, unless it's specifically meant for legacy use. This one is not meant for legacy use.

https://github.com/wildfly/wildfly/pull/13016/files#diff-c0d74669b8d8d3de259de044c118c547R13

If someone wants to use this in conjunction with legacy security it's possible to unset that param in a provisioning.xml:

```
<installation xmlns="urn:jboss:galleon:provisioning:3.0">
    <feature-pack location="wildfly@maven(org.jboss.universe:community-universe):current#21.0.0.Beta1-SNAPSHOT">
        <default-configs inherit="false"/>
        <packages inherit="false"/>
    </feature-pack>
    <config model="standalone" name="standalone.xml">
        <layers>
            <include name="batch-jberet"/>
            <include name="legacy-security"/>
        </layers>
		<feature spec="subsystem.batch-jberet">
	        <unset param="security-domain"/>
		</feature>
    </config>
    <options>
        <option name="optional-packages" value="passive+"/>
    </options>
</installation>
```

I don't see any way to do that using a layer; i.e. another layer this one optionally depends on that just adds that param, so you exclude the optional dep and that param goes away.  Maybe there's a way but I doubt I'd like such a layer anyway. Too much.

@jfdenise I'm interested in your opinion on whether https://github.com/wildfly/wildfly/pull/13016/files#diff-c0d74669b8d8d3de259de044c118c547R13 is the best way, versus other possible things like:

1) Add another feature group that adds that param and include that in the layer; i.e. don't do the param setting in the layer-spec.xml.

2) Add that param to the batch-jberet.xml file (i.e. to the feature-group spec) and then anyplace it's used where we don't want it (e.g. standalone.xml) then unset the param.

jfdenise:
@bstansberry , 
1) I would not touch at the existing batch-jberet.xml feature-group if we don't want to evolve the default configurations.
2) Could be cleaner to define a batch-jberet-elytron.xml if we want to share this logic in the future (inside a config or inside another layer).

bstansberry:
@jfdenise Thanks. I added batch-jberet-elytron.xml and used it in the layer.

testsuite/layers/pom.xml
```diff
@@ -2065,6 +2101,7 @@
                                             <model>standalone</model>
                                             <name>standalone.xml</name>
                                             <layers>
+												<layer>batch-jberet</layer>
```
yersan, line 2104: Tabs vs spaces

testsuite/integration/basic/pom.xml
```diff
@@ -1284,6 +1323,39 @@
...
+                                        <jboss.home.dir>${project.build.directory}/wildfly-batch</jboss.home.dir>
+                                        <jbossas.dist>${project.build.directory}/wildfly-batch</jbossas.dist>
+                                        <jboss.dist>${project.build.directory}/wildfly-batch</jboss.dist>
+                                        <!-- Override the standard module path that points at the shared module set from dist -->
+                                        <module.path>${project.build.directory}/wildfly-batch/modules${path.separator}${basedir}/target/modules</module.path>
+                                    </systemPropertyVariables>
+                                    <environmentVariables>
+                                        <JBOSS_HOME>${project.build.directory}/wildfly-batch</JBOSS_HOME>
+                                    </environmentVariables>
+                                    <redirectTestOutputToFile>true</redirectTestOutputToFile>
```
yersan, line 1346: Would you mind using `<redirectTestOutputToFile>${testLogToFile}</redirectTestOutputToFile>` instead? I see it useful for local debugging and verifications.

yersan (approved):
Added WildFly proposals here: https://github.com/wildfly/wildfly-proposals/pull/324

Layer looks good to me, just added two minor details.

ee-galleon-pack/src/main/resources/feature_groups/batch-jberet-elytron.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature-group-spec name="batch-jberet-elytron" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature-group name="batch-jberet">
+        <feature spec="subsystem.batch-jberet">
+            <param name="security-domain" value="ApplicationDomain"/>
```
yersan, line 5: @chengfang Notice you will have to do something similar for EJB3 subsystem

testsuite/integration/basic/pom.xml
```diff
@@ -1284,6 +1323,39 @@
...
+                                    <includes>
+                                        <include>org/jboss/as/test/integration/batch/**/*TestCase.java</include>
+                                    </includes>
+                                    <excludes>
+                                        <!-- Test app has a manifest dep on org.jboss.remoting3 module -->
```
yersan, line 1351: From the recent discussions taking about this, we should be able to replace these org.jboss.remoting3 by a non-deprecated module. @bstansberry do you want me to driving this PR instead of you? I can create another one to supersedes this.

bstansberry:
@yersan Yes, please take this over. Thanks!

yersan:
This PR is superseded by https://github.com/wildfly/wildfly/pull/13434

bstansberry:
Closing in favor of #13434

PR #13021, opened 2020-02-19
[19.0.x] WFLY-13122 Include MicroProfile Fault Tolerance subsystem in the defa…

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-13122

Upstream
https://github.com/wildfly/wildfly/pull/13012

### DIFF
--- a/galleon-pack/src/main/resources/feature_groups/domain-full-ha.xml
+++ b/galleon-pack/src/main/resources/feature_groups/domain-full-ha.xml
@@ -7,6 +7,7 @@
 
     <feature spec="profile">
         <param name="profile" value="full-ha"/>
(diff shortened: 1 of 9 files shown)


rhusar:
We might not actually want this for 19.x, I dono.

bstansberry:
@rhusar Sorry for not commenting on this for so long.

I don't think we want it in 19. We provide standalone-microprofile.xml.  Fault Tolerance in an app in combo with use of other EE specs not in standalone-microprofile.xml might be problematic, so even though having it does no harm I think it's better to defer doing this to WF 20 and think about it a bit, get the docs of what works / what doesn't solidified etc.

rhusar:
Agreed – closing (if anything changes this can be simply reopened).

PR #13025, opened 2020-02-19
[WFLY-13136]: Can't create a Pooled Connection Factory using JGroups

ehsavoie:
* Looking for the new resources as the old ones are empty thus we can't
   access the DiscoveryGroup configuration.

Jira: https://issues.redhat.com/browse/WFLY-13136


bstansberry (approved):

bstansberry:
Note to mergers: I already merged the master variant of this, but kicked off a re-run of a test job before merging this one.

PR #13026, opened 2020-02-19
[19.0.x][WFLY-13128] Add a Galleon layer for batch-jberet

bstansberry:
https://issues.redhat.com/browse/WFLY-13128

### DIFF
--- a/ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
+++ b/ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="batch-jberet">
+    <dependencies>
+        <!-- CDI should be optional but to be conservative in the first cut at this, require it
+        <layer name="cdi" optional="true"/>
+        -->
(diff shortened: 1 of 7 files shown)


ee-galleon-pack/src/main/resources/layers/standalone/batch-jberet/layer-spec.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="batch-jberet">
+    <dependencies>
+        <!-- CDI should be optional but to be conservative in the first cut at this, require it
+        <layer name="cdi" optional="true"/>
+        -->
```
jamezp: I missed this on the other PR, but I don't think CDI should be optional. Batch requires some kind of DI. I guess though if the user provides one that does make it optional.

bstansberry:
I'm going to close this as there is some question about what the proper configuration should be with respect to security integration. In an integration configured for elytron we need the change here:

https://github.com/wildfly/wildfly/blob/master/feature-pack/src/main/resources/content/docs/examples/enable-elytron.cli#L9

Generally I want our layers oriented toward an elytron-based security configuration, so perhaps adding that should be part of the standard batch-jberet layer. But we need to think a bit about what we should do if the legacy-security layer is involved.

PR #13035, opened 2020-02-20
[19.0.x][WFLY-13127] Upgrade smallrye openapi to 1.2.0 to bring in https://gi…

bstansberry:
…thub.com/smallrye/smallrye-open-api/issues/248 fix

https://issues.redhat.com/browse/WFLY-13127

Upstream is #13014


bstansberry:
I'll point this on hold until Paul Ferraro determines if it's the right way to go. He can remove the label if it is.

PR #13036, opened 2020-02-20
[WFLY-13137] Facilitate use of a distinct maven groupId for artifacts…

bstansberry:
… associated with org.wildfly:wildfly-galleon-pack

https://issues.redhat.com/browse/WFLY-13137

This doesn't have much practical effect on WildFly, but it will make it easier in EAP branches to use different GAVs for the MicroProfile content.

### DIFF
--- a/microprofile/fault-tolerance-smallrye/cdi/pom.xml
+++ b/microprofile/fault-tolerance-smallrye/cdi/pom.xml
@@ -25,15 +25,17 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
         <!--
         Maintain separation between the artifact id and the version to help prevent
         merge conflicts between commits changing the GA and those changing the V.
         -->
         <version>20.0.0.Beta1-SNAPSHOT</version>
+        <relativePath>../../../pom.xml</relativePath>
     </parent>
 
     <artifactId>wildfly-microprofile-fault-tolerance-smallrye-cdi</artifactId>
+
     <name>WildFly: MicroProfile Fault Tolerance - CDI</name>
 
     <dependencies>
--- a/microprofile/fault-tolerance-smallrye/extension/pom.xml
+++ b/microprofile/fault-tolerance-smallrye/extension/pom.xml
@@ -25,15 +25,17 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
         <!--
         Maintain separation between the artifact id and the version to help prevent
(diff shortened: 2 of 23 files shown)


microprofile/fault-tolerance-smallrye/cdi/pom.xml
```diff
@@ -25,15 +25,17 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
```
jmesnil: Shouldn't you keep the same parent and set the relative path to `../pom.xml`?
`wildfly-microprofile-fault-tolerance-smallrye-cdi` and `wildfly-microprofile-fault-tolerance-smallrye` seems to be closely related to have the same GAV grouping.

microprofile/fault-tolerance-smallrye/extension/pom.xml
```diff
@@ -25,15 +25,17 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
```
jmesnil: same remark as above

microprofile/fault-tolerance-smallrye/pom.xml
```diff
@@ -48,4 +49,8 @@
         <module>cdi</module>
         <module>extension</module>
     </modules>
+
```
jmesnil: ok, that answers my questions above :)

pom.xml
```diff
@@ -9533,5 +9536,20 @@
...
             </pluginRepositories>
         </profile>
+
+        <!-- Configures testsuite to use the non-default Galleon pack to provision EE content -->
+        <profile>
+            <id>test.ee.galleon.pack.profile</id>
+            <activation>
+                <property>
+                    <name>testsuite.ee.galleon.pack.artifactId</name>
+                    <value>wildfly-ee-alleon-pack</value>
```
jmesnil: typo: should be `wildfly-ee-galleon-pack`

jmesnil (changes requested):
Looks good to me (in particular, TCK for MP Config/Health/Metrics/OpenTracing are bound to the `testsuite.ee.galleon.pack.groupId` groupId) but has the tests been properly run?

The typo for `testsuite.ee.galleon.pack.artifactId` should have made the test fail somehow, right?

pom.xml
bstansberry: Fixed

bstansberry:
@jmesnil https://ci.wildfly.org/viewLog.html?buildId=188279&tab=buildResultsDiv&buildTypeId=WF_GalleonLinuxAltFpJdk8 is a custom job against this PR using the nightly job that's meant to use this testsuite.ee.galleon.pack.xxx stuff.

It wouldn't have caught that problem though, because in WF all the version numbers are the same. To detect that problem I'd need to create a branch with different version numbers and run it against that.

PR #13015, opened 2020-02-19
WFLY-13099 Standard configuration files for MicroProfile use cases

bstansberry:
https://issues.redhat.com/browse/WFLY-13099

PR #13043, opened 2020-02-22
[19.0.x] WFLY-13099 Standard configuration files for MicroProfile use cases

bstansberry:
https://issues.redhat.com/browse/WFLY-13099

Upstream: #13015


bstansberry:
retest this please

bstansberry:
retest this please

bstansberry:
retest this please

OndrejKotek:
I have tried this with `jboss.dist` and got inconsistent results.
```
mvn clean install -DskipTests
cd testsuite
mvn clean test -Dts.standalone.microprofile -Djboss.dist=/home/okotek/test/server -fae
```

Web and Smoke modules seem to ignore the server referenced by `jboss.dist`. I have checked `WildFly Full <version> (WildFly Core <version>) starting` message in `target/wildfly*/standalone/log/server.log` files of the modules.

Also, in case the referenced server dosn't contain the new configs, I would expect all the test modules to fail. However, Web, Smoke, and **MicroProfile - OpenAPI TCK** pass:
```
[INFO] WildFly Test Suite: Aggregator ..................... SUCCESS [  1.540 s]
[INFO] WildFly Test Suite: Integration (parent) ........... SUCCESS [  0.998 s]
[INFO] WildFly Test Suite: Integration - Web .............. SUCCESS [ 33.075 s]
[INFO] WildFly Test Suite: Integration - Smoke ............ SUCCESS [ 23.196 s]
[INFO] WildFly Test Suite: Integration - Basic ............ FAILURE [ 23.915 s]
[INFO] WildFly Test Suite: Integration - MicroProfile ..... FAILURE [  3.656 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK . SUCCESS [  0.181 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK - Config FAILURE [  4.677 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK - Fault Tolerance FAILURE [  6.073 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK - Health FAILURE [  5.913 s]
[INFO] WildFly Test Suite: Integration - MicroProfile - JWT FAILURE [  7.486 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK - Metrics FAILURE [  4.516 s]
[INFO] WildFly Test Suite: Integration - MicroProfile - OpenAPI TCK SUCCESS [ 49.448 s]
[INFO] WildFly Test Suite: Integration - MicroProfile TCK - OpenTracing FAILURE [  4.862 s]
[INFO] WildFly Test Suite: Integration - Clustering ....... FAILURE [ 10.986 s]
```

OndrejKotek:
Also, it seems that `modules` directory of the referenced server is used by Basic test module for testing somehow, the `modified` timestamp is updated.

bstansberry:
@OndrejKotek That's for those findings. I've added 3 commits that should address the issues with smoke, web and microprofile-tck/openapi.

I see nothing that would alter whether the modules dir gets written to in this PRs change to ts/integration/basic. Does this not occur with master/19.0.x?  There may be a test writing temporary modules there. If so we should file an issue to get that cleaned up.

bstansberry:
retest this please

OndrejKotek:
@bstansberry Thank you for the fix. Smoke still passes despite the new configs missing in the server referenced by `jboss.dist`. According to `target/wildfly/standalone/log/server.log` the referenced server is used, however there is `standalone-microprofile.xml` config (used) and original `modules` (not used?) in `target/wildfly`.

Regarding `modules` directory of the referenced server used by Basic test module, it's an issue unrelated to this PR, I'll file it separately.

bstansberry:
@OndrejKotek  Thanks. I'll try again with 'smoke'. I thought I'd made it fail when pointing -Djboss.dist at output from the ee-dist module but I probably created a false negative somehow. Or somehow broke it again.

Thanks for filing something re writing to 'modules'. That kind of thing has come up before and probably got cleaned up for some specific test where it came up but we probably never cleaned up the whole testsuite.

bstansberry:
@OndrejKotek I added a commit dealing with 'smoke'. It would fail if -Djboss.dist pointed at a dist without MP, but the failure would be because modules referred to in the config file would not be found. The smoke ts is different from the others because by default it doesn't copy stuff from jboss.dist into target/wildfly. It uses Galleon to provision a server. For the ts.standalone.microprofile layer I turned that off and went back to the copying approach. So there will no longer be a target/wildfly/standalone/configuration/standalone-microprofile.xml if jboss.dist didn't have one. That means the server will immediately fail to boot due to a non-existent config file.

PR #13046, opened 2020-02-24
WFLY-13159 Datasource resolution fails for application layer alias when using Hibernate bytecode enhancement

scottmarlow:
https://issues.redhat.com/browse/WFLY-13159
https://issues.redhat.com/browse/JBEAP-18803

This is a bug fix that addresses applications that use bytecode enhancing and a DataSource defined by the application (e.g. resource reference).

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/JPA_Reference_Guide.adoc
@@ -846,6 +846,8 @@ unit bootstrap, which improves JPA integration with CDI. Setting the
 wildfly.jpa.twophasebootstrap hint to false, disables the two phase
 bootstrap (for the persistence unit that contains the hint).
 
+|wildfly.jpa.applicationdatasource |set to true when using an application defined DataSource or resource reference to a global DataSource.
+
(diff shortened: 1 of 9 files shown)


bstansberry:
@scottmarlow Perhaps a test case?

scottmarlow:
@bstansberry org.jboss.as.test.integration.jpa.datasourcedefinition.DataSourceDefinitionJPATestCase reproduces the (application datasource via @DataSourceDefinition) bug but we could add additional tests as well that use a resource reference.

scottmarlow:
added test case that uses resource reference.

scottmarlow:
retest this please

scottmarlow:
The resourceref.DataSourceResourceReferenceTestCase failure looks relevant, I'll see if I can recreate that locally.

scottmarlow:
I fixed the added test case that was added for this, so we now have coverage of the two forms of application datasource with JPA bytecode enabled (DataSourceDefinition + resource ref).

scottmarlow:
The Linux JDK11 failure is in XSiteSimpleTestCase.test, that doesn't seem related to this pr.

bstansberry (approved):

PR #13047, opened 2020-02-25
[WFLY-13161] CLIENT-CERT login does not work in intermediate elytron setup

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-13161

Two comments:

1. The support for X509 evidences was added as possibly supported (if you think supported is better just let me know).
2. I doubted a lot between setting the credential to the final user certificate or the full chain. Finally I decided to use the first option (`getFirstCertificate` method) because I have checked and plain undertow/picketbox configuration just sends the client certificate (only one), so I'm doing the same. Nevertheless the `BaseCertLoginModule` accepts both (the test works for both options). The same, comment me if you prefer the full chain.

@darranl I'm going to ask for your review cos you already know about this issue. Change that if needed.


rmartinc:
@darranl Remember to take a look to this PR when you have time. ;-)
I wanted your comments to go for downstream.

darranl (approved):

PR #13052, opened 2020-02-26
WFLY-13152 WFLYEJB0094: EJB 3.1 FR 5.4.2 MessageDrivenBean does not i…

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-13152

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MessageDrivenComponentDescriptionFactory.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MessageDrivenComponentDescriptionFactory.java
@@ -29,7 +29,6 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.Properties;
 import java.util.Set;
-
 import javax.ejb.MessageDriven;
 import javax.jms.MessageListener;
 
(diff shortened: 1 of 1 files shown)


yersan (commented):
I think the CompositeIndex of the EAR modules could include the indexes of the modules they have access to. So anyone can beneficiate directly using the composite index. 

With this patch, this issue is still likely to fail if, for example, the ear uses two EJB modules and one module contains the parent class implementing the MessageListener interface and the other uses it.

Maybe a correct place to fix this could be the `CompositeIndexProcessor` instead, adding thereto each EAR module the indexes of the other modules that are shared with it. I'm not sure yet if the deployment exclusions or `ear-subdeployments-isolated` enabled should influence this behavior. It sounds reasonable to take it into account.

@chengfang What's your opinion on the above points?

chengfang:
Not sure if [CompositeIndexProcessor](https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/deployment/annotation/CompositeIndexProcessor.java#L176) already handles inter-module dependencies when building the index.  Looks like it should with its `buildSubdeploymentDependencyMap` method.  Various sharing and unsharing config in [jboss-deployment-structure](https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/jboss-deployment-structure-1_3.xsd#L47) will certainly affect this issue.  The standard way to share classes between ejb jars is to include common classes in ear lib jars, but if users configure `ear-subdeployments-isolated` to false and have ejb jars depend on other ejb jars, it should also be supported.  One issue with proactively adding dependencies to a module's `CompositeIndex` is we may be adding too much entries for no gains in most cases, as having ejb super classes in ear lib jar is edge case, let alone in other peer ejb jars.

yersan:
> Not sure if CompositeIndexProcessor already handles inter-module dependencies when building the index. Looks like it should with its buildSubdeploymentDependencyMap method. Various sharing and unsharing config in jboss-deployment-structure will certainly affect this issue. 

My understanding is it only handles modules that are explicitly added via class-path or deployment-structure. Hence my suggestion to additionally compute the indexes of the sub-deployment modules that belong to the same EAR (jar in libs to ejb-jars and wars, ejb-jars between them and to wars).

Anyway, by your inputs being this an edge case, unnecessary and overcomplicated. Your change already fixes the problem having a superclass defined into the EAR parent module.

bstansberry (approved):

PR #13065, opened 2020-03-02
[WFLY-13137] Use ee.maven.groupId in dependency management where ee.m…

bstansberry:
…aven.version is used

Follow-up work on https://issues.redhat.com/browse/WFLY-13137.  Use ee.maven.groupId in dependency management, and then the MP-related artifacts should use it in their pom's dependency sectioins.

### DIFF
--- a/microprofile/fault-tolerance-smallrye/executor/pom.xml
+++ b/microprofile/fault-tolerance-smallrye/executor/pom.xml
@@ -25,12 +25,13 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
         <!--
         Maintain separation between the artifact id and the version to help prevent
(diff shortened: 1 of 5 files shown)


microprofile/fault-tolerance-smallrye/executor/pom.xml
```diff
@@ -25,12 +25,13 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-fault-tolerance-smallrye</artifactId>
+        <artifactId>wildfly-parent</artifactId>
```
rhusar: This is wrong, why this change? This breaks `mvn clean install -rf :wildfly-microprofile-fault-tolerance-smallrye`.

bstansberry: We might be able to get rid of it.  It's a long story.

rhusar: OK I trust you! Just wanted to point it out in case it was unintentional.

PR #13067, opened 2020-03-02
[19.0.x][WFLY-13172][WFLY-13173]  Secured management and health

jfdenise:
Fix for https://issues.redhat.com/browse/WFLY-13172 and https://issues.redhat.com/browse/WFLY-13173
@bstansberry @pferraro @jmesnil I removed the dependency to management that was present in microprofile-config. It seemed not right. The only layers that depend on secure-management are metrics and health.


rhusar:
You might want to prefix the PR name with "[19.0.x]..."

bstansberry:
@jfdenise Seems ok. Can you squash these? The 2nd commit doesn't do what the msg says, it seems to just be a correction of a mistake in the 1st. So might as well just combine the two commits.

jfdenise:
@bstansberry , squashed.

jmesnil (commented):
@jfdenise it looks good to me, thanks

PR #13074, opened 2020-03-04
Cleanup documentation for "Using WildFly as a Load Balancer" section;…

rhusar:
… cleanup "load-balancer"

No Jira needed.

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow_using_as_a_Load_Balancer.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow_using_as_a_Load_Balancer.adoc
@@ -1,6 +1,6 @@
 [[Undertow_using_as_a_Load_Balancer]]
 = Using WildFly as a Load Balancer
-:wildflyVersion: 10.1
+:wildflyVersion: 18.0
 
 WildFly 10 added support for using the Undertow subsystem as a load
 balancer. WildFly supports two different approaches, you can either
@@ -11,16 +11,16 @@ to dynamically update the hosts.
 [[general-overview]]
 == General Overview
 
-Wildly uses Undertow's proxy capabilities to act as a load balancer.
+WildFly uses Undertow's proxy capabilities to act as a load balancer.
 Undertow will connect to the back end servers using its built in client,
 and proxies requests.
 
 The following protocols are supported:
 
-* http
-* ajp
-* http2
-* h2c (clear text HTTP2) +
+* `http`
+* `ajp`
+* `http2`
+* `h2c` (clear text HTTP2) +
 Of these protocols h2c should give the best performance, if the back end
 servers support it.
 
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Undertow_using_as_a_Load_Balancer.adoc
```diff
@@ -1,6 +1,6 @@
 [[Undertow_using_as_a_Load_Balancer]]
 = Using WildFly as a Load Balancer
-:wildflyVersion: 10.1
+:wildflyVersion: 18.0
```
bstansberry: @jamezp 

I'm not sure what this item was meant for in this file. Probably it can be removed, as AIUI the version string is meant to be controlled from the pom. But I note that it's quasi-camel-case here, and elsewhere in the adoc files, while the pom uses lower case, as does the README note on using it:

https://github.com/wildfly/wildfly/blob/master/docs/pom.xml#L69

rhusar: It's only used to generate a link at the bottom of the page at this page. Let me test if this can reference something else.

jamezp: So this is actually an interesting one and I didn't think to look for other references to wildscribe, however I will. Since we're moving Wildscribe into the documentation we'll need to change the link. However it should be as simple as linking to `wildscribe//subsystem/undertow/index.html` so the version should not be required.

jamezp: Doesn't look like too many other references, but I'll submit a PR to update the others.

bstansberry: FWIW, if I build the docs in master,  for that link I get https://wildscribe.github.io/WildFly/20/subsystem/undertow/index.html and the published page at https://docs.wildfly.org/18/Admin_Guide.html has https://wildscribe.github.io/WildFly/18/subsystem/undertow/index.html. So whatever is in this line of the adoc seems to not take effect, and the camel case vs lower case thing I noted also doesn't matter.

Unfortunately though the https://wildscribe.github.io/WildFly/18/subsystem/undertow/index.html link is wrong as it would need to be 18.0 not 18. Hopefully James' wildscribe//subsystem/undertow/index.html suggestion fixes that.

rhusar: @bstansberry The difference is that if you view the file on github - it will resolve to whatever version I am changing here as opposed to when you build, you get the link with the version you are building.

Lets try the James' way.

bstansberry: Thanks Rado; I keep forgetting about the view-it-on-github use case.

PR #13075, opened 2020-03-05
WFLY-13193 Move resteasy-client-microprofile to dedicated module

asoldano:
https://issues.redhat.com/browse/WFLY-13193

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
@@ -33,6 +33,7 @@ import org.jboss.modules.ModuleIdentifier;
 public class JaxrsSubsystemDefinition extends SimpleResourceDefinition {
     public static final ModuleIdentifier RESTEASY_ATOM = ModuleIdentifier.create("org.jboss.resteasy.resteasy-atom-provider");
     public static final ModuleIdentifier RESTEASY_CDI = ModuleIdentifier.create("org.jboss.resteasy.resteasy-cdi");
+    public static final ModuleIdentifier RESTEASY_CLIENT_MICROPROFILE = ModuleIdentifier.create("org.jboss.resteasy.resteasy-client-microprofile");
     public static final ModuleIdentifier RESTEASY_CRYPTO = ModuleIdentifier.create("org.jboss.resteasy.resteasy-crypto");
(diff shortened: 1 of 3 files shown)


bstansberry:
@asoldano I think you will need to add something to this or Galleon won't provision this module when it creates a slimmed server:

https://github.com/wildfly/wildfly/blob/master/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java#L72

If needed @jfdenise can provide more details during morning in Europe tomorrow.

bstansberry:
Once you have something you're good with please open a 19.0.x PR so we can get Beta 3 done not too late tomorrow.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
```diff
@@ -78,6 +79,7 @@ public void registerAdditionalRuntimePackages(ManagementResourceRegistration res
                     RuntimePackageDependency.required(JSON_API.getName()),
                     RuntimePackageDependency.optional(RESTEASY_ATOM.getName()),
                     RuntimePackageDependency.optional(RESTEASY_JAXRS.getName()),
+                    RuntimePackageDependency.optional(RESTEASY_CLIENT_MICROPROFILE.getName()),
```
jfdenise: @asoldano @bstansberry , in case we want to create a microprofile-rest-client galleon layer, this dependency should be passive. It would mean that when provisioning the jaxrs layer, microprofile-rest-client would be no more provisioned (no org.jboss.resteasy.resteasy-client-microprofile nor org.eclipse.microprofile.restclient modules).
The only way to provision microprofile-rest-client support would be by provisioning the "microprofile-rest-client" layer that would depend on jaxrs/cdi/microprofile-config. This would have an impact on jaxrs existing layer and layers that depend on it.
This is really an UX improvement (ability to pick and choose mp spec), the gain in term of size is small (100k).

bstansberry: @jfdenise Is there any reason this kind of adjustment needs to happen now, e.g. some incompatibility later?

AIUI the goal of this PR is not adding a new layer, so changes to support a new layer can come with that work.

jfdenise: @bstansberry , this can happen later.

bstansberry (approved):

PR #13076, opened 2020-03-05
Update the Wildscribe links to use the version of Wildscribe associat…

jamezp:
…ed with the current docs release.

No JIRA required, however I can create one if desired.

@rhusar this is likely going to conflict with #13074, but which ever is merged first we can fix or we can combine them in a single PR. Either works for me.


bstansberry:
@jamezp I went ahead and merged #13074

jamezp:
Perfect. I've rebased and pushed up the rebase.

bstansberry (approved):
@jamezp As long as you checked locally that these links work when docs are actually built, this seems fine.

PR #13077, opened 2020-03-05
Update the Wildscribe links to use the version of Wildscribe associat…

jamezp:
…ed with the current docs release.

No JIRA required, however I can create one if desired.


bstansberry (approved):
@jamezp As long as you checked locally that these links work when docs are actually built, this seems fine.

PR #13085, opened 2020-03-05
WFLY-13205 Remove clustering EC2 example server profiles

rhusar:
https://issues.redhat.com/browse/WFLY-13205

next up:  WFLY-13206 Replace Azure example server profile with a script


rhusar:
@pferraro for your review.

pferraro (approved):

bstansberry:
These are used downstream so we need a plan for what to do about that.

darranl:
Do we know what we are going to do to review if we can deal with this change?

pferraro:
Certainly, the existing server examples outdated.  S3_PING was deprecated a while ago - replaced by NATIVE_S3_PING from jgroups-extras.  AIUI, these example profiles are just documentation - as such, would be more maintainable as CLI scripts that modify an existing HA profile configuration.

darranl:
@rhusar / @pferraro This is now the least recently updated PR in the queue, do we have a path forward or could we close for now and resubmit later if now is not a suitable time for this change.

darranl:
This also has a merge conflict now.

rhusar:
> @rhusar / @pferraro This is now the least recently updated PR in the queue, do we have a path forward or could we close for now and resubmit later if now is not a suitable time for this change.

Yes, I believe so, now with galleon I think it's needed more than ever and the preconfigured profiles should be gone. Just didn't make it up on my priority list, but hopefully soon.

darranl:
@rhusar Time for a random period ping as I look at old PRs again ;-) I assume this has not made it up the priority list yet?

PR #13086, opened 2020-03-06
Fix for WFLY-13210, open-tracing layer must depend on microprofile-config

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-13210
Without microprofile-config, we can't deploy a jaxrs endpoint.

### DIFF
--- a/ee-galleon-pack/src/main/resources/layers/standalone/open-tracing/layer-spec.xml
+++ b/ee-galleon-pack/src/main/resources/layers/standalone/open-tracing/layer-spec.xml
@@ -2,6 +2,7 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="open-tracing">
     <dependencies>
         <layer name="cdi"/>
+        <layer name="microprofile-config"/>
     </dependencies>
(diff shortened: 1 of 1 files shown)


bstansberry:
@jfdenise or @ehsavoie or @jmesnil  Please explain further why this is necessary.

jfdenise:
@bstansberry @ehsavoie @jmesnil I have pasted the exception I observed in the JIRA : https://issues.redhat.com/browse/WFLY-13210

ehsavoie:
There are 2 MP config values used for MP opentracing

bstansberry (approved):
@jfdenise @ehsavoie @jmesnil 

I'm ok with this but it seems like more is needed.  The server boot should not progress into Stage.RUNTIME if a capability required by a subsystem is not present.  The problem shouldn't go undetected until an app is deployed.

I'm not saying fixing that should be part of this PR or this JIRA but it sounds like we need a separate JIRA.

ehsavoie:
The issue is in ServiceLoader so it is 'just' some configuration issue in module.xml

PR #13089, opened 2020-03-06
[WFLY-13185]:Add microprofile restclient dependency to jaxrs deployment to fix NPE issue

jimma:
https://issues.redhat.com/browse/WFLY-13185

PR #13092, opened 2020-03-09
[WFLY-13185]:Add microprofile restclient dependency to jaxrs deployment

jimma:
https://issues.redhat.com/browse/WFLY-13185

Upstream: #13089


bstansberry (changes requested):
@jimma I believe you also need to add a line equivalent to this:

https://github.com/wildfly/wildfly/pull/13075/files#diff-aaad68233e8e9273c54f93f079cc27cbR82

@jfdenise Please review

jfdenise:
@jimma , as @bstansberry pointed, you need to advertise to galleon any module that are not hard dependencies of the extension module.xml (directly or transitively) that you are injecting to deployments. This seems the case there. This is done calling RuntimePackageDependency.optional(MP_REST_CLIENT.getName());

jimma:
@bstansberry @jfdenise  Fixed. Thanks.

jfdenise (approved):

bstansberry (approved):

PR #13112, opened 2020-03-13
WFLY-13238 Mutable session attribute fix for ATTRIBUTE granularity and non-transactional caches

pferraro:
https://issues.redhat.com/browse/WFLY-13238


pferraro:
Replacing with a fix incorporating WFLY-13233.

PR #13114, opened 2020-03-13
WFLY-13240 Clustering TS: Default to non-multicast discovery protocol…

rhusar:
… for 'bridge' channel used in XSite tests

Resolves
https://issues.redhat.com/browse/WFLY-13240


rhusar:
@pferraro For your review. 

If only our GitHub organization was setup correctly and I could use its features to request a review... but no.

pferraro (approved):

PR #13121, opened 2020-03-17
WFLY-13147 Avoid processing deployments multiple times in Weld subsys…

manovotn:
…tem.

JIRA - https://issues.redhat.com/browse/WFLY-13147

_This basically a copy of what the issue suggsested as code changes_. All in all those changes make sense in avoiding loading dep modules that we know were/will be processed as separate units. Simple tests pass for me (as in tests in this repo executed with `-DallTests`).

However, I don't have a big enough sample to actually measure the impact on re-deploy times on a kind of application that it was reported against - @bstansberry does WFLY have some perf tests for deployment where this could be measured?
Or really, any more "obscure" deployment tests that we could/should run against.

CCing @mkouba just in case he has some thoughts about these changes.

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
@@ -125,7 +125,11 @@ public class ExternalBeanArchiveProcessor implements DeploymentUnitProcessor {
 
         final HashSet<URL> existing = new HashSet<URL>();
 
+        // build a set of all deployment unit names, used later on to skip processing of some dependencies
+        final Set<String> depUnitNames = new HashSet<>();
+        final String prefix = "deployment.";
         for (DeploymentUnit deployment : deploymentUnits) {
+            depUnitNames.add(prefix + deployment.getName());
             try {
                 final ExplicitBeanArchiveMetadataContainer weldDeploymentMetadata = deployment.getAttachment(ExplicitBeanArchiveMetadataContainer.ATTACHMENT_KEY);
                 if (weldDeploymentMetadata != null) {
@@ -162,6 +166,14 @@ public class ExternalBeanArchiveProcessor implements DeploymentUnitProcessor {
                 return;
             }
             for (DependencySpec dep : module.getDependencies()) {
+                if (!(dep instanceof ModuleDependencySpec)) {
+                    continue;
+                }
+                if (depUnitNames.contains(((ModuleDependencySpec) dep).getName())) {
+                    // dep.getName() returns something like deployment.123abc.ear
+                    // we want to skip processing this as it will be (or already was) processed as another dep. unit
+                    continue;
+                }
                 final Module dependency = loadModuleDependency(dep);
                 if (dependency == null) {
                     continue;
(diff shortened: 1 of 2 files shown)


bstansberry:
@manovotn WildFly as a project doesn't have separate boot perf tests. Our testsuite plus the EE TCKs is basically what we have. @msve do you guys have anything like that?

@ropalka I know last you you did some work on the number of services EAP deployments were creating. If you have some sort of deployment speed test left over from that that might be useful, please let @manovotn  know.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
```diff
@@ -162,6 +162,14 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                 return;
             }
             for (DependencySpec dep : module.getDependencies()) {
+                if (!(dep instanceof ModuleDependencySpec)) {
+                    continue;
+                }
+                if (deploymentUnits.stream().anyMatch(d -> (((ModuleDependencySpec) dep).getName()).endsWith(d.getName()))) {
+                    // d.getName() returns deployment.123abc.ear, whereas dep.getName() returns 123abc.ear
```
bstansberry: This comment is confusing because for using the example values the code would amount to "123abc.ear".endsWith("deployment.123abc.ear") which would never be true.

manovotn: Oh yea, it's the other way around of course, my bad. Either way, I wasn't planning to keep it here, it was just a temporary explanation of why the matching works based on my debugging...

ropalka:
Hi @bstansberry @manovotn - I didn't measure the boot time when I was dealing with number of services. What I have been doing was I was just counting number of created services during the deployment phase.

ucswntr:
Hi @bstansberry @manovotn , are there any news regarding the patch? Our colleagues are already  looking forward to reducing the startup times during their development/testing cycles.

KR Christoph

manovotn:
I know about this issue but I keep being pulled to other issues and PRs.

If @bstansberry doesn't have anything against it, we can let it in like this. The changes don't break existing tests and we know that they lower deployment times for @ucswntr scenario. 

I've fixed the misleading comment and rebased the PR against master.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
```diff
@@ -162,6 +162,14 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                 return;
             }
             for (DependencySpec dep : module.getDependencies()) {
+                if (!(dep instanceof ModuleDependencySpec)) {
+                    continue;
+                }
+                if (deploymentUnits.stream().anyMatch(d -> (((ModuleDependencySpec) dep).getName()).endsWith(d.getName()))) {
```
bstansberry, line 168: My only "maybe" suggestion is earlier in this method when looping through the DeploymentUnits, built up a Set<String> containing "deployment." + d.getName().  Then here just do a contains check against that set.

bstansberry:
@manovotn I'm fine with adding this. I made a small comment above but if you don't like that, please just @ ping me and I'll approve this and it can go in.

manovotn:
@bstansberry updated, see if you like it now :-)

PR #13122, opened 2020-03-17
WFLY-12018 Fix NotSerializableException when a distributable web application uses web sockets.

pferraro:
https://issues.redhat.com/browse/WFLY-12018


bstansberry:
LGTM but I'll give @fl4via a chance to comment before merging.

AIUI what this is doing is storing the value of any "io.undertow.websocket.current-connections" attribute in a kind of side storage (the 'local context') that does not get replicated (or, I assume, persisted in the case of non-HA distributable webapps).

I assume in the case of HA failover or restoration of a locally persisted session it is expected that the "io.undertow.websocket.current-connections" attribute will now have a null value.

pferraro:
@fl4via Any comments?  Or can we merge this?

PR #13129, opened 2020-03-20
[WFLY-12851] Upgrade PicketLink bindings from 2.5.5.SP12-redhat-00009-jbossorg-1 to 2.5.5.SP12-redhat-00012

ivassile:
https://issues.redhat.com/browse/WFLY-12851


pom.xml
```diff
@@ -428,7 +428,7 @@
         <version.org.ow2.asm>7.1</version.org.ow2.asm>
         <version.org.picketbox.picketbox-commons>1.0.0.final</version.org.picketbox.picketbox-commons>
         <version.org.picketlink>2.5.5.SP12-redhat-00006</version.org.picketlink>
-        <version.org.picketlink.bindings>2.5.5.SP12-redhat-00009-jbossorg-1</version.org.picketlink.bindings>
+        <version.org.picketlink.bindings>2.5.5.SP12-redhat-00012</version.org.picketlink.bindings>
```
bstansberry: Is there a reason any more for this to be distinct from version.org.picketlink?

ivassile: We've decoupled PicketLink and PicketLink Binding with https://github.com/wildfly/wildfly/pull/12684 in WF and now the components can be upgraded independently. We also apply this change in EAP.

bstansberry: OK, I see there is no 2.5.5.SP12-redhat-00012 of the other one.

bstansberry (approved):

PR #13131, opened 2020-03-22
[WFLY-12402] Clean up test deployment use of management classes and t…

bstansberry:
…hereby increase number of tests that can run in the ts.layers profile with Galleon slimmed servers

1) Eliminate no-longer needed dependencies on the org.jboss.as.cli module, which typically is not provisioned in ts.layers provisioning.
2) General cleanup of tests that used this or ones near those (primarily jca/ra/ds related tests) to better clarify between code that needs 'management' functionality and code that does not.
a) Deployments usually do not, particularly those in @RunAsClient tests. So remove the management code from the deployments as much as possible.
b) However, if the test class goes in the deployment and a @ServerSetup is on that class, the ServerSetupTask impl class and its superclasses must be present, even though they are not executed on the server side. Server side code processing the deployment will try and examine them.  Other classes used in the setup tasks do not need to be in the deployment though, as the setup task executes on the test driver side.
c) A number of tests were subclassing AbstractMgmtTestBase directly or transitively but were not using any of its functionality, instead relying on a ServerSetupTask for such things. In such cases the test class was changed to no longer subclass unnecessarily.
3) Following from this a large number of tests that previously could not run in the ts.layers profile now can, so they have been enabled.

https://issues.redhat.com/browse/WFLY-12402

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -440,6 +440,7 @@
                                 <include>org/jboss/as/test/integration/jaxb/**/*Test.java</include>
                                 <include>org/jboss/as/test/integration/jaxrs/**/*TestCase.java</include>
                                 <include>org/jboss/as/test/integration/jaxrs/**/*Test.java</include>
+                                <include>org/jboss/as/test/integration/jca/**/*TestCase.java</include>
                                 <include>org/jboss/as/test/integration/json/**/*TestCase.java</include>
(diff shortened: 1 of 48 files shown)


jamezp (approved):

bstansberry:
I put the 'hold' label on this because it's currently based on the tmp branch I used for batch PR test #13292. Once I merge those I'll rebase on master. I just wanted CI to run this in combo with all those others.

PR #13133, opened 2020-03-23
[WFLY-13273] Add tests for WFCORE-4860. The tests are required here a…

jamezp:
…s multiple deployments are required for testing.

https://issues.redhat.com/browse/WFLY-13273

Tests for https://github.com/wildfly/wildfly-core/pull/4112. This can be merged before as the tests should currently pass to prove the behavior is the same between the current implementation and the changed implementation.


jamezp:
These failures seem legitimate. I'll try to reproduce locally.

jamezp:
These failures are due to no EJB subsystem in the configs used. These will either need to be ignored or modified to either no use EJB's or know the subsystem is not present and ignore those missing messages.

bstansberry (approved):
The exclude was the right way to handle ts.layers.

Tests could check for system props ts.layers or ts.standalone.microprofile and if either is present behave differently, but I don't know if that's worth it. The logging subsystem is the same with all of these setups so there's not a lot of value in repeating the tests.

bstansberry:
@jamezp If this can be merged the core-upgrade-needed label can be removed.

PR #13140, opened 2020-03-25
[WFLY-13283] suspending WFTC transaction on response of REST-AT bridge

ochaloup:
https://issues.redhat.com/browse/WFLY-13283

### DIFF
--- a/rts/src/main/java/org/wildfly/extension/rts/jaxrs/ImportWildflyClientGlobalTransactionFilter.java
+++ b/rts/src/main/java/org/wildfly/extension/rts/jaxrs/ImportWildflyClientGlobalTransactionFilter.java
@@ -22,23 +22,44 @@
 package org.wildfly.extension.rts.jaxrs;
 
 import java.io.IOException;
+import javax.annotation.Priority;
+import javax.transaction.Status;
 import javax.transaction.SystemException;
+import javax.ws.rs.Priorities;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.ext.Provider;
 import org.wildfly.extension.rts.logging.RTSLogger;
+import org.wildfly.transaction.client.ContextTransactionManager;
 import org.wildfly.transaction.client.LocalTransactionContext;
 
 
 /**
- * Filter which is expected to be called after {@link InboundBridgeFilter} is processed.<br>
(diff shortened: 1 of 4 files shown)


ochaloup:
@kabir do you think this could be merged? I don't know who should I ask to review this in particular. In past I was asking these kind of fixes with David.

bstansberry:
@asoldano Please review.

@ochaloup As this relates to JAX-RS integration I've asked the REST component lead.

asoldano (commented):
Hi @ochaloup ,
wrt the filter priority, is there any expectation regarding other user provided filters being executed before or after yours?
Besides that, the javadoc changes in WildflyTransactionClientTxBridgeIntegrationHandler look unrelated to the other modifications on this PR (I see they're about WS), am I wrong?
Cheers

ochaloup:
@asoldano 
The point of the priority of this change (the filter in WildFly) is that it has to be processed after the Narayana one.
There is not specific expectation regarding of the user provided filters. But if I understand correctly you point me to the fact that `Priorities.USER` is probably not a good priority when this seems to be rather a "container internal filter". Do I understand correctly?
The reason was to go with `USER + 1` as the default JAX-RS filter priority is USER and unfortunately the original Narayana bridge filter (https://github.com/jbosstm/narayana/blob/5.10.4.Final/rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridgeFilter.java#L50) specifies no priority, aka. the `USER` level. 
If we talk about the user filters - as this is about the transactions - probably user could expect the transaction is already available for his processing in any filter he creates.
Maybe the filter priority in Narayana should be changed first, you think?

Regarding the javadoc, you are right. The comments pays for the similar functionality in XTS. I changed the PR in this regard.

ochaloup:
@asoldano what do you think now about the PR. Should the priority be changed?

ochaloup:
@asoldano please, what is your current standpoint on this PR? And thanks for the review.

ochaloup:
@asoldano @bstansberry is there something specific that this PR should be updated with to be merged?

bstansberry:
retest this please

I'm deferring to @asoldano on this one as I have no expertise.

ochaloup:
I'm closing this PR as the Narayana inbound bridge filter that does all the handling uses wrong `USER` priority (https://github.com/jbosstm/narayana/blob/5.10.4.Final/rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridgeFilter.java#L50).
First, there is needed a change in the Narayana codebase (https://issues.redhat.com/browse/JBTM-3321) and then there could be created an integration issue that uses priority bigger than the `InboundBridgeFilter`.

PR #13143, opened 2020-03-25
WFLY-13260 Remove Servlet specification dependency from wildfly-clustering-web-spi

pferraro:
https://issues.redhat.com/browse/WFLY-13260

This change removes any dependency from the wildfly-clustering-web-spi and its implementations (i.e. wildfly-clustering-web-hotrod and wildfly-clustering-web-infinispan) on a specific version of the Servlet specification.
This is needed for Tomcat 10 support in wildfly-clustering-tomcat.


bstansberry (approved):
@pferraro Seems ok, but not studied in huge depth. If you have someone in mind to look at this please ping them; otherwise please merge when you are ready.

PR #13149, opened 2020-03-26
[WFLY-13293] When deploying "ROOT.war" in EAP7.x, the context root value output through jboss-cli is not valid

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-13293

I have moved the final calculation of the app context from the info service to the deployment processor (this way both use the same context and teh issue is resolved). Let me know if you see something wrong.

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
@@ -185,12 +185,12 @@ public class UndertowDeploymentProcessor implements DeploymentUnitProcessor, Fun
             deploymentName = deploymentUnit.getParent().getName() + "." + deploymentUnit.getName();
         }
 
-        final Map.Entry<String,String> severHost = defaultModuleMappingProvider.getMapping(deploymentName);
+        final Map.Entry<String,String> serverHost = defaultModuleMappingProvider.getMapping(deploymentName);
         String defaultHostForDeployment;
         String defaultServerForDeployment;
-        if (severHost != null) {
-            defaultServerForDeployment = severHost.getKey();
-            defaultHostForDeployment = severHost.getValue();
+        if (serverHost != null) {
+            defaultServerForDeployment = serverHost.getKey();
+            defaultHostForDeployment = serverHost.getValue();
         } else {
             defaultServerForDeployment = this.defaultServer;
             defaultHostForDeployment = this.defaultHost;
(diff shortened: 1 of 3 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
```diff
@@ -198,8 +199,15 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
 
         String serverInstanceName = warMetaData.getMergedJBossWebMetaData().getServerInstanceName() == null ? defaultServerForDeployment : warMetaData.getMergedJBossWebMetaData().getServerInstanceName();
         String hostName = hostNameOfDeployment(warMetaData, defaultHostForDeployment);
+        CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+        ServiceName hostServiceName = support.getCapabilityServiceName(Capabilities.CAPABILITY_HOST, serverInstanceName, hostName);
+        ServiceController<?> hostService = phaseContext.getServiceRegistry().getService(hostServiceName);
```
bstansberry: I don't have time to properly review this now, but this doesn't feel right. DUPs looking up services isn't great as it's only safe if the DeploymentUnitPhaseService running the DUP has a dependency on the service being looked up. Which is fragile.

rmartinc: Thanks for looking this Brian! I tried first to fill the context root inside the service instead of the processor but the problem is that the context path is also used to name/locate the service (as an alias) [here](https://github.com/wildfly/wildfly/blob/master/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java#L316). So I didn't see any other alternative that trying to get the host and know the defaultWebModule inside the processor to create the alias with the right and final context. I'll ping you later.

bstansberry: @rmartinc Have a look at DefaultDeploymentMappingProvider. It was written for this sort of problem. I think that can be adapted.  Add some sort of reverse lookup function to it.

It's a public class but it's meant only for use internal to the subsystem so it can changed as needed.

rmartinc: Thanks a lot @bstansberry! I'm very sorry but, although the solution was just before my eyes, I didn't see it. I should have noticed that class. :-(

No need to change the class at all, with the current map it's possible to see if the deployment is the default one for that server/host.

I have performed the changes and rebased. Let's wait the tests. And let me know if you need something else.

rmartinc: Ah! I didn't remove the public modifier because the class is used in two packages (`org.wildfly.extension.undertow.deployment` and `org.wildfly.extension.undertow`). I forgot to mention that.

bstansberry: Re the public modifier, that's fine; I wasn't looking to have that changed. I just meant not to worry that the class was public so you had to worry about API compatibility. You could have changed it as needed. But not needing to change it is good too!

bstansberry (approved):

PR #13153, opened 2020-03-27
[WFLY-13297] Provide the deployment's PropertyReplacer to WeldInjecti…

bstansberry:
…onServices for its use in handling @Resource injection

https://issues.redhat.com/browse/WFLY-13297

### DIFF
--- a/weld/common/src/main/java/org/jboss/as/weld/util/ResourceInjectionUtilities.java
+++ b/weld/common/src/main/java/org/jboss/as/weld/util/ResourceInjectionUtilities.java
@@ -31,6 +31,7 @@ import javax.enterprise.inject.spi.Annotated;
 import javax.enterprise.inject.spi.InjectionPoint;
 
 import org.jboss.as.weld.logging.WeldLogger;
+import org.jboss.metadata.property.PropertyReplacer;
 import org.jboss.weld.injection.ParameterInjectionPoint;
 
 public class ResourceInjectionUtilities {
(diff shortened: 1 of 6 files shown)


bstansberry:
@manovotn WDYT?

weld/common/src/main/java/org/jboss/as/weld/util/ResourceInjectionUtilities.java
```diff
@@ -50,14 +51,21 @@ public static String getResourceName(String jndiName, String mappedName) {
         }
     }
 
+    /** @deprecated use {@link #getResourceName(InjectionPoint, PropertyReplacer)}  */
+    @Deprecated
```
manovotn: Couldn't we straight up remove this method? It's not any public API and within WFLY I only see a single usage that you already replaced.

manovotn (approved):
Looks good!

weld/common/src/main/java/org/jboss/as/weld/util/ResourceInjectionUtilities.java
bstansberry: If you're ok with removing it I'll do it in a separate PR.

I'm cautious about that kind of thing because some layered product might be doing odd things, but really I think using this method would have been unsupported hacking. But I expect this PR will be ported to EAP for bug fix branches so to be super-cautious I'll clean it up in a separate PR.

manovotn: I see what you mean, but I think we could remove it.

bstansberry: @manovotn  https://github.com/wildfly/wildfly/pull/13160

PR #13157, opened 2020-03-31
[WFLY-13308] remove ejb client bom from dependency management

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-13308
19.0.x PR: https://github.com/wildfly/wildfly/pull/13158

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -890,13 +890,6 @@
 
             <dependency>
                 <groupId>${ee.maven.groupId}</groupId>
-                <artifactId>wildfly-ejb-client-bom</artifactId>
-                <version>${ee.maven.version}</version>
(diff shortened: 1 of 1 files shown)


bstansberry:
@emmartins The description of WFLY-13308 talks about the ${ee.maven.groupId} etc being wrong, but AIUI the real issue here is there's just no reason to have this dependency on WF dependencyManagement at all.

emmartins:
@bstansberry I really just copied and adapted JIRA description from older issue reported wrt jms-client-bom being in wildfly's dep management too, it certainly shouldn't be there no matter the version

bstansberry (approved):

PR #13161, opened 2020-04-01
[WFLY-13316]: MicroProfile OpenTracing integration requires scope request.

ehsavoie:
* Removing the producer and using a CDI extension injecting
    'synthetic' beans into the deployments.
 * ServletContext for CDI injection is no longer used.

Jira: https://issues.redhat.com/browse/WFLY-13316


microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -19,26 +19,45 @@
...
+        TRACERS.put(classLoader, tracerInstance);
+    }
+
+    public void registerTracerBean(@Observes AfterBeanDiscovery abd) {
+        abd.addBean().addTransitiveTypeClosure(Tracer.class).produceWith(i -> {
+            return TRACERS.get(Thread.currentThread().getContextClassLoader());
+        });
     }
 
+    // TODO I suppose this somehow prevents other tracer impls from being picked up? Needs confirmation
```
manovotn, line 49: Remove this TODO, according to the log message, it is intentional to remove any other Trace implementation.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
```diff
@@ -210,6 +211,7 @@ private void injectTracer(DeploymentPhaseContext deploymentPhaseContext, Capabil
                 tracer = WildFlyTracerFactory.getTracer(tracerConfigurationName, serviceName);
             }
         }
+        TracingCDIExtension.registerApplicationTracer(moduleCL, tracer);
         deploymentUnit.addToAttachmentList(ServletContextAttribute.ATTACHMENT_KEY, new ServletContextAttribute(SMALLRYE_OPENTRACING_SERVICE_NAME, serviceName));
         deploymentUnit.addToAttachmentList(ServletContextAttribute.ATTACHMENT_KEY, new ServletContextAttribute(SMALLRYE_OPENTRACING_TRACER, tracer));
```
manovotn, line 216: You could probably remove this now, the only usage is in `TracerDynamicFeature` which again needs req. context active (not sure if it can be a problem there). Instead, you could either `@Inject` the `Tracer` if that's supported (not sure now, I might just be confused from Quarkus) or use `CDI.current()` to grab it.

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -19,26 +19,45 @@
...
-        String extensionName = TracingCDIExtension.class.getName();
-        bbd.addAnnotatedType(manager.createAnnotatedType(TracerProducer.class), extensionName + "-" + TracerProducer.class.getName());
+    private static final Map<ClassLoader, Tracer> TRACERS = Collections.synchronizedMap(new WeakHashMap<>());
+
+    public static void registerApplicationTracer(ClassLoader classLoader, Tracer tracerInstance) {
+        TRACERS.put(classLoader, tracerInstance);
+    }
+
+    public void registerTracerBean(@Observes AfterBeanDiscovery abd) {
+        abd.addBean().addTransitiveTypeClosure(Tracer.class).produceWith(i -> {
```
manovotn, line 44: Maybe also add `.scope(ApplicationScoped.class)`. This won't have any functional impact (apart from using proxy) as you are already sharing the instance, but it is a "cleaner" way to declare a bean that in fact is app-wide.

manovotn (changes requested):

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
bstansberry: @ehsavoie Did this input get dropped?

manovotn: We talked on zulip. Apparently the proxy interferes with some places where the contextual instance is needed.

ehsavoie: I'm currently trying to fix the test suite which is talking more time then I thought

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
ehsavoie: It broke a lot of tests related to CDI

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
ehsavoie: This change breaks with ear as the tracer will now be shared across all the ear and so all its wars

manovotn (approved):

bstansberry (approved):

PR #13163, opened 2020-04-01
[WFLY-12899] Set arbitrary property in surefire env rather depend on …

baranowb:
…env.USER - or similar

ISSUE: https://issues.redhat.com/browse/WFLY-12899


bstansberry (approved):

bstansberry:
The Windows - JDK 11 issue is unrelated to this.

PR #13165, opened 2020-04-01
[WFLY-13318] / [WFLY-13304] NPE Activating JWT without realm name and test cases.

darranl:
https://issues.redhat.com/browse/WFLY-13318

https://issues.redhat.com/browse/WFLY-13304


pom.xml
```diff
@@ -265,6 +265,7 @@
         <version.com.google.guava>25.0-jre</version.com.google.guava>
         <version.com.h2database>1.4.197</version.com.h2database>
         <version.com.microsoft.azure>6.1.0</version.com.microsoft.azure>
+        <version.com.nimbus.jose-jwt>5.4</version.com.nimbus.jose-jwt>
```
bstansberry, line 268: This is a couple years old; project is up to 8.11

https://mvnrepository.com/artifact/com.nimbusds/nimbus-jose-jwt

If this is the one we need, ok, but I'm checking as we'll get prompts to upgrade just because, and if security scanners identify anything with this one we'll need to deal with it.

darranl: I will quickly fix that one, once we move to a later version of SmallRye JWT we can likely drop it as it is just being used for the token generation on the client side of the test.

bstansberry:
@darranl 

PR fails CI at https://ci.wildfly.org/viewLog.html?buildId=196102&buildTypeId=WF_PullRequest_GalleonLinuxJdk11
which is due to...
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile/pom.xml#L207
just need to add microprofile-jwt to the set of layers

bstansberry (approved):

PR #13168, opened 2020-04-02
WFLY-12709: MP REST Client TCK addition to testsuite

asoldano:
https://issues.redhat.com/browse/WFLY-12709 Add MP REST Client TCK modulel

### DIFF
--- a/testsuite/integration/microprofile-tck/rest-client/pom.xml
+++ b/testsuite/integration/microprofile-tck/rest-client/pom.xml
@@ -193,7 +193,7 @@
                 <configuration>
                     <offline>true</offline>
                     <jboss-home>${basedir}/target/wildfly</jboss-home>
-                    <stdout>${project.build.directory}/wildfly-wiremock.log</stdout>
+                    <stdout>${project.build.directory}/wildfly/standalone/log/wildfly-wiremock.log</stdout>
                     <java-opts>${modular.jdk.args}</java-opts>
                     <system-properties>
                         <maven.repo.local>${settings.localRepository}</maven.repo.local>
@@ -273,6 +273,11 @@
                         <!-- We are not going to start a Jetty server within WildFly; test coverage ensured by having
                              the tck run in RESTEasy too -->
                         <exclude>org.eclipse.microprofile.rest.client.tck.ssl.SslContextTest</exclude>
+                        <!-- https://github.com/eclipse/microprofile-rest-client/issues/265 -->
+                        <exclude>org.eclipse.microprofile.rest.client.tck.timeout.TimeoutBuilderIndependentOfMPConfigTest</exclude>
+                        <exclude>org.eclipse.microprofile.rest.client.tck.timeout.TimeoutTest</exclude>
+                        <exclude>org.eclipse.microprofile.rest.client.tck.timeout.TimeoutViaMPConfigTest</exclude>
+                        <exclude>org.eclipse.microprofile.rest.client.tck.timeout.TimeoutViaMPConfigWithConfigKeyTest</exclude>
                     </excludes>
                 </configuration>
             </plugin>
--- a/testsuite/integration/microprofile-tck/rest-client/pom.xml
+++ b/testsuite/integration/microprofile-tck/rest-client/pom.xml
@@ -1,0 +1,378 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2019, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ts-integ-mp</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>21.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+    <artifactId>wildfly-ts-integ-mp-rest-client</artifactId>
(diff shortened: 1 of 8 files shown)


testsuite/integration/microprofile-tck/rest-client/pom.xml
```diff
@@ -0,0 +1,275 @@
...
+                        </scripts>
+                        </configuration>
+                     </execution>
+                  </executions>
+                  <configuration>
+                      <offline>true</offline>
+                      <jboss-home>${basedir}/target/wildfly</jboss-home>
+                      <stdout>${project.build.directory}/wildfly-wiremock.log</stdout>
+                      <java-opts>${modular.jdk.args}</java-opts>
+                      <system-properties>
```
bstansberry, line 195: @asoldano 

add

<maven.repo.local>${settings.localRepository}</maven.repo.local>

bstansberry: for example 

https://github.com/wildfly/wildfly/blob/master/testsuite/integration/clustering/pom.xml#L163
https://github.com/wildfly/wildfly/blob/master/testsuite/compat/pom.xml#L341
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/manualmode/pom.xml#L166

etc

bstansberry:
@asoldano 

This differs a bit structurally vs the other mp-tck modules in terms of things we added during the late MP 3.3 work.

1) In the standard test execution, the other TCKs test against the standard build, but the using the standalone-microprofile.xml config.
2) When -Dts.layers is set (which is what the 'Galleon Linux - JDK 11' job does, plus some nightly jobs testing master), then a ts.layers profile is executed. It turns off copying the standard build and instead uses Galleon to provision the minimal server needed.

The effect of this is we are testing our standard build and standalone-microprofile.xml, but also testing a server that has been heavily slimmed by galleon, which provides some coverage of that use case.

What you have here is close to that. And you probably don't want to precisely follow that, as the 'test the standard build' thing uses build/target/wildfly-xxx/modules for it's modules path, and you don't want your CLI 'module add' thing adding a module in 'build'.  But I think a few tweaks will make it match close enough:

1) In your Galleon provisioning, i.e. https://github.com/wildfly/wildfly/pull/13168/files#diff-bc7faecd1bde652428d2c20d6c775041R231 add a 'config' section for standalone-microprofile.xml.

2) Create a ts.layers profile, activated with -Dts.layers, e.g. 
https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/opentracing/pom.xml#L146

3) Within it add another 'galleon-maven-plugin' section:

a) disable your main 'mp-server-provisioning' execution by adding an execution with the same name but ```<goal>none</goal>
b) add one like https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/opentracing/pom.xml#L170. This provisions a slimmer server. But in the 'layers' section (e.g. https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/opentracing/pom.xml#L206) if it is not necessary to provision the 'observability' layer to pass TCK, let's not.

4) In arquillian.xml configure it to use standalone-microprofile.xml

https://github.com/wildfly/wildfly/pull/13168/files#diff-f001c2a53e6f4e789d218226f8a63958R9

5) Your script should modify standalone-microprofile.xml since that is what is tested:

https://github.com/wildfly/wildfly/pull/13168/files#diff-b7026ba05c3e7d355c69a0e5eeced845R1

asoldano:
hi @bstansberry , I tried applying the changes you're proposing, however I can't find a combination of galleon layers allowing passing the TCK, basically because of the need of org.apache.commons.lang3 module. Is there a layer including that?
Anyway, looking at the CI results, I'm seeing some tests failing because of timeouts, as if the tests would need to be more lenient to allow our env passing it. So while we sort that, I've created a PR with the RESTEasy upgrade only (https://github.com/wildfly/wildfly/pull/13174) which we might want to start merging.

bstansberry (commented):
@asoldano A look at 'git grep commons.lang3 | grep module.xml' shows all uses lead to webservices stuff.  There's no layer for that (yet), but in any case adding a future webservices layer to ts.layers would pretty much defeat the purpose of running it that way.

I'll poke to see if there's a way to force that package in using the galleon maven plugin config.

I had some minor review comments too that I forgot to post.

Good call on splitting out the upgrade.

bstansberry:
@asoldano re commons.lang3, to fix it, in the element analogous to https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/health/pom.xml#L143, add this:

```
<!-- The wiremock module we add as part of the test fixture needs Apache Commons Lang 
         but the layers we want don't result in it being provisioned, so explicitly tell Galleon to 
         provision it. -->
<included-packages>
  <package>org.apache.commons.lang3</package>
 </included-packages>
```

I took what you have here, put that in, reduced what's provisioned down to layers "jaxrs-server" and "microprofile-config" and the TCK passed locally.

testsuite/integration/microprofile-tck/rest-client/pom.xml
```diff
@@ -0,0 +1,373 @@
...
+                <executions>
+                    <execution>
+                        <id>add-wiremock-module</id>
+                        <phase>process-test-resources</phase>
+                        <goals>
+                            <goal>execute-commands</goal>
+                        </goals>
+                        <configuration>
+                            <commands>
+                                <command>module add --name=com.github.tomakehurst.wiremock --resources=${project.build.directory}/wildfly/wiremock/wiremock.jar:${project.build.directory}/wildfly/wiremock/junit.jar --module-xml=${basedir}/module.xml</command>
```
bstansberry, line 175: @asoldano https://github.com/wildfly/wildfly/pull/13208/commits/33957519ca393e5c1740555ac46743acb5038fdb gets past the Windows failure we're seeing in CI on this PR but then the tests fail anyway with what sounds like some problem with the beans.xml deployment overlay.

```
[16:46:01][Step 2/3] [INFO] Running TestSuite
[16:46:08][Step 2/3] [INFO] RequestHandlerClass from context returned com.github.tomakehurst.wiremock.http.AdminRequestHandler. Normalized mapped under returned 'null'
[16:46:36][Step 2/3] [ERROR] Tests run: 340, Failures: 44, Errors: 0, Skipped: 296, Time elapsed: 35.353 s <<< FAILURE! - in TestSuite
[16:46:36][Step 2/3] [ERROR] arquillianBeforeClass(org.eclipse.microprofile.rest.client.tck.AdditionalRegistrationTest)  Time elapsed: 8.903 s  <<< FAILURE!
[16:46:36][Step 2/3] org.jboss.arquillian.container.spi.client.container.DeploymentException: 
[16:46:36][Step 2/3] Cannot deploy AdditionalRegistrationTest.war: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"AdditionalRegistrationTest.war\".STRUCTURE" => "WFLYSRV0153: Failed to process phase STRUCTURE of deployment \"AdditionalRegistrationTest.war\"
[16:46:36][Step 2/3]     Caused by: org.jboss.as.server.deployment.DeploymentUnitProcessingException: WFLYSRV0196: Failed to get content for deployment overlay WEB-INF/lib/*.jar/META-INF/beans.xml at WEB-INF/lib/*.jar/META-INF/beans.xml
[16:46:36][Step 2/3]     Caused by: java.io.FileNotFoundException: C:\\BuildAgent\\work\\f5da3564a57e9d74\\testsuite\\integration\\microprofile-tck\\rest-client\\target\\wildfly\\standalone\\tmp\\vfs\\temp\\tempae0fca84ca4e4199\\content-90ee14a7fedaf8cb\\WEB-INF\\lib\\*.jar\\META-INF\\beans.xml (The filename, directory name, or volume label syntax is incorrect)"}}}}
[16:46:36][Step 2/3] 	at org.jboss.as.arquillian.container.ArchiveDeployer.deployInternal(ArchiveDeployer.java:184)
[16:46:36][Step 2/3] 	at org.jboss.as.arquillian.container.ArchiveDeployer.deployInternal(ArchiveDeployer.java:162)
```

bstansberry: If this is too much hassle I think just using a profile to turn this off on windows would be ok.

bstansberry (approved):

asoldano:
The single failure on Windows is unrelated, merging as agreed with Brian.

PR #13170, opened 2020-04-02
[WFLY-13334]:  MP Opentracing tests are not passing with a SecurityMa…

ehsavoie:
…nager

 * Adding the missing permissions

Jira: https://issues.redhat.com/browse/WFLY-13334

### DIFF
--- a/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
+++ b/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
@@ -31,6 +31,7 @@ import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.BeforeShutdown;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ProcessAnnotatedType;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 public class TracingCDIExtension implements Extension {
 
@@ -42,7 +43,7 @@ public class TracingCDIExtension implements Extension {
 
     public void registerTracerBean(@Observes AfterBeanDiscovery abd) {
         abd.addBean().addTransitiveTypeClosure(Tracer.class).produceWith(i -> {
-            return TRACERS.get(Thread.currentThread().getContextClassLoader());
+            return TRACERS.get(WildFlySecurityManager.getCurrentContextClassLoaderPrivileged());
         });
     }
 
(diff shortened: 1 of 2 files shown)


bstansberry (dismissed):
As we just chatted, we shouldn't need all those permissions.xml additions to the test deployments, TracerCDIExtension.beforeShutdown needs a privileged block for its getClassLoader() call.

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -42,7 +45,18 @@ public static void registerApplicationTracer(ClassLoader classLoader, Tracer tra
...
+                loader = Thread.currentThread().getContextClassLoader();
+            } else {
+                loader = AccessController.doPrivileged(new PrivilegedAction<ClassLoader>() {
+                    @Override
+                    public ClassLoader run() {
+                        return Thread.currentThread().getContextClassLoader();
+                    }
+                });
+            }
+            return TRACERS.get(loader);
```
jamezp, line 59: This could just be `TRACERS.get(WildFlySecurityManager.getCurrentContextClassLoader())`

ehsavoie: yes that was the issue

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -42,7 +45,18 @@ public static void registerApplicationTracer(ClassLoader classLoader, Tracer tra
 
     public void registerTracerBean(@Observes AfterBeanDiscovery abd) {
         abd.addBean().addTransitiveTypeClosure(Tracer.class).produceWith(i -> {
-            return TRACERS.get(Thread.currentThread().getContextClassLoader());
+            ClassLoader loader;
```
darranl, line 48: This entire block can be removed and instead just call WildFlySecurityManager.getCurrentContextClassLoaderPrivileged()

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -57,6 +71,17 @@ public void skipTracerBeans(@Observes ProcessAnnotatedType<? extends Tracer> pro
      * @param bs
      */
     public void beforeShutdown(@Observes final BeforeShutdown bs) {
-        TRACERS.remove(Thread.currentThread().getContextClassLoader());
+        ClassLoader loader;
```
darranl, line 74: This entire block can be removed and instead just call WildFlySecurityManager.getCurrentContextClassLoaderPrivileged()

Additionally with this current approach two new classes needed to be defined and loaded to achieve a common task.

darranl (changes requested):
This can be simplified to use a single call against the WildFlySecurityManager.

darranl (approved):

PR #13173, opened 2020-04-03
[WFLY-19905] Move module org.apache.xml-resolver from WildFly Core to WildFly

parsharma:
JIRA: https://issues.redhat.com/browse/WFLY-11905

### DIFF
--- a/ee-galleon-pack/pom.xml
+++ b/ee-galleon-pack/pom.xml
@@ -3978,6 +3978,11 @@
             </exclusions>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>xml-resolver</groupId>
+            <artifactId>xml-resolver</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/pom.xml
+++ b/pom.xml
@@ -240,6 +240,7 @@
         <version.verifier.plugin>1.1</version.verifier.plugin>
         <version.xml.plugin>1.0.1</version.xml.plugin>
 
+
         <!--
             *Non-plugin* Dependency versions. Please keep alphabetical.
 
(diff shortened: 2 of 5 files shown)


bstansberry:
this is ok to test

ee-galleon-pack/pom.xml
```diff
@@ -3978,6 +3978,11 @@
             </exclusions>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>xml-resolver</groupId>
+            <artifactId>xml-resolver</artifactId>
+ 	    <scope>provided</scope>
```
bstansberry, line 3984: There's something odd here with whitespace. When I use the View File option and see the whole file these line up. Perhaps there is a mix of tabs and spaces here?

pom.xml
```diff
@@ -443,7 +444,7 @@
         <version.sun.jaxb>2.3.3-b02</version.sun.jaxb>
         <version.sun.saaj-impl>1.4.1.SP1</version.sun.saaj-impl>
         <version.wsdl4j>1.6.3</version.wsdl4j>
-        <version.xml-resolver>1.2</version.xml-resolver>
+	<version.xml-resolver>1.2</version.xml-resolver>
```
bstansberry, line 447: Please correct the whitespace.

pom.xml
```diff
@@ -589,6 +590,12 @@
                 <version>${ee.maven.version}</version>
             </dependency>
 
+	   <dependency>
```
bstansberry, line 593: Please correct the whitespace.

bstansberry (changes requested):
Thanks. I made some comments about whitespace but the work in general looks good!

ee-galleon-pack/pom.xml
parsharma: May be, I will check all the comments and update accordingly.

yersan (approved):
Changes look good to me once ident issues are resolved.

The Jira should be followed up by a removal of `org.apache.xml-resolver` on wildfly-core.

bstansberry:
Thanks @parsharma .  Please squash these down to a single commit and add WFLY-19905 to the commit message and this is ready to merge.

PR #13167, opened 2020-04-01
[WFLY-13241] Upgrade IronJacamar from 1.4.20.Final to 1.4.21.Final

ivassile:
https://issues.redhat.com/browse/WFLY-13241

PR #13187, opened 2020-04-07
Revert "[WFLY-13241] Upgrade IronJacamar from 1.4.20.Final to 1.4.21.Final"

bstansberry:
Reverts wildfly/wildfly#13167 as it introduces https://issues.redhat.com/browse/WFLY-13340


bstansberry:
@maeste @ivassile AIUI there is currently no downstream requirement for 1.4.21 as the WFLY-13340 issue means 1.4.22 is required. So it seems better to move WF back to 1.4.20 until WFLY-13340 is resolved.

maeste:
@bstansberry  @ivassile yup rigth, @fl4via  is already on top of WFLY-13340, but it's fine to revert to 1.4.20 until it's fixed

kabir:
Superseded by https://github.com/wildfly/wildfly/pull/13189

PR #13190, opened 2020-04-08
[19.0.x] [WFLY-12834] Add documentation on the enabled-protocols attribute and indicate how it works when OpenSSL 1.1.0 or higher is in use

fjuma:
https://issues.redhat.com/browse/WFLY-12834

Depends on: https://github.com/wildfly/wildfly-core/pull/4138

master PR: https://github.com/wildfly/wildfly/pull/13152


bstansberry:
@jmesnil Can you approve this PR? Mostly from a process point of view. I believe the WFCORE related to this will be in the release used for the next 19 release, so this doc change is appropriate. But you know better.

jmesnil (approved):
Right, I merged Farah's PR for WildFly Core 11.0.1.Final (that will be upgraded for upcoming WildFly 19.y.z release)

PR #13195, opened 2020-04-08
[WFLY-13345] Ignore classloading failures when processing component i…

bstansberry:
…nterceptor annotations

https://issues.redhat.com/browse/WFLY-13345

I'd like some thought on this one from someone besides myself.  I think the kinds of problems that this now ignores are conceptually in the same category as the ClassLoadingUtils.loadClass problems we were already ignoring so the same handling seems ok.  Basically MethodAnnotationAggregator/ClassReflectionIndex are doing deeper probing of the class than simple classloading does but the kinds of problems it might hit are of the same general category.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/InterceptorAnnotationProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/InterceptorAnnotationProcessor.java
@@ -38,6 +38,7 @@ import org.jboss.as.ee.component.ComponentDescription;
 import org.jboss.as.ee.component.EEApplicationClasses;
 import org.jboss.as.ee.component.EEModuleDescription;
 import org.jboss.as.ee.component.InterceptorDescription;
+import org.jboss.as.ee.logging.EeLogger;
 import org.jboss.as.ee.metadata.MetadataCompleteMarker;
 import org.jboss.as.ee.metadata.MethodAnnotationAggregator;
 import org.jboss.as.ee.metadata.RuntimeAnnotationInformation;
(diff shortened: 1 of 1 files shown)


bstansberry:
@chengfang Do you have any thoughts on this?

ee/src/main/java/org/jboss/as/ee/component/deployers/InterceptorAnnotationProcessor.java
```diff
@@ -82,13 +83,14 @@ private void processComponentConfig(final EEApplicationClasses applicationClasse
...
+            handleAnnotations(applicationClasses, deploymentReflectionIndex, componentClass, description);
         } catch (Throwable e) {
             //just ignore the class for now.
             //if it is an optional component this is ok, if it is not an optional component
             //it will fail at configure time anyway
+            EeLogger.ROOT_LOGGER.debugf(e,"Ignoring failure to handle interceptor annotations for %s", description.getComponentClassName());
             return;
         }
-        handleAnnotations(applicationClasses, deploymentReflectionIndex, componentClass, description);
     }
```
chengfang, line 94: LGTM. Some minor points: we can now remove `throws DeploymentUnitProcessingException` since the method already catches all exceptions. `return;` is now also redundant. `final <class> componentClass` can be combined with assignment.

chengfang (approved):

ee/src/main/java/org/jboss/as/ee/component/deployers/InterceptorAnnotationProcessor.java
bstansberry: Thanks; I cleaned those things up.

PR #13197, opened 2020-04-09
WFLY-13359 Upgrade JGroups to 4.2.2.Final

rhusar:
Resolves https://issues.redhat.com/browse/WFLY-13359

Most importantly brings in fix for 'JGRP-2412 GMS: reduce likelihood of merges on concurrent new member startup' which could resolve the issue we have been seeing on CI.


rhusar:
For @pferraro to merge.

pferraro (approved):
We can analyze later whether the changes in this release are sufficient to re-enable the default transport bundler.

PR #13200, opened 2020-04-09
[WFLY-13360] : Log INFO when EJB does not implement Business interface

panossot:
[WFLY-13360] : https://issues.redhat.com/browse/WFLY-13360

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
@@ -49,6 +49,7 @@ import org.jboss.metadata.ejb.spec.EjbJarVersion;
 import org.jboss.modules.Module;
 
 import static org.jboss.as.ee.component.Attachments.EE_MODULE_DESCRIPTION;
+import static org.jboss.as.ejb3.deployment.processors.ViewInterfaces.potentialNotExposedViewInterfacesInfo;
 import static org.jboss.as.ejb3.deployment.processors.ViewInterfaces.getPotentialViewInterfaces;
 
 /**
@@ -150,6 +151,8 @@ public class BusinessViewAnnotationProcessor implements DeploymentUnitProcessor 
                 sessionBeanComponentDescription.addLocalBusinessInterfaceViews(toString(potentialBusinessInterfaces));
             }
         }
+
+        potentialNotExposedViewInterfacesInfo(sessionBeanClass);
     }
 
     private Collection<Class<?>> getRemoteBusinessInterfaces(final DeploymentUnit deploymentUnit, final Class<?> sessionBeanClass) throws DeploymentUnitProcessingException {
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/ViewInterfaces.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/ViewInterfaces.java
@@ -29,6 +29,7 @@ import java.io.Serializable;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import org.jboss.as.ejb3.logging.EjbLogger;
 
 /**
  * @author <a href="mailto:cdewolf@redhat.com">Carlo de Wolf</a>
@@ -56,6 +57,7 @@ class ViewInterfaces {
             }
             potentialBusinessInterfaces.add(klass);
         }
+
         return potentialBusinessInterfaces;
     }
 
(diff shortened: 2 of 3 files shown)


panossot:
Almost done ... Needs some logging modifications ...

panossot:
Should be done now ...

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/ViewInterfaces.java
```diff
@@ -85,4 +87,48 @@
...
+                    }
+                    potentialNotExposedBusinessInterfaces.add(klass);
+                    EjbLogger.ROOT_LOGGER.potentialInterfaceExposure(klass.getCanonicalName());
+                }
+            }
+
+            superClass = superClass.getSuperclass();
+        }
+
+        return potentialNotExposedBusinessInterfaces;
```
bstansberry, line 132: AFAICT this return value is never used, and the var is never read in this method, so the var can be dropped.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
```diff
@@ -218,6 +221,16 @@ private void processViewAnnotations(final DeploymentUnit deploymentUnit, final C
...
 
+    /**
+     * Returns interfaces implemented by a bean that are eligible to be business interfaces and not exposed
+     *
+     * @param sessionBeanClass The bean class
+     * @return A collection of potential business interfaces that are not exposed
+     */
+    private static Set<Class<?>> getPotentialNotExposedBusinessInterfaces(Class<?> sessionBeanClass) {
+        return getPotentialNotExposedViewInterfaces(sessionBeanClass);
+    }
```
bstansberry, line 232: I'm not clear on why this method is needed as it just calls another method.

bstansberry:

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -3204,4 +3204,8 @@
     @LogMessage(level = WARN)
     @Message(id = 515, value = "[EJB3.2 spec, section 4.9.2] Singleton session beans are not allowed to implement 'javax.ejb.SessionBean' interface. This interface on bean '%s' is going to be ignored and should be removed.")
     void singletonCantImplementSessionBean(String className);
+
+    @LogMessage(level = WARN)
+    @Message(id = 516, value = "Potential Interface view not exposed... : %s")
```
bstansberry, line 3209: I'm not sure if WARN is appropriate here as this might be emitted due to a legitimate design choice by the application author.

I think if we log a WARN or above for something that might be ok with the user it should be possible to turn off the log category without losing other important information. That would mean a separate log category. That seems like overkill for this though. Perhaps INFO suffices, since my impression is the goal here is to help users understand why something doesn't work as they expect. If something isn't working reading the INFO level logging emitted as the app is deployed is a reasonable thing to expect.

chengfang:
It's hard to guess the users' intent when some indirect interfaces are not exposed as business interfaces: users might have forgot to do it while still expecting it as a business interface, or the users really do not want that indirect interface as business interface.

Since this is intended to be an extra help for users, I think we should respect the currect application configuration as much as possible, and only offer to help when it's most likely to be user's oversight, that is, when no interface is directly implemented, and no business interface is specified.  This is the time when we currently assume it only expose a local bean no interface view.  We can probably add such check here:

https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java#L143

panossot:
Second testcase for possible additional interface view from session bean that already exposes one ...
: https://github.com/jboss-set/eap-additional-testsuite/pull/135

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/ViewInterfaces.java
bstansberry: Resolved.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
bstansberry: Resolved.

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
bstansberry: Resolved.

bstansberry:
The concerns I raised here have been addressed but I'll defer any approval to Cheng Fang as it's his subsystem and his comments are about the substance of the change.

panossot:
@chengfang  I have added a second testcase where this happens also when ejbs exist other than the ones with no interface view ...

wolfc:
Not a bug.

bstansberry:
As there's questions as to whether this is appropriate, I'm going to close it to get the queue better under control. If there's agreement reached offline about this, then please file a new PR.

PR #13203, opened 2020-04-09
WFLY-13015 Consolidate common instances of DiscardAttributeChecker that discard an attribute set to the default value

pferraro:
https://issues.redhat.com/browse/WFLY-13015

Follows up on https://issues.redhat.com/browse/WFCORE-4814

N.B. Requires wildfly-core 12.0.0.Beta1.


bstansberry:
This looks good to me, but @chengfang or @tadamski please have a look.

@pferraro This part seems fine, but I think DiscardAttributeChecker.DEFAULT can be made more efficient in core.

See:

https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/registry/AbstractResourceRegistration.java#L172

https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/registry/AbstractResourceRegistration.java#L209

Basically, in order to deal with the possibility of override registrations, calls to an MRR typically result in it internally working up to the root of the MRR tree and then walking back down to the target address.  So this:

```
ImmutableManagementResourceRegistration registration = context.getResourceRegistrationFromRoot(address);
            AttributeDefinition definition = registration.getAttributeAccess(PathAddress.EMPTY_ADDRESS, attributeName).getAttributeDefinition();
```

is less efficient than:

```
ImmutableManagementResourceRegistration registration = context.getResourceRegistrationFromRoot(PathAddress.EMPTY_ADDRESS);
            AttributeDefinition definition = registration.getAttributeAccess(address, attributeName).getAttributeDefinition();
```

pferraro:
@bstansberry Thanks Brian.  I've created https://issues.redhat.com/browse/WFCORE-4918 to address this in WFCORE.

chengfang (approved):

bstansberry (approved):

PR #12538, opened 2019-08-27
[WFLY-12218] Add support for automatically adding and updating credentials in a credential store

fjuma:
https://issues.jboss.org/browse/WFLY-12218

Just FYI for reviewers, there are two commits in this PR for each of the sub-task JIRAs below - one commit is for the model and schema bump and the second commit updates the new schema to reference the "urn:wildfly:credential-reference:1.1" namespace.

Sub-task JIRAs:
https://issues.redhat.com/browse/WFLY-13363 (jgroups)
https://issues.redhat.com/browse/WFLY-13364 (datasources)
https://issues.redhat.com/browse/WFLY-13365 (resource-adapters)
https://issues.redhat.com/browse/WFLY-13366 (agroal)
https://issues.redhat.com/browse/WFLY-13367 (mail)
https://issues.redhat.com/browse/WFLY-13368 (messaging-activemq)
https://issues.redhat.com/browse/WFLY-13369 (undertow)

This PR also includes https://github.com/wildfly/wildfly/pull/13204 and reverts https://github.com/wildfly/wildfly/pull/13204/commits/56d2394985036f5a5b75aa352a6e533fb1d4f1c1.

PR #13204, opened 2020-04-10
[WFLY-12510] Include wildfly-credential-reference_1_1.xsd when validating config

jmesnil:
This PR reverts https://github.com/wildfly/wildfly/commit/ea885a15dbb2adf3cd9ab9afc7832bed672e4669 and includes wildfly-credential-reference_1_1.xsd during the resolution of namespaces in the StandardConfigsXMLValidationUnitTestCase tests


darranl (commented):
This PR appears to duplicate the correction also included in this PR https://github.com/wildfly/wildfly/pull/12538

PR #13205, opened 2020-04-10
[WFLY-13370] Eliminate fault-tolerance/executor dependence on ee subs…

bstansberry:
…ystem module

https://issues.redhat.com/browse/WFLY-13370


bstansberry:
@rhusar FYI

rhusar (changes requested):
Thanks @bstansberry , the idea is good, but the PR needs a small change as the tests fail with

```
    Caused by: java.lang.NoClassDefFoundError: javax/enterprise/concurrent/ManagedThreadFactory
```

I assume `galleon-pack/src/main/resources/modules/system/layers/base/org/wildfly/microprofile/fault-tolerance-smallrye/executor/main/module.xml` needs to be updated as well, but I can look during next working day in detail :-)

pferraro (approved):

rhusar (approved):
Looks good now!

PR #13218, opened 2020-04-16
[WFLY-13381] At EJB3 remote AssociationImpl, we need to take into acc…

fl4via:
…ount the possibility of identity being null before setting up connection in SecurityActions.

With EJB over Remoting, it is always true that the identity won't be null because of Sasl anonymous user. But the same does not apply to EJB over HTTP. In this
case, security identity is null if http-remoting-connector lacks the security-realm attribute (at remoting subsystem model config):
<http-connector name="http-remoting-connector" connector-ref="default"/>

Jira: https://issues.redhat.com/browse/WFLY-13381
EAP Jira: https://issues.redhat.com/browse/JBEAP-16383


darranl (approved):
Looks good for me.

Later on in the call stack there is an assumption that if this RemoteConnection is set it will always return a SecurityIdentity so safer overall to just not set this if all it will do is return null.

fl4via:
Thanks for the feedback, @darranl !

PR #13219, opened 2020-04-16
Remove licenses for artifacts that are not part of final zip

rgdoliveira:


rgdoliveira:
@bstansberry @jamezp if you can please review this one.

jamezp (approved):
This seems okay to me, but I'd like to get @scottmarlow and @jfdenise to review as well to make sure these aren't meant to end up in the final distribution.

jfdenise:
@jamezp , the artifacts are not in the zip, +1 to remove them from the licenses.

jfdenise (approved):

bstansberry (approved):
Hibernate 5.3 is what goes in the dist; 5.1 is an option in the source that people can install themselves.

PR #13221, opened 2020-04-17
[WFLY-13267] Convert EJB3 subsystem transformers to use chained transformers

rachmatowicz:
This PR does the following:
- adds in transformer tests for the ejb3 subsystem for EAP versions 7.0.0, 7.1.0 and 7.2.0
- refactors the transformers to use the chained transformer approach, to simplify understanding of the transformers and make maintenance easier  

For more details, see: https://issues.redhat.com/browse/WFLY-13267

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
@@ -28,6 +28,12 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OPE
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UNDEFINE_ATTRIBUTE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_1_2_1;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_1_3_0;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_3_0_0;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_4_0_0;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_5_0_0;
+import static org.jboss.as.ejb3.subsystem.EJB3Model.VERSION_6_0_0;
 import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.ALLOW_EXECUTION;
 import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.CLIENT_MAPPINGS_CLUSTER_NAME;
 import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.DEFAULT_SFSB_CACHE;
@@ -57,24 +63,22 @@ import org.jboss.as.controller.transform.ResourceTransformationContext;
 import org.jboss.as.controller.transform.SubsystemTransformerRegistration;
 import org.jboss.as.controller.transform.TransformationContext;
 import org.jboss.as.controller.transform.description.AttributeConverter;
+import org.jboss.as.controller.transform.description.ChainedTransformationDescriptionBuilder;
 import org.jboss.as.controller.transform.description.DiscardAttributeChecker;
 import org.jboss.as.controller.transform.description.RejectAttributeChecker;
 import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
-import org.jboss.as.controller.transform.description.TransformationDescription;
 import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
 import org.jboss.as.ejb3.logging.EjbLogger;
 import org.jboss.as.threads.PoolAttributeDefinitions;
 import org.jboss.dmr.ModelNode;
 
 /**
+ * EJB Transformers used to transform current model version to legacy model versions for domain mode.
+ *
  * @author Tomaz Cerar (c) 2017 Red Hat Inc.
+ * @author Richard Achmatowicz (c) 2020 Red Hat Inc.
  */
 public class EJBTransformers implements ExtensionTransformerRegistration {
-    private static final ModelVersion VERSION_1_2_1 = ModelVersion.create(1, 2, 1);
-    private static final ModelVersion VERSION_1_3_0 = ModelVersion.create(1, 3, 0);
-    private static final ModelVersion VERSION_3_0_0 = ModelVersion.create(3, 0, 0);
-    private static final ModelVersion VERSION_4_0_0 = ModelVersion.create(4, 0, 0);
-    private static final ModelVersion VERSION_5_0_0 = ModelVersion.create(5, 0, 0);
 
     @Override
     public String getSubsystemName() {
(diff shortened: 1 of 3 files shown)


rachmatowicz:
Hi @pferraro , any feedback on how to improve this would be appreciated. In other words, waiting on review.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
```diff
@@ -123,7 +123,6 @@
             new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.ENABLE_STATISTICS, ModelType.BOOLEAN, true)
                     .setAllowExpression(true)
                     .setDeprecated(ModelVersion.create(5))
-                    .setFlags(AttributeAccess.Flag.ALIAS)
```
bstansberry, line 126: This attribute is an alias.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemXMLPersister.java
```diff
@@ -489,8 +491,10 @@ private void writeStrictMaxPoolConfig(final XMLExtendedStreamWriter writer, fina
 
         writer.writeAttribute(EJB3SubsystemXMLAttribute.NAME.getLocalName(), strictMaxPoolModel.getName());
 
-        StrictMaxPoolResourceDefinition.MAX_POOL_SIZE.marshallAsAttribute(strictMaxPoolModelNode, writer);
-        StrictMaxPoolResourceDefinition.DERIVE_SIZE.marshallAsAttribute(strictMaxPoolModelNode, writer);
+        if (strictMaxPoolModelNode.hasDefined(MAX_POOL_SIZE))
+            StrictMaxPoolResourceDefinition.MAX_POOL_SIZE.marshallAsAttribute(strictMaxPoolModelNode, writer);
+        if (strictMaxPoolModelNode.hasDefined(DERIVE_SIZE))
+            StrictMaxPoolResourceDefinition.DERIVE_SIZE.marshallAsAttribute(strictMaxPoolModelNode, writer);
```
bstansberry, line 497: AFAICT the changes to this file should not be necessary as the behavior of these AttributeDefinitions is to not write anything if the attribute isn't defined.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -77,6 +77,14 @@ protected void performBoottime(OperationContext context, ModelNode operation, Mo
         installRuntimeServices(context, model);
     }
 
+    @Override
+    protected void populateModel(ModelNode operation, ModelNode model) throws OperationFailedException {
+        EJB3RemoteResourceDefinition.CLIENT_MAPPINGS_CLUSTER_NAME.validateAndSet(operation, model);
+        EJB3RemoteResourceDefinition.CONNECTOR_REF.validateAndSet(operation, model);
+        EJB3RemoteResourceDefinition.THREAD_POOL_NAME.validateAndSet(operation, model);
+        EJB3RemoteResourceDefinition.EXECUTE_IN_WORKER.validateAndSet(operation, model);
+    }
```
bstansberry, line 86: I think this move introduces merge conflicts because the whole method is gone now.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemXMLPersister.java
rachmatowicz: OK, i'll remove the changes. Thanks.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
rachmatowicz: Not sure I understand. When I open the file in the PR, I see the method. Do you mean it disappears after being merged?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
rachmatowicz: This one is tricky. There is only one instance of ALIAS in the whole of Wildfly and its this one. I discussed with PF and RH and they said don't use it. If I do use it, the transformer tests fail. Is there some description somewhere of how to handle attributes with an ALIAS flag? The description above the definition says, "this attribute is an alias to something else". How are aliases handled in general?

rachmatowicz: What is happening is the model comparison fails: in 6.0.0, statistics-enabled=expression and enable-statistics=expression; in 5.0.0, statistics-enabled=expression and enable-statistics=undefined. In 6.0.0, enable-statistics is flagged as ALIAS and there are custom read-attribute and write-attribute handlers for enable-statistics which instead read and write from statistics-enabled. This explains why the values are equal in 6.0.0. Exactly the same setup in 5.0.0 (EAP 7.2.0, 7.1.0). The operation transformers are creating the correct model, but the resource transformers are creating a model where enable-statistics is undefined. Is this a case where a value needs to be injected if it is undefined?

rachmatowicz: I created an attribute converter for enable-statistics which will check if it is undefined and if it is, set the value to the value of statistics-enabled. I assumed that the undefined value was being created on the initial model transfer; this seems to fix it.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
bstansberry: When I wrote that comment Github was reporting this PR had a merge conflict.  At the time, github was showing the change this PR was making to this file was moving this method from lower in the file to here. I assumed that was the merge conflict because Paul had done a different change to this file that removed this method altogether.

It looks like now the conflict has been resolved but the resolution resulted in restoring this method. I don't think it should be restored. Give the rest of Paul's change all it's doing here is what the superclass implementation will do.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
bstansberry: You've revealed that a bug fix in WF 15 / EAP 7.3 (https://issues.redhat.com/browse/WFCORE-4183) fixes another problem besides the one it targeted. :-) When you invoke read-resource on an EAP 7.2 or earlier controller you get the value of aliases even though the op has the 'include-aliases="false" param set.

This is better to correct in DefaultModelFixer as opposed to changing the production code to send a value that the legacy host does not need. The result from the read-resource operation on the legacy controller is not correct due to that bug, so better to correct it in the test.

With https://github.com/bstansberry/wildfly/commit/fd3d27fe6cdf1f3fe453b72993026e35ac3c7c21 added to what you have here the EJB3 subsystem tests pass. Feel free to squash that in if you agree it is correct.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
rachmatowicz: Removed the method populateModel() as it is no longer in master..

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
rachmatowicz: Added in your code snippet to fix the issue in the ModelFixer. Thanks, Brian.

bstansberry (approved):

rachmatowicz:
I think there is a problem here with overwriting changes from WFLY-13015, fixing them now.

rachmatowicz:
This is going to take a while; now my tests are failing because of the changes to DiscardAttributeChecker.DEFAULT replacing the original default values. I've got to go through and check each change...

chengfang (approved):

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
```diff
@@ -83,202 +88,199 @@ public String getSubsystemName() {
...
-
-        TransformationDescription.Tools.register(builder.build(), subsystemRegistration, version);
-    }
-
-    private static void registerTransformers_3_0_0(SubsystemTransformerRegistration subsystemRegistration) {
-        final ResourceTransformationDescriptionBuilder builder = TransformationDescriptionBuilder.Factory.createSubsystemInstance();
-        builder.getAttributeBuilder().setValueConverter(AttributeConverter.Factory.createHardCoded(new ModelNode("hornetq-ra"), true), EJB3SubsystemRootResourceDefinition.DEFAULT_RESOURCE_ADAPTER_NAME)
+        subsystemBuilder.getAttributeBuilder()
+                .setDiscard(new DiscardAttributeChecker.DiscardAttributeValueChecker(ModelNode.TRUE), EJB3SubsystemRootResourceDefinition.LOG_EJB_EXCEPTIONS)
+                .addRejectCheck(RejectAttributeChecker.DEFINED, EJB3SubsystemRootResourceDefinition.LOG_EJB_EXCEPTIONS)
```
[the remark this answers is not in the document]
rachmatowicz: @chengfang Thanks, that instance is fixed in the current state of the PR.

bstansberry:
I removed the "Ready for merge" based on the last couple comments.

rachmatowicz:
It's ready now...I just missed a couple of places where Pauls previous commit got reverted by my commit during the merge

PR #13222, opened 2020-04-17
[WFLY-13386] Hung process instances and associated server.log WARN "Failed to reinstate timer"

elguardian:
upstream
https://issues.redhat.com/browse/WFLY-13386

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -828,31 +828,29 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service<Datab
                             while (resultSet.next()) {
                                 try {
                                     String id = resultSet.getString(1);
-                                    if (!existing.remove(id)) {
+                                    TimerImpl oldTimer = listener.getTimerService().getTimer(id);
+                                    if (!existing.remove(id) && oldTimer == null) {
                                         synchronized (DatabaseTimerPersistence.this) {
                                             knownTimerIds.get(timedObjectId).add(id);
                                         }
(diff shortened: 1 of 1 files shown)


bstansberry:
@chengfang or @tadamski Please review

elguardian:
added some comments for clarification.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -844,15 +844,22 @@ public void run() {
...
-                                                    timers.remove(oldTimer.getId());
+                                                    knownTimerIds.get(timedObjectId).remove(oldTimer.getId());
                                                     listener.timerSync(oldTimer, holder.timer);
                                                 }
-
+                                            } else if (validDBTimer && oldTimer == null) {
+                                                // timer migration among nodes.
+                                                // persistence tier and timer subsystem is out of sync
+                                                // we were able to delete it from the persistence tier but it does not exists in timer service
+                                                // need to add it again to the persistence tier and register it in timer service if is valid
```
chengfang, line 858: I don't understand why "we were able to delete it from the persistence tier", because the `dbTimer` is just retrieved from db.

Also why the comment "need to add it again to the persistence tier"? It already exists in db and so we just need to add it to timer service of the current node, which is what the statement below does.

elguardian: @chengfang I ment this class. Sometimes the database timer persistence class is out of sync from the TimerService (it did have the timer associated but it was not registered in the timer service causing a null pointer exception when you are trying to get the timer service).. It is true that it might be missleading the comment

chengfang (approved):

bstansberry:
retest this please

elguardian:
Customer still found another problem. I will throw another PR for this.

PR #13223, opened 2020-04-17
[WFLY-13385] increasing CLI timeout for CLI wrapper operations in testsuite

ochaloup:
https://issues.redhat.com/browse/WFLY-13385

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CloneProfileTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/management/cli/CloneProfileTestCase.java
@@ -29,6 +29,7 @@ import org.jboss.as.test.integration.management.base.AbstractCliTestBase;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import org.jboss.as.test.shared.TimeoutUtil;
 import org.jboss.logging.Logger;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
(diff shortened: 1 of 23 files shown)


kabir:
@ochaloup I am not really familiar with this, but it looks like there is a way to centralise this config a bit more: https://github.com/wildfly/wildfly-core/blob/master/cli/src/main/java/org/jboss/as/cli/impl/CliConfigImpl.java#L105. I see a few instances of jboss-cli.xml scattered around the codebase, so perhaps this can go in there? I have not checked the jboss-cli parser. Perhaps @jfdenise knows?

jfdenise:
@kabir @ochaloup , these failures are related to your CI, right? It doesn't seem that they fail in wildfly CI. As Kabir, suggested, it could be done by introducing a new https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/jboss-cli.xml (to put somewhere in the testsuite) containing the element `<connection-timeout>20000</connection-timeout>`. When launching tests in CI, passing `-Djboss.cli.config=<path to config>`. 
NB: The effect would be for all tests, not only the tests you updated.

ochaloup:
@jfdenise @kabir yes this sounds better. As I'm not familiar with the `jboss-cli.xml` file - it means that I can update the `jboss-cli.xml` with the proposed element and it will have the desired outcome for our CI testing, right?
(If so, I would change the WFLY issue to the WFCORE and create a new PR for the core.)

jfdenise:
@ochaloup , we shouldn't update the existing default jboss-cli.xml, I was suggesting to define a new one (copy the default one and add the connection-timeout element) to be used by testsuite (and configured in your CI).

ochaloup:
@jfdenise ok, I see now. The configuration should be done only on our CI.
Closing this PR then.

PR #13226, opened 2020-04-18
[WFLY-9440] Register requirement for jacc-policy capability

tadamski:
https://issues.redhat.com/browse/WFLY-9440

Similiar fix is needed for Undertow, but for now I wasn't sure whether it makes sense to add implementation of CapabilityReferenceRecorder for use with boolean attributes to core.

- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ApplicationSecurityDomainDefinition.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ApplicationSecurityDomainDefinition.java
@@ -31,6 +31,7 @@ import java.util.function.Function;
 import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.CapabilityServiceBuilder;
+import org.jboss.as.controller.CapabilityReferenceRecorder;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
@@ -49,6 +50,7 @@ import org.jboss.as.controller.access.management.SensitiveTargetAccessConstraint
 import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.operations.validation.StringLengthValidator;
 import org.jboss.as.controller.registry.AttributeAccess;
+import org.jboss.as.controller.registry.ImmutableManagementResourceRegistration;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.ejb3.security.ApplicationSecurityDomainConfig;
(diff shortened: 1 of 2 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ApplicationSecurityDomainDefinition.java
```diff
@@ -236,4 +239,62 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
             return null;
         };
     }
+
+    private static class BooleanCapabilityReferenceRecorder implements CapabilityReferenceRecorder {
```
darranl, line 243: Would this class be better in WildFly Core so it can be shared?  I think there was a duplicate issue raised for the Undertow subsystem so we know at least two will need a solution for this.

tadamski: I wasn't sure because the usage is scarce and imo this class is a bit hacky. But I can push it to core as well.

darranl: Lets see what others reviewing think, but if we know at least two subsystems need the same solution something common may be preferred.

Although one thing we may keep in mind is at some point we will be removing these application-security-domain resources but we are not quite at that point yet.

kabir:
@tadamski I see failures in Ejb3TransformersTestCase.testRejectionsEAP640

tadamski:
@kabir fixed

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ApplicationSecurityDomainDefinition.java
bstansberry: I'm doubtful about putting this in the core 'controller' project. It is shared across two things but I don't think it's a great config style and I'm concerned about promoting it. We need to update Galleon's wildfly plugin to understand that 'true' means there's the requirement.  That's a necessary but pretty low priority thing.

In any case I don't think adding a shared thing should block this; it can be separate.

bstansberry:
@tadamski Please @ ping me when the conflict is resolved so we can stop holding you up by leaving it open.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ApplicationSecurityDomainDefinition.java
darranl: Ok that makes sense @bstansberry TBH I am also thinking about the steps needed to remove these application-security-domain resources so if this is the only use then we just need to make sure any replacement doesn't trigger the same kind of issue to solve.

bstansberry:
@tadamski Can you rebase this so we can clear it out of the PR queue?

tadamski:
@bstansberry rebased

bstansberry (approved):

darranl (approved):

darranl:
Just added an approval as well - just found some other JACC issues that need refactoring.

darranl:
I am having second thoughts if we should actually be doing it like this, I believe the enable-jacc attribute would also be valid if JACC is provided by legacy security and not just via the resource in the Elytron subsystem.

darranl:
I have a different issue I am looking at here:
  https://issues.redhat.com/browse/WFLY-13615

These ApplicationSecurityDomain resources seem to be trying to solve something similar but I am not sure it is correct that the dependency is on the Elytron variant.  Instead I believe the actual JACC services should be depending on the Elytron or PicketBox variant depending on which one is installed instead.

darranl:
@tadamski FYI this presently has a merge conflict.

tadamski:
@darranl, rebased and fixed

darranl:
Thanks @tadamski once CI is completed we will get this in, I did try a local run with my own JACC refactoring and that also looked good.

PR #13240, opened 2020-04-23
Fix for WFLY-13406, Update layersTest to recognize bootable jar runtime

jfdenise:
Impact of fix for https://issues.redhat.com/browse/WFLY-13406
The newly introduced jboss module is not referenced from the module graph when starting from extentions.


bstansberry (approved):

darranl (dismissed):
@jfdenise Could you please quickly create a WFLY issue to track this and update the commit message and title accordingly?  I will then get on with the next stage of merges - as soon as we get this one in the WildFly Core once can quickly follow and get that out the way.

jfdenise:
@darranl , done.

PR #13241, opened 2020-04-27
 Test Case for the issue WFLY-13375, if connection factory or JMS destination is configured for third party resource adapter

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-13554
Issue: https://issues.redhat.com/browse/WFLY-13375

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/JndiViewOperationJMSTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/JndiViewOperationJMSTestCase.java
@@ -1,0 +1,124 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.test.integration.jca.deployment;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.test.integration.jca.JcaMgmtBase;
+import org.jboss.as.test.integration.jca.rar.MultipleAdminObject1;
+import org.jboss.dmr.ModelNode;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.jboss.shrinkwrap.api.spec.ResourceAdapterArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.*;
+
+/**
+ * Test the situation when connection factory or JMS destination is configured for resource adapter
+ *
+ * @author <a href="parsharma@redhat.com">Parul Sharma</a>
+ */
+
+@RunWith(Arquillian.class)
+@RunAsClient
+public class JndiViewOperationJMSTestCase extends JcaMgmtBase {
+
+    static final String rarDeploymentName = "eis.rar";
+
+    @Deployment(name = "rar", order = 1)
+    public static Archive<?> deploytRar() {
+        ResourceAdapterArchive raa = ShrinkWrap.create(ResourceAdapterArchive.class, rarDeploymentName);
+
+        JavaArchive ja = ShrinkWrap.create(JavaArchive.class, "lib.jar");
+        ja.addPackage(MultipleAdminObject1.class.getPackage());
+        raa.addAsLibrary(ja);
+
+        raa.addAsManifestResource(JndiViewOperationJMSTestCase.class.getPackage(), "ra.xml", "ra.xml");
+        return raa;
+    }
+
+    @Test
+    public void testValueOfJms() throws Exception {
+        ModelNode address = null;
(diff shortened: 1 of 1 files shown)


darranl (changes requested):
I have not checked the code itself yet but can the commit message please be updated to something more meaningful describing what the test actually does in a brief sentence.

Obviously the Jira issue does need to be in the commit message but when you come to debug or maintain some five or ten year old classes and the source history ends up full of messages like "Test Case for the issue WFLY-XXXX" it becomes difficult to follow without also having to manually open each Jira issue in turn.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/JndiViewOperationJMSTestCase.java
```diff
@@ -0,0 +1,96 @@
...
+
+     @Test
+     public void testValueOfJms() throws Exception{
+        ModelNode address = getAddress();
+        ModelNode operation = getOperation(address);
+          try{
+             ModelNode result=executeOperation(operation);
+             ModelNode value = result.get("java: contexts", "java:", "JmsXA","value");
+             Boolean val = !(value.equals(null));
+             Assert.assertTrue(val);
```
yersan, line 77: I suggest some changes here:
1.  If `result` is `undefined`, then `val` will be `true` and the test will pass, getting a false negative.
2. `value.equals(null)` never will be true, because in that case you will get a NPE.
3. `Boolean val`, you can use a primitive and avoid auto-boxing/unboxing.

yersan (changes requested):

parsharma:
@darranl @yersan All review comments are incorporated.

soul2zimate:
Does the failure in IronJacamarDeploymentTestCase relates to this change ?

parsharma:
No, I don't think so this test case is related to my change.

soul2zimate:
retest this please

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/ironjacamar.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ironjacamar>
+    <connection-definitions>
+        <connection-definition
+                class-name="org.jboss.as.test.integration.jca.rar.MultipleManagedConnectionFactory1"
+                jndi-name="java:jboss/name1" pool-name="ij_Pool1">
+        </connection-definition>
+    </connection-definitions>
```
yersan, line 8: This connection definition seems it is unused on your test, is so, you can remove it

yersan (changes requested):
@parsharma in case this helps:

The admin-object definition added in the ironjacamar.xml file is affecting to other tests. My vague understanding is you need to properly clean them up because it looks like they are not cleaned at all upon undeployment.

I don't know if it is a hidden bug in JCA or if this behaviour is expected, at any case, this object definition is affecting to `IronJacamarDeploymentTestCase`.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/ironjacamar.xml
gaol: IMO there should be no relationship between the 2 test cases, because the deployment will be undeployed anyway before starting the next one, right? And updating the jndi bindings to something else does not work either.

yersan: > because the deployment will be undeployed anyway before starting the next one, right? 

Yes it is undeployed and that was my initial thought but even upon an undeployment `IronJacamarDeploymentTestCase` fails after being affected by the deployment of `ironjacamar.xml` on `JndiViewOperationJMSTestCase`. I haven't got into the details, but seems it is due to the admin-objects defined there.

parsharma: @yersan yes, it is failing because of admin object.

gaol: @parsharma For now, please use resource adapter's CLI like: 
```
/subsystem=resource-adapters/resource-adapter=eis.rar:add(archive=eis.rar)
/subsystem=resource-adapters/resource-adapter=eis.rar/admin-objects=ij_Pool:add(jndi-name="java:jboss/Name3", class-name="org.jboss.as.test.integration.jca.rar.MultipleAdminObject1Impl")
/subsystem=resource-adapters/resource-adapter=eis.rar:activate()
```
to activate the resource adapter instead of using `ironjacamar.xml`, there maybe a hidden bug in `resource-adapter` subsystem or `ironjacamar` when it is activted using `ironjacamar.xml` as @yersan pointed. :)

You can refer to [this test case](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/ear/RarInsideEarReDeploymentTestCase.java#L74-L100) on how to do it. HTH

parsharma: @gaol  Thanks, your above suggestion works.

yersan (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/deployment/JndiViewOperationJMSTestCase.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private ModelNode getOperation(ModelNode address) {
+        ModelNode operation = new ModelNode();
+        operation.get(OP).set("jndi-view");
+        operation.get(OP_ADDR).set(address);
+        return operation;
+    }
+
+    private void setup() throws Exception {
```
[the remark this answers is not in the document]
parsharma: I think deployment, undeployment and clean up is done by arquillian.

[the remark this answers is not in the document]
gaol: @parsharma The configuration done in `@Test` method won't be clean up by Arquillian. The deployment archive with `@Deployment` will be undeployed by Arquillian. So it should be cleaned. It can be done either by cleaning it within your `@Test` method, or you can use `SnapshotRestoreSetupTask` for the configuration, you can refer to [TracerEnableTestCase#TracerEnableSetup](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/tracer/TracerEnableTestCase.java#L91) on the 2nd option.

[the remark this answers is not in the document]
parsharma: @gaol yes, yesterday I discussed with @yersan, he also suggested same by rolling back changes in the @test method itself. I referred to the same class you have suggested RarInsideEarReDeploymentTestCase. They are also removing address in finally, incase test class fails it will rollback all the changes.

parsharma:
Please let me know if any further changes needed.

yersan (approved):
Looks good to me, thanks @parsharma and @gaol for your help

PR #13249, opened 2020-04-29
[WFLY-13358] Tests for WFCORE-4950 - Regression: Legacy Ldap Realm securing EJB with JDK8 not working

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-13358

Tests for WFCORE-4950. The test class `LdapLegacyTestCase` should fail using jdk-8 until the WFCORE is merged and the version upgraded in wildfly.

PR for the WFCORE issue: https://github.com/wildfly/wildfly-core/pull/4187

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/LdapLegacyTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/security/LdapLegacyTestCase.java
@@ -1,0 +1,269 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


rmartinc:
Added that test is skipped if the elytron configuration is used (it's an only picketbox test and that made fail it with elytron).

rmartinc:
I updated the test to use ldaps instead of plain ldap. This change has the weird effect to make it work in current version too, because the TCCL was only changed previously if a SSLContext was specified. But I think it is more complete this way. And now the execution path is the same for both (ldap o ldaps).

rmartinc:
I think everything is OK, the fail in windows jdk-11 is not related to the test (a spurious one). And it fails in jdk8 (normal and secmgr) as expected.

bstansberry:
@rmartinc There are LDAP related test failures in the linked CI jobs so I assume those are related.

Please reach out to folks who work on the security component to get a code review. Thanks!

rmartinc:
@bstansberry The errors made sense because the upgrade in core was needed. I see that now it's in place so I have just rebased to current master and re-launch the tests. 

@darranl could you please take a look to this PR when you have time? (It's a legacy ejb/ldap-realm setup that re-uses some ldap server classes created for other tests.)

rmartinc:
@darranl Please review this PR or assign it to someone in security. It's needed for 7.2.CP3 and we are in a hurry. I'll do any required change asap.

darranl (approved):
Approved

The one failing test was a known issue.

PR #13253, opened 2020-04-30
[WFLY-13420] Enable Elytron mixed domain test suites

yersan:
Unignore Elytron tests and fix the elytron subsystem configuration used in the DC.

Jira issue: https://issues.redhat.com/browse/WFLY-13420

PR #13257, opened 2020-04-30
[WFLY-13298] Set up mixed domain testing against EAP 7.3

yersan:
supersedes https://github.com/wildfly/wildfly/pull/13253

Adds mixed domain tests against EAP 7.3.0

Jira issue: https://issues.redhat.com/browse/WFLY-13298

### DIFF
--- a/testsuite/mixed-domain/src/test/java/org/jboss/as/test/integration/domain/mixed/eap730/DomainAdjuster730.java
+++ b/testsuite/mixed-domain/src/test/java/org/jboss/as/test/integration/domain/mixed/eap730/DomainAdjuster730.java
@@ -24,6 +24,7 @@ package org.jboss.as.test.integration.domain.mixed.eap730;
 
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXTENSION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.operations.common.Util.createEmptyOperation;
 import static org.jboss.as.controller.operations.common.Util.createRemoveOperation;
 
 import java.util.ArrayList;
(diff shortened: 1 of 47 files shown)


testsuite/mixed-domain/src/test/java/org/jboss/as/test/integration/domain/mixed/eap730/DomainAdjuster730.java
```diff
@@ -43,16 +46,36 @@
...
+    private void adjustUndertow(final List<ModelNode> ops, final PathAddress subsystem) {
+        // EAP 7.0 and earlier required explicit SSL configuration. Wildfly 10.1 added support
+        // for SSL by default, which automatically generates certs.
+        // This could be removed if all hosts were configured to contain a security domain with SSL
+        // enabled.
+        final PathAddress httpsListener = subsystem
+                .append("server", "default-server")
+                .append("https-listener", "https");
+        ops.add(Util.getEmptyOperation(ModelDescriptionConstants.REMOVE, httpsListener.toModelNode()));
+    }
```
bstansberry, line 72: I don't understand this one. This resource exists in domain.xml for all versions from EAP 7.1 onward. Which is consistent with the comment above. So I'd think this method would be part of DomainAdjuster700.

Currently, it's here, it's also in DomainAdjuster720 (which means the removal gets run twice) and it's also left over as an unused method in DomainAdjuster710.

yersan: I only executed the 7.3.0 tests in local so I did not realise that this adjustment must be moved from previous DomainAdjusters to the new ones. I have removed it now from DomainAdjuster710 and DomainAdjuster720 and added it only on DomainAdjuster730 to avoid the duplicity and the 'resource not exist' error.

We need this adjustment because the host slave configuration file used in the mixed domain tests does not add the SSL identity for the ApplicationRealm (to make the configuration compatible across all previous server releases including 7.0.0 and below), so we need to remove the undertow https to be able to deploy an application on the servers managed by the slave which does not have an SSL context available.

bstansberry: Thanks!

bstansberry (approved):

PR #13262, opened 2020-05-04
[WFLY-13424]: External pooled connection factory won't be properl injected if it has several JNDI entries.

ehsavoie:
* if the bound name is not the first one then we get a reference to a
  reference service instead ofg the pcf service.
* resolving this so that we get the proper resource adapter service name.

Jira: https://issues.redhat.com/browse/WFLY-13424

### DIFF
--- a/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/deployment/JMSConnectionFactoryDefinitionInjectionSource.java
+++ b/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/deployment/JMSConnectionFactoryDefinitionInjectionSource.java
@@ -49,6 +49,7 @@ import java.util.Set;
 import javax.resource.spi.TransactionSupport;
 import org.apache.activemq.artemis.api.core.DiscoveryGroupConfiguration;
 import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.ra.ActiveMQRAConnectionFactoryImpl;
 
 import org.jboss.as.connector.deployers.ra.ConnectionFactoryDefinitionInjectionSource;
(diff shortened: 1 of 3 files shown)


bstansberry:
This could use a test.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/deployment/JMSConnectionFactoryDefinitionInjectionSource.java
```diff
@@ -420,6 +423,21 @@ static String getDefaulResourceAdapter(DeploymentUnit deploymentUnit) {
...
+                    ManagedReference ref = ((ContextListAndJndiViewManagedReferenceFactory) pcf).getReference();
+                    Object ra = ref.getInstance();
+                    if (ra instanceof ActiveMQRAConnectionFactoryImpl) {
+                        bindInfo = ContextNames.bindInfoFor(((ActiveMQRAConnectionFactoryImpl) ra).getReference().getClassName());
+                        binder = deploymentUnit.getServiceRegistry().getService(bindInfo.getBinderServiceName());
+                        if (binder != null) {
+                            pcf = binder.getService().getValue();
+                            if (pcf != null && pcf instanceof ConnectionFactoryReferenceFactoryService) {
+                                return ((ConnectionFactoryReferenceFactoryService) pcf).getName();
+                            }
```
bstansberry, line 437: Minor point; ignore if you want -- this L435-437 block duplicates L423-425 and could be dropped if this entire new code block from L426 was moved above L423.

ehsavoie:
Adding a test is difficult as it requires an external broker otherwise it will default to the internal one.

bstansberry (approved):

bstansberry:
retest this please

PR #13264, opened 2020-05-04
Pull request 12903

bstansberry:
Follow on to #12903 that includes the related WFCORE-4653 Picketbox component upgrade.


bmaxwell:
There are 2 tests failing which are in EJBContextMultipleSDTestCase that was added to test the fix JBEAP-17862, both errors are basically the test failed authentication.

bstansberry:
@bmaxwell I commented on that in the main PR:

https://github.com/wildfly/wildfly/pull/12903#issuecomment-623774542

Something like this needs to be done:

https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/container/interceptor/security/SwitchIdentityTestCase.java#L96

The test is for a fix of legacy security behavior and is internally asserting various Picketbox details so there's no point trying to run it with a pure elytron security configuration.

darranl:
Superseded by https://github.com/wildfly/wildfly/pull/13271

PR #13267, opened 2020-05-04
WFLY-10942 IIOPTimeoutTestCase fails intermittently

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-10942

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
@@ -40,8 +40,8 @@ import javax.ejb.Singleton;
 @Remote
 @ManagedBean
 public class TransactionCheckerSingleton implements TransactionCheckerSingletonRemote {
-    private int committed, prepared, rolledback;
-    private int  synchronizedBegin, synchronizedBefore, synchronizedAfter,
+    private volatile int committed, prepared, rolledback;
+    private volatile int synchronizedBegin, synchronizedBefore, synchronizedAfter,
         synchronizedAfterCommitted, synchronizedAfterRolledBack;
     private Collection<String> messages = new ArrayList<>();
 


testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
```diff
@@ -40,8 +40,8 @@
 @Remote
 @ManagedBean
 public class TransactionCheckerSingleton implements TransactionCheckerSingletonRemote {
-    private int committed, prepared, rolledback;
-    private int  synchronizedBegin, synchronizedBefore, synchronizedAfter,
+    private volatile int committed, prepared, rolledback;
+    private volatile int synchronizedBegin, synchronizedBefore, synchronizedAfter,
```
pferraro, line 44: While making these volatile might improve the situation, it will not solve it as the ++ unary operator is not atomic.  I would suggest changing these to use AtomicInteger.

testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
```diff
@@ -40,8 +40,8 @@
...
         synchronizedAfterCommitted, synchronizedAfterRolledBack;
     private Collection<String> messages = new ArrayList<>();
```
pferraro, line 46: I suspect this might need to use a concurrent collection (or at least synchronized wrapper).

pferraro (changes requested):

testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
chengfang: Since this is a singleton ejb with container-managed concurrency, all method access are serialized and no two methods will be accessing the instance state. So I think make them volatile should work. If you agree, we can add volatile to `messages` field as well.

testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
chengfang: See above comment.

testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TransactionCheckerSingleton.java
pferraro: OK - in that case, volatile is sufficient.  `messages` should not need to be volatile since it is effectively final.

pferraro (approved):

bstansberry:
In the Windows test job that got cancelled this test ran and passed.

PR #13279, opened 2020-05-11
[WFLY-13446] Dependency of wildfly-weld-common would be better to be ${project.groupId}

gaol:
Issue: https://issues.redhat.com/browse/WFLY-13446

Use `${project.groupId}` instead of `org.wildfly` for dependency in the same project.


yersan (approved):
Thank you @gaol !

PR #13280, opened 2020-05-11
WFLY-13447 Fix the comparator implementation for sorting undertow filter-ref

msfm:
JIRA: https://issues.redhat.com/browse/WFLY-13447

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/LocationService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/LocationService.java
@@ -90,7 +90,7 @@ public class LocationService implements Service<LocationService>, FilterLocation
     }
 
     protected static HttpHandler configureHandlerChain(HttpHandler rootHandler, List<UndertowFilter> filters) {
-        filters.sort((o1, o2) -> o1.getPriority() >= o2.getPriority() ? 1 : -1);
+        filters.sort((o1, o2) -> Integer.compare(o1.getPriority(), o2.getPriority()));
         Collections.reverse(filters); //handler chain goes last first
         HttpHandler handler = rootHandler;
         for (UndertowFilter filter : filters) {


bstansberry:
@fl4via Please review

bstansberry:
I believe this will alter the order of filters for the common case where the default 'priority' is used. I expect the initial sort of the filters list is based on the order of the configuration elements, and the current sort will result in the one that appears first in the config executing before any other filter with the same priority. This change will result in it executing last.

I suspect to retain the current semantics some kind of wrapper object would be needed that expands the range of priorities such that none are equal...

bstansberry:
Something like this would preserve the existing semantics while still using a valid Comparator:

https://github.com/bstansberry/wildfly/commit/b7db21ecfdb267ce3eb399818c6598d861248075

msfm:
@bstansberry Thank you for the review!

----

https://docs.wildfly.org/19.1/wildscribe/subsystem/undertow/server/host/filter-ref/index.html says:

> priority: Defines filter order. A lower number instructs the server to be included earlier in the handler chain than others with higher numbers. Values range from 1, indicating the filter will be handled first, to 2147483647, resulting in the filter being handled last.

It does not define the order for the same priority. If the current sort will result in the filter that appears first in the config executing before any other filter with the same priority, it's just an undocumented implementation detail. 
I think users should not rely on it as such undocumented behavior may change without notice. If a user would like to control the order, they need to specify the priority explicitly.

----

However, as far as I tested, it seems my suggested change works fine. 

I've double-checked results between the current (original) implementation, my proposed change, and your suggested change in a java program like https://gist.github.com/msfm/ae3d2c27e8437f7282624c9813c568d4

```
$ javac ArrayListSortTest.java
$ java ArrayListSortTest <random-priority-range> <list-size>
```

For example:
-  `java ArrayListSortTest 10 31` -  the result from my suggested change is the same as the results from the original one and yours.
-  `java ArrayListSortTest 20 300` -  the result from my suggested change is the same as the results from yours. (Note that the original implementation can throw "IllegalArgumentException: Comparison method violates its general contract!" or the ordering result can be wrong if the list size is larger than 32.)

Please point out if I'm misunderstanding or overlooking anything.

----

By the way, I've found an issue inside your suggested change. The priority value from FilterRefWrapper can overflow `Integer.MAX_VALUE` if a large priority value is specified and many filter-ref are defined. (For example, when a user specifies `2147483647` to the priority in order to run the filter at the end.) This can result in an unexpected ordering.

```
    private static class FilterRefWrapper implements UndertowFilter {

        ...

        @Override
        public int getPriority() {
            return (wrapped.getPriority() * spacer) + ordinal;
        }
```

bstansberry:
@msfm Very nice! 

Good point about the problem with the large priority value. I thought about overrunning Integer.MAX_VALUE but was silly and just focused on the size of the list, which wouldn't realistically be huge. But the configured priority could easily be Integer.MAX_VALUE.

And in any case, as you say that approach produces the same results as yours, but with lots more lines of code. :)

If you add System.setProperty("java.util.Arrays.useLegacyMergeSort", "true"); to your ArrayListSortTest then you see the difference I was concerned about in ordering between originalFix and proposedFix1 / proposedFix2.  But TimSort has been the default since JDK 7 so having the execution order match the configuration element order has not been the default WF behavior for years, if it ever was. So as you say counting on that behavior is counting on something undocumented.

ropalka (approved):
LGTM

bstansberry:
Thanks, @msfm !

PR #13285, opened 2020-05-13
[WFLY-13386] Hung process instances and associated server.log WARN "Failed to reinstate timer 'kie-server.kie-server.EJBTimerScheduler' "

elguardian:
If activate the refresh task in the database persistence there are at least
two threads compiting for registering a timer.

The refresh task is executing at the same time a new timer is being registered
if this happens the same timer will be try to registered by ejb and refresh task

this is due the fact the ejb action register in the database in its own tx (outside ejb tx)
the refresh task tries to load from its own tx

so if this order happens
1 ejb thread register the timer in the database
2 refresh task thread  load the timer from the dabase
3 ejb thread registers the timer in the dmr
4 refresh task thread register the timer in the dmr (FAIL as duplicated)

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
@@ -117,7 +117,7 @@ public class TimerServiceImpl implements TimerService, Service<TimerService> {
     /**
      * All timers which were created by this {@link TimerService}
      */
-    private final Map<String, TimerImpl> timers = Collections.synchronizedMap(new HashMap<String, TimerImpl>());
+    private final Map<String, TimerImpl> timers = new HashMap<String, TimerImpl>();
 
     /**
      * Holds the {@link java.util.concurrent.Future} of each of the timer tasks that have been scheduled
@@ -543,7 +543,9 @@ public class TimerServiceImpl implements TimerService, Service<TimerService> {
     }
 
     public TimerImpl getTimer(final String timerId) {
-        return timers.get(timerId);
+        synchronized (this.timers) {
+            return this.timers.get(timerId);
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


elguardian:
@chengfang do you mind to have a look. This one fix the nasty issue of the refresh task and duplicated resources. Also fixes the problem of null point while sync.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
```diff
@@ -1133,12 +1128,23 @@ public TimerServiceResource getResource() {
         return resource;
     }
 
-    private void registerTimerResource(final String timerId) {
-        this.resource.timerCreated(timerId);
+    private boolean registerTimerResource(final TimerImpl timer) {
+        synchronized (this.timers) {
+            if (this.timers.containsKey(timer.getId())) {
```
chengfang, line 1133: `timers` is already declared as a synchronized map. We may want to switch to a regular `HashMap` and guard all access to it to avoid double synchronize it.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
```diff
@@ -1133,12 +1128,23 @@ public TimerServiceResource getResource() {
...
+                this.timers.put(timer.getId(), timer);
```
chengfang, line 1134: If the timer id already exists in the internal cache `timers`, why do we put the timer to `timers` again? It looks harmless though.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
elguardian: done

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
elguardian: ok. removed

chengfang (approved):

PR #13298, opened 2020-05-18
[WFLY-13256] Upgrade BouncyCastle from 1.62 to 1.65

bstansberry:
https://issues.redhat.com/browse/WFLY-13256

It's a known issue that this upgrade results in some odd behavior with Picketlink STS tests on JDK 13, where running certain tests before others causes those others to fail. As picketlink use is deprecated and our support for JDK 13 is meant for developer evaluation, myself, Darran, Alessio, Jim Ma and Ashley discussed this and I believe it's more important to get this in than to worry about the PicketLink + JDK 13 scenario.

Custom JDK 13 test run against this branch is at https://ci.wildfly.org/viewQueued.html?itemId=205928


jamezp:
This looks like it produces some issues with the WS tests.

jamezp (approved):

PR #13300, opened 2020-05-18
WFLY-13357 Execution of concurrent batch jobs containg partitioned st…

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-13357

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/thread/pool/WildFlyJobExecutor.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/thread/pool/WildFlyJobExecutor.java
@@ -23,7 +23,9 @@
 package org.wildfly.extension.batch.jberet.thread.pool;
 
 import org.jberet.spi.JobExecutor;
+import org.jberet.spi.JobTask;
 import org.jboss.as.threads.ManagedJBossThreadPoolExecutorService;
+import org.wildfly.extension.requestcontroller.ControlPointTask;
 
 /**
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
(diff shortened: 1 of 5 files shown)


jamezp:
The test failure seems related.

chengfang:
Yes, it seems like a test bug.

`org.jboss.as.test.integration.batch.transaction.BatchTransactionTimeoutTestCase#testThreadIsAvailableForNextJob` currently sets a max thread count to 1, while the updated job executor requries at least 2.

chengfang:
The test bug `BatchTransactionTimeoutTestCase#testThreadIsAvailableForNextJob` has been fixed by increasing the `max-threads` from 1 to 3, and changing the batch job step to be a 2-partition chunk step.

chengfang:
@jamezp can you review it?

jamezp (commented):
@chengfang Does this mean the thread pools `max-threads` always has to be 2 or higher?

chengfang:
@jamezp Yes, See https://github.com/jberet/jsr352/blob/master/jberet-core/src/main/java/org/jberet/runtime/runner/JobExecutionRunner.java#L144 .I think this is a reasonable requriement for the users.

jamezp:
@chengfang This is a change in behavior then that we likely want to look at changing on the resource. It shouldn't be allowed to set to anything less than 2 in the model. We also need some documentation to some extent about this.

chengfang:
This is already documented via attribute description:
https://wildscribe.github.io/WildFly/18.0/subsystem/batch-jberet/thread-pool/index.html

"The thread pool used for batch jobs. Note that the max-thread attribute should always be greater than 3. Two threads are reserved to ensure partition jobs can execute."

But this min value of `max-thread` is not currently enforced in CLI.

jamezp (approved):

PR #13311, opened 2020-05-20
[WFLY-13495] Add missing dependencies for last json-patch update

jbliznak:
https://issues.redhat.com/browse/WFLY-13495

This is a follow up for https://github.com/wildfly/wildfly/pull/12818

By updating to json-patch 1.9 we also need other dependencies to be added which was missed in the original PR. Details are explained at https://issues.redhat.com/browse/RESTEASY-2126

I checked that these changes fix the problem reported at jira, what I am not sure about is whether I layouted the modules changes correctly so I kindly ask someone to confirm that.
Open questions:
* Do we need exclusions on maven dependencies? I can see that current changes generate unexpected license record for com.google.code.findbugs:jsr305 
* Does the order on ee-galleon-pack/pom.xml matter? 

The alternative simplified version without additional modules (as we just need these new dependencies on "classpath" and don't use them in other modules) is proposed at https://github.com/wildfly/wildfly/compare/master...jbliznak:wfly-13495-fix-fge-modules-flatten (this is basically how it is tested on RESTEasy now).

### DIFF
--- a/ee-galleon-pack/pom.xml
+++ b/ee-galleon-pack/pom.xml
@@ -287,14 +287,44 @@
 
         <dependency>
             <groupId>com.github.fge</groupId>
-            <artifactId>json-patch</artifactId>
+            <artifactId>btf</artifactId>
             <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>com.github.fge</groupId>
             <artifactId>jackson-coreutils</artifactId>
             <scope>provided</scope>
+            <exclusions>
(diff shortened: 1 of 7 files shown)


jbliznak:
OK, it looks like had `-Denforcer.skip` so that answer the question for needed exclusions which popped up when enforcer is enabled, fixing now.

bstansberry:
@asoldano @ronsigal @jbliznak Let's discuss in zulip

bstansberry:
@jbliznak re ordering, in the ee-galleon pack and feature-pack files please put the new items in the right alphabetical order by groupid / artifactid. That's a hard requirement for the licenses.xml file. For the others it's just being tidy.

jbliznak:
@bstansberry reordered, thanks

jamezp:
This required a rebase and is replaced by #13312

PR #13312, opened 2020-05-20
[WFLY-13495] Add missing dependencies for last json-patch update

darranl:
Replacement PR for https://github.com/wildfly/wildfly/pull/13311

### DIFF
--- a/ee-galleon-pack/pom.xml
+++ b/ee-galleon-pack/pom.xml
@@ -287,14 +287,44 @@
 
         <dependency>
             <groupId>com.github.fge</groupId>
-            <artifactId>json-patch</artifactId>
+            <artifactId>btf</artifactId>
             <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>com.github.fge</groupId>
             <artifactId>jackson-coreutils</artifactId>
             <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.fge</groupId>
+            <artifactId>json-patch</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.fge</groupId>
+            <artifactId>msg-simple</artifactId>
+            <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
--- a/feature-pack/pom.xml
+++ b/feature-pack/pom.xml
@@ -220,12 +220,40 @@
 
         <dependency>
             <groupId>com.github.fge</groupId>
-            <artifactId>json-patch</artifactId>
+            <artifactId>btf</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>com.github.fge</groupId>
             <artifactId>jackson-coreutils</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.fge</groupId>
+            <artifactId>json-patch</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.fge</groupId>
+            <artifactId>msg-simple</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
--- a/feature-pack/src/license/full-feature-pack-licenses.xml
+++ b/feature-pack/src/license/full-feature-pack-licenses.xml
@@ -124,6 +124,17 @@
     </dependency>
     <dependency>
       <groupId>com.github.fge</groupId>
+      <artifactId>btf</artifactId>
+      <licenses>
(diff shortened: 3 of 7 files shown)


asoldano (approved):
I've tested this and reviewed this and I think it's fine.

ronsigal:
Here's one thing I can see so far. I took the org.wildfly:wildfly-parent pom.xml, stripped out all of the modules, and ran mvn dependency:tree four times, each with exaclty one dependency:
* com.github.fge:btf
* com.github.fge:msg-simple
* com.github.fge:jackson-coreutils
* com.github.fge:json-patch

I got four trees, each was a subtree of the next, and the final tree was
```
[INFO] org.wildfly:wildfly-parent:pom:20.0.0.Beta1-SNAPSHOT
[INFO] \- com.github.fge:json-patch:jar:1.9:compile
[INFO]    \- com.github.fge:jackson-coreutils:jar:1.8:compile
[INFO]       +- com.github.fge:msg-simple:jar:1.1:compile
[INFO]       |  \- com.github.fge:btf:jar:1.2:compile
[INFO]       +- com.fasterxml.jackson.core:jackson-databind:jar:2.10.4:compile
[INFO]       |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.10.4:compile
[INFO]       |  \- com.fasterxml.jackson.core:jackson-core:jar:2.10.4:compile
[INFO]       +- com.google.guava:guava:jar:25.0-jre:compile
[INFO]       \- com.google.code.findbugs:jsr305:jar:3.0.2:test
```

ronsigal:
When I look at module.xml files

* com.github.fge.btf: no dependencies
* com.github.fge.msg-simple: depends on and exports com.github.fge.btf
* com.github.fge.jackson-coreutils: depends on and exports com.github.fge.msg-simple
* com.github.fge.json-patch: depends on but doesn't export com.github.fge.jackson-coreutils
* org.jboss.resteasy.resteasy-jackson2-provider: depends on but doesn't export com.github.fge.json-patch

ronsigal:
Also,  com.github.fge:btf, com.github.fge:msg-simple, com.github.fge:jackson-coreutils, and  com.github.fge:json-patch all have
```
    <properties>
        <property name="jboss.api" value="private"/>
    </properties>
```

ronsigal:
So, if I understand the wildfly module structure (which I don't, really), it looks like everything is hidden behind org.jboss.resteasy.resteasy-jackson2-provider.

jamezp:
> So, if I understand the wildfly module structure (which I don't, really), it looks like everything is hidden behind org.jboss.resteasy.resteasy-jackson2-provider.

That is correct and the best way to handle it. If they are not private we have to support them. For example say a new version of RESTEasy decides it doesn't want to use one of those API's it can be removed if it's private.

bstansberry:
This question shouldn't block merging this, but I'm wondering, @jbliznak, why the 'export="true"' stuff is needed. If X exports Y that implies the consumers of X need classes or resources from Y, but the maven dependency info from @ronsigal doesn't show an obvious need for that here.

jbliznak:
@bstansberry the reason I added the export=true is simply it doesn't work for me without it.

eg. with this the original problem reported in jira is still there
```
diff --git a/feature-pack/src/main/resources/modules/system/layers/base/com/github/fge/jackson-coreutils/main/module.xml b/feature-pack/src/main/resources/modules/system/layers/base/com/github/fge/jackson-coreutils/main/module.xml
index 292abb72af..9cdf274551 100755
--- a/feature-pack/src/main/resources/modules/system/layers/base/com/github/fge/jackson-coreutils/main/module.xml
+++ b/feature-pack/src/main/resources/modules/system/layers/base/com/github/fge/jackson-coreutils/main/module.xml
@@ -33,7 +33,7 @@
     </resources>
 
     <dependencies>
-        <module name="com.github.fge.msg-simple" export="true"/>
+        <module name="com.github.fge.msg-simple"/>
         <module name="com.fasterxml.jackson.core.jackson-annotations"/>
         <module name="com.fasterxml.jackson.core.jackson-core"/>
         <module name="com.fasterxml.jackson.core.jackson-databind"/>
```

ronsigal:
re: "why the 'export="true"' stuff is needed."

I was wondering about that, but I suppose if A needs a class from B and a class from C, then either

* A could import B and C, or
* A could import B and B could export C.

@jbliznak, what happens if you replace
```
<module name="com.github.fge.msg-simple" export="true"/>
```
with 
```
<module name="com.github.fge.msg-simple"/>
```
in jackson-coreutils and add
```
<module name="com.github.fge.msg-simple"/>
```
to json-patch?

ronsigal:
And, similarly, replace
```
 <module name="com.github.fge.btf" export="true"/>
```
with
```
 <module name="com.github.fge.btf"/>
```
in msg-simple, and add
```
<module name="com.github.fge.btf"/>
```
to jackson-coreutils?

ronsigal:
I guess my question is, does
```
[INFO]    \- com.github.fge:jackson-coreutils:jar:1.8:compile
[INFO]       +- com.github.fge:msg-simple:jar:1.1:compile
[INFO]       |  \- com.github.fge:btf:jar:1.2:compile
```
imply that jackson-coreutils **doesn't** depend directly on btf, i.e., doesn't explicitly use a class from btf?

jbliznak:
OK so the export is needed because `json-patch` directly accesses the `msg-simple` classes
https://github.com/java-json-tools/json-patch/blob/1.9/src/main/java/com/github/fge/jsonpatch/JsonPatchMessages.java#L22-L24

Furthermore, `jackson-coreutils` directly accesses `btf` https://github.com/fge/jackson-coreutils/blob/1.8/src/main/java/com/github/fge/jackson/JsonNodeReader.java#L30

@ronsigal hope it clarifies.

ronsigal:
re: "hope it clarifies."

Yes, that's exactly what I was wondering. I think our last posts crossed in the mail.

jbliznak:
Yes, these artifacts have either wrong encapsulation or wrong pom.xml which should state direct dependencies (so that we can easily design our modules tree)

ronsigal:
So msg-simple has
```
    <dependency>
      <groupId>com.github.fge</groupId>
      <artifactId>btf</artifactId>
      <version>1.2</version>
      <scope>compile</scope>
    </dependency>
```
and jackson-coreutils has
```
    <dependency>
      <groupId>com.github.fge</groupId>
      <artifactId>msg-simple</artifactId>
      <version>1.1</version>
      <scope>compile</scope>
    </dependency>
```
and compile scope is transitive, so, **in effect**, msg-simple is exporting btf.

bstansberry:
Ah, that's a good point, mvn dependency:tree isn't going to show that kind of thing.

PR #13316, opened 2020-05-21
[WFLY-13423] Remove the uses of AbstractOutboundConnectionService fro…

yersan:
…m ejb3 subsystem

It requires a core upgrade. Depends on: https://github.com/wildfly/wildfly-core/pull/4213

- Provisions `org.jboss.as.remoting` always when EJB3RemoteResourceDefinition is present.
- Replaces the uses of `org.jboss.as.remoting.AbstractOutboundConnectionService` in favor of `org.jboss.as.network.OutboundConnection` and get it from the capability

Jira issue: https://issues.redhat.com/browse/WFLY-13423

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EJBClientDescriptorMetaDataProcessor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EJBClientDescriptorMetaDataProcessor.java
@@ -40,6 +40,7 @@ import java.util.Properties;
 
 import javax.security.auth.callback.CallbackHandler;
 
+import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.domain.management.CallbackHandlerFactory;
 import org.jboss.as.domain.management.SecurityRealm;
 import org.jboss.as.ee.metadata.EJBClientDescriptorMetaData;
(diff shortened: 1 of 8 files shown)


yersan:
retest this please

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EJBClientDescriptorMetaDataProcessor.java
```diff
@@ -133,6 +136,19 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
+                for (EJBClientDescriptorMetaData.RemotingReceiverConfiguration receiverConfiguration : ejbClientDescriptorMetaData.getRemotingReceiverConfigurations()) {
+                    final String requiredCap = RuntimeCapability.buildDynamicCapabilityName(OUTBOUND_CONNECTION_CAPABILITY_NAME, receiverConfiguration.getOutboundConnectionRef());
+                    if (!support.hasCapability(requiredCap)) {
+                        throw EjbLogger.DEPLOYMENT_LOGGER.deploymentRequiresCapability(
+                                deploymentUnit.getName(),
+                                requiredCap);
+                    }
+            }
+        }
+
```
bstansberry, line 151: Is this validation necessary?  This isn't a management model consistency check, where we need to fail in Stage.MODEL if something is wrong. This is runtime stuff. If the reference is invalid won't MSC report the service wiring failure? I'm reluctant to start adding additional validation. It has the potential to produce nicer failures (although a "missing capability" message isn't particularly nicer than a "missing service" message) but it's either going to end up being done in a few random spots or it will be a very large addition of potentially fragile code.

yersan: It is unnecessary, the intention was fail as soon as possible with a nicer message.
I agree on your points so I'll remove this and let the missing service message in case a user runs into this situation.

bstansberry (approved):

PR #13321, opened 2020-05-22
[WFLY-13508] Fix layer name for naming, it should be 'naming' instead …

yersan:
…of 'ee'

Jira issue: https://issues.redhat.com/browse/WFLY-13508


kabir (approved):

jfdenise (approved):
Good catch.

rhusar:
> This is not a problem at all since currently layers names are discovered by the directory name it resides, however, the wrong name here could lead to confusion.

Hm, so these aren't used? What was the intention for them then?

yersan:
I asked about the intentionality on Zulip as well, see it on Galleon stream, not sure about the reason or if it is a bug.

bstansberry:
Thanks @yersan.

PR #13323, opened 2020-05-25
[WFLY-13249] WebSecurityCERTTestCase fails on IBM JDK when elytron is enabled

SoniaZaldana:
https://issues.redhat.com/browse/WFLY-13249


testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/security/cert/WebSecurityCERTTestCase.java
```diff
@@ -70,6 +72,16 @@
 @Category(CommonCriteria.class)
 public class WebSecurityCERTTestCase {
 
+    @BeforeClass
+    public static void setup() throws Exception {
+        System.setProperty("com.ibm.jsse2.overrideDefaultTLS", "true");
```
fjuma, line 77: Just a small comment, `System.setProperty` returns the previous value of the property or `null` if there wasn't a previous value. This can be saved in a static variable. Then in the tear down method, if this variable is non-null, `setProperty` can be used again to set it back to its original value. Otherwise, `clearProperty` can be used.

fjuma: It would also be good to only set this property if IBM JDK is in use.

PR #13324, opened 2020-05-26
[WFLY-13516] Adding integration test for recovery for ejb remoting context propagation and JVM crash

ochaloup:
https://issues.redhat.com/browse/WFLY-13516

This is an integration test for issue found at https://issues.redhat.com/browse/JBEAP-19408

/CC @bstansberry @mmusgrov @tomjenkinson


bstansberry:
I put the 20.x label on this just indicate it's ok for 20.0.0.Final if it gets approved via normal code review. It's just tests so it's safe to add late in the cycle. It's not a blocker for 20 or anything like that though.

PR #13328, opened 2020-05-27
WFLY-11767 Remove unused dependencies from org.jboss.as.connector

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-11767


bstansberry:
@maeste Any concerns?

The org.jboss.jts module provides org.jboss.narayana.jts:narayana-jts-idlj and none of the packages in that artifact are used by org.wildfly:wildfly-connector. The org.jboss.narayana.jts:narayana-jts-idlj artifact's META-INF/services has an entry for the CDI Extension interface, but that's not something org.wildfly:wildfly-connector would be loading. So this looks unused to me.

maeste:
@bstansberry LGTM

bstansberry:
Thanks @RanabirChakraborty.

RanabirChakraborty:
Thank you! @bstansberry I have marked it as Resolved.

PR #13329, opened 2020-05-27
WFLY-13525 Upgrade Eclipse MicroProfile Fault Tolerance TCK to 2.1.1-RC4

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-13525

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -347,7 +347,8 @@
         <version.org.cryptacular>1.2.4</version.org.cryptacular>
         <version.org.eclipse.jdt.core.compiler>4.6.1</version.org.eclipse.jdt.core.compiler>
         <version.org.eclipse.microprofile.config.api>1.4</version.org.eclipse.microprofile.config.api>
-        <version.org.eclipse.microprofile.fault-tolerance.api>2.1.1-RC1</version.org.eclipse.microprofile.fault-tolerance.api>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -347,7 +347,8 @@
         <version.org.cryptacular>1.2.4</version.org.cryptacular>
         <version.org.eclipse.jdt.core.compiler>4.6.1</version.org.eclipse.jdt.core.compiler>
         <version.org.eclipse.microprofile.config.api>1.4</version.org.eclipse.microprofile.config.api>
-        <version.org.eclipse.microprofile.fault-tolerance.api>2.1.1-RC1</version.org.eclipse.microprofile.fault-tolerance.api>
+        <version.org.eclipse.microprofile.fault-tolerance.api>2.1</version.org.eclipse.microprofile.fault-tolerance.api>
```
kabir, line 350: @rhusar I guess this has the same contents as the 2.1.1-RC1 version?

rhusar: @kabir Yes, the API is the same, there are some javadoc improvements in later versions, but I don't want to risk MP FT doesn't release a final before we do and to cause confusion. So splitting the API and TCK modules – which in all honesty should be versioned separately but oh well – gives us flexibility.

pferraro (approved):
I think decoupling the spec API version and the TCK version to pull in TCK fixes from a micro is sensible.

PR #13334, opened 2020-05-29
WFLY-13538 Port Java 9+ version of ConcurrentDirectDeque from Undertow

pferraro:
https://issues.redhat.com/browse/WFLY-13538

Ports the Java 9+ variant of FastConcurrentDirectDeque from Undertow and uses a multi-release jar for the org.wildfly:wildfly-clustering-ee-cache module.

### DIFF
--- a/clustering/ee/cache/build-include-jdk-misc
+++ b/clustering/ee/cache/build-include-jdk-misc
@@ -1,0 +1,1 @@
+https://github.com/jboss/jboss-parent-pom#mr-jar-sun-misc
--- a/clustering/ee/cache/build-test-java10
+++ b/clustering/ee/cache/build-test-java10
@@ -1,0 +1,1 @@
+https://github.com/jboss/jboss-parent-pom#mr-jar-testing
--- a/clustering/ee/cache/build-test-java11
+++ b/clustering/ee/cache/build-test-java11
@@ -1,0 +1,1 @@
+https://github.com/jboss/jboss-parent-pom#mr-jar-testing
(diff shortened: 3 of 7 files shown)


bstansberry:
@dosoudil FYI and comment.

@pferraro This should be discussed on the dev list. When we release WildFly we build using JDK 8. That's a pretty substantial thing to change for one class.

darranl:
I mentioned on the dev list thread as well but wouldn't this be better in wildfly-common where a multi-release jar is already possible then WildFly can use it without a change to the WildFly build.

pferraro:
To reiterate what was said on the dev list thread, I already pursued the option of housing these collections in wildfly-common, and it was decided to keep wildfly-common free of collection implementations.

scottmarlow:
We do have other collections included in WildFly already, probably more in components.

wildfly:
```
ag -rli doug.*lea
web-common/src/main/java/org/jboss/as/web/common/ConcurrentReferenceHashMap.java
naming/src/main/java/org/jboss/as/naming/util/FastCopyHashMap.java
security/subsystem/src/main/java/org/jboss/as/security/lru/PortableConcurrentDirectDeque.java
security/subsystem/src/main/java/org/jboss/as/security/lru/FastConcurrentDirectDeque.java
clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/PortableConcurrentDirectDeque.java
clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/FastConcurrentDirectDeque.java
```
wildfly-core:
```
ag -rli doug.*lea
controller/src/main/java/org/jboss/as/controller/registry/FastCopyHashMap.java
```

kabir:
Adding the hold label so noone accidentally merges this. It looks like it needs discussion

pferraro:
Removed hold label, since the discussion has come to a resolution.

PR #13338, opened 2020-06-02
[WFLY-13549] Apply class filtering to remote EJB and Naming invocations

bstansberry:
https://issues.redhat.com/browse/WFLY-13549

Also incorporates necessary upgrades:

https://issues.redhat.com/browse/WFLY-13522
https://issues.redhat.com/browse/WFLY-13551
https://issues.redhat.com/browse/WFLY-13552

@chengfang @tadamski @fl4via @ropalka @asoldano @jamezp FYI

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/iiop/EjbIIOPService.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/iiop/EjbIIOPService.java
@@ -23,6 +23,8 @@ package org.jboss.as.ejb3.iiop;
 
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InvalidClassException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -47,11 +49,13 @@ import org.jboss.ejb.client.StatelessEJBLocator;
 import org.jboss.ejb.iiop.EJBMetaDataImplIIOP;
 import org.jboss.ejb.iiop.HandleImplIIOP;
 import org.jboss.ejb.iiop.HomeHandleImplIIOP;
+import org.jboss.marshalling.ClassResolver;
 import org.jboss.marshalling.Marshaller;
 import org.jboss.marshalling.MarshallerFactory;
 import org.jboss.marshalling.MarshallingConfiguration;
 import org.jboss.marshalling.ModularClassResolver;
 import org.jboss.marshalling.OutputStreamByteOutput;
+import org.jboss.marshalling.Unmarshaller;
 import org.jboss.marshalling.river.RiverMarshallerFactory;
 import org.jboss.metadata.ejb.jboss.IIOPMetaData;
 import org.jboss.metadata.ejb.jboss.IORSecurityConfigMetaData;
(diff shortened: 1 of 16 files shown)


jamezp (approved):

bstansberry:
retest this please

bstansberry:
The LocalViewRemoteInvocationTestCasetestServerInVMInvocationOnLocalViewOfSFSB failures that will be reported are because a call that is expected to fail is indeed failing, but with an EJBException instead of the expected ISE. The message of the EJBException is along the expected lines (complains about local view for a remote bean) so it's perhaps ok, but I don't see why this PR would result in a different call path here. I'm debugging.

chengfang:
The EJBException seen in the failure was reported by ejb client, while the expected IllegalStateException is to be reported by [LocalEjbReceiver](https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/remote/LocalEjbReceiver.java#L104) (judgeing from [AS7-5103](https://issues.redhat.com/browse/AS7-5103)). So looks like the check that should've been performed by `LocalEjbReceiver` was not there and hence the error leaks to and was caught by client.

bstansberry:
@chengfang AIUI EJBClientInvocationContext resolves an EJBReceiver. It loops through the available ones and uses the first that supports the scheme of the destination URL. LocalTransportProvider should be the last one it checks as it's the last added to the list. It only accepts 'local' scheme.  But before it was being selected and now it seems RemoteTransportProvider is.  That implies the URL has a different scheme from before.  I'm not seeing how that happens yet, but it seems like it must be.

A guess I had is this had something to do with the naming changes (different object resolved leads to different state and code paths.). But eliminating the use of the filtering in naming has no effect on this.

bstansberry:
@chengfang Now I suspect this may have something to do with the wildfly-http-client upgrade so I filed #13442 to test that on its own.

The session creation call hits the LocalEjbReceiver but the method invocation goes to RemoteEjbReceiver.  I don't see how my changes could have any bearing on what URL is used for the invocation.

chengfang:
I checked out and rebuild locally with this pull, and ran this test, but it passed locally for me:

```
testsuite/integration/basic > 

mvn test -Dtest=org.jboss.as.test.integration.ejb.remote.view.LocalViewRemoteInvocationTestCase#testServerInVMInvocationOnLocalViewOfSFSB
```

I also tested it in a [sample](https://github.com/chengfang/hello-local-remote.git) with similar scenario (look up local sfsb with ejb: jndi name, and call it), deployed to the WF built above, got the expected `IllegalStateException`.

bstansberry:
@chengfang It's failing every time locally for me. (Another test in testsuite/int/basic always fails for me, some local env issue, and that prevented me realizing the surefire execution this one is in wasn't running.)

The reason it fails is with this PR RemotingEJBDiscoveryProvider is discovering the remote+http URL while with master it does not. (With both this PR and master the local discovery provider from AssociationService also discovers "local:-".) The failure happens because when the bean is invoked it uses the first URL discovered. With master the only one is "local:-" but with this PR it is the remote+http URL.

It's a pain figuring out why this is. I still have no clue how this PR could be the cause. (I've separately tested just the component upgrades, individually and as a set and they don't cause the problem.)

bstansberry:
@chengfang This should fix it:

https://github.com/wildfly/wildfly/pull/13338/files#diff-c8a90d2dcf5f441488f745f6716c16e5R39

The reason I couldn't figure out how adding the filter stuff could affect this is that it didn't. :) It was a new test altering the server's state and thus affecting the other.

chengfang:
> @chengfang This should fix it:
> 
> https://github.com/wildfly/wildfly/pull/13338/files#diff-c8a90d2dcf5f441488f745f6716c16e5R39
> 
> The reason I couldn't figure out how adding the filter stuff could affect this is that it didn't. :) It was a new test altering the server's state and thus affecting the other.

Interesting! I thought each test case starts its own server independent of each other. Anyway, great to sort it out.

PR #13340, opened 2020-06-02
[WFLY-13551] Upgrade wildfly-naming-client to 1.0.13.Final

bstansberry:
https://issues.redhat.com/browse/WFLY-13551

This incorporated in #13338 -- I file it separately here just to get CI running while 13338 waits for another component's release.


chengfang:
Linux with sec JDK8 had 1 failure, which is a known transient failure (EjbInvocationStatisticsTestCase fails [WFLY-10173](https://issues.redhat.com/browse/WFLY-10173)).

PR #13302, opened 2020-05-19
Fix for WFLY-13490, Galleon layer for web console

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-13490

New layer in ee-galleon-pack, updated docs and layers test.

PR #13343, opened 2020-06-03
[WFLY-13490] Add Galleon layer for web console

yersan:
This supersedes: https://github.com/wildfly/wildfly/pull/13302

It is the same PR but this one includes the integration tests available on WildFly tests suite for the Web Console.

One of those tests is ignored in the case of layers is enabled. The ignored test verifies header overriding for a non management endpoint, using the metrics endpoint. However when the web console is supplied via a Galleon Layer there is no other non management endpoint, so the test is ignored for layers provisioning.

Jira issue: https://issues.redhat.com/browse/WFLY-13490
Proposal: https://github.com/wildfly/wildfly-proposals/pull/312

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
@@ -165,6 +165,10 @@ Layers that you combine with "use-case tailored" layers to extend the capabiliti
 * _observability_: Support for MicroProfile monitoring and configuration features. 
 Includes Health support (optional), _microprofile-config_ (optional), _microprofile-metrics_ (optional) and _open-tracing_ (optional).
 
+Tools layers:
+
+* _web-console_: Support for HAL web console.
+
 ==== WildFly servlet layers
 
 Basic layers:
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-console/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-console/layer-spec.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="web-console">
+    <dependencies>
+        <layer name="secure-management"/>
+    </dependencies>
+    <packages>
+        <package name="org.jboss.as.console"/>
+    </packages>
+</layer-spec>
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1201,6 +1201,84 @@
                                 </configuration>
                             </execution>
 
+                            <!-- Provision a server with Web Console -->
+                            <execution>
+                                <id>web-console-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${project.build.directory}/wildfly-web-console</install-dir>
+                                    <record-state>false</record-state>
+                                    <log-time>${galleon.log.time}</log-time>
+                                    <offline>true</offline>
+                                    <plugin-options>
+                                        <jboss-maven-dist/>
+                                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                                        <optional-packages>passive+</optional-packages>
+                                    </plugin-options>
+                                    <feature-packs>
+                                        <feature-pack>
+                                            <groupId>${testsuite.ee.galleon.pack.groupId}</groupId>
+                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
+                                            <version>${testsuite.ee.galleon.pack.version}</version>
+                                            <inherit-configs>false</inherit-configs>
+                                            <inherit-packages>false</inherit-packages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <configurations>
+                                        <config>
+                                            <model>standalone</model>
+                                            <name>standalone.xml</name>
+                                            <layers>
+                                                <layer>web-console</layer>
+                                                <layer>datasources-web-server</layer>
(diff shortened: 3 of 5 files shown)


yersan:
Added also `org.jboss.as.patching` since it is necessary to supply the `/core-service=patching` generated at runtime.

yersan:
@bstansberry @hpehl would you mind taking a look and approve in case? I don't plan to apply code changes as it is now. Feature issue is in "Review" state, thanks!

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="web-console">
+    <dependencies>
+        <layer name="secure-management"/>
+        <layer name="core-management"/>
+    </dependencies>
+    <packages>
+        <package name="org.jboss.as.console"/>
+        <package name="org.jboss.as.patching"/>
```
bstansberry, line 9: @yersan Why this?

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="web-console">
+    <dependencies>
+        <layer name="secure-management"/>
+        <layer name="core-management"/>
```
bstansberry, line 5: Why this?

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
yersan: HAL is not prepared to show the "Pathching" menu entry if `/core-service=patching` resource does not exist.

It is a harmless error, we can live with it, the user would see an `"WFLYCTL0030: No resource definition is registered for address [(\"core-service\" => \"patching\")]"`, which indeed is fine. I don't know if we plan to use patching in the future for a bootable jar installation.

I just added it to avoid displaying this error, however, probably the users will receive a different one if they try to patch a bootable jar. The `org.jboss.as.patching` package has no impact on the disk footprint.

What do you prefer to do with this package?
Now I'm more inclined to remove it and live with the error.

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
bstansberry: I'm fine with keeping this but a HAL issue to allow it to gracefully live without this would be good.

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
bstansberry: @yersan IIRC with an RPM installation EAP doesn't install this (because RPM-based EAP is updated using RPM), so HAL should be able to handle its absence gracefully. I'd prefer that HAL corrects this.

yersan: https://issues.redhat.com/browse/HAL-1706

ee-galleon-pack/src/main/resources/layers/standalone/web-console/layer-spec.xml
yersan: https://issues.redhat.com/browse/HAL-1707

yersan:
retest this please

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
```diff
@@ -165,6 +165,10 @@ Layers that you combine with "use-case tailored" layers to extend the capabiliti
 * _observability_: Support for MicroProfile monitoring and configuration features. 
 Includes Health support (optional), _microprofile-config_ (optional), _microprofile-metrics_ (optional) and _open-tracing_ (optional).
 
+Tools layers:
+
+* _web-console_: Support for HAL web console.
```
bstansberry, line 170: @hpehl WDYT about wording more like this:

"Support for loading the HAL web console from the /console context on the HTTP management interface. Not required to use a HAL console obtained independently and configured to connect to the server."

Maybe too wordy. Or not wordy enough. :) I'm just wanting to get across that this isn't needed for HAL to work; it just gives you a copy of HAL.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-console/layer-spec.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="web-console">
+    <dependencies>
+        <layer name="secure-management"/>
```
yersan, line 4: Replace this by `management` once https://github.com/wildfly/wildfly-core/pull/4296 is merged

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
hpehl: @bstansberry 
Sounds good to me. I don't think it's too wordy.

tommaso-borgato:
@bstansberry tagging you in case you want to give a look

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-console/layer-spec.xml
bstansberry: Actually this can be done once https://github.com/wildfly/wildfly-core/pull/4295 is merged and available, and it is.

bstansberry:
Superseded by #13520, which just adjusts the two items discussed in comments here.

PR #13359, opened 2020-06-09
[WFLY-13576] Upgrade WildFly Galleon Plugins from 4.2.6.Final to 4.2.8.Final

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13576


bstansberry:
Thanks @yersan.

PR #13339, opened 2020-06-02
[WFLY-13522] Upgrade jboss-ejb-client to 4.0.33.Final

bstansberry:
https://issues.redhat.com/browse/WFLY-13522

PR #13289, opened 2020-05-14
WFLY-13469 Decouple distributed SFSB cache from client mappings registry

pferraro:
https://issues.redhat.com/browse/WFLY-13469

Pre-requisite of WFLY-13132

PR #13360, opened 2020-06-09
WFLY-13132 Wrong/Incomplete CLUSTER_TOPOLOGY update sent to EJB client

rachmatowicz:
... when clients connect to the same server/cluster using different protocols.
https://issues.redhat.com/browse/WFLY-13132

Includes subtasks:
Provide SPI and implementation for building a client mappings registry per connector
https://issues.redhat.com/browse/WFLY-13470

Requires:
#13289
#13339


pferraro:
We will defer to the topic branch.

PR #13361, opened 2020-06-09
[WFLY-13579] adjust testsuite waiting times with the timeout factor

ochaloup:
https://issues.redhat.com/browse/WFLY-13579

Making the `Thread.sleep` and other timeout calls in testsuite to be parametrized by timeout factor (defined in {{TimeoutUtil}} shared testsuite config class).

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/management/deployments/EjbInvocationStatisticsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/management/deployments/EjbInvocationStatisticsTestCase.java
@@ -48,6 +48,7 @@ import org.jboss.as.arquillian.container.ManagementClient;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.test.shared.TimeoutUtil;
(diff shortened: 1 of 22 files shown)


ochaloup:
@kabir @bstansberry could this change be considered for merging or is it something that's not suitable?

bstansberry:
@ochaloup Adding TimeoutUtil to test classes that get deployed means you also have to add permissions.xml settings to the deployment to allow TimeoutUtil to do its thing (i.e. read system properties.) I assume needing this is why there are lots of failures in the security manager job.

Please ask @chengfang or @tadamski to review the EJB bits and @ehsavoie to look at the messaging ones.  That's the bulk of it.

chengfang (approved):

ochaloup:
@bstansberry my bad. thanks Brian for pointing that. I fixed the permissions and hopefully the security manager tests will pass.
@chengfang thank you for the review

bstansberry (approved):
LGTM. I don't need a review from @ehsavoie (although he's welcome to do one); I see now a lot of the messaging stuff was more about mdbs, so Cheng Fang knows that part. I'm comfortable oking the rest.

ochaloup:
Regarding the test failures there are only for Windows JDK11. Those seems not to be caused by these changes.

ehsavoie (approved):

bstansberry:
Thanks, @ochaloup

PR #13371, opened 2020-06-17
servlet-feature-pack-licenses: merge org.wildfly.core:wildfly-event-logger licenses in one entry only

rgdoliveira:


rgdoliveira:
@jamezp @bstansberry can you please take a look in this one?

bstansberry:
@rgdoliveira Eliminating the duplicate declaration is fine but it seems wrong that org.wildfly.core:wildfly-event-logger and org.wildfly.core:wildfly-elytron-integration are in this file at all. They are already in core-feature-pack-license.

rgdoliveira:
@bstansberry I'm afraid we need those 2 licenses there.

If we remove them, the maven license plugin will try to inject them in the licenses.xml generated (docs/licenses/licenses.xml) and the URL for the apache license, for example, would be [1] instead of [2] making the URL inconsistent as it is defined in core-feature-pack-licenses.xml as [2].

[1] http://repository.jboss.org/licenses/apache-2.0.txt
[2] http://www.apache.org/licenses/LICENSE-2.0

bstansberry (approved):

bstansberry:
Thanks, @rgdoliveira

PR #13372, opened 2020-06-17
[WFLY-13606] socket-level keepalive flag in OpenJDK ORB connections

communik:
Restored the ability to use socket-level keepalive flag in OpenJDK ORB connections after it became impossible in Wildfly 11.

Wildfly version 11 brought in the following change: https://github.com/wildfly/wildfly/commit/09b79c4034d13dcc76f9e70b13823efeec1bd630#diff-cabdeec060c4801f721b35660a6081a3R347
which starts to implicitly declare the property com.sun.CORBA.transport.ORBSocketFactoryClass to value org.wildfly.iiop.openjdk.security.NoSSLSocketFactory due to default value of the iiop-openjdk wildfly setting support-ssl being false (see iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPSubsystemAdd.java).
But the code which handles the keepalive property is only found in DefaultSocketFactoryImpl of opendk-orb module, so it is basically never executed.

### DIFF
--- a/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/security/SocketFactoryBase.java
+++ b/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/security/SocketFactoryBase.java
@@ -37,9 +37,25 @@ import java.net.Socket;
 import java.net.SocketException;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
+import java.security.PrivilegedAction;
 
 public abstract class SocketFactoryBase implements ORBSocketFactory {
     protected ORB orb;
+    private static final boolean keepAlive;
+    static {
+        keepAlive = java.security.AccessController.doPrivileged(
+                new PrivilegedAction<Boolean>() {
+                    @Override
+                    public Boolean run () {
+                        String value =
+                                System.getProperty("com.sun.CORBA.transport.enableTcpKeepAlive");
+                        if (value != null)
+                            return new Boolean(!"false".equalsIgnoreCase(value));
+
+                        return Boolean.FALSE;
+                    }
+                });
+    }
 
     @Override
     public void setORB(ORB orb) {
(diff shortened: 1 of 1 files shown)


jamezp:
We would need a JIRA for this. @tadamski could you work with @communik and see if this seems reasonable?

communik:
I'm actually trying to restore my RedHat account, unable to create the issue in JIRA without one. Ready to provide any information.

tadamski:
@jamezp @communik To be perfectly honest, I was working on OpenJDK without JIRA issues and there even is no dedicated ORB project. I will create such project for subsequent JIRAs.

@communik Thanks for the fix! It looks good - I'm merging it.

tadamski (changes requested):
@communik Thanks for the fix! 

I have created the JIRA: https://issues.redhat.com/browse/WFLY-13606

Could you please update the commit title and PR description with it's number?

communik:
@tadamski done. Thanks for the JIRA issue!

tadamski (approved):

tadamski:
@communik Thanks. Merged.

tadamski (approved):

bstansberry:
Thanks, @communik

bstansberry:
@communik 

The WildFly project would like to change its license from the GNU Lesser General Public License version 2.1 to the Apache Software License 2.0. We feel that ASL 2.0 is a better option for the project and our user community, as we find a project's use of LGPL can be a barrier to adoption in some parts of the software community, while we see no such barriers from use of ASL.

Do you agree to a relicensing of your contributions to the WildFly, WildFly Core or JBoss AS 7 projects from LGPL 2.1 to ASL 2.0?

I'm happy to answer any questions you may have about this, and I hope you are doing well. Thank you for your contributions to WildFly and open source!

Best regards,
Brian Stansberry
WildFly Project Lead

PR #13380, opened 2020-06-22
[WFLY-13609] : Disable microprofile tests when XP dist is not used.

panossot:
[WFLY-13609] : https://issues.redhat.com/browse/WFLY-13609

### DIFF
--- a/testsuite/integration/manualmode/pom.xml
+++ b/testsuite/integration/manualmode/pom.xml
@@ -133,128 +133,187 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <!-- Update server profile with shared configuration changes -->
+            <plugin>
+                <groupId>org.wildfly.plugins</groupId>
+                <artifactId>wildfly-maven-plugin</artifactId>
+                <version>${version.org.wildfly.plugin}</version>
+                <executions>
+                    <execution>
+                        <id>ts.config-as.configure-wildfly</id>
+                        <phase>process-test-resources</phase>
+                        <goals>
+                            <goal>execute-commands</goal>
+                        </goals>
+                        <configuration>
+                            <offline>true</offline>
+                            <scripts>
+                                <script>manualmode-all.cli</script>
+                            </scripts>
+                            <jboss-home>${wildfly.dir}</jboss-home>
+                            <stdout>${project.build.directory}/wildfly-plugin.log</stdout>
+                            <java-opts>${modular.jdk.args}</java-opts>
+                            <system-properties>
+                                <maven.repo.local>${settings.localRepository}</maven.repo.local>
+                                <module.path>${jboss.dist}/modules</module.path>
+                            </system-properties>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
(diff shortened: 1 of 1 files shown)


fabiobrz:
Hi @panossot - not sure if I understand this correctly but the notification triggered something in my mind from the last weeks, so pointing you at https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/pom.xml#L111
WDYT=

panossot:
Hello @fabiobrz. Some tests were failing... I will check with this profile as well.

panossot:
@fabiobrz I have modified the pr...

fabiobrz:
Hi @panossot - and sorry for the late reply.
Let me clarify what I meant with my previous - and maybe too short :) - message.
I noticed that basically you were adding a Maven profile to skip some tests - and you even confirmed it later as you said that some of them were failing.
From the PR description I see that those failures are due to the fact that with non-XP distribution we don't have MicroProfile subsystems available, right? 
In such case I am wondering what the following profile is used for:

* https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/pom.xml#L111

Am I missing something about this?
Actually In this PR I see you're disabling just one est in case a property activates the new profile, but both the PR description and the JIRAs talk about disabling whole MP tests for non-XP distribution. 
WDYT?

fabiobrz:
Hi @panossot - I see it now, you're in `integration/manual`, actually, while I was looking at the MP TCKs. Sorry about that. Then ignore my previous comments, thanks :)

Maybe just a hint for the property/profile name would be to follow the WIldfly convention - i.e. https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile-tck/pom.xml#L85

Some time ago I created one similar profile in another TS and after having the PR merged I noticed the naming convention in Wildfly, so just my two cents here. 
WDYT?

panossot:
@fabiobrz the activation property (disable.microprofile.tests) should not change, so that it is in harmony with the equivalent one used for microprofile tests ... For the profile ids you can propose whatever you would like...

fabiobrz:
> @fabiobrz the activation property (disable.microprofile.tests) should not change, so that it is in harmony with the equivalent one used for microprofile tests ... For the profile ids you can propose whatever you would like...

Then I am perfectly ok with that. Thanks for your explanation!

bstansberry:
@panossot This breaks using -DnoManualMode to disable manual mode tests.

Also, it duplicates a lot of fussy surefire configuration stuff.

How about something like this:

1) Everything currently in the 'manualmode.integration.tests.profile' profile gets pulled out of a profile and becomes part of the top level 'build' configuration.

2) New 'no-manual-mode' profile, activated by -DnoManualMode that overrides the default executions by setting changing their phase to 'none', as https://github.com/wildfly/wildfly/blob/master/testsuite/integration/microprofile/pom.xml#L301 does

3) The 'disable.microprofile.profile' does a minimal override of the default-test surefire execution that only adds the exclude.

panossot:
The changes should be done ...

bstansberry (approved):
LGTM @panossot Thanks!

bstansberry:
Thanks, @panossot

PR #13390, opened 2020-06-30
[WFLY-13626] make transaction subsytem CMR attributes optional

ochaloup:
https://issues.redhat.com/browse/WFLY-13626

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem15Parser.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem15Parser.java
@@ -155,22 +155,9 @@ class TransactionSubsystem15Parser extends TransactionSubsystem14Parser {
                 case START_ELEMENT: {
                     switch (Element.forName(reader.getLocalName())) {
                         case CM_TABLE: {
-                            for (Attribute attribute : Attribute.values()) {
-                                switch (attribute) {
-                                    case NAME: {
-                                        addAttribute(reader, cmrOperation, CMResourceResourceDefinition.CM_TABLE_NAME);
-                                        break;
-                                    }
-                                    case CM_TABLE_BATCH_SIZE:
-                                        addAttribute(reader, cmrOperation, CMResourceResourceDefinition.CM_TABLE_BATCH_SIZE);
-                                        break;
-                                    case CM_TABLE_IMMEDIATE_CLEANUP:
-                                        addAttribute(reader, cmrOperation, CMResourceResourceDefinition.CM_TABLE_IMMEDIATE_CLEANUP);
-                                        break;
-                                    default:
-                                        break;
-                                }
-                            }
+                            addAttribute(reader, cmrOperation, CMResourceResourceDefinition.CM_TABLE_NAME);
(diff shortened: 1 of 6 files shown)


ochaloup:
@kabir if you have a minute would you be so kind and check if the fix on the txn parser is fine?

kabir:
@ochaloup Only looked quickly while on the meeting, but the original code looks fine to me. If an attribute has a default value (and/or is optional) then it does not need to be part of the add operation.

What is the problem you're trying to fix?

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem15Parser.java
```diff
@@ -181,7 +168,7 @@ private void parseCM(XMLExtendedStreamReader reader, List<ModelNode> operations)
     }
 
     private void addAttribute(XMLExtendedStreamReader reader, ModelNode operation, SimpleAttributeDefinition attributeDefinition) throws XMLStreamException {
-        String value = rawAttributeText(reader, attributeDefinition.getXmlName(), null);
+        String value = rawAttributeText(reader, attributeDefinition.getXmlName(), attributeDefinition.getDefaultValue().asString());
```
bstansberry, line 171: @ochaloup The parser can't do this. It needs to preserve what it read from the xml, not introduce default values into the management model that were not what the user put in the xml. The subsystem can use a default value from the AttributeDefinition to configure the runtime services, but what's stored in the management model should reflect the user provided configuration.

It appears this method is only called by the parseCM method, so I think just eliminating the 'else' block that throws missingRequired would be fine.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem30Parser.java
```diff
@@ -185,7 +172,7 @@ private void parseCM(XMLExtendedStreamReader reader, List<ModelNode> operations)
     }
 
     private void addAttribute(XMLExtendedStreamReader reader, ModelNode operation, SimpleAttributeDefinition attributeDefinition) throws XMLStreamException {
-        String value = rawAttributeText(reader, attributeDefinition.getXmlName(), null);
+        String value = rawAttributeText(reader, attributeDefinition.getXmlName(), attributeDefinition.getDefaultValue().asString());
```
bstansberry, line 175: Same comment as I made for TransactionSubsystem15Parser.

bstansberry (changes requested):

ochaloup:
@kabir the trouble was that attributes behaves as required ones instead of optional
@bstansberry oh, I see. By your suggestion I removed the exception thrown on the missing parameter. It works fine.
I think the parser with my original change was preserving the XML part. But I see it all went against the design principle.

How would you asses the current state?

bstansberry (approved):
@ochaloup LGTM; thanks.

The issue with storing a default value that wasn't read from the xml in the model is if a later management op that modifies the model is invoked, the server will write over the standalone.xml in order to persist the updated model. And when it does that it will write the default value in the xml.

bstansberry:
Thanks, @ochaloup

PR #13391, opened 2020-06-30
WFLY-13627 Distributed sessions/SFSBs stored in non-transactional invalidation-cache should schedule expirations locally

pferraro:
https://issues.redhat.com/browse/WFLY-13627


ivassile:
We are doing EAP 7.3.2.CR2 and need to include some critical issues. Can we get this issue merged upstream so we can include it in CR2? Thanks!

PR #13392, opened 2020-07-01
WFLY-13628 Invalidation caches need to consider keys in the cache store when reassigning ownership

pferraro:
https://issues.redhat.com/browse/WFLY-13628


rhusar (approved):

ivassile:
We are doing EAP 7.3.2.CR2 and need to include some critical issues as quickly as possible. Can we get this issue merged upstream so we can include it in CR2? Thanks!

PR #13394, opened 2020-07-02
[WFLY-13623][WFLY-13624] Organize feature pack code to facilitate separate EE 8 and EE 9 variants

bstansberry:
https://issues.redhat.com/browse/WFLY-13623
Also https://issues.redhat.com/browse/WFLY-13624

Also https://issues.redhat.com/browse/WFLY-13622 which is needed to allow license generation from the multiple license configuration files the result from the refactoring described below.

This is essentially a code reorganization:

Convert the 'xxx-feature-pack' maven module to a parent module containing submodules related to both the legacy xxx fp and the galleon one. The top level core-galleon-pack maven module is eliminated. Here xxx can mean "servlet", "ee" or "" for the case of the top level FP.

Under 'xxx-feature-pack' create a 'common' submodule that holds content that is expected to be usable in the legacy fp, the current EE 8 galleon FP, and a future EE 9 galleon FP. This includes the various module.xml files, a license information file for the artifacts referenced in those module.xml files, other shared content, and the dependency set declaration in the pom for all the artifacts referenced in the module.xml files.

Under 'xxx-feature-pack' create a 'ee-8-api' submodule that holds content that is expected to be usable in the legacy fp and the current EE 8 galleon FP but not by a future EE 9 galleon FP. This includes the various module.xml files, a license information file for the artifacts referenced in those module.xml files,, and the dependency set declaration in the pom for all the artifacts referenced in the module.xml files.

Under 'xxx-feature-pack' create a 'pruned' submodule that holds content that is expected to be usable in the legacy fp and the current EE 8 galleon FP but not by a future EE 9 galleon FP. This includes the various module.xml files, a license information file for the artifacts referenced in those module.xml files,, and the dependency set declaration in the pom for all the artifacts referenced in the module.xml files. The usage of this is the same as 'ee-8-api' but I use a separate maven module to distinguish the EE APIs -- which will exist in both an EE 8 and EE FP but will differ in details -- from items we just don't want to carry forward into the EE 9 FP.

Under 'xxx-feature-pack' create a 'galleon-common' submodule that holds content that is expected to be usable in the current EE 8 galleon FP and a future EE 9 galleon FP. It is not meant to include galleon content that would not be used in both FP variants, e.g. deprecated config items that won't appear in the EE 9 variant without a conscious decision to bring them in.

Under 'xxx-feature-pack' create a 'galleon-pruned' submodule that holds content that is expected only to be usable in the current EE 8 galleon FP. This likely will only exist in the servlet-feature-pack as the goal is to allow the artifact it produces to be shared between the servlet-galleon-pack and the ee-galleon-pack. The goal being to sever the dependency relationship ee-galleon-pack has on servlet-galleon-pack.

Under 'xxx-feature-pack' create a 'galleon-feature-pack' submodule that replaces the previous top level 'xxx-galleon-pack' module. It copies in content from the 'common', 'ee-8-api', 'galleon-common' and 'galleon-legacy' sibling maven modules discussed above, and also depends on the 'common' and 'ee-8-api' poms, as a source of dependency information. It may also have some content of its own that is unique to the EE 8 galleon FP.

Under 'core-feature-pack' create a 'legacy-feature-pack' submodule that replaces the function of the previous top level 'core-feature-pack' module. It copies in content from the 'common' and 'ee-8-api' sibling maven modules discussed above, and also depends on the 'common' and 'ee-8-api' poms, as a source of dependency information. It also has some content of its own that is unique to the legacy FP.


bstansberry:
@kabir This breaks the 'Dependency Tree' tests because the test is looking for a 'feature-pack' maven module that no longer exists. I could change the test config to use 'ee-feature-pack' but then all the other PRs will fail.

kabir:
@bstansberry Probably best to wait until this one is merged

jamezp (approved):

PR #13397, opened 2020-07-06
[WFLY-12342] Add server probes for readiness health check

jmesnil:
[WFLY-12342] Add server probes for readiness health check

Add 3 healthcheck probes to determine the readiness of the server:

* server-state returns UP when the server is "running"
* boot-errors returns UP if there were no boot-errors
* deployments-status returns UP if all deployments status are OK

JIRA: https://issues.jboss.org/browse/WFLY-12342

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
@@ -19,18 +19,45 @@ element to the xml or by using the following CLI operation:
 [standalone@localhost:9990 /]/extension=org.wildfly.extension.microprofile.health-smallrye:add
 ----
 
-== Management Operation
+== Management Operations
 
-The healthiness of the application server can be queried by calling the `check`:
+The healthiness of the application server can be queried by calling 3 different operations:
+
+* `check` to check both the liveness and readiness of the runtime
+* `check-live` to check only the liveness of the runtime
+* `check-ready` to check only the readiness of the runtime
+
 
 [source,options="nowrap"]
 ----
 [standalone@localhost:9990 /] /subsystem=microprofile-health-smallrye:check
 {
-    "outcome" => "success", <1>
+    "outcome" => "success",
     "result" => {
-        "outcome" => "UP", <2>
-        "checks" => []
+        "status" => "UP",
+        "checks" => [
+            {
+                "name" => "empty-liveness-checks",
+                "status" => "UP"
+            },
+            {
+                "name" => "server-state",
+                "status" => "UP",
+                "data" => {"value" => "running"}
+            },
+            {
+                "name" => "boot-errors",
+                "status" => "UP"
+            },
+            {
+                "name" => "deployments-status",
+                "status" => "UP"
+            },
+            {
+                "name" => "empty-readiness-checks",
+                "status" => "UP"
+            }
+        ]
     }
 }
 ----
@@ -40,6 +67,7 @@ The healthiness of the application server can be queried by calling the `check`:
 == HTTP Endpoints
 
 The MicroProfile Health Check specifications defines three HTTP endpoints:
+
 * `/health` to test both the liveness and readiness of the runtime.
 * `/health/live` to test the liveness of the runtime.
 * `/health/ready` to test the readiness of the runtime.
@@ -58,7 +86,7 @@ If the application server is healthy, it will return a `200 OK` response:
 $ curl -v http://localhost:9990/health
 < HTTP/1.1 200 OK
 ...
-{"outcome":"UP","checks":[]}
+{"status":"UP","checks":[{"name":"empty-liveness-checks","status":"UP"},{"name":"server-state","status":"UP","data":{"value":"running"}},{"name":"boot-errors","status":"UP"},{"name":"deployments-status","status":"UP"},{"name":"empty-readiness-checks","status":"UP"}]}
 ----
 
 If the application server (and its deployment) is not healthy, it returns `503 Service Unavailable`
@@ -79,28 +107,44 @@ created by the `add-user` script. For example:
 $ curl -v --digest -u myadminuser:myadminpassword http://localhost:9990/health
 < HTTP/1.1 200 OK
 ...
-{"outcome":"UP","checks":[]}
+{"status":"UP","checks":[{"name":"empty-liveness-checks","status":"UP"},{"name":"server-state","status":"UP","data":{"value":"running"}},{"name":"boot-errors","status":"UP"},{"name":"deployments-status","status":"UP"},{"name":"empty-readiness-checks","status":"UP"}]}
 ----
 
 If the authentication fails, the  server will reply with a `401 NOT AUTHORIZED` response.
 
-=== Global Status when probes are not defined
+=== Default Server Procedures
 
-By default, the HTTP endpoints will return `UP` if there are no health check probes defined by the application.
+WildFly provides some readiness procedures that are checked to determine if the application server is ready to serve requests:
 
-However, it is possible to change this behaviour.
-The `/health/live` HTTP endpoint (as well as the `check-live` management operation) can return `DOWN` when there are no liveness check probes by setting the `empty-liveness-checks-status` attribute
-on the `subsystem` resource to `DOWN`.
-Similarly, the `/health/ready` HTTP endpoint (as well as the `check-ready` management operation) can return `DOWN` when there are no readiness check probes by setting the `empty-readiness-checks-status` attribute
-on the `subsystem` resource to `DOWN`.
+* `boot-errors` checks that there were no errors during the server boot sequence
+* `deployments-status` checks that all deployments were deployed without errors
+* `server-state` checks that the server state is `running`
+* `empty-readiness-checks` determines the status when there are no readiness check procedures deployed to the server. The outcome of this procedure is determined by the `empty-readiness-checks-status` attribute. If the attribute is
+   `UP` (by default), the server can be ready when there are no readiness checks in the deployments. Setting the `empty-readiness-checks-status` attribute to `DOWN` will make this procedure fail when there are no readiness checks in the deployments.
 
-This allows application to ensure that the HTTP liveness and readiness endpoints will return `DOWN` until their probes are deployed and used to determine the actual
-liveness and readiness state of the application.
+If a deployment does not provide any readiness checks, WildFly will automatically add one for each deployment (named `ready-<deployment name>`) which always returns `UP`.
 
 [NOTE]
-There is a specific behaviour for deployments that provides no _readiness_ probes. In that the case, WildFly automatically adds a readiness probe for the deployment that always return
-`UP`. This allows existing deployments that do not provide readiness probes to configure WildFly so it will not be ready until this deployment is actually deployed.
+====
+This allows applications that does not provide readiness checks to still be able to inform cloud containers when they are ready to serve requests.
+Setting empty-readiness-checks-status` to `DOWN` ensures that the server will not be ready until the application is deployed. At that time, the `ready-<deployment name>`
+will be added (which returns `UP`) and the `empty-readiness-checks` procedure will no longer be checked as there is no a readiness check procedure provided by the deployment.
+====
 
+WildFly also provide a liveness procedure that is checked to determine if the application server is live:
+
+* `empty-liveness-checks` determines the status when there are no liveness check procedures deployed to the server. The outcome of this procedure is determined by the `empty-liveness-checks-status` attribute. If the attribute is
+`UP` (by default), the server can be live when there are no liveness checks in the deployments.  Setting the `empty-liveness-checks-status` attribute to `DOWN` will make this procedure fail when there are no liveness checks in the deployments.
+
+It is possible to disable all these server procedures by using the MicroProfile Config property `mp.health.disable-default-procedures`.
+
+[NOTE]
+====
+The MicroProfile Config property `mp.health.disable-default-procedures` is read at 2 different times:
+
+1. __When the server starts__, to determine if its server procedures should be disabled or enabled. It can be set using the system property `mp.health.disable-default-procedures` or the environment variable `MP_HEALTH_DISABLE_DEFAULT_PROCEDURES`. Setting this property in a deployment is ignored at that time.
+2. __When an application is deployed__, to determine if WildFly should add a readiness check if the deployment does not provide any. At that time, setting this property in a `microprofile-config.properties` file in the deployment would be taken into account. (with the usual priority rules for MicroProfile Config properties).
+====
 
 == Component Reference
 
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
@@ -22,16 +22,27 @@
 
 package org.wildfly.extension.microprofile.health;
 
+import static org.wildfly.extension.microprofile.health.MicroProfileHealthSubsystemDefinition.CLIENT_FACTORY_CAPABILITY;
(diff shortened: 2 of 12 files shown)


jmesnil:
Dev analysis is at https://github.com/wildfly/wildfly-proposals/pull/237

kabir (approved):
Although we're missing process to be able to merge it, the code looks good

testsuite/integration/microprofile-tck/health/src/test/resources/arquillian.xml
```diff
@@ -5,7 +5,7 @@
     <container qualifier="jboss" default="true" >
         <configuration>
             <property name="jbossHome">${jboss.home}</property>
-            <property name="javaVmArguments">${microprofile.jvm.args}</property>
+            <property name="javaVmArguments">${microprofile.jvm.args} -D__MP_HEALTH_TCK_DISABLE_SERVER_CHECKS=true</property>
```
mchoma, line 8: Why is it important to disable it? Was new output breaking MP TCK?

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
```diff
@@ -40,30 +51,52 @@
...
         this.emptyReadinessChecksStatus = emptyReadinessChecksStatus;
     }
 
     @Override
     public void start(StartContext context) {
         healthReporter = new HealthReporter(emptyLivenessChecksStatus, emptyReadinessChecksStatus);
+
+        modelControllerClient = modelControllerClientFactory.get().createClient(managementExecutor.get());
+
+        if (System.getProperty("__MP_HEALTH_TCK_DISABLE_SERVER_CHECKS") == null) {
```
mchoma, line 87: So it basically disable all three checks. Can we remove `_TCK_`part so it does not look like some workaround flag. Then we can also advertise that as feature compatibility flag. If for some reason new behaviour  make problem to someone.

testsuite/integration/microprofile-tck/health/src/test/resources/arquillian.xml
jmesnil: The TCK must be run with any vendor probes disabled: https://github.com/eclipse/microprofile-health/blob/master/tck/running_the_tck.asciidoc#disabling-default-vendor-procedures

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
jmesnil: that's a good point, I should replace that code by a proper call to MP Config and check the `mp.health.disable-default-procedures` config property instead

testsuite/integration/microprofile-tck/health/src/test/resources/arquillian.xml
jmesnil: fixed

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
jmesnil: fixed

bstansberry:
@jmesnil https://github.com/wildfly/wildfly-proposals/pull/237 describes some community docs for this. Can you add those here?

jmesnil:
> @jmesnil [wildfly/wildfly-proposals#237](https://github.com/wildfly/wildfly-proposals/pull/237) describes some community docs for this. Can you add those here?

Sorry for the omission. I've amended the commit with them (see https://github.com/wildfly/wildfly/blob/bb25e9aa793f2c19658af1444872bcf313fab057/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc#default-server-readiness-procedures)

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
```diff
@@ -79,29 +99,36 @@ created by the `add-user` script. For example:
...
+* `server-state` checks that the server state is `running`
+
+The server is ready when all these procedures are `UP`.
+
+It is possible to disable them by using the MicroProfile Config property `mp.health.disable-default-procedures`.
+
 === Global Status when probes are not defined
 
-By default, the HTTP endpoints will return `UP` if there are no health check probes defined by the application.
+By default, the HTTP endpoints and management operations will return `UP` if there are no health check procedures defined by the applications deployed on WildFly.
```
mchoma, line 121: is this sentence true for management operation after changes of this RFE? /health/ready can return `DOWN` in case of failed deployment

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
```diff
@@ -79,29 +99,36 @@ created by the `add-user` script. For example:
...
 
 However, it is possible to change this behaviour.
-The `/health/live` HTTP endpoint (as well as the `check-live` management operation) can return `DOWN` when there are no liveness check probes by setting the `empty-liveness-checks-status` attribute
+The `/health/live` HTTP endpoint (as well as the `check-live` management operation) can return `DOWN` when there are no liveness check procedures in the deployments by setting the `empty-liveness-checks-status` attribute
 on the `subsystem` resource to `DOWN`.
-Similarly, the `/health/ready` HTTP endpoint (as well as the `check-ready` management operation) can return `DOWN` when there are no readiness check probes by setting the `empty-readiness-checks-status` attribute
+Similarly, the `/health/ready` HTTP endpoint (as well as the `check-ready` management operation) can return `DOWN` when there are no readiness check procedures in the deployments by setting the `empty-readiness-checks-status` attribute
```
bstansberry, line 126: Do the 'Default Server Readiness Procedures' count as "procedures in the deployments"? If they were not disabled with the MP Config property, would setting this attribute to DOWN result in DOWN or UP?

bstansberry (changes requested):
Reading the docs I still have questions about how the 'Default Server Readiness Procedures' impact things, so I think some clarification is needed.

jmesnil:
@bstansberry @mchoma I've rewritten the doc to make it better to read. The `empty-liveness-checks` and `empty-readiness-checks` procedures are server check procedures whose behaviour is controlled by the `empty-liveness-checks-status` and `empty-readiness-checks-status` management attribute.
They are impacted by the MP Config property like the new server check procedures added by this RFE.

jmesnil:
> @jmesnil Is `ready-<deployment name>` subset of `deployments-status` ? Can we get rid of that? `ready-<deployment name> always returns UP` means it does not reflect deployment errors?

They serve different purpose.
`ready-<deployment name>` is added by WildFly for *each* deployment if the deployment does not contain any readiness check. It only returns `UP` for this deployment. However if the deployment provides a readiness check, this one will be used to determine if the deployment is ready or not.
`deployments-status` checks that *all deployments* have been deployed without error.

> Is it possible `empty-readiness-checks-status` make work with `deployments-status` ?
> This is somehow backward incompatible change, but would IMO simplify whole logic of health check.

I'm not sure it's a good idea. The goal of this RFE is to provides the *same* checks that the EAP probes provide using the health subsystem instead of external scripts.
Any change of behaviour would have to be in a way that the semantics of the "healthiness" of the app server are preserved regardless of the way they are checked (external scripts and health subsystem).

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
```diff
@@ -19,18 +19,45 @@ element to the xml or by using the following CLI operation:
...
+            {
+                "name" => "boot-errors",
+                "status" => "UP"
+            },
+            {
+                "name" => "deployments-status",
+                "status" => "UP"
+            },
+            {
+                "name" => "empty-readiness-checks",
```
mchoma, line 57: So why does this show up in WF21. In previous version it did not show?

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
```diff
@@ -79,28 +107,44 @@ created by the `add-user` script. For example:
...
+
+WildFly provides some readiness procedures that are checked to determine if the application server is ready to serve requests:
+
+* `boot-errors` checks that there were no errors during the server boot sequence
+* `deployments-status` checks that all deployments were deployed without errors
+* `server-state` checks that the server state is `running`
+* `empty-readiness-checks` determines the status when there are no readiness check procedures deployed to the server. The outcome of this procedure is determined by the `empty-readiness-checks-status` attribute. If the attribute is
+   `UP` (by default), the server can be ready when there are no readiness checks in the deployments. Setting the `empty-readiness-checks-status` attribute to `DOWN` will make this procedure fail when there are no readiness checks in the deployments.
+
+If a deployment does not provide any readiness checks, WildFly will automatically add one for each deployment (named `ready-<deployment name>`) which always returns `UP`.
```
mchoma, line 125: > > @jmesnil Is `ready-<deployment name>` subset of `deployments-status` ? Can we get rid of that? `ready-<deployment name> always returns UP` means it does not reflect deployment errors?
> 
> They serve different purpose.
> `ready-<deployment name>` is added by WildFly for _each_ deployment if the deployment does not contain any readiness check. 
And my point is if we can't use deployments-status for this purpose?
 
> It only returns `UP` for this deployment. However if the deployment provides a readiness check, this one will be used to determine if the deployment is ready or not.
> `deployments-status` checks that _all deployments_ have been deployed without error.

few notes:
MP should serve one deployment.
I do have detailed info in `deployments-status` about deployments so I am not loosing this information per deployment anyway.
```
            {
                "name" => "deployments-status",
                "status" => "UP",
                "data" => {"microprofile-config.war" => "OK"}
            },
```
 
Can it happen  `ready-<deployment name>` is UP but `deployments-status` is DOWN (with one deployment)? Can it happen  `ready-<deployment name>` is DOWN but `deployments-status` is UP (with one deployment)? In other words do they have still same result?

> 
> > Is it possible `empty-readiness-checks-status` make work with `deployments-status` ?
> > This is somehow backward incompatible change, but would IMO simplify whole logic of health check.
> 
> I'm not sure it's a good idea. The goal of this RFE is to provides the _same_ checks that the EAP probes provide using the health subsystem instead of external scripts.
> Any change of behaviour would have to be in a way that the semantics of the "healthiness" of the app server are preserved regardless of the way they are checked (external scripts and health subsystem).


I am not sure if this should be added under this RFE (mainly because of backward compatibility questions), but thinking about it with fresh context. Just it was hard to me to understand all combinations. And I am thinking how to simplify this unnecessary complexity. 

Also about `empty-readiness-checks-status`. Cant problem which is `empty-readiness-checks-status` fixing happen to `deployments-status` ,`boot-errors`, `server-state`? I mean that :check is reporting at beginning UP, but after some of this check finished it becomes DOWN? Don't we need to force that somehow globally? To report DOWN, until all checks finished? (I don't know if that is feasible)

jmesnil: WildFly supports multiple deployments. We should strive to take that into account when providing MP features if that does not conflict with the spec.

* If the deployment has no readiness checks*, `ready-<deployment name>` and `deployments-status` (assuming there is a single deployment) will have the same outcome
* If the deployment has readiness checks, the outcome will be determined by them.

> Cant problem which is empty-readiness-checks-status fixing happen to deployments-status ,boot-errors, server-state?

There is a short window where it is true: when `server-state` is `UP` (the server is running) and the deployment are not finished (deployments-status will not report any errors as the deployments are processed.
This windows can be alleviated by setting `empty-readiness-checks-status` to `DOWN` so that the whole server readiness will be `DONE` until deployments have been successfully deployed.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
jmesnil: There was a bug in the code that I fixed in this PR

mchoma: Thanks that explain some changes in tests, which I wasn't expecting.

jmesnil:
I've removed the `missing-reqs` label as the pre-check for EAP7-1322 is done

claudioweiler:
Hi. Just setup 21.0.0, and my /health/ready is reporting wrong DOWN status:
`{"status":"DOWN","checks":[
  {"name":"boot-errors","status":"UP"},
  {"name":"server-state","status":"DOWN"},
  {"name":"deployments-status","status":"DOWN"},
  {"name":"ready-deployment.sst-ear-1.0.0-SNAPSHOT.ear.sst.war","status":"UP"}
]}`

We can see that 'deployments-status' is DOWN but the only deployment is UP. I confirm that this is the only deployment. Or, without deployment:
`{"status":"DOWN","checks":[{"name":"deployments-status","status":"DOWN"},{"name":"server-state","status":"DOWN"},{"name":"boot-errors","status":"UP"},{"name":"empty-readiness-checks","status":"UP"}]}`

The 'server-state' shows DOWN too, but server is OK, deployment is running, console running, deployment status in console is green.

I don't know if it's relevant, but we use management console protected by Keycloak, and we already has problems with this setup: https://issues.redhat.com/browse/WFLY-11933

mchoma:
@claudioweiler this is really weird. Please can you create WFLY jira to continue discussion. Can you simplify reproducing steps to minimum? I have tried enabling RBAC from reproducer of WFLY-11933, but this make no difference.

I have tried again on empty WFLY21 and it works as expected ```[standalone@localhost:9990 /] /subsystem=microprofile-health-smallrye:check
{
    "outcome" => "success",
    "result" => {
        "status" => "UP",
        "checks" => [
            {
                "name" => "empty-readiness-checks",
                "status" => "UP"
            },
            {
                "name" => "deployments-status",
                "status" => "UP"
            },
            {
                "name" => "boot-errors",
                "status" => "UP"
            },
            {
                "name" => "server-state",
                "status" => "UP",
                "data" => {"value" => "running"}
            },
            {
                "name" => "empty-liveness-checks",
                "status" => "UP"
            }
        ]
    }
}
```

What is your configuration? Can you attach  server log?

**Comment by claudioweiler** [comment]
Hi @mchoma I will try to reproduce it at minimum to create an issue.

Just to add more info, I tested it with jboss-cli and it really returns UP, so, the DOWN status is received in management address request (http://server:9990/health/ready), or in management console (HAL).

```shell
[standalone@localhost:9990 /] /subsystem=microprofile-health-smallrye:check-ready
{
    "outcome" => "success",
    "result" => {
        "status" => "UP",
        "checks" => [
            {
                "name" => "boot-errors",
                "status" => "UP"
            },
            {
                "name" => "server-state",
                "status" => "UP",
                "data" => {"value" => "running"}
            },
            {
                "name" => "deployments-status",
                "status" => "UP",
                "data" => {
                    "sst-ear-1.0.0-SNAPSHOT.ear" => "OK",
                    "sst-cliente.war" => "OK"
                }
            },
            {
                "name" => "ready-deployment.sst-ear-1.0.0-SNAPSHOT.ear.sst.war",
                "status" => "UP"
            }
        ]
    }
}
#---------------------------------------
$ curl http://localhost:9990/health/ready
{"status":"DOWN","checks":[{"name":"boot-errors","status":"UP"},{"name":"server-state","status":"DOWN"},{"name":"deployments-status","status":"DOWN"},{"name":"ready-deployment.sst-cliente.war","status":"UP"},{"name":"ready-deployment.sst-ear-1.0.0-SNAPSHOT.ear.sst.war","status":"UP"}]}
```

claudioweiler:
Issue registered: https://issues.redhat.com/browse/WFLY-14034

PR #13399, opened 2020-07-07
[WFLY-13647] Annotation processing error sun.reflect.annotation.TypeNotPresentExceptionProxy does not indicate issue

gaol:
Issue: https://issues.redhat.com/browse/WFLY-13647

There is a bug with `Class.getAnnotation()` call in `JDK < 11` [1]  which throws `ArrayStoreException` when the annotation refers to some classes not found in current class path. The `ArrayStoreException` confuses user why it failed because the root cause is missing.

This PR tries to catch the exception and give more readable feedback in this case.

[1] https://bugs.openjdk.java.net/browse/JDK-7183985

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
@@ -153,7 +153,14 @@ public class BusinessViewAnnotationProcessor implements DeploymentUnitProcessor 
     }
 
     private Collection<Class<?>> getRemoteBusinessInterfaces(final DeploymentUnit deploymentUnit, final Class<?> sessionBeanClass) throws DeploymentUnitProcessingException {
-        final Remote remoteViewAnnotation = sessionBeanClass.getAnnotation(Remote.class);
+        final Remote remoteViewAnnotation;
+        try {
+            remoteViewAnnotation = sessionBeanClass.getAnnotation(Remote.class);
+        } catch (ArrayStoreException e) {
+            // https://bugs.openjdk.java.net/browse/JDK-7183985
+            // Class.findAnnotation() has a bug under JDK < 11 which throws ArrayStoreException
+            throw EjbLogger.ROOT_LOGGER.missingClassInAnnotation(Remote.class.getSimpleName(), sessionBeanClass.getName());
+        }
         if (remoteViewAnnotation == null) {
             Collection<Class<?>> interfaces = getBusinessInterfacesFromInterfaceAnnotations(sessionBeanClass, Remote.class);
             if (!interfaces.isEmpty()) {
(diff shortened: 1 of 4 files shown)


bstansberry:
@chengfang or @tadamski or @fl4via please review.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/BusinessViewAnnotationProcessor.java
```diff
@@ -197,12 +211,18 @@ private void processViewAnnotations(final DeploymentUnit deploymentUnit, final C
...
-            if (iface.getAnnotation(annotation) != null) {
-                businessInterfaces.add(iface);
+            try {
+                if (iface.getAnnotation(annotation) != null) {
+                    businessInterfaces.add(iface);
+                }
+            } catch (ArrayStoreException e) {
+                // https://bugs.openjdk.java.net/browse/JDK-7183985
+                // Class.findAnnotation() has a bug under JDK < 11 which throws ArrayStoreException
+                throw EjbLogger.ROOT_LOGGER.missingClassInAnnotation(annotation.getSimpleName(), iface.getName());
```
chengfang, line 225: I suspect we don't need exception handling here. When `@Remote @Local` is used in a business interface, they do not take any arguments.  The no-arg annotation is to indicate that the annotated business interface is a local or remote business interface.  For ex,

```java
@Remote 
public interface Calculator {...}
```

gaol: I think it should be good if users do it in correct way(like they should package classes referred by the annotation). :)
but what happens if users referred some classes in the business interfaces' Remote annotation by mistake? is it still a valid deployment?

chengfang: If `@Remote` or `@Local` on business interface has param, then it's an application's error, though the spec does not say what should happen in this case. WildFly just ignores the param, which happens after the annotation processing. So to tolerate this type of user errors, it's okay to keep this checking.

chengfang (approved):

bstansberry:
@chengfang This needs a conflict resolved.

chengfang:
@gaol can you update your local repo and increment the log id in `EjbLogger`  `@Message(id = 520` to `521`, since 520 is already taken?

The failure in Windows JDK11 seem unrelated intermittent failure. I just created [WFLY-13682](https://issues.redhat.com/browse/WFLY-13682) (JBossWSStringReplaceTestCase.incorrectValueOfModifyWsdlAddressOpeningElement failed intermittently on Windows) to track it separately.

gaol:
@chengfang Thanks for the instructions, I just updated the PR to fix the conflicts. :)

chengfang:
Linux JDK8 and Linux JDK11 jobs failed on July 20, 2020, with 12 failures in ` MixedDomainDeployment700TestCase`. I think they are unrelated failures and can be ignored for this PR. I've just reopened the following JIRA to track it separately:

[WFLY-12649](https://issues.redhat.com/browse/WFLY-12649)
Deal with mixed-domain failures in the OpenJ9 jobs

PR #13401, opened 2020-07-08
WFLY-13653 Upgrade smallrye-health to 2.2.2

xstefank:
https://issues.redhat.com/browse/WFLY-13653

The implementation was rewritten to be fully reactive (based on Mutiny).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -299,7 +299,7 @@
         <version.io.rest-assured>3.0.6</version.io.rest-assured>
         <version.io.smallrye.smallrye-config>1.6.2</version.io.smallrye.smallrye-config>
         <version.io.smallrye.smallrye-fault-tolerance>4.2.1</version.io.smallrye.smallrye-fault-tolerance>
-        <version.io.smallrye.smallrye-health>2.2.0</version.io.smallrye.smallrye-health>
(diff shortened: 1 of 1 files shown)


bstansberry (commented):
@jmesnil Please review.

jmesnil:
Diff to the previous integrated release: https://github.com/smallrye/smallrye-health/compare/2.2.0...2.2.2

jmesnil:
Based on the diff, I see some changes that needs to be taken into consideration:
* this version of smallrye-health consumes `smallrye-config:1.8.1` and adds a dependency to `smallrye-common:1.0.2`
* the `smallrye-health`artifact depends on `smallrye-health-api` which adds a new dependency to `mutiny`
* new experimental features (async health check, wellness) have been introduced to smallrye-health.
* there is no change in the supported version of Eclipse MicroProfile Health.
* I have not see any critical bug fix or higher


@bstansberry I don't think that there is a need to bump this version of smallrye-health that would incorporate new dependencies as we will not expose the new experimental smallrye features in WildFly for the time being.
I suggest we bump `smallrye-health` when a new version of `microprofile-health` is released.

xstefank:
fair enough. I really just wanted to test mutiny based impl against wildfly on all platforms. Closing.

PR #13402, opened 2020-07-08
[WFLY-13650] Allow expressions for ordinal attribute

jmesnil:
JIRA: https://issues.redhat.com/browse/WFLY-13650


bstansberry (approved):

kabir:
@jmesnil @bstansberry Should this not have a model version bump?

pferraro:
This also needs a model transformer to reject expressions for legacy hosts.

jmesnil:
@kabir @pferraro you are both right, I'll update the PR to minor bump the management model and add a transformer for it.

pferraro (approved):
LGTM.

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigExtension.java
```diff
@@ -55,7 +55,8 @@
     private static final String RESOURCE_NAME = MicroProfileConfigExtension.class.getPackage().getName() + ".LocalDescriptions";
 
     protected static final ModelVersion VERSION_1_0_0 = ModelVersion.create(1, 0, 0);
-    private static final ModelVersion CURRENT_MODEL_VERSION = VERSION_1_0_0;
+    protected static final ModelVersion VERSION_1_1_0 = ModelVersion.create(1, 1, 0);
+    private static final ModelVersion CURRENT_MODEL_VERSION = VERSION_1_1_0;
```
rhusar, line 59: Why minor? We have been doing a major bump for major WildFly versions.

rhusar (changes requested):

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigExtension.java
jmesnil: I thought about this and I'm not sure that change requires a major bump. I'm not modifying the management model in a significant way: there is only 1 flag changed on a single attribute.
I thinks this bug fix fits a minor version bump but if we have required a major bump no matter the changes, I can amend my PR.

bstansberry: The basic reasons we do major version bumps is:

1) To ensure we always have space for new API versions downstream, in whatever downstream variants we may care about.
2) KISS principle.

You have a really strong sense of possible downstream needs so I'm happy to defer to your judgement.

jmesnil: I'm fine with keeping a minor bump for that single change... If there were any other changes done to this subsystem before WildFLy 21 is released, then I'll make major bump at that time

PR #13404, opened 2020-07-08
WFLY-13655 LOG INFO level message if the Hibernate second level cache is enabled 

scottmarlow:
https://issues.redhat.com/browse/WFLY-13655

This is a minor change to show when the second cache is enabled.

### DIFF
--- a/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/JpaLogger.java
+++ b/jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/JpaLogger.java
@@ -17,8 +17,12 @@
 
 package org.jboss.as.jpa.hibernate5;
 
+import static org.jboss.logging.Logger.Level.INFO;
+
 import org.jboss.logging.BasicLogger;
 import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageLogger;
 
 /**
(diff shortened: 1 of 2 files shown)


kabir (approved):

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/JpaLogger.java
```diff
@@ -31,8 +35,17 @@
...
      */
     JpaLogger JPA_LOGGER = Logger.getMessageLogger(JpaLogger.class, "org.jipijapa");
 
+    /**
+     * Inform that the Hibernate second level cache is enabled.
+     *
+     * @param puUnitName the persistence unit name
+     */
+    @LogMessage(level = INFO)
+    @Message(id = 20253, value = "second level cache enabled for %s")
```
jamezp, line 48: This should be "Second", upper case S.

jamezp (changes requested):
I'm not really sure I understand the use case for this change. Reading the JIRA and associated JIRA's I didn't get out of it that we should be logging a message like this at a INFO message.

darranl:
@scottmarlow FYI this one is awaiting a response for the review from @jamezp

bstansberry:
@scottmarlow If you'd like this in, please adjust the capitalization. Thanks.

scottmarlow:
I somehow missed the feedback from last summer.  Correcting now.

jpa/hibernate5_3/src/main/java/org/jboss/as/jpa/hibernate5/JpaLogger.java
scottmarlow: Thanks James!

scottmarlow:
> @scottmarlow If you'd like this in, please adjust the capitalization. Thanks.

@bstansberry just did that, thanks!

scottmarlow:
> I'm not really sure I understand the use case for this change. Reading the JIRA and associated JIRA's I didn't get out of it that we should be logging a message like this at a INFO message.

@jamezp the idea is to make it more obvious when the second level cache is enabled, since the second level caching comes with a cost and applications may not know that it is enabled.

PR #13406, opened 2020-07-09
WFLY-13641 Do not try to index beans if module from jboss-deployment-…

manovotn:
…structure doesnt include META-INF.

JIRA - https://issues.redhat.com/browse/WFLY-13641

The processor in this PR goes over all dependencies and tries to scan for beans based on presence of `beans.xml` and the appropriate discovery mode. However, if a dependency is added via `jboss-deployment-structure`, it needs to have `meta-inf="import"` declared for CDI beans to be picked up. See https://docs.wildfly.org/17/Developer_Guide.html#CDI_Reference
This contract is currently not honored and you would see warning when Weld fails to index those classes.

This PR attempts to fix it. I am not 100% sure this is correct solution, e.g. if it couldn't be solved at some earlier point so that `ExternalBeanArchiveProcessor` wouldn't see that dependency at all. I have ran this against the issue reproducer and with basic WFLY tests (leaving full TS to PR checks).

Also, there is no test attached because the original issue reports warning being logged (and no hard failure) when this wasn't in place.

As usual, I am open to comments and suggestions :)

### DIFF
--- a/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDependencyProcessor.java
+++ b/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDependencyProcessor.java
@@ -93,7 +93,12 @@ public class TracingDependencyProcessor implements DeploymentUnitProcessor {
             moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, module, false, false, true, false));
         }
         for (String module : EXPORTED_MODULES) {
-            moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, module, false, true, true, false));
+            ModuleDependency modDep = new ModuleDependency(moduleLoader, module, false, true, true, false);
+            // io.opentracing.contrib.opentracing-interceptors needs to be processed by ExternalBeanArchiveProcessor
+            if (module.contains("opentracing-interceptors")) {
+                modDep.addImportFilter(s -> s.equals("META-INF"), true);
+            }
+            moduleSpecification.addSystemDependency(modDep);
         }
         for (String module : WildFlyTracerFactory.getModules()) {
             moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, module, true, false, true, false));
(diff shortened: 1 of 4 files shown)


bstansberry:
@ehsavoie FYI. What this is doing seems to have a negative effect on the Open Tracing integration (see the test failures.)

@jamezp FYI. I assume the logging test failures are because of a missing 'meta-inf' in things like https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/logging/config/LoggingConfigSharedTestCase.java#L141. I don't understand this issue enough to know if that's a flaw in the test artifact or in this fix.

jamezp:
@bstansberry @manovotn Yes that's the issue. The generated XML needs a `meta-inf="import"` attribute and value.
```
"           <module name=\"deployment." + EJB_DEPLOYMENT_NAME + "\" meta-inf=\"import\" />\n" +
```

manovotn:
@jamezp you're right, that fixes the logging tests (I will add it to this PR momentarily).

@ehsavoie I am not sure what's the issue with opentracing though. Both, TCKs and some integration tests fail, none of which use similar test setup. I assume opentracing relies on having a dependency added to the deployment and beans being picked up but does so in a manner that doesn't pass the check I added. The addition seem to be done inside `TracingDependencyProcessor`? 
There seems to be a way to add import/export filters to those programmatically added dependencies, but I don't know if it is a way to go really...

manovotn:
@ehsavoie I've added [third commit](https://github.com/wildfly/wildfly/pull/13406/commits/1422d6146786d0fe420057ed12fa091a5b052f21) that shows what we could do while adding opentracing deps in order to make it pass the new requirement in `ExternalBeanArchiveProcessor`. This (locally) passes the set of tests that originally failed.

I am not sure if other added dependencies need similar treatment, but TCKs don't seem to show any fails ATM. Let's see what full CI run thinks about that.

manovotn:
Apparently, the CI was still running with only the first commit I added. I have attempted to stop that execution and then force pushed here. Hope that will make TC re-run it with current state.

ehsavoie:
I think you fix is correct, the oother dependencies don't seem to bring in CDI beans so they should be skipped.

jamezp (approved):

bstansberry:
@emmartins Is there any kind of issue labeling system for possible migration issues?  So we can flag them for later follow-up. This seems like one; i.e. apps like the logging test case one that we've been discussing in the comments that may be relying on the broken behavior where a contract wasn't being enforced.

emmartins:
Nope, closest thing possible is to add Migration component in JIRA. In the
past I have used an aggregator JIRA to link all app migration issues but
think a label would make more sense.

jamezp:
+1 for a label. I created an `affects-migration` label. If anyone disagrees with that please let me know.

manovotn:
The JDK 11CI failure looks pretty unrelated: `javax.net.ssl.SSLException: Received fatal alert: internal_error`

ehsavoie (approved):

PR #13407, opened 2020-07-15
[WFLY-13601] JACC Not Correctly Activated Without Legacy Security.

darranl:
This PR is a set of changes required in WildFly to address issues where JACC and JASPI were not being activated correctly in a server provisioned without legacy security.  There will also be a PR against WildFly Core but that will depend upon these changes.

The primary Jira issue is https://issues.redhat.com/browse/WFLY-13601

This also includes the following issues:
https://issues.redhat.com/browse/WFLY-13648 - A new module to act as an optional intermediary.
https://issues.redhat.com/browse/WFLY-13610 - EJB DUPs were not checking if a deployment was theirs.
https://issues.redhat.com/browse/WFLY-13660 - Use capabilities in legacy security for JACC for more consistent DUP behaviour.
https://issues.redhat.com/browse/WFLY-13615 - Adjust DUPs to take into account the availability of JACC.

I did consider a more generic solution for the two security subsystems to negotiate which is responsible for JACC but TBH this is a temporary situation.  As legacy security is removed these will be revisited to remove the need for two different capabilities.

I am starting with the hold label as we will run some additional testsuites which are not a part of CI for some additional verification.


darranl:
Removing the hold label as other than the CI issue I think we will be better getting this in so the rest of the chain of PRs can follow.

darranl:
This will be showing with additional commits from https://github.com/wildfly/wildfly/pull/13419, this was to resolve a conflict this PR would have with the first but allow this one to get back into CI for a further run.

darranl:
I have just pushed again after merging with master to resolve today's merge conflict, I tried using -Dts.layers locally as that seemed most at risk from combining the changes but WeldJpaInjectionTestCase is failing for me against the jpa-layers.surefire execution as the bean is not being injected leading to a NPE.  I just double checked against master and I am seeing the same error there.

jmesnil:
@fjuma Could you please review?

darranl:
@fjuma This PR also currently contains another unmerged PR, the following comparison should be the real changes for this PR:

https://github.com/darranl/wildfly/compare/WFLY-13319_Base...darranl:WFLY-13601

darranl:
Actually GitHub for some reason is showing some changes not really related to the JACC PR so things like the EJBDefaultSecurityDomain deployer are not as relevant to this PR but the other commits are.

fjuma (approved):

jmesnil:
@darranl thanks

PR #13408, opened 2020-07-15
[WFLY-13671] Revert "[WFLY-13648] Temporarily mark 'org.jboss.as.security-integration` as an unreferenced module."

darranl:
https://issues.redhat.com/browse/WFLY-13671

This will be mergable after a WildFly Core release is available with https://github.com/wildfly/wildfly-core/pull/4267


bstansberry (approved):

bstansberry:
Note to reviewers: this looks like a big PR but actually all commits but the tiny last one are already in master.

PR #13410, opened 2020-07-17
[WFLY-13177] ManagedExecutorService: Wrong activeRequestCount at RequestController on RejectedExecutionException

istudens:
Issue: https://issues.redhat.com/browse/WFLY-13177


istudens:
@emmartins could you review this please? thanks

emmartins:
@istudens sure will do

istudens:
retest this please

istudens:
retest this please

emmartins:
Hi @istudens , just looked briefly but I believe there are few major issues with the PR:
1. the issue should apply to all wraps of tasks wrt the request controller, not just the submits, since those begin the request
2. you are converting all exceptions to rejected exceptions, when you should just catch any RejectedExecutionException *after* the begin request completes, in the submit/execute, and in such catch *complete the request* and rethrow the exception.

istudens:
@emmartins I've just added the same logic to the execute method. Could you point me to a code line related to your second comment please? The logic with converting all exceptions to rejected exceptions was there before my fix and it looks good to me. I've only added another case for rejected exception when ControlPoint returns RunResult.REJECTED. Then, ManagedExecutorServiceImpl contains a new logic which catches a rejected exception thrown by super.submit/execute and only in that case it calls controlPoint.requestCompelete which decreases active request counter. If a rejected exception is thrown from doWrapOrThrowException method itself, there is no need to decrease it.

istudens:
retest this please

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -67,6 +87,24 @@ public static Runnable doWrap(Runnable runnable, ControlPoint controlPoint) {
         }
     }
 
+    public static <T> Callable<T> doWrapOrThrowException(Callable<T> callable, ControlPoint controlPoint) {
+        if (controlPoint == null || callable == null) {
+            return callable;
+        }
+        try {
+            RunResult result = controlPoint.forceBeginRequest();
```
emmartins, line 95: Exceptions here should be throw, not be converted to RejectedExeception

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -67,6 +87,24 @@ public static Runnable doWrap(Runnable runnable, ControlPoint controlPoint) {
...
+            if (result == RunResult.REJECTED) {
+                ROOT_LOGGER.rejectedDueToMaxRequests();
```
emmartins, line 97: This is ok.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -67,6 +87,24 @@ public static Runnable doWrap(Runnable runnable, ControlPoint controlPoint) {
...
+            }
+            final ControlledCallable controlledCallable = new ControlledCallable(callable, controlPoint);
```
emmartins, line 99: Exceptions in this line and below should not be converted to RejectedException, and request must be completed.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -67,6 +87,24 @@ public static Runnable doWrap(Runnable runnable, ControlPoint controlPoint) {
         }
     }
 
+    public static <T> Callable<T> doWrapOrThrowException(Callable<T> callable, ControlPoint controlPoint) {
```
emmartins, line 90: instead of adding a new method you should fix the existent method, any usage of it has the issue

ee/src/main/java/org/jboss/as/ee/concurrent/ManagedExecutorServiceImpl.java
```diff
@@ -58,17 +60,35 @@ public ManagedExecutorServiceImpl(String name, ManagedThreadFactoryImpl managedT
 
     @Override
     public <T> Future<T> submit(Callable<T> task) {
-        return super.submit(doIdentityWrap(doWrap(task, controlPoint)));
+        final Callable<T> callable = doWrapOrThrowException(task, controlPoint);
+        try {
```
emmartins, line 64: this is not ok, any exception should complete the request

emmartins (changes requested):
Hi @istudens, here are the PR issues wrt 2.
Please note that you may have received additional requests to changes, now not found cause I deleted/edited some. While I don't agree with exceptions in the "wrap" turned into rejected exception, maybe there was a valid reason for such decision, and since that is not related with the original issue let's keep it that way.

istudens:
retest this please

darranl (dismissed):
Can you please change the text in the test case commit.

In five years time when someone looks at the history "a test case" doesn't reveal much, ideally a short sentence saying what the test does.

istudens:
My fault, I was pushing into another repo. Now it is up to date.
Thanks for your comments.

istudens:
retest this please

istudens:
retest this please

ee/src/main/java/org/jboss/as/ee/concurrent/ManagedExecutorServiceImpl.java
```diff
@@ -58,22 +59,46 @@ public ManagedExecutorServiceImpl(String name, ManagedThreadFactoryImpl managedT
...
+        }
     }
 
     @Override
     public void execute(Runnable command) {
-        super.execute(doIdentityWrap(doWrap(command, controlPoint)));
+        final Runnable runnable = doWrap(command, controlPoint);
+        try {
+            super.execute(doIdentityWrap(runnable));
+        } catch (RejectedExecutionException e) {
```
emmartins, line 98: should be catching Exception here too

emmartins (changes requested):

ee/src/main/java/org/jboss/as/ee/concurrent/ManagedExecutorServiceImpl.java
istudens: I am sorry I missed this.

emmartins:
retest this please

emmartins (approved):

bstansberry (approved):

bstansberry:
Thanks, @istudens!

PR #13411, opened 2020-07-17
[WFLY-13675] Make dependency on PicketBox optional for Batch subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-13675

### DIFF
--- a/batch-jberet/pom.xml
+++ b/batch-jberet/pom.xml
@@ -76,6 +76,10 @@
             <artifactId>jboss-msc</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.picketbox</groupId>
+            <artifactId>picketbox</artifactId>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>wildfly-weld-spi</artifactId>
         </dependency>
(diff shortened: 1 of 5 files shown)


batch-jberet/pom.xml
```diff
@@ -75,6 +75,10 @@
             <groupId>org.jboss.msc</groupId>
             <artifactId>jboss-msc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.picketbox</groupId>
+            <artifactId>picketbox</artifactId>
```
jamezp, line 80: Should this have `<optional>true</optional>`?

darranl: I suspect not as it will not compile without it - but maybe I should set the scope to provided to stop it leaking out.

jamezp: I think optional is still valid as it's optional at runtime. If the legacy security is desired, then the dependency is not optional. However it should work at runtime without it.

jamezp:
I'm not sure why those mixed domain tests are failing. They're passing for me locally so I kicked them off one more time.

darranl:
I think those test failures are unrelated, they have just popped up on a load of jobs.

batch-jberet/pom.xml
darranl: TBH here we are talking about class loading for a subsystem, by that point the pom definition has been dropped and we have moved to module definitions for class loading.  I will add scope provided to all the picketbox dependencies as I add them so we can avoid transitive dependencies creeping in.  I will add optional if you really want it but as we don't use maven dependencies at runtime I think that is redundant information.

jamezp: I've got no real strong opinion either way really. We do use `<optional>true</optional>` for the log tooling dependencies. I don't recall why we decided to do that though.

jamezp (approved):

PR #11717, opened 2018-10-05
[WFLY-11125] Convert weld subsystem module dependencies on security modules to optional

bstansberry:
https://issues.jboss.org/browse/WFLY-11125

PR #13412, opened 2020-07-17
[WFLY-13677] Make dependency on PicketBox optional for the Weld subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-13677

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldSecurityServices.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldSecurityServices.java
@@ -94,6 +94,10 @@ public class WeldSecurityServices implements Service, SecurityServices {
 
     @Override
     public org.jboss.weld.security.spi.SecurityContext getSecurityContext() {
+        if (securityManagerSupplier == null) {
+            return SecurityServices.super.getSecurityContext();
+        }
+
         SecurityContext ctx;
         if (WildFlySecurityManager.isChecking()) {
             ctx = AccessController.doPrivileged((PrivilegedAction<SecurityContext>) () -> SecurityContextAssociation.getSecurityContext());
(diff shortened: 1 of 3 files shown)


bstansberry:
@manovotn Please review.  I believe I did something along these lines for https://issues.redhat.com/browse/WFLY-11125 (see discussion on #11717) but in the end withdrew it just to be safe because it wasn't urgent and I didn't really understand the call flows.  @darranl FYI

weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldSecurityServices.java
```diff
@@ -94,6 +94,10 @@ public void cleanup() {
 
     @Override
     public org.jboss.weld.security.spi.SecurityContext getSecurityContext() {
+        if (securityManagerSupplier == null) {
```
manovotn, line 97: So there seem to be only one case when this can be invoked - from within `SecurityServices#getSecurityContextAssociator()` when `elytronDomation` is null. Can this even be a case? E.g. can you at any time have neither elytron nor picketbox on CP?

If so, then this is ok, I am just asking since my knowledge of security has the quality of a swiss chesse - it has lots of holes :)

darranl: It gets hard to achieve due to default security domain attributes on some subsystems but yes it should be possible that a deployment is deployed with no security activated so there could be a case legacy security is not available and no Elytron domain is associated either.

manovotn: Ok, thanks for quick clarification.

manovotn (approved):
LGTM, I think it should be safe to do it this way.

PR #13414, opened 2020-07-17
[WFLY-13085] Add documentation on the TLS 1.3 support for the OpenSSL TLS provider

fjuma:
https://issues.redhat.com/browse/WFLY-13085

Requires https://github.com/wildfly/wildfly-core/pull/4271 to be merged.

### DIFF
--- a/docs/src/main/asciidoc/_elytron/OpenSSL.adoc
+++ b/docs/src/main/asciidoc/_elytron/OpenSSL.adoc
@@ -28,7 +28,7 @@ that will use the OpenSSL TLS provider:
 reload
 ----
 
-== OpenSSL Library Location
+=== OpenSSL Library Location
 
 WildFly will search for the OpenSSL library using the standard system library search path. If you'd like to specify a
 custom location for the OpenSSL library, the `org.wildfly.openssl.path` property can be specified during WildFly startup.
@@ -39,3 +39,18 @@ If OpenSSL was loaded successfully, a message similar to the one below will occu
 ----
 15:37:59,814 INFO [org.wildfly.openssl.SSL] (MSC service thread 1-7) WFOPENSSL0002 OpenSSL Version OpenSSL 1.1.1d FIPS  10 Sep 2019
 ----
+
+=== TLSv1.3
+
+TLSv1.3 will be enabled by default when using OpenSSL 1.1.1 or higher and when running against JDK 11 or higher.
+OpenSSL will use the following TLSv1.3 cipher suites by default:
+
+[source,options="nowrap"]
+----
+TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
+----
+
+The `cipher-suite-names` attribute for an SSL context can also be used to configure the TLSv1.3 cipher suites
+you would like to use. See the sections on <<configuring-a-server-sslcontext,Configuring a server SSLContext>>
+and <<configuring-a-client-sslcontext,Configuring a client SSLContext>>
+for more details.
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/_elytron/OpenSSL.adoc
```diff
@@ -38,4 +38,19 @@ If OpenSSL was loaded successfully, a message similar to the one below will occu
...
+----
+
+=== TLSv1.3
+
+TLSv1.3 will be enabled by default when using OpenSSL 1.1.1 or higher and when running against JDK 11 or higher.
+OpenSSL will use the following TLSv1.3 cipher suites by default:
+
+[source,options="nowrap"]
+----
+TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256
```
jstourac, line 50: Just curious - I guess this will need some change into wildfly/wildfly-core code/configuration, right?

docs/src/main/asciidoc/_elytron/OpenSSL.adoc
```diff
@@ -38,4 +38,19 @@ If OpenSSL was loaded successfully, a message similar to the one below will occu
...
+----
+
+The `cipher-suite-names` attribute for an SSL context can also be used to configure the TLSv1.3 cipher suites
+you would like to use. See the sections on <<Using_the_Elytron_Subsystem.adoc#configuring-a-server-sslcontext,Configuring a server SSLContext>>
+and <<Using_the_Elytron_Subsystem.adoc#configuring-a-client-sslcontext,Configuring a client SSLContext>>
```
jstourac, line 55: When I generate this docs, the links don't seem to work - the `Using_the_Elytron_Subsystem.html` does not exist on my side. Is this expected at this moment or am I doing something wrong?

docs/src/main/asciidoc/_elytron/OpenSSL.adoc
fjuma: No additional changes will be needed for this. Support for the `cipher-suite-names` attribute for SSL contexts was already added as part of WFCORE-4172.

docs/src/main/asciidoc/_elytron/OpenSSL.adoc
fjuma: When viewing this file from GitHub, the link does work correctly. I haven't tried generating the docs locally though. What is the command you are using to do that?

docs/src/main/asciidoc/_elytron/OpenSSL.adoc
jstourac: +1 you are right, of course

docs/src/main/asciidoc/_elytron/OpenSSL.adoc
jstourac: You are right, GitHub preview works just fine. It's just HTML output that seems broken. I used following steps:
```
cd doc
mvn clean install
```
and looked on the html results in Firefox - when you naviage to the link in question, then the target link is not valid as it contains wrong filename, e.g.:
```
Using_the_Elytron_Subsystem.html#configuring-a-server-sslcontext
```
instead of:
```
WildFly_Elytron_Security.html#configuring-a-server-sslcontext
```
Maybe I just generate the doc in the wrong way.

fjuma: You're right, I just tried generating the docs and ran into the same issue. I've fixed it now. Thanks!

jstourac (approved):

jstourac (approved):

darranl:
@fjuma Do we really need a core update with the tests in core before this can be merged?

fjuma:
@darranl This can be merged before the tests are merged.

darranl (approved):

PR #13424, opened 2020-07-24
[WFLY-13676] Increase the test coverage of jpa and jpa-distribute gal…

yersan:
…leon layers

Jira issue: https://issues.redhat.com/browse/WFLY-13676

### DIFF
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -810,8 +810,8 @@
                                 <configuration>
                                     <!-- Tests to execute. -->
                                     <includes>
-                                        <include>org/jboss/as/test/clustering/cluster/jpa2lc/*TestCase.java</include>
-                                        <include>org/jboss/as/test/clustering/cluster/jpa2lcCustomRegion/*TestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/jpa/*TestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/jpa/custom/*TestCase.java</include>
                                     </includes>
                                     <!-- Parameters to test cases. -->
                                     <systemPropertyVariables>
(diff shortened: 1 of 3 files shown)


testsuite/integration/clustering/pom.xml
```diff
@@ -810,8 +810,8 @@
                                 <configuration>
                                     <!-- Tests to execute. -->
                                     <includes>
-                                        <include>org/jboss/as/test/clustering/cluster/jpa2lc/*TestCase.java</include>
-                                        <include>org/jboss/as/test/clustering/cluster/jpa2lcCustomRegion/*TestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/jpa/*TestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/jpa/custom/*TestCase.java</include>
```
bstansberry, line 814: I think you meant this to be an addition of more tests instead of a replacement.

testsuite/integration/clustering/pom.xml
```diff
@@ -986,8 +986,7 @@
                                                 <layer>jpa</layer>
                                             </excluded-layers>
                                             <layers>
-                                                <layer>cloud-server</layer>
-                                                <layer>web-clustering</layer>
+                                                <layer>jaxrs-server</layer>
                                                 <layer>h2-default-datasource</layer>
                                                 <layer>jpa-distributed</layer>
```
yersan, line 991: Notice I've also picked up a slimmer layer configuration for the jpa clustering tests. These tests are only using the hibernate container cache exposed by jaxrs resources, so we need neither `web-clustering` nor `cloud-server` to run it.

testsuite/integration/clustering/pom.xml
bstansberry: Ok; good. In e026da92aaf9fb5e4f36e123674c5b8ff89405f5 Paul changed the package.

testsuite/integration/clustering/pom.xml
bstansberry: OK.

bstansberry (approved):

PR #13425, opened 2020-07-24
[WFLY-13698] NullPointerException due to JSFDependencyProcessor addin…

moulalis:
Issue: https://issues.redhat.com/browse/WFLY-13698

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
@@ -64,15 +64,21 @@ public class JSFDependencyProcessor implements DeploymentUnitProcessor {
         final DeploymentUnit tl = deploymentUnit.getParent() == null ? deploymentUnit : deploymentUnit.getParent();
         final ModuleSpecification moduleSpecification = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
         final ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        //Set default when no default version has been set on the war file
+        String jsfVersion = JsfVersionMarker.getVersion(tl).equals(JsfVersionMarker.NONE)? JSFModuleIdFactory.getInstance().getDefaultSlot() : JsfVersionMarker.getVersion(tl);
+        String defaultJsfVersion = JSFModuleIdFactory.getInstance().getDefaultSlot();
+
         if(JsfVersionMarker.isJsfDisabled(deploymentUnit)) {
+            if (jsfVersion.equals(defaultJsfVersion) && !moduleIdFactory.isValidJSFSlot(jsfVersion)) {
+                throw JSFLogger.ROOT_LOGGER.invalidDefaultJSFImpl(defaultJsfVersion);
+            }
             addJSFAPI(JsfVersionMarker.JSF_2_0, moduleSpecification, moduleLoader);
             return;
         }
         if (!DeploymentTypeMarker.isType(DeploymentType.WAR, deploymentUnit) && !DeploymentTypeMarker.isType(DeploymentType.EAR, deploymentUnit)) {
             return;
         }
-        //Set default when no default version has been set on the war file
-        String jsfVersion = JsfVersionMarker.getVersion(tl).equals(JsfVersionMarker.NONE)? JSFModuleIdFactory.getInstance().getDefaultSlot() : JsfVersionMarker.getVersion(tl);
+
         if (jsfVersion.equals(JsfVersionMarker.WAR_BUNDLES_JSF_IMPL)) {
             //if JSF is provided by the application we leave it alone
             return;
@@ -80,7 +86,7 @@ public class JSFDependencyProcessor implements DeploymentUnitProcessor {
         //TODO: we should do that same check that is done in com.sun.faces.config.FacesInitializer
         //and only add the dependency if JSF is actually needed
 
-        String defaultJsfVersion = JSFModuleIdFactory.getInstance().getDefaultSlot();
+
         if (!moduleIdFactory.isValidJSFSlot(jsfVersion)) {
             JSFLogger.ROOT_LOGGER.unknownJSFVersion(jsfVersion, defaultJsfVersion);
             jsfVersion = defaultJsfVersion;
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
@@ -1,0 +1,86 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2020, Red Hat, Inc. and/or its affiliates, and individual
+ * contributors by the @authors tag. See the copyright.txt in the
+ * distribution for a full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.test.integration.jsf.deployment;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
+import org.jboss.arquillian.container.spi.client.container.DeploymentException;
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.jsf.logging.JSFLogger;
+import org.jboss.as.test.shared.ServerReload;
+import org.jboss.as.test.shared.SnapshotRestoreSetupTask;
+import org.jboss.dmr.ModelNode;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup(NonExistingJsfImplDeploymentTestCase.SetupTask.class)
+public class NonExistingJsfImplDeploymentTestCase {
+
+    @ContainerResource
+    private ManagementClient managementClient;
+
+    @ArquillianResource
+    private Deployer deployer;
+
+    @Deployment(testable = false, managed = false, name = "test_jar")
+    public static Archive<?> deploy() {
+        JavaArchive archive = ShrinkWrap.create(JavaArchive.class);
+        archive.addClass(NonExistingJsfImplDeploymentTestCase.class);
+        return archive;
+    }
+
+    private static final PathAddress JSF_ADDRESS = PathAddress.pathAddress(ModelDescriptionConstants.SUBSYSTEM, "jsf");
+
+    @Test
+    public void testDeploymentDoesntFailBecauseOfNPE() throws Exception {
+        try {
+            deployer.deploy("test_jar");
+            Assert.fail("Expected DeploymentException not thrown");
+        } catch (Exception e) {
+            Assert.assertTrue(e instanceof DeploymentException);
+            Assert.assertTrue(e.getMessage().contains(JSFLogger.ROOT_LOGGER.invalidDefaultJSFImpl("idontexist").getMessage()));
+        }
+    }
+
+    static class SetupTask extends SnapshotRestoreSetupTask {
+        @Override
+        protected void doSetup(ManagementClient client, String containerId) throws Exception {
+            ModelControllerClient mcc = client.getControllerClient();
+            ModelNode writeJSFAttributeOperation = Operations.createWriteAttributeOperation(JSF_ADDRESS.toModelNode(), "default-jsf-impl-slot", "idontexist");
+            ModelNode response = mcc.execute(writeJSFAttributeOperation);
+            Assert.assertEquals(SUCCESS, response.get(OUTCOME).asString());
+            ServerReload.executeReloadAndWaitForCompletion(client);
+        }
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -458,7 +458,6 @@
         <!-- Apache xml-resolver -->
         <version.xom>1.2.10</version.xom>
 
-
     </properties>
 
     <dependencyManagement>
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -412,6 +412,11 @@
             <artifactId>wildfly-jaxrs</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>wildfly-jsf</artifactId>
+            <scope>test</scope>
+        </dependency>
         <!-- To be able to start a 'remote' broker -->
         <dependency>
             <groupId>org.apache.activemq</groupId>


jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
```diff
@@ -107,12 +107,16 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
+            ModuleDependency jsfAPI = new ModuleDependency(moduleLoader, jsfModule, false, false, false, false);
+            moduleSpecification.addSystemDependency(jsfAPI);
+        }catch (NullPointerException exception){
+            throw new DeploymentUnitProcessingException("JSF module not found for JSF version "+jsfVersion);
+        }
 
-        ModuleIdentifier jsfModule = moduleIdFactory.getApiModId(jsfVersion);
-        ModuleDependency jsfAPI = new ModuleDependency(moduleLoader, jsfModule, false, false, false, false);
-        moduleSpecification.addSystemDependency(jsfAPI);
     }
```
bstansberry, line 120: I assume the NPE is coming from moduleSpecification.addSystemDependency(jsfAPI).

This addJSFAPI method is called from two places, L68 and L93. The L93 call has logic before it (L89-91) that checks for an invalid module slot and fails, which makes me think this is failing due to the L68 call. So does adding logic similar to L89-91 before the L68 call fix the problem?

moulalis: @bstansberry  It may fix. As per discussion in JBEAP-19759 we considered  throw an exception, the suspect was at 
L113 ModuleIdentifier jsfModule = moduleIdFactory.getApiModId(jsfVersion), not sure is it because of jsfVersion.  Currently we are not able to reproduce this issue but to be at safer side we thought handle exception. What do you suggest?

bstansberry: It should be possible to write a test that leads to L68 being called, which is then a reproducible failure that can be fixed.

Since we already have logic that is meant to handle a bad jsfVersion before it leads to this kind of failure it should be applied in both relevant cases.  One reason for that is that logic already has an i18n failure message, and we don't want to add another message for essentially the same problem.

Catching an NPE and rethrowing is generally not good, because the NPE means some var was unexpectedly null. Better to prevent reaching that state by understanding how the var was null and dealing with that.

moulalis: @bstansberry  I've wrote testcase which lead to call L68. Still not able to reproduce NPE, as it is passing default JSF version "Mojarra-2.0". There is one way of getting NPE if moduleSpecification object is null.  Please suggest.

bstansberry: AIUI, the value of JSFModuleIdFactory.defaultSlot is driven by the standalone.xml default-jsf-impl-slot attribute.

https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFSubsystemAdd.java#L61
https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFSubsystemAdd.java#L66
https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFVersionProcessor.java#L96

The bug is because the configuration has an invalid value for that attribute, so JSFModuleIdFactory.defaultSlot is a meaningless value.

This should result in ModuleIdFactory.getApiModId returning null, resulting in null getting passed into https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java#L133

That won't NPE in the code this PR is touching, but the bug report isn't about an NPE in that code. The bug report is about an NPE from org.jboss.modules.DependencySpec L637, and JSFDependencyProcessor.java#L133 passing null into the ModuleDependency constructor should result in DependencySpec L637 trying to process that null and an NPE.

ModuleIdFactory.getApiModId should return null for JSF version "Mojarra-2.0":

https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFModuleIdFactory.java#L170 calls 'computeSlot'.

computeSlot returns the (invalid) defaultSlot value:

https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFModuleIdFactory.java#L165

https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFModuleIdFactory.java#L170 does a map lookup in its 'apiIds' map using the invalid value, which will return null.  The 'apiIds' map is populated by scanning the module path so there's not going to be an entry in it for some invalid value.

moulalis: @bstansberry There is come thing wrong with this branch after I solved conflict. I will try to solve this tomorrow . If not then I will close this,  create new PR and tag this PR to it.   I have removed try-catch block and now code looks like below

```

@Override
    public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
        moduleIdFactory = JSFModuleIdFactory.getInstance();
        final DeploymentUnit tl = deploymentUnit.getParent() == null ? deploymentUnit : deploymentUnit.getParent();
        final ModuleSpecification moduleSpecification = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
        final ModuleLoader moduleLoader = Module.getBootModuleLoader();
        String jsfVersion = JsfVersionMarker.getVersion(tl).equals(JsfVersionMarker.NONE)? JSFModuleIdFactory.getInstance().getDefaultSlot() : JsfVersionMarker.getVersion(tl);
        String defaultJsfVersion = JSFModuleIdFactory.getInstance().getDefaultSlot();
        if(JsfVersionMarker.isJsfDisabled(deploymentUnit)) {
            if (jsfVersion.equals(defaultJsfVersion) && !moduleIdFactory.isValidJSFSlot(jsfVersion)) {
                throw JSFLogger.ROOT_LOGGER.invalidDefaultJSFImpl(defaultJsfVersion);
            }
            addJSFAPI(JsfVersionMarker.JSF_2_0, moduleSpecification, moduleLoader);
            return;
        }

```

darranl:
I think something has gone wrong with the rebase, this now shows as pulling in a load more commits but also have a conflict in a clustering pom which should not be affected by a JSF change.

moulalis:
> I think something has gone wrong with the rebase, this now shows as pulling in a load more commits but also have a conflict in a clustering pom which should not be affected by a JSF change.

@darranl  everything is fixed now.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
moulalis: @bstansberry I have added latest changes. Please review.

darranl:
Thank you - CI is now running again to see if we get a clean run.

darranl:
@bstansberry / @fjuma This has a clean CI run now, I think ready for your approval and we can merge.

jsf/subsystem/src/test/java/org/jboss/as/jsf/deployment/JSFDepProcessorTestCase.java
```diff
@@ -0,0 +1,57 @@
+package org.jboss.as.jsf.deployment;
```
spyrkob, line 1: @moulalis I think that should have a licence header

moulalis: @spyrkob I've added header.

fjuma:
@moulalis Looks ok. Have you verified that the newly added test throws an NPE without your fix?

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
```diff
@@ -56,31 +56,38 @@
...
         final ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        //Set default when no default version has been set on the war file
+        String jsfVersion = JsfVersionMarker.getVersion(tl).equals(JsfVersionMarker.NONE)? JSFModuleIdFactory.getInstance().getDefaultSlot() : JsfVersionMarker.getVersion(tl);
+        String defaultJsfVersion = JSFModuleIdFactory.getInstance().getDefaultSlot();
+
         if(JsfVersionMarker.isJsfDisabled(deploymentUnit)) {
             addJSFAPI(JsfVersionMarker.JSF_2_0, moduleSpecification, moduleLoader);
+            if (jsfVersion.equals(defaultJsfVersion) && !moduleIdFactory.isValidJSFSlot(jsfVersion)) {
+                throw JSFLogger.ROOT_LOGGER.invalidDefaultJSFImpl(defaultJsfVersion);
+            }
```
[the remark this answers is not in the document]
fjuma: Just FYI, this does appear above the addJSFAPI call when viewing the changes from the "Files changed" tab but appears below when viewing from the "Conversation" tab so might just be some sort of strange GitHub issue.

jsf/subsystem/src/test/java/org/jboss/as/jsf/deployment/JSFDepProcessorTestCase.java
```diff
@@ -0,0 +1,57 @@
...
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@PowerMockIgnore("jdk.internal.reflect.*")
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({JsfVersionMarker.class, JSFModuleIdFactory.class})
+public class JSFDepProcessorTestCase {
```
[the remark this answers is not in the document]
fjuma: I think it would be better to create a deployment that's going to fail there. The deployment will fail if JSF has been disabled for it. The method that's used for determining whether or not JSF should be activated for a deployment can be found here:

https://github.com/wildfly/wildfly/blob/master/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFVersionProcessor.java#L149-L164

From that method, we can see that if the deployment isn't a WAR then JSF should not be activated. So it looks like it should be possible to create a test that sets the `default-jsf-impl-slot` to some invalid value and then attempts to deploy an EAR.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
```diff
@@ -56,31 +56,38 @@
 
     private static final ModuleIdentifier JSF_SUBSYSTEM = ModuleIdentifier.create("org.jboss.as.jsf");
 
-    private JSFModuleIdFactory moduleIdFactory = JSFModuleIdFactory.getInstance();
+    private JSFModuleIdFactory moduleIdFactory;
 
     @Override
     public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+        moduleIdFactory = JSFModuleIdFactory.getInstance();
```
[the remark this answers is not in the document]
moulalis: @bstansberry  Yes. As it was unable to mock outside method.

jsf/subsystem/src/test/java/org/jboss/as/jsf/deployment/JSFDepProcessorTestCase.java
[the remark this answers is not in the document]
spyrkob: I think this can be tested in Arquillian with something like: https://github.com/spyrkob/wildfly/commit/9acf4ad3df71d5fcadebb1bafefacd271ef2c2c9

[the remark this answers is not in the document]
moulalis: @bstansberry I have tested Arquilian test suggested by @spyrkob, its working. The test case which I have written using Mockito is actually testing the code blocks of JSFDependencyProcessor.java . Please suggest if you are okay with the arquilian test then I'll move the changes.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFDependencyProcessor.java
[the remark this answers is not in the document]
bstansberry: Thanks, Farah. That's very strange.

jsf/subsystem/src/test/java/org/jboss/as/jsf/deployment/JSFDepProcessorTestCase.java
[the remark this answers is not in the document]
bstansberry: @moulalis Sorry; I've let this one sit idle. I prefer an arquillian integration test vs complex use of mocks. Thanks.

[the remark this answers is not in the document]
moulalis: @bstansberry I've added the changes with arquillian.

bstansberry:
@fjuma Please approve.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+    private Deployer deployer;
+
+    @Deployment(testable = false, managed = false, name = "test_jar")
+    public static Archive<?> deploy() {
+        JavaArchive archive = ShrinkWrap.create(JavaArchive.class);
+        archive.addClass(NonExistingJsfImplDeploymentTestCase.class);
+        return archive;
+    }
+
+    private static final PathAddress RA_ADDRESS = PathAddress.pathAddress(ModelDescriptionConstants.SUBSYSTEM, "jsf");
```
fjuma, line 63: s/RA_ADDRESS/JSF_ADDRESS

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+                throw e;
+            }
+        }
+    }
+
+    static class SetupTask extends SnapshotRestoreSetupTask {
+        @Override
+        protected void doSetup(ManagementClient client, String containerId) throws Exception {
+            ModelControllerClient mcc = client.getControllerClient();
+            ModelNode addRaOperation = Operations.createWriteAttributeOperation(RA_ADDRESS.toModelNode(), "default-jsf-impl-slot", "idontexist");
```
fjuma, line 89: Minor nit but `addRaOperation` should be reworded to something like `addJSFOperation`.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+            addRaOperation.get("archive").set("wf-ra-ely-security.rar");
+            addRaOperation.get("transaction-support").set("NoTransaction");
```
fjuma, line 91: Looks like the above two lines are a result of a copy/paste? Looks like they should be removed.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+        archive.addClass(NonExistingJsfImplDeploymentTestCase.class);
+        return archive;
+    }
+
+    private static final PathAddress RA_ADDRESS = PathAddress.pathAddress(ModelDescriptionConstants.SUBSYSTEM, "jsf");
+
+    @Test
+    public void testDeploymentDoesntFailBecauseOfNPE() throws Exception {
+        try {
+            deployer.deploy("test_jar");
```
fjuma, line 68: Since the test expects the deployment to fail, there should be an `Assert.fail` line under this one, e.g., something like:

`Assert.fail("Expected DeploymentException not thrown");`

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+        } catch (Exception e) {
```
fjuma, line 69: Should catch `DeploymentException` explicitly since that is what this test is expecting.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,97 @@
...
+                StringWriter out = new StringWriter();
+                e.printStackTrace(new PrintWriter(out));
+                String stackTrace = out.getBuffer().toString();
+
+                if (stackTrace.contains("Caused by: java.lang.NullPointerException")) {
+                    throw e;
+                }
+            } else {
+                throw e;
+            }
```
fjuma, line 81: This `if` block can be removed. If you want to check that the expected message occurs, you could use `e.getMessage()` and check if it contains the message from `invalidDefaultJSFImpl()`

fjuma (changes requested):

moulalis (commented):
@fjuma Please review.

fjuma (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/deployment/NonExistingJsfImplDeploymentTestCase.java
```diff
@@ -0,0 +1,85 @@
...
+        archive.addClass(NonExistingJsfImplDeploymentTestCase.class);
+        return archive;
+    }
+
+    private static final PathAddress JSF_ADDRESS = PathAddress.pathAddress(ModelDescriptionConstants.SUBSYSTEM, "jsf");
+
+    @Test
+    public void testDeploymentDoesntFailBecauseOfNPE() throws Exception {
+        try {
+            deployer.deploy("test_jar");
```
fjuma, line 68: Sorry, just realized that I don't see the `Assert.fail(...)` line added below `deployer.deploy(...)`, as mentioned before.

fjuma (changes requested):

fjuma (approved):

moulalis:
@bstansberry / @darranl could you please merge, as it is approved.

gaol:
@bstansberry would you please check if it is OK with you now? thanks :)

bstansberry:
Thanks, @moulalis !

PR #13428, opened 2020-07-26
[WFLY-13700] Add ability to configure stale transaction timeout for l…

tadamski:
https://issues.redhat.com/browse/WFLY-13700

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2011, Red Hat, Inc., and individual contributors
+ * Copyright 2020, Red Hat, Inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
@@ -281,6 +281,13 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
             .setAllowExpression(true)
             .setRequires(CommonAttributes.USE_JDBC_STORE).build();
 
+    public static final SimpleAttributeDefinition STALE_TRANSACTION_TIME = new SimpleAttributeDefinitionBuilder(CommonAttributes.STALE_TRANSACTION_TIME, ModelType.INT, true)
+            .setValidator(new IntRangeValidator(0))
+            .setMeasurementUnit(MeasurementUnit.SECONDS)
+            .setDefaultValue(new ModelNode().set(600))
+            .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
+            .setAllowExpression(true).build();
+
 
     private final boolean registerRuntimeOnly;
 
@@ -305,7 +312,7 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
             OBJECT_STORE_RELATIVE_TO, OBJECT_STORE_PATH, JTS, USE_HORNETQ_STORE_PARAM, USE_JOURNAL_STORE_PARAM, USE_JDBC_STORE, JDBC_STORE_DATASOURCE,
             JDBC_ACTION_STORE_DROP_TABLE, JDBC_ACTION_STORE_TABLE_PREFIX, JDBC_COMMUNICATION_STORE_DROP_TABLE,
             JDBC_COMMUNICATION_STORE_TABLE_PREFIX, JDBC_STATE_STORE_DROP_TABLE, JDBC_STATE_STORE_TABLE_PREFIX,
-            JOURNAL_STORE_ENABLE_ASYNC_IO, ENABLE_STATISTICS, HORNETQ_STORE_ENABLE_ASYNC_IO
+            JOURNAL_STORE_ENABLE_ASYNC_IO, ENABLE_STATISTICS, HORNETQ_STORE_ENABLE_ASYNC_IO, STALE_TRANSACTION_TIME
     };
 
     static final AttributeDefinition[] attributes_1_2 = new AttributeDefinition[] {USE_JDBC_STORE, JDBC_STORE_DATASOURCE,
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionTransformers.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionTransformers.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc.
+ * Copyright 2020 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import org.jboss.as.controller.transform.description.TransformationDescriptionBu
 import static org.jboss.as.txn.subsystem.TransactionExtension.CURRENT_MODEL_VERSION;
 import static org.jboss.as.txn.subsystem.TransactionSubsystemRootResourceDefinition.MAXIMUM_TIMEOUT;
 import static org.jboss.as.txn.subsystem.TransactionSubsystemRootResourceDefinition.OBJECT_STORE_RELATIVE_TO;
+import static org.jboss.as.txn.subsystem.TransactionSubsystemRootResourceDefinition.STALE_TRANSACTION_TIME;
 
 /**
  * @author Emmanuel Hugonnet (c) 2017 Red Hat, inc.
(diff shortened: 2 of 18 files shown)


transactions/src/main/resources/org/jboss/as/txn/subsystem/LocalDescriptions.properties
```diff
@@ -87,3 +87,5 @@ transactions.commit-markable-resource.jndi-name=JNDi name of this CMR resource
 transactions.commit-markable-resource.xid-location=Name of database table associated to this CMR resource
 transactions.commit-markable-resource.batch-size=Batch size for this CMR resource
 transactions.commit-markable-resource.immediate-cleanup=Immediate cleanup associated to this CMR resource
+
+transactions.stale-transaction-time=The time for which transaction is held in memory after it is being completed.
```
ochaloup, line 91: only for transactions which contains the remote enlistments (aka. which contains the EJB remote call). This should be emphasized in the description as for any other transaction (aka. the txn with DB) the transaction is removed imediatelly after it's committed.

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
```diff
@@ -0,0 +1,352 @@
...
+            <xs:element name="recovery-environment" type="recovery-environment" maxOccurs="1"/>
+            <xs:element name="coordinator-environment" type="coordinator-environment" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="object-store" type="object-store" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="jts" type="jts-Type" minOccurs="0" maxOccurs="1"/>
+            <xs:choice minOccurs="0" maxOccurs="1">
+                <xs:element name="use-journal-store" type="use-journal-store-Type" minOccurs="0" maxOccurs="1"/>
+                <xs:element name="jdbc-store" type="jdbc-store-Type" minOccurs="0" maxOccurs="1"/>
+            </xs:choice>
+            <xs:element name="commit-markable-resources" type="cmr-resources-Type" minOccurs="0" maxOccurs="1"></xs:element>
+            <xs:element name="stale-transaction-time" type="stale-transaction-time" minOccurs="0" maxOccurs="1"/>
```
ochaloup, line 55: have you been considering to create rather some "WFTC element" which could gathering all the WFTC setup?
Just thinking - until now the setup here was mostly about Narayana. This one is specific for WFTC - so just thinking.

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
```diff
@@ -0,0 +1,352 @@
...
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
+
+    <xs:complexType name="stale-transaction-time">
+        <xs:annotation>
+            <xs:documentation>
+                <![CDATA[
+                The time after which completed transactions are removed from the memory.
```
ochaloup, line 345: the  same comment as for description in cli - this is only for transactions with remote enlistments (remote EJB XAResource enlisted)

ochaloup (changes requested):
I would like ask for adjusting the description of the stale property and then I have one question about where the settings belong to in the subsystem.

transactions/src/main/resources/org/jboss/as/txn/subsystem/LocalDescriptions.properties
tadamski: fixed

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
tadamski: fixed

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
tadamski: fixed

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
```diff
@@ -0,0 +1,352 @@
...
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
+
+    <xs:complexType name="client">
+        <xs:annotation>
+            <xs:documentation>
+                <![CDATA[
+                The time after which completed transactions that contain remote enlistments are removed from the memory.
```
ochaloup, line 345: this annotation element should be part of the `<xs:attribte ...` element, or shouldn't?

ochaloup (changes requested):
Thanks for the changes. They addressed my points perfectly. I'm just adding one more as I think the documentation of the property should be inside of the attribute now.

transactions/src/main/resources/schema/wildfly-txn_6_0.xsd
tadamski: fixed

ochaloup (approved):
Seems fine to me. Thanks

bstansberry:
Transformation tests in TransactionSubsystemTestCase would be good.

tadamski:
@bstansberry I wasn't able to make transformation test work properly IMO because of the bug so I have created https://github.com/wildfly/wildfly-core/pull/4302.

bstansberry:
@tadamski https://issues.redhat.com/browse/WFCORE-5089 is fixed in the version of core WF master is using now. Thanks for that.

tadamski:
rebased

darranl (changes requested):
This PR is for a Feature Request but I don't see a corresponding RFE for the changes?

Also when bumping a subsystem version can you please separate the bump into a separate commit to the changes, it is not possible to use the GitHub UI to see things like the changes to the schema.

bstansberry:
@tadamski sorry -- i forgot that we were going to handle this without all the kanban process. 

Can you rebase this and ping me so we can get it in? Thanks.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -281,6 +281,13 @@
             .setAllowExpression(true)
             .setRequires(CommonAttributes.USE_JDBC_STORE).build();
 
+    public static final SimpleAttributeDefinition STALE_TRANSACTION_TIME = new SimpleAttributeDefinitionBuilder(CommonAttributes.STALE_TRANSACTION_TIME, ModelType.INT, true)
+            .setValidator(new IntRangeValidator(0))
+            .setMeasurementUnit(MeasurementUnit.SECONDS)
+            .setDefaultValue(new ModelNode().set(600))
+            .setFlags(AttributeAccess.Flag.RESTART_NONE)
```
bstansberry, line 288: This should be RESTART_ALL_SERVICES -- ReloadRequiredWriteAttributeHandler handles the writes.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionTransformers.java
```diff
@@ -39,6 +40,8 @@
     static final ModelVersion MODEL_VERSION_EAP70 = ModelVersion.create(3, 0);
     static final ModelVersion MODEL_VERSION_EAP71 = ModelVersion.create(4, 0);
     static final ModelVersion MODEL_VERSION_EAP72 = ModelVersion.create(5, 0);
+    static final ModelVersion MODEL_VERSION_5_1_0 = ModelVersion.create(5, 1);
+    static final ModelVersion MODEL_VERSION_EAP73 = ModelVersion.create(6, 0);
```
bstansberry, line 44: This would be MODEL_VERSION_EAP74

bstansberry (changes requested):
I made a couple comments re things I just noticed.

Also, please add coverage of this to TransactionSubsystemTestCase. Perhaps that would be adding the relevant config to full-expressions.xml and adding a full-6.0.0.xml file and a transformer test method that uses it.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
tadamski: fixed

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionTransformers.java
tadamski: fixed

tadamski:
hi @bstansberry, this PR should work fine now

bstansberry (approved):

mchoma:
> This PR is for a Feature Request but I don't see a corresponding RFE for the changes?

I agree here.

mchoma:
> @tadamski sorry -- i forgot that we were going to handle this without all the kanban process.

Why it was decided so? I see model changes in MR so I would expect that we need documentation and tests for it.

PR #13429, opened 2020-07-27
[WFLY-13703] Increase the test coverage of LayersTestCase for the all…

yersan:
… layers case

Jira issue: https://issues.redhat.com/browse/WFLY-13703

@bstansberry (cc @jfdenise  ) l Having this issue merged soon would be great to add the new layers into the current lists. Most of the layers included here were already being tested because one layer depends on another, but to me it has some utility see all of them in a single place, any new layer would be added to those list.


bstansberry (approved):

bstansberry:
Thanks, @yersan

PR #13235, opened 2020-04-22
[WFLY-13353][WFLY-13128] Add a Galleon layer for the POJO subsystem

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-13353
WildFly proposal: https://github.com/wildfly/wildfly-proposals/pull/309

It also amend a minor fix for JBeret Galleon layer community docs.

PR #13434, opened 2020-07-28
[WFLY-13128] Add a Galleon layer for batch-jberet

yersan:
This PR supersedes https://github.com/wildfly/wildfly/pull/13016

Jira issue: https://issues.redhat.com/browse/WFLY-13128

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
@@ -124,6 +124,7 @@ Basic layers:
 
 Basic layers:
 
+* _batch-jberet_: Tech preview support for Jakarta Batch.
 * _bean-validation_: Support for Jakarta Bean Validation. Optionally depends on _cdi_.
 * _cdi_: Support for CDI. Optionally depends on _bean-validation_.
 * _datasources_: Support for datasources.
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1304,6 +1304,46 @@
                                 </configuration>
                             </execution>
 
+                            <!-- Provision a cloud-profile server with batch-jberet as well -->
+                            <execution>
+                                <id>batch-jberet-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${project.build.directory}/wildfly-batch</install-dir>
+                                    <record-state>false</record-state>
+                                    <log-time>${galleon.log.time}</log-time>
+                                    <offline>true</offline>
+                                    <plugin-options>
+                                        <jboss-maven-dist/>
+                                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                                        <optional-packages>passive+</optional-packages>
+                                    </plugin-options>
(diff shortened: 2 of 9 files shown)


yersan:
Hi @chengfang , we do not plan to add more changes to this PR, would you mind taking a look when you have a chance and approve if you are fine with them? Thanks !

testsuite/integration/basic/pom.xml
```diff
@@ -1326,6 +1365,37 @@
...
+                                    <systemPropertyVariables>
+                                        <jboss.install.dir>${basedir}/target/wildfly-batch</jboss.install.dir>
+                                        <jboss.home>${project.build.directory}/wildfly-batch</jboss.home>
+                                        <jboss.home.dir>${project.build.directory}/wildfly-batch</jboss.home.dir>
+                                        <jbossas.dist>${project.build.directory}/wildfly-batch</jbossas.dist>
+                                        <jboss.dist>${project.build.directory}/wildfly-batch</jboss.dist>
+                                        <!-- Override the standard module path that points at the shared module set from dist -->
+                                        <module.path>${project.build.directory}/wildfly-batch/modules${path.separator}${basedir}/target/modules</module.path>
+                                    </systemPropertyVariables>
+                                    <environmentVariables>
```
chengfang, line 1385: system property `jboss.inst` is also used by some of [these tests](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/deployment/JobControlTestCase.java#L85), we may need to pass it too. But looks like these tests passed without it, probably because we ran them without security manager?

yersan: Thanks for reviewing Cheng. Yes, they it fails with the security manager. 

@bstansberry what is the position against whether we should run the layers tests under security manager? currently we do not have a CI with this configuration

bstansberry: That would require an entire separate CI job. I don't think it's worth it to add a per-PR job for that. Maybe a nightly job, but I'd be concerned it makes the set of controls of what's enabled overly complex. (Probably they already are!)  Something to look at in the future though.

yersan: Ok, I leave the sec manager concerns for now.

>  I'd be concerned it makes the set of controls of what's enabled overly complex. (Probably they already are!)

Yes, I think they are already.

yersan:
This PR is created on top of https://github.com/wildfly/wildfly/pull/13554

It includes a patch to fix intermittent failures of JDR report test case and adds on top the jberet Galleon layer

testsuite/integration/basic/pom.xml
yersan: Hello @chengfang Would you mind to approval the changes on this PR?

docs/src/main/asciidoc/_admin-guide/Galleon_provisioning.adoc
```diff
@@ -124,6 +124,7 @@ Basic layers:
 
 Basic layers:
 
+* _batch-jberet_: Tech preview support for Jakarta Batch.
```
yersan, line 127: @chengfang Just for your info, we will remove the "Tech preview" from this community documentation on this PR: https://github.com/wildfly/wildfly/pull/13235

It should not be here in the documentation.

chengfang (approved):

bstansberry (approved):

PR #13438, opened 2020-07-31
WFLY-13719 Add a system property to restore old behavior (Don't creat…

soul2zimate:
…e bindings from resource annotations if they're not used on a component)

Please see issue: https://issues.redhat.com/browse/WFLY-13719

An attempt to fix WFLY-13719 by adding a system property option (not sure that's the best name) to restore old behavior.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
@@ -63,6 +63,7 @@ import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 import java.lang.reflect.Modifier;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -86,6 +87,7 @@
 public class ModuleJndiBindingProcessor implements DeploymentUnitProcessor {
 
     private final boolean appclient;
+    private static final boolean IGNORE_UNUSED_RESOURCE_BINDING = Boolean.valueOf(WildFlySecurityManager.getPropertyPrivileged("jboss.ee.ignore-unused-resource-binding", "false"));
```
bstansberry, line 90: Not required but possibly useful would be to make this an instance var, not static, so we (e.g. in a test) or users (for ??? reason) can manipulate this behavior using a server reload.

bstansberry (commented):
I'm fine with this. I have a minor suggestion but won't be bothered at all if you prefer this as is.

bstansberry:
@chengfang or @tadamski Please review

soul2zimate:
Thanks, I have changed it to instance variable and added a test for the system property.

chengfang (approved):

bstansberry (approved):

PR #13417, opened 2020-07-21
[WFLY-13683] Remove microprofile-metrics dependency from jpa, jpa-dis…

yersan:
…tributed and web-clustering Galleon layers

The dependency of microprofile-metrics in jpa,jpa-distributed and web-clustering was initially added in an early version of the Infinispan 10.1.x upgrade. However, this initial requirement was removed and infinispan subsystem no longer requires metrics. 

Jira issue: https://issues.redhat.com/browse/WFLY-13683

PR #13440, opened 2020-07-31
[WFLY-13676] Fix WebJPATestCase execution when using jpa Galleon layer

yersan:
This PR follows up on https://github.com/wildfly/wildfly/pull/13424

Recently we merged https://github.com/wildfly/wildfly/pull/13417 which effectively removed metrics from jpa layers. This removal affects to the jpa Galleon layer execution that tests WebJPATestCase

Jira issue:  https://issues.redhat.com/browse/WFLY-13676

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1183,7 +1183,7 @@
                                             <name>standalone.xml</name>
                                             <layers>
                                                 <layer>jpa</layer>
-                                                <layer>datasources-web-server</layer>
+                                                <layer>jaxrs-server</layer>
                                                 <layer>h2-default-datasource</layer>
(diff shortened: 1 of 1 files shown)


bstansberry:
@yersan @scottmarlow The deployment in WebJPATestCase is not attempting to use BV, so it shouldn't be necessary for BV to be installed for the test to work.

Perhaps #13444 will fix it.

bstansberry:
@yersan While we discuss this I disabled the test via #13445.

yersan:
Right @bstansberry , I did not check `WebJPATestCase` when I was fixing this, I incorrectly assumed the reason of the error. 

Closing this in favor of https://github.com/wildfly/wildfly/pull/13444

PR #13442, opened 2020-07-31
[WFLY-13717] Add expressions for the host/port attributes for the out…

yersan:
…bound socket binding used for mail

Jira issue: https://issues.redhat.com/browse/WFLY-13717


bstansberry:
@yersan Adding entries to https://github.com/wildfly/wildfly/blob/master/dist/src/test/java/org/wildfly/dist/subsystem/xml/StandardConfigsXMLValidationUnitTestCase.java#L207 will fix the test failures.

bstansberry:
Thanks, @yersan

PR #13450, opened 2020-08-03
[WFLY-13730] Work around problem by hiding requirement relationship f…

bstansberry:
…rom Galleon

https://issues.redhat.com/browse/WFLY-13730


darranl:
If you create a follow up issue I can remove the dependency on the JACC service entirely meaning the resource will no longer need a capability reference.

bstansberry:
@chengfang @tadamski FYI in case this is causing you any problems with your layers work.

bstansberry:
@darranl I filed https://issues.redhat.com/browse/WFLY-13733.

darranl:
@jmesnil can you see about combining this with the core upgrade, I then need to follow up and delete the capability dependency entirely.

bstansberry:
This was included in #13467 13467

PR #13459, opened 2020-08-04
[WFLY-13741] Update the standalone-load-balancer configurations to depend on the legacy-management layer.

darranl:
https://issues.redhat.com/browse/WFLY-13741


bstansberry (approved):
I commented on the JIRA just to record some semi-relevant info, but the change itself LGTM.

PR #13463, opened 2020-08-06
[WFLY-13747] Add tests for the remoting layers, these require Undertow.

darranl:
https://issues.redhat.com/browse/WFLY-13747
https://issues.redhat.com/browse/WFLY-13755

It may be more appropriate to move these layers into WildFly, we can leave the subsystem where it is but the http-connector which we have chosen to use in our default configuration is only usable if Undertow is available.

### DIFF
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -2801,6 +2801,7 @@
                                                 <layer>jsonp</layer>
                                                 <layer>jsonb</layer>
                                                 <layer>legacy-management</layer>
+                                                <layer>legacy-remoting</layer>
                                                 <layer>legacy-security</layer>
                                                 <layer>logging</layer>
                                                 <layer>mail</layer>
@@ -2903,6 +2904,7 @@
                                                 <layer>jsonp</layer>
                                                 <layer>jsonb</layer>
                                                 <layer>legacy-management</layer>
+                                                <layer>legacy-remoting</layer>
                                                 <layer>legacy-security</layer>
                                                 <layer>logging</layer>
                                                 <layer>mail</layer>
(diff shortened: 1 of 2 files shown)


darranl:
Depends on https://github.com/wildfly/wildfly-core/pull/4291

yersan (changes requested):
Could you also include in this PR `legacy-remoting`  in `all-layers-provisioning-full ` /  `all-layers-jpa-distributed-provisioning-full` Galleon layers plugin execution?

darranl:
@yersan Done - looks to be passing locally.

yersan (approved):
Thanks Darran. Galleon tests pass in my local env too, so LGTM

jamezp:
@darranl Given your [comment](https://github.com/wildfly/wildfly/pull/13463#issuecomment-669934577)I added the core-upgrade-needed. If I misunderstood please feel free to remove it.

bstansberry:
I killed CI on this as it's a waste running all these jobs for something that only touches docs and testsuite/layers.

I'll start the Galleon one manually. If you want to test again in the future if things are busy just start that one.

testsuite/layers/pom.xml
```diff
@@ -2941,6 +2943,81 @@
...
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>legacy-remoting-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${layers.install.dir}/remoting</install-dir>
```
yersan, line 2991: This install dir duplicates the previous one, I would suggest rename it as 

`${layers.install.dir}/legacy-remoting`

yersan (changes requested):

testsuite/layers/pom.xml
darranl: FYI @yersan has the fix for this on a follow up PR with additional testing so I won't add a commit here for that one.

PR #13466, opened 2020-08-07
WFLY-13752 exclude element under includes and include element under excludes

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-13752

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -654,7 +654,6 @@
                                         <include>org/jboss/as/test/integration/ee/injection/support/jpa/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/ee/injection/resource/substitution/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/ejb/iiop/**/*TestCase*.java</include>
-                                        <exclude>org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java</exclude>
                                         <include>org/jboss/as/test/integration/*/security/**/*TestCase.java</include>
                                         <include>org/jboss/as/test/integration/ejb/**/mdb/**/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/ejb/log/*TestCase.java</include>
@@ -668,7 +667,6 @@
                                         <include>org/jboss/as/test/integration/jca/bootstrap/**/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/jmx/full/**/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/messaging/**/*TestCase*.java</include>
-                                        <include>org/jboss/as/test/integration/microprofile/opentracing/**/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/management/cli/**/*TestCase.java</include>
                                         <include>org/jboss/as/test/integration/naming/remote/**/*TestCase.java</include>
                                         <include>org/jboss/as/test/integration/ee/injection/resource/resourceref/*TestCase.java</include>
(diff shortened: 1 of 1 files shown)


testsuite/integration/basic/pom.xml
```diff
@@ -654,7 +654,7 @@
                                         <include>org/jboss/as/test/integration/ee/injection/support/jpa/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/ee/injection/resource/substitution/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/ejb/iiop/**/*TestCase*.java</include>
-                                        <exclude>org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java</exclude>
+                                        <include>org/jboss/as/test/integration/microprofile/opentracing/**/*TestCase*.java</include>
                                         <include>org/jboss/as/test/integration/*/security/**/*TestCase.java</include>
```
chengfang, line 658: Another possibility is, `<include>` might be a typo; it should be `<exclude>`. If that's the case, the fix is simply replacing `in` with `ex`.

The test `org/jboss/as/test/integration/microprofile/opentracing/**/*TestCase` is already included here. So this change will simply add a duplicate.

testsuite/integration/basic/pom.xml
```diff
@@ -731,7 +731,7 @@
                                         <exclude>org/jboss/as/test/integration/jca/bootstrap/**/*TestCase*.java</exclude>
                                         <exclude>org/jboss/as/test/integration/jmx/full/**/*TestCase*.java</exclude>
                                         <exclude>org/jboss/as/test/integration/messaging/**/*TestCase*.java</exclude>
-                                        <include>org/jboss/as/test/integration/microprofile/opentracing/**/*TestCase*.java</include>
+                                        <exclude>org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java</exclude>
                                         <exclude>org/jboss/as/test/integration/management/cli/**/*TestCase.java</exclude>
```
chengfang, line 735: Another possibility is, `<include>` might be a typo; it should be `<exclude>`. If that's the case, the fix is simply replacing `in` with `ex`.

The test `org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java` is already excluded here. So this change will simply add a duplicate.

testsuite/integration/basic/pom.xml
bstansberry: There are no tests in that ejb/**/jts package so both occurrences of the exclude can be removed.

```
$ git grep org.jboss.as.test.integration.ejb | grep jts
testsuite/integration/basic/pom.xml:                                        <exclude>org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java</exclude>
testsuite/integration/basic/pom.xml:                                        <exclude>org/jboss/as/test/integration/ejb/**/jts/**/*TestCase*.java</exclude>
```

If you annotate the file with git and look at the commit that added those two lines, the commit didn't add any tests in those packages either. So it looks like they were always cruft.

bstansberry: That git grep ^^^ was a search of the entire source tree in master.

bstansberry: There's also no package org.jboss.as.test.integration.microprofile in testsuite/integration/basic so those includes/excludes are not correct. The relevant tests are in org.wildfly.test.integration.microprofile.opentracing (wildfly, not jboss.as).

bstansberry: The basic idea here is if a test can pass running against plain standalone.xml, then it should not appear in either of the two sections this PR touches.  If it can only pass run against standalone-full.xml then it gets an "exclude" in basic-integration-default-web.surefire and an "include" in basic-integration-default-full.surefire.

parsharma: Then should we remove those exclude/include?

bstansberry: @parsharma Yes, they can be removed. The ejb/**/jts elements can just be removed as they don't have any relationship to any tests at all.

The org.jboss.as.test.integration.microprofile ones were meant to relate to tests, but they don't because the package was wrong. But that means the real tests are getting executed successfully in the basic-integration-default-web.surefire. And that means they don't need to be executed in the basic-integration-default-full.surefire execution, which is for tests that can't pass in basic-integration-default-web.surefire because the tests need things  in the standalone-full.xml config.  These tests don't, since they pass, so no special elements for them are needed.

bstansberry (changes requested):
@parsharma These are meaningless and should be removed as well:

https://github.com/wildfly/wildfly/blob/5ecacf8ffb313c09bec1d8fc41f482acb94e04e6/testsuite/integration/basic/pom.xml#L670

https://github.com/wildfly/wildfly/blob/5ecacf8ffb313c09bec1d8fc41f482acb94e04e6/testsuite/integration/basic/pom.xml#L722

bstansberry (approved):

bstansberry:
Thanks, @parsharma

PR #13476, opened 2020-08-11
[WFLY-13749] Add the bean-validation Galleon layer as an optional dep…

yersan:
…endency for jpa and jpa-distributed Galleon layers

Bean validation is not provisioned by default by the jpa / jpa-distributed Galleon layers. Users have to add this layer to their provisioning configuration to make Bean Validation work out of the box with a JPA deployment. This patch makes the layers be more according with the default use case. See Jira issue description to get more context about the motivation of this change.

Jira issue: https://issues.redhat.com/browse/WFLY-13749

### DIFF
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/jpa-distributed/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/jpa-distributed/layer-spec.xml
@@ -2,6 +2,7 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="jpa-distributed">
     <dependencies>
         <layer name="datasources"/>
+        <layer name="bean-validation" optional="true"/>
     </dependencies>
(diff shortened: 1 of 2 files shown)


bstansberry:
@scottmarlow Please review. The technical approach is fine; I'm mostly asking for an ok on what this achieves: provisioning BV by default if the user asks for jpa, with the option to exclude BV if not wanted. It's ok with me.

scottmarlow:
Sorry for the delay, this looks fine and I agree with the jira description:
```
-  If the user excludes the bean-validation Galleon layer from jpa / jpa-distributed Galleon layers, the deployment will succeed but bean validation won't work in the application code, because by default JPA uses ValidationMode.AUTO

-  If the user excludes the bean-validation Galleon layer from jpa / jpa-distributed Galleon layers, and configures JPA with ValidationMode.CALLBACK, the deployment will fail because the persistent provider will thrown an exception.

```

scottmarlow (approved):

PR #13478, opened 2020-08-12
[WFLY-12765]:Webservice deployment with undertow application security…

jimma:
… domain fals

https://issues.redhat.com/browse/WFLY-12765

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/service/EndpointService.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/service/EndpointService.java
@@ -71,6 +71,8 @@ import org.jboss.wsf.spi.deployment.Endpoint;
 import org.jboss.wsf.spi.deployment.EndpointType;
 import org.jboss.wsf.spi.management.EndpointMetricsFactory;
 import org.jboss.wsf.spi.security.EJBMethodSecurityAttributeProvider;
+import org.wildfly.extension.undertow.Capabilities;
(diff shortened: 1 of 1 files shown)


jimma:
@darranl  Can you please review ?

darranl (approved):
I think this will work but it will only be a temporary fix.

We should not be sharing the application security domain resources across subsystems like this but I see it has been that way for a while.

We need to reach the point where any subsystem which needs the SecurityDomain within a DeploymentUnitProcessor obtains it from the following attachment:
  https://github.com/wildfly/wildfly-core/blob/master/server/src/main/java/org/jboss/as/server/security/SecurityMetaData.java

I don't believe it will be too far away that we make the application security domain resources optional, and then in the future remove them entirely.  This fix wont be ready for those kinds of changes.

Also I suspect in the not too distant future we will be looking more at deployment defined security policies which again will be mapped to the SecurityMetaData attachment.

I am adding an approval for this one as I think it does solve the immediate problem but can you please create a follow up Jira to migrate the WS DUPs to use SecurityMetaData so all these references into the Undertow and EJB subsystems can be dropped.

jimma:
@darranl  created https://issues.redhat.com/browse/WFLY-13765 to follow up. Thanks.

jimma:
@bstansberry  There are couple of NPE failures. I'll have a look tomorrow.

bstansberry:
Thanks, @jimma

PR #13487, opened 2020-08-17
[WFLY-13150] Create a Galleon layer for distributable web

yersan:
Galleon layer to provide the distributable web subsystem configured with a local web container cache.

Jira issue: https://issues.redhat.com/browse/WFLY-13150
WildFly proposal: https://github.com/wildfly/wildfly-proposals/pull/340

### DIFF
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-passivation/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-passivation/layer-spec.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" ?>
-<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="distributable-web">
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="web-passivation">
     <dependencies>
         <layer name="web-server"/>
         <layer name="transactions"/>
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -618,8 +618,30 @@
                                     <includes>
                                         <include>org/jboss/as/test/clustering/single/**/*TestCase.java</include>
                                     </includes>
+                                    <excludes>
+                                        <exclude>org/jboss/as/test/clustering/single/web/NonHaWebSessionPersistenceTestCase.java</exclude>
+                                    </excludes>
                                     <systemPropertyVariables combine.children="append">
                                         <arquillian.launch>single</arquillian.launch>
+                                        <jboss.server.config.file.name>standalone.xml</jboss.server.config.file.name>
+                                    </systemPropertyVariables>
+                                </configuration>
+                            </execution>
+                            <!-- Single node clustering tests for containers with manual lifecycle. -->
+                            <execution>
+                                <id>ts.surefire.clustering.single.manual</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <runOrder>alphabetical</runOrder>
+                                    <includes>
+                                        <include>org/jboss/as/test/clustering/single/web/NonHaWebSessionPersistenceTestCase.java</include>
+                                    </includes>
+                                    <systemPropertyVariables combine.children="append">
+                                        <wildfly1>wildfly-1</wildfly1>
+                                        <arquillian.launch>node-non-ha</arquillian.launch>
                                         <jboss.server.config.file.name>standalone.xml</jboss.server.config.file.name>
                                     </systemPropertyVariables>
                                 </configuration>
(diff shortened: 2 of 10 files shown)


testsuite/integration/clustering/pom.xml
```diff
@@ -832,6 +832,42 @@
...
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <!-- Tests to execute. -->
+                                    <includes>
+                                        <include>org/jboss/as/test/clustering/cluster/web/**/CoarseSessionPassivationTestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/web/**/FineSessionPassivationTestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/web/**/CoarseTransactionalSessionActivationTestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/web/**/NonDistributableTestCase.java</include>
+                                        <include>org/jboss/as/test/clustering/cluster/web/**/FineTransactionalSessionActivationTestCase.java</include>
```
yersan, line 850: Hello @pferraro @rhusar ,

Could you point me out to any missing test that should be included for this Galleon layer? The above list are the tests I found suitable to work with this kind of configuration but I want to be sure I'm not missing anything, I'm not very familiar with the clustering test suite. Thanks!

pferraro: Everything within org/jboss/as/test/clustering/cluster/web/** should be testable with this profile.

yersan: @pferraro Notice that the web-passivation Galleon layer don't provision `jgroups` and configures a local infinispan web container cache. My understanding is the web session is not shared across all the cluster members, most of the tests under org/jboss/as/test/clustering/cluster/web/** assert values for all the cluster nodes.

yersan: @pferraro @rhusar 
I've adapted the code to use the single arquillian launch. Since the web-passivation Galleon layer uses a local infinispan web container cache, jgroups is not provisioned, that means, we cannot use any of the available tests that requires a cluster.

That made me to identify only one test available for a single node configuration [SimpleWebTestCase](https://github.com/wildfly/wildfly/blob/master/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/web/SimpleWebTestCase.java). Could you let me know if that test would be enough to test the capabilities giving by the web-passivation Galleon layer or, by the contrast, which other tests available on the clustering test suite are suitable to test this layer?

pferraro: This test should be used to validate this configuration: https://github.com/wildfly/wildfly/blob/master/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/NonHaWebSessionPersistenceTestCase.java

yersan: ok, thank you Paul!

yersan:
Hello @pferraro @rhusar Could you be so kind and review this feature so we can move it to Ready To Merge? Thanks!

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/web/NonHaWebSessionPersistenceTestCase.java
```diff
@@ -99,8 +103,11 @@ public void testSessionPersistence(@ArquillianResource(SimpleServlet.class) @Ope
             Assert.assertFalse(Boolean.valueOf(response.getFirstHeader("serialized").getValue()));
             response.getEntity().getContent().close();
 
-            stop(CONTAINER_SINGLE);
-            start(CONTAINER_SINGLE);
+            NodeUtil.stop(this.controller, CONTAINER_SINGLE);
+            NodeUtil.start(this.controller, CONTAINER_SINGLE);
+            if (Boolean.getBoolean("ts.bootable")) {
```
fabiobrz, line 108: Hi @yersan - I might be missing something here, but IIUC the deployment is operated by Arquillian only when `ts.bootable` is set. What happens for Galleon - i.e. when `ts.layers` is set?

yersan: Hi @fabiobrz This test requires a server restart to later validate the session information is correctly recovered. This test is used for a bootable JAR, Galleon trimmed or bare metal server. The deployment is always managed by the test itself, I mean Arquillian will not automatically deploy the application, we deploy it manually on @Before and @After methods, but the deployment survives to a server restart when we are not in the Bootable JAR scenario.

When we are testing the Bootable JAR we need to redeploy it after stopping the server, hence this `if`.

fabiobrz: Thanks @yersan you’re right, the jar must be redeployed to the hollow bootable jar server.

rhusar (approved):
LGTM

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/web-passivation/layer-spec.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" ?>
-<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="distributable-web">
```
pferraro, line 2: FYI - this should use the distributable-web-local feature-group, not distributable-web.
Currently, that feature-group is buggy, as it specifies parameters that are not valid.
See https://github.com/wildfly/wildfly/pull/13818/commits/4aaa6d224e3705f80fcfb73bb4811e9f97b49ca1#diff-fae6397e0df9fed6b193c54885a6b257568d2a361e472a17c55566afd18aa256L18

pferraro (changes requested):

pferraro (approved):
After discussing with Brian, we can fix the distributable-web-local stuff after this is merged.

PR #13489, opened 2020-08-17
WFLY-11808 Unable to do jndi lookup when starting batch job from web …

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-11808

Add support for jndi lookup in a batch job when started from Admin Console or CLI, by using the namespace context selector from the deployment.

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentService.java
@@ -31,6 +31,7 @@ import org.jberet.spi.BatchEnvironment;
 import org.jberet.spi.JobExecutor;
 import org.jberet.spi.JobTask;
 import org.jberet.spi.JobXmlResolver;
+import org.jboss.as.naming.context.NamespaceContextSelector;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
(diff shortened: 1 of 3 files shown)


batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentService.java
```diff
@@ -61,14 +62,20 @@
...
-    public BatchEnvironmentService(final ClassLoader classLoader, final JobXmlResolver jobXmlResolver, final String deploymentName, final boolean legacySecurityPresent) {
+    public BatchEnvironmentService(final ClassLoader classLoader,
+                                   final JobXmlResolver jobXmlResolver,
+                                   final String deploymentName,
+                                   final EEModuleDescription eeModuleDescription,
+                                   final boolean legacySecurityPresent) {
         this.classLoader = classLoader;
         this.jobXmlResolver = jobXmlResolver;
         this.deploymentName = deploymentName;
+        this.eeModuleDescription = eeModuleDescription;
```
jamezp, line 78: I think we should probably just pass the `NamespaceContextSelector` rather than the `EEModuleDescription`. It seems like we shouldn't be holding a reference to that in a service.

jamezp (changes requested):
Minus the one comment this looks okay.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentService.java
chengfang: Changed to pass `NamespaceContextSelector` instead of `EEModuleDescription`.

chengfang:
The test `org.jboss.as.test.integration.deployment.excludesubsystem.ExcludeEESubsystemTestCase#testDeploy` failed due to NPE. It is because when EE subsystem is excluded (as in this test), `eeModuleDescription` from the attachment is not set. So will need to check null before calling `eeModuleDescription.getNamespaceContextSelector()`.

jamezp (approved):

michpetrov:
@chengfang I don't see errors with the quickstart, however I still see a lookup failing:
 - https://github.com/jberet/jsr352/blob/d3544af2218f5916ac945b095814186a1ef8e614/jberet-core/src/main/java/org/jberet/operations/AbstractJobOperator.java#L141

Here the `applicationName` is null, the context has not been switched yet. If that's not an issue then I don't have further comments.

chengfang:
> @chengfang I don't see errors with the quickstart, however I still see a lookup failing:
> 
>     * https://github.com/jberet/jsr352/blob/d3544af2218f5916ac945b095814186a1ef8e614/jberet-core/src/main/java/org/jberet/operations/AbstractJobOperator.java#L141
> 
> 
> Here the `applicationName` is null, the context has not been switched yet. If that's not an issue then I don't have further comments.

`applicationName` will always be null when running in non-EE environment. Since this job execution is started from a management operation, not as part of the applicatio invocation, so a `null` `applicationName` is okay. In the future, we can improve jberet-core's `AbstractJobRepository` to not rely on jndi lookup; instead it can retrieve this value from `BatchEnvironment`.

PR #13495, opened 2020-08-20
WFLY-13786 DistributableSessionManager.getSession(...) should not propagate exceptions

pferraro:
https://issues.redhat.com/browse/WFLY-13786

This should resolve some intermittent failures in clustering testsuite.


pferraro:
@bstansberry That failure isn't related - but I've seen that one fail too often.  I'll look into stabilizing it this week.

bstansberry (approved):

PR #13500, opened 2020-08-22
WFLY-13426 Convert clustering modules to use ProtoStream marshalling

pferraro:
https://issues.redhat.com/browse/WFLY-13426

N.B.  This only affects marshalling of internal objects.  Support for ProtoStream marshalling of user objects is tracked by WFLY-13077.


clustering/server/src/main/java/org/wildfly/clustering/server/group/JGroupsAddressSerializer.java
```diff
@@ -1,6 +1,10 @@
 /*
  * JBoss, Home of Professional Open Source.
+<<<<<<< HEAD
  * Copyright 2019, Red Hat, Inc., and individual contributors
+=======
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+>>>>>>> 9aad2a3ba7... ProtoStream support for wildfly-clustering-server module
```
bstansberry, line 7: Unresolved conflict

pferraro: Fixed.

bstansberry:
@pferraro merge conflict now

clustering/marshalling/protostream/src/main/resources/org.wildfly.clustering.marshalling.protostream.proto
```diff
@@ -0,0 +1,29 @@
+package org.wildfly.clustering.marshalling.protostream;
+
+import "java.lang.proto";
+
+/**
+ * @TypeId(0)
```
rhusar, line 6: @pferraro Should we be using these IDs or is the Javadoc not specific enough when ti says 'globally'? Looking at some of the documentation:

> This numeric identifier must be globally unique so it can be used to identify the type instead of the fully qualified name.

> Values in the range 0..65535 (inclusive) are reserved for internal use by Protostream and other projects from the Infinispan organisation and should not be used by application developers.

clustering/server/src/test/java/org/wildfly/clustering/server/ServiceLoaderTestCase.java
```diff
@@ -51,10 +51,11 @@ public void load() {
         load(DistributedCacheServiceConfiguratorProvider.class);
         load(LocalGroupServiceConfiguratorProvider.class);
         load(LocalCacheServiceConfiguratorProvider.class);
+        load(SerializationContextInitializer.class);
     }
 
     private static <T> void load(Class<T> targetClass) {
         ServiceLoader.load(targetClass, ServiceLoaderTestCase.class.getClassLoader())
-                .forEach(object -> LOGGER.trace("\t" + object.getClass().getName()));
+                .forEach(object -> System.out.println("\t" + object.getClass().getName()));
```
rhusar, line 59: May want to get rid of this System.out running loose here in this test case.

clustering/web/hotrod/src/main/resources/org.wildfly.clustering.web.hotrod.session.proto
```diff
@@ -0,0 +1,15 @@
+package org.wildfly.clustering.web.hotrod.session;
+
+/**
+ * @TypeId(210)
```
rhusar, line 4: This doesn't seem right – the TypeId is the same as in `src/main/resources/org.wildfly.clustering.web.infinispan.session.proto` yet the message type is differnet. Is that intentional?

pferraro: Yes, this is intentional.  These types are mutually exclusive.

clustering/marshalling/protostream/src/main/resources/org.wildfly.clustering.marshalling.protostream.proto
pferraro: We do not reuse any of Infinispan's types, thus the ID space is our own.

clustering/server/src/test/java/org/wildfly/clustering/server/ServiceLoaderTestCase.java
pferraro: Yeah...

rhusar (approved):
Looks good to me, I don't see any other issues. 👍

PR #13503, opened 2020-08-25
[WFLY-13212] Boot messages related to activation of MP subsystems are…

baranowb:
… inconsistent

https://issues.redhat.com/browse/WFLY-13212

### DIFF
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/_private/MicroProfileConfigLogger.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/_private/MicroProfileConfigLogger.java
@@ -22,7 +22,7 @@
 
 package org.wildfly.extension.microprofile.config.smallrye._private;
 
-import static org.jboss.logging.Logger.Level.INFO;
+import static org.jboss.logging.Logger.Level.DEBUG;
 
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.logging.BasicLogger;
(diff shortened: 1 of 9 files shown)


bstansberry:
@maeste WDYT?

The change to DEBUG may have been inspired by my comment on the issue. I'm fine with it, but if you or the relevant individual component leads disagree, that's fine. We already provide a random mix of such messages at INFO so I have no strong case for my personal preference. I do think the MP subsystems should be consistent though.

I think you may know better whether "Eclipse" or "WildFly" or nothing in the various messages is best. My impression is the first use of "MicroProfile" should preface it with "Eclipse' and then the 'Eclipse' can be dropped, but i learned that more in the context of docs and articles, which is different from log messages. Clearly we wouldn't add logic to figure out "who's first to log".

kabir:
@bstansberry @maeste I think the MP ones should be at INFO level because:

* In EAP they come in via the XP so they may not always be available.
* Additionally the reactive ones in my https://github.com/wildfly/wildfly/pull/13783 are not enabled by default.

So having them at INFO level IMO makes it clearer whether or not they are enabled.

Following the above logic it could be argued that now that we have Galleon provisioning/slimming of the server, we should make all subsystem start messages INFO level but that is wider discussion than this PR.

bstansberry:
@kabir OK.

maeste:
@bstansberry @kabir I agree they should be at INFO level

regarding the name, Wildfly Microprofile should be fine. It's our implementation/integration, so the name should be fine, even if including the prefix Eclipse would be even better: for version >=4.0 the official name of the specs is Eclipse MicroProfile

baranowb:
So should this one be purged or changed?
@kabir @bstansberry  ^^

bstansberry:
@baranowb I'll defer to @kabir and @maeste on this PR but personally, not having thought about it hard, I don't see any reason to not just update this branch.

rhusar (changes requested):
-1

We do *not* put 'WildFly' in log messages. The MP Config/JWT (and VaultLogger) appear to be only places where this slipped in.

bstansberry:
Per https://github.com/eclipse/microprofile-marketing/issues/451#issuecomment-797656178 it seems that adding the 'Eclipse' before MicroProfile is no longer appropriate.

We should just get rid of the qualifiers, both Eclipse and WildFly. The latter is kind of confusing since we don't do that for Jakarta EE, plus it interferes with rebranding for EAP.

kabir:
@bstansberry I have updated the PR to remove all WildFly/Eclipse prefixes from the MicroProfile log messages. Additionally I moved the MP subsystem start messages to INFO.

(@baranowb - Since I am an admin, I was able to force push to your branch in case you wonder how I updated this PR)

kabir:
@bstansberry This should be good to go now

PR #13504, opened 2020-08-26
[WFLY-13188] Defer vendor metric registration until completion of boot

bstansberry:
https://issues.redhat.com/browse/WFLY-13188

### DIFF
--- a/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricCollector.java
+++ b/microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricCollector.java
@@ -154,7 +154,7 @@ public class MetricCollector {
             MetricID metricID = new MetricID(metricMetadata.metricName, tags);
 
             registration.addRegistrationTask(() -> registerMetric(metricMetadata, resourceAddress, attributeName, unit, attributeDescription, isCounter, tags));
-            registration.addUnregistrationTask(() -> MetricRegistries.get(MetricRegistry.Type.VENDOR).remove(metricID));
+            registration.addUnregistrationTask(metricID);
         }
 
         for (String type : current.getChildTypes()) {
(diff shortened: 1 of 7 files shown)


xstefank (approved):
I am no expert in this area. However, I don't see any issues with this PR.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricCollector.java
```diff
@@ -88,6 +94,26 @@ public MetricRegistration collectResourceMetrics(final Resource resource,
...
+                    registration.unregister();
+                    processStateNotifier.removePropertyChangeListener(this);
+                }
+
+            }
+        };
+        this.processStateNotifier.addPropertyChangeListener(listener);
+        // If server is already running, we won't get a change event so register now
+        if (ControlledProcessState.State.RUNNING == this.processStateNotifier.getCurrentState()) {
+            registration.register();
```
spyrkob, line 115: @bstansberry it's probably a stupid question, but is there a risk of registering the metric twice if the state flips to RUNNING after listener is added but before the condition?

bstansberry: @spyrkob No, there isn't, but it's probably better to code defensively against it by making the method synchronized.

It's not a stupid question. :)

This method is called in the context of a management op that modifying the system. The switch to RUNNING also occurs in the context of such a management op (a big composite op that runs during boot) and the switch to RUNNING doesn't happen until the code that would call this method has completed. The system won't allow concurrent execution of 2 such management ops.

spyrkob: Great, thanks for the explanation :)

spyrkob (approved):
As far as I can tell this looks good, but tbh I'm not very familiar with MP subsystem

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MetricCollector.java
bstansberry: @spyrkob I added the synchronized

PR #13506, opened 2020-08-27
[WFLY-13759] improvements for ManagedExecutorServiceMetricsTestCase

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-13759
Changes: replaces true with equals assertions, and retries stats reads up to 3 times (500ms retry wait period)


emmartins:
@bstansberry guess we can start like this, and tune the retries in the future if needed

bstansberry (approved):

PR #13516, opened 2020-08-31
[WFLY-13305] - Adjust testsuite for bootable jar

fabiobrz:
This PR is related to: https://issues.redhat.com/browse/WFLY-13305

Here we are adding a _Bootable JAR_ profile to each of those modules that already define a Galleon profile.

The original Galleon profile layers definition has been kept for bootable JAR packaged server instances and the same goes for involved `surefire` executions - apart from a few exceptions in `integration/basic`.

CC @jfdenise - can you please review?


fabiobrz:
retest this please

fabiobrz:
Hi @jamezp, I see that there are some failures which might be related to https://issues.redhat.com/browse/WFARQ-78. 
WDYT?
In it should be confirmed, is there any plan to have the Beta3 released?

jamezp:
> Hi @jamezp, I see that there are some failures which might be related to https://issues.redhat.com/browse/WFARQ-78.
> WDYT?
> In it should be confirmed, is there any plan to have the Beta3 released?

@fabiobrz yeah the issues are a combo of https://issues.redhat.com/browse/WFARQ-78 and https://issues.redhat.com/browse/WFARQ-79. I can release a Beta3 today.

jamezp:
@fabiobrz I've released 3.0.0.Beta3 which should fix the test issues seen.

fabiobrz:
Thanks @jamezp!

fabiobrz:
retest this please

jamezp:
retest this please

fabiobrz:
retest this please

fabiobrz:
Looks like current failures are not related to Bootable JAR profile.

fabiobrz:
@bstansberry I just rebased on the updated `master`. CI checks were green before than that, could you (or someone else) please review?

rhusar:
@fabiobrz @bstansberry @jfdenise If I am looking correctly, the existing CI runs *must* be ignored for now since none of them specify `ts.bootable` which means none of these tests are run. What's the reason for not tucking this along with `ts.layers`?

fabiobrz:
Thanks @rhusar - yep, no `ts.bootable` in CI for now I guess. We're lettng them run on our internal pipelines though. I thought this PR was just checked for being merged (i.e. according to the current CI status) and `ts.bootable` profile would be added later. But I might be missing something.

rhusar:
@fabiobrz There is a couple of ways to do this – but the most reasonable one is custom runs on the CI until a job is added on the CI. Though still not clear to me why another flag is needed.

fabiobrz:
> @fabiobrz There is a couple of ways to do this – but the most reasonable one is custom runs on the CI until a job is added on the CI. Though still not clear to me why another flag is needed.

Thanks for feedback @rhusar - regarding the additional flag, it had been used already in the core TS back when additional tests were added there by @jfdenise. Anyway to me it looks lile a good thing that the twos are being separated because, for instance, functional reasons (e.g.  layers config in bootable could be slightly different in some cases) and non-functional (e.g. tests duration). WDYT?

rhusar:
There are no functional reasons, it's really just an aspect of breaking tests into units controllable by properties, so in that sense more properties allow for better granularization. Looks to me the simplest would be to add that new flag to the 'galleon build' CI job as well and be done. All I am saying is so we don't merge something that hasn't been tested by the CI. But surely @bstansberry knows what works best here.

fabiobrz:
> All I am saying is so we don't merge something that hasn't been tested by the CI. But surely @bstansberry knows what works best here.

Makes sense @rhusar. Anyway, as you can see it fails CI checks even though the `ts.bootable` profile are not executed, so it won't be merged anyway, I guess :)

bstansberry:
The CI failures appear to just be job timeouts due to CI being overloaded last week. Later today I'll start those again.

fabiobrz:
Thanks @bstansberry! Please notice that I had to bump Wildfly Arquillian to 3.0.0.Beta3 because we need the bootable container adapter which was introduced with this version.

CC @jamezp

jfdenise:
@fabiobrz , https://github.com/wildfly/wildfly/compare/0f64b5332f5e6b7aa51597ab1abf06c4cd4d0adb..3990729b9c43279382e4d9614c20ca1a6497dd1a#diff-4b8d9b59b6ff474b31a2557a6ee42c72R766
This execution has been added to the bootable JAR profile. I must be in the default profile and a bootable JAR execution with sar should be added.

fabiobrz:
> @fabiobrz , https://github.com/wildfly/wildfly/compare/0f64b5332f5e6b7aa51597ab1abf06c4cd4d0adb..3990729b9c43279382e4d9614c20ca1a6497dd1a#diff-4b8d9b59b6ff474b31a2557a6ee42c72R766
> This execution has been added to the bootable JAR profile. I must be in the default profile and a bootable JAR execution with sar should be added.

Thanks for the hint @jfdenise! I fixed the Galleon execution but didn't add any bootable packaging/execution for `sar`, yet. 
This is not mandatory for pre-checking the feature and we can definitely add it later IMHO.

bstansberry:
This doesn't seem like another aspect of ts.layers as it's doing a lot of other modification and then I don't think is running all the stuff ts.layers does.

So ts.bootable makes sense.

Here are nightly jobs to run this:

https://ci.wildfly.org/viewType.html?buildTypeId=WF_BootableJarLinuxJdk8
https://ci.wildfly.org/viewType.html?buildTypeId=WF_BootableJarWindowsJdk11

Now they point at Fabio's branch. Once this is merged I'll point them at master.

This LGTM but I'm concerned about merging the dep on the arq Beta without talking with @jamezp first about the status there. He'll be back soon.

fabiobrz:
> This LGTM but I'm concerned about merging the dep on the arq Beta without talking with @jamezp first about the status there. He'll be back soon.

Hi @bstansberry - thanks for the info and the CI runs! :)
BTW, I agree on what you noted about Wildfly Arquillian Beta and also I'd like to know what you think about the bootable JAR plugin, which is still Beta as well. Will it stay as such since WildFly Final is prepared?

jfdenise:
If the rule is no Beta dependency in a WildFly Final (I think that is the case). I can make a 2.0.0.Final just before WF21 becomes Final, so can be referenced in your PR, then do a 2.0.1 to upgrade the plugin examples to use 21.0.0.Final.

rhusar:
In case this might help with the timing - remember a non-final version is fine as a test dependency. So if the non-final version bumps are reverted in main pom.xml and placed in the `ts.bootable` maven profiles that might be acceptable. But only use as last resort.

bstansberry:
Thanks for the suggestions, guys. I expect there shouldn't be an issue with moving WF ARQ to a Final version before WF 21 Final.  And once I get a chance to talk to James Perkins I figure I'd be fine with having the WF ARQ beta in master for a bit. Your suggestion @rhusar about using a profile for the bootable jar plugin is good to consider. Jean-Francois, we can talk in a bit about the relationship between the plugin and the main WF code base.

jfdenise:
@fabiobrz so we discussed that the WF21 release should contain a dependency on 2.0.0.Beta7 wildfly-jar-maven plugin. Dependency defined inside the ts.bootable profile (as @rhusar suggested). Once plugin 2.0.0.Final is released, we upgrade dependency in WF.

fabiobrz:
Hi @jfdenise, @rhusar, and thanks for the hint! Do you think the property for the version should be added to a common profile definition - e.g. in the testsuite pom file - or to each `ts.bootable` profile definition? I'd go for the the first one but I'd like to hear from you.

fabiobrz:
@jfdenise - while I was here I made the change to see whether it passes CI checks. 
Let me know whether you think I should move the version definition.
BTW I used 2.0.0.Beta6 though, since I don't see any Beta7 out. Am I missing something?

fabiobrz:
retest this please

jfdenise:
@fabiobrz , I released 2.0.0.Beta7. The master branch has been switched to 21.0.0.Final-SNAPSHOT version, you need to update your PR. Having everything in the ts.bootable is fine to me.

fabiobrz:
> @fabiobrz , I released 2.0.0.Beta7. The master branch has been switched to 21.0.0.Final-SNAPSHOT version, you need to update your PR. Having everything in the ts.bootable is fine to me.

Hi @jfdenise and thanks for your feedback. I've updated the PR to include the new version.

bstansberry:
@fabiobrz When ts.bootable runs in a job does QE mind if the tests in all the other maven modules are turned off?

We've got a lot of different CI jobs that execute more than they have to so I plan to clean that up. I'm "gathering requirements" so I can figure out how to do that most efficiently.

bstansberry:
@fabiobrz The -Dts.bootable profile works differently than I understood when I added the nightly CI jobs; i.e. it does not turn on running the maven modules where it adds executions. Which is ok. But when I changed the CI jobs to turn on running those a number of tests fail:

https://ci.wildfly.org/viewLog.html?buildId=223511&tab=buildResultsDiv&buildTypeId=WF_BootableJarLinuxJdk8
https://ci.wildfly.org/viewType.html?buildTypeId=WF_BootableJarWindowsJdk11

fabiobrz:
Hi @bstansberry and sorry for the late reply.

> @fabiobrz When ts.bootable runs in a job does QE mind if the tests in all the other maven modules are turned off?
> 
> We've got a lot of different CI jobs that execute more than they have to so I plan to clean that up. I'm "gathering requirements" so I can figure out how to do that most efficiently.

I'd say no. We have the chance to execute several different jobs covering modules. so no issues, I think.

> @fabiobrz The -Dts.bootable profile works differently than I understood when I added the nightly CI jobs; i.e. it does not turn on running the maven modules where it adds executions. Which is ok. But when I changed the CI jobs to turn on running those a number of tests fail:

I need to check this one and I'll let you know as soon as I have an idea on what's going on.

fabiobrz:
Hi @bstansberry  - I am experimenting with a different branch in order to fix and it seems I have all the modules running now.
Please notice that since I added the modules which should be run to the testsuite/integration/pom.xml `ts.bootable` profile configuration, I think that the usual `mvn` command line can be used, i.e. leaving the initial TeamCity CI configuration (just `-Dts.bootable`). Let me know it this seems reasonable to you too.

fabiobrz:
> Hi @bstansberry - I am experimenting with a different branch in order to fix and it seems I have all the modules running now.
> Please notice that since I added the modules which should be run to the testsuite/integration/pom.xml `ts.bootable` profile configuration, I think that the usual `mvn` command line can be used, i.e. leaving the initial TeamCity CI configuration (just `-Dts.bootable`). Let me know it this seems reasonable to you too.

Hi @bstansberry so I've renamed my local experimental branch (rebased on master already) and squashed some commits after I had a green bar on Jenkins.
As said earlier, the relevant modules were added to the `ts.bootable` profile so now something similar to the following command should be enough in order to run all the tests:
```
$ mvn install -Dmaven.repo.local=path-to-maven-repo-local -Dmaven.test.failure.ignore=true -Dts.bootable
```
so IMHO the TeamCity CI initial configuration could be resumed. 

On a side note, I have excluded 6 tests: 3 of them (security related) were excluded already from the correspondent `ts.layers` execution and the remaining 3 (about MP config module based configuration) require complex server setup which in the context of bootable JAR is something that would happen differently (although we've got some test coverage for similar cases).
Let me know what you think.

CC @jfdenise, @asoldano

bstansberry (approved):

bstansberry:
Thanks, @fabiobrz!

fabiobrz:
Thank you too for merging this PR @bstansberry!

PR #13517, opened 2020-08-31
 [WFLY-13814]: Drop messaging subsystem's HTTPUpgradeService's use of SimpleHttpUpgradeHandshake, drop dep on org.jboss.as.remoting.

ehsavoie:
* Replacing the code in HTTPUpgradeService.
 * Updating the MSC API usage by removing the use of the deprecated API.

    Jira: https://issues.redhat.com/browse/WFLY-13814


bstansberry (changes requested):
@ehsavoie This looks fine but I believe now the module.xml file no longer needs the dep on the org.jboss.as.remoting module.

bstansberry:
@ehsavoie This should be changed as well:

https://github.com/wildfly/wildfly/blob/master/messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/HTTPAcceptorDefinition.java#L60

Note also that that style of register the additional packages is deprecated (javadoc describes the preferred approach) so this may be an opportunity to change that.

bstansberry (approved):

bstansberry:
Thanks, @ehsavoie

PR #13522, opened 2020-09-01
[WFLY-13559]: Header response has changed and missing fields.

ehsavoie:
* resteasy.providers was overwritten instead of being appended to

Jira: https://issues.redhat.com/browse/WFLY-13559

### DIFF
--- a/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
+++ b/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
@@ -158,16 +158,34 @@ public class TracingDeploymentProcessor implements DeploymentUnitProcessor {
         return null;
     }
 
-    private void addContextParameter(JBossWebMetaData jbossWebMetaData, ParamValueMetaData restEasyProvider) {
+    private void addResteasyProvidersParameter(JBossWebMetaData jbossWebMetaData) {
         if (jbossWebMetaData == null) {
             return;
         }
+        ParamValueMetaData restEasyDynamicFeature = getResteasyProvidersParam(jbossWebMetaData);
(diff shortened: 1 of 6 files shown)


microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingDeploymentProcessor.java
```diff
@@ -158,18 +158,40 @@ private String getTracerConfiguration(DeploymentPhaseContext deploymentPhaseCont
...
             contextParams = new ArrayList<>();
         }
-        contextParams.add(restEasyProvider);
+        ParamValueMetaData restEasyDynamicFeature = getResteasyProvidersParam(jbossWebMetaData);
+        if (restEasyDynamicFeature.getParamValue() != null && !restEasyDynamicFeature.getParamValue().trim().isEmpty()) {
+            restEasyDynamicFeature.setParamValue(restEasyDynamicFeature.getParamValue() + ",org.wildfly.microprofile.opentracing.smallrye.TracerDynamicFeature");
+        } else {
+            restEasyDynamicFeature.setParamValue("org.wildfly.microprofile.opentracing.smallrye.TracerDynamicFeature");
+        }
+        contextParams.add(restEasyDynamicFeature);
```
bstansberry, line 175: This may add a 2nd 'resteasy.providers' ParamValueMetaData to the list, which may work but seems fragile.

ehsavoie: correct

bstansberry: @ehsavoie So will it stay this way or do you expect to change it?

ehsavoie: I'm fixing this and updating the test as it wasn't breaking without the fix ;)

bstansberry (approved):

PR #13525, opened 2020-09-02
[WFLY-13632] If a CMR resource is deployed concurrently then it will overwrite the list of resources that are expected to behave as CMR

gaol:
~Issue: https://issues.redhat.com/browse/WFLY-13633~

Issue: https://issues.redhat.com/browse/WFLY-13632

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/CMResourceAdd.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/CMResourceAdd.java
@@ -64,6 +64,10 @@ class CMResourceAdd extends AbstractAddStepHandler {
      */
     @Override
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
+        if (!context.isBooting()) {
+            context.reloadRequired();
+            return;
+        }
         PathAddress address = PathAddress.pathAddress(operation.get(OP_ADDR));
         final String jndiName = address.getLastElement().getValue();
         final String tableName = CMResourceResourceDefinition.CM_TABLE_NAME.resolveModelAttribute(context, model).asString();
@@ -76,9 +80,5 @@ class CMResourceAdd extends AbstractAddStepHandler {
                 .addDependency(TxnServices.JBOSS_TXN_JTA_ENVIRONMENT, JTAEnvironmentBean.class, service.getJTAEnvironmentBeanInjector())
                 .setInitialMode(ServiceController.Mode.ACTIVE)
                 .install();
-
-        if (!context.isBooting()) {
-            context.reloadRequired();
-        }
     }
 }
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/JTAEnvironmentBeanServlet.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/JTAEnvironmentBeanServlet.java
@@ -1,0 +1,47 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2020, Red Hat Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


gaol:
@tomjenkinson @bstansberry @mmusgrov Could anyone please review? Thanks :)

bstansberry:
This LGTM but let's wait a bit in case @mmusgrov has any concerns.

gaol:
@tomjenkinson doen't [jtaBean.setCommitMarkableResourceJNDINames(connectableResourceJNDINames);](https://github.com/wildfly/wildfly/pull/13525/files#diff-104e9239790ebc24b14cf99ab71fc053R94) update the underlying collection in narayana ?

gaol:
@tomjenkinson OK, I understand now. Thank you. It is more than I expected, and it needs changes in Narayana first to keep all consistent.

mmusgrov:
Yes Tom is correct and if we cannot perform a server restart then we would also need a fix in Narayana.

mmusgrov:
@gaol Hi, I created a [JBTM issue](https://issues.redhat.com/browse/JBTM-3357) to ensure that these property getters are dynamic.

bstansberry:
@gaol [~gaol]. Note that in WildFly CMResourceAdd has a bug in how it deals with reload-required:

https://github.com/wildfly/wildfly/blob/master/transactions/src/main/java/org/jboss/as/txn/subsystem/CMResourceAdd.java#L80

That entire method should really be an if/else block -- it either the server isn't booting and it just calls context.setReloadRequired() or it is booting and it modifies the runtime state by adding the service.

Right now my *impression* from the discussion is that doesn't matter too much because if CMResourceAdd runs after boot, the service it adds doesn't actually update the state that Narayana is really using. So it's doing pointless work, but it's not modifying the actual runtime behavior and then falsely telling the user it didn't by setting reload-required.

That should be fixed regardless, but particularly if things change in how Narayana handles updates made to JTAEnvironmentBean. Basically, what the handlers for the 'add' and 'remove' ops on these resources do should match what happens inside Narayana. If a change doesn't take effect without a process restart, then they should set restart-required. If only a reload is needed, then reload-required. If they are able immediately to correctly change what Narayana is doing, then no reload-required or restart-required.

The synchronization stuff here is good regardless, because during boot multiple CMResourceService instances can be installed, and they need to coordinate.

gaol:
@bstansberry Thanks for the review :)  

Here are summary of the changes:

* I updated the `CMResourceAdd` part to address the issue Brian mentioned.
* I updated the PR summary and description to target https://issues.redhat.com/browse/WFLY-13632 only.
* The https://github.com/jbosstm/narayana/pull/1670 may solve https://issues.redhat.com/browse/WFLY-13633 together with this fix.

This PR doesn't fix https://issues.redhat.com/browse/WFLY-13633 because updates to `JTAEnvironmentBean` by `CMResourceService` does not change the **static references** to `JTAEnvironmentBean.commitMarkableResourceJNDINames`([for example](https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L1748)) in other places in narayana.

After `:reload` operation, the static references to `JTAEnvironmentBean.commitMarkableResourceJNDINames`([for example](https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L1748)) are still the same because they are cached as the static class variables.

Together with https://github.com/jbosstm/narayana/pull/1670, which changes the static class fields to final instance fields, the `:reload` operation should be able to re create all needed instances with updated `JTAEnvironmentBean`.

But I think setting the state to `restart-required` should also be able to fix the problem, I am not sure if changes to `JTAEnvironmentBean` should lead to a hard restart or not.

mmusgrov:
This PR will protect against the data integrity issue if we mark the server as `restart-required`.
@gaol would you be able to update your PR to include that change?

transactions/src/main/java/org/jboss/as/txn/subsystem/CMResourceAdd.java
```diff
@@ -64,6 +64,10 @@ protected void populateModel(ModelNode operation, ModelNode model) throws Operat
      */
     @Override
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
+        if (!context.isBooting()) {
+            context.reloadRequired();
```
mmusgrov, line 68: Can we change this to restart to force Narayana to reread its config?

bstansberry:
Superseded by #13620

PR #13526, opened 2020-09-02
[WFLY-13807] EE: Add component class info to interface based proxy to…

tadamski:
… preserve validation metadata

https://issues.redhat.com/browse/WFLY-13807

This fix makes the component class the superclass of the interface proxy to preserve validation metadata. Currently in such case the superclass is set to Object, so validation metadata is lost. We cannot return proxy directly because it does not contain generics information.

This change lead to following problem: ViewService interceptor chains are hashed using Method object. As a result, after changing the superclass to the bean class some invocations failed if they invoked method was the interface method. I have modified view service so that in such case it matches the method based on the signature. Furthermore, I have to modify AuthorizationInterceptor to not throw an Exception if invokedMethod and componentMethod don't equal (after the fix one those can be bean and interface methods).

This PR is a big hacky. The alternative would be to make views hash their methods based on descriptions/signatures - this possibly may be a major refactor though and I'm not sure if it would be feasible.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/DefaultComponentViewConfigurator.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/DefaultComponentViewConfigurator.java
@@ -90,7 +90,7 @@ class DefaultComponentViewConfigurator extends AbstractComponentConfigurator imp
 
             //we define it in the modules class loader to prevent permgen leaks
             if (viewClass.isInterface()) {
-                proxyConfiguration.setSuperClass(Object.class);
+                proxyConfiguration.setSuperClass(configuration.getComponentClass());
                 proxyConfiguration.addAdditionalInterface(viewClass);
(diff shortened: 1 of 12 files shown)


gaol:
@tadamski would you please check if the test failures are related?

tadamski:
@gaol  I have pushed a possible fix - the tests should pass now

tadamski:
retest this please

gaol:
retest this please

gaol:
@tadamski The test still fails, would you please check again? Thanks :)

@darranl, There is a downstream issue waiting for this PR, if there are other changes needed to pass this CI, we need to evaluate what should be back ported to fix this issue. :)

chengfang:
@gaol I'll take a look.

bstansberry:
Superseded by #13600

PR #13381, opened 2020-06-22
[WFLY-13539] Add Galleon layer for JBoss Diagnostic Reporting support

yersan:
Adds JDR Galleon layer and increases the test coverage.

Jira issue: https://issues.redhat.com/browse/WFLY-13539
WildFly proposal: https://github.com/wildfly/wildfly-proposals/pull/311

PR #13528, opened 2020-09-02
[WFLY-13539] Add Galleon layer for JBoss Diagnostic Reporting support

bstansberry:
https://issues.redhat.com/browse/WFLY-13539

Follows up on #13381 with an additional commit to update the test improvements in that PR to not assert things that are validly not true. Running the test in other environments where a server had no ServiceModuleLoader present the test would incorrectly fail.


yersan (approved):

yersan:
@bstansberry would you mind rebasing this? the PR hash is a bit behind and I would rather prefer to pass the pre-check with https://github.com/wildfly/wildfly-core/pull/4295 incorporated by wildfly-core on the topic branch. I can rebase the PR on my local repository, but then we will not have a shareable commit hash as reference for the pre-check, I mean, if QE wants to use the same hash to do a pre-check, likely it won't be the same.

CI is fine since it merged the PR with master when the changes were already there.

bstansberry:
@yersan Rebased

PR #13529, opened 2020-09-03
[WFLY-13356]:Add webservices galleon layer

jimma:
JIRA: https://issues.redhat.com/browse/WFLY-13356
Analysis doc: https://github.com/wildfly/wildfly-proposals/pull/334

### DIFF
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml
@@ -2,7 +2,7 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="webservices">
     <dependencies>
         <layer name="web-server"/>
-        <layer name="ejb" optional="true"/>
+        <layer name="ejb-lite" optional="true"/>
         <layer name="messaging-activemq" optional="true"/>
     </dependencies>
     <feature-group name="webservices"/>
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="webservices">
+    <dependencies>
+        <layer name="web-server"/>
+        <layer name="ejb" optional="true"/>
+        <layer name="messaging-activemq" optional="true"/>
+    </dependencies>
+    <feature-group name="webservices"/>
+    <packages>
+        <package name="org.jboss.as.webservices.server.integration"/>
+        <package name="org.jboss.ws.saaj-impl"/>
+        <package name="org.jboss.ws.cxf.sts"/>
+        <!-- ws tools provision -->
+        <package name="bin.wstools"/>
+        <package name="org.jboss.ws.tools.common"/>
+        <package name="org.jboss.ws.tools.wsconsume"/>
+        <package name="org.jboss.ws.tools.wsprovide"/>
+    </packages>
+
+</layer-spec>
--- a/testsuite/integration/ws/pom.xml
+++ b/testsuite/integration/ws/pom.xml
@@ -350,6 +350,45 @@
                                 </configuration>
                             </execution>
                             <execution>
+                                <id>webservices-ejb-provisioning</id>
+                                <goals>
+                                    <goal>provision</goal>
+                                </goals>
+                                <phase>compile</phase>
+                                <configuration>
+                                    <install-dir>${project.build.directory}/wildfly-webservices-ejb</install-dir>
+                                    <record-state>false</record-state>
+                                    <log-time>${galleon.log.time}</log-time>
+                                    <offline>true</offline>
+                                    <plugin-options>
+                                        <jboss-maven-dist/>
+                                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                                        <optional-packages>passive+</optional-packages>
+                                    </plugin-options>
+                                    <feature-packs>
+                                        <feature-pack>
+                                            <groupId>${project.groupId}</groupId>
+                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
+                                            <version>${testsuite.ee.galleon.pack.version}</version>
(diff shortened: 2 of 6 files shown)


bstansberry:
@jimma It's ok to add a provisioning execution and a surefire execution that runs with legacy security provisioned in order to execute a few more tests, but there needs to be a test execution without legacy security. Users of this layer should not need to have legacy security. We do not want people using legacy security.

In my EE 9 branch I was able to get a lot of testsuite/integration/ws passing against a config with no legacy security subsystem -- https://github.com/bstansberry/wildfly/blob/ee-9/testsuite/integration/ws/pom.xml#L285. I had to exclude some tests but not nearly all. I'm pretty sure I didn't do anything special to make it work, as you'd made a needed change in master to allow it. In that branch I'm deliberately avoiding changing any subsystem java code.

jimma:
@bstansberry  thanks for review. I'll move ws tests to run with elytron security.

jimma:
@bstansberry @chengfang I already rebased the latest ejb layer change and fixed couple of things.  Please review.

jimma:
retest this please.

bstansberry:
@jimma Cheng Fang updated #13524 so this will need an update. I pushed up #13535 which is Emmanuel's messaging work updated to base on the latest EJB.  So that can become the base for this.

jimma:
@bstansberry  I just rebased with https://github.com/wildfly/wildfly/pull/13535 and pushed.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="webservices">
+    <dependencies>
+        <layer name="web-server"/>
+        <layer name="ejb" optional="true"/>
```
bstansberry, line 5: Is remote EJB support the common case with webservices? If so this is ok; otherwise ejb-lite is better as it has fewer dependencies. If a user wanted @Remote they could just add the 'ejb' layer and have that.

I see a few tests in testsuite/integration/ws use @Remote EJBs but if that's just for test coverage and it's not the typical use case those could be run in a separate execution with a server that included the 'ejb' layer.

Sorry if this has been discussed and I missed it or forgot.

jimma: I tried with ejb-lite and there are a lot of missing remote failures from testsuite/integration/ws module like : 
```
org.jboss.arquillian.container.spi.client.container.DeploymentException: Cannot deploy stateless-ws-endpoint-example.jar: 
{"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => 
{"WFLYCTL0412: Required services that are not installed:" => ["org.wildfly.ejb.remote","org.wildfly.transactions.remote-
transaction-service"],"WFLYCTL0180: Services with missing/unavailable dependencies" => 
["jboss.deployment.unit.\"stateless-ws-endpoint-
example.jar\".component.SimpleStatelessWebserviceEndpointImpl.CREATE is missing [org.wildfly.transactions.remote-
transaction-service]","jboss.deployment.unit.\"stateless-ws-endpoint-
example.jar\".component.SimpleStatelessWebserviceEndpointImpl.VIEW.\"org.jboss.as.test.integration.ws.ejb.SimpleState
lessWebserviceEndpointIface\".REMOTE is missing [org.wildfly.ejb.remote]"]}}}

```

bstansberry: When I tried, 5 out of 58 tests failed.

I think those are the tests where the EJBs are annotated with @Remote. If so ,they don't pass for a valid reason; they'd be testing something outside the intended use case of the layer. But the presence of those tests doesn't mean that's the use case our layer should target. They could be excluded or tested in a separate execution with the 'ejb' layer added, as you did with the tests that required legacy security. The layer should be designed to be as slim as possible while still meeting the expected end user use cases.

We wouldn't want to base on ejb-lite and make the user provision both 'webservices' and 'ejb' instead of just 'webservices' if wanting remote EJB is the common use case. On the other hand, 'ejb' brings the remoting subsytem and an endpoint, which adds both weight and security risk. If a user didn't want that they'd have to use 'webservices' and 'ejb-lite' and exclude 'ejb', which is fairly painful and not intuitive. It's doable though, so if wanting remote EJBs is the common use case you're targeting it's ok.

jimma: @bstansberry  Looked at layer dependency twice, and I agree with you that we should make sure the layer to be as slim as possible. I added the "ejb" layer dependency as user can get a out of box webservices layer which works for most of the causes, but this will increase the installation size and make the layer exclusion painful for user. I didn't think of the scenario if user wants to exclude something.  I pushed the change to change the dependency to "ejb-lite" as you suggested.

bstansberry (approved):

bstansberry:
@jimma A bunch of galleon layers PRs have completed pre-check and have been merged or are ready for merge. But they conflict with each, as you can see with the conflict on your branch.

https://github.com/wildfly/wildfly/pull/13546 is a branch that has all but this one in it with conflicts resolved. Please rebase your branch on top of that. I expect I'll be merging it into master this evening my time.

testsuite/integration/ws/pom.xml
```diff
@@ -254,9 +254,258 @@
...
+                                            <inherit-packages>false</inherit-packages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <configurations>
+                                        <config>
+                                            <model>standalone</model>
+                                            <name>standalone.xml</name>
+                                            <layers>
+                                                <layer>cloud-server</layer>
+                                                <layer>ejb</layer>
```
tommaso-borgato, line 382: @jimma this was added wit commit "[WFLY-13356]:Use ejb-lite dependency" ... ee-feature-pack/galleon-common/src/main/resources/layers/standalone/webservices/layer-spec.xml correcly uses `ejb-lite` layer, while the testsuite/integration/ws/pom.xml uses `ejb` layer, is that intentional?

jimma: @tommaso-borgato  yes. Adding ejb layer to get remote support.

bstansberry:
@jimma Another rebase needed; should be the last one

jimma:
@bstansberry  Rebased.

PR #13538, opened 2020-09-07
[WFLY-13578] adding integration failure tests for WFTC remote call with txn propagation

ochaloup:
https://issues.redhat.com/browse/WFLY-13578


ochaloup:
@bstansberry the failing test was the `BeforeShutdownJNDILookupTestCase` and the reason was that it uses the method `controller.kill`. See https://github.com/wildfly/wildfly/blob/21.0.0.Final/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/weld/extension/BeforeShutdownJNDILookupTestCase.java#L69
As the `controller.kill` is in fact implemented in way of calling the `stop` in the WFLY Arq then calling `stop` and `kill` do the same. I think the test `BeforeShutdownJNDILookupTestCase` expects the server not being killed immediatelly (like `kill -9`) but it assumes a shutdown hook to be invoked on container shutdown. I changed the call for `stop`.

The original version of the PR was adding an arquillian extension (https://github.com/ochaloup/wildfly/commit/fb9776674d99b3f0259338f844080cc220ccb5a8#diff-0567bd51c80833a49443f175c57da2e53a8974617f4821e0cb0be988ca0a4a8aR44) which makes the `kill` to do nothing. It was necessary because there was a bug in WFLY Arquilian (https://issues.redhat.com/browse/WFARQ-78). With the new release of the managed container (3.0.x) it's fixed and updating the WFLY arq fixes the test and does not make a need to add the arquillian extension which workarounds the `kill` behaviour.

Then I'm not sure why the WFLY Arq is not used with new version in whole WFLY - https://github.com/wildfly/wildfly/blob/21.0.0.Final/pom.xml#L446 (https://issues.redhat.com/browse/WFLY-13937) but hopefully it could work for the manual mode module at least (as it's needed for my newly created testcase).

ochaloup:
@bstansberry is this something that could be merged?
Or could be somebody so kind to review this first? e.g. @chengfang or @tadamski :-)

tadamski (approved):
LGTM

testsuite/integration/manualmode/pom.xml
```diff
@@ -25,6 +25,9 @@
         <jbossas.ts.integ.dir>${basedir}/..</jbossas.ts.integ.dir>
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}/..</jbossas.project.dir>
+
+        <!-- redefinition of WildFly Arquillian version to get fixes for the container stop/kill functionality -->
+        <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
```
bstansberry, line 30: @ochaloup Please remove this as it is the normal version now.

bstansberry (approved):
@ochaloup I've got a minor change requested. Once that's done and it passes CI please ask  @wildfly/mergers-full (not me personally) to get this merged.

ochaloup:
@wildfly/merger-full the tests passed, can you please take a look?

ochaloup:
@wildfly/mergers-full The tests passed. Would you be so kind to check this, please?

bstansberry:
@ochaloup When I combined this with a bunch of PRs to test the set before merging, I saw failures of two tests added here:

https://ci.wildfly.org/viewLog.html?buildId=235039&buildTypeId=WF_PullRequest_WindowsJdk11

That was on #13829

I don't have any particular reason to think that was because of the other commits in that PR. It seems more probable it was just an intermittent failure exposed by running the test another time.

bstansberry:
I initiated a re-test but I just cancelled it as I want to keep CI free today for things that are critical for WF 22. So ignore the cancellation failures.

bstansberry:
@tomjenkinson @ochaloup Previously the tests in this PR failed on some test runs (see my comment of Dec 16), so I'm occasionally doing a retest to help guard against the possibility that these will introduce disruptive intermittent failures in CI.

This doesn't need to be in the WF 23 Beta, so if my latest retest looks ok it's probably good to merge it soon after we tag the beta so we have two weeks of it running all the time in CI to detect and fix any intermittent failures.

PR #13542, opened 2020-09-09
WFLY-12586 Duplicated classes in test suite, content slightly differs in some cases

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-12586


testsuite/shared/src/main/java/org/jboss/as/test/integration/common/WebInfLibClass.java
```diff
@@ -0,0 +1,4 @@
+package org.jboss.as.test.integration.common;
```
yersan, line 1: Would you mind moving also the license header to the common file? I would say it is not mandatory for merging, but we should keep the pattern we already have.

darranl, line 1: Actually if the header is missing this is mandatory for merging, I know some other projects have relaxed that requirement but the WildFly Core and WildFly repos have mixed licensing so I believe we do still require the header to correctly identify on a file by file basis.

darranl (changes requested):
Missing header.

testsuite/shared/src/main/java/org/jboss/as/test/integration/common/WebInfLibClass.java
parsharma: Sure. I will move the license header.

darranl (approved):

PR #13550, opened 2020-09-10
[WFLY-13841] Check the compressed size of JDR test case Zip entries t…

yersan:
…o validate its size

Jira issue: https://issues.redhat.com/browse/WFLY-13841

Intermittently [ZipEntry.getSize()](https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#getSize--) is not returning the size of the ZipEntry. This patch tries to use the [ZipEntry. getCompressedSize()](https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#getCompressedSize--) as a second opportunity to verify if the entry is empty.

I've been unable to identify exactly the reason why sometimes it does not returns the size, but this patch fixed the executions in my local envrionement.


yersan:
Closing, this solution is still invalid

PR #13557, opened 2020-09-11
[WFLY-13846] Eliminate the connector dependency on the legacy security subsystem's module.

darranl:
https://issues.redhat.com/browse/WFLY-13846

I did consider if these ServiceName definitions should maybe in a common class but TBH they are used in various packages in the connector subsystem so preferred to keep visibility restricted - at some point all these references will be deleted anyway.

### DIFF
--- a/connector/pom.xml
+++ b/connector/pom.xml
@@ -70,10 +70,6 @@
             <artifactId>wildfly-transaction-client</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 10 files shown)


darranl:
Just temporarily on hold, I realise I haven't cleaned the pom dependency in there as well.

bstansberry (approved):
This LGTM.

CI didn't run so there may be some issue it finds but what's done here seems good.

bstansberry:
retest this please

PR #13559, opened 2020-09-11
[WFLY-13850] Fix maven coordinate variables for the provisioning of m…

yersan:
…icroprofile-platform on the test suite

Jira issue: https://issues.redhat.com/browse/WFLY-13850

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1363,9 +1363,9 @@
                                     </plugin-options>
                                     <feature-packs>
                                         <feature-pack>
-                                            <groupId>${testsuite.ee.galleon.pack.groupId}</groupId>
-                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
-                                            <version>${testsuite.ee.galleon.pack.version}</version>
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

rhusar (approved):

testsuite/integration/basic/pom.xml
```diff
@@ -1363,9 +1363,13 @@
                                     </plugin-options>
                                     <feature-packs>
                                         <feature-pack>
-                                            <groupId>${testsuite.ee.galleon.pack.groupId}</groupId>
-                                            <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
-                                            <version>${testsuite.ee.galleon.pack.version}</version>
+                                            <feature-pack>
```
bstansberry, line 1366: This adds another level of xml instead of just replacing the values in the 3 elements. This is why the Galleon job fails.

bstansberry (changes requested):

testsuite/integration/basic/pom.xml
yersan: Fixed, an oversight on my side since I did not test it locally before pushing it.

bstansberry (approved):

PR #13564, opened 2020-09-14
WFLY-13856 [WARNING] 'profiles.profile[ts.clustering.layers.profile].plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.wildfly.plugins:wildfly-maven-plugin @ line 1255, column 29

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-13856

The diff might seem very off, but this fixes incorrect maven usage where new executions were added along with plugin definitions rather than being added as new executions to the existing plugin definitions. This fixes that.

### DIFF
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -1178,62 +1178,6 @@
                     <plugin>
                         <groupId>org.wildfly.plugins</groupId>
                         <artifactId>wildfly-maven-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


rhusar:
According to git blame - https://github.com/wildfly/wildfly/commit/773be0569d80667471dd784936045270f1365003 - @chengfang and @bstansberry for review/merge. Thanks!

pferraro (approved):

chengfang (approved):

PR #13568, opened 2020-09-15
[WFLY-13659] Reduction of dependencies on the org.jboss.as.security module.

darranl:
This PR contains commits for the following issues:

https://issues.redhat.com/browse/WFLY-13848
https://issues.redhat.com/browse/WFLY-13851
https://issues.redhat.com/browse/WFLY-13852
https://issues.redhat.com/browse/WFLY-13862
https://issues.redhat.com/browse/WFLY-13868

Two further modules "vault-tool" and "appclient" need to be checked but this batch is ready for review / CI / merge if appropriate.


bstansberry:
@darranl Your new artifact needs to be added to servlet-feature-pack/galleon-feature-pack/pom/xml and servlet-feature-pack/legacy-feature-pack/pom.xml

darranl:
@bstansberry Hopefully CI agrees but I have corrected the dependency issues causing some test failures, I think this is now ready for review.

The only remaining dependencies I see are appclient and picketlink.

PicketLink is a PicketBox based integration, I think it just depends on service name definitions but TBH a clean up there doesn't add any value to us.

Appclient is an interesting one, seems to have the dependency but I couldn't find any code using it - I don't think it is a priority for this clean up but maybe we could drop it to optional.

darranl:
The hibernate failure looks to have happened a couple of times - investigating.

darranl:
Actually the job didn't run yet - I can't reproduce the failure locally and I am building with Java 8 on Linux.  I am just trying a merge with master in case there is an interaction I am not seeing.

darranl:
@ehsavoie Could you please give this one a review for the messaging changes, especially around what we have learned about class loading.

bstansberry (approved):

PR #12887, opened 2020-01-08
[WFLY-12190] EJB: EJB over HTTP discovery configuration

tadamski:
https://issues.redhat.com/browse/EAP7-1236
https://issues.redhat.com/browse/WFLY-12190

Depends on:
https://github.com/wildfly/wildfly/pull/13305

Thanks for submitting your Pull Request!

Please make sure your PR meets the following requirements:
- [ ] Pull Request title is properly formatted: `[WFLY-XYZ] Subject` or `WFLY-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue(s)
- [ ] Pull Request contains description of the issue(s)
- [ ] Pull Request does not include fixes for issues other than the main ticket
- [ ] Attached commits represent units of work and are properly formatted

For bigger changes, major and minor component upgrades make sure your PR also meets following requirements:
- [ ] Pull Request requires a change to the documentation
- [ ] Documentation have been updated accordingly
- [ ] Tests were added to cover changes

For new features ensure as well:
- [ ] Analysis was done
- [ ] Test Plan has been done
- [ ] Tests were verified in advance

If you are not an active contributor of the WildFly project you can request sponsorship by one of the members to help guide you through the process.

PR #13579, opened 2020-09-17
Pulls

bstansberry:
#12887 
#13552 
#13578


darranl:
The only failure seems to be one I have seen a few times.

PR #13586, opened 2020-09-18
Pulls

bstansberry:
#13500


darranl:
@rhusar / @pferraro Is this the PR supposed to address the kind of failures coming up in the "Linuc MicroProfile - JDK 8" job or is that a different one?

Not sure if that job needs kicking off again or not.

rhusar:
@darranl That's a different one, the WFLY-13869 doesn't seem to have resolved the issue completely.

bstansberry:
This was just me testing #13500 again against an updated master. The CI jobs that were running at the time on #13500 were against an outdated master.

PR #13588, opened 2020-09-18
WFLY-13883 Clustering TS: Enable Infinispan Server integration tests …

rhusar:
…on Windows platform

Resolves
https://issues.redhat.com/browse/WFLY-13883


pferraro:
This throws an java.lang.UnsupportedOperationException at org.infinispan.commons.jdkspecific.ProcessInfo.of(ProcessInfo.java:30) :/

rhusar:
@pferraro If that mismatch was the case, why would it not be a problem on the 'Linux - JDK 11 ' job?

rhusar:
Anyway, this needs more fixes apparently so closing.

PR #13590, opened 2020-09-21
[WFLY-13866]: Several traces created in Jaeger for just one method call.

ehsavoie:
* the current scope is 'lost' on async calls. Ficxing that by passing
  the scope manager on the new threads.

Jira: https://issues.redhat.com/browse/WFLY-13866

### DIFF
--- a/microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
+++ b/microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright 2020 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.extension.microprofile.faulttolerance;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Miniature context propagation API, adapted from MicroProfile Context Propagation and SmallRye Context Propagation.
+ * @author Ladislav Thon (c) 2020 Red Hat, Inc.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
+ */
+final class MiniContextPropagation {
+
+    private MiniContextPropagation() {
+        // avoid instantiation
+    }
+
+    public static ExecutorService executorService(ContextProvider contextProvider, ExecutorService delegate) {
+        return new ConPropExecutorService(contextProvider, delegate);
+    }
+
+    public static ScheduledExecutorService scheduledExecutorService(ContextProvider contextProvider, ScheduledExecutorService delegate) {
+        return new ConPropScheduledExecutorService(contextProvider, delegate);
+    }
+
+    @FunctionalInterface
+    interface ContextProvider {
+
+        ContextSnapshot capture();
+
+        ContextProvider NOOP = () -> ContextSnapshot.NOOP;
+
+        static ContextProvider compound(ContextProvider... contextProviders) {
+            return new CompoundContextProvider(Arrays.asList(contextProviders));
+        }
+    }
+
+    @FunctionalInterface
+    interface ContextSnapshot {
+
+        ActiveContextSnapshot activate();
+
+        ContextSnapshot NOOP = () -> ActiveContextSnapshot.NOOP;
+    }
+
+    @FunctionalInterface
+    interface ActiveContextSnapshot {
+
+        void deactivate();
+
+        ActiveContextSnapshot NOOP = () -> {
+        };
+    }
+
+    // ---
+    private static final class CompoundContextProvider implements ContextProvider {
+
+        private final List<ContextProvider> contextProviders;
+
+        CompoundContextProvider(List<ContextProvider> contextProviders) {
+            this.contextProviders = contextProviders;
+        }
+
+        @Override
+        public ContextSnapshot capture() {
+            List<ContextSnapshot> snapshots = new ArrayList<>();
+            for (ContextProvider contextProvider : contextProviders) {
+                snapshots.add(contextProvider.capture());
+            }
+
+            return () -> {
+                List<ActiveContextSnapshot> activeSnapshots = new ArrayList<>();
+                for (ContextSnapshot snapshot : snapshots) {
+                    activeSnapshots.add(snapshot.activate());
+                }
+
+                return () -> {
+                    for (int i = activeSnapshots.size() - 1; i >= 0; i--) {
+                        activeSnapshots.get(i).deactivate();
+                    }
+                };
+            };
+        }
+    }
+
+    // ---
+    private static final class ConPropRunnable implements Runnable {
+
+        private final ContextSnapshot contextSnapshot;
+        private final Runnable delegate;
+
+        ConPropRunnable(ContextSnapshot contextSnapshot, Runnable delegate) {
+            this.contextSnapshot = contextSnapshot;
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void run() {
+            ActiveContextSnapshot activeSnapshot = contextSnapshot.activate();
+            try {
+                delegate.run();
+            } finally {
+                activeSnapshot.deactivate();
+            }
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


microprofile/opentracing-extension/src/test/java/org/wildfly/extension/microprofile/opentracing/Subsystem_2_0_ParsingTestCase.java
```diff
@@ -152,6 +152,7 @@ protected void initializeExtraSubystemsAndModel(ExtensionRegistry extensionRegis
             super.initializeExtraSubystemsAndModel(extensionRegistry, rootResource, rootRegistration, capabilityRegistry);
             Map<String, Class> capabilities = new HashMap<>();
             capabilities.put(WELD_CAPABILITY_NAME, WeldCapability.class);
+            capabilities.put("org.wildfly.microprofile.config", Void.class);
```
rhusar, line 155: Please use org.wildfly.extension.microprofile.opentracing.SubsystemDefinition#MICROPROFILE_CONFIG_CAPABILITY_NAME instead and make it public.

microprofile/opentracing-extension/src/test/java/org/wildfly/extension/microprofile/opentracing/Subsystem_3_0_ParsingTestCase.java
```diff
@@ -164,6 +164,7 @@ protected void initializeExtraSubystemsAndModel(ExtensionRegistry extensionRegis
             super.initializeExtraSubystemsAndModel(extensionRegistry, rootResource, rootRegistration, capabilityRegistry);
             Map<String, Class> capabilities = new HashMap<>();
             capabilities.put(WELD_CAPABILITY_NAME, WeldCapability.class);
+            capabilities.put("org.wildfly.microprofile.config", Void.class);
```
rhusar, line 167: Please use org.wildfly.extension.microprofile.opentracing.SubsystemDefinition#MICROPROFILE_CONFIG_CAPABILITY_NAME instead and make it public.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/opentracing/OpenTracingFaultToleranceTestCase.java
```diff
@@ -0,0 +1,180 @@
...
+            });
+            for (int i = 0; i < spans.size(); i++) {
+                JsonObject object = spans.get(i);
+                String operationName = object.getString("operationName");
+                Assert.assertNotNull(object.toString(), object.getString("traceId"));
+                Assert.assertNotNull(object.toString(), object.getString("spanId"));
+                Assert.assertEquals(object.toString(), parentId, Integer.parseInt(object.getString("parentId")));
+                switch (i) {
+                    case 0: {
+                        Assert.assertEquals(object.toString(), operationName, "GET:org.wildfly.test.integration.microprofile.faulttolerance.opentracing.application.HelloResource.get");
```
rhusar, line 81: Please future-proof this by using `Java.lang.Class.getCanonicalName()` instead of string constants that will inevitably break with next refactoring.

rhusar (changes requested):

microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Miniature context propagation API, adapted from MicroProfile Context Propagation and SmallRye Context Propagation.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
```
Ladicek, line 33: You stole my code! :-D

Really, I'd suggest to:
- make this class package-private
- add a big TODO that this should be replaced by Context Propagation proper when it's integrated in WildFly

rhusar: Certainly a note can be added, but it's not that straightforward – even though if Context Propagation is integrated in WildFly it doesn't mean that the users must provision it in theirs server while still anticipating that the tracing context propagation works - which is what we are doing in the PR.

Ladicek: Yea, the thing is, what this PR does (which is the same thing I did in Thorntail) won't work with other things that also use context propagation. Thorntail is dead, but WildFly isn't, so it deserves better treatment. I'd suggest that if users provision Fault Tolerance and OpenTracing, but don't provision Context Propagation, then they should _NOT_ expect traces being propagated across FT `@Asynchronous` calls. Obviously, that becomes a breaking change if this PR is merged, so...

Again, I could do this hack in Thorntail, because it's dead, but you need to be more careful with WildFly.

ehsavoie: @Ladicek  You mean I took responsibility of this piece of code in WildFly ? ;)

ehsavoie: @Ladicek I disagree with your POV on `@Asynchronous` having tracer spans coming out of nowhere doesn't seem something that make sense. The fact that this issue arose is proof of that to me. If context propagation is provided we should rely on it insted of the current mechanism but otherwise I think we should still provide it.

Ladicek: I agree with you, but the same argument applies to all cases where context propagation is / can be  used, such as RestClient + OpenTracing, @Asynchronous + @RequestScoped, etc. etc. etc. It doesn't make any sense to build a special case support for everything.

On the other hand, of course, Context Propagation is not in MicroProfile yet, so this is all a gray area.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/opentracing/application/MyService.java
```diff
@@ -0,0 +1,65 @@
...
+    @Fallback(fallbackMethod = "fallback")
+    public String hello() {
+        try (Scope scope = tracer.buildSpan("hello").startActive(true)) {
+            scope.span().log("attempt " + counter.getAndIncrement());
+            throw new RuntimeException("failed");
+        }
+
+//        tracer.buildSpan("hello").start();
+//        tracer.scopeManager().activeSpan().log("attempt " + counter.getAndIncrement());
+//        throw new RuntimeException("failed");
```
rhusar, line 36: Maybe this can be gone?

rhusar (changes requested):

microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
bstansberry: @rhusar WDYT here?

microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Miniature context propagation API, adapted from MicroProfile Context Propagation and SmallRye Context Propagation.
+ * @author Ladislav Thon (c) 2020 Red Hat, Inc.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
+ */
+public final class MiniContextPropagation {
```
bstansberry, line 36: If this and other new classes can be package private that is better.

It's a private module but locking things down is better.

microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/MiniContextPropagation.java
rhusar: I am on the fence here, spent a lot of time thinking about this, here is some summary:

* All of this is out of the spec – there is no mention in FT nor OpenTracing.
* We can't leverage Context Propagation for this as its not part of the platform yet – we will when it does make it to the platform.
* If we bend over backwards now to support this use case, going forward, we will have to deal with complexity once Context Propagation is available.
* Yet, it does make sense to propagate the context in this scenario, but if we are doing that, we should evaluate all cases across the board, rather than a single one.
* The original issue was reported by QE, not a real world user use case; so e.g. introducing this in XP2 is a behavior change which might not be expected.

I think it boils down to what users are expecting to happen and @Ladicek 's comment "I'd suggest that if users provision Fault Tolerance and OpenTracing, but don't provision Context Propagation, then they should NOT expect traces being propagated across FT @Asynchronous calls." suggests that we shouldn't move forward with this.

bstansberry:
Per the discussion starting at https://github.com/wildfly/wildfly/pull/13590#discussion_r492612456 we've decided not to proceed with this approach and we'll defer dealing with WFLY-13866 issue until we have a more general idea of how MP Context Propagation will integrate (e.g. https://issues.redhat.com/browse/WFLY-13921), or at least a broader review of places where the same kind of thing would be needed.

PR #13595, opened 2020-09-22
[WFLY-13884] JdrReportManagmentTestCase fails when using the wildfly-…

yersan:
…ee-galleon pack

This patch adds a new entry to the JDR report to store the product directory name and get the rid of the test regular expression used to get entries under sos_string. Another option could be just navigate to the sos_string first subfolder but I thought adding the entry from the report could be more secure to be more confident with the report content under test.

Jira issue: https://issues.redhat.com/browse/WFLY-13884

### DIFF
--- a/jdr/jboss-as-jdr/src/main/java/org/jboss/as/jdr/JdrRunner.java
+++ b/jdr/jboss-as-jdr/src/main/java/org/jboss/as/jdr/JdrRunner.java
@@ -108,6 +108,7 @@ public class JdrRunner implements JdrReportCollector {
             }
             versionWriter.close();
             this.env.getZip().add(new ByteArrayInputStream(versionStream.toByteArray()), "version.txt");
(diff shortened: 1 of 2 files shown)


jdr/jboss-as-jdr/src/main/java/org/jboss/as/jdr/JdrRunner.java
```diff
@@ -108,6 +108,7 @@ public JdrReport collect() throws OperationFailedException {
             }
             versionWriter.close();
             this.env.getZip().add(new ByteArrayInputStream(versionStream.toByteArray()), "version.txt");
+            this.env.getZip().add(new ByteArrayInputStream(this.env.getZip().getProductDirName().getBytes(StandardCharsets.UTF_8)), "product.txt");
```
bstansberry, line 111: @bmaxwell Any objections to this? It's adding a file to the JDR zip.

bmaxwell: No issues

PR #13597, opened 2020-09-23
[WFLY-13891]:Make security service dependency optional for webservice…

jimma:
https://issues.redhat.com/browse/WFLY-13891


bstansberry:
@darranl FYI

@jimma So AIUI the goal here is to allow a completely unsecured webservice? Just confirming.

jimma:
@bstansberry  Yes. This change is to fix failures when webservice failed in server installation with webservices galleon layer which depends on web-server layer. It generates the standalone.xml like :https://gist.github.com/jimma/82ce0749c55a2d5b616e170cbe7d5b9c and undertow doesn't have elytron security reference.

bstansberry (approved):
@jimma @darranl This looks ok, but it led me to file https://issues.redhat.com/browse/WFLY-13928.

bstansberry:
Thanks, @jimma

PR #13599, opened 2020-09-23
[WFLY-13802] Minor fixes for Undertow deployments to allow web applications to be deployed when the EE subsystem is not installed.

darranl:
https://issues.redhat.com/browse/WFLY-13802

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/ComponentRegistry.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/ComponentRegistry.java
@@ -71,9 +71,14 @@ public class ComponentRegistry {
     }
 
     public ManagedReferenceFactory createInstanceFactory(final Class<?> componentClass) {
+        return createInstanceFactory(componentClass, false);
+    }
+
+    public ManagedReferenceFactory createInstanceFactory(final Class<?> componentClass, final boolean optional) {
         final ManagedReferenceFactory factory = componentsByClass.get(componentClass);
         if (factory == null) {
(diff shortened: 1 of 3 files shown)


bstansberry:
@darranl This seems ok but perhaps shouldn't resolve the JIRA or maybe we need a follow-on JIRA. I haven't built with just the 'undertow' layer to confirm, but I just had a look and believe it will add the servlet-container resource. A servlet container without the ee subsystem (which brings in naming) seems wrong. 

There are valid uses for a non-servlet-container HTTP server; undertow-load-balancer is an example and there may be others that are reasonably attainable. But the 'undertow' layer doesn't seem to be that. As part of our Galleon docs work we should figure out what it is. :)

I'm not opposed to this change though.

bstansberry:
I guess a servlet container + cdi with no ee/naming is an understandable use case.

darranl:
I think there are a few layers that need a more in-depth analysis, elytron is one and it seems undertow may be a second.

With this change it also makes it possible to deploy wars containing static content which may not require other EE relates resources, also simple servlet based deployments are fine as long as they don't require the injection of other EE resources.

There is still a down side here in that the undertow layer still brings in the module for the EE subsystem although I don't think that is as big a deal as the security subsystem is being removed which was why we wanted to prevent accidental inclusion.

For both elytron and undertow I think we may have some distinct use cases that justify a couple of separate subsystems, otherwise we end up with only partly usable subsystems which is harder to understand.  I definitely don't think we want an exercise of breaking all subsystems down but the reason I mention elytron and undertow is they both provide a foundation for other layers which means they both have a set of key use cases they "could" provide.

PR #13600, opened 2020-09-23
[WFLY-13807] EE: Add component class info to interface based proxy to preserve validation metadata

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-13807

This PR supercedes https://github.com/wildfly/wildfly/pull/13526, and fixed the 2 test failures in #13526

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/ViewService.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/ViewService.java
@@ -38,6 +38,7 @@ import org.jboss.invocation.InterceptorContext;
 import org.jboss.invocation.InterceptorFactory;
 import org.jboss.invocation.Interceptors;
 import org.jboss.invocation.SimpleInterceptorFactoryContext;
+import org.jboss.invocation.proxy.MethodIdentifier;
 import org.jboss.invocation.proxy.ProxyFactory;
 import org.jboss.msc.inject.Injector;
 import org.jboss.msc.service.Service;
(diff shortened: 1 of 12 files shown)


ee/src/main/java/org/jboss/as/ee/component/ViewService.java
```diff
@@ -193,11 +195,24 @@ public Object invoke(InterceptorContext interceptorContext) throws Exception {
...
             return interceptor.processInvocation(interceptorContext);
         }
 
+        private Method matchMethod(Method invokedMethod) {
+            if (invokedMethod.getDeclaringClass().isInterface()) {
+                for (Method beanMethod : proxyFactory.getCachedMethods()) {
+                    if (MethodIdentifier.getIdentifierForMethod(beanMethod).equals(MethodIdentifier.getIdentifierForMethod(invokedMethod))) {
+                        return beanMethod;
+                    }
+                }
```
bstansberry, line 209: This is on the request path and should be optimized to a single map lookup.

bstansberry: I was referring to the entire for loop.

chengfang: This path is only executed when `invokedMethod.getDeclaringClass().isInterface()` is true, i.e., when the bean is 2.x ejb, or the ejb bean class has no default constructor. So for the majority of ejb use case, this part is not touched. So I updated this part to populate a map on-demand to save the mapping.

bstansberry: Nice; I like that approach.

bstansberry (approved):

PR #13609, opened 2020-09-28
WFLY-13911 Intermittent deployment failures in clustering testsuite on Linux

pferraro:
https://issues.redhat.com/browse/WFLY-13911

### DIFF
--- a/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/ThreadPoolServiceConfigurator.java
+++ b/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/ThreadPoolServiceConfigurator.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ThreadFactory;
 
 import org.infinispan.commons.executors.BlockingThreadPoolExecutorFactory;
 import org.infinispan.commons.executors.ThreadPoolExecutorFactory;
+import org.infinispan.commons.util.ProcessorInfo;
 import org.infinispan.configuration.global.ThreadPoolConfiguration;
 import org.infinispan.configuration.global.ThreadPoolConfigurationBuilder;
 import org.jboss.as.clustering.context.DefaultThreadFactory;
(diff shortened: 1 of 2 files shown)


clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/ThreadPoolServiceConfigurator.java
```diff
@@ -61,7 +61,9 @@ public ServiceConfigurator configure(OperationContext context, ModelNode model)
         long keepAliveTime = this.definition.getKeepAliveTime().resolveModelAttribute(context, model).asLong();
         boolean nonBlocking = this.definition.isNonBlocking();
         if (this.definition == ThreadPoolResourceDefinition.NON_BLOCKING) {
-            maxThreads *= ProcessorInfo.availableProcessors();
+            int availableProcessors = ProcessorInfo.availableProcessors();
+            minThreads *= availableProcessors;
+            maxThreads *= availableProcessors;
```
bstansberry, line 66: The read-resource-description text description of these 2 attributes should mention that this multiplication occurs. (I think continuing to share the text across pool types is ok; the text can mention that the behavior is only for a non-blocking pool.)

I don't think this point should block this change getting in for WF 21 if that is wanted; it can come later.

bstansberry (approved):
I have one suggestion but that should not block this.

As WF 21 is the first appearance of this resource type we should get this default value change in so we don't have to do an API version bump/transformer in WF 22.

bstansberry:
@pferraro ping just in case you don't see my comments above otherwise.

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/ThreadPoolServiceConfigurator.java
pferraro: That's valid.  I've opened https://issues.redhat.com/browse/WFLY-13927 to track this.

PR #13616, opened 2020-09-30
[WFLY-13917] The org.jboss.resteasy.resteasy-crypto module requires t…

jamezp:
…he bouncy castle dependencies be exported.

https://issues.redhat.com/browse/WFLY-13917

### DIFF
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcmail/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/bouncycastle/bcmail/main/module.xml
@@ -36,6 +36,6 @@
         <module name="javax.mail.api" optional="true"/>
         <module name="javax.activation.api" optional="true"/>
         <module name="org.bouncycastle.bcpkix"/>
-        <module name="org.bouncycastle.bcprov"/>
+        <module name="org.bouncycastle.bcprov" services="import"/>
     </dependencies>
 </module>
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-crypto/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-crypto/main/module.xml
@@ -22,7 +22,7 @@
   ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
   -->
 
-<module xmlns="urn:jboss:module:1.5" name="org.jboss.resteasy.resteasy-crypto">
+<module xmlns="urn:jboss:module:1.9" name="org.jboss.resteasy.resteasy-crypto">
     <properties>
         <property name="jboss.api" value="private"/>
     </properties>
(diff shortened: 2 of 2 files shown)


darranl:
It sounds like these modules may need a review, if resteasy-crypto is being added to the classpath of the deployment this means private API is being made available (This isn't the only case, just the one I happen to see today).

Instead does the deployment really need these modules or is RestEasy relying in the ClassLoader of the deployment when really it should be using it's own?

jamezp:
> Instead does the deployment really need these modules or is RestEasy relying in the ClassLoader of the deployment when really it should be using it's own?

This I'm not entirely sure of. I don't really know how the providers are loaded, but they seem to be processed on the deployments class loader. I'm not yet familiar enough with JAX-RS or RESTEasy to know if that is correct or not. Maybe @asoldano or @ronsigal have an idea there.

asoldano:
I'm thinking we should put this on hold for a while, as it's really an improvement to avoid requiring users to add the dependency manually to their deployment, right? I think we need to evaluate which are the documented usage scenarios of resteasy-crypto and verify if the BC api are really needed for those or not. @liweinan has some experience on this, but he will be back after the holidays in China are over.

darranl:
Also if we are going to have a review I think we should review what is the relationship between the cryptographic requirements of Rest Easy and the security services provided by the application server.

From the perspective of the security team we are moving to focus more on the streamlining of common security requirements so will likely be exploring our relationship with JAX-RS quite a bit over the next 12 months.

jamezp:
My main concern is for a user to use a `multipart/signed` content type for an accept header they have to add a dependency to a private module. It's been like this forever as far as I can tell, but seems wrong to me. An example is the [MultipartSignedWriter](https://github.com/resteasy/Resteasy/blob/master/security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/smime/MultipartSignedWriter.javal) which directly uses Bouncy Castle.

As to why it needs to be on the deployment class path, that part I do not know. My guess was it had something to do with it being a provider. That is a guess though as I'm still learning the JAX-RS spec and RESTEasy :)

bstansberry:
I generally don't like the 'hold' label as it leads to PRs that should progress getting ignored, but I'll put it on this one given the open discussion above and the fact that all the participants have the perms to remove it when appropriate. :)

darranl:
Reviewing bouncy castle use is something we want to look at this year, maybe now is time to resurrect this discussion?

darranl:
@jamezp this looks like a stalled one to catch up on soon.

darranl:
@jamezp / @fjuma This maybe sounds like a discussion for the new year to check what is really needed here with the other work that is happening?

jamezp:
+1 I'll take another look too as I know RESTEasy better now :)

fjuma:
@jamezp Is this PR still relevant? There haven't been any updates on it in a long time and there are also conflicts that need to be resolved.

jamezp:
I'll go ahead and close it. I would imagine it still doesn't work, but doesn't seem like it's an issue for anyone.

PR #13619, opened 2020-10-02
Rollback ejb client to 4.0.33.Final.

pferraro:
An experiment to see if this upgrade is causing TwoConnectorsEJBFailoverTestCase to hang.


pferraro:
Results from several CI runs seems to confirm that rolling back this upgrade resolves the hanging issues with TwoConnectorsEJBFailoverTestCase.  I've removed the hold label.

bstansberry (approved):

PR #13620, opened 2020-10-02
[WFLY-13632] [WFLY-13633] Deploying CMR resources is not thread safe

mmusgrov:
Issue: https://issues.redhat.com/browse/WFLY-13633
Issue: https://issues.redhat.com/browse/WFLY-13632

This PR replaces pr #13525. The new PR pulls in the commit made by @gaol plus two extra commits that force a restart-required (see the commits in the PR by @bstansberry) plus one commit by myself to add a manual mode test for Brian's commits.

The reason for Brian's two commits is:

The first triggers restart-required for add or remove, but does not change the management API description. The current state is the API description is wrong, plus the behaviour is wrong. With this commit the behaviour is okay, description remains wrong.

The second commit fixes the description and bumps the API version accordingly.

Brian did this split in case we want to port this to EAP 7.3.  Bumping the API versions in a CP is very problematic; so much so that we have a rule against it. So if we do port this we can leave off the last commit.


bstansberry (approved):
This LGTM.

Looking more carefully at the test, I think the same kind of thing could run in the 'transactions' maven module testsuite, which would execute faster. But this is fine.

gaol:
I am OK with this PR as well, thanks :)

bstansberry:
Thanks, @mmusgrov and @gaol !

PR #13621, opened 2020-10-02
WFLY-13800 Upgrade Infinispan to 11.0.4.Final

pferraro:
https://issues.redhat.com/browse/WFLY-13800

Also resolves:
WFLY-13810 Intermittent failure in ExternalizerTestCase
https://issues.redhat.com/browse/WFLY-13810

### DIFF
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -45,7 +45,7 @@
         <surefire.forked.process.timeout>5400</surefire.forked.process.timeout>
         <test-group>org/jboss/as/test/clustering/cluster</test-group>
         <version.org.jboss.byteman>4.0.6</version.org.jboss.byteman>
-        <version.org.infinispan.server>11.0.3.Final</version.org.infinispan.server>
+        <version.org.infinispan.server>11.0.4.Final</version.org.infinispan.server>
         <ts.surefire.clustering.ha.additionalExcludes>nothing-by-default</ts.surefire.clustering.ha.additionalExcludes>
     </properties>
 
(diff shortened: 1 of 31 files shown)


testsuite/integration/clustering/pom.xml
```diff
@@ -45,7 +45,7 @@
         <surefire.forked.process.timeout>5400</surefire.forked.process.timeout>
         <test-group>org/jboss/as/test/clustering/cluster</test-group>
         <version.org.jboss.byteman>4.0.6</version.org.jboss.byteman>
-        <version.org.infinispan.server>11.0.3.Final</version.org.infinispan.server>
+        <version.org.infinispan.server>${version.org.infinispan}</version.org.infinispan.server>
```
rhusar, line 48: This essentially removes a fix for https://issues.redhat.com/browse/JBEAP-20157 - please don't do that, just bump the version here.

rhusar (changes requested):

testsuite/integration/clustering/pom.xml
pferraro: Not really.  The version for version of the server can be set independently.

rhusar: Yes, but that would require every prod build to be updated and kept in sync which is not feasible.

testsuite/integration/clustering/pom.xml
```diff
@@ -45,7 +45,7 @@
         <surefire.forked.process.timeout>5400</surefire.forked.process.timeout>
         <test-group>org/jboss/as/test/clustering/cluster</test-group>
         <version.org.jboss.byteman>4.0.6</version.org.jboss.byteman>
-        <version.org.infinispan.server>11.0.4.Final</version.org.infinispan.server>
+        <version.org.infinispan.server>11.0.3.Final</version.org.infinispan.server>
```
rhusar: We really do need to use 11.0.4.Final - because 11.0.3.Final suffers from the shutdown problem fixed by https://issues.redhat.com/browse/ISPN-12374.

rhusar (changes requested):

rhusar (approved):
Looks good now!

bstansberry (approved):

PR #13622, opened 2020-10-02
[WFLY-13932] Upgrade Bouncy Castle to 1.66

jamezp:
https://issues.redhat.com/browse/WFLY-13932

https://www.bouncycastle.org/releasenotes.html

The core upgrade is https://github.com/wildfly/wildfly-core/pull/4350. These should not rely on each other though.


darranl:
@jamezp I think we can close this one?  I have added the following PR (and another in core) to make sure the versions are defined together:
    https://github.com/wildfly/wildfly/pull/13780

At the moment by having the version defined in two locations the WildFly build is producing a distribution mixing both BouncyCastle versions even though there are dependencies between the modules.

PR #13627, opened 2020-10-06
[WFLY-13807] EJB validation: don't normalize if proxy is inteface based

tadamski:
https://issues.redhat.com/browse/WFLY-13807

This fix allows for correct validation of:
Class based resources with generic parameters (WFLY-11566)
Interface based resources (WFLY-13807)
As a result, it fixes the bugs that we have JIRAs for.

Interface based resources with generic parameters is the scenario that is going to fail but currently I can't think of any solution that will fix it.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/validation/EjbBeanValidationTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/validation/EjbBeanValidationTestCase.java
@@ -26,6 +26,7 @@ import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.as.test.integration.common.HttpRequest;
+import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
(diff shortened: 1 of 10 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/validation/EjbBeanValidationTestCase.java
```diff
@@ -60,10 +65,9 @@
         return ShrinkWrap.create(WebArchive.class, "ejbvalidation.war")
                 .addPackage(HttpRequest.class.getPackage())
                 .addClasses(
-                        EjbBeanValidationTestCase.class,
-                        TestApplication.class,
-                        TestResource.class
-                );
+                        EjbBeanValidationTestCase.class, TestApplication.class, TestResource.class, EchoResourceImpl.class, EchoResource.class,
+                        DummySubclass.class, DummyAbstractClass.class, DummyClass.class, DummyFlagImpl.class, DummyFlag.class)
+                .addAsWebInfResource("web.xml");
```
chengfang, line 70: `addAsWebInfResource("web.xml")` caused deployment failure, since there is no such file. This test case does not need web.xml so this line can be deleted.

tadamski: fixed

chengfang (approved):

gaol:
@bstansberry @chengfang any chance to merge it and verify that there is no tck failures with this fix? Thanks :)

PR #13632, opened 2020-10-07
DRAFT test WFLY-13894 Upgrade jboss-ejb-client to 4.0.35.Final

chengfang:
This is a DRAFT pull to test the upgrade of jboss-ejb-client from ejb-client-4.0.33 to ejb-client-4.0.35 (latest), to check if all CI jobs pass.


chengfang:
Restarted `Linux SM - JDK 8` job, which previously failed with 100+ failures, now passed.

pferraro:
In that failed "Linux SM - JDK 8" job, the org.jboss.as.test.clustering.cluster.ejb.remote.TwoConnectorsEJBFailoverTestCase hung.  This caused the remaining 100+ tests to fail, as the server could not be started (because of the previous hung process).
This was the reason we rolled back this release.

PR #13638, opened 2020-10-09
Remove duplicate dependency in ejb3 module

rhusar:
Causes:

```
[WARNING] Some problems were encountered while building the effective model for org.wildfly:wildfly-ejb3:jar:21.0.0.Final-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.wildfly.wildfly-http-client:wildfly-http-ejb-client:jar -> duplicate declaration of version (?) @ line 249, column 21
```

### DIFF
--- a/ejb3/pom.xml
+++ b/ejb3/pom.xml
@@ -247,11 +247,6 @@ projects that depend on this project.-->
         </dependency>
 
         <dependency>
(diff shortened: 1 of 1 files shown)


pferraro (approved):

bstansberry:
Thanks, @rhusar !

PR #13640, opened 2020-10-09
[WFLY-13953] Upgrade WildFly Arquillian to 3.0.1.Final

jamezp:
https://issues.redhat.com/browse/WFLY-13953

This includes #13639. This also may require a new JIRA.


bstansberry:
@jamezp Needs a rebase. We should probably get this in fairly soon so it bakes.

bstansberry:
@jamezp Is there any reason not to move forward with this?

jamezp:
Other thank I forgot about it? Nope :)

PR #13653, opened 2020-10-16
WFLY-13973 Writing default-mdb-instance-pool to ejb3 subsystem throws…

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-13973

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemDefaultPoolWriteHandler.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemDefaultPoolWriteHandler.java
@@ -22,11 +22,17 @@
 
 package org.jboss.as.ejb3.subsystem;
 
+import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.DEFAULT_MDB_INSTANCE_POOL;
+import static org.jboss.as.ejb3.subsystem.EJB3SubsystemModel.DEFAULT_SLSB_INSTANCE_POOL;
+import static org.jboss.as.ejb3.subsystem.EJB3SubsystemRootResourceDefinition.DEFAULT_MDB_POOL_CONFIG_CAPABILITY;
+import static org.jboss.as.ejb3.subsystem.EJB3SubsystemRootResourceDefinition.DEFAULT_SLSB_POOL_CONFIG_CAPABILITY;
+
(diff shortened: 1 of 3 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemDefaultPoolWriteHandler.java
```diff
@@ -78,44 +81,63 @@ protected void recordCapabilitiesAndRequirements(OperationContext context, Attri
...
+                        String newSLSBRequirementName = RuntimeCapability.buildDynamicCapabilityName(STRICT_MAX_POOL_CONFIG_CAPABILITY_NAME, resolvedNewValue.asString());
+                        context.registerAdditionalCapabilityRequirement(newSLSBRequirementName, DEFAULT_SLSB_POOL_CONFIG_CAPABILITY_NAME, DEFAULT_SLSB_INSTANCE_POOL);
+                    }
+                } catch (OperationFailedException ofe) {
+                    // if the new value cannot be resolved, deregister the old value only (in the finally clause)
+                    EjbLogger.ROOT_LOGGER.defaultPoolExpressionCouldNotBeResolved(DEFAULT_SLSB_INSTANCE_POOL, model.get(DEFAULT_SLSB_INSTANCE_POOL).asString());
+                } finally {
+                    // de-register the old requirement
+                    String oldSLSBRequirementName = RuntimeCapability.buildDynamicCapabilityName(STRICT_MAX_POOL_CONFIG_CAPABILITY_NAME, oldValue.asString());
+                    context.deregisterCapabilityRequirement(oldSLSBRequirementName, DEFAULT_SLSB_POOL_CONFIG_CAPABILITY_NAME, DEFAULT_SLSB_INSTANCE_POOL);
```
kabir, line 102: @chengfang I think if you execute a :write-attribute with the same value that was there, you end up removing the capability requirement set a few lines above. Perhaps you need some guard to check that the new and old requirement names are not the same?

Essentially AbstractWriteAttributeHandler doesn't do any comparison itself of the old and new values to check

chengfang: I added a check to compare old and new requirement name, and only de-register it if they differ. I did it for both default-slsb-instance-pool and default-mdb-instance-pool.

kabir (approved):

PR #13662, opened 2020-10-20
[WFLY-13132] Small fix to test case for TwoConnectorsEJBFailoverTestCase

rachmatowicz:
This PR does the following:
- adds a missing sleep delay to the existing text case TwoConnectorsEJBFailoverTestCase 

For more details: see https://issues.redhat.com/browse/WFLY-13132


rachmatowicz:
@chengfang Here is the PR for the change you requested.

chengfang (approved):

PR #13672, opened 2020-10-23
WFLY-14002 Add a property for the galleon-maven-plugin offline config…

xstefank:
…uration

https://issues.redhat.com/browse/WFLY-14002


bstansberry:
This seems ok but I think there are many more provisioning executions than are covered here.

xstefank:
@bstansberry thank you. Ready for review again.

bstansberry:
Thanks, @xstefank !

PR #13674, opened 2020-10-24
[WFLY-14008] Have EJB depend an internal variant of the JAX-RPC API m…

bstansberry:
…odule so we can drop the public one in the EE9 dist

https://issues.redhat.com/browse/WFLY-14008


bstansberry:
@chengfang @tadamski Is this ok? Doing this will let us fix some EE 9 TCK issues without having to expose the JAX-RPC module to deployments in the EE 9 server.

tadamski (approved):
@bstansberry It looks good to me.

PR #13675, opened 2020-10-27
[WFLY-13695] Upgrade Wildfly HTTP Client to 1.1.4.Final

gaol:
Issue: https://issues.redhat.com/browse/WFLY-13695

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -447,7 +447,7 @@
         <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
         <version.org.wildfly.core>14.0.0.Beta3</version.org.wildfly.core>
         <version.org.wildfly.extras.creaper>1.6.1</version.org.wildfly.extras.creaper>
-        <version.org.wildfly.http-client>1.1.2.Final</version.org.wildfly.http-client>
+        <version.org.wildfly.http-client>1.1.4.Final</version.org.wildfly.http-client>
         <version.org.wildfly.naming-client>1.0.13.Final</version.org.wildfly.naming-client>
         <version.org.wildfly.transaction.client>1.1.13.Final</version.org.wildfly.transaction.client>
         <version.org.yaml.snakeyaml>1.26</version.org.yaml.snakeyaml>


gaol:
Linux with SM test is hanging with `failed to deploy exception`.

bstansberry:
@gaol @tadamski It looks like the problem is in org.jboss.as.test.multinode.clientinterceptor.multiple.MultipleClientInterceptorTestCase which looks to be having a problem on the 'multinode-client' server. The logs for that server show what looks like a reload happening after the test deployment is undeploy, which is consistent with what the test class code says -- there's a ServerSetupTask that does a reload in tearDown.

That reload doesn't seem to progress. It gets to "INFO  [org.wildfly.extension.undertow] (MSC service thread 1-5) WFLYUT0004: Undertow 2.2.2.Final stopping" and then stops logging until finally there's logging of the container being stopped by OS signal a couple minutes later.

The build log shows that test taking > 100 seconds (2-3 secs is normal) but not failing. The next test fails when Arquillian is unable to deploy the test deployment to the 'multinode-client' server.  Failures continue after that.

Why this upgrade would lead to this result I have no idea. But it seems to happen consistently with this PR.

gaol:
@tadamski @bstansberry some tests result in my local:

* I can reproduce the hangs consistently using command `mvn clean install -Dversion.org.wildfly.http-client=1.1.3.Final -Dtest=EjbOver*TestCase -Dsecurity.manager=true` under directory: `testsuite/integration/multinode/`. This will execute 3 test cases. the client node stack trace: https://gist.github.com/gaol/43d51d5edef68b6180f35e68540f3a0a, each time it hangs at test: `org.jboss.as.test.multinode.ejb.http.EjbOverHttpTestCase`. Once, the `EjbOverHttpTestCase` finished without errors, but mvn command did not terminate.

* I **cannot** reproduce the hangs using command `mvn clean install -Dversion.org.wildfly.http-client=1.1.3.Final -Dtest=EjbOver*TestCase` under directory: `testsuite/integration/multinode/`. This will execute 3 test cases as well **without** enabling security manager.

* I **cannot** reproduce the hangs using command `mvn clean install -Dversion.org.wildfly.http-client=1.1.3.Final -Dtest=EjbOverHttpTestCase -Dsecurity.manager=true` under directory: `testsuite/integration/multinode/`. This will execute 1 test case as well **with** enabling security manager.

* I **cannot** reproduce the hangs without upgrading wildfly-http-client in above cases.

gaol:
https://issues.redhat.com/browse/WEJBHTTP-48 is created to track the issue

fl4via:
The failures in Windows seem to be unrelated, but I'm not succeeding in triggering a new run: "Unable to collect changes" is the error that is showing up in TeamCity when I click on Run.

gaol:
Thank you @fl4via for the checking !
@bstansberry would you please double check if this is good to upgrade in wildfly? there is a downstream upgrade relying on approval of this one, thanks :)

fl4via:
Never mind, the problem I mentioned is gone. I triggered a new Windows JDK11 build.

PR #13676, opened 2020-10-28
WFLY-13998 Fix handling of mutable distributable session attributes across concurrent requests when backing cache is non-transactional

pferraro:
https://issues.redhat.com/browse/WFLY-13998


luck3y:
Note there may be an issue with status reporting from CI currently, aggregator job is here: https://ci.wildfly.org/buildConfiguration/WF_PullRequestAggregator/228150?buildTab=dependencies&mode=timeline&range=1603892394000-1603904731000

PR #13679, opened 2020-10-29
[WFLY-14018] Allow injection for Validator/Converter/Behavior if managed=true

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-14018

Minor change to allow injection of Validator/Converter/Behavior if the `managed` property is `true`. If it's one of those components and the property is true the exception is not thrown (continue) and weld tries to inject that object as usual.

@fjuma check this when you have time. As in the previous JSF issue no hurries at all. With this fix included I can run all the mojarra TS that is executed when a PR is sent in upstream (eclipse).

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFAnnotationProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFAnnotationProcessor.java
@@ -32,6 +32,7 @@ import org.jboss.as.server.deployment.annotation.CompositeIndex;
 import org.jboss.as.web.common.ServletContextAttribute;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.modules.Module;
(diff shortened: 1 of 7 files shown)


jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFAnnotationProcessor.java
```diff
@@ -50,6 +50,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.jboss.jandex.AnnotationValue;
```
rhusar, line 53: We keep imports sorted to avoid unnecessary conflicts, see where `org.jboss...` is imported above.

rhusar (changes requested):

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFAnnotationProcessor.java
rmartinc: Thanks @rhusar, I'll move it upward tomorrow.

rmartinc: Done!

fjuma:
@rmartinc Thanks for working on this! The changes look good. Would it be possible to get a test added to the WildFly testsuite as well?

rmartinc:
@fjuma Yep, it's possible, I didn't add it because it was in the mojarra test-suite and I didn't want to repeat it. I'll work on it and update the PR with the test.

rmartinc:
@fjuma The reproducer is added as a JSF basic test case now.

fjuma (approved):
Thanks @rmartinc!

PR #13686, opened 2020-11-03
WFLY-14039 Obtain ComponentView with AccessController.doPrivileged

boris-unckel:
Adding the missing AccessController.doPrivileged if a SecurityManager is present. Fixing https://issues.redhat.com/browse/WFLY-14039

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/invocation/AbstractInvocationHandler.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/invocation/AbstractInvocationHandler.java
@@ -27,6 +27,8 @@ import static org.jboss.as.webservices.util.ASHelper.getMSCService;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.UndeclaredThrowableException;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Collection;
 import java.util.concurrent.Callable;
 
@@ -81,7 +83,13 @@ abstract class AbstractInvocationHandler extends org.jboss.ws.common.invocation.
             synchronized (this) {
                 cv = componentView;
                 if (cv == null) {
-                    cv = getMSCService(componentViewName, ComponentView.class);
+                    SecurityManager sm = System.getSecurityManager();
+                    if (sm == null) {
+                        cv = getMSCService(componentViewName, ComponentView.class);
+                    } else {
+                        cv = AccessController.doPrivileged(
+                                (PrivilegedAction<ComponentView>) () -> getMSCService(componentViewName, ComponentView.class));
+                    }
                     if (cv == null) {
                         throw WSLogger.ROOT_LOGGER.cannotFindComponentView(componentViewName);
                     }


bstansberry:
/ok-to-test

@jimma Please review.

@boris-unckel Is something catching and suppressing the SecurityException? It's interesting that this hasn't resulted in any tests failing when we run the testsuite with the security manager enabled.

boris-unckel:
> @boris-unckel Is something catching and suppressing the SecurityException? It's interesting that this hasn't resulted in any tests failing when we run the testsuite with the security manager enabled.
Unfortunately I could not manage to figure it out, we just have the stacktrace created for trace purpose. There is one interesting point why I wrote the patch: When we provided the  getCurrentServiceContainer by permission.xml in the EAR, it still failed. When we put it into the minimum-permission-set of the container it worked. We neither have observed this behaviour with any other Permission nor the same permission at a different code of call.

bstansberry:
Thanks, @boris-unckel. That's interesting. I'm not sure what to make of that! @darranl @fjuma FYI.

boris-unckel:
> Thanks, @boris-unckel. That's interesting. I'm not sure what to make of that! @darranl @fjuma FYI.
@bstansberry 
It's in the same class, my apologies for the too fast answer:
`   protected void handleInvocationException(final Throwable t) throws Exception {
      if (t instanceof MBeanException) {
         throw ((MBeanException) t).getTargetException();
      }
      if (t instanceof Exception) {
         if (t instanceof InvocationTargetException) {
            throw (Exception) t;
         } else if (t instanceof SecurityException) {
             throw (Exception) t;
         }  else {
            SOAPFaultException ex = findSoapFaultException(t);
            if (ex != null) {
               throw new InvocationTargetException(ex);
            }
            throw new InvocationTargetException(t);
         }
      }
      if (t instanceof Error) {
         throw (Error) t;
      }
      throw new UndeclaredThrowableException(t);
    }`

The part with `       } else if (t instanceof SecurityException) {
             throw (Exception) t;
         } `
is missing. 
How to deal with this? Provide it as second commit in the same pull request or new pull request?
The AccessController is still necessary, the correct throw does not fix the reason for the missing permission.

bstansberry:
@boris-unckel Thanks. I'll defer to @jimma (webservices lead for WildFly) re what to do about that. I expect it's not a problem, though. The code seems designed to wrap things in InvocationTargetException and rethrow and that isn't necessarily wrong. If it was suppressing the exception (i.e. not wrapping and rethrowing) that would concern me more.

Looking closer at the JIRA I also have a better sense of why we may not have seen this when running our testsuite. Looking at the stack trace I see non-WildFly-provided code -- at a glance some servlet filters. So that's the code that is lacking privileges. And it's not typical deployment code (a Servlet or EJB or something) so our testsuite is less likely to set up such a scenario.

I think your patch looks fine, but I'll give a bit of time for Jim to comment if he'd like before merging.

bstansberry:
@boris-unckel I need to ask a question before merging your first patch:

boris-unckel:
> @boris-unckel I need to ask a question before merging your first patch:
> 
> We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?
Yes, I agree to the terms of the MIT License for all my contributions.

boris-unckel:
> @boris-unckel Thanks. I'll defer to @jimma (webservices lead for WildFly) re what to do about that. I expect it's not a problem, though. The code seems designed to wrap things in InvocationTargetException and rethrow and that isn't necessarily wrong. If it was suppressing the exception (i.e. not wrapping and rethrowing) that would concern me more.

Yes, in terms of visibility you're right. Looking at the Java Security Manager usecase, I would prefer that any SecurityExeption breaks the run of the application. Therefor it should not be handled like application logic.

> Looking closer at the JIRA I also have a better sense of why we may not have seen this when running our testsuite. Looking at the stack trace I see non-WildFly-provided code -- at a glance some servlet filters. So that's the code that is lacking privileges. 
Still I think the point is that one cannot grant the privileges, just for this special case.

jimma:
@boris-unckel  The change looks good to me.  
And I agree with @bstansberry,  it's good to wrap this exception and throw it in AbstractInvocationHandler.  Apache cxf's ServiceInvokerInterceptor will finally look at these wrapped exception , convert them to soap fault and returns to client.

darranl:
On that last point we should not really be looking to get internal security exceptions into a format to send to a remote client, errors such as this have no relationship with the client so the client does not need to know about an internal security failure.

Log and throw is generally a bad practice but when it comes to security exceptions such as this we should be logging to allow diagnosis of the problem and if we can not proceed converting to a more appropriate error which is suitable for the client.

jimma:
@darranl  Good point. 
@boris-unckel  What's the client side fault message ?  Is there some detailed info has been exposed to client side ?

bunckelvhv:
> @boris-unckel What's the client side fault message ? Is there some detailed info has been exposed to client side ?

@jimma I'm answering with my business account, beeing at work now. We can't relate that to a specific client side issue. It's a multi instance, multi server environment and the message had no traceid to follow it across server borders. I just know for sure that the patch is resolving the issue.

In general I would prefer to have SecurityExceptions breaking the application: In dev/test you provide a permission, the WildFily Security Manager is in log-only mode. In production you generate an alert in the IDS out of the exception, the WildFly Security manager is active. Aside from false/positive or misconfiguration: A SecurityException happens because the system is under attack.

bstansberry:
> 
> > Looking closer at the JIRA I also have a better sense of why we may not have seen this when running our testsuite. Looking at the stack trace I see non-WildFly-provided code -- at a glance some servlet filters. So that's the code that is lacking privileges.
> 
> Still I think the point is that one cannot grant the privileges, just for this special case.

Yes, agreed. It was just bothering me a bit why this wasn't triggering testsuite failures, as maybe that meant something about the bug wasn't understood. But I understand why our testsuite might not be setting up this call pattern.

bstansberry:
@jimma @boris-unckel @darranl How would you like to proceed with this? From the discussion of the exception handling it sounds like something should be done. Shall we log a JIRA and pursue that via a separate issue? Or do it here?

My instinct is a separate issue is better and changing exception handling seems like something that needs careful consideration, while this problem of a missing privileged block is quite specific and straightforward.

boris-unckel:
> @jimma @boris-unckel @darranl How would you like to proceed with this? From the discussion of the exception handling it sounds like something should be done. Shall we log a JIRA and pursue that via a separate issue? Or do it here?

I will file a separate JIRA, provide a separate patch with the proposed exception handling (see above). Maybe the testsuite with activated JSM reacts.
> 
> My instinct is a separate issue is better and changing exception handling seems like something that needs careful consideration, while this problem of a missing privileged block is quite specific and straightforward.

The solution with a privileged block does not fade away just because we do some more detailed analysis.

boris-unckel:
@jimma Please check [WFLY-14072](https://issues.redhat.com/browse/WFLY-14072) and [PL 13703](https://github.com/wildfly/wildfly/pull/13703)

bstansberry:
Thanks, @boris-unckel!

PR #13688, opened 2020-11-05
[WFLY-14040] Fully specify the 'ee' layer in the wildfly-ee feature p…

bstansberry:
…ack variant

https://issues.redhat.com/browse/WFLY-14040

@jfdenise Please review.

### DIFF
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/ee/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/ee/layer-spec.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" ?>
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="ee">
+    <dependencies>
+        <layer name="naming"/>
+        <layer name="jsonb" optional="true"/>
+    </dependencies>
+    <feature-group name="ee"/>
     <packages>
        <!-- ee optional injected modules in DU that are located in full FP -->
        <package name="javax.resource.api" optional="true"/>


jfdenise:
@bstansberry , I don't fully get the picture. You plan to copy the servlet fp src content into the ee one? The dependencies (naming, ee feature-group) that are contained in the layer come from servlet feature-pack. You will have to do the same for the undertow layer BTW.

bstansberry:
@jfdenise Yes, the source is copied in. See the pom for ee-9/feature-pack in this branch.

I'll check the undertow layer.

jfdenise (approved):

bstansberry:
@jfdenise Thank you for your review and question.

I told you this in chat but for others who may read this I misspoke re the relevant branch. https://github.com/bstansberry/wildfly/tree/ee-9 is the branch I was referring to. https://github.com/bstansberry/wildfly/blob/ee-9/ee-9/feature-pack/pom.xml#L120 shows how it is copying src from various spots to create a FP with no deps other than the wildfly-core FP. (We could use src zips to do the same, but I started this branch before the src zip thing was merged in galleon, plus doing it this way is efficient if all the source is in the same project.)

Re: undertow, that one wasn't an issue because for a different reason (not willing to support vault in the ee 9 fp) I redefined that layer: https://github.com/bstansberry/wildfly/blob/ee-9/ee-9/feature-pack/src/main/resources/layers/standalone/undertow/layer-spec.xml

In general undertow + Galleon (not necessarily layers) is a bit different in that FP because of wanting to clearly eliminate legacy security.

PR #13694, opened 2020-11-09
[WFLY-14034] FIx Health checks when RBAC is enabled

jmesnil:
Create a LocalModelControllerClient by calling createSuperUserClient so
that the server readiness checks can be performed even when RBAC is
enabled.
This local client is used only by server checks hard-coded in the
ServerReadinessProbes probes.

JIRA: https://issues.redhat.com/browse/WFLY-14034

### DIFF
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
@@ -85,7 +85,10 @@ public class HealthReporterService implements Service<HealthReporter> {
         final boolean defaultServerProceduresDisabled = ConfigProvider.getConfig().getOptionalValue("mp.health.disable-default-procedures", Boolean.class).orElse(false);
         healthReporter = new HealthReporter(emptyLivenessChecksStatus, emptyReadinessChecksStatus, defaultServerProceduresDisabled);
 
-        modelControllerClient = modelControllerClientFactory.get().createClient(managementExecutor.get());
+        // we use a SuperUserClient for the local model controller client so that the server checks can be performed when RBAC is enabled.
+        // a doPriviledged block is not needed as these calls are initiated from the management endpoint.
+        // The user accessing the management endpoints must be authenticated (if security-enabled is true) but the server checks are not executed on their behalf.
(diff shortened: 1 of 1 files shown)


microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/HealthReporterService.java
```diff
@@ -85,7 +85,7 @@ public void start(StartContext context) {
         final boolean defaultServerProceduresDisabled = ConfigProvider.getConfig().getOptionalValue("mp.health.disable-default-procedures", Boolean.class).orElse(false);
         healthReporter = new HealthReporter(emptyLivenessChecksStatus, emptyReadinessChecksStatus, defaultServerProceduresDisabled);
 
-        modelControllerClient = modelControllerClientFactory.get().createClient(managementExecutor.get());
+        modelControllerClient = modelControllerClientFactory.get().createSuperUserClient(managementExecutor.get(), true);
```
darranl, line 88: I haven't got time today to check into the code but I suspect this will be triggering a permissions check here.  

The normal reaction to that would be "I must add a doProvileged" but I think this code is correct as the complete call stack will be initiated from the management endpoint.

But assuming my two assumptions are true it may be worth adding a comment so no one helpfully adds a "missing doPrivileged" at a later point.

jmesnil: @darranl I've added a comment based on your remark.

I've manually tested that if RBAC is enabled, this code works (when security-enabled is false).
When security-enabled is true, the authentication fails when the HTTP endpoint is accessed and this code is not invoked.

bstansberry:
This seems ok. The health endpoint doesn't expose information that someone in, for example, the Operator role couldn't see.

To be really cautious perhaps the op ModelNode constants in ServerReadinessProbes could include the 'roles' operation header, set to Operator. That shouldn't have any practical impact on those ops though; doing it would more be a hint to later devs who might add something new.

bstansberry:
Once Darran oks this it seems like a good candidate for next week's 21.0.1.

darranl (approved):

darranl:
I have added an approval, I think most importantly would be ensuring the end user doesn't have an option that modifies the operation being called.

PR #13697, opened 2020-11-10
WFLY-14055 Implement caching inside ExternalBeanArchiveProcessor to i…

manovotn:
…mprove deployment times for big applications.

JIRA link - https://issues.redhat.com/browse/WFLY-14055

This attempts to solve an issue with deployment times for massive applications using Weld.
The numbers plus reproducer can be seen in the JIRA issue.

Locally I have only verified this against basic integration suite, so if more issues popup, I'll look into it.

@mkouba I'd like to hear your take on this. I am aware that map of maps isn't exactly nice, but it seems to get the job done here.

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/BeanDeploymentArchiveImpl.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/BeanDeploymentArchiveImpl.java
@@ -239,11 +239,12 @@ public class BeanDeploymentArchiveImpl implements WildFlyBeanDeploymentArchive {
             return true;
         }
 
+        final String thatIdentifier = that.getModule().getName();
         // basic check whether the module is our dependency
         for (DependencySpec dependency : module.getDependencies()) {
             if (dependency instanceof ModuleDependencySpec) {
                 ModuleDependencySpec moduleDependency = (ModuleDependencySpec) dependency;
-                if (moduleDependency.getIdentifier().equals(that.getModule().getIdentifier())) {
+                if (moduleDependency.getName().equals(thatIdentifier)) {
                     return true;
                 }
             }
@@ -254,7 +255,7 @@ public class BeanDeploymentArchiveImpl implements WildFlyBeanDeploymentArchive {
                 ModuleDependencySpec moduleDependency = (ModuleDependencySpec) dependency;
                 // moduleDependency might be an alias - try to load it to get lined module
                 Module module = loadModule(moduleDependency);
-                if (module != null && module.getIdentifier().equals(that.getModule().getIdentifier())) {
+                if (module != null && module.getName().equals(thatIdentifier)) {
                     return true;
                 }
             }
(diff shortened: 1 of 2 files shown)


manovotn:
@mkouba pointed out that `Module` is not a good map key because it doesn't implement `equals`/`hashCode`. Instead I should be using something else. According to jboss modules docs, Module names are a way to go. I will fix this momentarily.

manovotn:
Should be good now, let's see what does the CI say to it.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
```diff
@@ -325,9 +328,13 @@ private Index tryLoadIndex(URL indexUrl) {
      * @param dependencyModule
      * @return the map of exported resources
      */
-    private Map<URL, URL> findExportedResources(Module dependencyModule) {
+    private Map<URL, URL> findExportedResources(Module dependencyModule, Map<String, Map<URL, URL>> exportedResourcesCache) {
+        if (exportedResourcesCache.containsKey(dependencyModule)) {
```
bstansberry, line 332: This will always return false as the keys are dependencyModule.getName().

bstansberry (dismissed):

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/ExternalBeanArchiveProcessor.java
manovotn: Thanks for spotting this @bstansberry! Should be fixed now.

manovotn:
I have also added the optimization @wolfc suggested on the issue (in a slightly tweaked manner) as a separate commit. Locally, I am seeing another equally massive improvement with that.

I haven't had the chance to run WFLY testsuite locally though, so let's see what CI says.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/BeanDeploymentArchiveImpl.java
```diff
@@ -239,11 +239,12 @@ public boolean isAccessible(BeanDeploymentArchive target) {
             return true;
         }
 
+        final String thatIdentifier = that.getModule().getName();
         // basic check whether the module is our dependency
         for (DependencySpec dependency : module.getDependencies()) {
             if (dependency instanceof ModuleDependencySpec) {
                 ModuleDependencySpec moduleDependency = (ModuleDependencySpec) dependency;
-                if (moduleDependency.getIdentifier().equals(that.getModule().getIdentifier())) {
+                if (moduleDependency.getName().equals(thatIdentifier)) {
```
manovotn, line 247: This is the only part where I am not certain - the comparison takes the name of `ModuleDependencySpec` and that of a `Module`. But I assume that if their identifiers were comparable and those were replaced by String names, then the names should be comparable as well. Feel free to prove me wrong.

bstansberry: @manovotn LGTM.  I'll just assume the existing equals is valid. If not, that's not the concern of this PR. ;) 

Re the change in both classes getIdentifier() is deprecated with javadoc saying to use getName(). And internally both implement getIdentfier() as ModuleIdentifier.fromString(name). So this change is valid according to both the implementation and the javadoc.

bstansberry (approved):

PR #13698, opened 2020-11-10
Protect Managed*ExecutorServiceImpl life cycle methods

boris-unckel:
Change WildFly implementation of ManagedExecutorServiceImpl and ManagedScheduledExecutorService according to org.glassfish.enterprise.concurrent.AbstractManagedExecutorServiceAdapter
Fixes [WFLY-14057](https://issues.redhat.com/browse/WFLY-14057)

### DIFF
--- a/ee/src/main/resources/org/jboss/as/ee/subsystem/LocalDescriptions.properties
+++ b/ee/src/main/resources/org/jboss/as/ee/subsystem/LocalDescriptions.properties
@@ -55,7 +55,6 @@ managed-executor-service.hung-thread-count=The number of executor threads that a
 managed-executor-service.max-thread-count=The largest number of executor threads.
 managed-executor-service.task-count=The approximate total number of tasks that have ever been submitted for execution.
 managed-executor-service.thread-count=The current number of executor threads.
-managed-executor-service.refuse-lifecycle-operation=Lifecycle operation not supported.
 
 managed-scheduled-executor-service=A managed scheduled executor service
 managed-scheduled-executor-service.add=Adds the scheduled executor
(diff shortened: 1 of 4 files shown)


bstansberry:
@emmartins Please review.

ee/src/main/resources/org/jboss/as/ee/subsystem/LocalDescriptions.properties
```diff
@@ -55,6 +55,7 @@ managed-executor-service.hung-thread-count=The number of executor threads that a
 managed-executor-service.max-thread-count=The largest number of executor threads.
 managed-executor-service.task-count=The approximate total number of tasks that have ever been submitted for execution.
 managed-executor-service.thread-count=The current number of executor threads.
+managed-executor-service.refuse-lifecycle-operation=Lifecycle operation not supported.
```
bstansberry, line 58: I don't think this should be here.  This file provides the English language version of free-form text strings used in the WildFly management API description (i.e. what you see if you do /:read-resource-description or read-operation-description in the CLI). It's not used for anything else.

boris-unckel: Thank you. I have reverted the change to LocalDescriptions.properties.

emmartins:
@bstansberry current implementation is correct, those lifecycle methods are internal use and should not be changed. App code gets an adapter with lifecycle methods protected.

boris-unckel:
Issue is not existing - the adapter is already in use. Thank you @emmartins for clarification.

PR #13708, opened 2020-11-13
Fix for WFLY-13967, Provide an simple overview of what a layer is at the start of the layers table

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-13967
@bstansberry , when this is good, I will open PR against 21.x.
JF

### DIFF
--- a/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
+++ b/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
@@ -1,6 +1,15 @@
 [[wildfly_layers]]
 === WildFly Layers
 
+A Galleon layer is a name that identifies a server capability (e.g.: jaxrs, 
+ejb, microprofile-config, jpa) or an aggregation of such capabilities. A layer captures a server capability in the form of:
+
+* A piece of server XML configuration (e.g.: extension, configured subsystem, interfaces) that describes the capability.
+* A set of modules and other filesystem content that implements the capability.
+
+When you are using a layer, it delivers these pieces of information in order for 
+Galleon to assemble a server containing only the required configuration and modules. 
+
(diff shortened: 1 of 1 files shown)


xstefank (approved):
+1

docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
```diff
@@ -1,6 +1,15 @@
 [[wildfly_layers]]
 === WildFly Layers
 
+A Galleon layer is a name that identifies a server capability (e.g.: jaxrs, 
+ejb, microprofile-config, jpa) or an aggregation of such capabilities. A layer captures a server capability in the form of:
+
+* A piece of server XML configuration (e.g.: extension, configured subsystem, interfaces) that describes the capability.
+* A set of JBoss modules that implements the capability.
```
bstansberry, line 8: s/JBoss modules/JBoss Modules modules/g

or just "modules".

More words, but perhaps

* A set of modules and other filesystem content that implements the capability.

Maybe that's too much information. Layers bring things other than modules is the point; e.g. core-tools, vault-tools etc.

jfdenise: @bstansberry , I focused on the modules/configs. I added what you suggested. I think it is vague enough to cover the "other files".

bstansberry (approved):

bstansberry:
@jfdenise Thanks. Re a 21.x PR for that to be useful someone would have to rebuild the docs and update the wildfly.github.io site. We can do that if you'd like.

jfdenise:
I was mentioning 21 because of the Fix Release of this fix. I am not sure it is important enough to release the web site.

PR #13710, opened 2020-11-16
[WFLY-14077] Some tests hang indefinitely when using security manager…

spyrkob:
JIRA: https://issues.redhat.com/browse/WFLY-14077

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/wildfly/test/manual/elytron/seccontext/AbstractSecurityContextPropagationTestBase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/wildfly/test/manual/elytron/seccontext/AbstractSecurityContextPropagationTestBase.java
@@ -576,7 +576,8 @@ public abstract class AbstractSecurityContextPropagationTestBase {
                 .addClasses(EntryServlet.class, WhoAmIServlet.class, WhoAmI.class, ReAuthnType.class, SeccontextUtil.class)
                 .addAsManifestResource(createPermissionsXmlAsset(new ElytronPermission("authenticate"),
                         new ElytronPermission("getPrivateCredentials"), new ElytronPermission("getSecurityDomain"),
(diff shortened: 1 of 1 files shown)


darranl (changes requested):
Please add a complete stack trace somewhere of the failure that leads to this permission being required.

This does not look like a permission we would expect end users to be needing to grant to their deployment.

spyrkob:
@darranl I attached the stacktrace to WFLY-14077

bstansberry:
@spyrkob This does not look like a missing deployment permission. It looks like a JBoss Marshalling bug where it cannot initialize internal classes. The Protocol class has a static field of type Unsafe which results in failure to initialize if deployment classes are on the stack.

@ropalka FYI.

ropalka:
Did you see testsuite/integration/iiop/src/test/java/org/jboss/as/test/iiop/security/IIOPSecurityInvocationTestCase.java @spyrkob how to add missing / additional permissions?

spyrkob:
@ropalka yes I followed that for this PR, but it sounds like this is not the way to go if I understand @bstansberry's comment right. 
Should I close this PR and open a bug in Marshalling?

ropalka:
Aha, I see. I didn't have a look to your PR proposal.

ropalka:
How could I reproduce the problem @spyrkob ? I'm running: 
---

cd wildfly/testsuite/integration/manualmode
mvn clean  test -Dtest=**/seccontext/*TestCase -DallTests -Dsecurity.manager=true

---

but no luck

spyrkob:
@ropalka are you using Java 11? 

I'm getting failures for example on `mvn clean test -Dtest=AuthenticationForwardingSFSFTestCase#testServletBasicToEjbForwardedIdentity -DallTests -Dsecurity.manager`

ropalka:
@spyrkob you can close this PR - I have a proper marshalling fix ready

spyrkob:
Thanks @ropalka

PR #13716, opened 2020-11-17
 [WFLY-9213]: Implement the Pause method for a Topic via Management APIs

ehsavoie:
* Adding a `pause` method on JMS Topics
* Adding a `resume` method on JMS Topics
* Adding a `paused` runtime attribute on JMS Topics

Jira: https://issues.jboss.org/browse/WFLY-9213
Proposal: https://github.com/wildfly/wildfly-proposals/pull/269


messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/JMSTopicControlHandler.java
```diff
@@ -388,6 +407,30 @@ private String listSubscribersInfosAsJSON(final DurabilityType durability, Addre
       }
    }
 
+   private void pause(AddressControl control, boolean persist) {
+       try {
+       control.pause(persist);
+       } catch (Exception e) {
+         throw new RuntimeException(e);
```
bstansberry, line 414: Perhaps add a separate catch for RuntimeException and just rethrow it instead of wrapping?

Same comment applies to resume and isPaused.

bstansberry:
My minor comment re RuntimeException aside, this LGTM.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/jms/JMSTopicControlHandler.java
kabir: There is actually a way to rethrow both without wrapping in a RuntimeException - see https://github.com/resteasy/Resteasy/blob/43f46228bbf8278a96648b48309f71d5ccad1e4c/resteasy-core/src/main/java/org/jboss/resteasy/core/SynchronousDispatcher.java#L186

ehsavoie: ok, using Kabir's trick. This should be applied to the rest of the code though in a separate commit

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
```diff
@@ -234,6 +234,22 @@ JMS endpoints can easily be created through the CLI:
 }
 ----
 
+
+Pausing and resuming Queues and Topics
+
+When a queue is paused, it will receive messages but will not deliver them. When it's resume, it'll begin delivering the queued messages, if any.
+When a topic is paused, it will receive messages but will not deliver them. Newly added subscribers will be paused too until the topic is resumed. When it is resumed, delivering will occur again. The `persist` parameter ensure that the opic stays paused on the rstart of the server.
```
honza-kasik, line 241: Just a typos: `opic` -> `topic`, `rstart` -> `restart`.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Messaging.adoc
```diff
@@ -234,6 +234,22 @@ JMS endpoints can easily be created through the CLI:
 }
 ----
 
+
+Pausing and resuming Queues and Topics
+
+When a queue is paused, it will receive messages but will not deliver them. When it's resume, it'll begin delivering the queued messages, if any.
```
honza-kasik, line 240: Possible typo: `resume` -> `resumed`?

honza-kasik (changes requested):
Thanks @ehsavoie ! Implemented tests look good to me, although they could be documented a bit - I can imagine that when something happens it may be difficult for someone not too familiar with this feature to immediately understand. IMHO adding at least a JIRA issue ID to JavaDoc would be great. WDYT?

Requesting changes just for the typos in documentation.

honza-kasik (approved):
Thank you, this LGTM from QE point of view (tests and docs) although it seems some automated checks haven't passed yet.

PR #13724, opened 2020-11-18
[WFLY-14089]: Make NetworkHealthTestCase less fragile by not checking the full trace.

ehsavoie:
* Checking error codes instead of full error message.

Jira: https://issues.redhat.com/browse/WFLY-14089


bstansberry (approved):

bstansberry:
Not urgent but I filed https://issues.redhat.com/browse/WFLY-14090 to consider reusing existing utilities for this log scanning stuff.

PR #13728, opened 2020-11-22
[WFLY-14111] Fix for 10 flaky tests due to AbstractUndertowSubsystemTestCase.java

Vidishab18:
The test `org.wildfly.extension.undertow.AbstractUndertowSubsystemTestCase#testRuntime` is flaky because it has `Assert.assertEquals("default-alias", new ArrayList<>(host.getAllAliases()).get(1))` where `getAllAliases()` returns `HashSet`, so the test checks that "default-alias" is at the second index of the list formed from the set. However, per the documentation "HashSet makes no guarantees as to the iteration order of the set", so converting the `HashSet` into a `List` doesn't ensure the deterministic order of elements, and `get(1)` may return a different value. The modified code checks that "default-alias" is in the `HashSet` itself.

The modified code helps with the following 10 flaky tests in subclasses of `AbstractUndertowSubsystemTestCase`:
* `org.wildfly.extension.undertow.UndertowSubsystem100TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem30TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem80TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem90TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem60TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem70TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem50TestCase#testRuntime` 
* `org.wildfly.extension.undertow.UndertowSubsystem31TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem40TestCase#testRuntime`
* `org.wildfly.extension.undertow.UndertowSubsystem110TestCase#testRuntime`

### DIFF
--- a/undertow/src/test/java/org/wildfly/extension/undertow/AbstractUndertowSubsystemTestCase.java
+++ b/undertow/src/test/java/org/wildfly/extension/undertow/AbstractUndertowSubsystemTestCase.java
@@ -22,7 +22,6 @@
 package org.wildfly.extension.undertow;
 
 import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
 import java.util.Properties;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


darranl:
If you could please create a Jira issue in the WFLY project and reference it in the commit message and PR that would be great.

bstansberry:
Beyond Darran's JIRA request, I've got a general question we ask for first commits:

Vidishab18:
The Jira issue raised : [](https://issues.redhat.com/browse/WFLY-14111 )
Yes agree to the terms of MIT License, http://www.opensource.org/licenses/mit-license.php.

bstansberry:
Thanks @Vidishab18. Can you amend the commit message to reference WFLY-14111?

Vidishab18:
I have changed the commit message.

bstansberry:
Thanks, @Vidishab18 !

PR #13731, opened 2020-11-23
[WFLY-14108] Base metrics subsystem and microprofile-metrics move

jmesnil:
* Add base metrics subsystem that exposes WildFly management model and
  JVM metrics (using JMX) to the /metrics endpoint on the management
  port (with the Prometheus format)
  * JMX Metrics are configured using the jmx-metrics.properties
  * Names of the base metrics were computed to match the existing ones
  from the microprofile-metrics subsystem. A Prometheus server can
  consume metrics from a base WildFly server or a one with the
  MicroProfile feature pack without any differences for the WildFly
  management model and JVM metrics.
* Move the existing microprofile-metrics-smallrye subsystem to the
  org.wildfly:wildfly-mp-feature-pack-galleon-common feature pack with
  other MicroProfile subsystems
  * add version 2.1.0 to microprofile-metrics-smallrye extension
  as the security-enabled attribute is deprecated and superseded by
  the attribute in the base "metrics" subsystme which is the one that
  actual registers the HTTP management endpoint
* Add tests for the new metrics subsystem in the
  testsuite/integration/basic
* Move tests related to the  microprofile-metrics-smallrye in
  testsuite/integration/microprofile

JIRA: https://issues.redhat.com/browse/WFLY-14108


jmesnil:
The failures are related to the PR. I'm investigating them.

ee-feature-pack/galleon-feature-pack/wildfly-feature-pack-build.xml
```diff
@@ -125,7 +125,6 @@
                 <extension>org.wildfly.extension.messaging-activemq</extension>
                 <extension>org.wildfly.extension.microprofile.config-smallrye</extension>
                 <extension>org.wildfly.extension.microprofile.health-smallrye</extension>
-                <extension>org.wildfly.extension.microprofile.metrics-smallrye</extension>
```
bstansberry, line 128: Is the new extension not being added here because org.wildfly.extension.microprofile.metrics-smallrye shouldn't have been in the domain section in the first place?

metrics/src/main/java/org/wildfly/extension/metrics/MetricsContextService.java
```diff
@@ -20,77 +20,86 @@
...
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StopContext;
+import org.wildfly.extension.metrics.internal.PrometheusExporter;
+import org.wildfly.extension.metrics.internal.WildFlyMetricRegistry;
 
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2018 Red Hat inc.
  */
-public class MetricsContextService implements Service {
+public class MetricsContextService implements Service<MetricsContextService> {
```
bstansberry, line 45: This needs the deprecated Service interface instead of ServiceBuilder.provides?

metrics/src/main/java/org/wildfly/extension/metrics/MetricsCollectorService.java
```diff
@@ -0,0 +1,99 @@
...
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
+import org.wildfly.extension.metrics.api.AbstractMetricCollector;
+import org.wildfly.extension.metrics.internal.BaseMetricCollector;
+import org.wildfly.extension.metrics.internal.WildFlyMetricRegistry;
+
+/**
+ * Service to create a metric collector
+ */
+public class MetricsCollectorService implements Service<AbstractMetricCollector> {
```
bstansberry, line 48: General comment: non-public and final are good wherever possible.

I didn't look carefully; this class may be fine and maybe all are. This is a general reminder comment. :)

metrics/src/main/java/org/wildfly/extension/metrics/api/MetricID.java
```diff
@@ -0,0 +1,75 @@
+package org.wildfly.extension.metrics.api;
```
bstansberry, line 1: A few of these have no copyright header.

metrics/src/main/java/org/wildfly/extension/metrics/api/WildFlyMetric.java
```diff
@@ -0,0 +1,84 @@
...
+    private ModelNode readAttributeValue(PathAddress address, String attributeName) {
+        final ModelNode readAttributeOp = new ModelNode();
+        readAttributeOp.get(OP).set(READ_ATTRIBUTE_OPERATION);
+        readAttributeOp.get(OP_ADDR).set(address.toModelNode());
+        readAttributeOp.get(ModelDescriptionConstants.INCLUDE_UNDEFINED_METRIC_VALUES).set(false);
+        readAttributeOp.get(NAME).set(attributeName);
+        ModelNode response = modelControllerClient.execute(readAttributeOp);
+        String error = getFailureDescription(response);
+        if (error != null) {
+            throw LOGGER.unableToReadAttribute(attributeName, address, error);
```
bstansberry, line 73: The way getValue() is implemented it should be possible to check for WFLYCTL0216 and return an undefined ModelNode. I believe that would eliminate the https://issues.redhat.com/browse/WFLY-11933 issue for this metrics impl. With MP Metrics it's harder.

I expect you know all this which is why you use OptionalDouble. ;)

jmesnil:
@bstansberry I've closed this PR. I underestimated the work to move the subsystem from one feature pack to another. I squashed the commits before opening this PR and it makes it very hard to figure out the regression that I introduced.
I'll open a new PR today with separate commits for the different stages of this RFE (move microprofile-metrics to the microprofile feature pack, refactor code, introduce a base metrics subsystem in the ee-feature-pack).
I think it will make the review simpler (and I'll squash after the PR is reviewed and before it is merged)

PR #13734, opened 2020-11-24
[WFLY-14115] Tests failures in multinode with security manager enabled affect other tests

gaol:
Issue: https://issues.redhat.com/browse/WFLY-14115

Summary of the PR:
* Adding FilerPermission on client node to be able to delete xa recovery file in case security manager is enabled.


tadamski:
@gaol Sorry, I somehow missed this PR which is equal to mine.

PR #13422, opened 2020-07-23
[WFLY-13695] Upgrade wildfly-http-client to 1.0.23.Final

fl4via:
Jira: https://issues.redhat.com/browse/WFLY-13695
20.x PR (in case one is needed): https://github.com/wildfly/wildfly/pull/13423

        Release Notes - WildFly EJB HTTP Client - Version 1.0.23.Final
                                                                                                                                            
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-47'>WEJBHTTP-47</a>] -         Ensure the HttpRemoteTransactionPeer captures current AuthenticationContext and SSLContext
</li>
</ul>
                                                                

        Release Notes - WildFly EJB HTTP Client - Version 1.0.22.Final
                                                                                                                                            
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-45'>WEJBHTTP-45</a>] -         UT000103 thrown when WildflyClientOutputStream size is exactly 1024 bytes
</li>
</ul>

PR #13736, opened 2020-11-24
[WFLY-14115] Multinode tests failures: grant delete permission for EJ…

tadamski:
…B XA recovery log files

https://issues.redhat.com/browse/WFLY-14115

This PR is an alternative to: https://github.com/wildfly/wildfly-http-client/pull/45
I believe we should just provide the tests with required permissions. I have checked that it prevents the hang.

### DIFF
--- a/testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/ejb/http/EjbOverHttpDescriptorTestCase.java
+++ b/testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/ejb/http/EjbOverHttpDescriptorTestCase.java
@@ -83,12 +83,12 @@ public class EjbOverHttpDescriptorTestCase {
         jar.addClasses(EjbOverHttpDescriptorTestCase.class);
         jar.addAsManifestResource("META-INF/jboss-ejb-client-http-connections.xml", "jboss-ejb-client.xml")
                 .addAsManifestResource("ejb-http-wildfly-config.xml", "wildfly-config.xml")
-                .addAsManifestResource(createPermissionsXmlAsset(createFilePermission("read,write",
-                        "jbossas.multinode.client", Arrays.asList("standalone", "data", "ejb-xa-recovery")),
-                        createFilePermission("read,write",
-                                "jbossas.multinode.client", Arrays.asList("standalone", "data", "ejb-xa-recovery", "-")),
-                        new SocketPermission(TestSuiteEnvironment.formatPossibleIpv6Address(System.getProperty("node0")) + ":" + serverPort,
-                                "connect,resolve")),
+                .addAsManifestResource(
+                        createPermissionsXmlAsset(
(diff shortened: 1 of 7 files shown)


bstansberry:
@tadamski Is your thinking here that if this client were an external application running with a security manager the user would need to give the app the permissions needed to read/write/delete these files used for tx recovery? So the same would be true for a client running inside the server?

I'm just asking to be sure I understand.

In any case, if it's correct for the app to have write perms it makes sense to have delete as well.

bstansberry:
@tadamski @gaol #13734 seems to be about the same thing as this one.

gaol:
@bstansberry please ignore #13734, just closed it. :)

tadamski:
@bstansberry Yes, actually in those case scenarios both client and server are running inside the server. OTOH, I believe such properties won't be required in standalone client - as it cannot be transaction coordinator it wouldn't require to deal with those logs.

bstansberry:
@tadamski I apologize for not coming back to this sooner. The point that a standalone client wouldn't need perms concerned me but I wasn't sure why and didn't have a good response at the time. 

Looking at the stack trace in https://github.com/wildfly/wildfly/pull/13422#issuecomment-680012845 I don't see why the ClientEjb used by TransactionInvocationTestCase should need special permissions. It's not doing anything exotic; it's just a BMT EJB that invokes UserTransaction.rollback().

When I look at FileSystemXAResourceRegistry$XAResourceRegistryFile I see that the one of the two constructors is doing a permission check that it can read,write the file's directory. I'm not sure why the other constructor doesn't. Anyway it seems that class contemplates ensuring it has perms for the dir and then opening a file using a privileged block, at least in some cases, i.e. the ones that result in fileChannel not being null. So it seems like in those same cases removeResource could use a privileged block to delete the file it created.

@mmusgrov @ochaloup wdyt?

ochaloup:
I'm not sure about the permission checks in the constructors but it's right that the ejb client needs the security manager permission for the file system access.
The `ClientEjb` makes a call from one server to another. The WFTC persists own data about transaction propagation that could be needed on transaction failure to recover correctly. Ie. it's the BMT EJB bean but the rollback is propagated to the second server and the results back then.
In case of failure the transaction recovery may come to the client (to WFTC) and asks if there were some unfinished operations which need to be rolled-back. The transaction manager for JTA stores data to Narayana object store only when the prepare 2PC phase finishes successfully. Otherwise recovery manager has no data and it only verifies with participants if there is something to get finished.

Adding the security manager permissions seem reasonable to me here.

tadamski:
@bstansberry hmm, TBH I thought that it is better to avoid privileged blocks when possible and I haven't noticed that the class already does it; I will prepare alternative PR then

bstansberry (approved):
https://issues.redhat.com/browse/WFTC-57 has been filed to look into what's concerned me about this, so I'm fine with this.

PR #13737, opened 2020-11-24
WFLY-14118 Distributed session manager can throw UnsupportedOperationException on concurrent calls to HttpServletRequest.getSession(...) for a given session

pferraro:
https://issues.redhat.com/browse/WFLY-14118


pferraro:
Not quite right.

PR #13738, opened 2020-11-25
[WFLY-14025] only add dependency to inject DefaultDataSource if defau…

bmaxwell:
…lt datasource binding is specified in ee subsystem

https://issues.redhat.com/browse/WFLY-14025

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/deployers/datasource/DefaultDataSourceResourceReferenceProcessor.java
+++ b/connector/src/main/java/org/jboss/as/connector/deployers/datasource/DefaultDataSourceResourceReferenceProcessor.java
@@ -22,6 +22,7 @@
 package org.jboss.as.connector.deployers.datasource;
 
 import org.jboss.as.ee.component.Attachments;
+import org.jboss.as.ee.component.EEModuleDescription;
 import org.jboss.as.ee.component.InjectionSource;
 import org.jboss.as.ee.component.LookupInjectionSource;
 import org.jboss.as.ee.component.deployers.EEResourceReferenceProcessor;
@@ -48,7 +49,9 @@ public class DefaultDataSourceResourceReferenceProcessor implements DeploymentUn
         if(deploymentUnit.getParent() == null) {
             final EEResourceReferenceProcessorRegistry eeResourceReferenceProcessorRegistry = deploymentUnit.getAttachment(Attachments.RESOURCE_REFERENCE_PROCESSOR_REGISTRY);
             if(eeResourceReferenceProcessorRegistry != null) {
-                eeResourceReferenceProcessorRegistry.registerResourceReferenceProcessor(RESOURCE_REFERENCE_PROCESSOR);
+                final EEModuleDescription eeModuleDescription = deploymentUnit.getAttachment(Attachments.EE_MODULE_DESCRIPTION);
+                if(eeModuleDescription != null && eeModuleDescription.getDefaultResourceJndiNames().getDataSource() != null)
+                    eeResourceReferenceProcessorRegistry.registerResourceReferenceProcessor(RESOURCE_REFERENCE_PROCESSOR);
             }
         }
     }


connector/src/main/java/org/jboss/as/connector/deployers/datasource/DefaultDataSourceResourceReferenceProcessor.java
```diff
@@ -48,7 +49,9 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
         if(deploymentUnit.getParent() == null) {
             final EEResourceReferenceProcessorRegistry eeResourceReferenceProcessorRegistry = deploymentUnit.getAttachment(Attachments.RESOURCE_REFERENCE_PROCESSOR_REGISTRY);
             if(eeResourceReferenceProcessorRegistry != null) {
-                eeResourceReferenceProcessorRegistry.registerResourceReferenceProcessor(RESOURCE_REFERENCE_PROCESSOR);
+                final EEModuleDescription eeModuleDescription = deploymentUnit.getAttachment(Attachments.EE_MODULE_DESCRIPTION);
+                if(eeModuleDescription != null && eeModuleDescription.getDefaultResourceJndiNames().getDataSource() != null)
```
gaol, line 53: This is always `false` because `DefaultDataSourceResourceReferenceProcessor` (Phase.STRUCTURE) gets executed before `DefaultBindingsConfigurationProcessor` (Phase.PARSE), the latter set the default datasource if any.

This will cause that `@Resource() DataSource ds` won't get injected.

bstansberry: I suggest we look into getting rid of DefaultBindingsConfigurationProcessor and collapse its logic into EEModuleInitialProcessor. When EEModuleInitialProcessor creates and attaches the EEModuleDescription all the default resource jndi names are known, as they all come from the subsystem config. There's no new information that comes in between then and Phase.PARSE when DefaultBindingsConfigurationProcessor runs. 

The approach @bmaxwell proposes here would then work. But it should be applied to the other analogous DUPs as well, which I believe will have the same problem.

tadamski:
@bstansberry @bmaxwell 
The suggested fix with refactor of  EEModuleInitialProcessor seems a little bit tricky to me because the object has to be created during EeExtension creation:
https://github.com/wildfly/wildfly/blob/master/ee/src/main/java/org/jboss/as/ee/subsystem/EeExtension.java#L76
 but  EEModuleInitialProcessor can be created only on boottime when appclient property is known.

OTOH just changing the phase of DefaultBindingsConfigurationProcessor seemed to work for me. I see that in reproducer default data source is injected when defined and no injected (without errors) if not.
I have created a PR with this fix (+ also modifing other default injections to work the same way):
https://github.com/wildfly/wildfly/pull/13744/files

gaol:
@tadamski or maybe we can use a singleton object holding the `EEDefaultResourceJndiNames` on the `DefaultBindingsAdd`, then creating the `EEModuleDescription` in `EEModuleInitialProcessor` ?

gaol:
Also with both proposed changes, in case of default datasource binding removed from ee subsystem, it will make the following deployment successful which I think it failed before the change:

```
public class TestServlet extends HttpServlet {
   @Resource
  private DataSource ds;
}
```

is it OK for the change?

I am preparing some tests in https://github.com/wildfly/wildfly/compare/master...gaol:WFLY-14025-brad?expand=1, so I want to be clear what exactly we want. thanks :)

bstansberry:
@gaol Can you ask your question on #13744?  That is the active PR and I'd like to close this one but I don't want your very good question to get lost.

It would be better if your tests came in via that PR as well.

gaol:
@bstansberry sure, I have move the question and sent PR to the bugfix branch with the tests. :) I think we can close this one.

bstansberry:
Closing; superceded by #13744

PR #13743, opened 2020-11-26
[WFLY-14130] proxy-list attribute ignored in modcluster subsystem

spyrkob:
JIRA: https://issues.redhat.com/browse/WFLY-14130

### DIFF
--- a/mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationServiceConfigurator.java
+++ b/mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ProxyConfigurationServiceConfigurator.java
@@ -370,7 +370,9 @@ import org.wildfly.clustering.service.SupplierDependency;
 
             });
         }
-        builder.mcmp().setProxyConfigurations(proxies);
+        if (!proxies.isEmpty()) {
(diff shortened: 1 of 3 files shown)


spyrkob:
@rhusar could you review when you have a moment please?

rhusar:
Will do.

spyrkob:
rebased the PR and fixed some formatting

@rhusar did you have a chance to look into this?

bstansberry (approved):

bstansberry:
@rhusar if CI looks good I'd like to merge this. If 'proxies' hasn't defined anything this prevents it overwriting what was configured from proxy-list with an empty list. If proxies did define something then it overwrites, which is the better choice for a preference. And at this point (Stage.RUNTIME) there really is no question of preference anyway as proxy-list defines proxies as an alternative, so a config that set both would be rejected in Stage.MODEL.

rhusar (approved):
Sorry, it looks like I forgot finish reviewing this. Yes, this is correct (org.jboss.modcluster.config.builder.MCMPHandlerConfigurationBuilder#setProxyConfigurations was being called twice and PROXIES was always overriding whatever the legacy PROXY_LIST has already set - adding the condition is sufficient fix since this is an alternative attribute).

Also, good stuff adding the test, hopefully it's stable in our CI :-)

PR #13744, opened 2020-11-27
[WFLY-14025] Don't create default binding injections if they are not …

tadamski:
https://issues.redhat.com/browse/WFLY-14025

Depends on: 
https://github.com/wildfly/wildfly-core/pull/4414

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/subsystem/DefaultBindingsAdd.java
+++ b/ee/src/main/java/org/jboss/as/ee/subsystem/DefaultBindingsAdd.java
@@ -36,6 +36,8 @@ import org.jboss.dmr.ModelNode;
  */
 public class DefaultBindingsAdd extends AbstractBoottimeAddStepHandler {
 
+    public static final int STRUCTURE_EE_DEFAULT_BINDINGS_CONFIG = 0x1B01;
+
     private final DefaultBindingsConfigurationProcessor defaultBindingsConfigurationProcessor;
 
     public DefaultBindingsAdd(DefaultBindingsConfigurationProcessor defaultBindingsConfigurationProcessor) {
@@ -73,7 +75,7 @@ public class DefaultBindingsAdd extends AbstractBoottimeAddStepHandler {
 
         context.addStep(new AbstractDeploymentChainStep() {
             protected void execute(DeploymentProcessorTarget processorTarget) {
-                processorTarget.addDeploymentProcessor(EeExtension.SUBSYSTEM_NAME, Phase.PARSE, Phase.PARSE_EE_DEFAULT_BINDINGS_CONFIG, defaultBindingsConfigurationProcessor);
+                processorTarget.addDeploymentProcessor(EeExtension.SUBSYSTEM_NAME, Phase.STRUCTURE, STRUCTURE_EE_DEFAULT_BINDINGS_CONFIG, defaultBindingsConfigurationProcessor);
             }
         }, OperationContext.Stage.RUNTIME);
     }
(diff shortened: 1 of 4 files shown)


messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/deployment/DefaultJMSConnectionFactoryResourceReferenceProcessor.java
```diff
@@ -47,8 +48,11 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
         final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
         if(deploymentUnit.getParent() == null) {
             final EEResourceReferenceProcessorRegistry eeResourceReferenceProcessorRegistry = deploymentUnit.getAttachment(Attachments.RESOURCE_REFERENCE_PROCESSOR_REGISTRY);
-            if(eeResourceReferenceProcessorRegistry != null) {
-                eeResourceReferenceProcessorRegistry.registerResourceReferenceProcessor(RESOURCE_REFERENCE_PROCESSOR);
+            if (eeResourceReferenceProcessorRegistry != null) {
+                final EEModuleDescription eeModuleDescription = deploymentUnit.getAttachment(Attachments.EE_MODULE_DESCRIPTION);
+                if (eeModuleDescription != null && eeModuleDescription.getDefaultResourceJndiNames().getDataSource() != null) {
```
gaol, line 53: Should it be `getJmsConnectionFactory()` ?

bstansberry: Yes, it should be getJmsConnectionFactory().

bstansberry (changes requested):
@tadamski Thanks. This needs the small change Lin Gao noticed.

messaging-activemq/src/main/java/org/wildfly/extension/messaging/activemq/deployment/DefaultJMSConnectionFactoryResourceReferenceProcessor.java
tadamski: fixed

gaol:
@tadamski @bstansberry do you think we can set the phase value manually here just like what was suggested in another PR, so that we can do back port if it is approved? thanks :)

bstansberry:
@tadamski @gaol That is fine with me.

tadamski:
fixed

bstansberry (approved):

gaol:
With the proposed changes, in case of default datasource binding removed from ee subsystem, it will make the following deployment successful which I think it failed before the change:

```java
public class TestServlet extends HttpServlet {
   @Resource
  private DataSource ds;
}
```

Is it OK for the change?

I made some tests to the bugfix branch: https://github.com/tadamski/wildfly/pull/1 so that it can come in with this PR, please let me know if the tests cover the cases. thanks :)

ee/src/main/java/org/jboss/as/ee/subsystem/DefaultBindingsAdd.java
```diff
@@ -36,6 +36,8 @@
  */
 public class DefaultBindingsAdd extends AbstractBoottimeAddStepHandler {
 
+    public static final int STRUCTURE_EE_DEFAULT_BINDINGS_CONFIG = 0x1B01;
```
darranl, line 39: Instead of a temporary change that needs cleaning up we should just aim to get the WildFly Core change through and use it after the next tag.

darranl (changes requested):

darranl:
The following PR to WildFly Core will be bringing in the required Phase ID https://github.com/wildfly/wildfly-core/pull/4414

darranl:
@bstansberry / @gaol FYI the Phase ID should be 0x1B10 and not 0x1B01 as merged.

darranl:
I have submitted the following PR to correct this - https://github.com/wildfly/wildfly/pull/13791

PR #13746, opened 2020-11-27
[WFLY-14135] Do not overwrite the default-stateful-bean-access-timeou…

yersan:
…t on ejb-dist-cache/ejb-local-cache Galleon layers

Jira issue: https://issues.redhat.com/browse/WFLY-14135

### DIFF
--- a/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/ejb-dist-cache/layer-spec.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/ejb-dist-cache/layer-spec.xml
@@ -8,7 +8,7 @@
     </dependencies>
 
     <feature spec="subsystem.ejb3">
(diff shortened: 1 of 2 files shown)


kabir (approved):

bstansberry:
@yersan I think the LayersTestCase failure is why the attribute is part of the layer.

bstansberry (changes requested):
The layers must pass LayersTestCase.

yersan:
I removed them because I expected the values were merged by Galleon when the affected layers were used in conjunction with ejb / ejb-lite. They are merged correctly, however we are validating each layer in isolation on the LayersTestCase and the attribute removed is mandatory, hence, the failure. I recovered them by using the expected value as timeout.

chengfang (approved):

bstansberry (approved):

PR #13751, opened 2020-11-30
WFLY-13387 Implemented Jaeger client metrics

tobiasstadler:
Implementation for https://issues.redhat.com/browse/WFLY-13387


microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerClientMetricsHandler.java
```diff
@@ -0,0 +1,218 @@
...
+                break;
+            default:
+                throw new IllegalStateException(TracingExtensionLogger.ROOT_LOGGER.unknownMetric(metric));
+        }
+
+        context.getResult().set(result);
+    }
+
+    private static ServiceName resolveDeploymentUnitServiceName(OperationContext context, ModelNode operation) {
+        PathAddress address = PathAddress.pathAddress(operation.require(ModelDescriptionConstants.OP_ADDR));
```
bstansberry, line 198: This should be

PathAddress address = context.getCurrentAddress();

The result is the same but ^^^ is more efficient and is the preferred idiom because it's simpler.

bstansberry:
@ehsavoie Please have a look.

We need to be sure that any metrics we add will continue to be available as long as we provide this subsystem.

Also this should probably require a 'statistics-enabled' config attribute to turn on.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/JaegerClientMetricsHandler.java
tobiasstadler: done

tobiasstadler:
> @ehsavoie Please have a look.
> 
> We need to be sure that any metrics we add will continue to be available as long as we provide this subsystem.
> 
> Also this should probably require a 'statistics-enabled' config attribute to turn on.

As far as I can see the last time the metrics changed was Feb. 2018. But I am not involved in the jaeger project and cannot say if/when the will change next.

Would statistics-enabled be an attribute in org.wildfly.extension.microprofile.opentracing.SubsystemDefinition or org.wildfly.extension.microprofile.opentracing.TracingDeploymentDefinition?

bstansberry:
@tobiasstadler I'll defer to @ehsavoie on the attribute placement as he knows the subsystem better. Generally those are at the top level because finer-grained configuration isn't particularly useful.

ehsavoie:
@tobiasstadlerI think it is better to define this at the subsystem level, having tracers with metrics and tracers without could lead to misunderstanding and JaegerEnvTracerConfiguration wouldn't be easily configurable that way.

darranl:
@tobiasstadler are you able to follow up on the comments from @ehsavoie ?  Also it looks like there is presently a merge conflict.

tobiasstadler:
@darranl Sorry, I din't have time so far.

darranl:
@jasondlee Are you able to comment on how relevent this is considering other changes in this area?

jasondlee:
@darranl I'm not sure I see any overlap, necessarily, between this and the other changes, other than the possibility that the version change my break this, but I would expect, given the age of this PR, that it is already using the version we're investigating. That aside, I'm not sure i understand what's going on here to have much of substance to add.

jasondlee:
MicroProfile OpenTracing has been removed from WildFly, so this request is no longer valid.

PR #13752, opened 2020-11-30
[WFLY-14114] JDBC: deregister all drivers on undeployment

tadamski:
https://issues.redhat.com/browse/WFLY-14114
Depends on: https://github.com/wildfly/wildfly-core/pull/4399

This hack allows to deregister all drivers registered by a given deployment.
Because those objects are created by a driver code and not guaranteed to be deregistered this lead to memory leaks.
OTOH DriverManager checks the caller class's classloader so I wasn't able to neither list nor deregister those drivers from connector code. To work this around I have added additional resource loader to a driver module and injected the adapter class code there. As a result, I was able to obtain and deregister all drivers registered by the driver module.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/deployers/ds/processors/DriverManagerAdapterProcessor.java
+++ b/connector/src/main/java/org/jboss/as/connector/deployers/ds/processors/DriverManagerAdapterProcessor.java
@@ -1,0 +1,112 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.connector.deployers.ds.processors;
+
+import org.jboss.as.connector._drivermanager.DriverManagerAdapter;
+import org.jboss.as.server.deployment.Attachments;
+import org.jboss.as.server.deployment.DeploymentPhaseContext;
+import org.jboss.as.server.deployment.DeploymentUnit;
+import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
+import org.jboss.as.server.deployment.DeploymentUnitProcessor;
+import org.jboss.as.server.deployment.module.ModuleSpecification;
+import org.jboss.modules.AbstractResourceLoader;
+import org.jboss.modules.ClassSpec;
+import org.jboss.modules.ResourceLoaderSpec;
+
+import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ * <p>
+ * https://issues.redhat.com/browse/WFLY-14114
+ * <p>
+ * This is a hack that allows us to get access to {@link java.sql.Driver} registered by the driver code. Those objects are created by the
+ * driver code and registered in {@link java.sql.DriverManager}. Driver objects are not guaranteed to be deregistered which leads to leaks.
+ * {@link java.sql.DriverManager} allows for obtaining the list of drivers, and deregistering a driver but only in a give classloading context.
+ * As a result, connector module can not neither list or deregister drivers from a deployed driver module.
+ * <p>
+ * To work this around, this hack modifies driver's module by injecting {@link DriverManagerAdapter} class to it. Because @{@link DriverManagerAdapter}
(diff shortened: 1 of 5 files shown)


bstansberry:
@ropalka Can you have a look at the AbstractResourceLoader impl please?

@tadamski This LGTM. Can you manually set the phase value here and follow up with a separate PR to move to the constant from WF Core when it is available? I'd prefer to get this merged without blocking on a core release.

tadamski:
@bstansberry I have fixed the Phase issue.

connector/src/main/java/org/jboss/as/connector/deployers/ra/RaDeploymentActivator.java
```diff
@@ -104,6 +108,7 @@ public void activateProcessors(final DeploymentProcessorTarget updateContext) {
         updateContext.addDeploymentProcessor(ResourceAdaptersExtension.SUBSYSTEM_NAME, Phase.PARSE, Phase.PARSE_RESOURCE_DEF_ANNOTATION_ADMINISTERED_OBJECT,
                 new AdministeredObjectDefinitionAnnotationProcessor());
         updateContext.addDeploymentProcessor(ResourceAdaptersExtension.SUBSYSTEM_NAME, Phase.DEPENDENCIES, Phase.DEPENDENCIES_RAR_CONFIG, new RarDependencyProcessor(appclient));
+        updateContext.addDeploymentProcessor(ResourceAdaptersExtension.SUBSYSTEM_NAME, Phase.CONFIGURE_MODULE, CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER, new DriverManagerAdapterProcessor());
```
ropalka, line 111: There is an agreed contract that all deployment phases and all processor ids are defined in Phase object.
Further more someone in the future might define the same DU processor ID in WildFly Core and it would collide. Please keep the contract.

connector/src/main/java/org/jboss/as/connector/deployers/ra/RaDeploymentActivator.java
```diff
@@ -55,6 +56,9 @@
  * @author <a href="mailto:stefano.maestri@redhat.com">Stefano Maestri</a>
  */
 public class RaDeploymentActivator {
+
+    public static final int CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER = 0x0180;
```
ropalka, line 60: Remove this. See comment below.

connector/src/main/java/org/jboss/as/connector/deployers/ds/processors/DriverManagerAdapterProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+            ByteArrayOutputStream bos = new ByteArrayOutputStream();
+            try {
+                final byte[] buffer = new byte[1024];
+                int read = 0;
+                while ((read = is.read(buffer)) != -1) {
+                    bos.write(buffer, 0, read);
+                }
+                return bos.toByteArray();
+            } finally {
+                is.close();
```
ropalka, line 93: calling is.close() might produce exception. Please define:

    private static void safeClose(final Closeable closeable) {
        if (closeable == null) return;
        try {
            closeable.close();
        } catch (IOException e) {
            // ignore
        }
    }

and use it instead of InputStream.close() method

ropalka (changes requested):

ropalka:
What would be also acceptable @tadamski is:
 * create PR against WildFly Core where you will define Phase.CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER = 0x0180
 * just fix safeClose() requirement in this PR 
 * create a follow up JIRA in WildFly requiring switch
   from RaDeploymentActivator.CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER
   to Phase.CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER
 * finally mention this JIRA in this PR request

ropalka:
Ah, I see @bstansberry suggested the same approach.

tadamski:
@ropalka this is only a temporary solution, to be able to release without core; there is already core PR with given phase created; I will fix safeClose in a moment

ropalka:
Just fix safeClose() requirement @tadamski and ignore other comments.

connector/src/main/java/org/jboss/as/connector/deployers/ra/RaDeploymentActivator.java
darranl: +1 in my recent DUP refactorings I have found a number of cases where the IDs used in WildFly don't get passed back to WildFly Core where we say we will come back to it later.

Please get a PR raised against WildFly Core and we should be able to get it merged quickly.

darranl (changes requested):

darranl:
I see the Core PR is merged, I have added the core-upgrade-needed label so we can merge this one after the next core upgrade.

darranl:
Also this may be a clean up task to consider at a later point, since the WildFly Core split this has always been hard to keep in sync.

From my recent work I think the individual phases need some more work to really define their entry and exit requirements, at the moment many are instead named in relation to the state of the class loader for the deployment.  

At the same point I think ti keep the projects cleaner we should have a hard requirement that in a single Phase all DUPs in WildFly Core execute before any DUPs in Wildfly - that way WildFly could become responsible for it's own numbering.  If any WildFly DUPs do need to execute before the WildFly Core DUPs either they need to move to an earlier phase or it is an indication that phase needs splitting in two.

tadamski:
safeClose issue fixed

ropalka:
I'll approve this PR @tadamski when you will link here WFCORE PR for adding constant 
Phase.CONFIGURE_JDBC_DRIVER_MANAGER_ADAPTER
plus WildFly JIRA issue to migrate to that Phase constant.

darranl:
@ropalka on this one I think we have plenty of time to just fix it correctly from the outset, the constant is in WildFly Core now and we are guaranteed that WildFly Core must be tagged before WildFly 22 can be tagged.

bstansberry:
@darranl OK now? I'd like to get this merged ASAP to give it as much bake as possible.

ropalka:
WildFly Core 14.0.0.Beta3 have been applied to WildFly master branch.
Please update this PR @tadamski to use constants defined there.

tadamski:
@ropalka fixed

ropalka (approved):

darranl (approved):
This PR is now using the Phase defined in WildFly Core.

gaol:
this has been synchronized to downstream PR, thanks all :)

PR #13753, opened 2020-11-30
[WFLY-11933] Fix metrics collector when RBAC is enabled

claudioweiler:
Fix https://issues.redhat.com/browse/WFLY-11933

This fix is based on @jmesnil PR: https://github.com/wildfly/wildfly/pull/13694

It uses a SuperUserClient to enable metrics collector to work when used with RBAC enabled.


bstansberry:
I think we need some kind of change in smallrye to provide a hook for WildFly to suppress processing of the metric in a given call context.

Metrics requests are different from health in that they provide information about potentially security-sensitive resources. We shouldn't provide super-user privileges to callers.

claudioweiler:
Hi @bstansberry.

I can't see this issue of security-sensitive resources. Actually, if RBAC are disable, all metrics are visible, and we have the `security-enable` attribute just for this, right?

So, if without RBAC it just works, setting RBAC should not change this behavior, because there are no user involved in the request. If I want that the metrics are protected, so I turn `security-enable` to true, and all requests will require an authenticated user.

See that my point is there are no user involved in this setup, then maybe this is a bigger problem with actual behavior, that even without RBAC this metrics should not be visible at all, but changing this behavior now could bring major issues to anyone that is relying on this and build all they monitoring system based on current behaviour.

bstansberry:
Thanks @claudioweiler and sorry it took so long to work through this issue.

Between the log spam fix in #14136, the proper handling of metrics output from https://issues.redhat.com/browse/WFLY-14694, and the identity propagation fix in https://issues.redhat.com/browse/WFCORE-5370, I think the basic handling of metrics when RBAC is enabled is now fitting the expected behavior with RBAC. Users who want to enable RBAC but still allow unauthenticated callers can map the 'anonymous' user to whatever RBAC role is appropriate for their setup. So I'm going to close this one.

PR #13756, opened 2020-12-01
[WFLY-14108] Introduce a base "metrics" subsystem and move microprofile-metrics to the microprofile feature-pack

jmesnil:
[WFLY-14108] Introduce a base "metrics" subsystem and move microprofile-metrics to the microprofile feature-pack

This PR is composed of a bunch of small commits during the development.
I will squash them in a single commit before merging but I kept them for the time being to help the review.

@bstansberry If you are reviewing this PR, could you please pay special attention to the 1st commit (that moves the microprofile-metrics extension from ee to full feature-pack (though the microprofile feature-pack). The rest of the commits are mostly code refactoring (moving some functionalities to the newly created `metrics` subsystem and use it from the `microprofile-metrics-smallrye` subsystem). The key change of all this PR is that the `observability` layer is now using the `metrics` layer instead of the `microprofile-metrics` layer.

I saw that some of my files are still missing copyright headers (I'll add them but I want to get a CI job running now).
I'll also add some diffs that shows that the metrics exposed by the `metrics` and `microprofile-metrics-subsystem` are identical (for WildFly management model and JVM metrics)

Finally, this PR will need a Core upgrade as the enums for the host-excludes is missing values for WildFly 20 & 21 (I've added & commented the relevant XML elements)

JIRA: https://issues.redhat.com/browse/WFLY-14108


ee-9/feature-pack/src/main/resources/feature_groups/standalone-ha.xml
```diff
@@ -22,7 +22,8 @@
     <feature-group name="management-interfaces"/>
     <feature-group name="access-control"/>
     <feature-group name="health"/>
-    <feature-group name="metrics" />
+    <!-- WFLY-14108 do not forget to put it back -->
```
jmesnil: All such comments were fixed in following commits after I created the new `metrics` subsystem.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/_private/MicroProfileMetricsLogger.java
```diff
@@ -39,7 +39,7 @@
  *
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2018 Red Hat inc.
  */
-@MessageLogger(projectCode = "WFLYMETRICS", length = 4)
+@MessageLogger(projectCode = "WFLYMPMETRICS", length = 4)
```
jmesnil, line 42: I changed the project code but we are fine as the new `metrics` subsystem reuse the `WFLYMETRICS` code and keep the same message id for the same errors and logs

microprofile/metrics-smallrye/src/main/resources/schema/wildfly-microprofile-metrics-smallrye_2_1.xsd
```diff
@@ -0,0 +1,57 @@
...
+           elementFormDefault="qualified"
+           attributeFormDefault="unqualified"
+           version="1.0">
+
+    <xs:element name="subsystem">
+        <xs:complexType>
+            <xs:attribute name="security-enabled" type="xs:boolean" default="true">
+                <xs:annotation>
+                    <xs:documentation>
+                        Deprecated. Use the security-enabled attribute on the metrics subsystem instead.
```
jmesnil, line 37: the only diff between this 2.1 and the 2.0 xsd is that deprecated documentation....

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
```diff
@@ -161,77 +161,88 @@
...
-                "org.jboss.as.threads",
-                "org.jboss.as.transactions",
-                "org.jboss.as.web",
-                "org.jboss.as.webservices",
-                "org.jboss.as.weld",
-                "org.jboss.as.xts",
-                // This module was added in EAP70, but we move it to the EAP62 extension list to allow the test passing
-                // without adding it to the host-exclude section. We don't want to expose it in the host-exclude.
-                "org.wildfly.extension.mod_cluster"
+        WILDFLY_19_0("WildFly19.0", WILDFLY_18_0,Arrays.asList(
```
jmesnil, line 164: I'm not sure why this was not spotted before but once I added the `org.wildfly.extension.metrics` to more recent WIldFly version, this was spotted in the CI

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml
```diff
@@ -2,7 +2,6 @@
 <layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="observability">
     <dependencies>
         <layer name="microprofile-config" optional="true"/>
-        <layer name="microprofile-metrics" optional="true"/>
```
bstansberry, line 5: We may have to redefine / override this in the 'wildfly' f-p as all of these layers are optional people might be excluding them using the 'microprofile-*' name.

If they are using 'wildfly-ee' then they have no choice; they have to change their exclude to 'metrics'.  But if they use 'wildfly' their existing exclude could work.

@jfdenise @yersan WDYT? Overriding a layer isn't nice either, and having a fixed definition for 'observability' is also a good thing.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+[[metrics-http-endpoint]]
+== HTTP Endpoint
+
+The Metric HTTP endpoint is accessible on WildFly HTTP management interface http://localhost:9990/metrics[http://localhost:9990/metrics].
+
+Secured access to the HTTP endpoint is controlled by the `security-enabled` attribute of the `/subsystem=metrics` resource.
+If it is set to `true`, the HTTP client must be authenticated.
+
+If the application server is healthy, it will return a `200 OK` response:
```
bstansberry, line 46: I'm not sure if 'healthy' is a good word here. If I'm thinking about metrics and hear 'health' I think in terms of what a liveness/readiness probe would return, which I *think* is not what you mean here. There may have been boot errors etc that would reflect poor 'health' but would not result in a non-200 from /metrics.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+If the authentication fails, the  server will reply with a `401 NOT AUTHORIZED` response.
+
+== Exposed Metrics
+
+The HTTP endpoint exposes the following metrics:
+
+* Base metrics - Metrics from JVM MBeans (read from their JMX MBeans)
+* Vendor metrics - WildFly Metrics from the management model subsystem and deployment subtrees.
+
+The HTTP endpoint exposes the metrics in the Prometheys format only.
```
bstansberry, line 82: s/Promethys/Prometheus

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+=== WildFly Metrics Description
+
+WildFly metrics names is based on the subsystem that provides them as well as the name of the attribute from the management model.
```
bstansberry, line 86: s/names is/names are/g

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
+[[MicroProfile_Metrics_SmallRye]]
+= Metrics Subsystem Configuration
+
+[NOTE]
+====
+This subsystem exposes only base metrics from WildFly Management Model and JVM MBeans.
```
bstansberry, line 6: s/from WildFly/from the WildFly/g

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+* Base metrics - Metrics from JVM MBeans (read from their JMX MBeans)
+* Vendor metrics - WildFly Metrics from the management model subsystem and deployment subtrees.
+
+The HTTP endpoint exposes the metrics in the Prometheys format only.
+
+=== WildFly Metrics Description
+
+WildFly metrics names is based on the subsystem that provides them as well as the name of the attribute from the management model.
+Their name can also be prepended with a `prefix` (specified on the `/subsystem=metrics` resource).
+Other information are stored using labels.
```
bstansberry, line 88: s/are stored/is stored/g

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+that can be deployed in WildFly.
+A corresponding metric will be exposed for it with the name and labels:
+
+* `wildfly_undertow_request_count_total{deployment="helloworld-rs.war",servlet="org.jboss.as.quickstarts.rshelloworld.JAXActivator",subdeployment="helloworld-rs.war"}`
+
+[NOTE]
+Some subsystems (such as `undertow` or `messaging-activemq`) do not enable their statistics by default
+as they have an impact on performance and memory usage. These subsystems provides a `statistics-enabled` attribute that must
+be set to `true` to enable them.
+For convenience, WildFly standalone configuration provides expression to enable the statistics by setting a
```
bstansberry, line 104: s/expression/expressions

metrics/src/main/java/org/wildfly/extension/metrics/MetricsCollectorService.java
```diff
@@ -49,47 +47,43 @@
...
         this.processStateNotifier = processStateNotifier;
-        this.exposedSubsystems = exposedSubsystems;
-        this.globalPrefix = globalPrefix;
+        this.metricCollectorConsumer = metricCollectorConsumer;
     }
 
     @Override
     public void start(StartContext context) {
-        modelControllerClient = modelControllerClientFactory.get().createClient(managementExecutor.get());
+        modelControllerClient = modelControllerClientFactory.get().createSuperUserClient(managementExecutor.get());
```
bstansberry, line 76: I don't think we can do this with metrics. :(

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public OptionalDouble getValue() {
+        ModelNode result = readAttributeValue(address, attributeName);
+        if (result.isDefined()) {
+            try {
+                return OptionalDouble.of(result.asDouble());
+            } catch (Exception e) {
+                throw MetricsLogger.LOGGER.unableToConvertAttribute(attributeName, address, e);
```
bstansberry, line 60: This will blow up the entire read.

With Smallrye I believe if a Metric throws an exception it is caught and logged but the loop reading all the metrics continues, so other metrics get reported. A failure here though will abort the PrometheusExporter loop.

Since you are providing an Optional it seems like this could log the failure and fall through.

RBAC failures could be logged at DEBUG/TRACE.

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
```diff
@@ -0,0 +1,86 @@
...
+    private ModelNode readAttributeValue(PathAddress address, String attributeName) {
+        final ModelNode readAttributeOp = new ModelNode();
+        readAttributeOp.get(OP).set(READ_ATTRIBUTE_OPERATION);
+        readAttributeOp.get(OP_ADDR).set(address.toModelNode());
+        readAttributeOp.get(ModelDescriptionConstants.INCLUDE_UNDEFINED_METRIC_VALUES).set(false);
+        readAttributeOp.get(NAME).set(attributeName);
+        ModelNode response = modelControllerClient.execute(readAttributeOp);
+        String error = getFailureDescription(response);
+        if (error != null) {
+            LOGGER.unableToReadAttribute(attributeName, address, error);
```
bstansberry, line 75: RBAC failures could be logged at a lower level.

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
```diff
@@ -0,0 +1,122 @@
...
+        this.metricRegistry = metricRegistry;
+        this.managementExecutor = managementExecutor;
+        this.exposeAnySubsystem = exposeAnySubsystem;
+        this.exposedSubsystems = exposedSubsystems;
+        this.prefix = prefix;
+    }
+
+    @Override
+    public void start(StartContext startContext) {
+        System.out.println("DeploymentMetricService.start");
```
bstansberry, line 77: ^^^

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
```diff
@@ -0,0 +1,122 @@
...
+        final Runnable task = new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    registration = new MetricRegistration(metricRegistry.get());
+                    System.out.println(">>> JFM DeploymentMetricService.start");
```
bstansberry, line 83: ^^^

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
```diff
@@ -0,0 +1,122 @@
...
+                    metricCollector.get().collectResourceMetrics(rootResource,
+                            managementResourceRegistration,
+                            // prepend the deployment address to the subsystem resource address
+                            address -> deploymentAddress.append(address),
+                            exposeAnySubsystem, exposedSubsystems, prefix,
+                            registration);
+                } catch (Throwable t) {
+                    System.out.println(">>> JFM DeploymentMetricService.error");
```
bstansberry, line 91: ^^^

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
```diff
@@ -0,0 +1,122 @@
...
+        } catch (RejectedExecutionException e) {
+            task.run();
+        } finally {
+            startContext.asynchronous();
+        }
+    }
+
+    @Override
+    public void stop(StopContext stopContext) {
+        System.out.println("JFM DeploymentMetricService.stop start ");
```
bstansberry, line 109: ^^^

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
```diff
@@ -0,0 +1,122 @@
...
+        registration.unregister();
+        System.out.println("JFM DeploymentMetricService.stop end");
```
bstansberry, line 111: ^^^

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/_private/MicroProfileMetricsLogger.java
```diff
@@ -57,12 +57,15 @@
     @Message(id = 2, value = "Failed to initialize metrics from JMX MBeans")
     IllegalArgumentException failedInitializeJMXRegistrar(@Cause IOException e);
 
+    // no longer used
```
bstansberry, line 60: No longer used methods can be commented to save metaspace.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileVendorMetricRegistry.java
```diff
@@ -0,0 +1,204 @@
...
+                }
+
+                @Override
+                public void inc(long n) {
+                }
+
+                @Override
+                public long getCount() {
+                    OptionalDouble value = metric.getValue();
+                    return  Double.valueOf(value.orElse(0)).longValue();
```
bstansberry, line 66: It would be nice if SmallRye had something like a

```
public interface OptionalMetric extends Metric {

  Optional<Metric> getOptional();
}
```

Then check for that when exporting and if that's the Metric type use the Optional, skipping export !optional.isPresent().

Responding with zero exposes the metric names which in theory could be privileged information.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
jmesnil: I agree. That looks like a bad copy/paste from the Microprofile-health guide.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml
jfdenise: @bstansberry +1 on redefining the observability layer in wildfly FP to add back the optional dependency on the microprofile-metrics.

jfdenise: I don't see the new "metrics" base layer in observability. User can't replace their exclusion to "metrics". I missed something there?

jmesnil: @jfdenise https://github.com/wildfly/wildfly/blob/9c2367ec9fad676658cea070bf6e9e8af95270cf/ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml#L4 or am I misunderstanding what you mean?

metrics/src/main/java/org/wildfly/extension/metrics/MetricsCollectorService.java
jmesnil: you are right, I changed the code in https://github.com/wildfly/wildfly/pull/13756/commits/9c2367ec9fad676658cea070bf6e9e8af95270cf to use a regular client and skip any metrics that can not be read due to RBAC

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
jmesnil: done

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
jmesnil: done

metrics/src/main/java/org/wildfly/extension/metrics/deployment/DeploymentMetricService.java
jmesnil: oops, sorry about that

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileVendorMetricRegistry.java
jmesnil: Yes, that's a big difference between the base `metrics` and the `microprofile-metrics-smallrye`subsystem.

In the base metrics, if the metric value is not defined, I skip exposing any info about that (including its name and description): https://github.com/wildfly/wildfly/blob/9c2367ec9fad676658cea070bf6e9e8af95270cf/metrics/src/main/java/org/wildfly/extension/metrics/PrometheusExporter.java#L48

With MicroProfile Metrics, any registered metric is considered valid and I have no point where I can "validated" the metric value before it is printed out.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml
jfdenise: @jmesnil , thank-you I should have looked at a commit, not the actual full PR. The 'metrics' is there, that answers my question.

yersan: > WDYT? Overriding a layer isn't nice either, and having a fixed definition for 'observability' is also a good thing.

microprofile-config and microprofile-health subsystems will be also moved out from wildfly-ee. I'm not sure what observability will contain in wildfly-ee. It seems it will tend to be a complete different layer on wildfly-ee vs wildfly fp. Re-define it sounds although it seems users will see a different server configuration when using wildfly-ee and wildfly, so the same scripts won't work for the same layer. Am I wrong?

jmesnil:
I've checked that this PR does not introduce (too many) changes in the metrics output compared to the existing branch.
I've added a gist at https://gist.github.com/jmesnil/6253ea2150f22f379ce5a9b7d569d795 to show 4 outputs:
* `master-standalone-full.txt`  is the output from the master branch with the `standalone-full.xml` config
* `master-standalone-microprofile.txt`  is the output from the master branch with the `standalone-microprofile.xml` config
* `current-standalone-full.txt`  is the output from this PR with the `standalone-full.xml` config
* `current-standalone-microprofile.txt`  is the output from the master branch with the `standalone-microprofilel.xml` config

There is no difference between  `master-standalone-microprofile.txt` & `current-standalone-microprofile.txt` (as they both use smallrye-metrics to output the metrics)/

There are *some* differences between `master-standalone-full.txt` and `current-standalone-microprofile.txt`:
* the order of the tags may have changed (that's ok)
* the `microprofile_scope="vendor"` tag is no longer present in the current base output (that's ok)
* the `vendor_memoryPool` memory pool names have changed. The master branch is silly and makes consuming these metrics very difficult. This PR has moved the *name* of the memory pool in a tag (which is more consistent with all the other metrics output).
  * Before, it was `vendor_memoryPool_Code_Cache_usage_bytes` & `vendor_memoryPool_Compressed_Class_Space_usage_bytes`(with different descriptions)
  * Now it is `vendor_memoryPool_usage_bytes{name="Code Cache"}` & `vendor_memoryPool_usage_bytes{name="Compressed Class Space"}` (with a single description)
* I'm pretty sure that the actual output for counter metrics with units is wrong. smallrye-metrics output the metric with the name `wildfly_undertow_total_request_time_total_seconds` while I think it should be wildfly_undertow_total_request_time_seconds_total`. I'm using the same convention in the base `metrics` subsystem so I'm bug-compatible but I'll open an issue agains smallrye-metrics (or microprofile-metrics) to clarify this later.

Note that I have only checked the output using `standalone-full.xml` and `standalone-microprofile.xml` with a deployed JAX-RS app (so I had some Undertow deployment metrics). It does not cover _all_ metrics that can be exposed by WildFly but it gives a pretty good idea about the compatibility.
As far as I am concerned, we're good (we could object about the `vendor_memoryPool` metrics but (a) they are not required by MicroProfile Metrics, (b) the current convention is inconsistent)

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml
bstansberry: In wildfly-ee observability will contain 'health' and 'metrics', the two most important aspects of observing a server. It won't have any form of tracing. Having MP Config was always more an implementation detail not a feature of 'observability'.

So, I do think the observability concept makes sense for wildfly-ee.

Yes, redefining the layer means what your configuration looks like will be different depending on what top-level FP you use. That does make it more unlikely someone can use the same script for both. Or the script becomes more complex.

I think if people were customizing they are going to have to adapt no matter what we do. For example the 'security-enabled' attribute on the mp-metrics-smallrye subsystem is no longer functional. And if you're using wildfly-ee the mp subsystems are just gone.

metrics/src/main/java/org/wildfly/extension/metrics/_private/MetricsLogger.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @LogMessage(level = WARN)
+    @Message(id = 3, value = "Unable to read attribute %s on %s: %s.")
+    void unableToReadAttribute(String attributeName, PathAddress address, String error);
+
+    @LogMessage(level = WARN)
+    @Message(id = 4, value = "Unable to convert attribute %s on %s to Double value.")
+    void unableToConvertAttribute(String attributeName, PathAddress address, @Cause Exception exception);
+
+    @LogMessage(level = INFO)
```
bstansberry, line 67: This still seems high.  IIRC when I turned on RBAC to look at WFLY-11933 there were tons of items logged. If something was polling fairly frequently it would be really spammy.

Hmm, I probably hadn't secured the endpoint so maybe I was getting logging for every management API metric. I suppose someone might do that though.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileVendorMetricRegistry.java
bstansberry: @jmesnil If you think the OptionalMetric idea is useful I can file an RFE for such an SPI in smallrye. It's kind of hacky because it extends 'Metric' just to be an acceptable type for the registry, but the normal Metric impl really isn't meant to be used.

bstansberry:
@jmesnil Thanks for the updates. I recommend calling out the vendor_memoryPool thing on your requirements doc, just because it says there will be no name changes. I agree the use of tags is better and the other way could be regarded as a bug.

metrics/src/main/java/org/wildfly/extension/metrics/_private/MetricsLogger.java
jmesnil: ok, I'll do that in a DEBUG log (and get rid of the i18ned message)

jmesnil:
> @jmesnil Thanks for the updates. I recommend calling out the vendor_memoryPool thing on your requirements doc, just because it says there will be no name changes. I agree the use of tags is better and the other way could be regarded as a bug.

done

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/observability/layer-spec.xml
yersan: My concern was oriented toward those cases thinking that since observability is going to have notable differences between the fp consumed we should have different names for such layer. I wanted to be sure it's assumable.

Then it means an user using the ee-wildfly fp that wants to incorporate MP specs moving to wildfly fp, would need to redefine / adapt any CLI scripts if he is using observability Galleon layer (or any layer that depends upon it, e.g cloud-profile and cloud-server)

bstansberry (approved):
LGTM.

bstansberry:
@jmesnil I filed https://github.com/jmesnil/wildfly/pull/4 to correct some issues I found.

bstansberry (changes requested):
I found some issues; please see https://github.com/jmesnil/wildfly/pull/4

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+----
+
+== Management Model
+
+The `/subsystem=metrics` resource defines three attributes:
+
+* `security-enabled` - a boolean to indicate whether authentication is required to access the HTTP metrics endpoint (described below). By default, it is `true`. The
+standalone configurations explicitly sets it to `false` to accept unauthenticated access to the HTTP endpoints.
+* `exposed-subsystems` - a list of strings corresponding to the names of subsystems that exposes their metrics in the HTTP metrics endpoints.
+By default, it is not defined (there will be no metrics exposed by subsystem. The special wildcard `*` can be used to expose metrics from _all_ subsystems. The standalone
```
mchoma, line 34: I do not see `*` rendered properly in github Preview mode.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
```diff
@@ -0,0 +1,105 @@
...
+configuration sets this attribute to `*`.
```
mchoma, line 35: Same. Not rendered properly.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Metrics.adoc
```diff
@@ -35,9 +35,9 @@ standalone configurations explicitly sets it to `false` to accept unauthenticate
 The Metric HTTP endpoint is accessible on WildFly HTTP management interface http://localhost:9990/metrics[http://localhost:9990/metrics].
 
 Secured access to the HTTP endpoint is controlled by the `security-enabled` attribute of the `/subsystem=microprofile-metrics-smallrye` resource.
-If it is set to `true`, the HTTP client must be authenticated.
+The value of this attribute will override the `security-enabled` attribute of the `/subsystem=metrics` resource (documented in <<metrics-http-endpoint,Metrics subsystem configuration guide>>).
```
mchoma, line 38: What about `exposed-subsystems` and `prefix` does same overriding logic apply as well in case of these properties.

If `metrics` subsystem is responsible for registring /metrics http endpoint and will be always present, do we really need this overriding logic?

jmesnil: `exposed-subsystems` and `prefix` are specific to the `microprofile-metrics-smallrye` and do not provide this overriding logic.

This special case to handle the `security-enabled` is to keep it backwards compatible. Users that change this value expects to be able to use it. Making it a no-op would have break their configuration.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
jmesnil: fixed (this seems to be an issue with Asciidoc rendering

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Metrics.adoc
jmesnil: fixed

jmesnil:
I've squashed all the commits (incuding the ones that Brian created in jmesnil#4)

bstansberry (approved):

jmesnil:
@bstansberry Pre-check is "Done" for EAP7-1566. I've removed the `missing-reqs` label and this PR can be merged.

PR #13758, opened 2020-12-01
[WFLY-14093] deprecate XTS

ochaloup:
https://issues.redhat.com/browse/WFLY-14093

### DIFF
--- a/xts/src/main/java/org/jboss/as/xts/XTSSubsystemDefinition.java
+++ b/xts/src/main/java/org/jboss/as/xts/XTSSubsystemDefinition.java
@@ -87,6 +87,7 @@ public class XTSSubsystemDefinition extends SimpleResourceDefinition {
                 XTSExtension.getResourceDescriptionResolver(null),
                 XTSSubsystemAdd.INSTANCE,
                 XTSSubsystemRemove.INSTANCE);
+        setDeprecated(ModelVersion.create(3,0,0));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


ochaloup:
@yersan can you please check if this fine as we discussed? Thanks a lot.

xts/src/main/java/org/jboss/as/xts/XTSSubsystemDefinition.java
```diff
@@ -87,6 +87,7 @@ protected XTSSubsystemDefinition() {
                 XTSExtension.getResourceDescriptionResolver(null),
                 XTSSubsystemAdd.INSTANCE,
                 XTSSubsystemRemove.INSTANCE);
+        setDeprecated(ModelVersion.create(3,0,0));
```
bstansberry, line 90: In https://github.com/wildfly/wildfly/blob/master/xts/src/main/resources/org/jboss/as/xts/LocalDescriptions.properties you need to add an 'xts.deprecated' entry that provides the message the user sees if they configure the deprecated resource.

ochaloup: @bstansberry thanks a lot for guidance! I can see the build failures now. Just, I would bet that I was able to build the WFLY with the older version of WFLY sources (~ 7 days ago). But I had to be somehow ignorant. Sorry for this omission.

It should be fixed now.

PR #13761, opened 2020-12-02
[WFLY-6660]: Allow to dump Artemis journal to file using WildFly Management API.

ehsavoie:
* Adding an admin-only operation to print the data of a broker.
* Testing the data printed in safe and full mode.

Jira: https://issues.redhat.com/browse/WFLY-6660
Proposal: https://github.com/wildfly/wildfly-proposals/pull/350

Should be rebased on https://github.com/wildfly/wildfly/pull/13906 when it is merged


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/PrintDataTestCase.java
```diff
@@ -0,0 +1,296 @@
...
+        for (String line : Files.readAllLines(dataPath)) {
+            if (!inMessages) {
+                inMessages = "M E S S A G E S   J O U R N A L".equals(line);
+            } else {
+                if (!hasQueueCount) {
+                    hasQueueCount = SAFE_QUEUE_COUNT.matcher(line).find();
+                }
+            }
+        }
+        Assert.assertTrue("Should have a message count of 1 for " + jmsQueueName, hasQueueCount);
```
jbliznak, line 219: This check is not enough, the test passes even with `printData(dataPath, false, false); // ~ secure = false` 

but I guess we want to check the secure info is _not printed_ and that is also described in analysis:
> - check that it displays the correct number of total messages and the correct number of message in the queue but doesn't display the queue infos.

jbliznak (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/PrintDataTestCase.java
jbliznak: Updated, thanks.

jbliznak (changes requested):
Tests are now testing what they should.

However, I have 2 more points:
First, nice to have - test methods are doing a lot of the same checks, might worth to refactor with methods.

Second, serious one - the tests are prone to leave server in incorrect state (reloaded to admin-only mode) if they fail - I suppose the reloading back to normal mode should be done in `@After` method or in a finally block.

ehsavoie:
@bstansberry ping for merge :)

jbliznak (approved):

PR #13765, opened 2020-12-02
[WFLY-14151] Move microprofile-health extension to the microprofile feature pack

jmesnil:
[WFLY-14151] Move microprofile-health extension to the microprofile feature pack

This PR will handle the move of microprofile-health extension to the microprofile feature pack and the creation of a health extension to provide base healthiness features in the EE feature pack

JIRA: https://issues.redhat.com/browse/WFLY-14151

Signed-off-by: Jeff Mesnil <jmesnil@redhat.com>

PR #13766, opened 2020-12-02
[WFLY-14148]: Move microprofile opentracing from wildfly-ee to wildfly.

ehsavoie:
* removing opentracing from the ee feature pack and moving it to the
  complete feature pack.

Jira: https://issues.redhat.com/browse/WFLY-14148


bstansberry (changes requested):
Besides my various comments below, the mixed-domain stuff will need some cleanup as domain.xml no longer has OT stuff to remove:

```
$ cd testsuite/mixed-domain/
yingtao:mixed-domain bstansberry$ git grep tracing
src/test/java/org/jboss/as/test/integration/domain/mixed/eap710/DomainAdjuster710.java:        list.addAll(removeMicroProfileOpenTracing(profileAddress.append(SUBSYSTEM, "microprofile-opentracing-smallrye")));
src/test/java/org/jboss/as/test/integration/domain/mixed/eap710/DomainAdjuster710.java:     * microprofile-opentracing subsystem and extension do not exist in previous versions, so remove it
src/test/java/org/jboss/as/test/integration/domain/mixed/eap710/DomainAdjuster710.java:        list.add(createRemoveOperation(PathAddress.pathAddress(EXTENSION, "org.wildfly.extension.microprofile.opentracing-smallrye")));
src/test/java/org/jboss/as/test/integration/domain/mixed/eap730/DomainAdjuster730.java:        adjustOpenTracing(ops, profileAddress.append(SUBSYSTEM, "microprofile-opentracing-smallrye"));
```

There are a bunch of other TS failures that I don't understand.

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile-ha.xml/config.xml
```diff
@@ -7,10 +7,19 @@
...
         <include name="web-clustering"/>
     </layers>
+    <feature spec="subsystem.microprofile-opentracing-smallrye">
+        <param name="default-tracer" value="jaeger"/>
+        <feature spec="subsystem.microprofile-opentracing-smallrye.jaeger-tracer">
+            <param name="jaeger-tracer" value="jaeger"/>
+            <param name="sampler-param" value="1"/>
+            <param name="sampler-type" value="const"/>
+        </feature>
+    </feature>
```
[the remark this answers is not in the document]
ehsavoie: for backward compatibility

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile.xml/config.xml
```diff
@@ -7,12 +7,21 @@
...
     <feature-group name="distributable-web"/>
     <feature-group name="infinispan-local-web"/>
+    <feature spec="subsystem.microprofile-opentracing-smallrye">
+        <param name="default-tracer" value="jaeger"/>
+        <feature spec="subsystem.microprofile-opentracing-smallrye.jaeger-tracer">
+            <param name="jaeger-tracer" value="jaeger"/>
+            <param name="sampler-param" value="1"/>
+            <param name="sampler-type" value="const"/>
+        </feature>
+    </feature>
```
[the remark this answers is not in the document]
ehsavoie: ok fixed

ee-9/feature-pack/wildfly-feature-pack-build.xml
```diff
@@ -123,7 +122,6 @@
                 <extension>org.wildfly.extension.microprofile.jwt-smallrye</extension>
                 <extension>org.wildfly.extension.microprofile.metrics-smallrye</extension>
                 <extension>org.wildfly.extension.microprofile.openapi-smallrye</extension>
-                <extension>org.wildfly.extension.microprofile.opentracing-smallrye</extension>
```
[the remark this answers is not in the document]
ehsavoie: ok

ehsavoie:
@bstansberry it should fix the test failures that is due to the fact that the configuration with layers differs from the one produced in full.

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile-ha.xml/config.xml
```diff
@@ -7,10 +7,12 @@
...
         <include name="microprofile-jwt"/>
         <include name="microprofile-openapi"/>
+        <include name="microprofile-opentracing"/>
         <include name="undertow-legacy-https"/>
         <exclude name="management-security-realm"/>
         <exclude name="jpa"/>
         <include name="jpa-distributed"/>
         <include name="web-clustering"/>
     </layers>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
bstansberry, line 17: The current standalone-microprofile-ha.xml uses the empty config.

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile.xml/config.xml
```diff
@@ -7,12 +7,14 @@
...
         <include name="microprofile-jwt"/>
         <include name="microprofile-openapi"/>
+        <include name="microprofile-opentracing"/>
         <include name="undertow-legacy-https"/>
         <exclude name="management-security-realm"/>
     </layers>
     <!-- TODO WFLY-13150 encapsulate these in a layer -->
     <feature-group name="distributable-web"/>
     <feature-group name="infinispan-local-web"/>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
bstansberry, line 17: The current standalone-microprofile-ha.xml uses the empty config.

microprofile/galleon-common/src/main/resources/layers/standalone/open-tracing/layer-spec.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="open-tracing">
```
bstansberry, line 2: I think inverting this is more useful (i.e. microprofile-opentracing depends open-tracing, with open-tracing having the features). In any case where the user adds either, it works. If anyone uses one of our aggregate layers, like microprofile-platform or observability, if they exclude either one it works.

The microprofile-opentracing would have to optionally depend open-tracing though; otherwise you could not exclude open-tracing.

In WF 24 we could drop open-tracing.

bstansberry (changes requested):
Could you diff the target/wildfly-.../standalone/configuration directory produced by ee-9/build vs build? There should be no differences related to OT. I'm pretty sure some things need to be added to ee-9/feature-pack.

microprofile/galleon-common/src/main/resources/layers/standalone/open-tracing/layer-spec.xml
ehsavoie: ok

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile.xml/config.xml
ehsavoie: ok, I'e realigned the produced configurations with the ones in 21

galleon-pack/src/main/resources/configs/standalone/standalone-microprofile-ha.xml/config.xml
ehsavoie: ok, I've realigned the produced configurations with the ones in 21

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/opentracing/OnOffOpenTracingTestCase.java
```diff
@@ -61,6 +61,7 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
+@Ignore("WFLY-14199")
```
bstansberry, line 64: This needs an import statement

bstansberry:
@ehsavoie this kind of thing will be needed for OT as well:

https://github.com/wildfly/wildfly/pull/13765/files#diff-352e26d1bdbe634b8174a0fac0966a47829c74adf0a6037e5f49ea5a8015e214

ee-9/feature-pack/src/main/resources/feature_groups/standalone-full.xml
```diff
@@ -3,6 +3,7 @@
 
     <feature-group name="standalone-full-x-messaging"/>
     <feature-group name="messaging-remote-activemq"/>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
jmesnil, line 6: @bstansberry is that intended?

I thought that MP specs would be made available only in the `standalone-microprofile` configurations even for EE9

ee-9/feature-pack/src/main/resources/feature_groups/standalone-ha.xml
```diff
@@ -26,6 +26,7 @@
 
     <feature-group name="servlet-unsecured-standalone-profile"/>
     <feature-group name="basic-ha-profile"/>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
jmesnil, line 29: idem

ee-9/feature-pack/src/main/resources/feature_groups/standalone-full.xml
bstansberry: @jmesnil @ehsavoie Good question. Let's discuss when we talk in a few minutes and we can post back here.

Emmanuel may have had this all along or may have added it in response to my request that he compare outputs and eliminate diffs.  But that was based on my unthinking idea that these specs are in the 'wildfly' and 'wildfly-preview' FPs so there's no reason the resulting configs should change. But they do, correctly I think, when it comes to health and metrics, so why not differ about this too?

ehsavoie: @jmesnil @bstansberry yes the goal was to have a minimal diff with the preview release. Maybe removing this from preview should come in a 2nd step ?

ee-9/feature-pack/src/main/resources/feature_groups/domain-ha-profile.xml
```diff
@@ -11,5 +11,6 @@
     </feature-group>
 
     <feature-group name="basic-ha-profile"/>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
jmesnil, line 14: that one might not be correct (see https://github.com/wildfly/wildfly/pull/13806#issuecomment-742650956)

ee-9/feature-pack/src/main/resources/feature_groups/domain-profile.xml
```diff
@@ -5,5 +5,6 @@
 
     <feature-group name="basic-profile"/>
     <feature-group name="infinispan-local"/>
+    <feature-group name="microprofile-opentracing-jaeger"/>
```
jmesnil, line 8: idem (https://github.com/wildfly/wildfly/pull/13806#issuecomment-742650956)

ee-9/feature-pack/src/main/resources/feature_groups/standalone-full.xml
jmesnil: ok, let's make the diff minimal and define the rules for inclusion before this preview distribution is finalized

bstansberry: @jmesnil @ehsavoie +1 to make the diff minimal for this OT change.

A summary, hopefully correct, of what we discussed.

1) The 'wildfly' and 'wildfly-preview' FPs were not meant to differ re: MP. Both contain the MP extensions and unlike, say, legacy security, we're not planning to drop the extensions. We could use wildfly-preview to preview some change in how we offer MP in our standard configs but that wasn't the original intent.

2) It's a valid question whether we want to continue offering the MP specs in our main distribution's standard configs other than standalone-microprofile[-ha].xml. That discussion would also encompass MP JWT, which we include in standalone[-xxx] and the various domain.xml profiles.

But changing those things seems beyond the scope of what this PR is about.

jmesnil (changes requested):
could you please cherry-pick https://github.com/wildfly/wildfly/pull/13806/commits/909a0bdca1884079818dc7bada97af4fabede847 that is missing to have a correct configuration file in dist-legacy?

ee-9/feature-pack/src/main/resources/feature_groups/domain-ha-profile.xml
bstansberry: I think this is correct and the problem is in the 'wildfly' feature pack. In current master both 'ee-9/dist' and 'dist' include MP OT in their standard configs, while with this PR 'dist' no longer does.  As discussed starting at https://github.com/wildfly/wildfly/pull/13766#discussion_r540134348 we should decide separately about keeping MP subsystems in our standard configs other than standalone-microprofile[-ha].xml.

ehsavoie:
@jmesnil done

jmesnil (approved):

jmesnil:
@ehsavoie there is a checkstyle error in this PR:

```
 [org.wildfly:wildfly-ts-integ-microprofile] [ERROR] /store/work/tc-work/f5da3564a57e9d74/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/opentracing/OnOffOpenTracingTestCase.java:55:1: Duplicate import to line 50 - org.junit.Ignore. [RedundantImport]
```

bstansberry (approved):

jmesnil:
@ehsavoie thanks!

PR #13772, opened 2020-12-03
WFLY-13792 Test for JSF deployment failure due to UnsupportedOperationExceptio…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-13792

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.test.integration.jsf.unsupportedwar;
+
+import java.net.URL;
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.util.EntityUtils;
+import org.hamcrest.CoreMatchers;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * <p>Tests the JSF deployment failure due to UnsupportedOperationException when
+ * <em>javax.faces.FACELETS_VIEW_MAPPINGS</em> is defined with something that
+ * does not include <em>*.xhtml</em>.</p>
+ *
+ * For details check https://issues.redhat.com/browse/WFLY-13792
+ *
+ * @author Ranabir Chakraborty <rchakrab@redhat.com>
+ *
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class FaceletsViewMappingsTestCase {
+
+    private static final String DEPLOYMENT = FaceletsViewMappingsTestCase.class.getSimpleName();
+
+    @ArquillianResource
+    private URL url;
+
+    @Deployment
+    public static WebArchive deployment() {
+        final Package DEPLOYMENT_PACKAGE = HelloBean.class.getPackage();
+        WebArchive war = ShrinkWrap.create(WebArchive.class, DEPLOYMENT + ".war");
+        war.addClasses(HelloBean.class, ConfigurationBean.class);
+        war.addAsWebInfResource(DEPLOYMENT_PACKAGE, "web.xml", "web.xml");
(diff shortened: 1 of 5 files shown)


gaol:
this depends on https://github.com/wildfly/wildfly/pull/13771 to pass

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @SuppressWarnings("deprecation")
+    @Test
+    public void testHelloWorld() throws Exception {
+        try (CloseableHttpClient httpclient = HttpClients.createDefault()) {
+            String Url = url.toExternalForm();
+            HttpGet httpget = new HttpGet(Url);
+            httpclient.execute(httpget);
+
+            HttpResponse response = httpclient.execute(httpget);
```
yersan, line 73: There are two `httpclient.execute(httpget)` requests here, which is unnecessary to perform this test.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
```diff
@@ -0,0 +1,80 @@
...
+            Assert.assertEquals("Operation failed", HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
```
yersan, line 74: Suggestion, instead of "Operation failed", could you add something more informative? that would helps if you are reviewing the logs. "Operation failed" sounds like there is a management operation involved. In addition, you could add the result of the `response.getStatusLine().getStatusCode()` so you can easily see in the logs what was returned by the previous HTTP request.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/HelloBean.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.jboss.as.test.integration.jsf.unsupportedwar;
+
+import java.io.Serializable;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Named;
+
+@Named(value = "helloBean")
+@RequestScoped
+public class HelloBean implements Serializable {
```
yersan, line 25: Harmless, but Serializable is not needed for RequestScope, feel free to remove it to make the code more clean.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/web.xml
```diff
@@ -0,0 +1,34 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<web-app version="4.0"
+	xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_4_0.xsd">
+
+	<context-param>
+		<param-name>javax.faces.DEFAULT_SUFFIX</param-name>
```
yersan, line 23: tabs vs spaces, could you use spaces to be consistent?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
```diff
@@ -0,0 +1,80 @@
...
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * <p>Tests the JSF deployment failure due to UnsupportedOperationException when
+ * <em>javax.faces.FACELETS_VIEW_MAPPINGS</em> is defined with something that
+ * does not include <em>*.xhtml</em>.</p>
```
yersan, line 41: Maybe this was only my impression, but when I read the javadoc of this test I was expecting that the test was going to assert that a deployment will fail with an `UnsupportedOperationException` when `javax.faces.FACELETS_VIEW_MAPPINGS` does not includes *.xhtml, I had to navigate to the Jira to understand it better.

If that is only my impression, that's fine, but my feeling was we can describe it differently. Something like verifies that an JSF application can be deployed if `javax.faces.FACELETS_VIEW_MAPPINGS` does not includes *.xhtml.

yersan (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
RanabirChakraborty: the idea of the testing is that the `javax.faces.FACELETS_VIEW_MAPPINGS` has some real value. Like FACELETS_VIEW_MAPPINGS set to *.jsf, then in the test just deploy the app and access the hello.jsf and check.

gaol:
retest this please

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/unsupportedwar/FaceletsViewMappingsTestCase.java
bstansberry: @RanabirChakraborty 

I still don't understand what this test is doing from the above text and looking at the JIRAs. The various JIRAs talk about *.jsp etc and it's not clear what the expected behavior is without reading a lot of discussion and then kind of guessing. Because it's not obvious in this case I think it's worthwhile having a clear explanation in the test.

I think the wording should be:

"Tests that JSF deployment failure does not occur when <em>javax.faces.FACELETS_VIEW_MAPPINGS</em> is defined with something that does not include <em>*.xhtml</em>."

That is what the test is doing.

RanabirChakraborty: The issue was that the default value of `javax.faces.FACELETS_VIEW_MAPPINGS` is *.xhtml and in wildfly adding the parameter FACELETS_VIEW_MAPPINGS that not containing the *.xhtml value triggers this issue. 

That's why in this test we are using *.jsf instead of default *.xhtml,(this test will eventually fail now, because the code fix in there in mojarra, but when the component upgrade will merge, this test should pass without failing)

PR #13774, opened 2020-12-03
 [WFLY-14156]: Remove MicroProfile REST Client from Wildfly ee.

ehsavoie:
* Making rest-client api optionnal in resteasy-cdi and removing it from the Jaxrs subsystem
 * Adding a new microprofile-rest-client in the microprofile galleon pack

Jira: https://issues.redhat.com/browse/WFLY-14156


microprofile/galleon-common/src/main/resources/layers/standalone/microprofile-rest-client/layer-spec.xml
```diff
@@ -0,0 +1,27 @@
...
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="microprofile-rest-client">
+    <packages>
+        <package name="org.jboss.resteasy.resteasy-client-microprofile" optional="true"/>
```
bstansberry, line 25: Is optional="true" needed? If someone asks for this layer there's no reason to exclude the package.

bstansberry (changes requested):
I believe in microprofile-galleon-common we need a new jaxrs-server layer that is like the one in ee-feature-pack but also optionally depends on microprofile-rest-client.

bstansberry:
@ehsavoie When I build this and do various comparisons of dists with and without this changeset or between various kinds of dists (legacy vs 'modern', regular vs ee-9) I see some issues:

1) A comparison of ee-9/build with and without this shows that standalone-microprofile.xml and standalone-microprofile-ha.xml are missing the jaxrs extension and subsystem in a build of this branch.

2) A comparision of the 'build' output with and without this shows that the opentracing extension and subsystem are missing in domain.xml and a bunch of configs in docs/examples/configs.

3) I built with -DlegacyBuild. The configs in build-legacy vs those in ee-build differ in ways they should not. The opentracing extension and subsystem are still present in build-legacy. I'll send a PR to your repo against the branch that just has the opentracing commit.

bstansberry:
https://github.com/ehsavoie/wildfly/pull/8

Please feel free to squash that in.

I misspoke above. Without ^^^ the build fails when -DlegacyBuild is used.

ehsavoie:
@bstansberry I've fixed your issues.
1. The problem is that we are overwriting the jaxrs layer from ee with the one from micropfile so I had to 'merge' them in e99 to provide the full support.
2. and 3. were fixed in WFLY-14148 PR

ee-9/feature-pack/pom.xml
```diff
@@ -192,6 +192,9 @@
                             <resources>
                                 <resource>
                                     <directory>${mp.galleon.common.resources.directory}</directory>
+                                    <excludes>
+                                        <exclude>layers/standalone/jaxrs</exclude>
+                                    </excludes>
```
bstansberry, line 197: Is this needed? The next execution should copy over this fine with the one from the local src/main/resources.

microprofile/galleon-common/src/main/resources/layers/standalone/jaxrs/layer-spec.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="jaxrs">
+    <dependencies>
+        <layer name="microprofile-rest-client" optional="true"/>
+    </dependencies>
+</layer-spec>
```
bstansberry, line 6: @jfdenise My instinct is if we override a layer from a depended-upon FP, instead of letting Galleon combine the layer bits we are better off completely redeclaring the layer, as is done at https://github.com/wildfly/wildfly/pull/13774/commits/df965e99da9e05cbb8682eb223dfc6dace4cf0dc#diff-2fa73c0a47166ab82aea627f723525fbb1b3d1a112e07cfa018fc33bc08ddb23R7

There's some places where completely overriding is needed because the layer-spec.xml from the 'lower' FP doesn't actually get used. And it seems better to consistently do it the same way.

Doing that would also remove the need for https://github.com/wildfly/wildfly/pull/13774/commits/df965e99da9e05cbb8682eb223dfc6dace4cf0dc#diff-2fa73c0a47166ab82aea627f723525fbb1b3d1a112e07cfa018fc33bc08ddb23

It's not a big deal though; should not block merging this.

jmesnil:
@ehsavoie could you please rebase and resolve conflicts?

bstansberry (approved):

jmesnil:
@ehsavoie thanks!

PR #13777, opened 2020-12-03
[WFLY-14007] Certificate Revocation Lists

SoniaZaldana:
Jira: https://issues.redhat.com/browse/WFLY-14007

Requires: https://issues.redhat.com/browse/WFCORE-5170
WildFly Core PR with tests: https://github.com/wildfly/wildfly-core/pull/4412

### DIFF
--- a/docs/src/main/asciidoc/_elytron/Using_the_Elytron_Subsystem.adoc
+++ b/docs/src/main/asciidoc/_elytron/Using_the_Elytron_Subsystem.adoc
@@ -1454,6 +1454,13 @@ Two-Way HTTPS is now enabled for applications.
 
 You can configure your trust-manager to use certificate-revocation-list (CRL) to check revocation status of obtained certificates.
 
+The supported attributes for a certificate-revocation-list include:
+
+* path: The path to the configuration file that is used to initialize the certificate revocation list.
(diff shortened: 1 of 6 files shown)


docs/src/main/asciidoc/_elytron/Using_the_Elytron_Subsystem.adoc
```diff
@@ -1471,6 +1471,16 @@ This will use CRLs obtained from distribution points referenced in your certific
 /subsystem=elytron/trust-manager=twoWayTM:write-attribute(name=certificate-revocation-list.path, value=intermediate.crl.pem)
 ----
 
+===== Configure multiple certificate revocation lists
```
fjuma, line 1474: I think it would be good to mention which attributes are supported for `certificate-revocation-list` both here and above.

It would also be good to update the documentation (or add it if it doesn't already exist) on how to configure certificate revocation lists via the Elytron authentication client.

SoniaZaldana: I have added the additional documentation.

fjuma (approved):

PR #13790, opened 2020-12-07
WFLY-14172 Update Weld 4 for EE 9 WFLY. Also try to align Weld SE ver…

manovotn:
…sion used by tests.

JIRA issue - https://issues.redhat.com/browse/WFLY-14172

Note that apart from just bumping the version, I have noticed that WFLY uses different Weld SE version for some tests (and this isn't overriden in EE 9 setup).
I have tried aligning these version to see what happens because I don't quite see why they'd need to be different.


bstansberry:
https://ci.wildfly.org/viewLog.html?buildId=233053&tab=buildResultsDiv&buildTypeId=WF_Ee9LinuxJdk8 is a run of the -Dts.ee9 job with this.

bstansberry:
I re-ran all the jobs that timed out.

PR #13792, opened 2020-12-07
WFLY-14175 Update Jakarta Activation 1.2.2

boris-unckel:
Fixes [WFLY-14175](https://issues.redhat.com/browse/WFLY-14175)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -272,7 +272,7 @@
         <version.com.nimbus.jose-jwt>8.11</version.com.nimbus.jose-jwt>
         <version.com.squareup.okhttp3>3.9.0</version.com.squareup.okhttp3>
         <version.com.squareup.okio>1.13.0</version.com.squareup.okio>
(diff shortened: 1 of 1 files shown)


yersan (approved):
Hello @boris-unckel , thanks for the PR and the other contributions. We usually provide a link to make a comparison between relevant commits/tags so we can easily check what's going in the new release.

For this upgrade we can use this one: https://github.com/eclipse-ee4j/jaf/compare/e5623f7365f4d10b38a9d931e7f7fbd190924439...1.2.2

The 1.2.2 release is just a preparation work for the Jakarta Specification Process without any functional changes from its previous release. This upgrade should be harmless, so approving from my side. The only relevant change could be this fix https://github.com/eclipse-ee4j/jaf/commit/70e71e6aab37cb0a9621fe19b7ce03190eff4f03 which affects to OSGi  headers.

PR #13794, opened 2020-12-07
WFLY-14176 Update Apache commons CLI 1.4

boris-unckel:
Fixes [WFLY-14176](https://issues.redhat.com/browse/WFLY-14176)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -289,7 +289,7 @@
         <version.com.sun.istack>3.0.10</version.com.sun.istack>
         <version.com.sun.xml.fastinfoset>1.2.13</version.com.sun.xml.fastinfoset>
         <version.commons-beanutils>1.9.4</version.commons-beanutils>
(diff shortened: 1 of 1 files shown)


bstansberry:
@darranl I think this is ok.  It's used by VaultTool and JDR and seems like it should be fine for both.

There's also this in the legacy security subsystem module:

https://github.com/wildfly/wildfly/blob/master/servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/security/main/module.xml#L43

I can't see any reason why that is needed. The module doesn't import any classes from it nor does commons-cli provide any Service impls that might be pulled in via ServiceLoader. 

Looking at the commit that added that line I don't see any reason for its addition either.

https://github.com/wildfly/wildfly/commit/26bac3161a3fc5629edad8d373a798349dcf5d13

bstansberry:
retest this please

bstansberry:
Thanks, @boris-unckel!

PR #13804, opened 2020-12-08
WFLY-14132 jboss-ejb3.xml is ignored for exploded deployments

chengfang:
JIRA: https://issues.redhat.com/browse/WFLY-14132

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EjbJarParsingDeploymentUnitProcessor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EjbJarParsingDeploymentUnitProcessor.java
@@ -29,7 +29,6 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
-
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLResolver;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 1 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EjbJarParsingDeploymentUnitProcessor.java
```diff
@@ -191,14 +189,15 @@ public void undeploy(DeploymentUnit unit) {
...
     }
 
-    private static VirtualFile getDescriptor(final VirtualFile deploymentRoot, final String descriptorName) {
+    private static VirtualFile getDescriptor(final DeploymentUnit deploymentUnit, final VirtualFile deploymentRoot, final String descriptorName) {
+        final String deploymentUnitName = deploymentUnit.getName();
         // Locate the descriptor
         final VirtualFile descriptor;
         // EJB 3.1 FR 20.4 Enterprise Beans Packaged in a .war
-        if (isWar(deploymentRoot)) {
+        if (deploymentUnitName.endsWith(WAR_FILE_EXTENSION)) {
```
bstansberry, line 197: Is dropping the case sensitivity bit ok? I don't recall us doing anything to control what case is used for any runtime-name the user provides via the management API.

chengfang: I believe in the EE spec, these file extensions (.war, .jar, .ear) are case-sensitive. So if WildFly follows this rule and convention (some parts are outside the spec but WildFly still follows this convention), I think make the case sensitive will be cleaner. Not sure how common is case-insensitive runtime-name.

chengfang: I changed to compare the extensions case-insensitive, trying to be consistent with existing behavior.

bstansberry (approved):

PR #13806, opened 2020-12-10
[WFLY-14203] Move microprofile-config-smallrye to the microprofile feature-pack

jmesnil:
JIRA: https://issues.redhat.com/browse/WFLY-14203

This PR is built on top of #13756, #13765, #13766 & #13774. Only the last commit for WFLY-14203 should be reviewed.


ee-9/feature-pack/src/main/resources/feature_groups/domain-ha-profile.xml
```diff
@@ -12,5 +12,6 @@
 
     <feature-group name="basic-ha-profile"/>
     <feature-group name="microprofile-opentracing-jaeger"/>
+    <feature spec="subsystem.microprofile-config-smallrye"/>
```
jmesnil, line 15: I added them as it is required by JWT.

jmesnil:
I made a diff between dist/target/wildfly-22.0.0.Beta1-SNAPSHOT/standalone/configuration  and ee-9/dist/target/wildfly-preview-22.0.0.Beta1-SNAPSHOT/standalone/configuration and there were no differences related to microprofile stuff (differences were about security/elytron, jsr77 and messaging)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-jaxrs/main/module.xml
```diff
@@ -47,6 +47,6 @@
         <module name="org.jboss.resteasy.resteasy-validator-provider" optional="true" services="export" export="true"/>
         <module name="org.jboss.logging"/>
         <module name="org.reactivestreams"/>
-        <module name="org.eclipse.microprofile.config.api"/>
+        <module name="org.eclipse.microprofile.config.api" optional="true"/>
```
jmesnil, line 50: @ronsigal Could you please confirm that it is fine with resteasy if this module is not present?

jmesnil:
@ehsavoie I spotted 2 diffs related to opentracing:

```
 <extension module="org.wildfly.extension.microprofile.opentracing-smallrye"/>
       
            <subsystem xmlns="urn:wildfly:microprofile-opentracing-smallrye:3.0" default-tracer="jaeger">
                <jaeger-tracer name="jaeger">
                    <sampler-configuration sampler-type="const" sampler-param="1.0"/>
                </jaeger-tracer>
            </subsystem>
```

are present in `ee-9/dist/target/wildfly-preview-22.0.0.Beta1-SNAPSHOT/domain/configuration/domain.xml` while they are not in `dist/target/wildfly-22.0.0.Beta1-SNAPSHOT/domain/configuration/domain.xml`

Could you check that in #13766 PR please?

bstansberry:
@jmesnil I looked at the last commit and it seemed fine.

Comparing the output of build-legacy with and without this, it looked fine. Same with ee-9/build.

Re: the diff of 'build' though there's the same question as was discussed on the MP OT starting at https://github.com/wildfly/wildfly/pull/13766#discussion_r540134348. The MP Config extension/subsystem is removed from all the standard configs except standalone-microprofile[-ha].xml.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-jaxrs/main/module.xml
ronsigal: Hi @jmesnil,

Sorry for the late reply. I have believed that it would be fine, but I took another look at the code and it looks ok. I was going to run the WildFly testsuite to convince myself even further, but then I noticed that resteasy-jaxrs/main/module.xml in RESTEasy is

```
    <dependencies>
        ...
        <module name="org.eclipse.microprofile.config.api" export="true" optional="true"/>
    </dependencies>
```
so, there's the answer. It's fine.

jmesnil:
mmh, the mixed-domain failures seems genuine. I'm investigating them

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-jaxrs/main/module.xml
jmesnil: thanks!

jmesnil:
The mixed domain tests were genuine because I remove the code that was removing `microprofile-config-smallrye` from 7.1.0 Domain. However, I realised after that microprofile-config-smallrye remains in domain mode and must be removed when we adjust to 7.1.0. After all tests pass, I've squashed the separate commit

PR #13808, opened 2020-12-10
WFLY-14061 Replace Java EE references in java doc with corresponding …

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-14061

Description: Remove Java EE usage in java doc

As a part of the migration from JavaEE to Jakarta, we need to clean up JavaEE references in the sources (javadoc, documentation tags of xsd schemas, comments, etc.), replacing them with JakartaEE corresponding one when possible. Special care needed in avoiding any claim about JavaEE certification/compliance.

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
@@ -186,8 +186,8 @@ The timer is will be started once at the specified time.
 
 In case of a server restart the timeout method of a persistent timer
 will only be called directly if the specified time is elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will be not
-longer available if JBoss is restarted or the application is redeployed.
+If the timer is not persistent, it will no longer be available 
+if JBoss is restarted or the application is redeployed.
 
 [[recurring-timer]]
 === Recurring timer
@@ -201,9 +201,8 @@ execution.
 In case of server downtime for a persistent timer, the timeout method
 will be called only once if one, or more than one, interval is
 elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will not
-longer be active after the server is restarted or the application is
-redeployed.
+If the timer is not persistent, it will no longer be active 
+after the server is restarted or the application is redeployed.
 
 [[calendar-timer]]
 === Calendar timer
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2012, Red Hat, Inc., and individual contributors
+ * Copyright 2020, Red Hat, Inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
@@ -128,7 +128,7 @@ public interface WSLogger extends BasicLogger {
     @Message(id = 17, value = "Invalid handler chain file: %s")
     void invalidHandlerChainFile(String fileName);
 
-    String WS_SPEC_REF_5_3_2_4_2 = ". See section 5.3.2.4.2 of \"Web Services for Java EE, Version 1.4\".";
+    String WS_SPEC_REF_5_3_2_4_2 = ". See section 5.3.2.4.2 of \"Jakarta Enterprise Web Services 2.0\".";
 
     @LogMessage(level = ERROR)
     @Message(id = 18, value = "Web service method %s must not be static or final" + WS_SPEC_REF_5_3_2_4_2)
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/JSF.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/JSF.adoc
@@ -1,26 +1,26 @@
-[[JSF]]
-= JSF Configuration
+[[Jakarta_Server_Faces]]
+= Jakarta Server Faces Configuration
 
 [abstract]
 
-JSF configuration is handled by the JSF subsystem. The JSF subsystem
-allows multiple JSF implementations to be installed on the same WildFly
+Jakarta Server Faces configuration is handled by the Jakarta Server Faces subsystem. The Jakarta Server Faces subsystem
+allows multiple Jakarta Server Faces implementations to be installed on the same WildFly
 server. In particular, any version of Mojarra or MyFaces that implements
-spec level 2.1 or higher can be installed. For each JSF implementation,
+spec level 2.1 or higher can be installed. For each Jakarta Server Faces implementation,
 a new slot needs to be created under `com.sun.jsf-impl`,
-`javax.faces.api`, and `org.jboss.as.jsf-injection`. When the JSF
-subsystem starts up, it scans the module path to find all of the JSF
-implementations that have been installed. The default JSF implementation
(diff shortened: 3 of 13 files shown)


scottmarlow:
@RanabirChakraborty I see that some [acronyms](https://jakarta.ee/legal/acronym_guidelines/) are not yet changed.  Would you like feedback on which ones also need to change?

RanabirChakraborty:
I have changed a few acronyms but still have a big chunk left to do. It's all over the places and hard to spot because many have the older specifications, which I guess I can't change (for example, somewhere It's still using the old EJB version `EJB 3.0`). It'll probably take a few more days to change all the acronyms properly.

bstansberry:
@RanabirChakraborty @scottmarlow @asoldano 

I think this work should be partitioned into chunks in some way, probably by spec/subsystem.

We need to be able to get PRs up, get them reviewed, get them merged, and avoid having individual ones touch wide parts of the codebase. This will allow us to avoid conflicts between these changes and other PRs people are doing. It also makes it more likely that component leads will help with the review

If this PR is correct I suggest we get it approved and merged.

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
```diff
@@ -186,7 +186,7 @@ The timer is will be started once at the specified time.
 
 In case of a server restart the timeout method of a persistent timer
 will only be called directly if the specified time is elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will be not
+If the timer is not persistent (since Jakarta Enterprise Beans 3.2 see 18.2.3) it will be not
```
bstansberry, line 189: "since ... 3.2" is not correct.

I think in general if we encountered problems related to the use of "since XXX" we should drop that bit of text, unless it adds a lot of benefit. We put out new releases of docs with each new WF release, so this kind of "since XXX" info generally isn't that useful. The reader isn't having to guess whether the text is relevant to their release.

The "see 18.2.3" part is problematic too. Here's the Jakarta Enterprise Beans 3.2 spec:

https://jakarta.ee/specifications/enterprise-beans/3.2/enterprise-beans-spec-3.2.html

There is no 18.2.3. There's not much at all.

Basically Jakarta EE 8 relies on the Java EE 8 spec docs for all the details. There's no real way to be correct about this kind of reference without referring to EJB.

I don't see a lot of value in the whole (since EJB3.1 see 18.2.3).

Also, since this line is being edited:

s/it will be not longer/it will no longer be/g

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
```diff
@@ -201,7 +201,7 @@ execution.
 In case of server downtime for a persistent timer, the timeout method
 will be called only once if one, or more than one, interval is
 elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will not
+If the timer is not persistent (since Jakarta Enterprise Beans 3.2 see 18.2.3) it will not
```
bstansberry, line 204: Same discussion as above, and

s/it will be not/it will no/g

webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
```diff
@@ -128,7 +128,7 @@
     @Message(id = 17, value = "Invalid handler chain file: %s")
     void invalidHandlerChainFile(String fileName);
 
-    String WS_SPEC_REF_5_3_2_4_2 = ". See section 5.3.2.4.2 of \"Web Services for Java EE, Version 1.4\".";
+    String WS_SPEC_REF_5_3_2_4_2 = ". See section 5.3.2.4.2 of \"Jakarta Enterprise Web Services 1.4\".";
```
bstansberry, line 131: @asoldano Technically, there is no such thing:

https://jakarta.ee/specifications/enterprise-ws/1.4/enterprise-ws-spec-1.4.html

There's Section 1 and that's it.

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
scottmarlow: The Jakarta Enterprise Beans 3.2 spec contents or lack of, is sadly caused by undetected errors translating the asciidoc => html/pdf files.  

IMO +1 for removing the whole `(since Jakarta Enterprise Beans 3.2 see 18.2.3)`.

webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
scottmarlow: It looks like the invalid asciidoc issues were fixed for EE 9 via https://jakarta.ee/specifications/enterprise-ws/2.0/enterprise-ws-spec-2.0.html.

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
scottmarlow: I'm asking if it is possible to fix the broken EE 8 SPEC doc pages.

scottmarlow: > I'm asking if it is possible to fix the broken EE 8 SPEC doc pages.

https://www.eclipse.org/lists/jakartaee-platform-dev/msg02286.html mentions that a full SPEC ballot process is needed to update the EE 8 SPEC doc.  

https://www.eclipse.org/mhonarc/lists/jakartaee-platform-dev/msg02287.html mentions that the only published spec doc for EE 8 was the Platform SPEC doc.

My question now is how important (to WildFly) is it that there be Jakarta EE 8 SPEC API documentation?

RanabirChakraborty:
@bstansberry Thank you for the review comments. I have changed them accordingly.

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
bstansberry: @scottmarlow IMHO it's not critical.

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
```diff
@@ -186,8 +186,8 @@ The timer is will be started once at the specified time.
 
 In case of a server restart the timeout method of a persistent timer
 will only be called directly if the specified time is elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will be not
-longer available if JBoss is restarted or the application is redeployed.
+If the timer is not persistent, It will no longer be available
```
bstansberry, line 189: s/It/it

docs/src/main/asciidoc/_developer-guide/EJB3_Reference_Guide.adoc
```diff
@@ -201,9 +201,8 @@ execution.
 In case of server downtime for a persistent timer, the timeout method
 will be called only once if one, or more than one, interval is
 elapsed. +
-If the timer is not persistent (since EJB3.1 see 18.2.3) it will not
-longer be active after the server is restarted or the application is
-redeployed.
+If the timer is not persistent, It will no longer be active
```
bstansberry, line 204: s/It/it

bstansberry (changes requested):
My previous substantive comments have been addressed. Thanks!

I have a couple tiny capitalization change, then LGTM.

RanabirChakraborty:
@bstansberry Change is done. I guess it's good to go now.

bstansberry (approved):

PR #13810, opened 2020-12-10
WFLY-14110 Distributed session manager reports HttpSession.getLastAccessedTime() incorrectly

pferraro:
https://issues.redhat.com/browse/WFLY-14110

### DIFF
--- a/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
+++ b/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
@@ -28,11 +28,31 @@ import java.time.Duration;
  * The more static aspects of a session's meta-data.
  * @author Paul Ferraro
  */
-public interface SessionCreationMetaData extends ImmutableSessionCreationMetaData {
-
+public interface SessionCreationMetaData extends ImmutableSessionCreationMetaData, AutoCloseable {
+    /**
+     * Sets the maximum duration of time this session may remain idle before it will be expired by the session manager.
+     * @param a maximum duration of time this session may remain idle before it will be expired by the session manager.
+     */
     void setMaxInactiveInterval(Duration duration);
 
+    /**
+     * Indicates whether or not this session has been invalidated.
+     * @return true, if this session was invalidated, false otherwise.
+     */
     boolean isValid();
 
+    /**
+     * Invalidates this session.
+     * @return true, if this session was previous valid, false otherwise
+     */
     boolean invalidate();
+
+    /**
+     * Indicates whether or not this session was newly created.
+     * @return true, if this session was newly created, false otherwise.
+     */
+    boolean isNew();
+
+    @Override
+    void close();
 }
--- a/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
+++ b/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
@@ -53,6 +53,9 @@ public interface SessionCreationMetaData extends ImmutableSessionCreationMetaDat
      */
     boolean isNew();
 
+    /**
+     * Signals the end of the transient lifecycle of this session, typically triggered at the end of a given request.
+     */
     @Override
     void close();
 }
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
@@ -29,6 +29,7 @@ import io.undertow.servlet.handlers.security.CachedAuthenticatedSessionHandler;
 import io.undertow.websockets.core.WebSocketChannel;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.function.Consumer;
 import java.util.Arrays;
@@ -59,6 +60,7 @@ public class DistributableSession implements io.undertow.server.session.Session 
     private final UndertowSessionManager manager;
     private final Batch batch;
     private final Consumer<HttpServerExchange> closeTask;
+    private final Instant startTime;
 
     private volatile Map.Entry<Session<LocalSessionContext>, SessionConfig> entry;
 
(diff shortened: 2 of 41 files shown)


clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
```diff
@@ -28,11 +28,31 @@
...
     boolean isValid();
 
+    /**
+     * Invalidates this session.
+     * @return true, if this session was previous valid, false otherwise
+     */
     boolean invalidate();
+
+    /**
+     * Indicates whether or not this session was newly created.
```
bstansberry, line 51: What does 'newly created' mean? First request for the session id anywhere in the cluster? Locally?

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
```diff
@@ -28,11 +28,31 @@
...
+     * @return true, if this session was newly created, false otherwise.
+     */
+    boolean isNew();
+
+    @Override
+    void close();
```
bstansberry, line 57: What is the expected use of this? From looking at the PR it seems it's meant to be called at some point and that changes what isNew() would return. But 'close()' as a method name doesn't say that, which is ok so long as the expected use of close() will have that effect. But the expected use is ???.

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
```diff
@@ -87,6 +90,11 @@ public void requestDone(HttpServerExchange exchange) {
                     }
                     // If batch is closed, close session in a new batch
                     try (Batch batch = (this.batch.getState() == Batch.State.CLOSED) ? batcher.createBatch() : this.batch) {
+                        if (this.startTime != null) {
+                            // According to §7.6 of the servlet specification:
+                            // The session is considered to be accessed when a request that is part of the session is first handled by the servlet container.
+                            session.getMetaData().setLastAccessedTime(this.startTime);
```
bstansberry, line 96: Is a new one of these DistributableSession objects created per request?

Must be or startTime would not change and the Batch object passed to the c'tor doesn't sound like something that would span requests.

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
pferraro: This is ultimately triggered by org.wildfly.clustering.web.session.Session#close which effectively ends the transient lifecycle of the session (typically spanning a single request).

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
pferraro: > Is a new one of these DistributableSession objects created per request?

Yes.

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
pferraro: As defined by: https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpSession.html#isNew--

pferraro: This property was formerly implied by the fact that the creationTime != the lastAccessedTime.  However, since the semantics of lastAccessedTime have changed, this must now be determined explicitly, hence this logic.

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
bstansberry: @pferraro Thanks. I meant to ask in my previous comment whether the javadoc should say something about this, for the benefit of future maintainers.

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
bstansberry: Thanks.

clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SessionCreationMetaData.java
pferraro: More javadoc cannot hurt.

bstansberry: Thanks. That helps a lot. I think of a session as a long lived thing. Object implementing these interfaces might or might not be, so my brain had trouble grokking what the semantics of close would be.

bstansberry (approved):

bstansberry:
I'm going to kill the CI jobs that just started from the javadoc change and toss this in a combo PR with a few other PRs.

bstansberry:
The Linux Elytron job's FineNonTransactionalSessionActivationTestCase failure has showed up quite a lot in association with this PR:

https://ci.wildfly.org/project.html?projectId=WF_PullRequest&buildTypeId=&tab=testDetails&testNameId=-4167790151802329749&order=TEST_STATUS_DESC&branch_WF_PullRequest=__all_branches__&itemsCount=50

There are failures for this PR or for #13819, a combo PR that initially included this. The failures led me to exclude this from #13819 which then passed CI.

bstansberry (changes requested):
@pferraro I'm going to assume the test failures I noted previously are somehow related.

pferraro:
I found the source of the problem, though it is unclear to me why this change makes the problem more prominent.

pferraro:
@bstansberry I've push a fix.  Will validate using several CI runs.

pferraro:
So far so good.  Windows failures are unrelated. /retest

bstansberry (approved):
@pferraro When you are satisfied with CI please go ahead and merge.

pferraro:
Damn it. https://ci.wildfly.org/viewLog.html?buildId=234736& triggered the same failure.  Adding hold label.

PR #13812, opened 2020-12-11
[WFLY-14210] Bump the jaxrs subsystems model version. Reject attribut…

jamezp:
…es against the 1.0.0 and 2.0.0 model versions. Model version 2.0.0 was incorrectly used in WildFly 19, 20 and 21. An added system property was added to allow the transformers to be bypassed.

https://issues.redhat.com/browse/WFLY-14210

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_1_0.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_1_0.java
@@ -25,37 +25,25 @@ import static org.jboss.as.controller.parsing.ParseUtils.requireNoAttributes;
 import static org.jboss.as.controller.parsing.ParseUtils.requireNoContent;
 
 import java.util.List;
-
 import javax.xml.stream.XMLStreamConstants;
 import javax.xml.stream.XMLStreamException;
 
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.operations.common.Util;
-import org.jboss.as.controller.persistence.SubsystemMarshallingContext;
 import org.jboss.dmr.ModelNode;
 import org.jboss.staxmapper.XMLElementReader;
-import org.jboss.staxmapper.XMLElementWriter;
 import org.jboss.staxmapper.XMLExtendedStreamReader;
-import org.jboss.staxmapper.XMLExtendedStreamWriter;
 
-public class JaxrsSubsystemParser_1_0 implements XMLStreamConstants, XMLElementReader<List<ModelNode>>, XMLElementWriter<SubsystemMarshallingContext> {
+public class JaxrsSubsystemParser_1_0 implements XMLStreamConstants, XMLElementReader<List<ModelNode>> {
 
-      /**
-       * {@inheritDoc}
-       */
-      @Override
-      public void readElement(final XMLExtendedStreamReader reader, final List<ModelNode> list) throws XMLStreamException {
-          // Require no attributes or content
-          requireNoAttributes(reader);
-          requireNoContent(reader);
-          list.add(Util.createAddOperation(PathAddress.pathAddress(JaxrsExtension.SUBSYSTEM_PATH)));
-      }
-
-      /**
-       * {@inheritDoc}
-       */
-      @Override
-      public void writeContent(final XMLExtendedStreamWriter streamWriter, final SubsystemMarshallingContext context) throws XMLStreamException {
-          context.startSubsystemElement(JaxrsExtension.NAMESPACE_1_0, true);
-      }
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void readElement(final XMLExtendedStreamReader reader, final List<ModelNode> list) throws XMLStreamException {
+        // Require no attributes or content
+        requireNoAttributes(reader);
+        requireNoContent(reader);
+        list.add(Util.createAddOperation(PathAddress.pathAddress(JaxrsExtension.SUBSYSTEM_PATH)));
+    }
 }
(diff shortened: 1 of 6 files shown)


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_1_0.java
```diff
@@ -56,6 +58,6 @@ public void readElement(final XMLExtendedStreamReader reader, final List<ModelNo
        */
       @Override
       public void writeContent(final XMLExtendedStreamWriter streamWriter, final SubsystemMarshallingContext context) throws XMLStreamException {
-          context.startSubsystemElement(JaxrsExtension.NAMESPACE_1_0, true);
+          context.startSubsystemElement(NAMESPACE, true);
```
bstansberry, line 61: Since you went down the refactoring path I'm going to punish you. ;)

j/k -- cleanup is good.

This class should no longer implement XMLElementWriter as it is never used as such.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsTransformers.java
```diff
@@ -45,12 +56,22 @@ public String getSubsystemName() {
...
+                    TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(subsystemRegistration.getCurrentSubsystemVersion());
+
+            //Differences between 3.0.0 and 2.0.0
+            ResourceTransformationDescriptionBuilder builder200 =
+                    chainedBuilder.createBuilder(subsystemRegistration.getCurrentSubsystemVersion(), MODEL_VERSION_2_0_0);
+            registerTransformers_2_0_0(builder200);
+
+            //Create chained transformers.
+            chainedBuilder.buildAndRegister(subsystemRegistration, new ModelVersion[] {MODEL_VERSION_2_0_0, MODEL_VERSION_1_0_0});
+        }
```
bstansberry, line 70: Something feels off here.

AIUI we've actually had a "3.0" (imagine some air quotes) since WF 19, and it looks like we've had a proper 2.0 since some time in 2017. And now you're officially creating a proper 3.0. 

We need transformation for hosts using 1.0 no matter what. We need the missing 3.0 -> 2.0 transformation unless the admin disabled it. If the admin disabled 3.0 -> 2.0 with the system property, then we must do 3.0 -> 1.0.

Maybe rename registerTransformers_2_0_0 to registerTransformersFrom_3_0_0 and then make sure it's invoked no matter what. What differs is the target version provided to chainedBuilder.createBuilder.

BTW I think names like registerTransformersFrom_3_0_0 are better in general. When I review these classes I usually waste time trying to figure out what these version # in these methods represents -- source or target.

bstansberry (approved):

bstansberry (dismissed):
Last time I hit "Approve" when I meant "Request changes".

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemParser_1_0.java
jamezp: Ah good call. I'll fix that as well. It's a quick change.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsTransformers.java
jamezp: Okay that makes sense to me. Something felt off to me as well and for some reason I was just not seeing it :) Now that you point it out it's fairly obvious though.

jamezp: Updated. I _think_ I've got it right. For some reason I confuse myself with transformers all the time :)

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsTransformers.java
```diff
@@ -45,12 +56,29 @@ public String getSubsystemName() {
...
+        ChainedTransformationDescriptionBuilder chainedBuilder =
+                TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(subsystemRegistration.getCurrentSubsystemVersion());
+
+        // Differences between 3.0.0 and 2.0.0
+        ResourceTransformationDescriptionBuilder builder300 =
+                chainedBuilder.createBuilder(subsystemRegistration.getCurrentSubsystemVersion(), target);
+        registerTransformers_3_0_0(builder300);
+
+        // Create chained transformers.
+        chainedBuilder.buildAndRegister(subsystemRegistration, modelVersions);
```
bstansberry, line 77: @kabir All of the above LGTM but can you confirm?

kabir: @bstansberry @jamezp Although I don't totally get the isBypassTransformers stuff I think it looks fine. 
It seems we only care about transforming one version delta (CURRENT to 1.0.0; or CURRENT to 2.0.0) and we account for that in the builder.
Then the registration process seems to take that into account. If tests pass both with and without org.wildfly.jaxrs.unsupported.bypass.2.0.0.transformer it should be ok

jamezp: @kabir The test actually fails if the property is set, but that should be expected. The reason for the property is to allow WildFly 19, 20 and 21 servers to **not** reject those attributes in a mixed domain. I doubt it will be used much TBH, but it seemed like we should have a way to bypass it just in case.

kabir (approved):

PR #13817, opened 2020-12-11
[WFLY-14194] Driver processor: only remove drivers explicitly registe…

tadamski:
…red by a module

https://issues.redhat.com/browse/WFLY-14194

With this fix:
a. JSTL TCK tests pass
b. drivers from driver modules are being correctly unregistered

so this let's the TCKs pass but TBH I'm not sure if now we are preventing all leaks: f.e. wouldn't TCK deployment cause a leak? 
OTOH even if it would and it is correct I think we cannot do much about it

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/deployers/ds/processors/DriverProcessor.java
+++ b/connector/src/main/java/org/jboss/as/connector/deployers/ds/processors/DriverProcessor.java
@@ -115,25 +115,33 @@ public final class DriverProcessor implements DeploymentUnitProcessor {
     /** {@inheritDoc} */
     @Override
     public void undeploy(final DeploymentUnit context) {
-        try {
-            /**
-             * https://issues.redhat.com/browse/WFLY-14114
-             *
-             * This hack allows to deregister all drivers registered by this module. See comments in {@link DriverManagerAdapterProcessor}
-             */
-            final Module module = context.getAttachment(org.jboss.as.server.deployment.Attachments.MODULE);
-            Class<?> driverManagerAdapterClass = module.getClassLoader().loadClass(DriverManagerAdapter.class.getName());
+        /**
+         * https://issues.redhat.com/browse/WFLY-14114
+         *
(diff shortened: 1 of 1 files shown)


scottmarlow:
In case it helps, [org.jboss.as.jpa.processor.PersistenceUnitServiceHandler](https://github.com/wildfly/wildfly/blob/master/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java#L143) is an example of attaching information (PersistenceAdaptorRemoval) to the DeploymentUnit during deployment that can later be referenced during undeployment for performing cleanup.

scottmarlow:
> so this let's the TCKs pass but TBH I'm not sure if now we are preventing all leaks: f.e. wouldn't TCK deployment cause a leak?
> OTOH even if it would and it is correct I think we cannot do much about it

Are you saying that any application that directly references a DataSource, the JDBC driver may be registered with the DriverManager and will not be unregistered from the DriverManager? 

If yes, I wonder if we should have a JCA management operation that helps:

1. List all registered JDBC drivers
2. Unregister a particular JDBC driver

bstansberry:
Good discussion, although let's not let it block getting this merged if what is here is valid (LGTM), if what we're talking about is an RFE, or even if it's a bug if it's a minor one.

What's here plus the earlier change this amends seems like a solid piece of work on its own. The drivers the DUP instantiates in deploy() are cleaned up in undeploy().

If we have other cases of server provided services instantiating drivers from deployments and not cleaning them up, that's good to look into.

Application code that's directly instantiating drivers and not cleaning them up is an application that has its own CL leak. Being helpful about that sounds more like an RFE. It seems like such an app would be better off doing their own cleanup, say in a @PreDestroy method vs having their poor admins be responsible for invoking a management op.

bstansberry (approved):

bstansberry:
Thanks, @tadamski !

gaol:
Thanks a lot for the quick fix :+1:  @tadamski

tadamski:
@scottmarlow That seems a good idea. I will create a JIRA for that.

PR #13818, opened 2020-12-11
WFLY-14218 Non-HA server configuration variants use wrong distributable-web feature group

pferraro:
https://issues.redhat.com/browse/WFLY-14218

### DIFF
--- a/clustering/infinispan/extension/src/main/resources/subsystem-templates/infinispan.xml
+++ b/clustering/infinispan/extension/src/main/resources/subsystem-templates/infinispan.xml
@@ -12,7 +12,6 @@
                     <file-store passivation="true" purge="false"/>
                 </local-cache>
                 <local-cache name="sso"/>
-                <local-cache name="routing"/>
             </cache-container>
             <cache-container name="ejb" aliases="sfsb" default-cache="passivation" module="org.wildfly.clustering.ejb.infinispan">
                 <local-cache name="passivation">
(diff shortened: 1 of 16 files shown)


pferraro:
@bstansberry Please review.

ee-feature-pack/galleon-common/src/main/resources/feature_groups/distributable-web-local.xml
```diff
@@ -15,10 +15,7 @@
             <param name="cache-container" value="web"/>
             <param name="cache" value="sso"/>
         </feature>
-        <feature spec="subsystem.distributable-web.routing.local">
-            <param name="cache-container" value="web"/>
-            <param name="cache" value="routing"/>
-        </feature>
+        <feature spec="subsystem.distributable-web.routing.local"/>
```
bstansberry, line 18: This comment is just a note for the record in case someone someday investigates this PR, or someone else reviews it now.

The two attributes that are removed above are not valid for the resource being configured. This had gone undiscovered because this feature group was never used prior to Paul's work on this issue.

bstansberry (changes requested):
@pferraro The change itself seems fine but I believe the same change is applicable to all the non-HA standard configs we build that provide distributable-web, whether from the 'wildfly', 'wildfly-preview', or 'wildfly-ee' Galleon feature packs or from the legacy ee-feature-pack/legacy-feature-pack. I don't think there's anything about the purpose of the microprofile configs that should result in config drift vs the others.

pferraro:
Ah, ok. I only noticed this after debugging bootable jar tests that use the mp configs. I will check the others as well.

pferraro:
@bstansberry I've updated this PR to fix all non-HA configuration variants, including the subsystem template for the legacy build

pferraro:
@bstansberry This is all set now.

PR #13843, opened 2021-01-04
[WFLY-14256] Correct what testsuite/layers provisioning executions ru…

bstansberry:
…n when -Dts.ee9 is used

https://issues.redhat.com/browse/WFLY-14256

### DIFF
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -661,7 +661,7 @@
                                 <goals>
                                     <goal>provision</goal>
                                 </goals>
-                                <phase>${provisioning.phase.preview.excluded}</phase>
+                                <phase>${provisioning.phase}</phase>
                                 <configuration>
                                     <install-dir>${layers.install.dir}/web-passivation</install-dir>
                                     <feature-packs>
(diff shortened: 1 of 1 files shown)


yersan (commented):
There are still `galleon-maven-plugin` executions in `testsuite/layers` which are using the `${servlet.feature.pack.artifactId}` to be run on `${provisioning.phase}.`

Based on the Jira description, these executions should use `${provisioning.phase.preview.excluded} `phase instead, because the servlet fp is an off-topic in the context of ee9.profile.

@bstansberry should not all these execution be changed accordingly on this PR?

bstansberry:
@yersan No, most executions that use ${servlet.feature.pack.artifactId} can run. The ts.ee9 profile changes servlet.feature.pack.artifactId to 'wildfly-preview-feature-pack'.  The ones that should use ${provisioning.phase.preview.excluded} are the ones where using 'wildfly-preview-feature-pack' won't work or is somehow meaningless.

So. 'test-standalone-reference-provisioning' is meaningless because it does not test a layer, it tests the full standalone.xml config and there is another execution that does the same thing, just not with ${servlet.feature.pack.artifactId}. So, running it with -Dts.ee9 is redundant, i.e. meaningless, so exclude it.

The 'vault-provisioning' execution will not work with 'wildfly-preview-feature-pack' as we don't provide it. So, disabled.

OTOH 'ee-provisioning' uses ${servlet.feature.pack.artifactId} but will run fine with 'wildfly-preview-feature-pack'.  That's the only execution that tests that layer by itself so we want it to run.

yersan (approved):
Thanks you @bstansberry for the detailed explanation, now I see the intention and how it works

PR #13845, opened 2021-01-04
[WFLY-14230] - Integrate MP OpenTracing 2.0 into EAP

jasondlee:
These changes upgrade OpenTracing in Wildfly to 2.0. Most tests (TCK and otherwise) are passing locally. One still failing due, we think, to a change in the response. Will finish discussion ASAP.


jasondlee:
Recreated clean PR

PR #13848, opened 2021-01-05
[WFLY-14271] timed object id throws null when tried to be restored

elguardian:
Jira: https://issues.redhat.com/browse/WFLY-14271
Upon activation of the timer service the creation of the hash set related to a timer object id can be null if there is a problem with database connection. This causes the map not to be filled and therfore any further registration of the timers fail to NullPointerException.
Added a new timerDeployed event in the persistence to signal before the load so always this is filled.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -326,6 +326,13 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service<Datab
 
     @Override
     public void addTimer(final TimerImpl timerEntity) {
+        String timedObjectId = timerEntity.getTimedObjectId();
+        synchronized (this) {
+            if(!knownTimerIds.containsKey(timedObjectId)) {
+                throw EjbLogger.EJB3_TIMER_LOGGER.timerCannotBeAdded(timerEntity);
+            }
+        }
+
         String createTimer = sql(CREATE_TIMER);
         Connection connection = null;
         PreparedStatement statement = null;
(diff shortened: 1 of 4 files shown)


elguardian:
this is ok to test

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -326,6 +326,13 @@ private String sql(final String key) {
 
     @Override
     public void addTimer(final TimerImpl timerEntity) {
+        String timedObjectId = timerEntity.getTimedObjectId();
+        synchronized (this) {
+            if(!knownTimerIds.containsKey(timedObjectId)) {
+                EjbLogger.EJB3_TIMER_LOGGER.timerCannotBeAdded(timerEntity);
```
bstansberry, line 332: you need to add the 'throw' keyword.

bstansberry: BTW I had a quick look because the PR title made me curious, and noted the ^^^, but please don't wait on me for approval on this; the EJB experts should do the code review.

elguardian: thx @bstansberry for the feedback. Sry I was testing CI in this case for not breaking anything before asking for Cheng

elguardian:
@chengfang do you mind to have a look. We have big customer on this.

@bstansberry cannot add reviewers

elguardian:
Jenkins retest this please

elguardian:
This is ok to test

bstansberry:
The github request review function is not very useful as there are far more people who should do reviews vs ones who can hit the merge button. Using @ pings works better

@tadamski may be able to help as well

chengfang (approved):

PR #13850, opened 2021-01-05
[WFLY-14261] Integrate MP Health 3.0 into EAP

xstefank:
https://issues.redhat.com/browse/WFLY-14261
https://github.com/wildfly/wildfly-proposals/pull/357

PR #13854, opened 2021-01-05
Fix the mp-4.0 GH action to build the PR branch

xstefank:
I noticed in https://github.com/wildfly/wildfly/pull/13850 that the PR itself wasn't built as I updated the MP Health to 3.0 which has now 21 TCK tests (and run had only 16 from Health 2.2) so after some checking the current approach with `pull_request_target` (https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) is probably not intended - "This event runs in the context of the base of the pull request, rather than in the merge commit as the pull_request event does." meaning this workflow now checkouts directly MP4 branch and not the PR branch. This change will make it check out the PR branch merged on top of MP4.

### DIFF
--- a/.github/workflows/mp-4.0.yml
+++ b/.github/workflows/mp-4.0.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - 'MP4'
-  pull_request_target:
+  pull_request:
     branches:
       - 'MP4'
 


xstefank:
@kabir can you take a look?

kabir:
I was using `pull_request_target` since that seems safer. I was using it elsewhere as secrets are not propagated for cross-repository PRs and I want access to those. But as this job isn't actually using the secrets we don't need to worry, so using `pull_request` should be ok.

PR #13857, opened 2021-01-06
[WFLY-14266] Resource-adapters: add report-directory config

tadamski:

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
@@ -44,6 +44,7 @@ import org.jboss.as.controller.operations.validation.StringLengthValidator;
 import org.jboss.as.controller.security.CredentialReference;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.jboss.dmr.ValueExpression;
 import org.jboss.jca.common.api.metadata.Defaults;
 import org.jboss.jca.common.api.metadata.common.Recovery;
 import org.jboss.jca.common.api.metadata.common.TransactionSupportEnum;
@@ -207,6 +208,8 @@ public class Constants {
     public static final String TEST_CONNECTION_IN_POOL = "test-connection-in-pool";
 
     public static final String CLEAR_STATISTICS = "clear-statistics";
+
+    public static final String REPORT_DIRECTORY_NAME = "report-directory";
 
 
     static final SimpleAttributeDefinition CLASS_NAME = new SimpleAttributeDefinitionBuilder(CLASS_NAME_NAME, ModelType.STRING, false)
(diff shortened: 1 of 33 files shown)


darranl:
@tadamski This one has been flagged as a feature, has this one kicked off in the RFE process yet?

tadamski:
@darranl Not yet. I have it on my radar.

darranl:
Just added "rebase-this" as there are conflicts but I suspect we have higher priorities for now.

tadamski:
@darranl  I have rebased this PR and started the RFE process in https://issues.redhat.com/browse/EAP7-1867

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
```diff
@@ -659,6 +662,14 @@ public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceM
             .setRestartAllServices()
             .build();
 
+    static SimpleAttributeDefinition REPORT_DIRECTORY = new SimpleAttributeDefinitionBuilder(REPORT_DIRECTORY_NAME, ModelType.STRING)
+            .setXmlName("path")
+            .setDefaultValue(new ModelNode(new ValueExpression("${jboss.server.log.dir}")))
+            .setRequired(false)
+            .setAllowExpression(true)
+            .setRestartAllServices()
+            .build();
```
bstansberry, line 671: Please add this to the builder:

.addArbitraryDescriptor(ModelDescriptionConstants.FILESYSTEM_PATH, ModelNode.TRUE)

This provides a hint to tools like the CLI that they can use the local FS to do tab completion suggestions.

bstansberry: @jmesnil or anyone... Plus @tadamski of course

This is non-standard in how it is handling paths; i.e. it violates the 2 attribute, path + relative-to convention.

OTOH I really like the default value idea.

The downside of not having relative-to is the path can't be relative to user-provided Path resources, as those don't have associated system properties. But if that is a problem we can compatibly add another report-directory-relative-to attribute later.

Note that with a 2 attribute setup, the default value Tomek added here is still valid; the 'relative-to' would have no default.

Perhaps this should be named report-directory-path to better align with a future report-directory-relative-to but I don't really think that is necessary.

@jamezp made a comment re possible effect on wildscribe that i'll look into a bit.

bstansberry: Wildscribe renders the default value as '${jboss.server.log.dir}' which isn't as correct as the 'expression "${jboss.server.log.dir}"' in the read-resource-description output but seems ok.

bstansberry: @tadamski Please add this to the builder:

.addArbitraryDescriptor(ModelDescriptionConstants.FILESYSTEM_PATH, ModelNode.TRUE)

connector/src/main/resources/schema/wildfly-resource-adapters_6_0.xsd
```diff
@@ -31,9 +31,10 @@
   <xs:element name="subsystem" type="subsystemType"/>
 
   <xs:complexType name="subsystemType">
-    <xs:all>
+    <xs:sequence>
```
bstansberry, line 34: The 6.0 xsd has been published so shouldn't be changed.

bstansberry (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
tadamski: added

bstansberry: Thanks. I resolved this conversation but I'm going to 'unresolve' it, not because it's not done, but just because I think the discussion comments are useful so no need to make the github ui hide them.

bstansberry (approved):

bstansberry:
Superseded by https://github.com/wildfly/wildfly/pull/15362

PR #13866, opened 2021-01-11
[WFLY-14238] FaceletsResourceResolver annotation is not taken into account

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-14238

The fix just adds the missing annotation (`javax.faces.view.facelets.FaceletsResourceResolver`) included in JSF 2.2+ to wildfly. Previously it was not there and the Jandex Annotation Provider does not return classes with this annotation inside a JSF app. The class is loaded programmatically (`loadClass`) because if not done this way there is a `NoClassDefFoundError` in the Jandex provider and the implementation defaults to the default JSF one. Test included.

@fjuma take a look when you have time.

### DIFF
--- a/jsf/injection/src/main/java/org/jboss/as/jsf/injection/AnnotationMap.java
+++ b/jsf/injection/src/main/java/org/jboss/as/jsf/injection/AnnotationMap.java
@@ -70,7 +70,10 @@ public class AnnotationMap {
             stringToAnnoMap.put(FacesBehavior.class.getName(), FacesBehavior.class);
             stringToAnnoMap.put(FacesBehaviorRenderer.class.getName(), FacesBehaviorRenderer.class);
 
-            // Put JSF 2.2 annotations below this line if any new ones are to be scanned.  So far none.
+            // Put JSF 2.2+ annotations below this line if any new ones are to be scanned.
+            // Load the class to avoid a NoClassDefFoundError if it is not present in the impl
+            ClassLoader loader = AnnotationMap.class.getClassLoader();
+            addAnnotationIfPresent(loader, "javax.faces.view.facelets.FaceletsResourceResolver");
         } catch (Exception e) {
             // Ignore.  Whatever classes are available have been loaded into the map.
         }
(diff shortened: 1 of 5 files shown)


jsf/injection/src/main/java/org/jboss/as/jsf/injection/AnnotationMap.java
```diff
@@ -70,7 +70,10 @@
             stringToAnnoMap.put(FacesBehavior.class.getName(), FacesBehavior.class);
             stringToAnnoMap.put(FacesBehaviorRenderer.class.getName(), FacesBehaviorRenderer.class);
 
-            // Put JSF 2.2 annotations below this line if any new ones are to be scanned.  So far none.
+            // Put JSF 2.2+ annotations below this line if any new ones are to be scanned.
+            // Load the class to avoid a NoClassDefFoundError if it is not present in the impl
+            ClassLoader loader = AnnotationMap.class.getClassLoader();
+            addAnnotationIfPresent(loader, "javax.faces.view.facelets.FaceletsResourceResolver");
```
fjuma, line 76: Could use `FaceletsResourceResolver.class.getName()` here.

fjuma (commented):
@rmartinc Thanks, looks good! I've just added one minor comment.

jsf/injection/src/main/java/org/jboss/as/jsf/injection/AnnotationMap.java
rmartinc: `FaceletsResourceResolver` is not present in JSF 2.1 and 2.0. If it is imported a `NoClassDefFoundError` is thrown and the Jandex provider is not loaded (default JSF one is used, `DelegatingAnnotationProvider` in 2.1 for example). I discovered this when I did some tests with the multi-jsf. So I decided to do a `loadClass` and don't import the class at all, to make it work with old 2.0 and 2.1 impls. WDYT?

fjuma: Ah, ok, that's fine then.

fjuma (approved):

jsf/injection/src/main/java/org/jboss/as/jsf/injection/AnnotationMap.java
rmartinc: I tested with ee9 too because of the name, and it's translated OK. :smiley:

fjuma: Great, was just thinking about that. Thanks very much, Ricardo!

bstansberry: :+

PR #13871, opened 2021-01-11
WFLY-14275 Large job repository is blocking deployment

chengfang:
https://issues.redhat.com/browse/WFLY-14275

This PR contains the following improvement to `org.wildfly.extension.batch.jberet.deployment.BatchJobExecutionResource`:

1, limit the max number of job instances (e.g., 100) to retrieve during refreshing process;
2, avoid excessive refreshing, by keeping track of the last refreshing time, and only refresh after certain interval (e.g., 3000 ms) elapses.

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchJobExecutionResource.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchJobExecutionResource.java
@@ -52,6 +52,19 @@ public class BatchJobExecutionResource implements Resource {
     // Should be guarded by it's instance
     private final Set<String> children = new LinkedHashSet<>();
 
+    /**
+     * Last time when job names were refreshed
+     */
+    private volatile long lastRefreshedTime;
+
+    /**
+     * The minimum interval in milliseconds in which the job names are to be refreshed.
+     * If the interval period has elapsed from the last refresh time,
+     * any incoming refresh request will be performed; otherwise, it is ignored
+     * and the current result is returned.
+     */
+    private final int refreshMinInterval = 3000;
+
     BatchJobExecutionResource(final WildFlyJobOperator jobOperator, final String jobName) {
         this(Factory.create(true), jobOperator, jobName);
     }
@@ -216,9 +229,13 @@ public class BatchJobExecutionResource implements Resource {
      * guarded.
      */
     private void refreshChildren() {
+        if (System.currentTimeMillis() - lastRefreshedTime < refreshMinInterval) {
+            return;
+        }
+
         final List<JobExecution> executions = new ArrayList<>();
         // Casting to (Supplier<List<JobInstance>>) is done here on purpose as a workaround for a bug in 1.8.0_45
-        final List<JobInstance> instances = jobOperator.allowMissingJob((Supplier<List<JobInstance>>)() -> jobOperator.getJobInstances(jobName, 0, jobOperator.getJobInstanceCount(jobName))
+        final List<JobInstance> instances = jobOperator.allowMissingJob((Supplier<List<JobInstance>>)() -> jobOperator.getJobInstances(jobName, 0, Integer.MAX_VALUE)
                 , Collections.emptyList());
         for (JobInstance instance : instances) {
             executions.addAll(jobOperator.getJobExecutions(instance));
@@ -228,5 +245,6 @@ public class BatchJobExecutionResource implements Resource {
             final String name = Long.toString(execution.getExecutionId());
             children.add(name);
         }
+        lastRefreshedTime = System.currentTimeMillis();
     }
 }


chengfang:
@jamezp  can you please review?  If it's too late for 22.0.0.Final, that's okay; we can try to include it in 22.0.1 and 23.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchJobExecutionResource.java
```diff
@@ -52,6 +52,19 @@
     // Should be guarded by it's instance
     private final Set<String> children = new LinkedHashSet<>();
 
+    /**
+     * Last time when job names were refreshed
+     */
+    private long lastRefreshedTime;
```
jamezp, line 58: Maybe not crucial, but it might be good to make this `volatile`.

jamezp (commented):
Just a general question @chengfang. Is there a reason this is being accessed multiple times during a deploy operation? That seems odd to me.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchJobExecutionResource.java
chengfang: I've seen the metrics collector `org.wildfly.extension.metrics.MetricCollector` triggers this refreshing multiple times during a deployment. Sometimes 2, I've also seen 3 times. Every time the stack trace is the same (see the comment in https://issues.redhat.com/browse/WFLY-14275).

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchJobExecutionResource.java
```diff
@@ -216,9 +229,13 @@ private boolean hasJobExecution(final String executionName) {
...
      */
     private void refreshChildren() {
+        if (System.currentTimeMillis() - lastRefreshedTime < refreshMinInterval) {
+            return;
+        }
+
         final List<JobExecution> executions = new ArrayList<>();
         // Casting to (Supplier<List<JobInstance>>) is done here on purpose as a workaround for a bug in 1.8.0_45
-        final List<JobInstance> instances = jobOperator.allowMissingJob((Supplier<List<JobInstance>>)() -> jobOperator.getJobInstances(jobName, 0, jobOperator.getJobInstanceCount(jobName))
+        final List<JobInstance> instances = jobOperator.allowMissingJob((Supplier<List<JobInstance>>)() -> jobOperator.getJobInstances(jobName, 0, 100)
```
jamezp, line 238: This actually changes the current behavior by limiting a maximum of 100 job. I'm not sure we should be changing this. At a minimum we should be making it configurable somehow with it defaulting to the current behavior.

chengfang: Yeah, I was't sure about this either. I think we can preserve the current behavior (I'll change the limit to be Integer.MAX, which will down match the actual count). Hope by avoiding excessive refresh, this can mitigate the long deployment problem.

In the long term, we should optimize this part to use a single operation (a query in jdbc job repository), which will need an upgrade of jberet-core. So the limit here will no longer be needed.

jamezp (approved):
While I'd like to know why the resources are accessed to often during a deploy, I think this is a decent workaround.

bstansberry:
@jamezp The stack trace @chengfang posted in https://issues.redhat.com/browse/WFLY-14275?focusedCommentId=15639993&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15639993 shows one reason for multiple invocations.

MetricsCollector may be suboptimal here:

https://github.com/wildfly/wildfly/blob/master/metrics/src/main/java/org/wildfly/extension/metrics/MetricCollector.java#L133-L134

The "if hasChildren then getChildren" idiom can be inefficient for this kind of runtime-only resource as it results in two executions of the slow logic. It might be more efficient in other cases (saves an object allocation with the default Resource impl). Overall though it is probably the wrong idiom.

bstansberry:
@jmesnil @jasondlee Do we have MetricsCollector collecting the container metrics multiple times?

1) I haven't thought hard about this, but doesn't the Stage.VERIFY collection in MetricsSubsystemAdd end up re-collecting all the deployment=* subtree metrics already collected in Stage.RUNTIME via DeploymentMetricProcessor/DeploymentMetricService? It walks the whole resource tree from the root.

2) If the MP Metrics subsystem is installed, isn't MicroProfileMetricsSubsystemAdd and that subsystem's DeploymentMetricProcessor/DeploymentMetricService also collecting the same set of metrics?

chengfang:
@jmesnil  @jasondlee any ideas on the above inquiry by Brian regarding MetricsCollector?

bstansberry:
@chengfang @jamezp I think we should go ahead with this, as with various competing priorities I don't know how quickly anything will be done about the questions I raised. I filed https://issues.redhat.com/browse/WFLY-14372 to track that.

PR #13872, opened 2021-01-12
[WFLY-14277] Upgrade to support MicroProfile JWT 1.2

darranl:
https://issues.redhat.com/browse/WFLY-14277
https://github.com/wildfly/wildfly-proposals/pull/358

This pull request is using a release candidate of SmallRye JWT so a couple more updates will be required to complete the work under WFLY-14277.

This pull request does add a new dependency, however this is just a new artefact from the SmallRye JWT project so not a true new dependency.

No community documentation is required as we do not specify the version of MicroProfile JWT supported in the documentation and the general approach has not changed with this update.

No test updates are required as this PR pulls in the latest version of the MicroProfile JWT TCK to add coverage for the latest areas required in the specification.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,7 +364,7 @@
         <version.org.eclipse.microprofile.fault-tolerance.api>2.1.1</version.org.eclipse.microprofile.fault-tolerance.api>
         <version.org.eclipse.microprofile.fault-tolerance.tck>${version.org.eclipse.microprofile.fault-tolerance.api}</version.org.eclipse.microprofile.fault-tolerance.tck>
         <version.org.eclipse.microprofile.health.api>3.0</version.org.eclipse.microprofile.health.api>
-        <version.org.eclipse.microprofile.jwt.api>1.1.1</version.org.eclipse.microprofile.jwt.api>
+        <version.org.eclipse.microprofile.jwt.api>1.2</version.org.eclipse.microprofile.jwt.api>
         <version.org.eclipse.microprofile.metrics.api>3.0</version.org.eclipse.microprofile.metrics.api>
         <version.org.eclipse.microprofile.openapi>2.0</version.org.eclipse.microprofile.openapi>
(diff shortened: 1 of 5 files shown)


kabir:
@darranl please link to the proposal in the description

testsuite/integration/microprofile-tck/jwt/tck-base-suite.xml
```diff
@@ -40,6 +40,9 @@
         </groups>
         <classes>
             <class name="org.eclipse.microprofile.jwt.tck.util.TokenUtilsTest" />
+            <class name="org.eclipse.microprofile.jwt.tck.util.TokenUtilsEncryptTest" />
```
kabir, line 43: @darranl Out of curiosity why are you listing all these classes here? It does not seem to be needed (to this extent) for the other specs

pom.xml
```diff
@@ -2155,6 +2155,11 @@
                 <artifactId>smallrye-jwt</artifactId>
                 <version>${version.io.smallrye.smallrye-jwt}</version>
             </dependency>
+            <dependency>
+                <groupId>io.smallrye</groupId>
+                <artifactId>smallrye-jwt-common</artifactId>
+                <version>${version.io.smallrye.smallrye-jwt}</version>
```
kabir, line 2161: This new dependency should be added to `microprofile/galleon-common/src/license/microprofile-feature-pack-licenses.xml`

testsuite/integration/microprofile-tck/jwt/tck-base-suite.xml
darranl: I believe this approach was from the instructions in MicroProfile JWT at the time 1.1 was initially added to identify the mandatory set of tests.  I would need to check but there may have been additional test on top of this mandatory set.

pom.xml
darranl: @kabir done

PR #13876, opened 2021-01-12
[WFLY-14282] HostExcludesTestCase fails when using wildfly-ee-galleon…

yersan:
…-pack server distribution

Jira issue: https://issues.redhat.com/browse/WFLY-14282


yersan:
Just for the record, a Job running the alternative wildfly-ee-galleon-pack is also in green and can be checked here: https://ci.wildfly.org/viewLog.html?buildId=237630&buildTypeId=WF_GalleonLinuxAltFpJdk8&tab=buildResultsDiv

bstansberry:
Thanks, @yersan, particularly for the link above to the custom test run; thank kind of thing is very helpful.

PR #13879, opened 2021-01-13
WFLY-14276 Store elytron security identity in local context, not as a replicated session attribute, as this is not serializable.

pferraro:
https://issues.redhat.com/browse/WFLY-14276


bstansberry:
@fjuma or @darranl Please review

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
```diff
@@ -56,15 +54,17 @@
     // These mechanisms can auto-reauthenticate and thus use local context (instead of replicating)
     private static final Set<String> AUTO_REAUTHENTICATING_MECHANISMS = new HashSet<>(Arrays.asList(HttpServletRequest.BASIC_AUTH, HttpServletRequest.DIGEST_AUTH, HttpServletRequest.CLIENT_CERT_AUTH));
     static final String WEB_SOCKET_CHANNELS_ATTRIBUTE = "io.undertow.websocket.current-connections";
+    static final String IDENTITY_CONTAINER_ATTRIBUTE = "org.wildfly.elytron.web.undertow.server.servlet.ServletSecurityContextImpl$IdentityContainer";
```
darranl, line 57: As the class name referenced here is private API we would not automatically check on class name references like this if we perform any refactoring, I would recommend a comment in the Undertow class to point us back here.

darranl:
I think the change made here is reasonable, it would mean on failover a complete session is replicated minus the authentication so authentication would need to be performed again.  I am not sure if this matches the original bug report as I think they also want the authentication to be replicated but we have the SSO implementation for that.

My only concern is what happens if the user authenticates as a different user, i.e. permission checks may have been performed as the session is populated but then authentication happens using a different user those prior checks may no longer be valid.

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
pferraro: I've opened https://issues.redhat.com/browse/ELYWEB-124 to track this.

pferraro:
> My only concern is what happens if the user authenticates as a different user, i.e. permission checks may have been performed as the session is populated but then authentication happens using a different user those prior checks may no longer be valid.

If an application uses relies on security checks for securing access to specific session attributes, they have to manage that access manually - there is nothing in the servlet specification that applies security context to specific session attributes.  Otherwise the server has no way of knowing which attributes are "secured" and which are not.

darranl (approved):

PR #13883, opened 2021-01-13
[WFLY-14206] Update MP Metrics to 3.0

jasondlee:
These changes will update MP Metrics in Wildfly to 3.0.

JIRA: https://issues.redhat.com/browse/WFLY-14206
Analysis Doc PR: wildfly/wildfly-proposals#354


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/async/requestcontext/AsynchronousRequestContextTest.java
```diff
@@ -61,6 +62,7 @@ public static WebArchive createTestArchive() {
     }
 
     @Test
+    @Ignore("Test is broken to to a dependency update. Will be fixed in WFLY-14281.")
```
kabir, line 65: I created https://issues.redhat.com/browse/WFLY-14296 so this does not get lost

PR #13887, opened 2021-01-14
[WFLY-14293]:Fix regression failures when deploy JMS  and session Endpoint

jimma:
https://issues.redhat.com/browse/WFLY-14293


asoldano (approved):
The fix looks correct to me, at least in the first part. I'm not 100% sure about the change in AbstractInvocationHandler, but I trust that's needed.
For the future, @jimma , I suggest you explicitly ask @rsearls to formally review your PRs.

PR #13888, opened 2021-01-14
[22.x][WFLY-14293]:Fix regression failures when deploy JMS and session endpoint

jimma:
https://issues.redhat.com/browse/WFLY-14293
Upstream: #13887


asoldano (approved):
Same comment I did in https://github.com/wildfly/wildfly/pull/13887 applies here.

PR #13899, opened 2021-01-20
[WFLY-14310] Use standard idioms for handling attribute resolution in…

bstansberry:
… JPASubsystemAdd

Remove the 'default-datasource=""' xml attribute from our standard configs, which was only necessary as a workaround of the fact that JPASubsystemAdd was incorrectly calling 'require' to read an optional attribute.

This also does a bit of code cleanup in the class and associated ResourceDefinition.

https://issues.redhat.com/browse/WFLY-14310

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
@@ -53,6 +53,23 @@ public class JPADefinition extends SimpleResourceDefinition {
             .addRequirements(JPAServiceNames.LOCAL_TRANSACTION_PROVIDER_CAPABILITY)
             .build();
 
+    protected static final SimpleAttributeDefinition DEFAULT_DATASOURCE =
+            new SimpleAttributeDefinitionBuilder(CommonAttributes.DEFAULT_DATASOURCE, ModelType.STRING, true)
+                    .setAllowExpression(true)
+                    .setXmlName(CommonAttributes.DEFAULT_DATASOURCE)
+                    .setValidator(new StringLengthValidator(0, Integer.MAX_VALUE, true, true))
+                    .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
+                    .build();
+
+    protected static final SimpleAttributeDefinition DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE =
+            new SimpleAttributeDefinitionBuilder(CommonAttributes.DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE, ModelType.STRING, true)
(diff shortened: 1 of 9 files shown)


bstansberry:
@scottmarlow Please review.

scottmarlow:
@bstansberry are the test failures caused by this change?

bstansberry:
@scottmarlow I'm going to assume so.

scottmarlow:
I'll try building this locally.

scottmarlow:
I'm running a few of the JPA tests locally with https://github.com/scottmarlow/wildfly/commit/ad0fbfbaf9babf47c6a2387f809793c1dbb042ae

bstansberry:
Thanks, @scottmarlow. Yes, that check should not be there as the xsd records the attribute as optional. And the rest of my PR removes the Stage.RUNTIME code that required it to be there despite what the xsd and attribute definition say.

This will teach me not to run the smoke tests before pushing something up.

I'll add a subsystem test of an empty subsystem element.

scottmarlow:
@bstansberry the local JPA tests are passing for me and the management console/jboss-cli output looks good (standalone.xml has expected content after setting default datasource and also correct after removing default datasource).

bstansberry:
@scottmarlow I've updated this. Adding new subsystem tests revealed a problem in how the xml is written if the none of the optional attributes are defined; I fixed that as well.

scottmarlow (approved):

jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
```diff
@@ -53,6 +53,23 @@
...
+
+    protected static final SimpleAttributeDefinition DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE =
+            new SimpleAttributeDefinitionBuilder(CommonAttributes.DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE, ModelType.STRING, true)
+                    .setAllowExpression(true)
+                    .setXmlName(CommonAttributes.DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE)
+                    .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
+                    .setValidator(new EnumValidator<ExtendedPersistenceInheritance>(ExtendedPersistenceInheritance.class,true,true))
+                    .setDefaultValue(new ModelNode(ExtendedPersistenceInheritance.DEEP.toString()))
+                    .build();
+
```
bstansberry, line 72: Note for the historical record: The reason this code is moved is because it's all static initialization, and the order in which that is done matters. We want these ADs initialized before the INSTANCE field on L73 is initialized, as that initialization in turn calls code that reads these AD variables.

PR #13910, opened 2021-01-22
[WFLY-14339] Add the javax.api module back as a module dependency to …

jamezp:
…the legacy security. Note the explicit required dependencies were not reverted and only the javax.api dependency was reintroduced.

https://issues.redhat.com/browse/WFLY-14339

### DIFF
--- a/servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/security/main/module.xml
+++ b/servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/security/main/module.xml
@@ -39,7 +39,7 @@
         <module name="java.naming"/>
         <module name="java.security.sasl"/>
         <module name="java.xml"/>
-        <!--WFLY-14219 Remove deprecated <module name="javax.api"/> -->
+        <module name="javax.api"/>
         <module name="javax.security.jacc.api"/>
         <module name="javax.servlet.api"/>
         <module name="javax.transaction.api"/>


scottmarlow (approved):

servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/security/main/module.xml
```diff
@@ -39,7 +39,7 @@
         <module name="java.naming"/>
         <module name="java.security.sasl"/>
         <module name="java.xml"/>
-        <!--WFLY-14219 Remove deprecated <module name="javax.api"/> -->
+        <module name="javax.api"/>
```
ropalka, line 42: This is too big dependency.
"javax.api" is a legacy aggregation module which aggregates:
"java.se", "java.xml", "java.xml.crypto"  and "jdk.xml.dom" JPMS modules.
In the JIRA I see:
"There could be cases where a provider is used that relies on API's provided by the javax.api module."
If such a provider is using classes from all four aggregated modules then 
define them explicitly as dependency instead of using "javax.api" legacy module.
It is preferred to reference JPMS modules instead of deprecated "javax.api" module.

ropalka (changes requested):

servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/security/main/module.xml
bstansberry: This is a temporary workaround to deal with a setup issue in an internal testsuite.  We should fix the setup issue but that is not an immediate priority, but continuing to pass that testsuite is.

There are still ~ 40 issues of javax.api so even if we forget to revert this when we get around to cleaning up the setup, we can get to it when we clean up those other 40.

darranl: @ropalka This module is also due to be removed in the not too distant future, I assume maybe at the same time if not before javax.api - would keeping this dependency for a short while be a problem?

ropalka: Got it @bstansberry .

ropalka: It will not be a problem @darranl to keep that dependency there for a short while.

PR #13915, opened 2021-01-24
[WFLY-14342] Use AttributeDefinition.resolveModelAttribute to read th…

bstansberry:
…e PROTOCOL_BINDINGS attribute from the model

[WFLY-14343] Don't pass 'undefined' into the runtime as a value of a property value when the model node is undefined.

Also update other add handlers in the webservices subsystem to use current idioms. I found the PROTOCOL_BINDINGS issue while reviewing all these add handlers so might as well update things a bit.

https://issues.redhat.com/browse/WFLY-14342
https://issues.redhat.com/browse/WFLY-14343


bstansberry:
@jimma Please review.

jimma:
It looks good. Thanks Brain.

PR #13920, opened 2021-01-25
[WFLY-11272] wildfly-microprofile-health-smallrye_1_0.xsd doesn't hav…

xstefank:
…e documentation for allowed attributes

https://issues.redhat.com/browse/WFLY-11272


bstansberry:
Thanks, @xstefank !

PR #13923, opened 2021-01-26
[WFLY-14006] Deprecate the extensions where the root subsystem is alr…

yersan:
…eady deprecated

Jira issue: https://issues.redhat.com/browse/WFLY-14006

@jmesnil @bstansberry The following extensions are suspicious, they are registered via [AbstractLegacyExtension.java#L101](https://github.com/wildfly/wildfly-core/blob/master/controller/src/main/java/org/jboss/as/controller/extension/AbstractLegacyExtension.java#L101), which means are allowed to be registered to manage a legacy server instance. As such, it looks like they should have been deprecated. However, their root subsystems have not been deprecated yet:

```
/extension=org.jboss.as.configadmin:add()
/extension=org.jboss.as.jaxr:add()
/extension=org.wildfly.extension.rts:add()
```

### DIFF
--- a/legacy/cmp/src/main/java/org/jboss/as/cmp/subsystem/CmpExtension.java
+++ b/legacy/cmp/src/main/java/org/jboss/as/cmp/subsystem/CmpExtension.java
@@ -55,7 +55,7 @@ public class CmpExtension extends AbstractLegacyExtension {
     @Override
     protected Set<ManagementResourceRegistration> initializeLegacyModel(final ExtensionContext context) {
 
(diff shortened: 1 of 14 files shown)


bstansberry:
@yersan Configadmin and Jaxr should be deprecated; good catch. The RTS subsystem is not based on AbstractLegacyExtension. I don't believe it is deprecated.

yersan:
retest this please

yersan:
@bstansberry Correct, rts subsystem should not have been added to the list. I fixed the others.

PR #13937, opened 2021-01-29
WFLY-14322 Replace 'Contexts and Dependency Injections|CDI' reference…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-14322
Link with: https://github.com/wildfly/wildfly/pull/13808

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/JTA_Reference.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/JTA_Reference.adoc
@@ -494,3 +494,4 @@ or you can list all the configuration options of the subsystem in `jboss-cli`
 ```bash
 /subsystem=transactions:read-resource-description(recursive=true)
 ```
+NOTE: References in this document to CDI refer to Jakarta Contexts and Dependency Injection unless otherwise noted.
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
@@ -60,7 +60,7 @@ public class ArtifactFactoryService extends AbstractArtifactFactory implements S
     public void destroy(final Object instance) {
         final Holder holder = contexts.remove(instance);
         if (holder == null) {
-            // This bean was not created via CDI, we need to invoke the JBeret cleanup
+            // This bean was not created via Jakarta Contexts and Dependency Injection, we need to invoke the JBeret cleanup
             super.destroy(instance);
         } else {
             // Only release the context for @Dependent beans, Weld should take care of the other scopes
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentProcessor.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentProcessor.java
@@ -144,7 +144,7 @@ public class BatchEnvironmentProcessor implements DeploymentUnitProcessor {
             final ArtifactFactoryService artifactFactoryService = new ArtifactFactoryService();
             final ServiceBuilder<ArtifactFactory> artifactFactoryServiceBuilder = serviceTarget.addService(artifactFactoryServiceName, artifactFactoryService);
 
-            // Register the bean manager if this is a CDI deployment
+            // Register the bean manager if this is a Jakarta Contexts and Dependency Injection deployment
             if (support.hasCapability(WELD_CAPABILITY_NAME)) {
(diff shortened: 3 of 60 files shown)


RanabirChakraborty:
@manovotn If you can review this PR :) (CC- @asoldano @scottmarlow)

bstansberry:
@RanabirChakraborty @asoldano 

This looks good in general. There are just a few adoc files where a many times repeated replacement of 'CDI' with the much more verbose 'Jakarta Contexts and Dependency Injection' is pretty clunky.

I note the following statement in https://jakarta.ee/legal/acronym_guidelines/

```
If a specification document (or related materials) contains references to an Oracle/JCP acronym, include a sentence such as the following: “References in this document to JMS refer to the Jakarta Message Service unless otherwise noted." The sentence needs to be a legible size and should appear early in the specification document (or related materials), not in an illegibly small footnote. A reader should not have to scroll, turn a page, or click a link to get to the sentence.
```

Is this something we can utilize here? (And in other cases, although this case may be one of the worst when it comes to verbosity.)

The docs where this might be useful are

JTA_Reference.adoc
Jakarta_Contexts_and_Dependency_Injection_Reference.adoc
Getting_Started_Developing_Applications_Presentation_Demo.adoc

Those are the ones with many replacements. The others either there is no repetition or it's not a lot. (In javadoc and code comments I'm not much concerned about being concise; we should just do what's easy and not worry about what I'm discussing here.)

RanabirChakraborty:
I guess then it'll be much easier to do the changes for big components (such as EJB) without doing the same changes repeatedly for a single file. I can mention it with one line at the beginning of the file. 

@bstansberry But there is one concern. Suppose there is EJB 3.1 and EJB 3.2 in the same file. but I can only change EJB 3.2 to Jakarta Enterprise Beans 3.2(https://jakarta.ee/specifications/enterprise-beans/). Then this line will not work I guess.

manovotn:
I am probably missing something but why do we need to perform this change?

CDI as an acronym if far better than `Jakarta Contexts and Dependency Injection' as in easier to read and understand for user. After all, the user facing API stays the same. If need be, we could add the one-liner that @bstansberry suggested - that's IMO a good move purely for clarification purposes.

Also, note that CDI was originally RH project/specification which we then donated to Jakarta and apart from being part of Java EE ecosystem, it wasn't owned by Oracle which to me sounds like we are not obliged to change the acronym anyway, WDYT?

bstansberry:
@manovotn I'm going to defer to @asoldano as to whether work related to CDI is necessary. It's a good question.

@RanabirChakraborty You should confirm with Alessio whether my suggestion is valid before applying it elsewhere (or even here.) I think what to do about EJB 3.1 / 3.2 details is better discussed on that JIRA with the guys who work on EJB.

asoldano (commented):
Thanks @RanabirChakraborty (and @bstansberry / @manovotn for your notes). I agree with the proposal of adding a clear / legible note "References in this document to CDI refer to the Jakarta Context and Dependency Injection unless otherwise noted" to the pages where the substitutions would make the text too hard to read.
This would be fine in general, I think; the cases where the specification version reference matters (EJB3.1 vs 3.2) should be discussed in the specific jiras with SME engineers.

RanabirChakraborty:
@bstansberry @asoldano Thanks for the review, done the required changes :)

rhusar (changes requested):
The copyright headers should not be updated.

RanabirChakraborty:
> The copyright headers should not be updated.
@rhusar Okay! so should I move the year as it was before, right?

RanabirChakraborty:
> The copyright headers should not be updated.
@rhusar Okay! so should I move the year as it was before, right?

bstansberry:
@RanabirChakraborty yes, copyright dates back to what they were

RanabirChakraborty:
Thanks, @bstansberry @rhusar Done the changes

docs/src/main/asciidoc/_developer-guide/JTA_Reference.adoc
```diff
@@ -494,3 +494,4 @@ or you can list all the configuration options of the subsystem in `jboss-cli`
 ```bash
 /subsystem=transactions:read-resource-description(recursive=true)
 ```
+NOTE: References in this document to CDI refer to the Jakarta Contexts and Dependency Injection unless otherwise noted
```
bstansberry, line 497: s/to the Jakarta/to Jakarta/g

Also the sentence should end in a period.

Please apply the same to the other uses of this sentence in the PR.

bstansberry (changes requested):
One more final change requested. :)

docs/src/main/asciidoc/_developer-guide/JTA_Reference.adoc
RanabirChakraborty: Thanks! Done :)

bstansberry (approved):
One more final change requested. :)

PR #13940, opened 2021-01-30
WFLY-14359 ProtoStream marshallers for Object[] should not use packed encoding

pferraro:
Also, WFLY-14270 ProtoStream proxy marshalling should not imply JBoss Modules
Ensures all java.lang.* marshallers use valid protobuf encoding.

https://issues.redhat.com/browse/WFLY-14359
https://issues.redhat.com/browse/WFLY-14270

PR #13941, opened 2021-01-30
WFLY-14381 Ensure Infinispan specific marshallers use valid protobuf encoding

pferraro:
https://issues.redhat.com/browse/WFLY-14381

Requires https://github.com/wildfly/wildfly/pull/13940


pferraro:
@rhusar Please review.

bstansberry (approved):

clustering/marshalling/protostream/src/main/java/org/wildfly/clustering/marshalling/protostream/EnumMarshaller.java
```diff
@@ -22,15 +22,16 @@
...
 
-import org.infinispan.protostream.ImmutableSerializationContext;
+import org.infinispan.protostream.impl.WireFormat;
 
 /**
  * ProtoStream marshaller for enums.
  * @author Paul Ferraro
  */
-public class EnumMarshaller<E extends Enum<E>> implements org.infinispan.protostream.EnumMarshaller<E>, Predictable<E> {
+public class EnumMarshaller<E extends Enum<E>> implements org.infinispan.protostream.EnumMarshaller<E>, ProtoStreamMarshaller<E> {
```
aimiebell, line 33: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of implemented interface (NM_SAME_SIMPLE_NAME_AS_INTERFACE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-implemented-interface-nm-same-simple-name-as-interface).
This class/interface has a simple name that is identical to that of an implemented/extended interface, except that the interface is in a different package (e.g., alpha.Foo extends beta.Foo). This can be exceptionally confusing, create lots of situations in which you have to look at import statements to resolve references and creates many opportunities to accidentally define methods that do not override methods in their superclasses.

PR #13935, opened 2021-01-28
WFLY-14315 Replace JavaServer Faces|JSF references in java doc with c…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-14315
Link with: https://github.com/wildfly/wildfly/pull/13808

PR #13952, opened 2021-02-02
WFLY-14325 Replace 'Batch Applications for the Java Platform|JBatch, …

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-14325
(Replace "Batch Applications for the Java Platform|JBatch, JavaMail API, JavaBeans Activation Framework|JAF, Bean Validation|JBV, Expression Language|JEL" references in java doc with corresponding Jakarta EE ones)
Link with: https://github.com/wildfly/wildfly/pull/13808


RanabirChakraborty:
CC- @asoldano @scottmarlow

docs/src/main/asciidoc/Getting_Started_Guide.adoc
```diff
@@ -44,22 +44,22 @@ Profile |WildFly {wildflyVersion} Full Platform |WildFly {wildflyVersion} Web Pr
 
 |JSR-372: Jakarta Server Faces 2.3 |X |X |X |X
 
-|JSR-341: Expression Language 3.0 |X |X |X |X
+|JSR-341: Jakarta Expression Language 3.0 |X |X |X |X
```
scottmarlow, line 47: Should we remove the JSR-341 (and other JSR-NNN) references?  

How about the Specification version number?

scottmarlow:
IMO, replacing `Bean Validation` with `Jakarta Bean Validation` looks good to me.  I asked whether we should also remove SPEC version references and JSR-NNN references as the SPEC versions change often and the JSR-NNN references could be kept but are referring to Java EE `Java Specification Requests` which don't have a clear `Jakarta EE` replacement.

docs/src/main/asciidoc/Getting_Started_Guide.adoc
RanabirChakraborty: Yes, I guess you are right. I should remove those JSR-NNN references because after all there is no JSR-NNN reference for Jakarta specifications.
And about the specification version number, I'm not completely sure what is the recent specification version for all, which we are going to use in Wildfly 23. That's why I just kept the original one which was there before and replaced it by the Jakarta one. 
and also I think it's good to have the specification version number so that one can check the documents and search the details about it.
@asoldano if you want to add anything.

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
```diff
@@ -0,0 +1,4 @@
+[[Jakarta_Mail_API]]
```
yersan, line 1: Not by this change, but all the blocks Ids which are using "(" and ")" characters make the documentation render badly. See https://docs.wildfly.org/22/JavaEE_Tutorial.html#JavaMail_API for example.


@RanabirChakraborty Would you mind removing these invalid characters to make the documentation render correctly? It is not really on the scope of this Jira, but it sounds appropriate to use this one.

I'm pointing you to the blocks IDs, ror example, replace `[[Java_Authorization_Contract_for_Containers_(JACC)]]` by `[[Java_Authorization_Contract_for_Containers_JACC]]` in Java_Authorization_Contract_for_Containers_(JACC).adoc file and so on.

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
```diff
@@ -0,0 +1,4 @@
...
+= Jakarta Mail API
```
yersan, line 2: Could you use `Jakarta Mail` instead of `Jakarta Mail API`?  It is more widely used without "API".

yersan (changes requested):
Thank you @RanabirChakraborty for taking care of this. 

Regarding Jakarta Mail, I missed this replacement on Getting_Started_Guide.adoc:79
s/JavaMail/Jakarta Mail

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
RanabirChakraborty: I have done the same changes here - https://github.com/wildfly/wildfly/pull/13953/files , please take a look :) I'll take care of others too.

RanabirChakraborty:
@yersan @scottmarlow thanks for your review changed it accordingly :)

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
yersan: I could be missing something but what I can see is you are changing the file names, but not the block IDs.

The file names can contain "(" and ")" characters, but the Block IDs, those are the headers with "[[" and "]]" markers cannot contain those characters, for example, Java_Authorization_Contract_for_Containers_(JACC).adoc, line 1, contains `[[Java_Authorization_Contract_for_Containers_(JACC)]]` but it should be `[[Java_Authorization_Contract_for_Containers_JACC]]`

RanabirChakraborty: Thanks! got it now :) 
https://github.com/wildfly/wildfly/pull/13953/files#diff-26d164d098fc71f48b3264385036a769956734ec202f9a1fe7ae1f3c8df090a7
did it for other specifications, forgot to do for JACC.

yersan (commented):
Hi @RanabirChakraborty , I caught two more related to the Mail subsystem:

https://github.com/wildfly/wildfly/blob/master/mail/src/main/resources/org/jboss/as/mail/extension/LocalDescriptions.properties#L8
https://github.com/wildfly/wildfly/blob/master/mail/src/main/resources/org/jboss/as/mail/extension/LocalDescriptions.properties#L20

Would you mind replacing the `JavaMail` with `Jakarta Mail` on those files as well?

RanabirChakraborty:
Thanks! @yersan done :)

docs/src/main/asciidoc/Getting_Started_Guide.adoc
chengfang: Having spec version is a bit hard to maintain as the spec evolves. We'll need to take care to adjust the spec version to the current one. For Jakarta Batch, it's versoin 1.0 in EE 8, and version 2.0 in EE 9.

chengfang:
A couple of files reference the old name of Jakarta Batch. We may want to replace it as well.

[[Batch_JSR-352]]
./_admin-guide/subsystem-configuration/Batch_JSR-352.adoc

include::subsystem-configuration/Batch_JSR-352.adoc[]
./_admin-guide/Subsystem_configuration.adoc

yersan:
Hello @RanabirChakraborty , following the criteria taken in the other PR related, it seems we should not upgrade the license headers after modifying these java docs and logger messages. Would you mind rolling back them to the original? thanks!

docs/src/main/asciidoc/Getting_Started_Guide.adoc
RanabirChakraborty: So, should I remove this spec version?

RanabirChakraborty:
@chengfang @yersan did the changes :)

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
yersan: Ok @RanabirChakraborty , it is difficult to track all of these changes across all the PRs. I understand your idea is to change all of these wrong headers on each PR, right? If so, take into account that the JSF one was not fixed, see https://github.com/wildfly/wildfly/pull/13935/files#diff-e3b6f2a3adda324b932ed5156478991589b9656814257e8b1891a4fc9ffe1657R1

In any case, I think the easier and quicker solution could be to review all of these at the end when WFLY-14061 is resolved to verify we haven't missed any.

yersan (approved):
There are conflicts by other PRs, but looks good on my side from Jakarta Mail and Jakarta Activation. Thank you @RanabirChakraborty

docs/src/main/asciidoc/_javaee-guide/Jakarta_Mail_API.adoc
RanabirChakraborty: Yeah @yersan sorry that was my bad. JSF was the first one to merge and at that time I haven't got many ideas bout the whole thing. I have it in my mind that I need to do a few more changes for JSF. I'll do that.
Yes, it's also a really good idea to review all of these after all the sub-tasks are merged. I'll send a separate PR with all the leftover changes and link the same to the main issue - https://issues.redhat.com/browse/WFLY-14061 and PR - https://github.com/wildfly/wildfly/pull/13808

docs/src/main/asciidoc/Getting_Started_Guide.adoc
bstansberry: @RanabirChakraborty Removing the spec versions is probably a good idea but it seems out of scope for what you're doing here. You're dividing your work by technology area, while this is something we should discuss and then do as a single piece of work.

asoldano: @RanabirChakraborty , +1 on leaving the removal of the spec version to a future effort.

chengfang (approved):

PR #13954, opened 2021-02-03
[WFLY-12274] add warning when MDB/Session bean has mappedName declara…

baranowb:
…tion in its definition

JIRA: https://issues.redhat.com/browse/WFLY-12274

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/SessionBeanComponentDescriptionFactory.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/SessionBeanComponentDescriptionFactory.java
@@ -174,6 +174,12 @@ public class SessionBeanComponentDescriptionFactory extends EJBComponentDescript
             }
 
             addComponent(deploymentUnit, sessionBeanDescription);
+
+            final AnnotationValue mappedNameValue = sessionBeanAnnotation.value("mappedName");
+            if (mappedNameValue != null && !mappedNameValue.asString().isEmpty()) {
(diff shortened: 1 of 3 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/SessionBeanComponentDescriptionFactory.java
```diff
@@ -174,6 +174,12 @@ private void processSessionBeans(final DeploymentUnit deploymentUnit, final List
             }
 
             addComponent(deploymentUnit, sessionBeanDescription);
+
+            final AnnotationValue mappedNameValue = sessionBeanAnnotation.value("mappedName");
+            if (mappedNameValue != null && !mappedNameValue.asString().isEmpty()) {
+                EjbLogger.ROOT_LOGGER.mappedNameNotSupported(mappedNameValue != null ? mappedNameValue.asString() : "",
+                        ejbName);
+            }
```
chengfang, line 182: `mappedNameValue != null ? mappedNameValue.asString() : "",` seems to be redundant, as mappedNameValue is already not null.

bstansberry:
@baranowb  @chengfang This one affects internationalized strings so it would be good to get it done Monday or Tuesday so we it's in what we hand off to people doing translations.

chengfang (approved):

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/SessionBeanComponentDescriptionFactory.java
chengfang: consider the above comment as optional minor improvement, not blocking one. I've approved this PR.

PR #13955, opened 2021-02-03
[WFLY-14403] introduce a test case to automatically test configurati…

istraka:
…tion and challenge expression values and write op
https://issues.redhat.com/browse/WFLY-14403

The core of test is from ExpressionSupportSmokeTestCase (domain TS).

### DIFF
--- a/testsuite/integration/src/test/scripts/manualmode-build.xml
+++ b/testsuite/integration/src/test/scripts/manualmode-build.xml
@@ -105,6 +105,28 @@
         <copy todir="target/jbossas-dwm-manual-1" overwrite="true">
             <fileset dir="target/wildfly"/>
         </copy>
+        <copy todir="target/jbossas-custom" overwrite="true">
+            <fileset dir="target/wildfly"/>
+        </copy>
+
+        <xslt destdir="target/jbossas-custom/docs/examples/configs"
+              style="${jbossas.ts.integ.dir}/src/test/xslt/changeLdapStore.xsl"
+              extension=".xml.mod"
+              useImplicitFileset="false">
+            <fileset dir="target/jbossas-custom/docs/examples/configs">
+                <include name="**/standalone-picketlink.xml"/>
+            </fileset>
+        </xslt>
+
+        <!-- Move processed files back. -->
+        <move todir="target/jbossas-custom/docs/examples/configs">
+            <fileset dir="target/jbossas-custom/docs/examples/configs">
+                <include name="**/standalone-picketlink.xml.mod"/>
+            </fileset>
+            <mapper type="glob" from="*.mod" to="*"/>
+        </move>
+
+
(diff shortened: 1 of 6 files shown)


darranl (changes requested):
This project is the upstream community project, please clone the JBEAP Jira issue to a WFLY jira issue and reference the WFLY issue in the commit and PR instead of the downstream JBEAP issue.

istraka:
@darranl done. `No runnable methods` error is also fixed.

testsuite/integration/src/test/scripts/manualmode-build.xml
```diff
@@ -105,6 +105,12 @@
         <copy todir="target/jbossas-dwm-manual-1" overwrite="true">
             <fileset dir="target/wildfly"/>
         </copy>
+        <copy todir="target/jbossas-custom" overwrite="true">
+            <fileset dir="target/wildfly"/>
+        </copy>
+        <copy todir="target/jbossas-custom/standalone/configuration" overwrite="true">
+            <fileset dir="src/test/resources/org/jboss/as/test/manualmode/model/custom-configurations"/>
```
bstansberry, line 112: @istraka Are all the files in this dir basically meant to be what we ship in docs/examples/configs, or have they been altered in some way to fit the needs of the test?

I think it's best if as much as possible we test the configs we provide in docs/examples/configs, as those are what will be maintained; copies can fall out of date. There are different ways that can be accomplished.

It's also possible to configure arquillian to directly use the file in docs/examples/configs. See https://issues.redhat.com/browse/WFARQ-75

istraka: IIRC I took them from 73 zip, so they are bit different. Yeah, we should not maintain it. Now they are taken `build/docs/...` automatically.  `standalone-picketlink.xml` however needs some changes so that `attribute` nodes in LDAP store can be set to an expression. Hence the xslt script.

Moreover, in order to use readOnlyServerConfig I must change to `jboss.server.config.file.name` property `config.file.name` due to [1]. Otherwise Arquillian throws exception since both `serverConfig` and `readOnlyServerConfig` are used.

[1] https://github.com/wildfly/wildfly-arquillian/blob/master/container-managed/src/main/java/org/jboss/as/arquillian/container/managed/ManagedContainerConfiguration.java#L33

istraka: See the latest changes.

istraka:
@darranl can you review, please?

darranl:
The commit message for the commit still references the JBEAP, this needs changing to reference the WFLY issue and remove the issue reference to the JBEAP.  I would suggest whilst doing that unless they represent distinct steps as you are changing one commit the four commits could probably be squashed into one.

istraka:
@darranl changed & squashed.

bstansberry (approved):

PR #13957, opened 2021-02-04
WFLY-14395 Bump messaging schema to 13.0

TomasHofman:
https://issues.redhat.com/browse/WFLY-14395


ehsavoie:
There is already https://github.com/wildfly/wildfly/pull/13906 which is used by 4 other PRs as base

PR #13961, opened 2021-02-05
[WFLY-14390] WildFly does not start when async-registration attribute in XTS subsystem is set to an expression

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-14390

Just removing the lines that checks for the true/false value and let the `parseAndSetParameter` do the job. The sample subsystem test file is updated to set a expression that defaults to true as before, this way the expression value is tested in the subsystem tests.

@ochaloup Please review the PR.


ochaloup (approved):
Seems fine to me.

PR #13966, opened 2021-02-08
[WFLY-14389] Resource-adapters: correctly marshall application elemen…

tadamski:
…t with expression value

https://issues.redhat.com/browse/WFLY-14389

This fix seem to work for me. In case of application being set to true, it is marshalled as an empty element. In other cases, expression included, it is fully marshalled.


bstansberry (changes requested):
Hi @tadamski -- please add coverage of this to the xml parsed by ResourceAdaptersSubsystemTestCase.

tadamski:
hi @bstansberry, I have attached an update to test configuration; with this configuration current master fails and this branch succeedes

bstansberry (approved):
Nice. :)

bstansberry:
Thanks, @tadamski

PR #13982, opened 2021-02-11
[WFLY-13779] Require reload when removing config source

jasondlee:
https://issues.redhat.com/browse/WFLY-13779

Flag the config-source and config-source-provider :remove operations with RESTART-ALL-SERVICES

### DIFF
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ClassConfigSourceRegistrationService.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ClassConfigSourceRegistrationService.java
@@ -1,0 +1,69 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.microprofile.config.smallrye;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.msc.Service;
+import org.jboss.msc.service.ServiceBuilder;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
+import org.wildfly.extension.microprofile.config.smallrye._private.MicroProfileConfigLogger;
+
+/**
+ * Service to register a ConfigSource reading its configuration from a directory (where files are property keys and
+ * their content is the property values).
+ *
+ * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
+ */
+class ClassConfigSourceRegistrationService implements Service {
+
+    private final String name;
+    private final ConfigSource configSource;
+    private final Registry<ConfigSource> sources;
+
+    ClassConfigSourceRegistrationService(String name, ConfigSource configSource, Registry<ConfigSource> sources) {
+        this.name = name;
+        this.configSource = configSource;
+        this.sources = sources;
+    }
+
+    static void install(OperationContext context, String name, ConfigSource configSource, Registry registry) {
+        ServiceBuilder<?> builder = context.getServiceTarget()
+                .addService(ServiceNames.CONFIG_SOURCE.append(name));
+
+        builder.setInstance(new ClassConfigSourceRegistrationService(name, configSource, registry))
+                .install();
+    }
+
+    @Override
+    public void start(StartContext startContext) {
+        MicroProfileConfigLogger.ROOT_LOGGER.loadConfigSourceFromClass(configSource.getClass());
+        this.sources.register(this.name, configSource);
+    }
+
+    @Override
+    public void stop(StopContext context) {
+        this.sources.unregister(this.name);
+    }
+}
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
@@ -30,11 +30,10 @@ import static org.wildfly.extension.microprofile.config.smallrye._private.MicroP
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Map;
 
+import io.smallrye.config.PropertiesConfigSource;
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.jboss.as.controller.AbstractAddStepHandler;
-import org.jboss.as.controller.AbstractRemoveStepHandler;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.AttributeMarshaller;
 import org.jboss.as.controller.AttributeMarshallers;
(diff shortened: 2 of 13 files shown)


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source/SetupTask.java
```diff
@@ -79,6 +80,8 @@
 
     @Override
     public void setup(ManagementClient managementClient, String s) throws Exception {
+        ServerReload.reloadIfRequired(managementClient);
```
jamezp, line 83: This one should not be required.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source_provider/SetupTask.java
```diff
@@ -51,6 +52,8 @@
 
     @Override
     public void setup(ManagementClient managementClient, String s) throws Exception {
+        ServerReload.reloadIfRequired(managementClient);
```
jamezp, line 55: Same here. This one should not be required.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source/SetupTask.java
jasondlee: You're right!

jamezp (approved):

bstansberry (changes requested):
@jasondlee The test failures appear to be related.

jasondlee:
> @jasondlee The test failures appear to be related.

That they do, but I swear I have those running locally. I'll dig.

jasondlee:
Well, I sort of did then, but I do now. :)

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
```diff
@@ -166,4 +130,60 @@ private static Class unwrapClass(ModelNode classModel) throws OperationFailedExc
...
+        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+            String name = context.getCurrentAddressValue();
+            ModelNode dirModel = DIR.resolveModelAttribute(context, model);
+            if (dirModel.isDefined()) {
+                context.removeService(ServiceNames.CONFIG_SOURCE.append(name));
+            } else {
+                sources.unregister(name);
+            }
+            super.performRuntime(context, operation, model);
+        }
```
bstansberry, line 186: @jasondlee This should just be ReloadRequiredRemoveStepHandler, no override.

A management op handler should do one of 4 mutually exclusive things with respect to the running behavior of a server:

1) Fail because something is wrong.
2) Determine it can do what was requested immediately and go ahead and change the runtime services in some way to bring their state in sync with the current configuration.
3) Determine that it can't do what's requested without a reload and therefore put the server in reload-required.
4) Determine that it can't do what's requested without process restart and therefore put the server in restart-required.

It shouldn't partially change the running services but then say reload/restart are required to finish the job.

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceProviderDefinition.java
```diff
@@ -108,4 +89,43 @@ private static Class unwrapClass(ModelNode classModel) throws OperationFailedExc
...
+            this.providers = providers;
+        }
+
+        @Override
+        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+            String name = context.getCurrentAddressValue();
+            providers.unregister(name);
+            super.performRuntime(context, operation, model);
+        }
+    }
```
bstansberry, line 130: Same comment as I made with ConfigSourceDefinitionRemoveHandler applies here as well.

bstansberry (changes requested):
I made some technical comments below. 

Also, can you update the commit message and PR title to change 'restart' to 'reload'?  They mean different things and in this kind of context that difference is important.

jasondlee:
> I made some technical comments below.
> 
> Also, can you update the commit message and PR title to change 'restart' to 'reload'? They mean different things and in this kind of context that difference is important.

On it

bstansberry (approved):

bstansberry:
The one test fails regularly on Windows, so we'll need to get that sorted:

https://ci.wildfly.org/project.html?projectId=WF_PullRequest&buildTypeId=&tab=testDetails&testNameId=-3684118276893383784&order=TEST_STATUS_DESC&branch_WF_PullRequest=__all_branches__&itemsCount=50

bstansberry:
I pushed up a commit that adds more info to the failure message when that test fails (i.e. what the span's operation name is that doesn't have the expected content.)

jasondlee:
I add some logging as well, and this is what I'm seeing:
```
*** 0: {traceId:1, spanId:3, parentId:2, operationName:"my-custom-method-operation-name"}
*** 1: {traceId:1, spanId:4, parentId:2, operationName:"my-custom-class-operation-name"}
*** 2: {traceId:1, spanId:2, parentId:0, operationName:"GET:/with-custom-operation-name/"}
```
I'm not sure what to make of that yet, but there's the raw data.

bstansberry:
@jasondlee I have no clue!  I will say that as a naive person the 'wrong' logging seems more 'right'!

jasondlee:
@bstansberry And I would agree. :) The more pressing question for me, I think, is why there's a difference in behavior. I guess that's step 1, then we can figure out what the right response should be. :) I suspect maybe the mock tracer, but hopefully I'll know more soon.

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceProviderDefinition.java
```diff
@@ -65,30 +66,10 @@
...
-                    @Override
-                    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
-                        String name = context.getCurrentAddressValue();
-                        providers.unregister(name);
-                    }
-                });
+        super(new SimpleResourceDefinition.Parameters(MicroProfileConfigExtension.CONFIG_SOURCE_PROVIDER_PATH,
+                MicroProfileConfigExtension.getResourceDescriptionResolver(MicroProfileConfigExtension.CONFIG_SOURCE_PROVIDER_PATH.getKey()))
+                .setAddHandler(new ConfigSourceProviderDefinitionAddHandler(providers))
+                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE));
```
jmesnil, line 72: using a `ReloadRequiredRemoveStepHandler.INSTANCE` will not be enough as you would also need to remove the current instance from the providers:

```java
new AbstractRemoveStepHandler() {		
                     @Override		
                     protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {		
                         String name = context.getCurrentAddressValue();		
                         providers.unregister(name);		
                     }		
                 });
```

jmesnil (changes requested):

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceProviderDefinition.java
bstansberry: It sounds like we are missing an MSC service to do the registration/unregistration then, so the service stop does a remove during a reload.

jmesnil: Brian is right, we need to have 3 more MSC service to take care of:
* ConfigSource loaded from class
* ConfigSource from Properties
* ConfigSourceProvide

All of them will be pretty similar to https://github.com/wildfly/wildfly/blob/master/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/DirConfigSourceRegistrationService.java

jasondlee: The reason all of that code was removed was because Brian said that if a reload is required, the handler shouldn't be changing the runtime state. Am I misunderstanding something here?

jasondlee: I talked to Brian and have a better understanding of what's going on here. Change is in progress...

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ClassConfigSourceRegistrationService.java
```diff
@@ -0,0 +1,69 @@
...
+                .addService(ServiceNames.CONFIG_SOURCE.append(name));
+
+        builder.setInstance(new ClassConfigSourceRegistrationService(name, configSource, registry))
+                .install();
+    }
+
+    @Override
+    public void start(StartContext startContext) {
+        MicroProfileConfigLogger.ROOT_LOGGER.loadConfigSourceFromClass(configSource.getClass());
+        this.sources.register(this.name, configSource);
```
bstansberry, line 62: This comment applies to all these services, and if @jmesnil thinks what you have here is fine, then that's good enough...

Doing the register call in the add handler might have been fine; the key thing was doing the unregister in the service stop.  Doing it in the service start subtly shifts the timing. I don't understand how these are used enough to know whether that's a possible problem. My uninformed guess is it's not a problem.

Doing it the way you do here is certainly more understandable though, start does X and stop undoes X, so if there's no concern from Jeff, it's preferable to do it this way.

jasondlee: I like the symmetry, encapsulation, whatever-other-buzz-word-might-be-appropriate, etc., but I'll certainly defer judgment to you or Jeff. :)

jmesnil: using a service to do the registration/unregistration is cleaner.

jasondlee:
The failure on Windows should now be resolved. For the record, it stems from the need to reload the server when removing ConfigSources. One test set a property that broke things for the test in question. That we didn't see it on Linux is, I think, a lucky coincidence.

PR #13987, opened 2021-02-12
[WFLY-14424] Test case for EJB client interacting with two clusters.

rachmatowicz:
This PR does the following:
- adds a test case to the Wildfly clustering testsuite to test EJB client interaction with two non-overlapping clusters

For more info, see: https://issues.redhat.com/browse/WFLY-14424

### DIFF
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -671,6 +671,7 @@
                                         <!-- These are run by ts.clustering.cluster.ha-infinispan-server.profile -->
                                         <exclude>${test-group}/**/web/remote/*TestCase.java</exclude>
                                         <exclude>${test-group}/**/sso/remote/*TestCase.java</exclude>
+                                        <exclude>${test-group}/**/ejb/remote/RemoteEJBClientTwoClustersTestCase.java</exclude>
                                         <!-- These are run by ts.clustering.cluster.byteman.profile -->
                                         <exclude>${test-group}/**/byteman/*TestCase.java</exclude>
                                         <exclude>%regex[${test-group}/(${ts.surefire.clustering.ha.additionalExcludes})/.*TestCase.class]</exclude>
@@ -737,6 +738,70 @@
                                     <includes>
                                         <include>${test-group}/**/web/remote/*TestCase.java</include>
                                         <include>${test-group}/**/sso/remote/*TestCase.java</include>
+                                    </includes>
+                                    <!-- Parameters to test cases. -->
+                                    <systemPropertyVariables>
+                                        <!-- required by arquillian.xml -->
+                                        <wildfly1>wildfly-1</wildfly1>
+                                        <wildfly2>wildfly-2</wildfly2>
+                                        <wildfly3>wildfly-3</wildfly3>
+                                        <wildfly4>wildfly-4</wildfly4>
+                                        <wildfly-load-balancer1>wildfly-load-balancer-1</wildfly-load-balancer1>
+                                        <!-- end required by arquillian.xml -->
+                                        <node0>${node0}</node0>
+                                        <node1>${node1}</node1>
+                                        <node2>${node2}</node2>
+                                        <node3>${node3}</node3>
+                                        <mcast>${mcast}</mcast>
+                                        <mcast1>${mcast1}</mcast1>
+                                        <mcast2>${mcast2}</mcast2>
+                                        <mcast3>${mcast3}</mcast3>
+                                        <arquillian.xml>arquillian.xml</arquillian.xml>
+                                        <arquillian.launch>clustering-all</arquillian.launch>
+                                        <jboss.server.config.file.name>standalone-ha.xml</jboss.server.config.file.name>
+                                        <!-- Override ${server.jvm.arg} to only include ${jvm.args.ip} but *not* include ${jvm.args.ip.server} -->
+                                        <server.jvm.args>${surefire.system.args} ${jvm.args.ip} ${jvm.args.other} ${jvm.args.timeouts} ${jvm.args.dirs} ${extra.server.jvm.args} -Dnode0=${node0} -Dnode1=${node1} -Dnode2=${node2} -Dnode3=${node3}</server.jvm.args>
+                                        <infinispan.server.home>${infinispan.server.home.template}</infinispan.server.home>
+                                        <infinispan.server.profile>${infinispan.server.profile.override}</infinispan.server.profile>
+                                    </systemPropertyVariables>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <!-- Run cluster tests with standalone-ha.xml profile with reuseForks == false -->
+        <profile>
+            <id>ts.clustering.cluster.ha-no-reuse-forks.profile</id>
+            <activation>
+                <property>
+                    <name>!ts.noClustering</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <version>${version.surefire.plugin}</version>
+                        <executions combine.children="append">
+                            <!-- Multi-node clustering tests -->
+                            <execution>
+                                <id>ts.surefire.clustering.ha-no-reuse-forks</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <!-- Fix test order to speed up execution by grouping tests that run against same group of containers -->
+                                    <runOrder>alphabetical</runOrder>
+                                    <!-- Don't reuse forked JVMs - required for EJB client tests which define custom client side configurations  -->
+                                    <reuseForks>false</reuseForks>
+                                    <!-- Tests to execute. -->
+                                    <includes>
+                                        <include>${test-group}/**/ejb/remote/RemoteEJBClientTwoClustersTestCase.java</include>
                                     </includes>
                                     <!-- Parameters to test cases. -->
                                     <systemPropertyVariables>
@@ -1756,6 +1821,7 @@
                                         <exclude>${test-group}/ejb/forwarding/*TestCase.java</exclude>
                                         <exclude>${test-group}/ejb/remote/GlobalAuthContextRemoteStatelessEJBFailoverTestCase.java</exclude>
                                         <exclude>${test-group}/ejb/remote/ThreadAuthContextRemoteStatelessEJBFailoverTestCase.java</exclude>
+                                        <exclude>${test-group}/ejb/remote/RemoteEJBClientTwoClustersTestCase.java</exclude>
                                     </excludes>
                                     <!-- Parameters to test cases. -->
                                     <systemPropertyVariables>
@@ -2404,6 +2470,8 @@
                                         <!-- These are run by ts.clustering.cluster.byteman.profile -->
                                         <exclude>${test-group}/**/byteman/*TestCase.java</exclude>
                                         <exclude>%regex[${test-group}/(${ts.surefire.clustering.ha.additionalExcludes})/.*TestCase.class]</exclude>
+                                        <!-- These are run by ts.clustering.cluster.ha-infinispan-server.profile -->
+                                        <exclude>${test-group}/**/ejb/remote/RemoteEJBClientTwoClustersTestCase.java</exclude>
 
                                         <!-- TODO see why these aren't working and fix or exclude with a clear explanation -->
                                         <exclude>${test-group}/ejb/remote/GlobalAuthContextRemoteStatelessEJBFailoverTestCase.java</exclude>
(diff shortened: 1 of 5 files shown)


rachmatowicz:
retest please

rachmatowicz:
There is a problem with the targets -Dts.layers and -Dts.standalone.microprofile which I will have to look at.

rachmatowicz:
Omitted renaming some execution instance names in the -Dts.layers and -Dts.standalone.microprofile profiles. Should be OK now.

rachmatowicz:
retest please

bstansberry:
I just stopped CI on this to redirect resources to more time-critical jobs. This one is not required for WF 23 Beta so it's lower priority until that's tagged.

rachmatowicz:
ready to merge! :-)

pferraro:
@rhusar Please review.

darranl:
Hi @rhusar are you able to give this one a quick review?  Looks like we should be able to get this one through.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/remote/RemoteEJBClientTwoClustersTestCase.java
```diff
@@ -0,0 +1,380 @@
...
+                    // distinct JChannel name for each node in cluster A
+                    .setup("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=clusterA)")
+                    .teardown("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=ejb)")
+                    .parent()
+                    // clusterB nodes
+                    .node(NODE_3, NODE_4)
+                    // distinct JChannel name in cluster B for each node as well as unique discovery address
+                    .setup("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=clusterB)")
+                    .setup("/socket-binding-group=standard-sockets/socket-binding=jgroups-mping:write-attribute(name=multicast-address,value=%s)", TESTSUITE_MCAST3)
+                    .teardown("/socket-binding-group=standard-sockets/socket-binding=jgroups-mping:write-attribute(name=multicast-address,value=%s", TESTSUITE_MCAST)
```
rhusar, line 375: There's a missing closing bracket in the `:write-attribute(..)` operation, but the client tolerates this.

rhusar (approved):
Looks good to me (with 2 comments).

testsuite/integration/clustering/pom.xml
```diff
@@ -604,9 +605,9 @@
             </build>
         </profile>
 
-        <!-- Run cluster tests with standalone-ha.xml profile with Infinispan Server provisioned -->
+        <!-- Run cluster tests with standalone-ha.xml profile with reuseForks == false -->
         <profile>
-            <id>ts.clustering.cluster.ha-infinispan-server.profile</id>
+            <id>ts.clustering.cluster.ha-noReuseForks.profile</id>
```
rhusar, line 610: The rest of the id-s are standardised using dash-separated-names instead of CamelCase. Do you want to rename these to minimise confusion?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/remote/RemoteEJBClientTwoClustersTestCase.java
rachmatowicz: Done.

testsuite/integration/clustering/pom.xml
rachmatowicz: Done.

rhusar:
@rachmatowicz It looks like this might be causing regression (intermittent?) in the https://ci.wildfly.org/buildConfiguration/WF_PullRequest_GalleonLinuxJdk11/251775? TransactionalRemoteStatelessTestCase - can you check if that's related?

rachmatowicz:
> @rachmatowicz It looks like this might be causing regression (intermittent?) in the https://ci.wildfly.org/buildConfiguration/WF_PullRequest_GalleonLinuxJdk11/251775 - TransactionalRemoteStatelessTestCase - can you check if that's related?

In fact for each test run failure, there is a different set of test cases which fail:
TransactionRemoteStatelessTestCase on galleon Linux
DatabaseTimerServiceMultiNodeExecutionDisabledTestCase on Linux with security manager
InvalidCharactersTestCase on Windows JDK 11

rachmatowicz:
SecurityAuditingTestCase: fails consistently across all platforms, except Galleon Linux
TransactionReoteStatelessTestCase, RemoteEJBClientTwoClustersTestCase: fails on Galleon Linux but passes on all other platforms

rachmatowicz:
My tests pass on all platforms except Galleon Linux. What is Galleon Linux? Can I reproduce it locally?

rhusar:
@rachmatowicz You just add the `-Dts.layers`.

rachmatowicz:
> @rachmatowicz Yes – you just add the `-Dts.layers`.

Thanks. I think the TransactionRemoteStatelessTestCase failure is due to this: https://issues.redhat.com/browse/WFCC-33.

rachmatowicz:
The problem was that there is a target ts.surefire.clustering.ejb which is running a mix of EJB client test cases. The EJB client codebase depends on a lot of static initializers for contexts such as Discovery, EJBClientContext, AuthenticationContext, etc. When a test requires a different initializer for one of these static contexts, and they share the same JVM, that's where the trouble begins.Tests which require a different static initialization need to run in a separate JVM. I had to exclude my test from ts.surefire.clustering.ejb as it is one of those tests which requires a different static configuration.

darranl:
@rhusar Are you still happy with the final changes?  I think we can get this one through.

testsuite/integration/clustering/pom.xml
```diff
@@ -2306,7 +2309,7 @@
                                 </configuration>
                             </execution>
                             <execution>
-                                <id>ts.surefire.clustering.ha-infinispan-server</id>
+                                <id>ts.surefire.clustering.ha-no-reuse-forks</id>
```
bstansberry, line 2312: @rachmatowicz This test doesn't pass for me with -Dts.ee9 so it needs to be added to the exclude list a bit below this.

I think the new test has the EJB client outside the appserver and that's not likely to work with what we have with EE 9 at the moment, because any jakarta.* namespace classes that get sent from the server to the client won't be understood by the client, which has javax.* namespace classes.

bstansberry (changes requested):
@rachmatowicz Please see my previous comment.

testsuite/integration/clustering/pom.xml
rachmatowicz: I've added it to the exclude list, i'll push the changes. Not sure about your last comment; there are many tests which use EJB client from a standalone test case interacting with a server. Is this going to be a general problem?

bstansberry: @rachmatowicz Sorry, I didn't get back to this for too long.

There are two aspects to my answer.

First, if we convert the codebase to jakarta namespace then this will no longer be a problem.

But, an important task for WF 26 is figuring out how to deal with remote EE 8 EJB clients accessing an EE > 8 server, as propagating java types to clients is part of remote EJB. This is something you, @chengfang, @tadamski, @ropalka and I need to talk about.

rachmatowicz:
The issue was that the new test needed to run in the "ha-no-reuse-forks" target, as opposed to the regular "ha" target. The original PR excluded the test from the "ha" target, but did not do so for the ts.ee9 profile, which caused the failures when running with -Dts.ee9. This has now been fixed.

testsuite/integration/clustering/pom.xml
```diff
@@ -708,9 +709,9 @@
             </build>
         </profile>
 
-        <!-- Run cluster tests with standalone-ha.xml profile with Infinispan Server provisioned -->
+        <!-- Run cluster tests with standalone-ha.xml profile with reuseForks == false -->
         <profile>
-            <id>ts.clustering.cluster.ha-infinispan-server.profile</id>
+            <id>ts.clustering.cluster.ha-no-reuse-forks.profile</id>
```
rhusar, line 714: If we are renaming this profile, the documentation at `docs/src/main/asciidoc/_testsuite/WildFly_Testsuite_Overview.adoc` should be updated accordingly as well.

rhusar (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/remote/RemoteEJBClientTwoClustersTestCase.java
```diff
@@ -0,0 +1,380 @@
...
+                    // distinct JChannel name for each node in cluster A
+                    .setup("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=clusterA)")
+                    .teardown("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=ejb)")
+                    .parent()
+                    // clusterB nodes
+                    .node(NODE_3, NODE_4)
+                    // distinct JChannel name in cluster B for each node as well as unique discovery address
+                    .setup("/subsystem=jgroups/channel=ee:write-attribute(name=cluster,value=clusterB)")
+                    .setup("/socket-binding-group=standard-sockets/socket-binding=jgroups-mping:write-attribute(name=multicast-address,value=%s)", TESTSUITE_MCAST3)
+                    .teardown("/socket-binding-group=standard-sockets/socket-binding=jgroups-mping:write-attribute(name=multicast-address,value=%s)", TESTSUITE_MCAST)
```
rhusar, line 375: You can just drop these - this isn't doing anything. The test is using the `ee` channel which is using `stack="tcp"`, which doesn't use MPING. You are separating clusters by setting a separate cluster name, not by multicast groups.

rhusar (changes requested):

testsuite/integration/clustering/pom.xml
rachmatowicz: We discussed this at the time, agreeing on the new name, and it was my understanding that this target was distinguished by running with no reuse forks and not some Infinispan class rule as mentioned in the documentation. If this has changed, i'm not going to rename it. Also, I don't see difference between this target and the full-ha target below it.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/remote/RemoteEJBClientTwoClustersTestCase.java
rachmatowicz: Fixed. Thanks for noting.

testsuite/integration/clustering/pom.xml
rhusar: @rachmatowicz I was just talking about updating the documentation, but you actually raised a more important point!

Yes, what you are describing is what we discussed at that time - basically I suggested to reuse and rename that profile which was using <reuseForks>false</reuseForks>. However, this has changed since, and as of 044c1532d0946a50905bcfddf2dffa755d5dd329 this workaround was removed. 

Looking more closely, you aren't actually putting that <reuseForks>false</reuseForks> back, so your test is passing even with reuse forks enabled! If this is the case, you can just put the test back to the `ts.clustering.cluster.ha` profile. Right?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/remote/RemoteEJBClientTwoClustersTestCase.java
rhusar: Still gotta push :)

testsuite/integration/clustering/pom.xml
rhusar: @rachmatowicz So, this might be working because in that execution, there are no other EJB client tests, thus the static EJB context is not present and the tests pass. So if you cannot add this onto other executions, I am going to ask you to create a new one. The `ha-infinispan-server` is now relied on to run the RHDG tests so let's not rename it.

rhusar (changes requested):
Requesting changes: https://github.com/wildfly/wildfly/pull/13987#discussion_r763163357

testsuite/integration/clustering/pom.xml
rachmatowicz: It does not work if I include in the ts.clustering.cluster.ha profile. I would prefer not to add another profile just for one test - on the other hand, this test does not play well with other EJB tests and so merits a reuse forks property. Thinking.

rhusar: @rachmatowicz Neither do I, but adding `<reuseForks>false</reuseForks>` to any of the existing test cases increasing testsuite runtime by minutes; the price is managing one extra profile.

darranl:
@rachmatowicz does your latest push mean this is ready for re-review?

fjuma:
@rachmatowicz Just wanted to check on the status of this PR. Looks like there's a conflict that needs to be resolved. Are any other changes still pending too?

pferraro:
@rachmatowicz We are in the process of cleaning up the PR queue and are going to close this for now.  Let's revisit this branch during the reassessment of the clustering testsuite.

PR #13988, opened 2021-02-12
WFLY-14430 Singleton lifecycle transition throws wrong exception type if target service was removed

pferraro:
https://issues.redhat.com/browse/WFLY-14430

This is one of the causes (if not *the* cause) of intermittent failures in singleton deployment tests.


rhusar (approved):
🎉

PR #13989, opened 2021-02-12
[WFLY-14399] Ensure the invocation of the common scripts are executed…

jamezp:
… in each script. Also add tests to ensure this.

https://issues.redhat.com/browse/WFLY-14399

Requires https://github.com/wildfly/wildfly-core/pull/4472
Analysis Doc: https://github.com/wildfly/wildfly-proposals/pull/369

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/General_configuration_concepts.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/General_configuration_concepts.adoc
@@ -213,3 +213,44 @@ element, the value in a `host.xml` taking precedence over anything from
 `domain.xml`, and a value in a `<server-group/>` element taking
 precedence over a value specified directly under the `domain.xml` root
 element.
+
+== Script Configuration Files
+
+Scripts are located in the `$JBOSS_HOME/bin` directory. Within this directory you will find script configuration files
+for standalone and domain startup scripts for each platform. These files can be used to configure your environment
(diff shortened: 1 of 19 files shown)


docs/src/main/asciidoc/_admin-guide/General_configuration_concepts.adoc
```diff
@@ -213,3 +213,40 @@ element, the value in a `host.xml` taking precedence over anything from
...
+`$JBOSS_HOME/bin` directory. While these configuration files are not present in the directory by default, they
+can be added. You can simply just add the `common.conf` configuration file for the script type you want to
+execute and all scripts in the directory will invoke the configuration script.
+
+* `common.conf` for bash scripts
+* `common.conf.bat` for Windows batch scripts
+* `common.conf.ps1` for PowerShell scripts
+
+You can also set the `COMMON_CONF` environment variable to have this configuration script live outside the
+`$JBOSS_HOME/bin` directory.
```
bstansberry, line 252: Possible additional paragraph:

Note that if you provide both a `common.conf` configuration file and a `standalone.conf` or `domain.conf` file, the standalone or domain startup scripts will call both files, first calling [whichever is first] and then calling [the other one].

jamezp: That makes sense to add. Thank you!

bstansberry (approved):

PR #13992, opened 2021-02-15
[WFLY-13043] Fix error propagation for scheduled tasks

baranowb:
ISSUE: https://issues.redhat.com/browse/WFLY-13043

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
+++ b/ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
@@ -172,10 +172,14 @@ public class ControlPointUtils {
             if (controlPoint == null) {
                 runnable.run();
             } else {
+                RuntimeException runnableException = null;
                 try {
                     if (controlPoint.beginRequest() == RunResult.RUN) {
                         try {
                             runnable.run();
+                        } catch (RuntimeException e) {
+                            runnableException = e;
+                            throw e;
                         } finally {
                             controlPoint.requestComplete();
                         }
@@ -183,11 +187,13 @@ public class ControlPointUtils {
                         throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(runnable);
                     }
                 } catch (RuntimeException re) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(re);  // note that this is now a DEBUG message as we rethrow re
-                    throw re;
-                } catch (Exception e) { // TODO when WFCORE-6153 is done, drop this block. The actual try code only throws RuntimeExceptions
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);  // note that this is now a DEBUG message as we rethrow e
-                    throw new RuntimeException(e);
+                    // WFLY-13043
+                    if (runnableException == null) {
+                        EeLogger.ROOT_LOGGER.failedToRunTask(runnable,re);
+                        return;
+                    } else {
+                        throw EeLogger.ROOT_LOGGER.failureWhileRunningTask(runnable, re);
+                    }
                 }
             }
         }
@@ -223,8 +229,7 @@ public class ControlPointUtils {
                         throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(callable);
                     }
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e); // note that this is now a DEBUG message as we rethrow e
-                    throw e;
+                    throw EeLogger.ROOT_LOGGER.failureWhileRunningTask(callable, e);
                 }
             }
         }
--- a/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
+++ b/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
@@ -23,7 +23,6 @@
 package org.jboss.as.ee.logging;
 
 
-import static org.jboss.logging.Logger.Level.DEBUG;
 import static org.jboss.logging.Logger.Level.ERROR;
 import static org.jboss.logging.Logger.Level.INFO;
 import static org.jboss.logging.Logger.Level.WARN;
@@ -1087,9 +1086,9 @@ public interface EeLogger extends BasicLogger {
     @Message(id = 109, value = "A class must not declare more than one AroundInvoke method. %s has %s methods annotated.")
     DeploymentUnitProcessingException aroundInvokeAnnotationUsedTooManyTimes(DotName className, int numberOfAnnotatedMethods);
 
-    @LogMessage(level = DEBUG) // this is only used in a log-and-throw so make it DEBUG
-    @Message(id = 110, value = "Failed to run scheduled task")
-    void failedToRunTask(@Cause Exception e);
+    @LogMessage(level = ERROR)
+    @Message(id = 110, value = "Failed to run scheduled task: %s")
+    void failedToRunTask(Object delegate, @Cause Exception e);
 
     @Message(id = 111, value = "Cannot run scheduled task %s as container is suspended")
     IllegalStateException cannotRunScheduledTask(Object delegate);
@@ -1227,4 +1226,8 @@ public interface EeLogger extends BasicLogger {
     @LogMessage(level = WARN)
     @Message(id = 135, value = "Failed to resume transaction.")
     void failedToResumeTransaction(@Cause Throwable cause);
+
+    @Message(id = 136, value = "Failed to run scheduled task: %s")
+    RuntimeException failureWhileRunningTask(Object delegate,@Cause Exception e);
+
 }
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/ManagedScheduledExecutorExceptionTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/ManagedScheduledExecutorExceptionTestCase.java
@@ -46,6 +46,7 @@ public class ManagedScheduledExecutorExceptionTestCase {
                 .addClasses(ManagedScheduledExecutorExceptionTestCase.class);
     }
 
+    private static final String loggerMessage = "WFLYEE0136";
     private static final String message = "WFLY-17186";
 
     private static void badMethod() {
@@ -74,7 +75,8 @@ public class ManagedScheduledExecutorExceptionTestCase {
             future.get();
             fail("Exception did not propagate");
         } catch (ExecutionException e) {
-            assertTrue(e.toString(), e.getCause().getMessage().contains(message));
+            assertTrue(e.toString(), e.getCause().getMessage().contains(loggerMessage));
+            assertTrue(e.toString(), e.getCause().getCause().getMessage().contains(message));
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             fail("Caught " + e);
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/scheduled/ExceptionInScheduleTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/scheduled/ExceptionInScheduleTestCase.java
@@ -1,0 +1,61 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.test.integration.ee.concurrent.scheduled;
+
+import jakarta.ejb.EJB;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Test case for managed executor's hung tasks termination feature.
+ *
+ * @author baranowb
+ */
+@RunWith(Arquillian.class)
+public class ExceptionInScheduleTestCase {
+
+    @EJB(lookup = "java:module/ExecFlawedTaskBean!org.jboss.as.test.integration.ee.concurrent.scheduled.ExecNumber")
+    private ExecNumber execNumber;
+
+    @Deployment
+    public static Archive<?> deploy() {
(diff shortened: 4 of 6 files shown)


baranowb:
Not ideal fix, but...

pferraro (changes requested):
@baranowb Please include a test that validates the intended behavior.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -184,8 +188,15 @@ public void run() {
                         throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(runnable);
                     }
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);
+                    // WFLY-13043
```
pferraro, line 191: I don't think any of this exception handling logic is necessary.  If you just get rid of this try/catch block entirely, then any runtime exceptions from run() will propagate to the underlying executor service.

pferraro (changes requested):
@baranowb Please include a test that validates the intended behavior.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
baranowb: It would, if for some reason beginRequest() did not indicate it throws Exception ( which it does not in any part of impl). Yeah, I dont like this code, but unless controlPoint change API it has to be like that.

darranl:
@pferraro Are you able to re-review this one?  A test case has been added and @baranowb responded to your comment about the exception handling.

pferraro:
@darranl Will do.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
pferraro: Agreed - ControlPoint has an unfortunate method signature...
However, you can still simplify this code to be more readable, e.g.

```java
try {
    if (controlPoint.beginRequest() == RunResult.RUN) {
        try {
            runnable.run();
        } catch (RuntimeException e) {
            throw new ExecutionException(e);
        } finally {
            controlPoint.requestComplete();
        }
    }
} catch (ExecutionException e) {
    throw EeLogger.ROOT_LOGGER.failureWhileRunningTask(runnable, e.getCause());
} catch (Exception e) {
    EeLogger.ROOT_LOGGER.failedToRunTask(runnable,e);
}
```

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -217,7 +228,7 @@ public T call() throws Exception {
                         }
                     }
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);
+                    EeLogger.ROOT_LOGGER.failedToRunTask(callable,e);
```
pferraro, line 231: Exceptions thrown by Callable.call() also need to be propagated to the caller.  Thus the same treatment as above is needed here.

pferraro (changes requested):

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
baranowb: It is, two lines bellow: 
 } catch (Exception e) {
                    EeLogger.ROOT_LOGGER.failedToRunTask(callable,e);
                }
                throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(callable);

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
baranowb: I guess, though I just dont like to re-throw exception few lines down the pipe

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
pferraro: There are two problems with that handling:

1. It generates an incorrect exception message.
2. It obfuscates the exception thrown by the caller.

e.g. If the control point allows execution, but the subsequent invocation of callable.call() throws an exception, the ControlledScheduledCallable.call() method will throw a wrapped IllegalStateException to the caller containing the message "Cannot run scheduled task %s as container is suspended".  The message is not correct, since the container was *not* suspended (1).  Additionally, the checked exception thrown by callable.call() is incorrectly wrapped as a IllegalStateException (2).

pferraro (changes requested):

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
baranowb: #1 - how can it now generate incorrect exception? This code was there before, only change is another param, to point what kind of task failed.
#2 - again, this has been there way before, has nothing to do with this PR.

pferraro: 1. I explained this in my previous message.  See the e.g. for details.
2. This PR aims to fix error propagation for scheduled tasks, does it not?  Thus we should address both run() and call() scenarios.

baranowb: Sort of, but only for ControlledScheduledRunnable. I got tad thrown away by: Additionally, the checked exception thrown by callable.call() is incorrectly wrapped as a IllegalStateException (2) - since it isnt. Original exception is being logged and new one is thrown, though with wrong message.

Something like: 
try {
                    if (controlPoint.beginRequest() == RunResult.RUN) {
                        try {
                            return callable.call();
                        } finally {
                            controlPoint.requestComplete();
                        }
                    }
                } catch (Exception e) {
                    EeLogger.ROOT_LOGGER.failedToProcureCallableResult(callable,e);
                    throw e;
                }
                throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(callable);
                
should be enough?

pferraro: > Original exception is being logged and new one is thrown

This is what I mean by "incorrectly wrapped as an ISE".  Callable.call() should throw the unadulterated checked exception - it should not wrap it as a runtime exception.

Any exception thrown by ControlPoint *should* be wrapped as a runtime exception, but any exception thrown by Callable.call() should be rethrown to the caller without wrapping.  For the latter, we do not want a redundant log message.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -171,21 +172,23 @@ public T call() throws Exception {
...
+                            throw new ExecutionException(e);
                         } finally {
                             controlPoint.requestComplete();
                         }
-                        return;
-                    } else {
-                        throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(runnable);
                     }
+                } catch (ExecutionException e) {
+                    throw EeLogger.ROOT_LOGGER.failureWhileRunningTask(runnable, (Exception)e.getCause());
```
pferraro, line 187: Formatting inconsistency - use whitespace between cast and target.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -171,21 +172,23 @@ public T call() throws Exception {
...
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);
+                    EeLogger.ROOT_LOGGER.failedToRunTask(runnable,e);
```
pferraro, line 189: Formatting inconsistency - use whitespace between comma and second argument

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -212,12 +215,18 @@ public T call() throws Exception {
...
                             return callable.call();
+                        } catch (Exception e) {
+                            throw new ExecutionException(e);
                         } finally {
                             controlPoint.requestComplete();
                         }
                     }
+                } catch(ExecutionException e) {
+                    //from callable
+                    throw (Exception)e.getCause();
```
pferraro, line 226: Formatting inconsistency - use whitespace between cast and target.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -212,12 +215,18 @@ public T call() throws Exception {
...
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);
+                    //from controlPoint.beginRequest(), Callable mandates exception
+                    throw EeLogger.ROOT_LOGGER.failedToProcureCallableResult(callable,e);
```
pferraro, line 229: Formatting inconsistency - use whitespace between comma and second argument

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -1204,4 +1204,11 @@
     @LogMessage(level = INFO)
     @Message(id = 130, value = "%s hung task %s not cancelled")
     void hungTaskNotCancelled(String executorName, String taskName);
+
+    @Message(id = 131, value = "Failed to run scheduled task: %s")
+    RuntimeException failureWhileRunningTask(Object delegate,@Cause Exception e);
```
pferraro, line 1209: Formatting inconsistency - use whitespace between comma and second argument

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -212,12 +215,18 @@ public T call() throws Exception {
                     if (controlPoint.beginRequest() == RunResult.RUN) {
                         try {
                             return callable.call();
+                        } catch (Exception e) {
+                            throw new ExecutionException(e);
                         } finally {
                             controlPoint.requestComplete();
                         }
                     }
+                } catch(ExecutionException e) {
```
pferraro, line 224: Formatting inconsistency - use whitespace to better distinguish betwee control statements and method calls.

pferraro (changes requested):
My only remaining comments are related to code formatting.
Otherwise, this looks good.

baranowb:
Sorry, was on PTO. Fixed.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -212,12 +215,18 @@ public T call() throws Exception {
...
                             controlPoint.requestComplete();
                         }
                     }
+                } catch (ExecutionException e) {
+                    // from callable
+                    throw (Exception) e.getCause();
                 } catch (Exception e) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(e);
+                    // from controlPoint.beginRequest(), Callable mandates exception
+                    throw EeLogger.ROOT_LOGGER.failedToProcureCallableResult(callable, e);
```
pferraro, line 229: This should use failureWhileRunningTask(...).
We don't want to throw a checked exception if ControlPoint.beginRequest() threw an exception.

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -1204,4 +1204,11 @@
     @LogMessage(level = INFO)
     @Message(id = 130, value = "%s hung task %s not cancelled")
     void hungTaskNotCancelled(String executorName, String taskName);
+
+    @Message(id = 131, value = "Failed to run scheduled task: %s")
+    RuntimeException failureWhileRunningTask(Object delegate, @Cause Exception e);
+
+    @Message(id = 132, value = "Failed to procure result, container failed: %s")
+    Exception failedToProcureCallableResult(Object delegate, @Cause Exception e);
```
pferraro, line 1212: I don't think this is needed.

pferraro (changes requested):

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
baranowb: Thats Callable contract.

pferraro: We only want to propagate a checked exception if the underlying Callable.call() throws a checked exception.  If this method were to throw a checked exception due to ControlPoint.beginRequest() failure - that would not comply with the Callable contract.

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -1204,4 +1204,8 @@
     @LogMessage(level = INFO)
     @Message(id = 130, value = "%s hung task %s not cancelled")
     void hungTaskNotCancelled(String executorName, String taskName);
+
+    @Message(id = 131, value = "Failed to run scheduled task: %s")
+    RuntimeException failureWhileRunningTask(Object delegate, @Cause Exception e);
```
pferraro, line 1209: This should probably throw an IllegalStateException, instead of a generic RuntimeException.

pferraro (changes requested):

darranl:
@baranowb are you able to continue this one?  It seems like we reached the end of review comments and just need to finish off CI.

fjuma:
@baranowb Have you had a chance to investigate the test failures for this one? There's also a conflict that needs to be addressed.

fjuma:
@pferraro Did you want to re-review this one? CI is now passing for this PR.

fjuma:
@pferraro Actually, ignore that last comment, just realized that all the jobs haven't been run yet.

baranowb:
Hmm, there was conflict. Oddly enough I see single request, but I just dont see where. Is it still valid?

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -172,22 +172,36 @@ public void run() {
...
                             controlPoint.requestComplete();
                         }
                     } else {
                         throw EeLogger.ROOT_LOGGER.cannotRunScheduledTask(runnable);
                     }
                 } catch (RuntimeException re) {
-                    EeLogger.ROOT_LOGGER.failedToRunTask(re);  // note that this is now a DEBUG message as we rethrow re
-                    throw re;
+                    // WFLY-13043
+                    EeLogger.ROOT_LOGGER.failedToRunTask(runnable,re);
```
bstansberry, line 191: If we're not going to always rethrow and we're wrapping re when we do rethrow, then I think eLogger.ROOT_LOGGER.failedToRunTask(runnable,re); should be moved into the 'if (runnableException == null) {' and logged at WARN or ERROR, not DEBUG.

testsuite/integration/basic/pom.xml
```diff
@@ -3069,6 +3069,7 @@
                                     <excludes>
                                         <!-- JpaJarFileTestCase requires custom hibernate module -->
                                         <exclude>org/jboss/as/test/integration/jpa/jarfile/JpaJarFileTestCase.java</exclude>
+                                        <exclude>org/jboss/as/test/integration/ee/concurrent/scheduled/ExceptionInScheduleTestCase.java</exclude>
```
bstansberry, line 3072: This test shouldn't need to be excluded. If it is excluded it needs a comment explaining why.

Perhaps remove the @Remote from ExecNumber? The test runs in container and shouldn't need @Remote.

ee/src/main/java/org/jboss/as/ee/concurrent/ControlPointUtils.java
```diff
@@ -172,22 +172,36 @@ public void run() {
...
+                    if (runnableException == null) {
+                        return;
+                    } else {
+                        throw EeLogger.ROOT_LOGGER.failureWhileRunningTask(runnable, re); // note that this is now a DEBUG message as we rethrow e
+                    }
                 } catch (Exception e) { // TODO when WFCORE-6153 is done, drop this block. The actual try code only throws RuntimeExceptions
```
bstansberry, line 197: WFCORE-6153 is now done and merged, so please just drop this entire catch block.

bstansberry (changes requested):
Thanks @baranowb -- please ping me or @pferraro or @fjuma when updated so we can merge. It's far enough back in the queue that it's likely to fall in a crack otherwise.

bstansberry (approved):

bstansberry:
Thanks @baranowb

PR #14020, opened 2021-02-22
Revert "[WFLY-14439] Move org.apache.thrift module from wildfly-ee to microprofile"

bstansberry:
Reverts wildfly/wildfly#14010

@jasondlee I'm reverting this because it should go in the MP4 branch or waiting until that branch is merged.


jasondlee:
So do I need to open a new PR targeting MP4?

bstansberry:
@jasondlee Unless @kabir says otherwise best is to wait until MP4 is merged, and then do it against master. I think MP4 branch is likely in a phase where it's best to leave it alone.

jasondlee:
Fair enough. I'll sit tight and open a new PR once MP4 is merged. I will, though, reject this PR on the JIRA.

PR #14029, opened 2021-02-23
WFLY-14467 Upgrade mod_cluster to 1.4.3.Final

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-14467


darranl:
I have re-triggered the two failed jobs.

One was a clustering related failure but the other appears to be a timeout, looking at the thread dumps in the log the tests were still running as normal.

bstansberry (approved):

PR #14034, opened 2021-02-23
[WFLY-14439] Move apache thrift module from wildfly-ee to microprofile

jasondlee:
https://issues.redhat.com/browse/WFLY-14439

Move org.apache.thrift module from wildfly-ee to microprofile


bstansberry:
@jasondlee Needs another rebase

I don't think we'll be breaking this further.

jasondlee:
@bstansberry Rebased and pushed. Thanks.

PR #14035, opened 2021-02-24
WFLY-14418 Add tests to batch-jberet subsystem for attributes that al…

chengfang:
…low expression

WFLY-14477 AttributeParser in batch subsystem should use a ModelNode as the value of an operation

JIRA:
https://issues.redhat.com/browse/WFLY-14418
https://issues.redhat.com/browse/WFLY-14477


chengfang:
Still seeing the "Failed to perform checkout on agent: Cannot find commit" error.  Restarting one job "Linux JDK 11" and see how it goes.

PR #14042, opened 2021-02-24
[WFLY-14484] Upgrade JBoss Modules XML schema version to 1.9 for lega…

yersan:
…cy module identifiers that contain a slot

Jira issue: https://issues.redhat.com/browse/WFLY-14484

Notice some of them are not flagged with jboss.api=private, so I'm not sure if we can deal with them bumping the XML NS version.


bstansberry:
@yersan We can change the NS as we please.  If we changed the effective module id for a public module that would be something to care about, but the xml format of a module.xml is an implementation detail.

bstansberry (approved):

PR #14047, opened 2021-02-25
[WFLY-14490] Disable the vendor metrics diagnostic logging from Micro…

bstansberry:
…ProfileVendorMetricsBootTestCase

I considered making this configurable vis a system property but then I'd have to set up security manager perms to allow the prop to be read by the deployment, and... meh.

https://issues.redhat.com/browse/WFLY-14490


rhusar (approved):
+1 these shouldn't be logged, I was wondering where these came from!

PR #14051, opened 2021-02-25
WFLY-14462 exclude-local-receiver and local-receiver-pass-by-value no…

chengfang:
…t in ejb3 subsystem schema

JIRA: https://issues.redhat.com/browse/WFLY-14462


kabir:
@chengfang I don't think it needs to block this PR, but I don't see these attributes tested in the ejb3 subsystem tests

chengfang:
These 2 attributes are tested in https://github.com/wildfly/wildfly/blob/master/ejb3/src/test/java/org/jboss/as/ejb3/subsystem/Ejb3SubsystemUnitTestCase.java#L253, which was recently added. Is there anything else needed?

kabir:
@chengfang thanks - I was lazy and just checked the xml

kabir (approved):

PR #14070, opened 2021-03-03
WFLY-14507 HotRod/DatabasePersistenceWebFailoverTestCase should use n…

rhusar:
…on-transactional caches

Resolves
https://issues.redhat.com/browse/WFLY-14507


rhusar:
@pferraro Actually, this doesn't work, lets look into this..

pferraro (approved):

PR #14075, opened 2021-03-04
WFLY-14474 HHH000431: Unable to determine H2 database version, certain features may not work when using ExampleDS

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-14474


scottmarlow (approved):

bstansberry:
@scottmarlow @parsharma  This is ok, because it's an optional dependency, but this kind of smells funny, i.e. like it is a workaround that happens to be useful because we ship the h2 module.

If there's something about this that makes com.h2database.h2 special (and thus this isn't just a workaround) it would be good to explain what that is in a comment on the JIRA. If it is just a workaround it would be good to note that in a JIRA comment too.

PR #14079, opened 2021-03-05
[WFLY-14524][WFLY-14525] Prevent vendor metric registry modifications…

bstansberry:
… while scrapes are in progress

https://issues.redhat.com/browse/WFLY-14524
https://issues.redhat.com/browse/WFLY-14525

@soul2zimate does this fix the problems?

### DIFF
--- a/metrics/src/main/java/org/wildfly/extension/metrics/MetricRegistry.java
+++ b/metrics/src/main/java/org/wildfly/extension/metrics/MetricRegistry.java
@@ -21,9 +21,37 @@
  */
 package org.wildfly.extension.metrics;
 
+/**
+ * Provides a registry of metrics.
+ */
 public interface MetricRegistry {
 
+    /**
+     * Registers the given metric. Calls will block if another thread
+     * {@link #readLock() holds the registry read lock}.
+     *
+     * @param metric the metric. Cannot be {@code null}
+     * @param metadata metadata for the metric. Cannot be {@code null}
+     */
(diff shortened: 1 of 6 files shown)


soul2zimate:
Thanks for the quick update,  I have manually tested this patch.

For the same scenario with `standalone-microprofile.xml` in [WFLY-14524](https://issues.redhat.com/browse/WFLY-14524), the server can't not properly shutdown. Both server shutdown process side and the `/metrics/` endpoint request loop side stuck with a deadlock scene. I have to kill the server to stop it.

The `ConcurrentModificationException` described in [WFLY-14525](https://issues.redhat.com/browse/WFLY-14525) seems fixed, I don't see the error anymore.

soul2zimate:
I have made a thread dump here https://gist.github.com/soul2zimate/550ff5432f82c18c5d4fef1032d5531e

bstansberry:
@soul2zimate Thanks so much for that!  It was late for me last night to manually test this. I think I fixed the problem and I'll try and manually test it in the next couple days.

I changed the order such that https://github.com/wildfly/wildfly/pull/14079/files#diff-cbdce83e7542d6148696847c354fcac837ee1345f7ae32cee35ed6dd75d28788R83 is after acquiring the lock, not before. A reader thread would first acquire the read lock and then do the synchronization on the registry that your thread dump revealed, so a writer thread should follow the same order.

soul2zimate:
The deadlock doesn't appear anymore with the latest change. :+1:

jmesnil (approved):

PR #14085, opened 2021-03-05
WFLY-14526 Distributable sessions cannot be invalidated via CLI

pferraro:
https://issues.redhat.com/browse/WFLY-14526

Adds an OOB distributed session, which is not (or no longer) associated with any request.


pferraro:
@rhusar Please review.

bstansberry (approved):

PR #14086, opened 2021-03-05
WFLY-14530 InfinispanSessionManager should not expose JNDI contexts to its Infinispan listeners

pferraro:
https://issues.redhat.com/browse/WFLY-14530

### DIFF
--- a/clustering/web/infinispan/src/main/java/org/wildfly/clustering/web/infinispan/session/InfinispanSessionManager.java
+++ b/clustering/web/infinispan/src/main/java/org/wildfly/clustering/web/infinispan/session/InfinispanSessionManager.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


clustering/web/infinispan/src/main/java/org/wildfly/clustering/web/infinispan/session/InfinispanSessionManager.java
```diff
@@ -129,7 +130,7 @@ public void accept(ImmutableSession session) {
 
     @Override
     public void start() {
-        this.executor = new DefaultExecutorService(this.getClass(), ExecutorServiceFactory.CACHED_THREAD);
+        this.executor = Executors.newCachedThreadPool(new DefaultThreadFactory(this.getClass()));
```
bstansberry, line 133: @pferraro Will this result in a meaningfully named thread? If not is this a regression or was it not meaningfully named before?

pferraro: This does not change the way threads are named.  The same thread factory is still used.

bstansberry (approved):

PR #14101, opened 2021-03-10
[WFLY-14546] Restore naming context

spyrkob:
Issue: https://issues.redhat.com/browse/WFLY-14546

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldExecutorServices.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldExecutorServices.java
@@ -24,10 +24,13 @@ package org.jboss.as.weld.services.bootstrap;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
+import org.jboss.as.naming.context.NamespaceContextSelector;
 import org.jboss.as.server.Services;
 import org.jboss.msc.Service;
 import org.jboss.msc.service.ServiceName;
@@ -35,6 +38,8 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.threads.JBossThreadFactory;
+import org.jboss.weld.Container;
+import org.jboss.weld.ContainerState;
 import org.jboss.weld.executor.AbstractExecutorServices;
 import org.jboss.weld.manager.api.ExecutorServices;
 import org.wildfly.security.manager.WildFlySecurityManager;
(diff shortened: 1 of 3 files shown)


weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldExecutorServices.java
```diff
@@ -115,4 +118,52 @@ public synchronized ExecutorService getTaskExecutor() {
...
+        protected void afterExecute(Runnable r, Throwable t) {
+            super.afterExecute(r, t);
+            if (r instanceof WeldTaskWrapper) {
+                NamespaceContextSelector.popCurrentSelector();
+            }
+        }
+
+        @Override
+        public void execute(Runnable command) {
+            WeldTaskWrapper task = new WeldTaskWrapper(command, NamespaceContextSelector.getCurrentSelector());
```
manovotn, line 149: Avoid wrapping if Weld isn't yet started, `org.jboss.weld.Container#instance()` can be used to check state.
You can see how we do similar check here - https://github.com/wildfly/wildfly/blob/master/weld/subsystem/src/main/java/org/jboss/as/weld/WeldProvider.java#L69-L70

spyrkob: Thanks @manovotn, I added the check

spyrkob:
@manovotn can you re-review this please? Should I ask anybody else for the review as well?

manovotn:
I am fine with this. It shouldn't affect bootstrap at least.
So the only question that stands is how this affects some event-heavy application because it is effectively an extra thread local set/unset for every event regardless of whether you use it or not. Might be interesting to see the comparison but I don't have such perf test at hand.

The only other person you can try to get review from is @mkouba ;-)

manovotn (approved):

weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldExecutorServices.java
```diff
@@ -115,4 +120,54 @@ public synchronized ExecutorService getTaskExecutor() {
...
+            if (Container.instance().getState() == ContainerState.INITIALIZED) {
+                WeldTaskWrapper task = new WeldTaskWrapper(command, NamespaceContextSelector.getCurrentSelector());
+                super.execute(task);
+            } else {
+                super.execute(command);
+            }
+        }
+    }
+
+    class WeldTaskWrapper implements Runnable {
```
mkouba, line 158: This class and `WeldExecutor` do not seem to need to access the enclosing class (`WeldExecutorServices`) so it might make sense to make it `static`.

mkouba:
Looks good. But I would be interested in the numbers as well ;-).

weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/WeldExecutorServices.java
bstansberry: @spyrkob Can you make this change? Thanks.

spyrkob: Sorry, I missed that. It's updated now.

bstansberry (approved):

PR #14106, opened 2021-03-11
WFLY-14542 Cache entries in heap can exceed max-active-sessions when distributed session manager configured with local, passivating caches

pferraro:
https://issues.redhat.com/browse/WFLY-14542

Also resolves: WFLY-14531 InfinispanSessionManager should be more discriminating about listener registration
https://issues.redhat.com/browse/WFLY-14531


pferraro:
@rhusar Please review.

rhusar (approved):
👍

PR #14112, opened 2021-03-12
[WFLY-14536] Add tests to connector subsystem for attributes that all…

tadamski:
…ow expression

https://issues.redhat.com/browse/WFLY-14536


connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
```diff
@@ -493,6 +493,7 @@ public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceM
             .setRestartAllServices()
             .build();
 
+    @Deprecated
```
bstansberry, line 496: This is package protected and unused so why not just delete it?

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
```diff
@@ -227,18 +227,15 @@
     static final SensitiveTargetAccessConstraintDefinition DS_SECURITY_DEF = new SensitiveTargetAccessConstraintDefinition(DS_SECURITY);
 
     static final SimpleAttributeDefinition DEPLOYMENT_NAME = SimpleAttributeDefinitionBuilder.create("deployment-name", ModelType.STRING)
-            .setAllowExpression(true)
```
bstansberry, line 230: This and JDBC_COMPLIANT are used in two different ways -- as an attribute in the jdbc-driver resource, and then as part of the reply to get-installed-driver/installed-driver-list.

It seems like different AttributeDefinitions should be used for these two ways.  This one can be used for get-installed-driver/installed-driver-list. Expressions make no sense there. The other AD can be used for the jdbc-driver resource. It should allow expressions. But AFAICT those attributes are cruft, not actually used in any service. So add setDeprecationData to the AD builder. And then in another WF release or two,  remove them.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
```diff
@@ -227,18 +227,15 @@
...
             .setRequired(false)
             .build();
 
     static final SimpleAttributeDefinition MODULE_SLOT = SimpleAttributeDefinitionBuilder.create("module-slot", ModelType.STRING)
-            .setAllowExpression(true)
```
bstansberry, line 235: This and the change to DRIVER_MODULE_NAME are breaking changes. AIUI what's driving this change is parsing the xml, distinguishing between the ':'  separating the name and slot vs a ':' that may be inside an expression. If so that seems like something that can be solved without a breaking change.

darranl:
@tadamski This one has been around for a while but we have some comments from Brian and a merge conflict, is this one we can clean up to get merged or should we close for now and resubmit once ready?

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
tadamski: fixed (deleted)

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
tadamski: done

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
tadamski: I have added the new parsing and updated the test.

tadamski:
hi @darranl, I have rebased the branch and fixed the issued mentioned by Brian

darranl:
Thanks, just running CI again then we can see if we can get merged.

bstansberry:
Thanks, @tadamski

PR #14115, opened 2021-03-12
[WFLY-14566] Fix JSFDeploymentProcessorTestCase on Java11+secmgr

spyrkob:
Issue: https://issues.redhat.com/browse/WFLY-14566

Adding "suppressAccessChecks" persmission to the JSFDeploymentProcessorTestCase to workaround an issue in commons-beanutils library. 
During myfaces startup commons-beanutil's MethodUtils attempts to call Method.setAccessible(true) in order to workaround accessibility issue. If that fails, there is logic to check java version and log additional warnings. This logic fails with Java 11+ (https://issues.apache.org/jira/browse/BEANUTILS-547)

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/version/JSFDeploymentProcessorTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/version/JSFDeploymentProcessorTestCase.java
@@ -22,6 +22,7 @@
 package org.jboss.as.test.integration.jsf.version;
 
 import java.io.FilePermission;
+import java.lang.reflect.ReflectPermission;
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


darranl (changes requested):
When any proposed changes the permissions granted to test deployments are submitted we need the PR or code change to be accompanied with a description.  Ideally I would like to see the original error including full stack trace.

If I am reading this correctly there is an issue in an underlying component - we should not be adding permissions to test cases to work around these issues.  If there is an underlying problem we should be looking to address that.

spyrkob:
Added more information in the PR description and the stack trace in JIRA.

The root cause is indeed an issue in commons-beanutils. If it's better to wait for the component fix, I'll close this PR.

darranl:
@fjuma Do you have any thoughts on this one?  It looks to me like the priority should be fixing https://issues.apache.org/jira/browse/BEANUTILS-547 rather than adding a permission to the deployment.

fjuma:
@spyrkob @darranl Since the root cause is in commons-beanutils, we should make sure the problem is fixed there instead of adding the permission to the deployment.

spyrkob:
Thanks, I'll close this and see if the component issue can be resolved

PR #14136, opened 2021-03-19
[WFLY-11933] Return properly configured ModelNode when RBAC is enabled

jasondlee:
https://issues.redhat.com/browse/WFLY-11933

Return empty ModelNode for attributes restricted by RBAC so that isDefined() answers correctly
(h/t @jmesnil)

### DIFF
--- a/metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
+++ b/metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
@@ -36,13 +36,18 @@ import org.jboss.as.controller.LocalModelControllerClient;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
 
 public class WildFlyMetric implements Metric {
+
+    private static final ModelNode UNDEFINED = new ModelNode();
 
     private LocalModelControllerClient modelControllerClient;
     private final PathAddress address;
     private final String attributeName;
+
+    static {
+        UNDEFINED.protect();
+    }
 
     public WildFlyMetric(LocalModelControllerClient modelControllerClient, PathAddress address, String attributeName) {
         this.modelControllerClient = modelControllerClient;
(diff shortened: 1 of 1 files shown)


metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
```diff
@@ -79,7 +78,7 @@ private ModelNode readAttributeValue(PathAddress address, String attributeName)
             } else{
                 LOGGER.unableToReadAttribute(attributeName, address, error);
             }
-            return new ModelNode(ModelType.UNDEFINED);
+            return new ModelNode();
```
jmesnil, line 81: I think, we would better having a constant (`static final ModelNode UNDEFINED = new ModelNode()`) to avoid having to create many instances of the same object here

jmesnil (changes requested):

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
jasondlee: Ah, ok. That's what you meant by using a constant. :P One moment...

jmesnil (approved):

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
```diff
@@ -36,10 +36,11 @@
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
 
 public class WildFlyMetric implements Metric {
 
+    private static final ModelNode UNDEFINED = new ModelNode();
```
bstansberry, line 42: Please add a static block:

static {
    UNDEFINED.protect();
}

That makes the node immutable and thus safe to pass around.

bstansberry (changes requested):

metrics/src/main/java/org/wildfly/extension/metrics/WildFlyMetric.java
jasondlee: Ah. Very cool. Done.

bstansberry (approved):

PR #14175, opened 2021-04-01
[WFLY-14609] CLI ...service=timer-service/timer=* throws NullPointerException

gaol:
Issue: https://issues.redhat.com/browse/WFLY-14609

This PR tries to provide more meaningful failure message instead of NPE in case of `*` is used for the timer id.


bstansberry:

This looks ok to me.

chengfang (approved):

PR #14177, opened 2021-04-01
WFLY-14672: remove the usage of org.codehaus.jackson

wolfc:
Issue: https://issues.redhat.com/browse/WFLY-14672

This removes the dependency on org.codehaus.jackson.

Upgrading Hibernate Search and removing Apache Avro would still need to be done, but this is probably something for another PR.


ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/apache/avro/main/module.xml
```diff
@@ -32,8 +32,6 @@
     </resources>
 
     <dependencies>
-        <module name="org.codehaus.jackson.jackson-core-asl"/>
-        <module name="org.codehaus.jackson.jackson-mapper-asl"/>
```
bstansberry, line 36: This is the part that needs to be sorted. Avro before 1.9.0 uses Codehaus Jackson; later releases use FasterXML. 

Avro is used by Hibernate Search.

@yrodiere IIRC the expectation is a later Avro would work ok with the Hibernate Search we are using but testing is needed. Is that right?

wolfc: As per https://github.com/hibernate/hibernate-search/commit/ca77d8a482761f22f9a2170d7fc2904ac6b2570c (6.0.2) the Avro serialization was removed. So upgrading Hibernate Search to 6.0.2.Final should eliminate this issue.

wolfc:
In accordance with the new review policy from Bill, this is now ready to be merged. :)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/apache/avro/main/module.xml
yrodiere: Sorry, I was on PTO, catching up now...

@wolfc Hibernate Search 6 offers similar features, but backward compatibility was clearly ruled out from the very start. Upgrading Search to 6.0.2.Final will break backward compatibility in almost every possible way (API, behavior, feature coverage, ...). I doubt you want that in WildFly, unless you are working on a new major release (in which case I'd love to help with the upgrade).

To keep backward compatibility, we need to stick to 5.x.

As to Avro... For sure we cannot upgrade Avro without patching Hibernate Search. I can upgrade, and in fact [I have a PR ready](https://github.com/hibernate/hibernate-search/pull/2106).

However, as I explained before, Avro is used for an old feature that I have little familiarity with, and is not very well tested. Evaluating the risk of regression would be challenging. The few tests we have do pass, that's all I can say...

wolfc: IMO we should start upgrading WildFly to the latest and the greatest.

yrodiere: That would be nice, though there are no such plans to the best of my knowledge and we haven't started working on a proper migration path (deprecation, ...).

In the meantime... Do you need me to release a new version of Search 5.10 with the Avro upgrade (which version exactly?), so that we can put that in WildFly?

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
```diff
@@ -87,14 +81,12 @@ public void registerAdditionalRuntimePackages(ManagementResourceRegistration res
...
                     RuntimePackageDependency.optional(RESTEASY_MULTIPART.getName()),
                     RuntimePackageDependency.optional(RESTEASY_YAML.getName()),
-                    RuntimePackageDependency.optional(JACKSON_CORE_ASL.getName()),
                     RuntimePackageDependency.optional(RESTEASY_CRYPTO.getName()),
                     RuntimePackageDependency.optional(JACKSON_DATATYPE_JDK8.getName()),
                     RuntimePackageDependency.optional(JACKSON_DATATYPE_JSR310.getName()),
                     // The following ones are optional dependencies located in org.jboss.as.jaxrs module.xml
                     // To be provisioned, they need to be explicitly added as optional packages.
                     RuntimePackageDependency.optional("org.jboss.resteasy.resteasy-jettison-provider"),
-                    RuntimePackageDependency.optional("org.jboss.resteasy.resteasy-jackson-provider"),
```
wolfc, line 97: Note that `org.jboss.resteasy.resteasy-jackson2-provider` is already present as it was added on line 79.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/apache/avro/main/module.xml
wolfc: At some point WildFly needs to move forward, might as well be now. At least the usage of Codehaus Jackson must be removed asap.
@bstansberry WDYT?

bstansberry: @wolfc We are not making the kinds of breaking changes @yrodiere describes as coming with Hibernate Search 6 in WildFly 24.  So we'd either defer this until at least WF 25 or we need a non-breaking Hibernate Search with the dep updated. Moving forward and crossing this codehaus jackson item off the list would be nice though if it's not too much work for @yrodiere.

Also, AIUI Hibernate Search 6 requires ORM 5.4, and moving to 5.4 wasn't something we were eager to do. Once @scottmarlow gets unburied from his EE 9.1 work (he's the critical man!) I want to get a reset on where we're going with that, but in any case i don't see moving off of ORM 5.3 in WF 24.

Re: what avro to use, perhaps 1.10.2, as it uses fasterxml jackson 2.12.x, same as WildFly. The avro 1.9 series uses jackson 2.10.

yrodiere: I updated [my Hibernate Search PR](https://github.com/hibernate/hibernate-search/pull/2106) to upgrade to Avro 1.10.2 and Jackson 2.12.3.

Now, as a reminder, doing this kind of upgrades in a maintenance branch (5.10) puts us in a very delicate situation. We will no longer have a maintenance branch of Hibernate Search 5.10 without the Avro upgrade. But Hibernate Search 5.10 is used in many older WildFly versions that still use Avro 1.7.

If we ever need to push a bugfix to Search 5.10, and get that fix in an older version of WildFly that uses Avro 1.7... Well, we'll have to get creative with the git branches and the versioning. Create a 5.10.10.SP1, maybe.

I think that's one of the reasons that led us to give up on this upgrade in [WFLY-14188](https://issues.redhat.com/browse/WFLY-14188).

So, before I proceed with the release, can @scottmarlow confirm this is what we want?

scottmarlow (approved):

PR #14180, opened 2021-04-04
[WFLY-14673] Utilize org.wildfly.common.Assert for Null-Checks

boris-unckel:


bstansberry:
@boris-unckel Thanks for this, but this kind of thing needs to be broken up by maven module or logically related group of maven modules (e.g. clustering) so the relevant component leads can review it their parts. Big cross-cutting refactors are too likely to introduce bugs.

boris-unckel:
> Thanks for this, but this kind of thing needs to be broken up by maven module or logically related group of maven modules (e.g. clustering) so the relevant component leads can review it their parts. Big cross-cutting refactors are too likely to introduce bugs.

@bstansberry Thanks for your fast check. Do you expect one Jira and several pull requests or do I need to split up the JIRA also? I would utilize the toplevel path under wildfly/[splitHere] to provide the PRs.

PR #14185, opened 2021-04-06
[WFLY-14625]: WARN message jaegertracing: FlushCommand execution failed!

ehsavoie:
* replacing Jaeger ThriftSenderFactory with our own to swallow
  exceptions thrown by the UdpSender which were swallowed before.
* some module.xml magic for the ServiceLoader to load our
  implementation.

Jira: https://issues.redhat.com/browse/WFLY-14625

### DIFF
--- a/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySender.java
+++ b/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySender.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2021 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.wildfly.extension.microprofile.opentracing.spi.sender;
+
+import io.jaegertracing.internal.JaegerSpan;
+import io.jaegertracing.internal.exceptions.SenderException;
+import io.jaegertracing.spi.Sender;
+import org.wildfly.extension.microprofile.opentracing.TracingExtensionLogger;
+
+/**
+ * Jaeger client Sende implementation to be able to 'swallow' exceptions when sending the spans to a Jaeger server.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
+ */
+public class WildFlySender implements Sender {
+
+    private Sender delegate;
+
+    WildFlySender(Sender sender) {
+        TracingExtensionLogger.ROOT_LOGGER.debugf("Creating new WildFly Sender with %s", sender);
+        this.delegate = sender;
+    }
+
+    @Override
+    public int append(JaegerSpan span) throws SenderException {
+        return delegate.append(span);
+    }
+
+    @Override
+    public int flush() throws SenderException {
+        try {
+            return delegate.flush();
+        } catch (SenderException ex) {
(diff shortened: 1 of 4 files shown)


microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySender.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Jaeger client Sende implementation to be able to 'swallow' exceptions when sending the spans to a Jaeger server.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
+ */
+public class WildFlySender implements Sender {
+
+    private Sender delegate;
+
+    WildFlySender(Sender sender) {
+        TracingExtensionLogger.ROOT_LOGGER.warnf("Creating new WildFly Sender with %s", sender);
```
bstansberry, line 32: If this is kept I think it's debug.

bstansberry (changes requested):
@ehsavoie Needs a small change.

This is the approach we want, yes? At least for now.

When this approach was first being discussed I wasn't sure if it was an experiment, so I'm just double-checking.

bstansberry:
@ehsavoie never mind about my question about whether this is an old experiment -- I see now this PR is only 1 day old!

ehsavoie:
@bstansberry there were 4 approaches:
 - use the logging subsystem to disable this trace :-1: 
 - ask jaeger client for a change (PR submitted) but they refused saying this was a logging configuration issue :-1: 
 - change our default config not to provide opentracing: this is not backward compatible :-1: 
 - this fix

bstansberry:
@ehsavoie Thanks. Please note the comment about the WARN log in the WildFlySender constructor.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySender.java
```diff
@@ -0,0 +1,56 @@
...
+package org.wildfly.extension.microprofile.opentracing.spi.sender;
+
+import io.jaegertracing.internal.JaegerSpan;
+import io.jaegertracing.internal.exceptions.SenderException;
+import io.jaegertracing.spi.Sender;
+import org.wildfly.extension.microprofile.opentracing.TracingExtensionLogger;
+
+/**
+ * Jaeger client Sende implementation to be able to 'swallow' exceptions when sending the spans to a Jaeger server.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
```
jamezp, line 25: Not really a huge deal, but it's 2021 :)

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySenderFactory.java
```diff
@@ -0,0 +1,40 @@
...
+package org.wildfly.extension.microprofile.opentracing.spi.sender;
+
+import io.jaegertracing.Configuration;
+import io.jaegertracing.spi.Sender;
+import io.jaegertracing.spi.SenderFactory;
+import io.jaegertracing.thrift.internal.senders.ThriftSenderFactory;
+
+/**
+ * Jaeger client SenderFactory implementation to be able to 'swallow' exceptions when sending the spans to a Jaeger server.
+ * @author Emmanuel Hugonnet (c) 2020 Red Hat, Inc.
```
jamezp, line 25: Same comment as above :)

jamezp:
One thing I'm still not completely clear on is why, I'm guessing it's the client itself not WildFly, it just attempts to start sending data even if no server is there to receive the data. That was my only point for not adding the subsystem by default. If there is no server to receive the data what's the point in attempting to send it?

That said I don't really know much about the spec or implementation so I could misunderstand some things :)

bstansberry:
@jamezp I think it's legit that it tries to send the data and logs if that fails. It's configured to send and that doesn't work.  Repeatedly logging isn't great, though.

Changing our default configs to not have the sender is the more correct solution, really, but it's an incompatible change.

bstansberry (approved):

bstansberry:
@jasondlee WDYT?

jasondlee:
@bstansberry Apologies. I thought this was closed. I think the change sounds reasonable. They're not willing to change on their end, and the "right" fix isn't backwards compatible, so we're left with this.

+1

PR #14186, opened 2021-04-06
[WFLY-14673] Utilize org.wildfly.common.Assert for Null-Checks (clust…

boris-unckel:
…ering)

The PR 14180 was too big to verify. It has been splitted up, here: clustering.

### DIFF
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/clustering/infinispan/spi/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/clustering/infinispan/spi/main/module.xml
@@ -45,6 +45,7 @@
         <module name="org.wildfly.clustering.marshalling.protostream"/>
         <module name="org.wildfly.clustering.marshalling.spi"/>
         <module name="org.wildfly.clustering.service"/>
+        <module name="org.wildfly.common"/>
         <module name="org.wildfly.security.elytron-private"/>
     </dependencies>
 </module>
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
@@ -21,6 +21,8 @@
  */
 package org.jboss.as.test.clustering.cluster;
 
+import static org.wildfly.common.Assert.checkNotNullArrayParam;
+
 import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Map;
@@ -257,10 +259,7 @@ public abstract class AbstractClusteringTestCase {
         String[] getContainers(String... nodes) {
             String[] containers = new String[nodes.length];
             for (int i = 0; i < nodes.length; ++i) {
-                String node = nodes[i];
-                if (node == null) {
-                    throw new IllegalArgumentException();
-                }
+                String node = checkNotNullArrayParam("nodes", i, nodes[i]);
                 containers[i] = node;
             }
             return containers;
--- a/clustering/ee/cache/src/main/java9/org/wildfly/clustering/ee/cache/scheduler/FastConcurrentDirectDeque.java
+++ b/clustering/ee/cache/src/main/java9/org/wildfly/clustering/ee/cache/scheduler/FastConcurrentDirectDeque.java
@@ -24,6 +24,8 @@
 
 package org.wildfly.clustering.ee.cache.scheduler;
 
+import static org.wildfly.common.Assert.checkNotNullParamWithNullPointerException;
+
 import java.io.Serializable;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
@@ -1257,7 +1259,7 @@ public class FastConcurrentDirectDeque<E> extends AbstractCollection<E> implemen
     @Override
     @SuppressWarnings("unchecked")
     public <T> T[] toArray(T[] a) {
-        if (a == null) throw new NullPointerException();
+        checkNotNullParamWithNullPointerException("a", a);
         return (T[]) toArrayInternal(a);
     }
 
--- a/clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/persistence/IndexedKeyFormatMapper.java
+++ b/clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/persistence/IndexedKeyFormatMapper.java
@@ -22,6 +22,8 @@
 
 package org.wildfly.clustering.infinispan.spi.persistence;
 
+import static org.wildfly.common.Assert.checkNotNullParam;
+
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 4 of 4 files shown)


bstansberry:
@pferraro Please review

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
```diff
@@ -289,10 +289,7 @@ public void stop(String... nodes) {
             String[] deployments = new String[nodes.length];
             for (int i = 0; i < nodes.length; ++i) {
                 String node = nodes[i];
-                String deployment = NODE_TO_DEPLOYMENT.get(node);
-                if (deployment == null) {
-                    throw new IllegalArgumentException(node);
-                }
+                String deployment = checkNotNullParam(String.valueOf(node), NODE_TO_DEPLOYMENT.get(node));
```
pferraro, line 292: This is not a proper use case of Assert. "deployment" is not a parameter.

pferraro (changes requested):

boris-unckel:
@pferraro Thanks for taking the time to review the PR. I agree that the variables you mentioned are not parameters. I have reverted the changes and updated the PR. I still think it was a valid approach: [Google Guavas Preconditions](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Preconditions.java#L894) and [Apache Commons Lang Validate](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/Validate.java) share the same idea. Aside from a dependency discussion they do not face the project reality where both IAE and NPE is in use. o.w.common.Assert perfectly adresses that, but is focussed on "param". I like the fluent API style of the check and the enforcement of an exception text.
Every reverted exception in this PR already had a meaningful text, so there is no loss for the concrete case.

clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/PortableConcurrentDirectDeque.java
```diff
@@ -1124,7 +1116,7 @@ public boolean addAll(Collection<? extends E> c) {
         // Copy c into a private chain of Nodes
         Node<E> beginningOfTheEnd = null, last = null;
         for (E e : c) {
-            checkNotNull(e);
+            checkNotNullParamWithNullPointerException("e", e);
```
pferraro, line 1119: "e" is not a parameter of this method.  Is there an Assert.checkNotNullArrayParam(...) equivalent for collections?

clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/FastConcurrentDirectDeque.java
```diff
@@ -1157,7 +1149,7 @@ public boolean addAll(Collection<? extends E> c) {
         // Copy c into a private chain of Nodes
         Node<E> beginningOfTheEnd = null, last = null;
         for (E e : c) {
-            checkNotNull(e);
+            checkNotNullParamWithNullPointerException("e", e);
```
pferraro, line 1152: "e" is not a parameter of this method. Is there an Assert.checkNotNullArrayParam(...) equivalent for collections?

clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/FastConcurrentDirectDeque.java
```diff
@@ -853,7 +845,7 @@ public FastConcurrentDirectDeque(Collection<? extends E> c) {
         // Copy c into a private chain of Nodes
         Node<E> h = null, t = null;
         for (E e : c) {
-            checkNotNull(e);
+            checkNotNullParamWithNullPointerException("e", e);
```
pferraro, line 848: "e" is not a parameter of this method. Is there an Assert.checkNotNullArrayParam(...) equivalent for collections?

clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/PortableConcurrentDirectDeque.java
```diff
@@ -820,7 +812,7 @@ public PortableConcurrentDirectDeque(Collection<? extends E> c) {
         // Copy c into a private chain of Nodes
         Node<E> h = null, t = null;
         for (E e : c) {
-            checkNotNull(e);
+            checkNotNullParamWithNullPointerException("e", e);
```
pferraro, line 815: "e" is not a parameter of this method. Is there an Assert.checkNotNullArrayParam(...) equivalent for collections?

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/clustering/infinispan/spi/main/module.xml
```diff
@@ -46,5 +46,6 @@
         <module name="org.wildfly.clustering.marshalling.spi"/>
         <module name="org.wildfly.clustering.service"/>
         <module name="org.wildfly.security.elytron-private"/>
+        <module name="org.wildfly.common"/>
```
pferraro, line 49: Nit: Please retain the lexicographical ordering.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/ProtocolMetricsHandler.java
```diff
@@ -234,7 +234,6 @@ public ProtocolMetricsHandler(FunctionExecutorRegistry<JChannel> executors) {
 
     @Override
     protected void executeRuntimeStep(OperationContext context, ModelNode operation) throws OperationFailedException {
-
```
pferraro, line 237: Nit: I assume this was unintentional?

pferraro (changes requested):

clustering/ee/cache/src/main/java/org/wildfly/clustering/ee/cache/scheduler/PortableConcurrentDirectDeque.java
boris-unckel: There is no equivalent method, but even when it would not be applicable to a non-parameter variable here. I have reverted the change.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/clustering/infinispan/spi/main/module.xml
pferraro: I'll fix this later.

pferraro (approved):

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/clustering/infinispan/spi/main/module.xml
boris-unckel: @pferraro Sorry, I did not recognize you meant the existing entries also. Next time better. Thanks for taking the time to review and merge.

PR #14190, opened 2021-04-06
[WFLY-14682] Upgrade ironjacamar to 1.4.31.Final 385a118

tadamski:
https://issues.redhat.com/browse/WFLY-14682

This is an IJ upgrade that implements:
https://issues.redhat.com/browse/WFLY-14347 (https://issues.redhat.com/browse/EAP7-1189)
+ additional performance fixes to: https://issues.redhat.com/browse/EAP7-168

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/util/ModelNodeUtil.java
+++ b/connector/src/main/java/org/jboss/as/connector/util/ModelNodeUtil.java
@@ -1,5 +1,27 @@
+    /*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
 package org.jboss.as.connector.util;
 
+import org.jboss.as.connector.logging.ConnectorLogger;
 import org.jboss.as.controller.ObjectListAttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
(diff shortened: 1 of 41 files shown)


connector/src/main/java/org/jboss/as/connector/util/ModelNodeUtil.java
```diff
@@ -41,16 +65,36 @@ public static String getResolvedStringIfSetOrGetDefault(final OperationContext c
...
+                moduleName = getResolvedStringIfSetOrGetDefault(operationContext, dataSourceNode, moduleNameAttribute);
+            }
+
+            ClassLoader moduleClassLoader = null;
+            if (moduleName != null) {
+                try {
+                    final ModuleIdentifier exceptionSorterModuleId = ModuleIdentifier.create(moduleName);
+                    moduleClassLoader = Module.getCallerModuleLoader().loadModule(exceptionSorterModuleId).getClassLoader();
+                } catch (ModuleLoadException exception) {
+                    throw new OperationFailedException("Wrong module name", exception);
```
bstansberry, line 90: This needs i18n

testsuite/integration/smoke/src/test/config/arq/arquillian.xml
```diff
@@ -7,7 +7,7 @@
     <container qualifier="jboss" default="true">
         <configuration>
             <property name="jbossHome">${jboss.install.dir}</property>
-            <property name="javaVmArguments">${server.jvm.args} -Djboss.inst=${basedir}/target/wildfly</property>
+	    <property name="javaVmArguments">${server.jvm.args} -Djboss.inst=${basedir}/target/wildfly</property>
```
bstansberry, line 10: Minor: looks like this change should be removed.

bstansberry (changes requested):
The one commit here is a management API change so the subsystem version needs to be incremented, transformers are needed, and the new attributes should be covered in the subsystem test.

Note that there are 3 different things in this PR, and one is a Feature Request that requires the kanban process.  The others are a micro component upgrade and an error reporting enhancement, neither of which require kanban.  So if you want the non-RFE commits merged faster you can file a separate PR with just those two.

connector/src/main/java/org/jboss/as/connector/util/ModelNodeUtil.java
tadamski: fixed

testsuite/integration/smoke/src/test/config/arq/arquillian.xml
tadamski: fixed

tadamski:
@bstansberry  I decided to combine those two cause I wanted to avoid doing IronJacamar release for each JIRA and with recent IJ upgrade those are needed to compile. Anyway, this should be fixed now.

tadamski:
hi @bstansberry, I received comment from Carlo that EAP7-1189 API change "violates backwards compatibility requirement for micro releases". Should this change be reverted?

bstansberry:
@tadamski If a change in an IJ maintenance branch introduced some kind of incompatibility, yes, it should be reverted.

If the incompatible change is in some kind user-accessible it's best to avoid doing it, even in a new minor.

tadamski:
@bstansberry I have reverted the API  modifing change and removed the associated commit from this PR

bstansberry:
Thanks @tadamski. This is still held up by needing to go through the RFE process for the WFLY-14347 commit though.

darranl:
@tadamski what is the current status of this one?  Are we still looking to get this through the RFE process?

tadamski:
@darranl I think this should be closed, it is going through RFE process and was replaced by:
https://github.com/wildfly/wildfly/pull/14598

bstansberry:
@tadamski If the WFLY-14313 fix is still needed can you break that commit out into a separate PR, and then close this one please? Thanks.

tadamski:
@bstansberry WFLY-14313 was already included as a part of another PR as well, I'm closing this one

PR #14194, opened 2021-04-07
[WFLY-14673] Utilize o.w.common.Assert for Null-Checks (ejb)

boris-unckel:

The PR 14180 was too big to verify. It has been splitted up, here: ejb

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/EjbDeploymentInformation.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/EjbDeploymentInformation.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.ejb3.deployment;
 
+import static org.wildfly.common.Assert.checkNotNullParam;
+
 import org.jboss.as.ee.component.ComponentView;
 import org.jboss.as.ejb3.component.EJBComponent;
 import org.jboss.as.ejb3.iiop.EjbIIOPService;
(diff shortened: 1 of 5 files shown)


bstansberry:
@chengfang Please review

ejb3/src/main/java/org/jboss/as/ejb3/deployment/EjbDeploymentInformation.java
```diff
@@ -106,10 +108,7 @@ public EJBComponent getEjbComponent() {
     }
 
     public ComponentView getView(String name) {
-        final InjectedValue<ComponentView> value = componentViews.get(name);
-        if (value == null) {
-            throw new IllegalArgumentException("View " + name + " was not found");
-        }
+        final InjectedValue<ComponentView> value = checkNotNullParam(String.valueOf(name), componentViews.get(name));
```
chengfang, line 111: String.valueOf() seems redundant.

boris-unckel: @chengfang thanks for your review. This is intended because the previous version did not null check 'name'. [org.wildfly.common.Assert.checkNotNullParam](https://github.com/wildfly/wildfly-common/blob/master/src/main/java/org/wildfly/common/Assert.java#L47) does check the not-null for the excetion text. If the String name contained null but null is a valid lookup value (i.E. a default comes back) the previous version would perfectly work. If you require me to change that I would do it with two checks:
`checkNotNullParam("name", name);
final InjectedValue<ComponentView> value = name, componentViews.get(name));`
The two check solution is OK, but it changes the behaviour.
Please have a look at it again and advise what is needed here.

chengfang: @boris-unckel thanks for explaining this. I'm fine with your original fix.

chengfang (approved):

PR #14208, opened 2021-04-09
[WFLY-14692] Tests security domain with dots in the name

dcihak:
Creates security domain with a name containing dots and tries to log in using the POST method through the secured form.

Test case for [WFCORE-4707](https://issues.redhat.com/browse/WFCORE-4707).

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/security/SecurityDomainDotNameServerSetupTask.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/security/SecurityDomainDotNameServerSetupTask.java
@@ -1,0 +1,100 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.test.integration.management.api.security;
+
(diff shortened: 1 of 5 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/security/SecurityDomainDotNameServerSetupTask.java
```diff
@@ -0,0 +1,78 @@
+package org.jboss.as.test.integration.management.api.security;
```
darranl, line 1: Missing copyright header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/security/SecurityDomainDotNameTestCase.java
```diff
@@ -0,0 +1,86 @@
+package org.jboss.as.test.integration.management.api.security;
```
darranl, line 1: Missing copyright header.

darranl (changes requested):
As a submission to the WildFly project this should be using a WFLY Jira issue to track the changes even if the component is just testsuite.

This test is testing deprecated code, when we remove legacy security we are more likely to just delete all legacy security tests.  

But the underlying issue sounds to be very generic, is this reproducible using non-deprecated code as this test will likely be deleted again within the next 12 months.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/security/SecurityDomainDotNameTestCase.java
dcihak: Copyright added.

dcihak:
@darranl Thanks for the review!
JIRA issue [WFLY-14692](https://issues.redhat.com/browse/WFLY-14692) was created to be linked with this test case. The original issue was related to the security subsystem and it seems it is not reproducible with the non-legacy configuration. As we need this test case for an authomatic verification of the 7.3.7 JBEAP issue [JBEAP-21215](https://issues.redhat.com/browse/JBEAP-21215) (will be merged into EAP AS TS) do you think it will be better to merge this into  wildfly-core testsuite?

darranl:
In general my preference is get the test case as close as possible to the code it is testing, from an engineers perspective it makes it faster for an engineer to identify if they have caused a regression in that code rather than waiting for a later testsuite to detect the regression.

Having said that this test does require the legacy security subsystem for it's reproducer and the life of the test upstream is potentially limited so I probably would not worry at this stage.

Do be aware once we reach WildFly 25 or 26 development we may be ready to pass the point where we stop accepting legacy security changes and tests upstream so at some point in the future these kind of tests may not need an upstream change although we are not quite there yet.

dcihak:
Retest this please.

dcihak:
@darranl We are ok with the limited life of the test case, but thanks for the information. I fixed the test for the case of Galleon JDK 11. Do we need some other changes to be done from my side? Thanks.

dcihak:
@darranl Hello, are there any changes required from your side or can we merge this? We would like to use this test case for the 7.3.7 JIRA verificaton which is currently in progress. All the checks have passed. Thanks.

darranl (approved):

PR #14214, opened 2021-04-13
[WFLY-14701] Narayana component upgrade to 5.11.1.Final

ochaloup:
https://issues.redhat.com/browse/WFLY-14701


ochaloup:
@bstansberry it seems there could be some trouble in test or Narayana. I'm investigating what's going on. I will let you know back.

ochaloup:
I'm closing this PR as it seems the trouble could be some corner case handling within txn reaper thread. I think we will need to find a fix and then proceed with a new release.

PR #14221, opened 2021-04-16
[23.x] Move to GF jaxb 2.3.4 and istack 3.0.12

bstansberry:
Just an experiment.


bstansberry:
For WildFly Preview, #14223 is a better approach and for standard WildFly there is no particular driver for doing this upgrade in 23.x.

PR #14232, opened 2021-04-19
[WFLY-14714] Specify bean-discovery-mode for RSO CDI Provider

kabir:
https://issues.redhat.com/browse/WFLY-14714

### DIFF
--- a/microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
+++ b/microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
@@ -28,6 +28,7 @@ import java.util.Iterator;
 import java.util.ServiceLoader;
 
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
 import javax.enterprise.inject.Produces;
 
 import org.eclipse.microprofile.reactive.streams.operators.spi.ReactiveStreamsEngine;
@@ -35,6 +36,7 @@ import org.eclipse.microprofile.reactive.streams.operators.spi.ReactiveStreamsEn
 /**
  * @author <a href="mailto:kabir.khan@jboss.com">Kabir Khan</a>
  */
+@Dependent
 public class ReactiveEngineProvider {
 
     /**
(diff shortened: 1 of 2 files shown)


bstansberry:
@kabir tests fail

kabir:
@bstansberry fixed

microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
```diff
@@ -43,7 +44,6 @@
      * @throws IllegalStateException if no implementations are found.
      */
     @Produces
-    @ApplicationScoped
```
manovotn, line 46: Keep this `@ApplicationScoped`. Removing the annotation means it will be `@Dependent` - the annotation is **not inherited** from the class itself. Using dependent would leads to creation of new bean of this type for every injection point whereas before you had one shared instance.

microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
```diff
@@ -35,6 +35,7 @@
 /**
  * @author <a href="mailto:kabir.khan@jboss.com">Kabir Khan</a>
  */
+@ApplicationScoped
```
manovotn, line 38: It is correct to place a bean defining annotation here but since I don't know the internals of its usage, I can't say whether `@Dependent` or `@ApplicationScoped`.
Though I can tell you that it was previously (with discovery mode `all`) considered a `@Dependent` bean.

microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
kabir: Done -thanks for the feedback!

microprofile/reactive-streams-operators-smallrye/cdi-provider/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/cdi/ReactiveEngineProvider.java
kabir: Done -thanks for the feedback!

PR #14246, opened 2021-04-22
[WFLY-14279] Don't prematurely validate the appclient file param. Hav…

bstansberry:
…e the teststuite pass expressions to the appclient process to validate they are handled

https://issues.redhat.com/browse/WFLY-14729

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/util/AppClientWrapper.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/util/AppClientWrapper.java
@@ -171,11 +171,15 @@ public class AppClientWrapper implements Runnable {
         }
         final ZipExporter exporter = archive.as(ZipExporter.class);
         exporter.exportTo(archiveOnDisk);
+        final String archivePath = archiveOnDisk.getAbsolutePath();
+
+        // We don't directly pass the archive file and deployment name to appclient's main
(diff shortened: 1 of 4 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/appclient/util/AppClientWrapper.java
```diff
@@ -202,6 +206,8 @@ private String getAppClientCommand() throws Exception {
                 " -Djboss.server.base.dir="+ asDist + "/appclient" +
                 " -Djboss.home.dir="+ asDist +
                 " " + this.appClientArgs + " " + archiveArg + " " + args;
```
yersan, line 208: `args` is the value used for AppClientSubsystemResourceDefinition.PARAMETERS, but I don't see it wrapped by an expression, shouldn't it be?

bstansberry: Thanks @yersan -- I missed that that one supported expression. One test method passes a non-empty arg so I'll wrap that value in an expression.

bstansberry: I've updated AbstractSimpleApplicationClientTestCase.simpleAppClientTest().

PR #14259, opened 2021-04-28
WFLY-14670 Fix intermittent failures in HotRod-based web session expiration tests

pferraro:
Follows up on WFLY-14670 to fix regressions in remote session expiration tests.
https://issues.redhat.com/browse/WFLY-14670


rhusar (approved):
👍

PR #14263, opened 2021-04-30
[WFLY-14750] Batch task not restarted after server resumed from suspended state

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-14750

Regression found created by WFCORE-4291. The new logic to start stopped tasks is OK for server boot but does not take into account app deployment time. If the app is deployed and then the server is suspended, the `running` property is never initialized (so it's false) so the tasks are not restarted at resume time (it waits for a server change event that is not going to happen). Just calling an `initialize` method to init the two boolean properties always with the current server values.

Test added with this use-case too.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/suspend/SuspendBatchletTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/suspend/SuspendBatchletTestCase.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2021 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.as.test.integration.batch.suspend;
+
+import java.io.FilePermission;
+import java.io.IOException;
+import java.util.List;
+import java.util.Properties;
+import java.util.PropertyPermission;
+import java.util.concurrent.TimeUnit;
+import javax.batch.operations.JobOperator;
+import javax.batch.runtime.BatchRuntime;
+import javax.batch.runtime.BatchStatus;
+import javax.batch.runtime.JobExecution;
+import javax.batch.runtime.JobInstance;
+import javax.batch.runtime.StepExecution;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.test.integration.batch.common.AbstractBatchTestCase;
+import org.jboss.as.test.shared.PermissionUtils;
+import org.jboss.as.test.shared.TimeoutUtil;
+import org.jboss.dmr.ModelNode;
+import org.jboss.remoting3.security.RemotingPermission;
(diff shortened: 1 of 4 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/suspend/SuspendBatchletTestCase.java
```diff
@@ -0,0 +1,134 @@
...
+    @Deployment
+    public static WebArchive createDeployment() {
+        return createDefaultWar("suspend-batchlet.war", SuspendBatchletTestCase.class.getPackage(), "suspend-batchlet.xml")
+                .addClass(LongRunningBatchlet.class)
+                .addAsResource(new StringAsset("Dependencies: org.jboss.dmr, org.jboss.as.controller, org.jboss.remoting\n"), "META-INF/MANIFEST.MF")
+                .addAsManifestResource(PermissionUtils.createPermissionsXmlAsset(
+                        new RemotingPermission("createEndpoint"),
+                        new RemotingPermission("connect"),
+                        new PropertyPermission("ts.timeout.factor", "read"),
+                        new FilePermission(System.getProperty("jboss.inst") + "/standalone/tmp/auth/*", "read")
```
chengfang, line 68: Looks like the test wanted to use timeout factor to adjust wait time, but `Timeout.adjust()` is not used in the test code.

rmartinc: @chengfang I think it is used by the `waitForTermination` (inherited from the `AbstractBatchTestCase`). It's used inside the test to wait for the termination of the job. If the permission is not added, the execution with `-Dsecurity.manager` fails with the following exception:

```
[ERROR] testSuspendResume(org.jboss.as.test.integration.batch.suspend.SuspendBatchletTestCase)  Time elapsed: 0.379 s  <<< ERROR!
java.lang.ExceptionInInitializerError
	at deployment.suspend-batchlet.war//org.jboss.as.test.integration.batch.common.AbstractBatchTestCase.waitForTermination(AbstractBatchTestCase.java:130)
	at deployment.suspend-batchlet.war//org.jboss.as.test.integration.batch.suspend.SuspendBatchletTestCase.checkJobExecution(SuspendBatchletTestCase.java:87)
	at deployment.suspend-batchlet.war//org.jboss.as.test.integration.batch.suspend.SuspendBatchletTestCase.testSuspendResume(SuspendBatchletTestCase.java:110)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
...
Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.util.PropertyPermission" "ts.timeout.factor" "read")" in code source "(vfs:/content/suspend-batchlet.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.suspend-batchlet.war" from Service Module Loader")
	at org.wildfly.security.elytron-private@1.15.3.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:309)
	at org.wildfly.security.elytron-private@1.15.3.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPropertyAccess(WildFlySecurityManager.java:493)
	at java.base/java.lang.System.getProperty(System.java:814)
	at java.base/java.lang.Integer.getInteger(Integer.java:1331)
	at java.base/java.lang.Integer.getInteger(Integer.java:1287)
	at deployment.suspend-batchlet.war//org.jboss.as.test.shared.TimeoutUtil.lambda$static$0(TimeoutUtil.java:44)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at deployment.suspend-batchlet.war//org.jboss.as.test.shared.TimeoutUtil.<clinit>(TimeoutUtil.java:44)
	... 152 more
```
I'm not using it for the sleep in the line 124, if you want I can adjust it (but I think 100ms is OK).

chengfang: For better configuration of testing and consistency, I think we should use adjusted timeout in both batchlet process method and test method.

rmartinc: Perfect @chengfang! I have created a `waitForJobRestarted` private method in the test  which is doing more or less the same that `waitForTermination`, but it waits for the job to be restarted after the resume.

chengfang (approved):

PR #14264, opened 2021-05-01
[WFLY-14751] Create a servlet-feature-pack/pruned module; move the or…

bstansberry:
…g.jboss.as.security module to it

https://issues.redhat.com/browse/WFLY-14751


pom.xml
```diff
@@ -1492,6 +1492,13 @@
                 <type>zip</type>
             </dependency>
 
+            <dependency>
+                <groupId>${ee.maven.groupId}</groupId>
+                <artifactId>wildfly-servlet-feature-pack-pruned</artifactId>
+                <version>${ee.maven.version}</version>
+                <type>zip</type>
+            </dependency>
```
yersan, line 1500: Is there any specific reason for adding only the zip artifact and not the pom type as well?

It would just avoid the need to specify a version on each child module that uses it, not a great advantage, so I'm just asking

bstansberry: Good question. I have no idea why I did that. It's such an odd way to do it that I suspect I had a reason, but I have no idea what it was! I'll do it the normal way and if something breaks, that will tell me why I did it. :)

bstansberry: @yersan If I do that the build fails:

```
[ERROR] Failed to execute goal on project wildfly-ee-galleon-pack: Could not resolve dependencies for project org.wildfly:wildfly-ee-galleon-pack:pom:24.0.0.Final-SNAPSHOT: Failure to find org.wildfly:wildfly-servlet-feature-pack-pruned:pom:24.0.0.Final-SNAPSHOT in https://repository.jboss.org/nexus/content/groups/public/ was cached in the local repository, resolution will not be reattempted until the update interval of jboss-public-repository-group has elapsed or updates are forced -> [Help 1]
```

The maven Reactor Summary shows that it intends to build wildfly-servlet-feature-pack-pruned towards the end; i.e. too late. I guess it feels free to do that because wildfly-ee-galleon-pack records the dep as provided scope (which is correct). I'm not sure why including the 'version' element in wildfly-ee-galleon-pack prevents the problem by causing the reactor to build things in the right order, but it does.

I see now that the wildfly-ee-galleon-pack pom includes 'version' elements for all the modules from which it pulls resources. I suppose when I reworked the FP construction during WF 22 I hit this problem and worked around it.

yersan (approved):

pom.xml
bstansberry: I'm using Maven 3.6.3.

To see what I saw it might be important that the artifact isn't already present in the local maven repo (e.g. from an earlier build.)

bstansberry:
I put the hold label on this because the picketbox removal work is in active progress and I don't want this to disrupt it.

bstansberry:
I no longer see the need to move the legacy security subsystem to pruned, as its current model-only form is valid for use in WildFly Preview.

PR #14266, opened 2021-05-02
[WFLY-14754] Upgrade ASM to 9.1

bstansberry:
https://issues.redhat.com/browse/WFLY-14754

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -458,7 +458,7 @@
         <version.org.jvnet.staxex>1.8.3</version.org.jvnet.staxex>
         <version.org.kohsuke.metainf-services>1.8</version.org.kohsuke.metainf-services>
         <version.org.opensaml.opensaml>3.3.1</version.org.opensaml.opensaml>
-        <version.org.ow2.asm>7.1</version.org.ow2.asm>
+        <version.org.ow2.asm>9.1</version.org.ow2.asm>
         <!-- WildFly overrides the picketbox version from core to use MRRC variants -->
         <version.org.picketbox>5.0.3.Final-redhat-00007</version.org.picketbox>
         <version.org.picketbox.picketbox-commons>1.0.0.final</version.org.picketbox.picketbox-commons>


ropalka (approved):
LGTM

bstansberry:
@scottmarlow WDYT?

scottmarlow:
Test results look good!

Dependencies on asm.asm module in WildFly appear to be:

1. io/smallrye/config/main/module.xml - Does anyone else need to review this dependency?
2. net/bytebuddy/main/byte-buddy-1.9.11.jar - ByteBuddy has its own shaded copy of ASM (https://github.com/raphw/byte-buddy/releases/tag/byte-buddy-1.10.21 contains JDK11 support but Hibernate ORM still uses ByteBuddy 1.9.11).
3. org/jboss/as/webservices/server/integration/main/module.xml - Does anyone else need to review this dependency?
4. org/hibernate/bytecodetransformer/main/module.xml - the unit test passed so we are good.
5. org/apache/cxf/ws-security/main/module.xml - Does anyone else need to review this dependency?
6. org/apache/cxf/main/module.xml - Does anyone else need to review this dependency?
7. org/apache/cxf/services-sts/main/module.xml - Does anyone else need to review this dependency?
8. org/apache/cxf/impl/main/module.xml9 - Does anyone else need to review this dependency?

bstansberry:
Thanks, @scottmarlow. I wasn't aware of the byte-buddy shading. But our module for it has no dep on the asm.asm module, so any calls that go into bytebuddy are going to load it's own shaded classes, so we're good there.

@jimma WDYT?  My look at CXF led me to believe the version we use already depends on ASM 9.1, so this basically brings us into alignment.

Smallrye Config 2.0.2's pom actually depends on ASM 9.0, so this PR would bring us to a closer match than we currently have.

jimma:
> @jimma WDYT? My look at CXF led me to believe the version we use already depends on ASM 9.1, so this basically brings us into alignment.

+1. We are good to align this version to 9.1.

PR #14267, opened 2021-05-03
Try ByteBuddy 1.10.22 as 1.10.19 or later is needed for SE 17 support

bstansberry:
just a test for now

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -365,7 +365,7 @@
         <version.org.apache.xalan>2.7.1.jbossorg-5</version.org.apache.xalan>
         <version.org.awaitility.awaitility>4.0.2</version.org.awaitility.awaitility>
         <version.org.bitbucket.jose4j>0.7.2</version.org.bitbucket.jose4j>
-        <version.org.bytebuddy>1.9.11</version.org.bytebuddy>
+        <version.org.bytebuddy>1.10.22</version.org.bytebuddy>
         <version.org.codehaus.jettison>1.4.0</version.org.codehaus.jettison>
         <version.org.cryptacular>1.2.4</version.org.cryptacular>
         <version.org.eclipse.jdt.core.compiler>4.6.1</version.org.eclipse.jdt.core.compiler>


bstansberry:
@scottmarlow @Sanne WDYT? To be conservative perhaps we could initially try this in WF Preview, which is where we're initially focusing our work for getting SE 17 to run.

scottmarlow:
@bstansberry @Sanne 
I tried build Hibernate ORM (5.3.x branch) locally with this change and it looks like code changes would be needed.  
```
diff --git a/gradle/libraries.gradle b/gradle/libraries.gradle
index bd5d0b7701..ed34fd2f30 100644
--- a/gradle/libraries.gradle
+++ b/gradle/libraries.gradle
@@ -26 +26 @@ ext {
-    byteBuddyVersion = '1.9.11'
+    byteBuddyVersion = '1.10.22'
```

```
 ./gradlew clean install PublishToMavenLocal test
> Task :hibernate-core:compileJava
/home/smarlow/work/hibernate5/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java:517: error: cannot find symbol
                        if ( access != null && access.loadSilent().value() == AccessType.PROPERTY ) {
                                                     ^
  symbol:   method loadSilent()
  location: variable access of type Loadable<Access>
/home/smarlow/work/hibernate5/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/EnhancerImpl.java:526: error: cannot find symbol
                        else if ( access != null && access.loadSilent().value() == AccessType.FIELD ) {
                                                          ^
  symbol:   method loadSilent()
  location: variable access of type Loadable<Access>
/home/smarlow/work/hibernate5/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/BiDirectionalAssociationHandler.java:166: error: cannot find symbol
                if ( access != null && access.loadSilent().value() == AccessType.FIELD ) {
                                             ^
  symbol:   method loadSilent()
  location: variable access of type Loadable<Access>
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
3 errors
```

Sanne:
Right we'rebuilding ORM 5.3 with BB 1.9.11, while ORM 5.4 (and now 5.5) have the latest 1.10.21.

This should only be a matter of backporting patches - and these changes have been widely tested in the Quarkus community as well so while I'm afraid it's actually many patches, at least it's not experimental code.

However please note that this older version of ORM isn't being tested on more recent versions of the JDK.
Do you need this? In case we should probably open a more general backport JIRA as there might be more than BB that needs to be done.

An alternative path is of course to upgrade WildFly to a more recent build, if backwards compatibility is less so of a concern in this context.

bstansberry:
@Sanne We've chatted elsewhere about this, but to not leave this comment thread hanging, I'll say here that I think testing ORM 5.3.x on SE 17 would be useful.

I'm not sure if you or @scottmarlow would like a WFLY to track bringing in an ORM 5.3 with the mentioned changes, as a place to have further discussions. I opened this draft PR to what would happen with our CI just from the BB upgrade, but I should close it soon as we know this branch won't be merged as-is.

Sanne:
thanks @bstansberry !

Some Hibernate issues to track:
 - https://hibernate.atlassian.net/browse/HHH-14610
 - https://hibernate.atlassian.net/browse/HHH-14611

@scottmarlow I'm assuming you will want to own the WFLY steps?

ropalka:
I successfully tested WildFly on JDK17 with ByteBuddy version 1.11.0 @bstansberry @scottmarlow (just FYI)

bstansberry:
Thanks @ropalka. FYI @Sanne when I put up this experiment PR there was a 1.11.0 but I used 1.10 under the theory of making the smallest possible minor version bump to get over the SE 17 issue. Plus 1.11 has no bug fix releases yet. But I apologize; I never noted the fact that 1.11 exists. Which to use is of course up to you guys.

Sanne:
I'd suggest sticking to 1.10.x unless there's a real need for 1.11 : micro upgrades have been mostly uneventful in the Hibernate & Quarkus communities but we've also been bitten occasionally so I'd prefer the conservative approach.

such upgrades trickle very quickly into RHBQ

Sanne:
BTW I had tested BB 1.11 as well and yes it seems to work fine, so it's definitely an option if you want to. Just for context, remember we also have to test the impact of upgrading such components on native-image builds.

ropalka:
Thanks for info @Sanne - I will downgrade to 1.10.x series and give it a try.

ropalka:
I confirm 1.10.22 works like a charm. BTW, we need also to upgrade Javassist from 3.23.2-GA to 3.28.0-GA @bstansberry .
What do you think about upgrading Javassist @Sanne ? Does it affect Hibernate somehow? Did you test it already?

Sanne:
@ropalka I couldn't test Javassist yet - we would need Gradle upgraded first.

FWIW I expect issues with it - we might want to support only bytebuddy when running on recent JVM versions, but I guess we can see how the testing plays out.

ropalka:
I don't know how well hibernate javassist integration is tested in WildFly test suite, but the test suite passed on 100% for me @Sanne

scottmarlow:
> I don't know how well hibernate javassist integration is tested in WildFly test suite, but the test suite passed on 100% for me 

The WildFly testsuite only tests with bytebuddy integration.  It is best to try new Javassist upgrades with the Hibernate ORM project testsuite.

scottmarlow:
> @ropalka I couldn't test Javassist yet - we would need Gradle upgraded first.
> 
> FWIW I expect issues with it - we might want to support only bytebuddy when running on recent JVM versions, but I guess we can see how the testing plays out.

+1 to only support bytebuddy when running on recent JVM versions.

PR #14270, opened 2021-05-04
WFLY-13044: fix extension matching in security constraints

michpetrov:
Issue: [WFLY-13044](https://issues.redhat.com/browse/WFLY-13044)

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/WarJACCService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/WarJACCService.java
@@ -836,10 +836,9 @@ public class WarJACCService extends JaccService<WarMetaData> {
          */
         public boolean isExtensionFor(PatternInfo other) {
             int offset = other.pattern.lastIndexOf('.');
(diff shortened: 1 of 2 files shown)


bstansberry:
This PatternInfo class looks like an excellent candidate for some unit testing. It's static and package protected so a test class can easily manipulate instances of it.

darranl:
@michpetrov The last comment was regarding adding some testing, is that something you can add to this PR?

michpetrov:
@darranl I can add a test for the affected method but I don't feel confident about testing the entire class.

bstansberry:
@michpetrov That's fine. The most important thing is demonstrating the bug, and then other scenarios for the method.  I read the JIRA and then saw the fix and without doing a lot of analysis I have no clue if it's right, wrong, other effects it might have etc. It seemed like a unit test with some simple comments explaining the cases being tested would solve that.

bstansberry (approved):
Great; thanks @michpetrov !

bstansberry:
Thanks, @michpetrov

PR #14273, opened 2021-05-04
 [WFLY-14752]: Make messaging subsystem dependencies on org.picketbox module optional.

ehsavoie:
* Updating the module.xml to reflect this.
* Add missing cpapability reference on legacy security domain in
  ActiveMQ Server.
* Cleaning the pom.xml a bit
* Moving some tests that were forgotten to the proper module

Jira: https://issues.redhat.com/browse/WFLY-14752


ee-feature-pack/galleon-content/src/main/resources/feature_groups/standalone-activemq-colocated.xml
```diff
@@ -17,6 +17,9 @@
         <feature spec="subsystem.messaging-activemq">
             <feature spec="subsystem.messaging-activemq.server">
                 <param name="server" value="default"/>
+                <!-- <param name="security-domain" value="GLN_UNDEFINED"/>-->
+                <unset param="security-domain"/>
+                <param name="elytron-domain" value="ApplicationDomain"/>
```
darranl, line 22: I am not sure if we should be changing default configurations as part of this exercise, I think I may create a set of issues to track the individual configurations that reference legacy security.

bstansberry: Note: this is now commented out; Emmanuel left it that way deliberately to make it easier to resurrect. But, at least as of 1 minute ago the main page for this PR made it look like it was still uncommented. But it's not; so Darran's point has been addressed.

bstansberry: @ehsavoie FYI, in ee-9/feature-pack an alternative version of this file is used. It doesn't have the 'unset' element, but the generated config file looks ok.

I'm only mentioning that in case there's something about that difference that concerns you.

bstansberry (approved):

PR #14274, opened 2021-05-05
[WFLY-14758] Don't try to use java.security.acl.Group in WS subsystem…

bstansberry:
… on JDK 14+

https://issues.redhat.com/browse/WFLY-14758


bstansberry:
@jimma This and a fix for JBWS-4133 resolve the issues I'm seeing with webservices when running WildFly Preview tests with SE 15.

jimma:
@bstansberry  This looks good.  Thanks for the fix.

PR #14279, opened 2021-05-06
[WFLY-14771] Define commons-io dependency in root pom

gaol:
Issue: https://issues.redhat.com/browse/WFLY-14771


gaol:
The version is still `2.5` to be consistent with what wildfly-core has currently. There is a [need to upgrade commons-io](https://github.com/wildfly/wildfly-core/pull/4400), which shall be done once [new release of commons-io](https://github.com/apache/commons-io/tags) gets out.

PR #14285, opened 2021-05-06
[WFLY-14773] Don't add the resteasy-rxjava2 module in the RSO deployer

kabir:
https://issues.redhat.com/browse/WFLY-14773

~EDIT: I noticed that the module has actually been around for a while, and was not added by me as I thought. So let's see what @asoldano says. If he disagrees, I will change it to just contain the change to the deployer~
EDIT 2: This has now been updated to just remove resteasy-rxjava2 from the dependencies added by the RSO deployer

### DIFF
--- a/microprofile/reactive-streams-operators-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/deployment/ReactiveStreamsOperatorsDependencyProcessor.java
+++ b/microprofile/reactive-streams-operators-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/deployment/ReactiveStreamsOperatorsDependencyProcessor.java
@@ -64,7 +64,6 @@ public class ReactiveStreamsOperatorsDependencyProcessor implements DeploymentUn
         moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "io.smallrye.reactive.converters.api", true, false, false, false));
         moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "io.smallrye.reactive.converters.rxjava2", true, false, true, false));
 
-        moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "org.jboss.resteasy.resteasy-rxjava2", true, false, true, false));
     }
 
 


bstansberry:
@ronsigal Any concerns?

bstansberry:
@kabir I think there is more to clean out:

```
$ git grep resteasy-rxjava2
ee-feature-pack/common/pom.xml:            <artifactId>resteasy-rxjava2</artifactId>
ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml:      <artifactId>resteasy-rxjava2</artifactId>
ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-rxjava2/main/module.xml:<module xmlns="urn:jboss:module:1.9" name="org.jboss.resteasy.resteasy-rxjava2">
ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-rxjava2/main/module.xml:        <artifact name="${org.jboss.resteasy:resteasy-rxjava2}" />
microprofile/reactive-streams-operators-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/streams/operators/deployment/ReactiveStreamsOperatorsDependencyProcessor.java:        moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, "org.jboss.resteasy.resteasy-rxjava2", true, false, true, false));
pom.xml:                <artifactId>resteasy-rxjava2</artifactId>
```

ronsigal:
Hi @bstansberry,

Not concern so much as curiosity. What is rxjava2 being replaced with? Maybe there's something we should do in RESTEasy. I've mentioned to @kabir the possibility of creating a RESTEasy module based on reactive-streams-operators that is like resteasy-rxjava2 but which abstracts out the reactive implementation.

-Ron

ronsigal:
Ok, so that validates my idea of a resteasy-reactive-streams-operators module, I think. It would replace resteasy-rxjava2, and the implementation would get service loaded. 

One thing occurs to me. resteasy-rxjava2, besides FlowableRxInvoker, also has a SingleRxInvoker which, of course, returns Singles. I don't remember seeing an abstraction of Single in RSO. Maybe I missed it?

kabir:
Adding the hold label so nobody accidentally merges it until Alessio who added the module back in 2018 has a say

asoldano (approved):

ronsigal:
Hey @kabir,

I notice that the smallrye-reactive-streams-operators implementation of ReactiveStreamsEngine uses io.reactivex.Flowable from rxjava2. Is that the reason?

I have another question. Someone wants to introduce into RESTEasy a version of a SingleRxInvoker that returns an interface instead of a Single (or Mono, or whatever), so he's using Publisher. He said, "Honestly, I don't know why the reactive-streams API doesn't have a first-class 'single-value publisher' concept, but it doesn't seem too. I'm sure they have a reason." I've been wondering the same thing. Do you know of any interface that could be used?

-Ron

kabir:
@ronsigal I think my prior comment was wrong, but as this dependency is added by the server (and Alessio pasted https://github.com/wildfly/wildfly-proposals/blob/master/jaxrs/WFLY-10636_EAP7-1010.adoc) we should keep it. BTW the SmallRye Reactive Streams Operators implementation has moved to Mutiny (SmallRyes implementation of reactive streams). This is the ReactiveStreamsEngine used there https://github.com/smallrye/smallrye-mutiny/blob/main/reactive-streams-operators/src/main/java/io/smallrye/mutiny/streams/Engine.java.

As for the Single/Mono part, I have no idea. I noticed the same thing as you said though :)

PR #14286, opened 2021-05-06
[WFLY-14775] Add test for Metrics and RBAC

jasondlee:
https://issues.redhat.com/browse/WFLY-14775

Add new test verifying proper RBAC operation when accessing metrics

### DIFF
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/metrics/rbac/MicroProfileMetricsRbacTestCase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/metrics/rbac/MicroProfileMetricsRbacTestCase.java
@@ -1,0 +1,177 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.test.integration.microprofile.metrics.rbac;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.URL;
+
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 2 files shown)


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/metrics/rbac/MicroProfileMetricsRbacTestCase.java
```diff
@@ -0,0 +1,170 @@
...
+        executeOperation(Util.createAddOperation(
+                PathAddress.parseCLIStyleAddress("/core-service=management/access=authorization/role-mapping=Monitor"))
+        );
+        final ModelNode addOperation = Util.createAddOperation(
+                PathAddress.parseCLIStyleAddress("/core-service=management/access=authorization/role-mapping=Monitor/include=monitor")
+        );
+        addOperation.get("alias").set("testSuite");
+        addOperation.get("name").set("testSuite");
+        addOperation.get("type").set("USER");
+        executeOperation(addOperation);
```
jamezp, line 160: Again not critical here as since it's only two operations it's not too slow, but you can do composite operations so only one request is made.

```java
final CompositeOperationBuilder builder = CompositeOperationBuilder.create();
builder.addStep(Operations.createAddOperation(
        Operations.createAddress("core-service", "management", "access", "authorization", "role-mapping", "Monitor")));

final ModelNode addOperation = Operations.createAddOperation(
        Operations.createAddress("core-service", "management", "access=authorization", "role-mapping", "Monitor", "include", "monitor")
);
addOperation.get("alias").set("testSuite");
addOperation.get("name").set("testSuite");
addOperation.get("type").set("USER");
builder.addStep(addOperation);
executeOperation(builder.build());
```

jamezp (commented):
As mentioned these changes are not critical at all. Just opinion based :)

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/metrics/rbac/MicroProfileMetricsRbacTestCase.java
```diff
@@ -0,0 +1,170 @@
...
+        ServerReload.reloadIfRequired(mgmtClient);
+    }
+
+    @After
+    public void after() throws Exception {
+        client.close();
+        setRbacStatus("simple");
+        setEndpointSecurityStatus(false);
+
+        ServerReload.reloadIfRequired(mgmtClient);
```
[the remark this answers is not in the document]
jasondlee: That's a good point. Save a little bit of time in an already long test run.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/metrics/rbac/MicroProfileMetricsRbacTestCase.java
jasondlee: I like this change, but it would require I duplicate executeOperation() to take and Operation, so I'll probably leave it as is. Leaves the code simpler at the cost of a few ms. :)

jamezp: FWIW what I usually do is:
```
private void executeOperation(final ModelNode op) {
    executeOperation(Operation.Factory.create(op)):
}
```
Then just have the one that accepts `Operation` do the logic.

jasondlee: Pfft. I like that. Incoming. :)

jamezp (approved):

PR #14289, opened 2021-05-07
WFLY-14778 Updated the class with a dependency on http-deprecated Elytron module to no longer rely on deprecated classes

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-14778

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/ApplicationSecurityDomainDefinition.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/ApplicationSecurityDomainDefinition.java
@@ -95,7 +95,14 @@ import org.wildfly.security.auth.server.SecurityDomain;
 import org.wildfly.security.cache.IdentityCache;
 import org.wildfly.security.http.HttpExchangeSpi;
 import org.wildfly.security.http.HttpServerAuthenticationMechanismFactory;
-import org.wildfly.security.http.impl.ServerMechanismFactoryImpl;
+import org.wildfly.security.http.basic.BasicMechanismFactory;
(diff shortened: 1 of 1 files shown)


bstansberry (commented):
@fjuma or @darranl or @fl4via Please review

@parsharma This doesn't feel right. It is converting from a new construction of an object per call to toHttpAuthenticationFactory into a single construction that will live for the life of the VM (including across reloads.). INSTANCE.delegate is effectively a constant.

If INSTANCE.delegate references an immutable object (and Elytron believes it will remain one as Elytron evolves) then maybe using a constant is ok. But in that case it should be a private static final field.

parsharma:
Thanks, @bstansberry for the suggestion.

undertow/src/main/java/org/wildfly/extension/undertow/ApplicationSecurityDomainDefinition.java
```diff
@@ -354,7 +366,7 @@ public String getSecurityDomainName() {
     }
 
     private static HttpAuthenticationFactory toHttpAuthenticationFactory(final SecurityDomain securityDomain, final String realmName) {
-        final HttpServerAuthenticationMechanismFactory mechanismFactory = new FilterServerMechanismFactory(new ServerMechanismFactoryImpl(), SERVLET_MECHANISM);
+        final HttpServerAuthenticationMechanismFactory mechanismFactory = new FilterServerMechanismFactory(delegate, SERVLET_MECHANISM);
```
fjuma, line 369: @parsharma Thanks for working on this! Instead of using `delegate` here, the new construction can be done here.

fjuma (approved):

PR #14297, opened 2021-05-12
[WFLY-14691] Tests the data source statistics attribute MaxWaitCount during waiting requests.

dcihak:
Test enables statistics, sets min/max pool size on the ExampleDS datasource and tests the data source statistics attribute MaxWaitCount during waiting requests.

[WFLY-14691](https://issues.redhat.com/browse/WFLY-14691)


dcihak:
Please do not merge now, waits for the fix (https://github.com/ironjacamar/ironjacamar/pull/720) to be merged.

rhusar:
BTW @dcihak You can convert the PR to a draft which is not mergeable. The other option is to add a test as @ Ignore-d and reenable on component upgrade.

bstansberry:
@dcihak @moulalis @tadamski I see various tests of pool statistics in IronJacamar in the core/src/test/java/org/jboss/jca/core/connectionmanager/pool package. This seems like something that should be tested there, not in WildFly.

dcihak:
@rhusar Thank's for the information. @bstansberry I will move the testcase to the Ironjacamar TS if it's more suitable there.

bstansberry:
Thanks, @dcihak. If this can be tested in IJ instead I'd prefer it.

dcihak:
Retest this please.

dcihak:
@bstansberry @moulalis @tadamski org.jboss.as.test.integration.jca.statistics.DataSourceJdbcStatisticsTestCase and org.jboss.as.test.integration.jca.statistics.DataSourcePoolStatisticsTestCase seem to be very similar coverage, that's why I supposed this is the right place.

bstansberry:
@dcihak That test looks to be a simple smoke test that the management API is exposing statistics. This is a much more complex test that is relying on fairly complex timing, and which will take a fair bit of time to run, both of which add fragility to the WF testsuite. If there are regressions in the behavior being tested they are highly unlikely to be in the WildFly code exposing the IJ numbers. Those regressions would not be found when testing IJ; they'd be noticed only when an IJ upgrade is tested in WF, which is too late.

dcihak:
@bstansberry Thank's for clarification. Nevertheless, I will have to move the tests into our [internal testsuite](https://gitlab.hosts.mwqe.eng.bos.redhat.com/jbossqe-eap/tests-jca) as there is not a sufficient support in the ironjacamar repo or the adjustments wouldn't be trivial.

PR #14307, opened 2021-05-19
[WFLY-14802] Upgrade jboss-ejb-client to 4.0.40.Final

tadamski:
https://issues.redhat.com/browse/WFLY-14802


jamezp (approved):
All tests passed on #4308

PR #14314, opened 2021-05-25
[WFLY-14815] Temporarily disable broken assertions so Undertow can be…

fl4via:
… upgraded (see UNDERTOW-1837)

Jira: https://issues.redhat.com/browse/WFLY-14815
Another PR will follow to fix the test, the PR will only pass with Undertow upgrade (containing UNDERTOW-1837 fix)

PR #14319, opened 2021-05-26
[WFLY-14815] Fix ForwardedHandlerTestCase to comply with UNDERTOW-1837

fl4via:
Jira: https://issues.redhat.com/browse/WFLY-14815

This is the second part of the fix required to make ForwardedHandlerTestCase compliant with UNDERTOW-1837.
The first part is #14314 
Requires upgrade of WildFly Core with Undertow upgrade: https://github.com/wildfly/wildfly-core/pull/4604

### DIFF
--- a/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/ForwardedHandlerTestCase.java
+++ b/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/ForwardedHandlerTestCase.java
@@ -6,6 +6,7 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUB
 import static org.junit.Assert.assertEquals;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.URL;
 
 import org.apache.http.Header;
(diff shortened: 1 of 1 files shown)


bstansberry:
I'm going to kill the CI jobs on this as we know the tests won't pass yet.

spyrkob:
@fl4via should the expected values be set to something like 
```
final InetAddress address = InetAddress.getByName(url.getHost());
final String localAddrName = address.getHostName();
final String localAddr = address.getHostAddress() + ":" + url.getPort();
``` 
to get the values of address server is bound to?

fl4via:
@spyrkob possibly, I'm going to give it a try.

spyrkob:
The SM testsuite fails because resolving hostname requires SocketPermission(host, "resolve")). Since that's expected behaviour is it OK to just add the permission to the test deployment?

bstansberry:
@spyrkob @fl4via Yes, adding the permission to the deployment is the correct thing to do.

fl4via:
@spyrkob  thanks for your commit! @bstansberry I hope the tests will pass now :-)

PR #13687, opened 2020-11-04
[WFLY-13762] [Community] Tests for SSLContext to support delegation to alternate instances based on peer information.

skyllarr:
Smoke tests for https://issues.redhat.com/browse/WFLY-13762

**Requires**: https://issues.redhat.com/browse/WFCORE-5101 / https://github.com/wildfly/wildfly-core/pull/4311

____

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #14333, opened 2021-06-02
[WFLY-13762] Add documentation for SSLContext to support delegation to alternate instances based on peer information.

skyllarr:
https://issues.redhat.com/browse/WFLY-13762


fjuma:
@Skyllarr Just noticed that there is a separate PR with tests for WFLY-13762 (https://github.com/wildfly/wildfly/pull/13687). Since both of these PRs are for the same feature, it would be good to combine this PR with https://github.com/wildfly/wildfly/pull/13687 so both the tests and the docs are in one place.

bstansberry:
@Skyllarr I'm going to close this in an effort to get the PR queue down to a manageable size. Please add this commit to #13687.

PR #14335, opened 2021-06-02
[WFLY-14846] Documentation for automatic registration of client side / JVM wide default SSLContext

skyllarr:
https://issues.redhat.com/browse/WFLY-14846

### DIFF
--- a/docs/src/main/asciidoc/_client-guide/authentication-client.adoc
+++ b/docs/src/main/asciidoc/_client-guide/authentication-client.adoc
@@ -292,6 +292,47 @@ The individual match elements are: -
 * <match-domain-name name="..."/>:: Matches where the protocol of the URI is 'domain' and the scheme specific part of the URI is the name specified within this match element.
  <match-abstract-type name="..." authority="..." />:: Matches the abstract type and/or authority against the values specified within this match element.
 
+=== Elytron client default SSL context security provider
+
+You can register the `org.wildfly.security.auth.client.ElytronClientDefaultSSLContextProvider` security provider. This java security provider loads Elytron client configuration and provides an initialized default SSL context. You can make use of this SSL context by calling `SSLContext.getDefault()` if this provider is registered with high enough priority.
+
+The provider loads the SSL context from either current authentication context obtained from classpath, or from authentication context obtained from file whose path is passed into the security provider either programmatically or as an argument in `java.security` file. Argument passed to the provider directly has precedence over authentication context from classpath.
+
+The SSL context configured to match all rules is the one that will be initialized and returned by this provider:
+
+[source,xml,options="nowrap"]
+----
+<rule use-ssl-context="ssl-context-for-client-provider" />
+----
+
+Or:
+
+----
+myAuthenticationContext.withSsl(MatchRule.ALL, myDefaultSslContext);
+----
+
+Example of dynamic configuration of client default SSL context provider with configuration file path provided:
+
+[source,java,options="nowrap"]
+----
+Security.insertProviderAt(new ElytronClientDefaultSSLContextProvider(CONFIG_FILE_PATH), 1);
+----
+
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/_client-guide/authentication-client.adoc
```diff
@@ -292,6 +292,47 @@ The individual match elements are: -
 * <match-domain-name name="..."/>:: Matches where the protocol of the URI is 'domain' and the scheme specific part of the URI is the name specified within this match element.
  <match-abstract-type name="..." authority="..." />:: Matches the abstract type and/or authority against the values specified within this match element.
 
+=== Elytron client default SSL context security provider
+
+You can register `org.wildfly.security.auth.client.ElytronClientDefaultSSLContextProvider` security provider. This java security provider loads Elytron client configuration and provides initialized default SSL context that is available by calling `SSLContext.getDefault()`.
```
fjuma, line 297: Very minor comments:

s/You can register/You can register the

s/provides initialized/provides an initialized

Skyllarr: @fjuma Fixed. Thank you.

docs/src/main/asciidoc/_client-guide/authentication-client.adoc
```diff
@@ -292,6 +292,47 @@ The individual match elements are: -
...
+security.provider.1=org.wildfly.security.auth.client.ElytronClientDefaultSSLContextProvider CONFIG_FILE_PATH
+----
+
+When you have default SSL context configured with this provider, you can make use of it by calling:
+
+```
+SSLContext.getDefault();
+```
+
+If no default SSL context is configured or there is no configuration present, the provider will be ignored. The provider will also be ignored if the configuration is invalid and creates infinite loop. This can happen if the default SSL context is configured to be the default SSL context.
```
fabiobrz, line 334: This last sentence seems a bit unclear to me, can we do something to make it clearer when such a loop would happen maybe?

docs/src/main/asciidoc/_client-guide/authentication-client.adoc
```diff
@@ -292,6 +292,47 @@ The individual match elements are: -
 * <match-domain-name name="..."/>:: Matches where the protocol of the URI is 'domain' and the scheme specific part of the URI is the name specified within this match element.
  <match-abstract-type name="..." authority="..." />:: Matches the abstract type and/or authority against the values specified within this match element.
 
+=== Elytron client default SSL context security provider
+
+You can register the `org.wildfly.security.auth.client.ElytronClientDefaultSSLContextProvider` security provider. This java security provider loads Elytron client configuration and provides an initialized default SSL context that is available by calling `SSLContext.getDefault()`.
```
fabiobrz, line 297: I'd add a note here, to make clear that the provider needs to have high enough priority, WDYT?

fabiobrz (changes requested):
HI @Skyllarr - and thanks for this MR.

I've dropped a couple of comments, let me know what you think.

docs/src/main/asciidoc/_client-guide/authentication-client.adoc
Skyllarr: @fabiobrz I specified that the provider must be registered with high enough priority.

docs/src/main/asciidoc/_client-guide/authentication-client.adoc
Skyllarr: @fabiobrz I agree and I removed the sentence about looping for now. I will get back to it after other deliverables are done today. In the worst case. it can also be added just as a documentation update after the feature is in, as it is not such an important information for the feature.

fjuma (approved):

fabiobrz (approved):
Thanks for addressing my comments @Skyllarr 
Changes LGTM, hence I am approving the PR. 
Let me know in case you add anything regarding the loop managment case.

PR #14337, opened 2021-06-03
[WFLY-14849] The java8-test surefire execution should not use the mod…

bstansberry:
…ular.jdk.args prop, removing the need to mess with its value

https://issues.redhat.com/browse/WFLY-14849

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -203,7 +203,8 @@
         <!-- Surefire args -->
         <surefire.extra.args></surefire.extra.args>
         <surefire.jpda.args></surefire.jpda.args>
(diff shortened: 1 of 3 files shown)


bstansberry:
@pferraro @ropalka Please review.

This is a different approach to what we were discussing last week re getting java8-test to work while still allowing SE 16 builds to work as well.

With an SE 16 JDK this works fine for me:

mvn install -pl clustering/marshalling/protostream -am -Djava8.home=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home

https://ci.wildfly.org/viewLog.html?buildId=258115& is a custom run of the JDK 17 test job that's been reporting ~ 100 failures from this module.

pom.xml
```diff
@@ -9654,7 +9655,19 @@
                         </systemProperties>
                         <argLine>${surefire.system.args}</argLine>
                     </configuration>
-
+                    <executions>
```
ropalka, line 9658: This is exactly I was looking for @bstansberry !

ropalka (approved):

pferraro (approved):

PR #14344, opened 2021-06-07
WFLY-14861 Interrupting session expiration task on shutdown can leave orphaned entries in cache

pferraro:
https://issues.redhat.com/browse/WFLY-14861


rhusar (approved):
Looks good.

PR #14345, opened 2021-06-07
WFLY-14862 Session invalidation can leave stale cache entries for non-tx caches if terminated mid expiration

pferraro:
https://issues.redhat.com/browse/WFLY-14862


pferraro:
It's unclear why the changes made here would cause failures in the hotrod session expiration tests.  Putting on hold while I investigate.

rhusar (approved):
👍

PR #14346, opened 2021-06-07
[WFLY-14863] Undertow o.w.e.undertow.security.AccountImpl setRoles pa…

boris-unckel:
…ram self-assign

Fixes https://issues.redhat.com/browse/WFLY-14863

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/security/AccountImpl.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/security/AccountImpl.java
@@ -75,11 +75,6 @@ public class AccountImpl implements Account, Serializable {
         this.roles.addAll(roles);
     }
 
-    void setRoles(final Set<String> roles) {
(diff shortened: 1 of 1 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/security/AccountImpl.java
```diff
@@ -77,7 +77,7 @@ public AccountImpl(final Principal principal, Set<String> roles, final Object cr
 
     void setRoles(final Set<String> roles) {
```
darranl, line 78: Having a quick look I don't see anywhere this is used - can it be removed instead?

boris-unckel: Thanks for your feedback. I have removed the method as you suggested.

bstansberry (approved):

PR #14350, opened 2021-06-07
WFLY-14868 Coordinator change can cause missing session expiration tasks

pferraro:
https://issues.redhat.com/browse/WFLY-14868


bstansberry:
@pferraro Is this urgent for 24? Seems like it could use bake (plus a review from someone who understands.)

pferraro:
@bstansberry This should be included in 24.  @rhusar Can you review asap?

rhusar (approved):
Looks good to me.

PR #14355, opened 2021-06-09
WFLY-14877 Do not allow application to create a new session or change the identifier of a session after response is committed

pferraro:
https://issues.redhat.com/browse/WFLY-14877

### DIFF
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
@@ -25,6 +25,7 @@ import io.undertow.security.api.AuthenticatedSessionManager.AuthenticatedSession
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.session.SessionConfig;
 import io.undertow.server.session.SessionListener.SessionDestroyedReason;
+import io.undertow.servlet.UndertowServletMessages;
 import io.undertow.servlet.handlers.security.CachedAuthenticatedSessionHandler;
 
 import java.time.Duration;
@@ -297,6 +298,10 @@ public class DistributableSession implements io.undertow.server.session.Session 
 
     @Override
     public String changeSessionId(HttpServerExchange exchange, SessionConfig config) {
+        // Workaround for UNDERTOW-1902
+        if (exchange.isResponseStarted()) { // Should match the condition in io.undertow.servlet.spec.HttpServletResponseImpl#isCommitted()
+            throw UndertowServletMessages.MESSAGES.responseAlreadyCommited();
+        }
         Session<Map<String, Object>> oldSession = this.getSessionEntry().getKey();
         SessionManager<Map<String, Object>, Batch> manager = this.manager.getSessionManager();
         String id = manager.createIdentifier();
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
@@ -65,15 +65,17 @@ public class DistributableSessionManager implements UndertowSessionManager, Long
     private final SessionManager<Map<String, Object>, Batch> manager;
     private final RecordableSessionManagerStatistics statistics;
     private final StampedLock lifecycleLock = new StampedLock();
+    private final boolean allowOrphanSession;
 
     // Guarded by this
     private OptionalLong lifecycleStamp = OptionalLong.empty();
 
-    public DistributableSessionManager(String deploymentName, SessionManager<Map<String, Object>, Batch> manager, SessionListeners listeners, RecordableSessionManagerStatistics statistics) {
-        this.deploymentName = deploymentName;
-        this.manager = manager;
-        this.listeners = listeners;
-        this.statistics = statistics;
+    public DistributableSessionManager(DistributableSessionManagerConfiguration config) {
+        this.deploymentName = config.getDeploymentName();
+        this.manager = config.getSessionManager();
+        this.listeners = config.getSessionListeners();
+        this.statistics = config.getStatistics();
+        this.allowOrphanSession = config.isOrphanSessionAllowed();
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
```diff
@@ -154,6 +154,9 @@ public void accept(HttpServerExchange exchange) {
         if (config == null) {
             throw UndertowMessages.MESSAGES.couldNotFindSessionCookieConfig();
         }
+        if (exchange.isResponseComplete()) {
+            throw UndertowMessages.MESSAGES.responseComplete();
+        }
```
msfm, line 159: Same to the above comment.

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSession.java
```diff
@@ -280,6 +281,9 @@ public void invalidate(HttpServerExchange exchange) {
 
     @Override
     public String changeSessionId(HttpServerExchange exchange, SessionConfig config) {
+        if (exchange.isResponseComplete()) {
+            throw UndertowMessages.MESSAGES.responseComplete();
+        }
```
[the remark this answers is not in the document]
pferraro: @msfm Thanks for the feedback.

> Isn't this better to use exchange.isResponseStarted() instead of exchange.isResponseComplete()?

Yes - you are absolutely right.  We want the condition to match the implementation of ServletResponse.isCommitted().  Good catch.

> Also, I do not know if UndertowMessages is good to use here

This generally fine, so long as the context of the message is exactly the same.

> Even if it's ok to use Undertow logger, UndertowMessages.MESSAGES.responseAlreadyStarted() (The response has already been startd) or UndertowServletMessages.MESSAGES.responseAlreadyCommited() (Response already commited) might be suitable log message?

I think UndertowServletMessages.MESSAGES.responseAlreadyCommited() is most appropriate here - it is the same ISE message used for a duplicate invocation of sendRedirect(...) or sendError(...).

bstansberry:
@pferraro This has a conflict because I merged the PR for the Critical bug.

If this is needed for WF 24 we should get it in Monday (Jun 14).

pferraro:
@bstansberry This one is urgent for downstream, not WF24 specifically.  Merge conflict fixed.

bstansberry:
@pferraro I'm going to stick the 25 label on this for now, although it's fine for 24.0.1 if you want to port it to 24.x once that's created.

pferraro:
@bstansberry That's fine.

bstansberry:
@pferraro  This LGTM but that's based on 

1) My understanding based on a comment I read somewhere that this problem only affects distributable webapps. So the solution doesn't need to account for non-distributable. (Even if it does, that could be a separate fix.)

2) That the "jboss.web.tolerate-session-creation-after-redirect" prop used to disable the fix is ok with @fl4via with respect to whatever she might do for the same purpose on UNDERTOW-1902.

pferraro:
@bstansberry Technically, this *can* effect non-distributable web applications (hence UNDERTOW-1902), but only if a web application both attempts to create a new session after the response is closed (e.g. after HttpServletResponse.sendRedirect/sendError(...)) *and* if the created session has no timeout (e.g. <session-timeout>0</session-timeout>).  Given the low probability that an application would use both these conditions, the issue is less significant for non-distributed web applications.

Ideally, this condition would be handled within Undertow itself.  Thus, this fix is really a workaround.  The "jboss.web.tolerate-session-creation-after-redirect" system property is a temporary measure and only exists to maintain compatibility (see linked jiras) for those users that relied on non-spec behavior.  I anticipate that this will be superseded by a proper management attribute in the Undertow subsystem once a fix for UNDERTOW-1902 is available to WildFly.

pferraro:
Rebased on master.  Updated terminology to match proposed UNDERTOW-1902 fix.

rmartinc:
@pferraro @bstansberry We are preparing the next downstream CP. In order to integrate this into the CP an approval in upstream is at least needed. We have time until the beginning of the next week. Thanks!

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManagerFactory.java
```diff
@@ -49,6 +51,7 @@
  * @author Paul Ferraro
  */
 public class DistributableSessionManagerFactory implements io.undertow.servlet.api.SessionManagerFactory {
+    private static final String TOLERATE_SESSION_CREATION_AFTER_REDIRECT_PROPERTY = "jboss.web.tolerate-session-creation-after-redirect";
```
fl4via, line 54: I'm assuming this property is a temporary property so we can fix it for now until the config can be added to the model, am I right?

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
```diff
@@ -150,6 +152,13 @@ public void accept(HttpServerExchange exchange) {
         }
         // Workaround for UNDERTOW-1902
         if (exchange.isResponseStarted()) { // Should match the condition in io.undertow.servlet.spec.HttpServletResponseImpl#isCommitted()
+            // The servlet specification mandates that an ISE be thrown here
```
fl4via, line 155: @pferraro If I get it correctly, this workaround will go after UNDERTOW-1902 is merged and released, is that right?

fl4via (approved):
It looks okay to me, but I need to clarify some points

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManagerFactory.java
pferraro: Yes - I anticipate replacing this an Undertow management attribute that configures: https://github.com/pferraro/undertow/blob/master/servlet/src/main/java/io/undertow/servlet/api/DeploymentInfo.java#L1424

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
pferraro: After UNDERTOW-1902, this method will not be invoked unless DeploymentInfo.isOprhanSessionAllowed() returns true (so we do not need to worry about throwing an ISE), however, the distributed session manager still needs to return an OrphanSession in this case, since the session will never be scheduled for expiration.

rmartinc:
After chatting with @pferraro we have included this in the next CP. It's approved by @fl4via, the fix is slightly different for 7.3 and Paul doesn't expect big changes in this upstream PR. Thanks everybody!

PR #14360, opened 2021-06-10
[WFLY-13679] Make legacy security optional for "org.wildfly.iiop-openjdk"

darranl:
https://issues.redhat.com/browse/WFLY-13679

This also corrects the configuration in WildFly Preview which presently assumes it is using legacy security even though the subsystem is not provisioned.

### DIFF
--- a/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
+++ b/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
@@ -227,6 +227,11 @@ link:#gal.h2-datasource[h2-datasource] +
 |
 link:#gal.base-server[base-server] +
 
+|[[gal.iiop-openjdk]]iiop-openjdk
+|Support for IIOP
+|
+link:#gal.naming[naming] +
+
 |[[gal.io]]io
 |Support for XNIO workers and buffer pools.
 |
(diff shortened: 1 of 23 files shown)


darranl:
@chengfang / @tadamski Could you please review this PR?

The overall approach is that all configurations that require PicketBox classes have a requirement on the legacy security subsystem being present and we know if that is present PicketBox will be present - other than that we can assume the module dependency is optional.

chengfang:
Looking at failed CI jobs, those errors seem to be real:
```
org.jboss.as.cli.CommandLineException:
{
"outcome" => "failed",
"failure-description" => "WFLYCTL0367: Cannot remove capability 'org.wildfly.legacy-security' as it is required by other capabilities:
capability 'org.wildfly.iiop' requires it for attribute 'security' at address '/subsystem=iiop-openjdk'",
```

darranl:
I have just pushed again, hopefully this will address the failure that was appearing.

darranl:
Hi @chengfang / @tadamski CI is now green so I think we are ready to continue.

tadamski (commented):
LGTM but shouldn't iiop-openjdk (non legacy) feature group be set by default in all the configurations?

darranl:
At this time the intent was just to get to the point PicketBox can be flagged as being optional.

There will be a follow up task later where we change all the default configurations to be Elytron only but we have some other tasks to work through first.

tadamski (approved):
ahh ok, I'm approving then

PR #14369, opened 2021-06-14
[WFLY-14892] Flag io.smallrye.config module as jboss.api=private

jmesnil:
JIRA: https://issues.redhat.com/browse/WFLY-14892


jmesnil:
I raised the priority of this bug to critical but the omission has been here from the start.
That't be great if we could fix it in WildFly 24

PR #14375, opened 2021-06-15
[WFLY-14898] transactions subsystem log-store transactions attributes are read-only

ochaloup:
https://issues.redhat.com/browse/WFLY-14898


bstansberry:
@mmusgrov Please review.

@ochaloup I think just deleting ParticipantWriteAttributeHandler is ok if it's not used. It's a public class but this module is not meant for use by anyone else.

mmusgrov (approved):
It's a good change.

PR #14376, opened 2021-06-15
[WFLY-14899 - Document env var expression resolution

kabir:
https://issues.redhat.com/browse/WFLY-14899

Implementation: https://github.com/wildfly/wildfly-core/pull/4633 / https://issues.redhat.com/browse/WFCORE-5464

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/management-api/Description_of_the_Management_Model.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/management-api/Description_of_the_Management_Model.adoc
@@ -108,11 +108,14 @@ values described.
 attribute may be of type `ModelType.EXPRESSION`, instead of its standard
 type (see `type` and `value-type` above for discussion of an attribute's
 standard type.) A value of ModelType.EXPRESSION contains a
-system-property substitution expression that the server will resolve
+system-property or environment variable
+substitution expression that the server will resolve
 against the server-side system property map before using the value. For
 example, an attribute named max-threads may have an expression value of
 `${example.pool.max-threads:10`} instead of just 10. Default value if
 not present is false.
+See the <<Expression_Resolution, Expression Resolution>> subsection
+for a more thorough description.
 * `required` – boolean – true if the attribute must have a defined value
 in a representation of its portion of the model unless another attribute
 included in a list of `alternatives` is defined; false if it may be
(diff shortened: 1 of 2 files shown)


jmesnil:
I'm not sure if we can address this in your PR but this documentation would better belong to a "user" guide than an admin guide.

The intent of that feature is for the user that deploys their application and wants to tweak it. I'm not sure how easy it will be for users to find this out with our current doc structure.

jmesnil (approved):
doc looks good to me

docs/src/main/asciidoc/_admin-guide/management-api/Description_of_the_Management_Model.adoc
```diff
@@ -516,3 +522,27 @@ operation will go ahead and restart the runtime service. Otherwise
...
+* First we check if there is a system property with the name
+`my.example-expr`.
+If there is, we use its value as the result of the resolution.
+If not, we continue checking the next locations.
+* We convert the name `my.example-expr` to upper case, and replace all
+non-alphanumeric characters with underscores, ending up with
+`MY_EXAMPLE_EXPR`. We check if there is an environment variable with that
+name. If there is, we use its value as the result of the resolution.
+If not, we continue checking the next location.
+* We append `env.` onto the original name, i,e, `env.my.example-expr`
```
vsalbaba, line 540: Does this mean checking for literally `env.my.example-expr` environment variable, or is the SCREAMING_SNAKE_CASE conversion applied and we check for `ENV_MY_EXAMPLE_EXPR`

kabir: Actually this is slightly wrong, and was what always happened.

What we do is check if the original name starts with 'env.', and if it does we drop the suffix, and use the remainder as the env var that we look up.  So `env.blah` becomes `blah` and `env.HI_THERE` becomes `HI_THERE`

vsalbaba: So maybe a clarification is in order, so that it is in line with analysis [1] 
https://github.com/wildfly/wildfly-proposals/blob/3f1ec1e5ae9aadc2cd0b7dc34c4e7fe4327e3086/management/WFCORE-5464_Check_Env_Vars_On_Expression_Resolution.adoc#hard-requirements

kabir: I pushed a fix

vsalbaba: Much clearer now, thank you

PR #14380, opened 2021-06-16
[WFLY-14894] adding admin guide for transactions subsystem

ochaloup:
https://issues.redhat.com/browse/WFLY-14894

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Transactions.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Transactions.adoc
@@ -1,0 +1,425 @@
+[[Transactions_Subsystem]]
+= Transactions subsystem configuration
+
+*Required extension:*
+
+[source,xml,options="nowrap"]
+----
+<extension module="org.jboss.as.transactions"/>
+----
+
+*Basic subsystem configuration example:*
+
+[source,xml,options="nowrap"]
+----
+<subsystem xmlns="urn:jboss:domain:transactions:6.0">
+  <core-environment node-identifier="${jboss.tx.node.id:1}">
+    <process-id>
+      <uuid/>
+    </process-id>
+  </core-environment>
+  <recovery-environment socket-binding="txn-recovery-environment"
+                        status-socket-binding="txn-status-manager"/>
+   <coordinator-environment statistics-enabled="${wildfly.transactions.statistics-enabled:${wildfly.statistics-enabled:false}}"/>
+  <object-store path="tx-object-store"
+                relative-to="jboss.server.data.dir"/>
+</subsystem>
+----
+
+== Transaction subsystem configuration
+
+Transaction subsystem configures the behaviour of the transaction manager.
+https://narayana.io/[Narayana] is the transaction manager used in WildFly.
+The second component configured within the subsystem is
+https://github.com/wildfly/wildfly-transaction-client[WildFly Transaction Client]
+(the WFTC serves as an abstract layer to work with the transactional context).
+
+=== Configuration of Narayana component
+
+The structure of the transaction subsystem follows the structure of Narayana component.
+Narayana defines a separate configuration bean for every
+https://jbossts.blogspot.com/2018/01/narayana-periodic-recovery-of-xa.html#configuration[internal module].
+For example any configuration related to Narayana core is available through beans
+https://github.com/jbosstm/narayana/blob/master.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java[CoordinatorEnvironmentBean] and
+https://github.com/jbosstm/narayana/blob/master.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java[CoreEnvironmentBean],
+for JTA processing it is
+https://github.com/jbosstm/narayana/blob/master.Final/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java[JTAEnvironmentBean],
+for the transaction recovery setup it's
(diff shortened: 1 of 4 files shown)


ochaloup:
@mmusgrov @mayankkunwar @jmfinelli 
hi, would be somebody so kind to look at the documentation update in this PR and provide me with a feedback?

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Transactions.adoc
```diff
@@ -0,0 +1,429 @@
...
+  <recovery-environment socket-binding="txn-recovery-environment"
+                        status-socket-binding="txn-status-manager"/>
+   <coordinator-environment statistics-enabled="${wildfly.transactions.statistics-enabled:${wildfly.statistics-enabled:false}}"/>
+  <object-store path="tx-object-store"
+                relative-to="jboss.server.data.dir"/>
+</subsystem>
+----
+
+[cols=",",options="header"]
+|====
```
mmusgrov, line 30: We need to be consistent about how we define the model attributes and XML elements across our documentation. For example the Model Reference guide and the subsystem guide should be more or less identical. I only checked on or two of the following attributes but the ones I picked are sufficiently different to be worthwhile aligning.

You could even check with the wildfly team to see if it's possible to derive the text (for the model reference, the subsystem, and any other docs that refer to this config) from a common source.

ochaloup: The point of the documentation is to provide to user information on how to configure the WildFly. There could  be seen this approach in all the subsystem configuration guides: https://docs.wildfly.org/23/Admin_Guide.html#Subsystem_configuration

The Red Hat EAP tries to talk mostly only about the JBoss CLI commands but this is WildFly where we can provide a broader information - e.g. see the Agroal documentation https://docs.wildfly.org/23/Admin_Guide.html#agroal (or any other subsystem admin guide).

@mmusgrov 
I agree that the table should be more explicitly talks about the XML and CLI. I added a new column where the CLI name is referred. Would that be better that way?

> You could even check with the wildfly team to see if it's possible to derive the text from a common source.
I don't think it's the point of the documentation. The documentation should explain the topic in much more details and with examples which is not available a thing for the model reference.

mmusgrov: Take the first property/XML element, node-identifier. There are three distinct definitions of what this property is for (and it is not clear on a first reading whether they are the same or are subtly different):

*WildFly documentation:*
`Unique identification of the transaction manager (the WildFly) instance. Important for data consistency when running periodic recovery.`

*Model Reference:*
`Used to set the node identifier on the core environment. Each Xid that Transaction Manager creates will have this identifier encoded within it and ensures Transaction Manager will only recover branches which match the specified identifier. It is imperative that this identifier is unique between Application Server instances which share either an object store or access common resource managers.`

*EAP documentation:*
`The node identifier for the transaction manager. If this option is not set, you will see a warning upon server startup. This option is required in the following situations:
    - For JTS to JTS communications
    - When two transaction managers access shared resource managers
    - When two transaction managers access shared object stores
The node-identifier must be unique for each transaction manager as it is required to enforce data integrity during recovery. The node-identifier must also be unique for JTA because multiple nodes may interact with the same resource manager or share a transaction object store.`

All of these different definitions refer to the same property or element and they have the same meaning, ie they all set the arjuna CoreEnvironmentBean#nodeIdentifier property and there is no other meaning for it.

Or take an easier one (the second in the list), process-id-uid:

*WildFly documentation:*
`Strategy how transaction identifiers are generated. uuid means generation based on the process id.`

*Model Reference:*
`Indicates whether the transaction manager should use a UUID based process id.`

*EAP documentation:*
`Set to true to use the process identifier to create a unique identifier for each transaction. Otherwise, the socket-based mechanism is used. Defaults to true. See process-id-socket-max-ports for more information. To enable process-id-socket-binding, set process-id-uuid to false.`

Again all of these definitions refer to the same underlying arjuna property and they all have the same effect. Similarly for the other properties.

ochaloup: Yes, the all refer to the same underlying arjuna property and all of them refers the same thing, the description has the same meaning but said in other words with different level of detail.

The model reference normally do not contain a big level of detail and the detailed description is in documentation.

There are differences in EAP and WildFly documentation in all levels as EAP documentation is not written by developers and not maintained by community. I don't think there is a great reason trying to align them. Community works in a different pace and in different means than the EAP product doc.
I'm not sure if I'm permitted to take the documentation snippets just from EAP doc and paste them directly to community doc. If I can I can copy some parts but I would phrase differently some other parts and I want to do so as this is a community doc written by developers.

Does this perspective make sense?

I would like to talk here rather about what is wrong in the text or how to phrase the text better.

mmusgrov (changes requested):
Having different definitions for the same thing just creates confusion.
Can you cut and paste the definitions from the model reference please.

ochaloup:
@mmusgrov ok. I can take the definition from the model reference and then elaborate on the meaning in next few sentences. Would that be fine?

mmusgrov:
> @mmusgrov ok. I can take the definition from the model reference and then elaborate on the meaning in next few sentences. Would that be fine?

No, we don't want the definitions to diverge and elaboration will not help with that goal.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Transactions.adoc
ochaloup: I reworked the PR so I'm resolving this comment now. Please reopen if required.

ochaloup:
@mmusgrov I reworked the PR based on our discussion. The attribute definitions were changed in the model, the documentation guide was simplified to contain examples and few details on how the things work.
Please take a look and review the current version. Thank you.

mmusgrov (approved):
Even though the new definitions are a little different from the model definitions, the new the definitions are much clearer than the ones provided in the model definition. Thanks.

ochaloup:
@bstansberry @darranl can you please review this PR if it can be merged?

PR #14386, opened 2021-06-18
[WFLY-14922] Remove picketlink extension

bstansberry:
https://issues.redhat.com/browse/WFLY-14922


pferraro:
This looks good to me so far - CI will complain if anything is missing.

darranl (changes requested):
As a subsystem I was assuming this needed to live on as a skeleton so older domains could still be managed?

Has there been any discussion yet with @emmartins about the implications of complete removals?

I don't see any documentation changes in the PR but we do have some references to PicketLink in the documentation - this appears most to affect web services so may need to be discussed with @jimma.

For any removals I would like us to understand which tests are removed for two purposes:

1. Identification of lost functionality, especially functionality where we don't have a replacement strategy.
2. Tests targetting the removed functionality that can be repurposed to test new functionality or indicate the need for a new test.

I think quickstarts look clear.

On the WildFly side I think we need a clear statement somewhere for each removed component pointing to the new alternatives that should be used.

bstansberry:
@darranl OK, I'm going to close this then. I don't know why I didn't think about mixed-domain. I have it in my head that this was TP and has been on the chopping block forever and we can finally just get rid of it, but I keep forgetting it wasn't TP it was just long ago deprecated.

I don't know if any migration is feasible unless the OpenID Connect subsystem will be the target.  We don't ship any keycloak adapters and having a migrate op try and guess if the extension is present seems dicey.

There were some docs removed -- some WS content that details the PL integration.

PR #14390, opened 2021-06-22
WFLY-14843 Make EJB3 dependency on PicketBox Optional

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-14843


chengfang (approved):

bstansberry:
@chengfang @parsharma Will LocalEjbReceiver work properly without Picketbox? It seems like it would be used regardless of how the server is configured and will try and use PB classes.

darranl:
I am going to close this as we have the following PR to remove it entirely https://github.com/wildfly/wildfly/pull/14962

PR #14391, opened 2021-06-22
[WFLY-14913] ignore error on container kill command as process is for…

ochaloup:
as process is forcibly_ killed at the end and it's what we need to happen in test

https://issues.redhat.com/browse/WFLY-14913

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/transaction/preparehalt/TransactionPropagationFailureTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/transaction/preparehalt/TransactionPropagationFailureTestCase.java
@@ -177,9 +177,15 @@ public class TransactionPropagationFailureTestCase {
             log.debugf(expected,"Exception expected as the client server '%s' should be crashed", CLIENT_SERVER_NAME);
         }
 
-        // container was JVM crashed by the test, this call let the arquillian know that the container is not running
-        container.kill(CLIENT_SERVER_NAME);
-        // starting the container (as it was JVM crashed) for running transaction recovery
+        try {
+            // container was JVM crashed by the test, this call let the arquillian know that the container is not running
+            container.kill(CLIENT_SERVER_NAME);
+        } catch (Exception ignore) {
+            // ignoring the kill issue as the server process will be killed at the end (Arquillian uses Process.destroyForcibly)
+            // at this time the Arquillian container inner state could be 'KILLED_FAILED' which is still fine to restart the app server
+            log.debugf("Arquillian kill of " + CLIENT_SERVER_NAME + " failed, but the process is killed and we can ignore it", ignore);
+        }
+        // starting the container (JVM crashed by call Runtime.getRuntime().halt), test will be running transaction recovery on restarted server
         container.start(CLIENT_SERVER_NAME);
 
         RecoveryExecutor recoveryExecutor = new RecoveryExecutor(managementClientClient);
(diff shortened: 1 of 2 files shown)


rhusar (changes requested):
Still  appears to fail on Windows (Windows - JDK 11).

Nevertheless, isn't this attempting to be a  workaround for an issue with Arquillian?

ochaloup:
@rhusar ah, my wrong I haven't noticed the trouble. The failure seems to not be with Arquillian on Windows. It's some connection issue of Narayana recovery manager. I will investigate.

rhusar:
@ochaloup Anyway, I really don't think this test is doing what you want it to do.  There is just no kill, the implementation this defaults to regular stop as the log indicates, here is a code snipped from that impl class:

```java
public class DefaultServerKillProcessor implements ServerKillProcessor {
    private final Logger log = Logger.getLogger(DefaultServerKillProcessor.class.getName());

    @Override
    public void kill(Container container) throws Exception {
        log.warning("Triggering default implementation of a server kill() command => using stop() internally");
        container.getDeployableContainer().stop();
    }
}
```

Or am I misreading your intentions with the test?

ochaloup:
@rhusar yes, you are mislead, maybe by test, my code or my comments :)

The point is that the test makes the JVM killed. It's the call of `Runtime.halt()` (https://github.com/wildfly/wildfly/blob/master/testsuite/shared/src/main/java/org/jboss/as/test/integration/transactions/TestXAResource.java#L91) at specific time of the transaction processing. The `halt` **kills** the JVM - i.e., the word `kill` comes from that fact.

When this happens the test needs to start the server once again. The manager of the app server lifetime is Arquillian but as the kill came from a different place Arquillian does not know that's stopped.
I need to provide Arquillian with the information that the server is stopped and can be started again. Thus the `container.kill` call (I'm right now not sure if `container.stop` can be used but maybe it can be).

I can create my own Arquillian extension and override the `kill` processing (in way of "doing nothing" which makes the call to pass and the internal Arquillian state will be changed to killed). But I did not want to do so as some other tests could be calling `kill` in future and I would influence them (i.e., by `kill` doing nothing which is not intuitive).

So I want to let Arquillian know that the server is not running and thus it can be started again. I'm using the `kill/stop` that passes of doing nothing as container is already death.
If you have idea of better solution or perspective of what is the best practice here, please let me know.

ochaloup:
I have get running the WildFly testsuite here on CI three times and all win+jdk11 processed these times fine. I run the test on win+jdk11 on my machine separately and I haven't got the failures either.

The failure that was seen here was

```
java.lang.IllegalStateException: Error when triggering transaction recovery synchronous scan with RecoveryDriver null, based on the management client org.jboss.as.arquillian.container.ManagementClient@2ce03e86
  at org.jboss.as.test.integration.transactions.RecoveryExecutor.runTransactionRecovery(RecoveryExecutor.java:104)
  at org.jboss.as.test.integration.transactions.RecoveryExecutor.runTransactionRecovery(RecoveryExecutor.java:76)
  at org.jboss.as.test.manualmode.ejb.client.outbound.connection.transaction.preparehalt.TransactionPropagationFailureTestCase.prepareCrashOnClient(TransactionPropagationFailureTestCase.java:198)
```

which is not connected WFLY-14913 as the failure happens after Arquillian started the killed server succesfully. In this PR added better logging to be easier to understand the potential Win+jdk11 issue if happens again in some other test run.

I think this PR could be merged as considered addressing the WFLY-14913.

ochaloup:
@darranl @kabir do you think this could be merged?

rhusar (approved):

rhusar:
@ochaloup Ok, I see what you are doing, makes sense. However, this still seems to workaround the issue that our container controller doesn't implement `org.jboss.arquillian.container.spi.ServerKillProcessor#kill`. If then `org.jboss.arquillian.container.spi.Container#kill` actually worked, would you be able to leverage that in the test or is your use of `halt` very timing sensitive? I am planning to get this working so we can use it in our clustering tests eventually, so I am just wondering if there are more use case this would cover.

ochaloup:
@rhusar Having the `ServerKillProcessor#kill` working in Arquilian would be nicel I would use it but in other tests, not here. The `halt()` call needs to be executed in specific  time of transaction processing from inside of the app server. For that the Arquillian `kill` does not help me.

What I would need in this test is having a chance to inform Arquillian that the container is down when it was not launched by Arquillian. But it would be enhancement of some other interface while I'm not sure if it's generic enough and if it's something that somebody else than me needed.

rhusar:
@ochaloup Great. I will let you know if this moves anywhere as this is on the back burner (or you have a volunteer to implement the kill support let me know).

For this `halt` use-case, I am thinking a custom `ServerKillProcessor` that is a no-op should do the trick (and perhaps that's what should be happening in this test case). Would probably be useful when halt is called from Byteman as well.

PR #14401, opened 2021-06-28
[WFLY-14854] Add OpenTelemetry support

jasondlee:
https://issues.redhat.com/browse/WFLY-14854

This PR is not quite ready to merge. I opened it to get the CI builds. There are a number of clean ups that need to happen, and there are changes required (e.g., MP reactive messaging - https://github.com/wildfly/wildfly/pull/14475) that need to go in as well.


ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
```diff
@@ -0,0 +1,87 @@
...
+        <artifact name="${io.jaegertracing:jaeger-core}"/>
+<!--        <artifact name="${io.jaegertracing:jaeger-client}"/>-->
+        <artifact name="${io.opentelemetry:opentelemetry-exporter-jaeger}"/>
+        <artifact name="${io.grpc:grpc-okhttp}"/>
+        <artifact name="${io.grpc:grpc-core}"/>
+        <artifact name="${com.google.code.gson:gson}"/>
+        <artifact name="${com.google.android:annotations}"/>
+        <artifact name="${com.squareup.okhttp:okhttp}"/>
+        <artifact name="${com.squareup.okio:okio}"/>
+        <artifact name="${com.google.guava:guava}"/>
```
boris-unckel, line 40: This should be a module dependency: [/ee-feature-pack/common/src/main/resources/modules/system/layers/base/com/google/guava/main/module.xml](https://github.com/wildfly/wildfly/blob/master/ee-feature-pack/common/src/main/resources/modules/system/layers/base/com/google/guava/main/module.xml)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
```diff
@@ -0,0 +1,87 @@
...
+        <artifact name="${com.google.guava:failureaccess}"/>
```
boris-unckel, line 41: failureaccess is exported by the com.google.guava module and should be neither a artifact nore a dependency.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
```diff
@@ -0,0 +1,87 @@
...
+        <artifact name="${org.checkerframework:checker-compat-qual}"/>
+        <artifact name="${com.google.j2objc:j2objc-annotations}"/>
+        <artifact name="${com.google.errorprone:error_prone_annotations}"/>
+        <artifact name="${io.perfmark:perfmark-api}"/>
+        <artifact name="${org.codehaus.mojo:animal-sniffer-annotations}"/>
+        <artifact name="${com.google.code.findbugs:jsr305}"/>
```
boris-unckel, line 47: findbugs:jsr305 is a null/notnull annotation for compile time only. It should be removed here. Compare [https://github.com/wildfly/wildfly/blob/master/pom.xml#L1763-L1766](https://github.com/wildfly/wildfly/blob/master/pom.xml#L1763-L1766) and [https://github.com/wildfly/wildfly/blob/master/pom.xml#L1790-L1795](https://github.com/wildfly/wildfly/blob/master/pom.xml#L1790-L1795)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
```diff
@@ -0,0 +1,87 @@
...
+        <artifact name="${io.grpc:grpc-core}"/>
+        <artifact name="${com.google.code.gson:gson}"/>
+        <artifact name="${com.google.android:annotations}"/>
+        <artifact name="${com.squareup.okhttp:okhttp}"/>
+        <artifact name="${com.squareup.okio:okio}"/>
+        <artifact name="${com.google.guava:guava}"/>
+        <artifact name="${com.google.guava:failureaccess}"/>
+        <artifact name="${org.checkerframework:checker-compat-qual}"/>
+        <artifact name="${com.google.j2objc:j2objc-annotations}"/>
+        <artifact name="${com.google.errorprone:error_prone_annotations}"/>
```
boris-unckel, line 44: com.google.errorprone is is a static analysis tool for Java that catches common programming mistakes at compile-time. Compare with findbugs:jsr305 references.

pom.xml
```diff
@@ -2060,6 +2073,80 @@
...
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>30.0-android</version>
+            </dependency>
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>failureaccess</artifactId>
+                <version>1.0.1</version>
+            </dependency>
```
boris-unckel, line 2110: Duplicate [https://github.com/wildfly/wildfly/blob/master/pom.xml#L1758-L1795](https://github.com/wildfly/wildfly/blob/master/pom.xml#L1758-L1795) Is 30.0-**android** (vs 30.1-jre) intended?

pom.xml
```diff
@@ -2060,6 +2073,80 @@
...
+            <dependency>
+                <groupId>com.google.j2objc</groupId>
+                <artifactId>j2objc-annotations</artifactId>
+                <version>1.3</version>
+            </dependency>
+            <dependency>
+                <groupId>com.google.errorprone</groupId>
+                <artifactId>error_prone_annotations</artifactId>
+                <version>2.4.0</version>
+            </dependency>
```
boris-unckel, line 2125: Not needed at runtime. Is this really in use?

pom.xml
```diff
@@ -2060,6 +2073,80 @@
...
+            <dependency>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>animal-sniffer-annotations</artifactId>
+                <version>1.19</version>
+            </dependency>
+            <dependency>
+                <groupId>com.google.code.findbugs</groupId>
+                <artifactId>jsr305</artifactId>
+                <version>3.0.2</version>
+            </dependency>
```
boris-unckel, line 2140: Not needed at runtime. Is this really in use?

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
jasondlee: Correct. I dumped all of these in one module to make getting things working correctly a bit easier, so I do need to clean it up. TBH, I had lost track of that. :)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
jasondlee: Noted.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
jasondlee: Right, but it seems I had some weird CNFE errors around that, but much has changed in my modules of late, so I may have fixed the root cause. I'll remove this and see what happens.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
jasondlee: See above. :)

pom.xml
jasondlee: Definitely a dupe.

pom.xml
jasondlee: See above, but I'll double check all the deps.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/jaegertracing/main/module.xml
jasondlee: I've moved many of these into their own modules, though I have still bundled some artifacts together, as the added granularity didn't make sense to me, so there's still more to do there if you (or anyone else) deems it necessary.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
```diff
@@ -0,0 +1,51 @@
...
+        <artifact name="${io.opentelemetry:opentelemetry-exporter-jaeger}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-common}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-trace}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-proto}"/>
+    </resources>
+
+    <dependencies>
+        <module name="io.opentelemetry.api"/>
+        <module name="io.opentelemetry.context"/>
+        <module name="com.squareup.okhttp3"/>
```
boris-unckel, line 41: com.squareup.okhttp3 vs. com.squareup.okhttp2 (module above) ?

jasondlee: Good catch. Seems to work just fine with v3.

jasondlee: Fixed.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
```diff
@@ -0,0 +1,51 @@
...
+        <module name="org.slf4j"/>
+        <module name="io.jaegertracing"/>
+        <module name="io.grpc"/>
+        <module name="com.google.guava" />
+        <module name="com.google.code.gson"/>
+
+        <module name="java.logging"/>
+        <module name="sun.jdk"/>
```
boris-unckel, line 49: sun.jdk is deprecated. In [https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/modules/system/layers/base/sun/jdk/main/module.xml](https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/src/main/resources/modules/system/layers/base/sun/jdk/main/module.xml) you can find the modules to use instead.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry-api/main/module.xml
```diff
@@ -0,0 +1,48 @@
...
+        <module name="io.opentelemetry.context" />
+
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
+        <module name="org.jboss.weld.api"/>
+        <module name="org.wildfly.security.elytron"/>
+        <module name="javax.enterprise.api" />
+        <module name="java.logging"/>
+        <module name="javax.ws.rs.api" optional="true" services="import"/>
+        <module name="sun.jdk"/>
```
boris-unckel, line 46: deprecated, see above.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry-extension/main/module.xml
```diff
@@ -0,0 +1,62 @@
...
+        <module name="org.jboss.as.weld.common"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.metadata.common"/>
+        <module name="org.jboss.metadata.web"/>
+        <module name="org.jboss.modules"/>
+        <module name="org.jboss.staxmapper"/>
+
+        <module name="org.wildfly.security.elytron"/>
+
+        <module name="sun.jdk"/>
```
boris-unckel, line 60: deprecated, see above.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
jasondlee: Fixed.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry-api/main/module.xml
jasondlee: Not needed (copy and paste). Removed.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry-extension/main/module.xml
jasondlee: Same.

jasondlee:
I have rebased this change on @kabir's work, as there is some crossover. His PR (https://github.com/wildfly/wildfly/pull/14475) should probably be merged first, at which time I'll rebase this PR on master.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetrySubsystemDefinition.java
```diff
@@ -0,0 +1,160 @@
...
+ * For future reference: https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#jaeger-exporter
+ */
+
+public class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
+
+    private static final String[] ALLOWED_EXPORTERS = {"jaeger", "otlp"};
+    private static final String[] ALLOWED_SPAN_PROCESSORS = {"batch", "simple"};
+
+    public static final String[] MODULES = {
+            "org.wildfly.extension.opentelemetry-extension"
```
kabir, line 45: Although there are a few cases that don't follow this, we should try to put the MP spec extensions under `org.wildfly.extension.microprofile.` as the majority follow that pattern, so `org.wildfly.extension.microprofile.opentelemetry-extension` is better

jasondlee: OpenTelemetry, though, is not a MicroProfile spec, so I put it elsewhere. It *may* become an MP spec at some point, and we'll to figure out what means here if/when that happens.

kabir: Thanks for the clarification

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
```diff
@@ -0,0 +1,101 @@
...
+* `type`: The type of sampler to use
+** `on`: Always on (all traces are recorded)
+** `off`: Always off (no traces are recorded)
+** `ratio`: Return a ratio of the traces (e.g., 1 trace in 10000).
+* `sampler-arg`: The value use to configure the `ratio` sampler, which must be within [0.0, 1.0]. For example, if 1
+trace in 10,000 is to recorded, this value would be `0.0001`.
+
+=== Example Configuration
+
+The following XML is an example of the full configuration, including default values (WildFly does not persist default
```
bstansberry, line 49: WildFly does not include most attributes that have default values in its standard configuration files, so what you see in those files may look different.

(We do persist default values, if the user explicitly set the attribute to that value. Or we did if we created the standard config file.)

jasondlee: Right. I called that out in the docs to make sure users knew that and weren't confused. Would you like me to change something about this?

bstansberry (changes requested):
Hi @jasondlee -- here's some initial input. I haven't gotten to the java code part and need to step away so I'm starting with this.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
```diff
@@ -0,0 +1,101 @@
...
+received and the OpenTelemetry context propagation header (`traceparent`) is present, the request will traced as part of
+the remote trace context automatically.
+
+Likewise, all Jakarta REST Client calls will have the trace context added to outgoing request headers so that requests
+to external applications can be traced correctly (assuming the remote system properly handles OpenTelemetry trace context
+propagation). If the REST Client call is made to another application on the local WildFly server, or a remote server of
+the same version or later, the trace context will propagate automatically as described above.
+
+While automatic tracing may be sufficient in many cases, it will often be desirable to have traces occur throughout the
+user application. To support that, WildFly makes available the `OpenTelemetry` and `Tracer` instances, via CDI injection.
```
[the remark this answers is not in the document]
jasondlee: It is the object from which tracers are requested: https://opentelemetry.io/docs/java/manual_instrumentation/#set-up

Should someone need a tracer other than the one that's injected, he/she would use this object.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/deployment/OpenTelemetryExtensionLogger.java
```diff
@@ -0,0 +1,80 @@
...
+@MessageLogger(projectCode = "WFLYOTELEXT", length = 4)
+public interface OpenTelemetryExtensionLogger extends BasicLogger {
+    OpenTelemetryExtensionLogger OTEL_LOGGER = Logger.getMessageLogger(OpenTelemetryExtensionLogger.class,
+            OpenTelemetryExtensionLogger.class.getPackage().getName());
+
+    @LogMessage(level = INFO)
+    @Message(id = 1, value = "Activating OpenTelemetry Subsystem")
+    void activatingSubsystem();
+
+    @LogMessage(level = INFO) // DEBUG
```
bstansberry, line 42: DEBUG :)

Note that if it is below INFO it doesn't need to be in this class, and not having it here saves the need to translate it.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/deployment/OpenTelemetryExtensionLogger.java
```diff
@@ -0,0 +1,80 @@
...
+    @Message(id = 2, value = "OpenTelemetry Subsystem is processing deployment")
+    void processingDeployment();
+
+    @LogMessage(level = INFO) // DEBUG
```
bstansberry, line 46: DEBUG

ee-9/feature-pack/src/main/resources/feature_groups/domain-host-excludes.xml
```diff
@@ -5,70 +5,80 @@
...
+        <param name="excluded-extensions" value="[&quot;org.wildfly.extension.clustering.web&quot;,&quot;org.wildfly.extension.core-management&quot;,&quot;org.wildfly.extension.datasources-agroal&quot;,&quot;org.wildfly.extension.discovery&quot;,&quot;org.wildfly.extension.ee-security&quot;,&quot;org.wildfly.extension.elytron&quot;,&quot;org.wildfly.extension.health&quot;,&quot;org.wildfly.extension.metrics&quot;,&quot;org.wildfly.extension.microprofile.config-smallrye&quot;,&quot;org.wildfly.extension.microprofile.fault-tolerance-smallrye&quot;,&quot;org.wildfly.extension.microprofile.health-smallrye&quot;,&quot;org.wildfly.extension.microprofile.jwt-smallrye&quot;,&quot;org.wildfly.extension.microprofile.metrics-smallrye&quot;,&quot;org.wildfly.extension.microprofile.openapi-smallrye&quot;,&quot;org.wildfly.extension.microprofile.opentracing-smallrye&quot;,&quot;org.wildfly.extension.microprofile.reactive-messaging-smallrye&quot;,&quot;org.wildfly.extension.microprofile.reactive-streams-operators-smallrye&quot;,&quot;org.wildfly.extension.opentelemetry-extension&quot;]"/>
```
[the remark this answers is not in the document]
jasondlee: That was done to differentiate between this module and the user-exposed -api module. It _is_ verbose and redundant, so that can probably left off.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/opentelemetry-extension/layer-spec.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
jasondlee: TBH, I'm not sure. Galleon newbie move? Having done more with it since then, maybe we don't need them at all. I'll try removing them and seeing.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetryDeploymentDefinition.java
```diff
@@ -0,0 +1,42 @@
...
+    private OpenTelemetryDeploymentDefinition() {
+          super(new Parameters(OpenTelemetrySubsystemExtension.SUBSYSTEM_PATH,
+                  OpenTelemetrySubsystemExtension.getResourceDescriptionResolver())
+                  .setFeature(false));
+    }
+
+    @Override
+    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+    }
+}
```
bstansberry, line 42: This class should be removed, as it doesn't do anything. You don't need to add a resource in the /deployment=* tree unless it is providing something to an administrator, i.e. runtime state info or custom ops against a deployment.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetrySubsystemDefinition.java
```diff
@@ -0,0 +1,159 @@
...
+    };
+
+    public static final OpenTelemetrySubsystemDefinition INSTANCE = new OpenTelemetrySubsystemDefinition();
+
+    protected OpenTelemetrySubsystemDefinition() {
+
+        super(new SimpleResourceDefinition.Parameters(OpenTelemetrySubsystemExtension.SUBSYSTEM_PATH,
+                OpenTelemetrySubsystemExtension.getResourceDescriptionResolver())
+                .setAddHandler(OpenTelemetrySubsystemAdd.INSTANCE)
+                .setRemoveHandler(OpenTelemetrySubsystemRemove.INSTANCE));
```
bstansberry, line 151: You need to register a capability and have it depend on the "org.wildfly.weld" capability. That's how we ensure this subsystem can't be added if the weld subsystem is not and is what tells Galleon to reject an attempt to generate a config that does that.

See org.wildfly.extension.microprofile.config.smallrye.MicroProfileSubsystemDefinition.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetrySubsystemExtension.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public void initialize(ExtensionContext context) {
+        final SubsystemRegistration subsystem = context.registerSubsystem(SUBSYSTEM_NAME, CURRENT_MODEL_VERSION);
+        subsystem.registerXMLElementWriter(CURRENT_PARSER);
+
+        final ManagementResourceRegistration registration = subsystem.registerSubsystemModel(OpenTelemetrySubsystemDefinition.INSTANCE);
+        registration.registerOperationHandler(GenericSubsystemDescribeHandler.DEFINITION, GenericSubsystemDescribeHandler.INSTANCE);
+        subsystem.registerDeploymentModel(OpenTelemetryDeploymentDefinition.INSTANCE);
```
bstansberry, line 75: This can be dropped until there's a resource that does something.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetrySubsystemDefinition.java
bstansberry, line 151: Re the remove handler, use ReloadRequiredRemoveStepHandler.

observability/opentelemetry-extension/src/main/java/org/wildfly/extension/opentelemetry/extension/OpenTelemetrySubsystemRemove.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.dmr.ModelNode;
+
+/**
+ * Handler responsible for removing the subsystem resource from the model
+ *
+ * @author <a href="jasondlee@redhat.com">Jason Lee</a>
+ */
+public class OpenTelemetrySubsystemRemove extends AbstractRemoveStepHandler {
```
bstansberry, line 32: This class can be dropped; use ReloadRequiredRemoveStepHandler.

A resource that adds a DUP in its 'add' op can only have it's effect on the runtime removed via a reload or restart, which ReloadRequiredRemoveStepHandler handles.

bstansberry (changes requested):
Review continues; added some more comments.

A general comment re the java code is just a general point that reducing visibility of classes and members to the minimum is good, and declaring classes or methods final is good. Eliminates future compatibility discussions if we want to change things. Private modules give us freedom but you never know who will break rules.

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/opentelemetry-extension/layer-spec.xml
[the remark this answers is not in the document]
jasondlee: Looks like they're artifacts of "let's do what this other module is doing". I've removed them and things appear to be sane.

jasondlee:
> Review continues; added some more comments.
> 
> A general comment re the java code is just a general point that reducing visibility of classes and members to the minimum is good, and declaring classes or methods final is good. Eliminates future compatibility discussions if we want to change things. Private modules give us freedom but you never know who will break rules.

Agreed, and I usually try to do that (though I was likely less stringent with THIS code, as I was more concerned on making it work). Do you have any examples in mind, or should I just revisit everything and lock it down as much as possible?

bstansberry:
Re locking down, it was mostly general.

observability/opentelemetry-extension/src/main/resources/subsystem-templates/opentelemetry.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
bstansberry, line 1: This file can be dropped now. We no longer produce pre-Galleon legacy feature packs, which is what used these,

bstansberry (changes requested):

ee-9/feature-pack/src/main/resources/feature_groups/domain-host-excludes.xml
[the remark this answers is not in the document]
bstansberry: The opentelemetry-api module is jboss-api=private, so its name is not particularly relevant to users. The extension module name is because its part of the user managed config.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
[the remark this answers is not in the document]
bstansberry: Perhaps use the FQCN? A reader may not be familiar with the OT API.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
bstansberry: Sorry; the first sentence in my last post was my suggestion. :)

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/opentelemetry-extension/layer-spec.xml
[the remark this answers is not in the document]
bstansberry: There should be one, called 'opentelemetry'. I think this opentelemetry-extension one is the right content.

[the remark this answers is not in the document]
bstansberry: An execution for this needs to be added to testsuite/layers/pom.xml, and then the layer needs to be added to the all-layers-provisioning-full and all-layers-jpa-distributed-provisioning-full executions.

observability/opentelemetry-extension/src/main/resources/subsystem-templates/opentelemetry.xml
jasondlee: Removed.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
jasondlee: Added some clarification? :)

ee-feature-pack/galleon-common/src/main/resources/layers/standalone/opentelemetry-extension/layer-spec.xml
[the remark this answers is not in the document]
jasondlee: Ah. I missed the comment before this one. I removed those layers. Should I add them back? I'm not sure what you mean by execution. Example?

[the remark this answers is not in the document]
bstansberry: Please do add back 1 layer, 'opentelemetry'. We want users to be able to use layers to provision all functionality, so we need one.

By execution I mean meant maven plugin executions, specifically a galleon provisioning one added here:

https://github.com/wildfly/wildfly/blob/main/testsuite/layers/pom.xml#L1727

and then an element added here: 

https://github.com/wildfly/wildfly/blob/main/testsuite/layers/pom.xml#L2297

and here:

https://github.com/wildfly/wildfly/blob/main/testsuite/layers/pom.xml#L2398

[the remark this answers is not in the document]
bstansberry: A question here, where I'm interested in @jmesnil's input, is whether we want this as an optional dependency of the 'observability' layer.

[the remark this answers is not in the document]
jasondlee: Thanks. Will work on that now. Also interested in Jeff's input. Seems like it should be part of it, especially given our plans...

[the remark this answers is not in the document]
jasondlee: Done and pushed.

bstansberry (approved):
I'm approving this with respect to the last 3 commits (i.e. the OT ones, not the RM ones they are built on.) The RM ones approved in Kabir's PR.

jasondlee:
> I'm approving this with respect to the last 3 commits (i.e. the OT ones, not the RM ones they are built on.) The RM ones approved in Kabir's PR.

Thanks. Once @kabir's PR is merged, I'll rebase off of main to clean this up before merging.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/deployment/OpenTelemetryDependencyProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+
+    }
+
+    private void addDependencies(DeploymentUnit deploymentUnit) {
+        ModuleSpecification moduleSpecification = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
+        ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        for (String module : MODULES) {
+            moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, module, false, false,
+                    true, false));
+        }
```
pferraro, line 51: Why does the OpenTelemetry subsystem module need to be exposed to deployments?
Unless I'm mistaken, it seems to me that this is only needed to allow Weld to locate the CDI extension.
Would it not be better to register this extension manually (i.e. via WeldCapability#registerExtensionInstance(...))?
This way we avoid bloating the deployment classloader and do not expose unnecessary subsystem internals to deployments.

pferraro (changes requested):

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryParser_1_0.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.wildfly.extension.opentelemetry;
+
+import static org.jboss.as.controller.PersistentResourceXMLDescription.builder;
+
+import org.jboss.as.controller.PersistentResourceXMLDescription;
+import org.jboss.as.controller.PersistentResourceXMLParser;
+
+public class OpenTelemetryParser_1_0 extends PersistentResourceXMLParser {
```
pferraro, line 27: I suggest parameterize your parser with a schema version, rather than anticipate a parser class per version.  You'll save yourself a lot of code duplication in the long run.
e.g. https://github.com/wildfly/wildfly/blob/main/clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/DistributableWebXMLParser.java

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
```diff
@@ -0,0 +1,170 @@
...
+            .setAttributeGroup(GROUP_SAMPLER)
+            .setRestartAllServices()
+            .build();
+
+    public static final AttributeDefinition[] ATTRIBUTES = {
+            SERVICE_NAME, EXPORTER, ENDPOINT, SPAN_PROCESSOR_TYPE, BATCH_DELAY, MAX_QUEUE_SIZE, MAX_EXPORT_BATCH_SIZE,
+            EXPORT_TIMEOUT, SAMPLER, SAMPLER_ARG
+    };
+
+    public static final OpenTelemetrySubsystemDefinition INSTANCE = new OpenTelemetrySubsystemDefinition();
```
pferraro, line 155: ResourceDefinition instances are only needed during resource registration.  Retaining references to them via static references prevents their garbage collection.

pferraro (changes requested):

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/deployment/OpenTelemetryDependencyProcessor.java
jasondlee: To be honest, I was just following the pattern I had seen elsewhere. :) I'll take a look at the linked code and go from there.

pferraro: An alternative approach, leveraged in subsystems like messaging-activemq or smallrye-fault-tolerance: encapsulate your portable extension within a separate module, and add a deployment dependency on that module.

pferraro: Ideally, the only modules on which a deployment should depend are any public modules containing API (spec or otherwise) that a deployment intends to use.
If you register your portable extension via the WeldCapability directly, there should be no need to export that module to the deployment (since Weld no longer needs to resort to service loading to find your CDI extension).

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryParser_1_0.java
jasondlee: Done.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
jasondlee: Done.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/deployment/OpenTelemetryDependencyProcessor.java
jasondlee: Fixed.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
```diff
@@ -0,0 +1,66 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
```
OndrejKotek, line 25: The schema should provide documentation of the elements and attributes, like other schemas, e.g. Elytron subsystem.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
```diff
@@ -0,0 +1,66 @@
...
+                                    <xs:enumeration value="otlp"/>
+                                </xs:restriction>
+                            </xs:simpleType>
+                        </xs:attribute>
+                        <xs:attribute name="endpoint" type="xs:string"/>
+                    </xs:complexType>
+                </xs:element>
+                <xs:element name="span-processor" minOccurs="0" maxOccurs="1">
+                    <xs:complexType>
+                        <xs:attribute type="xs:string" name="type" default="batch"/>
```
OndrejKotek, line 49: Why isn't this restricted? Can there be custom values?

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
```diff
@@ -0,0 +1,66 @@
...
+                        <xs:attribute type="xs:integer" name="batch-delay" default="5000"/>
+                        <xs:attribute type="xs:integer" name="max-queue-size" default="2048"/>
+                        <xs:attribute type="xs:integer" name="max-export-batch-size" default="512"/>
+                        <xs:attribute type="xs:integer" name="export-timeout" default="30000"/>
+                    </xs:complexType>
+                </xs:element>
+                <xs:element name="sampler" minOccurs="0" maxOccurs="1">
+                    <xs:complexType>
+                                <xs:attribute type="xs:string" name="type"/>
```
OndrejKotek, line 58: Why isn't this restricted? Can there be custom values?

What is the default value?

observability/opentelemetry/src/main/resources/org/wildfly/extension/opentelemetry/LocalDescriptions.properties
```diff
@@ -0,0 +1,13 @@
...
+opentelemetry.batch-delay=The interval, in milliseconds, between two consecutive exports. Default is 5000.
+opentelemetry.endpoint=The URL for the exporter endpoint.
+opentelemetry.export-timeout=The maximum allowed time, in milliseconds, to export data.
+opentelemetry.exporter-type=The name of the exporter to use
+opentelemetry.max-export-batch-size=The maximum size of the export batch.
+opentelemetry.max-queue-size=The maximum size of the queue before it is exported.
+opentelemetry.remove=Operation Removes subsystem
+opentelemetry.sampler-arg=The argument to use to configure the sampler
+opentelemetry.sampler-type=The sampling strategy to use
+opentelemetry.service-name=Service name
```
OndrejKotek, line 12: This is not much helpful. How is this used?

observability/opentelemetry/src/main/resources/org/wildfly/extension/opentelemetry/LocalDescriptions.properties
```diff
@@ -0,0 +1,13 @@
+opentelemetry=OpenTelemetry subsystem
+opentelemetry.add=Operation Adds subsystem
+opentelemetry.batch-delay=The interval, in milliseconds, between two consecutive exports. Default is 5000.
+opentelemetry.endpoint=The URL for the exporter endpoint.
+opentelemetry.export-timeout=The maximum allowed time, in milliseconds, to export data.
+opentelemetry.exporter-type=The name of the exporter to use
+opentelemetry.max-export-batch-size=The maximum size of the export batch.
+opentelemetry.max-queue-size=The maximum size of the queue before it is exported.
```
OndrejKotek, line 8: What is the relationship between the queue and batch? Is a batch one unit in the queue?

boris-unckel:
@jasondlee A remark more like a question: New pom.xml and module.xml files have a GNU Lesser V2 license, new .java files have a Apache V2 license. I could bet for poms and modules this is C'n'P, for *.java it's the Eclipse preference :-)

What is the correct one? Or is the mix for some reason intended?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/MultiserviceTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.test.integration.observability.opentelemetry.application.Service1;
+import org.wildfly.test.integration.observability.opentelemetry.application.Service2;
+import org.wildfly.test.integration.observability.opentelemetry.application.TestApplication;
+
```
OndrejKotek, line 48: A simple description explaining the tested scenario would be helpful for understanding what is covered and potential investigation.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/MultiserviceTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+            + "<web-app xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns=\"http://java.sun.com/xml/ns/javaee\"\n"
+            + "         xsi:schemaLocation=\"http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd\"\n"
+            + "         metadata-complete=\"false\" version=\"3.0\">\n"
+            + "    <servlet-mapping>\n"
+            + "        <servlet-name>javax.ws.rs.core.Application</servlet-name>\n"
+            + "        <url-pattern>/*</url-pattern>\n"
+            + "    </servlet-mapping>"
+            + "</web-app>";
+
+    public static final String WEB_XML2 =
```
OndrejKotek, line 66: Is the value of this different from the value of `WEB_XML` above intentionally? If it's the case, why? If not, one constant will be enough.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
```diff
@@ -0,0 +1,168 @@
...
+    public static final SimpleAttributeDefinition EXPORT_TIMEOUT = SimpleAttributeDefinitionBuilder
+            .create(OpenTelemetryConfigurationConstants.EXPORT_TIMEOUT, ModelType.LONG, true)
+            .setAllowExpression(true)
+            .setAttributeGroup(GROUP_SPAN_PROCESSOR)
+            .setRestartAllServices()
+            .setDefaultValue(new ModelNode(30000))
+            .build();
+
+    public static final SimpleAttributeDefinition SAMPLER = SimpleAttributeDefinitionBuilder
+            .create(OpenTelemetryConfigurationConstants.SAMPLER_TYPE, ModelType.STRING, true)
```
OndrejKotek, line 136: Can we limit this with `setAllowedValues`?

observability/opentelemetry/src/main/resources/org/wildfly/extension/opentelemetry/LocalDescriptions.properties
jasondlee: I believe that language is an almost direct copy from the otel docs, but I've tried to make it clearer here. Will push changes momentarily.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
```diff
@@ -0,0 +1,168 @@
...
+            .setAllowExpression(true)
+            .setXmlName(OpenTelemetryConfigurationConstants.TYPE)
+            .setAttributeGroup(GROUP_SAMPLER)
+            .setRestartAllServices()
+            .build();
+
+    public static final SimpleAttributeDefinition SAMPLER_ARG = SimpleAttributeDefinitionBuilder
```
OndrejKotek, line 143: Isn't this too general? Can we rather call this ratio, set limits for the value, and set a default value?

observability/opentelemetry/src/test/java/org/wildfly/extension/opentelemetry/extension/SubsystemParsingTestCase.java
```diff
@@ -0,0 +1,149 @@
...
+import org.jboss.as.subsystem.test.AdditionalInitialization;
+import org.jboss.as.subsystem.test.KernelServices;
+import org.jboss.dmr.ModelNode;
+import org.junit.Assert;
+import org.junit.Test;
+import org.wildfly.extension.opentelemetry.OpenTelemetrySchema;
+import org.wildfly.extension.opentelemetry.OpenTelemetrySubsystemExtension;
+
+
+/**
```
OndrejKotek, line 28: This seems to be a default description.

observability/opentelemetry/src/test/java/org/wildfly/extension/opentelemetry/extension/SubsystemParsingTestCase.java
```diff
@@ -0,0 +1,149 @@
...
+ * Tests all management expects for subsystem, parsing, marshaling, model definition and other
+ * Here is an example that allows you a fine grained controler over what is tested and how. So it can give you ideas what can be done and tested.
+ * If you have no need for advanced testing of subsystem you look at {@link SubsystemBaseParsingTestCase} that testes same stuff but most of the code
+ * is hidden inside of test harness
+ *
+ * @author <a href="jasondlee@redhat.com">Jason Lee</a>
+ */
+public class SubsystemParsingTestCase extends AbstractSubsystemTest {
```
OndrejKotek, line 36: All the attributes of the new subsystem that allow expressions should be tested with expressions set. I suppose this is the right place.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
```diff
@@ -0,0 +1,66 @@
...
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="urn:wildfly:opentelemetry:1.0"
+           xmlns="urn:wildfly:opentelemetry:1.0"
+           elementFormDefault="qualified"
+           version="1.0">
+
+    <xs:element name="subsystem">
+        <xs:complexType>
```
OndrejKotek, line 32: Is it ok that configurations for the upcoming parts (like logging) will be mixed up with this one for tracing?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BasicOpenTelemetryTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import io.opentelemetry.api.trace.Tracer;
+
+@RunWith(Arquillian.class)
+public class BasicOpenTelemetryTestCase {
```
OndrejKotek, line 37: This test case covers the DUP. I would expect here also a real integration with a collector.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/MultiserviceTestCase.java
jasondlee: Good catch. That test changed a lot and this was orphaned.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
jasondlee: Perhaps, and I tend to agree with you, but this is what the otel spec calls it, so I left it as is in case it's ever used in the future for something else as well. I guess in that case we can just add another param on our side to cover it.

OndrejKotek: Thanks for the explanation. What about a dynamic validator for the value then? In case the sampler is `ratio` just 0.0 - 1.0 will be allowed?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/MultiserviceTestCase.java
jasondlee: Added.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
jasondlee: Changed.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
jasondlee: Restrictions added.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
jasondlee: Restrictions added.

observability/opentelemetry/src/main/resources/org/wildfly/extension/opentelemetry/LocalDescriptions.properties
jasondlee: Updated.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
jasondlee: An excellent question. Perhaps I should nest all of this under a <tracing> element? @bstansberry, thoughts?

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BasicOpenTelemetryTestCase.java
jasondlee: That would be ideal, but, as written, the extension doesn't support the kind of extensibility it would take to do that. The OpenTelemetry instance is created once and shared across all deployments. The pieces that would need to be swapped out in order to capture/process/verify the spans in a test are not changeable once the otel instance has been created. In order to support that, we would need to modify when/where the otel instance is created, as well as providing a mechanism for the application to provide (e.g., via CDI) alternative implementations for the various pieces that wired together inside the otel instance. All of that is doable, of course, but I chose NOT to support that in this first go in order to keep things simple, waiting to add that kind of support if/when users request it. We can do that NOW, though that might impact the schedule (hard to say). All that said, I _have_ tested manually with a Jaeger instance running locally and have seen the traces exported as expected. That is not as good as automated regression testing, of course, but it's at least a bit better than nothing. :) If you (or anyone) feels strongly that we should implement the test like this, please speak up. I may try to get started regardless and see what happens.

observability/opentelemetry/src/test/java/org/wildfly/extension/opentelemetry/extension/SubsystemParsingTestCase.java
jasondlee: Indeed. Copy-and-paste. Removed.

observability/opentelemetry/src/test/java/org/wildfly/extension/opentelemetry/extension/SubsystemParsingTestCase.java
jasondlee: Added

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
bstansberry: We can modify the xsd later if appropriate. There is no requirement to maintain an xml structure across releases, only to be able to parse old versions of the xsd. We make no guarantees that people can do xml manipulation compatibly across releases.

A more important question though is whether the subsystem resource definition structure is properly accounting for likely future expansion. THAT we do provide compatibility guarantees over time. At a glance it seems ok, as the configuration is a number of attributes that are scoped to the 'tracing' use case. New things could be added via child resources or via additional attributes whose names are understandable without clashing with the existing ones.

jasondlee: WFM. Thanks.

darranl:
@jasondlee FYI this one now also have a merge conflict I think following on from the legacy realm removal before the weekend.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
jasondlee: Added

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BasicOpenTelemetryTestCase.java
jasondlee: I discussed this with @bstansberry. We won't be expanding the RFE so that we can run a test case. If/when we do add this functionality, this kind of test can and will be added.

jasondlee: I spoke too quickly. There might be a way to test this pretty simply with the Jaeger collector. I'll get back to you.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
```diff
@@ -0,0 +1,102 @@
...
+trace in 10,000 is to recorded, this value would be `0.0001`.
+
+=== Example Configuration
+
+The following XML is an example of the full configuration, including default values (WildFly does not typically persist
+default values, so what you see in the configuration file may look different):
+
+[code,xml]
+-----
+<subsystem xmlns="urn:wildfly:opentelemetry:1.0" service-name="example">
```
spriadka, line 54: What about the `service-name` attribute? Is it configurable? Could you add it to the rest of the configurability information?

jasondlee: This *should* work now, though I still need to verify locally.

jasondlee:
> @jasondlee FYI this one now also have a merge conflict I think following on from the legacy realm removal before the weekend.

I have rebased from main, so conflicts should be resolved and the Pr should contain only otel changes.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Observability_Tracing.adoc
jasondlee: Confirmed.

testsuite/integration/manualmode/pom.xml
```diff
@@ -164,6 +164,27 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-context</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.16.0</version>
```
darranl, line 184: I think this version should still be managed from the wildfly parent pom.

jasondlee: Good catch. Put it there while experimenting and forgot to clean it up. Moved locally. Will push when the current CI jobs finish to make sure there isn't something else I need to clean up.

jasondlee:
> > @jasondlee FYI this one now also have a merge conflict I think following on from the legacy realm removal before the weekend.
> 
> I have rebased from main, so conflicts should be resolved and the Pr should contain only otel changes.

@darranl Changes pushed, so this should be ready, unless there's a spurious CI failure. It was all green when I pushed, fwiw. :)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry-api/main/module.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.wildfly.extension.opentelemetry-api">
```
mchoma, line 25: I am not expert in wildfly building. But does it mean module `org.wildfly.extension.opentelemetry-api` will be included in EAP Base, when this module is located in `wildfly-ee-feature-pack-common`?

mchoma: @jasondlee My understanding is this feature should end up in Microprofile part of server - EAP XP.  Can you confirm it will not end up in EAP Base part?

yersan: Thanks for raising it again @mchoma , I was preparing the conversions and faced the same. We could deal with them at conversion level, but this should be treated upstream instead, keeping all modules, licenses, and resources used by this extension at the galleon-pack level instead of the ee- variant, ideally handled as [wildfly-elytron-oidc-client](https://github.com/wildfly/wildfly/blob/795c1fc286afd32c1a474f6aeb430e5889eb3869/elytron-oidc-client/galleon-common/pom.xml#L40) does. Let's wait for Jason's feedback.

mchoma: I can see modules in ee-dist, what I believe is analogy of EAP Base
```
pwd
/wildfly/ee-dist/target/wildfly-25.0.0.Final-SNAPSHOT/modules/system/layers/base/org/wildfly/extension
[mchoma@localhost extension (master)]$ tree -L 1
.
...
├── opentelemetry
├── opentelemetry-api
...
```

jasondlee: It is my understanding that this will ship with XP4 (that port is underway at the moment) but that it will also ship with EAP 8 (though I'm not a PM so don't quote me on that :).

PR #14408, opened 2021-06-30
[WFLY-14957] Skip the Jakarta transformer if a system property is set.

jamezp:
https://issues.redhat.com/browse/WFLY-14957

This will be useful for projects which use WildFly for testing to ensure they didn't miss anything that has not been transformed.

### DIFF
--- a/ee-9/deployment-transformer/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java
+++ b/ee-9/deployment-transformer/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java
@@ -24,6 +24,7 @@ import java.nio.file.Path;
 
 import org.jboss.logging.Logger;
 import org.wildfly.galleon.plugin.transformer.JakartaTransformer;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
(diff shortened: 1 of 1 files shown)


ee-9/deployment-transformer/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java
```diff
@@ -57,4 +62,9 @@ public void print(String format, Object... args) {
         });
         */
     }
+
+    private boolean isEnabled() {
+        final String value = WildFlySecurityManager.getPropertyPrivileged("org.wildfly.skip.jakarta.transformer", "");
```
bstansberry, line 67: Please stick the word 'unsupported' in this property name.

How this stuff works can change at any time, and since the purpose of this AIUI is to ease use of WF Preview by outside projects, any projects that use this should understand that this isn't a reliable API.

bstansberry (changes requested):

ee-9/deployment-transformer/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java
jamezp: Done

bstansberry (approved):

PR #14411, opened 2021-07-01
[WFLY-14960] Upgrade commons-io from 2.5 to 2.10.0

gaol:

Issue: https://issues.redhat.com/browse/WFLY-14960

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -238,7 +238,7 @@
             versions, add the artifactId or other qualifier to the property name.
          -->
         <version.asciidoctor.plugin>1.5.6</version.asciidoctor.plugin>
-        <version.commons-io>2.5</version.commons-io>
+        <version.commons-io>2.10.0</version.commons-io>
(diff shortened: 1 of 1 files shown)


gaol:
@bstansberry @jimma @jamezp May I ask from you to evaluate this upgrade? thanks :)

jimma:
@gaol  This is good for webservice subsystem.

bstansberry:
@gaol I submitted https://github.com/wildfly/wildfly-core/pull/4650 to see if the core testsuite failure that sparked our hesitancy about moving on from 2.5 is still there. My impression though is that it is.

If it is we're going to have to decide what to do. The biggest issue before was that failure made us concerned about other failures, but if our actual production use of commons-io is looks ok we can deal with the core test thing in some other way.

jamezp (approved):
I'm sorry. I'm not sure how I missed this notification. This should be fine for RESTEasy.

PR #14414, opened 2021-07-01
[WFLY-14963] Remove redundant boolean literals

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-14963


bstansberry:
Personally, I am not in favor of this kind of thing.

A change like this means dozens of possible bugs, and these changes lines don't fit on the screen when doing code review so the odds of mistakes not being noticed are quite high.

boris-unckel:
@bstansberry Hello Brian, thanks for your feedback. I assume I cannot gain acceptance by splitting this and withdraw the PR.

PR #14415, opened 2021-07-01
WFLY-7115 KeyAffinityService blocks Infinispan's topology change thread

pferraro:
https://issues.redhat.com/browse/WFLY-7115

Replaces Infinispan's default KeyAffinityService with a custom implementation without the problem described in WFLY-7715.
Also encapsulates all key distribution logic into a single object to ensure agreement between disparate parties.

### DIFF
--- a/clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/affinity/DefaultKeyAffinityService.java
+++ b/clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/affinity/DefaultKeyAffinityService.java
@@ -214,7 +214,7 @@ public class DefaultKeyAffinityService<K> implements KeyAffinityService<K>, Supp
     }
 
     @TopologyChanged
-    public CompletionStage<Void> viewChanged(TopologyChangedEvent<?, ?> event) {
+    public CompletionStage<Void> topologyChanged(TopologyChangedEvent<?, ?> event) {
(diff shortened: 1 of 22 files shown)


pferraro:
@rhusar @rachmatowicz  Please review.

clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/affinity/DefaultKeyAffinityService.java
```diff
@@ -0,0 +1,301 @@
...
+                return keys.poll(nanos, TimeUnit.NANOSECONDS);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        return null;
+    }
+
+    @TopologyChanged
+    public CompletionStage<Void> viewChanged(TopologyChangedEvent<?, ?> event) {
```
rhusar, line 217: This method should rather be called `topologyChanged(...)`.

rhusar (approved):
I think this looks good!

clustering/infinispan/spi/src/main/java/org/wildfly/clustering/infinispan/spi/affinity/DefaultKeyAffinityService.java
pferraro: The name is arbitrary, but I agree with you.

rhusar (approved):
LGTM 👍

PR #14422, opened 2021-07-02
WFLY-14971 Singleton deployment tests still failing intermittently

pferraro:
https://issues.redhat.com/browse/WFLY-14971


pferraro:
Adding hold label until verified via sufficient CI runs.

pferraro:
So far, so good.  I will focus rerunning on Linux SM - JDK 8 jobs, as these test seem to fail most often with that configuration.

bstansberry:
@rhusar Please review.

pferraro:
This looks good now.  No singleton deployment test failures seen.

clustering/server/src/main/resources/org.wildfly.clustering.server.provider.proto
```diff
@@ -5,12 +5,6 @@ import "org.jgroups.util.proto";
 
 // IDs: 140 -144
 
-/**
- * @TypeId(140)
- */
-message GetLocalServicesCommand {
-}
```
rhusar, line 12: I am under impression that proto types should be commented out rather than removed so the IDs don't get inadvertently reused?

rhusar (approved):
Looks good!

clustering/server/src/main/resources/org.wildfly.clustering.server.provider.proto
pferraro: We are not yet guaranteeing serialization compatibility across major releases.

PR #14425, opened 2021-07-03
[WFLY-14968] Collapse if-statements (Security)

boris-unckel:

### DIFF
--- a/security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
+++ b/security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
@@ -323,37 +323,37 @@ public class SimpleSecurityManager implements ServerSecurityManager, Supplier<Se
         RunAs currentRunAs = current.getIncomingRunAs();
         boolean trusted = currentRunAs != null && currentRunAs instanceof RunAsIdentity;
 
-        if (trusted == false) {
+        if (!trusted && SecurityActions.remotingContextIsSet()) {
             /*
              * We should only be switching to a context based on an identity from the Remoting connection if we don't already
              * have a trusted identity - this allows for beans to reauthenticate as a different identity.
              */
-            if (SecurityActions.remotingContextIsSet()) {
-                // In this case the principal and credential will not have been set to set some random values.
-                SecurityContextUtil util = current.getUtil();
+            // In this case the principal and credential will not have been set to set some random values.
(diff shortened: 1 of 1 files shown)


security/subsystem/src/main/java/org/jboss/as/security/service/SimpleSecurityManager.java
```diff
@@ -323,37 +323,37 @@ public void push(final String securityDomain) {
         RunAs currentRunAs = current.getIncomingRunAs();
         boolean trusted = currentRunAs != null && currentRunAs instanceof RunAsIdentity;
 
-        if (trusted == false) {
+        if (trusted && SecurityActions.remotingContextIsSet()) {
```
darranl, line 326: I am just quickly scanning the diff so may have missed some lines but this seems to have lost the "trusted==false"?

boris-unckel: Thanks for the quick reply. I saw the failing test and it is resolved now.

darranl (approved):
FYI this code is planned to be deleted during WildFly 25 development anyway but should be Ok to merge.

TBH I am not a big fan of these kind of changes, I don't really agree about the readability motivation, separate lines make it a lot easier to read one step at a time and also a lot easier to navigate with a debugger.  I know when I am debugging code I am not overly familiar with the last thing I want to hit is a complex if statement with multiple operators including the very thin not operator.  Newer JVMs are much better at reporting the exact location is things like NPEs but older ones when trying to support users remotely if too much happens on a single line in the past it has been very difficult to deduce what may be null.  But this one does seem sufficiently constrained.

boris-unckel:
@darranl Thanks for your review and comments.
> FYI this code is planned to be deleted during WildFly 25 development anyway but should be Ok to merge.

The merge conflict is visible below and I have a local solution prepared, ready to push. I don't want to block other important tests, so I just push 1-2 at a time.

> 
> TBH I am not a big fan of these kind of changes, I don't really agree about the readability motivation, separate lines make it a lot easier to read one step at a time and also a lot easier to navigate with a debugger. ...

Yersay gave me a similar feedback and I'm already reworking all PRs: Separate lines for separate conditions, except OR expressions in parenthesis. Unnecessary parenthesis removed.

boris-unckel:
@bstansberry Hello Brian, the tests for this PR have passed, it is approved. Can you check and merge, please?

PR #14433, opened 2021-07-04
[WFLY-14968] Collapse if-statements (Weld)

boris-unckel:

### DIFF
--- a/weld/common/src/main/java/org/jboss/as/weld/util/Reflections.java
+++ b/weld/common/src/main/java/org/jboss/as/weld/util/Reflections.java
@@ -74,10 +74,9 @@ public class Reflections {
     public static boolean containsAnnotation(Class<?> javaClass, Class<? extends Annotation> requiredAnnotation) {
         for (Class<?> clazz = javaClass; clazz != null && clazz != Object.class; clazz = clazz.getSuperclass()) {
             // class level annotations
-            if (clazz == javaClass || requiredAnnotation.isAnnotationPresent(Inherited.class)) {
-                if (containsAnnotations(clazz.getAnnotations(), requiredAnnotation)) {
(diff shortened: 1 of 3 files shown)


weld/common/src/main/java/org/jboss/as/weld/util/Reflections.java
```diff
@@ -74,10 +74,9 @@ public static boolean isAccessible(String className, ClassLoader classLoader) {
     public static boolean containsAnnotation(Class<?> javaClass, Class<? extends Annotation> requiredAnnotation) {
         for (Class<?> clazz = javaClass; clazz != null && clazz != Object.class; clazz = clazz.getSuperclass()) {
             // class level annotations
-            if (clazz == javaClass || requiredAnnotation.isAnnotationPresent(Inherited.class)) {
-                if (containsAnnotations(clazz.getAnnotations(), requiredAnnotation)) {
+            if ((clazz == javaClass || requiredAnnotation.isAnnotationPresent(Inherited.class))
+                    && containsAnnotations(clazz.getAnnotations(), requiredAnnotation)) {
```
yersan, line 78: Notice here you are modifying the order of executions. On the left side, `containsAnnotations` is executed always if `clazz == javaClass || requiredAnnotation.isAnnotationPresent(Inherited.class)` is true, on the right side `containsAnnotations` is not executed in such case.

We should try to avoid these code changes, I would say it is harmless here, but if the side effect can be avoided, much better.

yersan: Ups, sorry, my fault, I did not see that the OR was enclosed with the parenthesis

yersan (approved):

boris-unckel:
@bstansberry This one has passed tests and is approved, too. Please, could you check and merge?

PR #14434, opened 2021-07-04
[WFLY-14968] Collapse if-statements (Misc)

boris-unckel:
Last one of the series.

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
@@ -45,11 +45,9 @@ public class JSFMetadataProcessor implements DeploymentUnitProcessor {
                 jsf = servlet;
             }
         }
-        if(jsf != null) {
-            if(jsf.getMultipartConfig() == null) {
-                //WFLY-2329 File upload doesn't work
-                jsf.setMultipartConfig(new MultipartConfigMetaData());
-            }
+        if (jsf != null && jsf.getMultipartConfig() == null) {
+            // WFLY-2329 File upload doesn't work
+            jsf.setMultipartConfig(new MultipartConfigMetaData());
         }
         if (disallowDoctypeDecl != null) {
             // Add the disallowDoctypeDecl context param if it's not already present
--- a/xts/src/main/java/org/jboss/as/xts/txnclient/WildflyTransactionClientTxBridgeIntegrationHandler.java
+++ b/xts/src/main/java/org/jboss/as/xts/txnclient/WildflyTransactionClientTxBridgeIntegrationHandler.java
@@ -73,12 +73,11 @@ public class WildflyTransactionClientTxBridgeIntegrationHandler implements Handl
         final Boolean isOutbound = (Boolean) messageContext.get(MessageContext.MESSAGE_OUTBOUND_PROPERTY);
 
         try {
-            if (isOutbound != null && isOutbound) {
-                // suspending context before returning to the client
-                if(ContextTransactionManager.getInstance() != null &&
-                        ContextTransactionManager.getInstance().getStatus() != Status.STATUS_NO_TRANSACTION) {
-                    ContextTransactionManager.getInstance().suspend();
-                }
+            if (isOutbound != null && isOutbound
+                    // suspending context before returning to the client
+                    && ContextTransactionManager.getInstance() != null
+                    && ContextTransactionManager.getInstance().getStatus() != Status.STATUS_NO_TRANSACTION) {
+                ContextTransactionManager.getInstance().suspend();
             }
         } catch (SystemException se) {
             XtsAsLogger.ROOT_LOGGER.cannotGetTransactionStatus(messageContext, se);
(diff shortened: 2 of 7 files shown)


xts/src/main/java/org/jboss/as/xts/txnclient/WildflyTransactionClientTxBridgeIntegrationHandler.java
```diff
@@ -73,12 +73,11 @@ public boolean handleMessage(MessageContext messageContext) {
...
-                // suspending context before returning to the client
-                if(ContextTransactionManager.getInstance() != null &&
-                        ContextTransactionManager.getInstance().getStatus() != Status.STATUS_NO_TRANSACTION) {
-                    ContextTransactionManager.getInstance().suspend();
-                }
+            if (((isOutbound != null) && isOutbound)
+                    // suspending context before returning to the client
+                    && ((ContextTransactionManager.getInstance() != null)
+                            && (ContextTransactionManager.getInstance().getStatus() != Status.STATUS_NO_TRANSACTION))) {
+                ContextTransactionManager.getInstance().suspend();
```
yersan, line 80: In favor of code's readability, which is the original intention of these changes, I would remove the unnecessary parenthesizes on these kinds of expressions, this one for example would be:

```
if (isOutbound != null && isOutbound
        // suspending context before returning to the client
        && ContextTransactionManager.getInstance() != null
        && ContextTransactionManager.getInstance().getStatus() != Status.STATUS_NO_TRANSACTION) {
```

Which, in my opinion, is more clear

boris-unckel: @yersan Hello Yeray, thanks for your feedback. Before I start to polish the PRs: Do you recommend to rely only on the operator precedence? Concrete for most cases I've seen: Keep parenthesizes only for existing OR || statements? Second question: I've used the automatic Eclipse formatter. Do you prefer manually formatted statements with one condition in each line and/or logical groups?

yersan: Hello @boris-unckel 

>  Do you recommend to rely only on the operator precedence? 

Not really, I saw that `if` the sentence, and was pretty clear to me that the parenthesizes were adding nothing, and without them code looks cleaner to me. I'm not trying to establish any rule here and it was just a personal point of view, which indeed is not good for a code review. So, do not feel forced to change the code if you think it is not necessary; notice I'm not requesting changes here.

> Do you prefer manually formatted statements with one condition in each line and/or logical groups?

You should autoformat your code with the help of your IDE, but in general, long lines are more difficult to read and make code review harder, I would say, just follow your common sense.

boris-unckel: @yersan I have adopted your recommendations, here and in all other PRs. Tests passed. Could you please review and approve the PR?

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
```diff
@@ -45,11 +45,10 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                 jsf = servlet;
             }
         }
-        if(jsf != null) {
-            if(jsf.getMultipartConfig() == null) {
-                //WFLY-2329 File upload doesn't work
-                jsf.setMultipartConfig(new MultipartConfigMetaData());
-            }
+        if (jsf != null
+                && jsf.getMultipartConfig() == null) {
```
yersan, line 49: A nitpick, these kinds of checks can be done in a single line. The idea of splitting them into several lines when you have multiple checks is to facilitate knowing which condition fails when we are debugging. However, checks like x != null && x.something does not require multiple lines, because if x is null, you will get an NPE. What's the point of being able to evaluate first if x != null is you know that the check gets an NPE?

xts/src/main/java/org/jboss/as/xts/txnclient/WildflyTransactionClientTxBridgeIntegrationHandler.java
yersan: A nitpick regarding `isOutbound != null && isOutbound`, these kinds of checks can be done in a single line. 

The idea of splitting them into several lines when you have multiple checks is to facilitate knowing which condition fails when we are debugging. However, checks like x != null && x.something does not require multiple lines, because if x is null, you will get an NPE. What's the point of being able to evaluate first if x != null is you know that the check gets an NPE?

boris-unckel: Can we agree on a compromise: I'll fix it whenever there is need to push --force. The coding/push effort is very low, obviously. But even on the weekend some CI runs had to be canceled due to higher prioritized workload. I didn't put in a newline for each and every condition, i.E. where a boolean is leading and the expression is short.

boris-unckel:
@yersan Hello Yeray, I have fixed the remaining issues. Please review and approve.

yersan (approved):
LGTM, thank you @boris-unckel

boris-unckel:
@bstansberry Another one with tests passed and approved. Could you check and merge, please?

PR #14446, opened 2021-07-05
WFLY-14844 Make appclient dependency on PicketBox Optional

sudeshnas93:
Issue: https://issues.redhat.com/browse/WFLY-14844


kabir (approved):

darranl:
@sudeshnas93 The testing of appclient is quite light within our testsuite, can I ask which additional checks have you performed to verify that appclient can be used without the picketbox module being present in the module hierarchy?

sudeshnas93:
> @sudeshnas93 The testing of appclient is quite light within our testsuite, can I ask which additional checks have you performed to verify that appclient can be used without the picketbox module being present in the module hierarchy?

Hello, I have checked wildfly testsuite, but did not find any error while building it. Also, I tried removing picketbox dependency from module.xml still there was no error. Do you suggest to check somewhere else in the testsuite?

bstansberry:
@darranl @sudeshnas93 I think the thing to do here is merge this and see what happens with TCK. It's the first day of WF 26 dev so today is the day we have maximum time to run an experiment.

I believe the goal is to remove the picketbox module altogether anyway, so we're not going down a strange path with this step.

PR #14448, opened 2021-07-05
[WFLY-14972] Check emptiness with Collection.isEmpty() (EJB3)

boris-unckel:
Fixes: https://issues.redhat.com/browse/WFLY-14972

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
@@ -383,7 +383,7 @@ public class EJBTransformers implements ExtensionTransformerRegistration {
             if (!fileStore.isDefined()) {//happens where default is not file-store
                 rejectIncompatibleDataStores(context, address);
             } else if ((untransformedModel.hasDefined(EJB3SubsystemModel.DATABASE_DATA_STORE)
-                    && untransformedModel.get(EJB3SubsystemModel.DATABASE_DATA_STORE).keys().size() > 0)
+                    && !untransformedModel.get(EJB3SubsystemModel.DATABASE_DATA_STORE).keys().isEmpty())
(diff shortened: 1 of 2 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBTransformers.java
```diff
@@ -383,8 +383,8 @@ public void transformResource(ResourceTransformationContext context, PathAddress
             if (!fileStore.isDefined()) {//happens where default is not file-store
                 rejectIncompatibleDataStores(context, address);
             } else if ((untransformedModel.hasDefined(EJB3SubsystemModel.DATABASE_DATA_STORE)
-                    && untransformedModel.get(EJB3SubsystemModel.DATABASE_DATA_STORE).keys().size() > 0)
-                    || untransformedModel.get(EJB3SubsystemModel.FILE_DATA_STORE).keys().size() > 1) {
+                    && !untransformedModel.get(EJB3SubsystemModel.DATABASE_DATA_STORE).keys().isEmpty())
+                    || !untransformedModel.get(EJB3SubsystemModel.FILE_DATA_STORE).keys().isEmpty()) {
                 rejectIncompatibleDataStores(context, address);
```
chengfang, line 388: the 2nd line of change: size() > 1 should be preserved (meaning it has 2 or more file data stores).

But I don't see much benefits of switching to isEmpty().

boris-unckel: > the 2nd line of change: size() > 1 should be preserved (meaning it has 2 or more file data stores).

I'll fix that.

> But I don't see much benefits of switching to isEmpty().

I have put a notice in the Jira: It's about performance - isEmpty() has O ( 1 ), 
some implementations of .size() have O ( n ). I did not check the concrete cases but refactored all occurrences. Special cases may vary, i.E. with only few elements in the collection.

chengfang (approved):

PR #14453, opened 2021-07-07
[WFLY-14975] Inappropriate "Collection" key/value issue

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-14975

### DIFF
--- a/picketlink/src/main/java/org/wildfly/extension/picketlink/common/model/AbstractResourceDefinition.java
+++ b/picketlink/src/main/java/org/wildfly/extension/picketlink/common/model/AbstractResourceDefinition.java
@@ -119,15 +119,12 @@ public abstract class AbstractResourceDefinition extends SimpleResourceDefinitio
             childResourceDefinitions.put(resourceDefinitionKey, childResources);
         }
 
-        if (!childResources.contains(resourceDefinitionKey)) {
-            for (ResourceDefinition childResource : childResources) {
-                if (childResource.getPathElement().getKey().equals(resourceDefinition.getPathElement().getKey())) {
-                    return;
-                }
+        for (ResourceDefinition childResource : childResources) {
+            if (childResource.getPathElement().getKey().equals(resourceDefinition.getPathElement().getKey())) {
+                return;
             }
-
-            childResources.add(resourceDefinition);
         }
+        childResources.add(resourceDefinition);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


boris-unckel:
@tadamski Hello Tomasz, could you please review this PR and approve? Tests have passed.

picketlink/src/main/java/org/wildfly/extension/picketlink/common/model/AbstractResourceDefinition.java
```diff
@@ -119,7 +119,7 @@ private void addChildResourceDefinition(ModelElement resourceDefinitionKey, Reso
             childResourceDefinitions.put(resourceDefinitionKey, childResources);
         }
 
-        if (!childResources.contains(resourceDefinitionKey)) {
+        if (!childResources.contains(resourceDefinition)) {
```
pferraro, line 122: Since childResources is an ArrayList, this "if" statement is completely redundant with the logic following it.

pferraro (changes requested):

picketlink/src/main/java/org/wildfly/extension/picketlink/common/model/AbstractResourceDefinition.java
pferraro: Sorry - I should have been more clear...
By redundant, I meant at the if statement should go away.
Because the ResourceDefinitions in question do not implement equals(...), the contains(...) method will use reference equality to determine if the resource definition is contained in the list.
Since the logic below is more permissive, we want to remove the if statement and instead use the logic that follows it (which determines equality based on the ResourceDefinition's path key.

pferraro (changes requested):

picketlink/src/main/java/org/wildfly/extension/picketlink/common/model/AbstractResourceDefinition.java
boris-unckel: ok, got that. NP to fix. Can you please remove the PR from the CI queue and ping me again?

pferraro: > can you remove the PR from the CI queue

Done.

pferraro (approved):

boris-unckel:
@bstansberry Hello Brian, another one with test passed and approved. Please, could you check and merge?

PR #14466, opened 2021-07-12
[WFLY-14809] Add Bouncycastle mail Jakarta version

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-14809

WildFly Preview Jobs:
https://ci.wildfly.org/viewLog.html?buildId=263192&tab=buildResultsDiv&buildTypeId=WF_WildFlyPreviewLinuxJdk11

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -74,6 +74,7 @@
         <version.jakarta.websocket.jakarta-websocket-api>2.0.0</version.jakarta.websocket.jakarta-websocket-api>
         <version.jakarta.ws.rs.jakarta-ws-rs-api>3.0.0</version.jakarta.ws.rs.jakarta-ws-rs-api>
         <version.jakarta.xml.bind.jakarta-xml-bind-api>3.0.0</version.jakarta.xml.bind.jakarta-xml-bind-api>
+        <version.org.bouncycastle>1.69</version.org.bouncycastle>
         <version.org.eclipse.yasson>2.0.1</version.org.eclipse.yasson>
         <version.org.glassfish.jakarta.el>4.0.0</version.org.glassfish.jakarta.el>
         <version.org.glassfish.jakarta.enterprise.concurrent>2.0.0</version.org.glassfish.jakarta.enterprise.concurrent>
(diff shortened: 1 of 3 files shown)


ee-9/feature-pack/pom.xml
```diff
@@ -74,6 +74,7 @@
         <version.jakarta.websocket.jakarta-websocket-api>2.0.0</version.jakarta.websocket.jakarta-websocket-api>
         <version.jakarta.ws.rs.jakarta-ws-rs-api>3.0.0</version.jakarta.ws.rs.jakarta-ws-rs-api>
         <version.jakarta.xml.bind.jakarta-xml-bind-api>3.0.0</version.jakarta.xml.bind.jakarta-xml-bind-api>
+        <version.org.bouncycastle>1.69</version.org.bouncycastle>
```
boris-unckel, line 77: In the past WF and WFCore had separate versions for bouncycastle. This has been refactored to a single point in WFCore. From my experience Bouncycastle always has exactly one version for all artifacts and separates JDK compatibility and/or standards always with a maven coordinate. Is it possible to rely on one version [wildfly-core/pom.xml#L173](https://github.com/wildfly/wildfly-core/blob/main/pom.xml#L173) ?

bstansberry: It's not possible to rely on the version from wildfly-core/pom.xml as its maven properties are not available to the WildFly build. It's dependencyManagement entries are as the root WF pom.xml  dependencyManagement imports core's pom, but that doesn't result in maven properties being available.

That said, this line should be removed because the same property from WildFly's root pom.xml is available here. We'd only override the prop here is we needed a different version in WildFly Preview.

jamezp: I actually don't see it defined in the root pom. Given that it kind of seems like the right place for it here because it's only going to replace a dependency for WildFly Preview.

yersan: Right, that's because the Jakarta dependency is not currently being brought in by wildfly-core, certainly, it could be an improvement and drive the bcjmail release from a core upgrade and/or override it in the preview when necessary. I don't have any argument against it, we will be polluting maven with dependencies unneeded for non-Jakarta EE flavors, but it sounds harmless if the dependencies themselves are not declared.

bstansberry: @yersan Oh, I assumed/misremembered that bcmail-jdk15on was declared as a dep in the full WF root pom.xml, not in the WF Core pom, since it is unused in core.

I understand why it isn't done that way, since the whole set presumably needs to be on the same version.  It probably makes sense to handle this artifact the same.

That said. let's do that via a follow up change rather than waiting another 2 weeks.

PR #14472, opened 2021-07-13
[WFLY-14912] Final step for vault removal.

darranl:
This is the final step for vault removal.

https://issues.redhat.com/browse/WFLY-14912
https://issues.redhat.com/browse/WFLY-14996

This depends on the following PR to WildFly Core:
    https://github.com/wildfly/wildfly-core/pull/4640


docs/src/main/asciidoc/_migration-guide/Vault.adoc
```diff
@@ -0,0 +1,271 @@
...
+
+=== Defining the credential store in the management model
+
+This newly created credential store can be added to the management model of a running server.
+
+[source]
+----
+/subsystem=elytron/credential-store=mystore:add( \
+    relative-to=jboss.server.config.dir, \
+    location=credentials.store, \
```
OndrejKotek, line 37: `location` is deprecated, `path` should be used instead

bstansberry:
I figure the Galleon - Linux job failure is related.

Here are WF Preview runs:

https://ci.wildfly.org/viewLog.html?buildId=267152&tab=buildResultsDiv&buildTypeId=WF_Ee9BootableJarLinuxJdk8
https://ci.wildfly.org/viewLog.html?buildId=267150&tab=buildResultsDiv&buildTypeId=WF_WildFlyPreviewLinuxJdk11

bstansberry:
The he Galleon - Linux job failure has a poor failure message. Here's what it means:

The test-standalone-reference-provisioning execution provisions using servlet-feature-pack's standalone.xml (i.e. no layers). So everything should be provisioned, equivalent to building servlet-dist.

The all-layers-provisioning execution provisions what's meant to be a complete set of layers.

The test compares the modules in the two provisioned installation and fails if any from the test-standalone-reference-provisioning are missing in the all-layers-provisioning. So javax.xml.stream.api no longer gets provisioned by all-layers-provisioning.

bstansberry:
LayersTestCase.NOT_USED is the list of modules that are expected not to be provisioned. So the fix is likely just adding javax.xml.stream.api to the list.

bstansberry:
@darranl I filed #14543 to let CI confirm my thinking about that test. Please do with it what you will.

darranl:
FYI I am trying to reproduce locally as well but think I hit an unrelated failure.

bstansberry (approved):

PR #14475, opened 2021-07-14
Reactive Messaging 2.0 Upgrade and related features

kabir:
This contains the following feature requests:

* Upgrade to Reactive Messaging 2.0
    * https://issues.redhat.com/browse/WFLY-14798 /  https://github.com/wildfly/wildfly-proposals/pull/395
* Kafka API to set/read key and other Kafka variables
    * https://issues.redhat.com/browse/WFLY-14932 /  https://github.com/wildfly/wildfly-proposals/pull/402
* The MP Reactive Messaging Kafka connector should be able to use an SSLContext configured in the Elytron subsystem
    * https://issues.redhat.com/browse/WFLY-14987 /  https://github.com/wildfly/wildfly-proposals/pull/411

I decided to put them together since the latter two depend on the [WFLY-14798](https://issues.redhat.com/browse/WFLY-14798) upgrade, and I want to get some eyes on this while waiting for testing to happen.

If desired, once we get closer to this being mergeable, I can break this up into individual pull requests. 

[WFLY-14932](https://issues.redhat.com/browse/WFLY-14932) and [WFLY-14987](https://issues.redhat.com/browse/WFLY-14987) have no inter-dependencies, so any order of merging these is fine. As already mentioned, [WFLY-14798](https://issues.redhat.com/browse/WFLY-14798) needs to go in before these two.


microprofile/galleon-common/src/main/resources/modules/system/layers/base/io/smallrye/reactive/messaging/main/module.xml
```diff
@@ -58,5 +58,7 @@
         <module name="org.reactivestreams"/>
         <module name="org.slf4j"/>
         <module name="org.wildfly.reactive.messaging.config" services="import"/>
+        <module name="io.smallrye.metrics"/>
+        <module name="org.eclipse.microprofile.metrics.api"/>
```
bstansberry, line 62: Unless I've missed something, this a class visibility only dependency then, with no config requirement (i.e. there's no requirement to add the MP Metrics extensions/subsystem.)

If so that's good. Adding libs to the server footprint isn't great, but requiring adding endpoint-exposing subsystems to the config is yuck.

microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/MicroProfileReactiveMessagingSubsystemDefinition.java
```diff
@@ -104,13 +112,40 @@ private AddHandler() {
...
             MicroProfileReactiveMessagingLogger.LOGGER.activatingSubsystem();
         }
+
+        private void installKafkaElytronSSLContextRegistryServiceIfPresent(OperationContext context) {
+            ClassLoader cl = WildFlySecurityManager.getClassLoaderPrivileged(this.getClass());
+            if (cl instanceof ModuleClassLoader) {
+                ModuleLoader loader = ((ModuleClassLoader)cl).getModule().getModuleLoader();
+                try {
+                    loader.loadModule("org.wildfly.reactive.messaging.kafka");
+                    ElytronSSLContextRegistryService.install(context.getServiceTarget(), context.getServiceRegistry(false));
```
bstansberry, line 144: Without looking at the method being called, I assume this should be context.getServiceRegistry(true) as it will modify.

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ElytronSSLContextRegistryService.java
```diff
@@ -0,0 +1,77 @@
...
+        target.addService(name)
+                .setInstance(INSTANCE)
+                .install();
+        INSTANCE.serviceRegistry = unmodifiableServiceRegistry;
+    }
+
+    static SSLContext getInstalledSSLContext(String name) {
+        ServiceController<SSLContext> controller = (ServiceController<SSLContext>)INSTANCE.serviceRegistry
+                .getService(getSSLContextName(name));
+        return controller.getValue();
```
bstansberry, line 63: This can/should throw ISE if the service is not UP.

Also, serviceRegistry.getService should return null if the service is not installed.

AIUI there are two users of this method, one trying to check if the service is installed, so it can add an MSC dep, and then another, which I guess won't be called until that dep is satisfied, which looks up the service value.

I think those two use cases should use separate method calls, and the impl or callers of both need to deal with the failure conditions MSC says can happen.

I think all of this needs generous code comments too.

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ElytronSSLContextRegistryService.java
```diff
@@ -0,0 +1,77 @@
...
+import org.jboss.msc.service.ServiceRegistry;
+import org.jboss.msc.service.ServiceTarget;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+
+/**
+ * @author <a href="mailto:kabir.khan@jboss.com">Kabir Khan</a>
+ */
+public class ElytronSSLContextRegistryService implements Service {
```
bstansberry, line 39: Why is this a Service? AFAICT no code adds a dep on it and it doesn't do anything in start/stop.

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ReactiveMessagingSslConfigProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+
+    private String getSSLEngineFactoryClassName() {
+        try {
+            Class<?> clazz = WildFlyKafkaSSLEngineFactory.class;
+            return clazz.getName();
+        } catch (Throwable throwable) {
+            return null;
```
bstansberry, line 111: A comment explaining why returning null is ok would be good.

I haven't verified but my assumption is the reason is the module that provides WildFlyKafkaSSLEngineFactory may not be provisioned, and that's ok.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Reactive_Messsaging_SmallRye.adoc
```diff
@@ -40,14 +42,69 @@ If you provision the `microprofile-reactive-messaging-kafka` Galleon layer it in
...
+}
+----
+
+
+In the above example we can now easily send data to the Reactive Messaging streams by calling `Emitter.send()`. Similarly, we can subscribe to the Publisher and receive the data. However, receiving still has a few shortcomings:
+
+* The above example will not work out of the box. When trying to send on the `Emitter`, you will get an error that there are no subscibers (which in turn runs the risk of causing overflow). This can be worked around by creating a subscription on the `Publisher`.
+* At present there can only be one subscription on the injected `Publisher`.
+
+The above points means that this `Publisher` is not usable directly as an asynchronous return value for e.g. a JAX-RS endpoint. As the JAX-RS request is what will create the subscription such a call would need to happen before anybody tries to `Emitter.send()`.
```
bstansberry, line 102: s/JAX-RS/Jakarta RESTFul Webservices/g 

Everywhere. :(

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Reactive_Messsaging_SmallRye.adoc
```diff
@@ -119,7 +176,109 @@ Next we will briefly discuss each of these entries. Remember the `to` channel is
...
+* `mp.messaging.connector.smallrye-kafka.sasl.jaas.config=...` - specifies how we will authenticate with Kafka. In order to not hardcode the credentials in our `microprofile-config.properties file` we are using the property substitution feature of MicroProfile Config. In this case, if you have defined the `USER` and `PASSWORD` environment variables they will be passed in as part of the configuration
+* `mp.messaging.connector.smallrye-kafka.wildfly.elytron.ssl.context=test` - this is not needed if Kafka is secured with a CA signed certificate. If you are using self-signed certificates, you will need to specify a truststore in the Elytron subsystem, and create an `SSLContext` referencing that. The value of this property is used to look up the `SSLContext` in the Elytron subsystem under `/subsystem=elytron/client-ssl-context=*` in the WildFly management model. In this case the property value is `test`, so we look up the `SSLContext` defined by `/subsystem=elytron/client-ssl-context=test` and use that configure the truststore to use for the connection to Kafka.
```
bstansberry, line 281: A lot of the stuff in this file raises the question of how we should structure docs in a MP Config / Jakarta Config world.

Most of what you've added is application developer documentation. But OTOH some of the config stuff is somewhat admin documentation. Dev ops, etc.

microprofile/reactive-messaging-smallrye/config/src/main/java/org/wildfly/microprofile/reactive/messaging/config/ReactiveMessagingConfigSource.java
```diff
@@ -65,6 +76,14 @@ public int getOrdinal() {
...
-        return PROPERTIES.keySet();
+        return properties.keySet();
+    }
+
+    public static void addProperties(Config config, Map<String, String> properties) {
+        for (ConfigSource source : config.getConfigSources()) {
+            if (source.getName().equals(NAME) && source instanceof ReactiveMessagingConfigSource) {
+                ((ReactiveMessagingConfigSource)source).properties.putAll(properties);
+            }
+        }
```
bstansberry, line 87: This method could use some comments related to scoping so it's understandable to non-experts.

AIUI the Config object passed in is deployment-scoped, and thus the ReactiveMessagingConfigSource is as well. So this method is populating a deployment-scoped object with deployment-scoped properties from the DUP.

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ReactiveMessagingSslConfigProcessor.java
```diff
@@ -0,0 +1,128 @@
...
+class ReactiveMessagingSslConfigProcessor implements DeploymentUnitProcessor {
+
+    private static final String RM_KAFKA_GLOBAL_PROPERTY_PREFIX = "mp.messaging.connector.smallrye-kafka.";
+    private static final String RM_INCOMING_PROPERTY_PREFIX = "mp.messaging.outgoing.";
+    private static final String RM_OUTGOING_PROPERTY_PREFIX = "mp.messaging.incoming.";
+    static final String SSL_CONTEXT_PROPERTY_SUFFIX = "wildfly.elytron.ssl.context";
+
+    private static final String SSL_ENGINE_FACTORY_CLASS = "ssl.engine.factory.class";
+
+    private static final AttachmentKey<Object> DEPLOYMENT_ATTACHEMENT_KEY = AttachmentKey.create(Object.class);
```
bstansberry, line 59: Nit: s/ATTACHEMENT/ATTACHMENT

testsuite/integration/microprofile-tck/reactive-messaging/src/test/java/org/wildfly/test/integration/mp/tck/reactive/messaging/WildFlyDeploymentExceptionTransformer.java
```diff
@@ -22,16 +22,24 @@
...
+ * Workaround for https://issues.redhat.com/browse/WFARQ-59 and https://issues.redhat.com/browse/WFARQ-36.
+ * The TCK wrongly assumes that deployment exceptions will always result in a thrown exception of a specific type.
+ * This class checks the exception thrown to make sure the root cause on the server is what is expected.
+ *
  * @author <a href="mailto:kabir.khan@jboss.com">Kabir Khan</a>
  */
 public class WildFlyDeploymentExceptionTransformer implements DeploymentExceptionTransformer {
     @Override
     public Throwable transform(Throwable throwable) {
+        System.out.println("----- " + throwable);
```
bstansberry, line 40: Better to remove this.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationWithoutReadinessOperationTestCase.java
```diff
@@ -30,12 +30,18 @@
...
 
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2019 Red Hat inc.
  */
 public class MicroProfileHealthApplicationWithoutReadinessOperationTestCase extends MicroProfileHealthApplicationReadyTestBase {
 
+    @BeforeClass
+    public static void blah() {
+        System.out.println("Hello");
+    }
```
bstansberry, line 43: ?

microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/MicroProfileReactiveMessagingSubsystemDefinition.java
kabir: No, this Service is just a wrapper around the service registry. It uses that to look up SSLContexts installed by the Elytron subsystem, so read-only is fine

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ElytronSSLContextRegistryService.java
kabir: Hmm, I didn't want to use a singleton - and also wanted the ServiceRegistry. Although, I see now I am really using a singleton anyway. I can change this

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ReactiveMessagingSslConfigProcessor.java
kabir: I think this is a hangover from before I put this class into the same module as WildFlyKafkaSSLEngineFactory - so the load should always succeed. Will fix

microprofile/reactive-messaging-smallrye/config/src/main/java/org/wildfly/microprofile/reactive/messaging/config/ReactiveMessagingConfigSource.java
kabir: Added

microprofile/galleon-common/src/main/resources/modules/system/layers/base/io/smallrye/reactive/messaging/main/module.xml
kabir: This is the class https://github.com/smallrye/smallrye-reactive-messaging/blob/main/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java which deals with it. It relies on a class from the API jar but in the absence of a CDI provided MetricsRegistry becomes a noop.

I have modified the io.smallrye.metrics dependency to be optional and verified my QS works for the following three scenarios:
* A standalone-microprofile.xml server. This has the microprofile-metrics subsystem, and its MetricsRegistry is injected
* A standalone.xml server. This does not have the microprofile-metrics subsystem, so no MetricsRegistry is injected. This works fine both:
    * out of the box
    * with the io.smallrye.metrics module deleted

bstansberry: Thanks.

microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/MicroProfileReactiveMessagingSubsystemDefinition.java
bstansberry: You've changed how this works, so my comment is no longer applicable.

But true was the correct value before -- it modified runtime state by installing a service, so the op needs the exclusive write lock, which 'true' triggers. It doesn't much matter as this OSH already gets the exclusive right lock, but I watch out for that so people don't see code and do the same in places where it might matter.

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ElytronSSLContextRegistry.java
```diff
@@ -0,0 +1,58 @@
...
+    private volatile ServiceRegistry serviceRegistry;
+
+    private ElytronSSLContextRegistry() {
+    }
+
+    public static void setServiceRegistry(ServiceRegistry serviceRegistry) {
+        serviceRegistry = serviceRegistry;
+    }
+
+    static SSLContext getInstalledSSLContext(String name) {
```
bstansberry, line 48: I still think this needs to be split into two methods.

https://github.com/wildfly/wildfly/pull/14475/files#diff-d9d0091e8a499da58a7b0ae2d67bc83b134bb6aaa06beae3fe69404b1ee3898eR86 does not need the SSLContext; it just needs a true | false indicating if the ServiceController exists.

If it does not exist, L51 below will NPE.

If it does exist, but is not started (which could happen as there's no clear dep on the service from the DUPS running ReactiveMessagingSslConfigProcessor) then MSC is allowed to throw an exception at L51, which would unnecessarily fail the deployment.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Reactive_Messsaging_SmallRye.adoc
bstansberry: Let's leave this point aside for later.

microprofile/reactive-messaging-smallrye/config/src/main/java/org/wildfly/microprofile/reactive/messaging/config/ReactiveMessagingConfigSource.java
bstansberry: Thanks.

bstansberry (changes requested):
@kabir I still have once concern about ElytronSSLContextRegistry but otherwise LGTM.

microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/MicroProfileReactiveMessagingSubsystemDefinition.java
kabir: For the record, it used `context.getServiceTarget()` to install the service itself. And (as it does now) it used the read-only registry to delegate the lookups to

microprofile/reactive-messaging-smallrye/kafka/src/main/java/org/wildfly/microprofile/reactive/messaging/config/kafka/ssl/context/ElytronSSLContextRegistry.java
kabir: I totally forgot about this point - sorry!

bstansberry (approved):

kabir:
Superseded by #14618, #14619 and #14620 which split up this PR into the individual features.

PR #14476, opened 2021-07-14
[WFLY-15008] Check null in TracingDeploymentProcessor.close()

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15008


boris-unckel:
@chengfang Hello Cheng, I have prepared the PR. I'm wondering about the CI status because the PRs number appears in the queue but here it's marked as "all checks have passed" with a single check. Please review and approve.

chengfang (approved):

rhusar:
@boris-unckel Looks like this regression was caused by your commit - https://github.com/wildfly/wildfly/commit/7a7174b3b278a0dfca9436f384169c7afa69529e#diff-993040b4ff01e1d0aab5d16ad446552b62f7045585a2e94eef9bd315cf621e41L284 - since instanceof is a null check as well.

boris-unckel:
@bstansberry Hello Brian, a prior change by me caused a regression. This is the fix, it's approved. Could you check and merge, please?

PR #14479, opened 2021-07-15
[WFLY-14957] Correct the default behavior

bstansberry:
Fix the https://issues.redhat.com/browse/WFLY-14957 fix


bstansberry:
@jamezp @chengfang fyi

PR #14484, opened 2021-07-16
WFLY-14902 Clustering TS: Enable authentication and authorization for Infinispan Server tests + enable on Windows + instances provisioned by testsuite never shutdown

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-14902
https://issues.redhat.com/browse/WFLY-13883
https://issues.redhat.com/browse/WFLY-13897

Requires
https://github.com/wildfly/wildfly/pull/14533

### DIFF
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -618,8 +618,6 @@
                                     <goal>test</goal>
                                 </goals>
                                 <configuration>
-                                    <!-- Workaround for @ClassRule Infinispan Server -->
-                                    <reuseForks>false</reuseForks>
                                     <!-- Fix test order to speed up execution by grouping tests that run against same group of containers -->
                                     <runOrder>alphabetical</runOrder>
                                     <!-- Tests to execute. -->
--- a/testsuite/integration/clustering/pom.xml
+++ b/testsuite/integration/clustering/pom.xml
@@ -46,6 +46,9 @@
         <test-group>org/jboss/as/test/clustering/cluster</test-group>
         <version.org.jboss.byteman>4.0.6</version.org.jboss.byteman>
         <version.org.infinispan.server>12.1.7.Final</version.org.infinispan.server>
+        <version.org.infinispan.server.driver>13.0.0.Dev03</version.org.infinispan.server.driver>
+        <!-- Avoids REST call timeout on cluster shutdown -->
+        <version.com.squareup.okhttp3>3.14.8</version.com.squareup.okhttp3>
         <ts.surefire.clustering.ha.additionalExcludes>nothing-by-default</ts.surefire.clustering.ha.additionalExcludes>
         <!-- properties to enable plugins shared by various bootable jar executions -->
         <bootable-jar-cloud-profile-packaging.phase>none</bootable-jar-cloud-profile-packaging.phase>
@@ -113,26 +116,14 @@
         <dependency>
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-server-testdriver-junit4</artifactId>
-            <version>${version.org.infinispan}</version>
+            <version>${version.org.infinispan.server.driver}</version>
             <exclusions>
+                <!-- Avoids issue on deploy - "Caused by: org.jboss.as.cli.CommandRegistry$RegisterHandlerException: The following command names could not be registered since they conflict with the already registered ones: cd" -->
                 <exclusion>
                     <groupId>org.infinispan</groupId>
-                    <artifactId>infinispan-server-runtime</artifactId>
+                    <artifactId>infinispan-cli-client</artifactId>
                 </exclusion>
             </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-client-rest</artifactId>
-            <version>${version.org.infinispan}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-commons</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-client-hotrod</artifactId>
         </dependency>
         <!-- Arquillian extension for Byteman -->
         <dependency>
@@ -409,7 +400,7 @@
                                     <outputDirectory>${basedir}/target/infinispan-server-${version.org.infinispan.server}/server/conf</outputDirectory>
                                     <resources>
                                         <resource>
-                                            <directory>src/test/resources/infinispan-server</directory>
+                                            <directory>src/test/resources/infinispan-server-overlay</directory>
                                             <filtering>true</filtering>
                                         </resource>
                                     </resources>
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/sso/remote/InfinispanServerSetupTask.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/sso/remote/InfinispanServerSetupTask.java
@@ -26,7 +26,6 @@ import static org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase.IN
 import static org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase.INFINISPAN_APPLICATION_USER;
 import static org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase.INFINISPAN_SERVER_ADDRESS;
 import static org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase.INFINISPAN_SERVER_PORT;
-import static org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase.INFINISPAN_SERVER_PROTOCOL_VERSION;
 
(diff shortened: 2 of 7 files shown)


testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/sso/remote/InfinispanServerSetupTask.java
```diff
@@ -36,9 +38,7 @@
...
-// Disable infinispan-server security until test driver issues are fixed
-//                .setup("/subsystem=infinispan/remote-cache-container=sso:add(default-remote-cluster=infinispan-server-cluster, module=org.wildfly.clustering.web.hotrod, properties={infinispan.client.hotrod.auth_username=testsuite-user,infinispan.client.hotrod.auth_password=testsuite-password}, protocol-version=%s)", INFINISPAN_SERVER_PROTOCOL_VERSION)
-                .setup("/subsystem=infinispan/remote-cache-container=sso:add(default-remote-cluster=infinispan-server-cluster, module=org.wildfly.clustering.web.hotrod, protocol-version=%s)", INFINISPAN_SERVER_PROTOCOL_VERSION)
+                .setup("/subsystem=infinispan/remote-cache-container=sso:add(default-remote-cluster=infinispan-server-cluster, module=org.wildfly.clustering.web.hotrod, properties={infinispan.client.hotrod.auth_username=%s, infinispan.client.hotrod.auth_password=%s}, protocol-version=%s)", INFINISPAN_APPLICATION_USER, INFINISPAN_APPLICATION_PASSWORD, INFINISPAN_SERVER_PROTOCOL_VERSION)
```
pferraro, line 41: Do we need to specify the protocol version any longer?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
```diff
@@ -434,43 +438,55 @@ public void test(@ArquillianResource(SessionOperationServlet.class) @OperateOnDe
...
                         }
-                    } finally {
-                        HttpClientUtils.closeQuietly(response);
                     }
+                    // Wait a second between retries
+                    Thread.sleep(TimeUnit.SECONDS.toMillis(1));
                 }
             }
-            Assert.assertTrue(destroyed);
+            Assert.assertTrue("Session has not been destroyed following expiration within " + maxAttempts + " seconds.", destroyed);
```
pferraro, line 489: The assertion message needs to be updated.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/remote/InfinispanServerSetupTask.java
```diff
@@ -37,9 +39,7 @@
...
-// Disable infinispan-server security until test driver issues are fixed
-//                .setup("/subsystem=infinispan/remote-cache-container=web:add(default-remote-cluster=infinispan-server-cluster, tcp-keep-alive=true, marshaller=PROTOSTREAM, module=org.wildfly.clustering.web.hotrod, properties={infinispan.client.hotrod.auth_username=testsuite-user, infinispan.client.hotrod.auth_password=testsuite-password}, protocol-version=%s, statistics-enabled=true)", INFINISPAN_SERVER_PROTOCOL_VERSION)
+                .setup("/subsystem=infinispan/remote-cache-container=web:add(default-remote-cluster=infinispan-server-cluster, tcp-keep-alive=true, marshaller=PROTOSTREAM, module=org.wildfly.clustering.web.hotrod, properties={infinispan.client.hotrod.auth_username=%s, infinispan.client.hotrod.auth_password=%s}, protocol-version=%s, statistics-enabled=true)", INFINISPAN_APPLICATION_USER, INFINISPAN_APPLICATION_PASSWORD, INFINISPAN_SERVER_PROTOCOL_VERSION)
```
pferraro, line 42: Same question re: protocol-version.

testsuite/integration/clustering/src/test/resources/infinispan.xml
```diff
@@ -47,27 +57,21 @@
...
                     </properties-realm>
                 </security-realm>
             </security-realms>
         </security>
 
-        <endpoints socket-binding="default">
-            <hotrod-connector name="hotrod"/>
-            <rest-connector name="rest"/>
-        </endpoints>
+        <endpoints socket-binding="default" security-realm="default"/>
```
pferraro, line 75: Why were the connectors removed?

testsuite/integration/clustering/pom.xml
```diff
@@ -113,26 +114,20 @@
...
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-commons</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.infinispan</groupId>
-            <artifactId>infinispan-client-hotrod</artifactId>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>3.14.8</version>
```
pferraro, line 130: Create a property for this version.

pferraro (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
rhusar: How so?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/sso/remote/InfinispanServerSetupTask.java
rhusar: The original idea was to eventually test against different datagrid versions requiring different protocol versions. But with the protocol negotiation after 3.0 in place, we should not need this any longer. I ll remove.

testsuite/integration/clustering/src/test/resources/infinispan.xml
rhusar: Because that's a typical infinispan server configuration. Declaring an empty <endpoints/> element implicitly enables the Hot Rod and REST connectors.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/remote/InfinispanServerSetupTask.java
rhusar: Same response, will remove.

testsuite/integration/clustering/pom.xml
rhusar: Do we need such a property? There is already a property for com.squareup.okhttp3 however its using an outdated version at the moment, hopefully this will eventually go away when jaegertracing uses a later version.

testsuite/integration/clustering/src/test/resources/infinispan.xml
rhusar: s/typical/default

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
pferraro: maxAttempts is not a measure of seconds.

testsuite/integration/clustering/src/test/resources/infinispan.xml
pferraro: OK - that's a strange choice.

testsuite/integration/clustering/pom.xml
pferraro: This was the properties (and their intent) are more visible.  This pom can override the property with no problem, no?

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
rhusar: Well, each attempt takes a second, so it is... changed wording to attempts if its causing confusion. Anyway, any message is better than no message since line numbers change one needs to look into the code to see when the test actually failed.

testsuite/integration/clustering/src/test/resources/infinispan.xml
rhusar: Not mine, theirs! I try to keep the infinispan config as close to their defaults as possible and this is their default now.

testsuite/integration/clustering/pom.xml
rhusar: I see what you mean now, removed the dependency and just override the property and it works. Originally I didn't realize we already manage this dependency in the main pom.xml; fixed now.

rhusar:
We have analyzed the test failure is caused by https://issues.redhat.com/browse/WFLY-15088 will include fix in this PR shortly.

rhusar:
@pferraro CI runs are passing now, lets do this :-)

pferraro (approved):

PR #14485, opened 2021-07-16
[WFLY-15035][WFLY-15036]  Use source code transformation to produce a native jakarta namespace variant of the mail subsystem artifact

bstansberry:
https://issues.redhat.com/browse/WFLY-15035
Also https://issues.redhat.com/browse/WFLY-15036

### DIFF
--- a/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/mail/main/module.xml
+++ b/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/mail/main/module.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2010, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.jboss.as.mail">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <!-- TODO implement WFGP-206 or use maven-resource-plugin filtering to drive this
+             value with an expression so we can use one source file for both javax and jakarta variants
+             of this module -->
+        <artifact name="${org.wildfly:wildfly-mail-jakarta}"/>
+    </resources>
+
+    <dependencies>
+        <!--WFLY-14219 Remove deprecated <module name="javax.api"/> -->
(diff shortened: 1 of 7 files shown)


bstansberry:
WF Preview tests:

https://ci.wildfly.org/viewLog.html?buildId=264432&buildTypeId=WF_Ee9BootableJarLinuxJdk8
https://ci.wildfly.org/viewLog.html?buildId=264434&buildTypeId=WF_WildFlyPreviewLinuxJdk11

First one passes; second has two failures we are seeing in many places so not likely relevant, discussed at https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/ClusteredMessagingTestCase.20failures

ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/mail/main/module.xml
```diff
@@ -0,0 +1,60 @@
...
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <!-- TODO implement WFGP-206 or use maven-resource-plugin filtering to drive this
+             value with an expression so we can use one source file for both javax and jakarta variants
+             of this module -->
+        <artifact name="${org.wildfly:wildfly-mail-jakarta}"/>
```
yersan, line 35: Would it be a possibility to use the same artifact instead of renaming it? I mean, instead of having a wildfly-mail-jakarta version, use just wildfly-mail.

That would avoid the need to overwriting the artifact in the module.xml and deal with the license stuff. 

After all the transformed artifact is built when the server is built. The problem is the artifact is installed in the local repository, so, if you are moving from Jakarta EE to Java EE locally, you need to build the server and you could have no certainty about which artifact are you using. Not sure if there could be other bad sides. So, just asking.

mail/src/test/java/org/jboss/as/mail/extension/MailTransformersTestCase.java
```diff
@@ -58,6 +62,13 @@ public MailTransformersTestCase() {
         super(MailExtension.SUBSYSTEM_NAME, new MailExtension());
     }
 
+    @BeforeClass
+    public static void assumeJavaxNamespace() {
+        //noinspection ConstantConditions
+        Assume.assumeTrue("Ignoring test that requires javax.* namespace",
```
yersan, line 68: It should be ignoring a test that requires jakarta.* namespace

yersan (changes requested):

mail/src/test/java/org/jboss/as/mail/extension/MailTransformersTestCase.java
bstansberry: No, the test requires that the javax namespace be in effect.  It fails in the wildfly-mail-jakarta module because it uses jakarta.* but the legacy controller needs javax.

bstansberry: Well, my last response didn't address your valid point. I'll make it "Ignoring test when run using the jakarta.* namespace EJB API"

ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/as/mail/main/module.xml
bstansberry: The requirement is to produce both javax.* and jakarta.* artifacts in the same build and deploy them to nexus. People using a jakarta.* namespace variant of WF should be able to provision a server solely with artifacts pulled from remote maven, with no provisioning-time transformation.

See https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/LE6CBNDI3GNXWVX4DOZLU5QNPZQWFJ7K/ for a discussion of how to avoid duplicating the module.xml. But the license stuff I think is unavoidable.

yersan: Yes, right, I understand. I raised an invalid approach here, although the wildfly-mail is built by the server, we still need to make a distinction, (version, classifier) so feature packs are able to retrieve the expected one.

yersan (approved):

bstansberry:
WFP: 

https://ci.wildfly.org/viewLog.html?buildId=267240&tab=buildResultsDiv&buildTypeId=WF_Ee9BootableJarLinuxJdk8
https://ci.wildfly.org/viewLog.html?buildId=267242&tab=buildResultsDiv&buildTypeId=WF_WildFlyPreviewLinuxJdk11

PR #14498, opened 2021-07-21
[WFLY-15047] EJB (Un)marshaller resource leak

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15047


chengfang:
marshaller and unmarshaller have a `finish()` method that does the cleanup. In fact, its `close()` method just calls `finish()`. So adding a call to `close()` will duplicate what's already done by `finish()` method for places where `finish()` is already called. So I think it's better to add a finally clause to make sure `finish()` is always called for cleanup.

boris-unckel:
@chengfang I should have checked that :-( I suggest to remove the finish() to avoid the duplication. try-with-resource is caring for the correct Exception/Suppressed Ex handling including to avoid NullPointer.

boris-unckel:
I have found a second issue here: `finish()` and `flush()` had the wrong order and `flush()` was sometimes in use, sometimes not. It's fixed now.

chengfang (approved):

boris-unckel:
Hello @bstansberry Tests passed for this PR and it's approved. Could you check and merge, please?

PR #14499, opened 2021-07-21
[WFLY-14956] adding transaction statistics attribute to group

ochaloup:
https://issues.redhat.com/browse/WFLY-14956

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionStatisticsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionStatisticsTestCase.java
@@ -1,0 +1,296 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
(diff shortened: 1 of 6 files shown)


ochaloup:
@mmusgrov please, can you take a look?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionStatisticsTestCase.java
```diff
@@ -0,0 +1,167 @@
...
+                .addAsManifestResource(createPermissionsXmlAsset(
+                    // ManagementClient needs the following permissions and a dependency on 'org.jboss.remoting3' module
+                    new RemotingPermission("createEndpoint"),
+                    new RemotingPermission("connect"),
+                    new FilePermission(System.getProperty("jboss.inst") + "/standalone/tmp/auth/*", "read")
+                ), "permissions.xml");
+    }
+
+    @Test
+    public void transactionStatisticsGrouped() throws Exception{
```
mmusgrov, line 116: Why did you only test the following attributes

```
number-of-transactions
number-of-committed-transactions
number-of-aborted-transactions
number-of-application-rollbacks
```

but not these ones:

```
average-commit-time
number-of-heuristics
number-of-inflight-transactions
number-of-nested-transactions
number-of-resource-rollbacks
number-of-system-rollbacks
number-of-timed-out-transactions
```

mmusgrov (approved):
A very useful operation. I approved it, though I did have one question about the test.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionStatisticsTestCase.java
ochaloup: Agree, it would be more appropriate to test all the properties being in group. And possibly trying to check if other statistics work.
I adjusted the PR to work with more statistics (those which are available without difficulties) and check if all of them are listed in the group. Can you  @mmusgrov look at again, please?

mmusgrov (approved):
The PR still looks good and still approved.

But my question had been about verifying that the group contains all of our stats. Something like the following:

```
ModelNode grouped = readGrouped();

and then test that the group contains all of our stats, something like

for each attribute in the group
    assert exists grouped.get(attribute name);

And I'd imagine that checking that just one of the stats are incremented correctly would suffice.
```

ochaloup:
@mmusgrov sure. I added what you talk about in a bit different manner but in the same functionality here in the PR: 
https://github.com/wildfly/wildfly/pull/14499/files#diff-7bb42ab4b13ca9b420e8ac3762335a6ceb56d71cc9cc1447a4e1fcb6e3c6d72eR179
I'm reading the expected values for all the statistic attributes. Does it make sense or do you think to create a separate `@Test` which will only iterate on existence?

mmusgrov:
I approved the PR so it fine to merge it.

PR #14504, opened 2021-07-23
[WFLY-15038] Kerberos authenticated database connections (e.g. to SQL Server) not reusable when using legacy security and remote EJBs

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-15038

Just adding an equals method to the wrapping `RemotingRemoteConnection` class. This way the picketbox cache for security domains is maintained when the EJB call is using the same connection as before. Indirect test added, not using kerberos, just checking the principal and credential obtained from picketbox are cachable.

@darranl Please do the review when you have time. No hurry, the elytron configuration is working better.


darranl:
I am going to close this as we are proceeding to the next stage to remove PicketBox support.

PR #14505, opened 2021-07-23
[WFLY-15055] Use javax module alias instead of the jakarta modules on…

yersan:
… the Bouncy Castle Jakarta Mail module

Jira issue: https://issues.redhat.com/browse/WFLY-15055

### DIFF
--- a/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/bouncycastle/bcmail/main/module.xml
+++ b/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/bouncycastle/bcmail/main/module.xml
@@ -32,8 +32,8 @@
     <dependencies>
         <!-- Required for the IBM J9 1.8 -->
         <module name="java.se"/>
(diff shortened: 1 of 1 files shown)


yersan:
https://ci.wildfly.org/viewLog.html?buildId=265385&buildTypeId=WF_WildFlyPreviewLinuxJdk11
https://ci.wildfly.org/viewLog.html?buildId=265396&buildTypeId=WF_Ee9BootableJarLinuxJdk8

chengfang:
Is it required to use javax alias module, or is it both will work but alias is better?

https://github.com/wildfly/wildfly/pull/14502 is similar in this regard, so we may also need to make the same change there.

bstansberry:
@chengfang I'm going to post on wildfly-dev about this in a minute.

bstansberry:
The post: https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/LE6CBNDI3GNXWVX4DOZLU5QNPZQWFJ7K/

PR #14509, opened 2021-07-24
[WFLY-15044] Avoid Nullpointer (Mail)

boris-unckel:

### DIFF
--- a/mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
+++ b/mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
@@ -41,6 +41,7 @@ import org.jboss.as.controller.CapabilityServiceTarget;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.controller.security.CredentialReference;
 import org.jboss.as.naming.ManagedReferenceFactory;
@@ -90,19 +91,20 @@ class MailSessionAdd extends AbstractAddStepHandler {
         installRuntimeServices(context, address, fullTree);
     }
 
-    private static void addCredentialStoreReference(ServerConfig serverConfig, OperationContext context, ModelNode model, ServiceBuilder serviceBuilder, String... modelFilter) throws OperationFailedException {
+    private static void addCredentialStoreReference(ServerConfig serverConfig, OperationContext context, ModelNode model, ServiceBuilder serviceBuilder, final PathElement credRefParentAddress) throws OperationFailedException {
         if (serverConfig != null) {
+            final String servertype = credRefParentAddress.getKey();
+            final String serverprotocol = credRefParentAddress.getValue();
             ModelNode filteredModelNode = model;
-            if (modelFilter != null && modelFilter.length > 0) {
-                for (String path : modelFilter) {
-                    if (filteredModelNode.get(path).isDefined())
-                        filteredModelNode = filteredModelNode.get(path);
-                    else
-                        break;
-                }
+
+            if (filteredModelNode.hasDefined(servertype, serverprotocol)) {
+                filteredModelNode = filteredModelNode.get(servertype, serverprotocol);
+            } else {
+                return;
             }
+
+            String keySuffix = servertype + KEY_DELIMITER + serverprotocol;
             ModelNode value = MailServerDefinition.CREDENTIAL_REFERENCE.resolveModelAttribute(context, filteredModelNode);
-            String keySuffix = modelFilter[0] + KEY_DELIMITER + modelFilter[1];
             if (value.isDefined()) {
                 serverConfig.getCredentialSourceSupplierInjector()
                         .inject(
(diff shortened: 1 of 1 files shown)


mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
```diff
@@ -93,16 +93,17 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
     private static void addCredentialStoreReference(ServerConfig serverConfig, OperationContext context, ModelNode model, ServiceBuilder serviceBuilder, String... modelFilter) throws OperationFailedException {
```
bstansberry, line 93: Unless @yersan disagrees, I think this is a bad method signature. The method requires 2 non-null string params, not an arbitrary length array. All the callers (correctly) pass two strings because that's the model structure this method is meant to process.

Also the 'break' at L102 should be a 'return'. If the model doesn't have the expected structure processing should stop.

boris-unckel: @bstansberry The tests with two Strings and && logic fails now. It seems [MailSessionAdd#L132](https://github.com/wildfly/wildfly/blob/2bfac3db7e000ae152872c2a6ba08c3ae4412eaa/mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java#L132) leads to actual "fallback logic" with the break.

mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
```diff
@@ -90,19 +90,19 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
...
-                for (String path : modelFilter) {
-                    if (filteredModelNode.get(path).isDefined())
-                        filteredModelNode = filteredModelNode.get(path);
-                    else
-                        break;
-                }
+
+            if (filteredModelNode.get(modelFilter1).isDefined()
+                    && filteredModelNode.get(modelFilter2).isDefined()) {
+                filteredModelNode = filteredModelNode.get(modelFilter2);
```
yersan, line 99: Notice `filteredModelNode` is being reassigned on each for loop; it is traversing the model assigned in L95 following a path.

The equivalent `if` to replace the for loop would be:

```
if (filteredModelNode.get(modelFilter1).isDefined()
        && filteredModelNode.get(modelFilter1).get(modelFilter2).isDefined()) {
    filteredModelNode = filteredModelNode.get(modelFilter1).get(modelFilter2);
}
```

mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
```diff
@@ -90,19 +90,19 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
         installRuntimeServices(context, address, fullTree);
     }
 
-    private static void addCredentialStoreReference(ServerConfig serverConfig, OperationContext context, ModelNode model, ServiceBuilder serviceBuilder, String... modelFilter) throws OperationFailedException {
+    private static void addCredentialStoreReference(ServerConfig serverConfig, OperationContext context, ModelNode model, ServiceBuilder serviceBuilder, final String modelFilter1, final String modelFilter2) throws OperationFailedException {
```
yersan, line 93: I agree this method signature could be different; it does not need a variable argument.

From the point of view of the caller, what `modelFilter` should be is an argument to tell the method to where to find the credential-reference model node on the current model, passed by the third argument named `model`. 
Probably a PathAdress object, e.g `final PathAddress credRefParentAddress` is more suitable there instead of `String modelFilter1, final String modelFilter2`

yersan (changes requested):

mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
bstansberry: +1 but it can be a PathElement, which is cheaper.

mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
bstansberry: Two DMR usage notes:

1) node.get(x).isDefined() is the same as node.hasDefined(x) except it doesn't create an undefined node (which is good)

2) get and hasDefined support varargs, and hasDefined will return false as soon as it can.

So

```
if (filteredModelNode.hasDefined(modelFilter1, modelFilter2)) {
    filteredModelNode = filteredModelNode.get(modelFilter1, modelFilter2);
}
```

boris-unckel: @yersan @bstansberry thanks for your feedback. I have changed here according to it.

mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java
boris-unckel: @yersan @bstansberry I can't change easily to a PathElement due to line [MailSessionAdd#L132](https://github.com/wildfly/wildfly/blob/24.0.0.Final/mail/src/main/java/org/jboss/as/mail/extension/MailSessionAdd.java#L132) . I have kept two Strings but renamed them to their content.

yersan: Hi @boris-unckel , PathElement object has a (String,String) static constructor, see https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/PathElement.java#L58

boris-unckel: Hello @yersan thanks for the direct pointer. It's resolved now.

boris-unckel:
Good Morning @yersan @bstansberry The findings should all be fixed now. Could you review and approve, please?

yersan (approved):
LGTM @boris-unckel , thank you!

PR #14517, opened 2021-07-28
Revert "[WFLY-11566] EJB: make a bean class a superclass to a proxy"

tadamski:
I'm reverting the recent commit because of TCK failures. Because the fix was related to support case I'm working on alternative currently.


chengfang (approved):

bstansberry:
I re-ran the Windows JDK 11 job. I don't know why it failed so badly and doubt this PR is related, but might as well retry.

PR #14538, opened 2021-08-09
[WFLY-15095] Test cases to verify and validate the uses of jboss.serv…

yersan:
…er.[base,log,data,temp].dir properties as managed server JVM options

Jira issue: https://issues.redhat.com/browse/WFLY-15095

This requires: https://github.com/wildfly/wildfly-core/pull/4703

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
@@ -3,7 +3,7 @@
 
 Configuration of the JVM settings is different for a managed domain and
 a standalone server. In a managed domain, the domain controller
-components are responsible for starting and stoping server processes and
+components are responsible for starting and stopping server processes and
 hence determine the JVM settings. For a standalone server, it's the
 responsibility of the process that started the server (e.g. passing them
 as command line arguments).
@@ -78,6 +78,33 @@ heap size.
    }
 }
 ----
+
+=== Using filesystem locations as JVM options on domain mode
+The link:Admin_Guide{outfilesuffix}#controlling-filesystem-locations-with-system-properties[Controlling filesystem locations with system properties] section describes the available system properties associated with relevant WildFly file system paths. In addition to all the domain mode properties, the following server specific properties are also available for resolution as JVM options:
+
+* jboss.server.base.dir
+* jboss.server.log.dir
+* jboss.server.data.dir
+* jboss.server.temp.dir
+
+This ability is useful when you need to configure JVM settings without specifying a specific server name. For example, if you want to redirect the GC logging to a file to the default log server directory, you can configure the following JVM option at host level:
+
+[source,options="nowrap"]
+----
+[domain@localhost:9990 /] /host=master/jvm=default:add-jvm-option(jvm-option="-Xlog:gc:file=${jboss.server.log.dir}/gc.log")
+{
+    "outcome" => "success",
+    "result" => undefined,
+    "server-groups" => {"main-server-group" => {"host" => {"master" => {"server-two" => {"response" => {
+        "outcome" => "success",
+        "response-headers" => {
+            "operation-requires-restart" => true,
+            "process-state" => "restart-required"
+        }
+    }}}}}}
+}
+----
+
 
 == Standalone Server
 
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
@@ -1,0 +1,239 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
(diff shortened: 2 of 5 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        return op;
+    }
+
+    public static WebArchive createDeployment() {
+        WebArchive war = ShrinkWrap.create(WebArchive.class, PROP_SERVLET_APP_WAR);
+        war.addClass(PropertiesServlet.class);
+        war.addAsManifestResource(createPermissionsXmlAsset(
+                new PropertyPermission("test.jboss.server.base.dir", "read"),
```
jedla97, line 201: Is there something why not use `org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ`.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
```diff
@@ -0,0 +1,209 @@
...
+                new PropertyPermission("test.jboss.server.log.dir", "read"),
+                new PropertyPermission("test.jboss.server.data.dir", "read"),
+                new PropertyPermission("test.jboss.server.temp.dir", "read")
+        ), "permissions.xml");
+
+        return war;
+    }
+}
```
jedla97, line 209: Please add new line to end of file. At least github show that there is no new line (only in this file).

jedla97 (changes requested):
Hi, I would suggest change change some of the string literal to constant. It will make future maintance easier. I was talking about this with my peer and both of us are in agreedment that when same string literal is used 3+ times it shuld be replaced by constant. For exemple in `JVMServerPropertiesTestCase.java` is `"by-server"` used 6 times. 

Otherwise except some small changes it's look good.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
yersan: Well, `ModelDescriptionConstants` is meant for constants used by the management model, this is just a java property permission action. We cannot use it here.

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
yersan: done

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/JVMServerPropertiesTestCase.java
yersan: Although "read" is used more than 3 times here, I think replacing it with a constant is quite pointless, since the action name is never expected to be changed. If we use a constant will force the reader to look at what such a constant value is, so I think just "read" is a better option here.

@jedla97 Let me know if you think otherwise, thank you for the review

yersan (commented):
@jedla97 I replaced some strings with constants and addressed the missing newline. I've also added a pending commit for the community doumentation. Could you take a look and let me know your opinion? Thanks!

I've created the integration Jobs, but at the moment CI is pretty busy and I cannot paste here yet the Job links. I'll add them soon.

docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
```diff
@@ -79,6 +79,33 @@ heap size.
 }
 ----
 
+=== Using filesystem locations as JVM options on domain mode
+The link:Command_line_parameters.html#controlling-filesystem-locations-with-system-properties[Controlling filesystem locations with system properties] section describes the available system properties associated with relevant WildFly file system paths. In addition to all the domain mode properties, the following server specific properties are also available for resolution as JVM options:
```
jedla97, line 83: Hi, the link redirect me to github 404 page.

jedla97 (commented):
Hi @yersan thanks for fixing it. I looked into the changes and it LGTM. With that "read" string you are right.  Also I look into the community doumentation file and except the link everything look good.

docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
yersan: Good catch @jedla97, I saw the link working fine in my IDE but it was not working in the final documentation generated. I've fixed the link and rebased the branch. Thanks!

yersan:
Hello @jedla97, integration Jobs revealed that we were using a no longer valid configuration for the domain / host configuration files, they were based on the removed security realms see https://issues.redhat.com/browse/WFLY-15001

When I have the Jobs done, I'll paste the links here

yersan:
Rebased to get the wildfly-core recent changes

yersan:
@bstansberry you can now remove the "core-upgrade-needed" label

PR #14534, opened 2021-08-06
WFLY-14923 Update JPA handling to support `initialize-in-order`

scottmarlow:
https://issues.redhat.com/browse/WFLY-14923

PR #14539, opened 2021-08-09
WFLY-14923 Update JPA handling to support initialize-in-order

scottmarlow:
https://issues.redhat.com/browse/WFLY-14923

Replaces https://github.com/wildfly/wildfly/pull/14534

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
@@ -27,10 +27,15 @@ import static org.jboss.as.jpa.messages.JpaLogger.ROOT_LOGGER;
 import java.lang.instrument.ClassFileTransformer;
 import java.lang.instrument.IllegalClassFormatException;
 import java.nio.ByteBuffer;
+import java.security.AccessControlContext;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.security.ProtectionDomain;
 
 import org.jboss.as.jpa.messages.JpaLogger;
 import org.jipijapa.plugin.spi.PersistenceUnitMetadata;
+import org.wildfly.security.manager.WildFlySecurityManager;
+import org.wildfly.security.manager.action.GetAccessControlContextAction;
 
 /**
  * Helps implement PersistenceUnitInfo.addClassTransformer() by using DelegatingClassFileTransformer
@@ -52,30 +57,40 @@ public class JPADelegatingClassFileTransformer implements ClassFileTransformer, 
     @Override
     public ByteBuffer transform(ClassLoader classLoader, String className, ProtectionDomain protectionDomain, ByteBuffer classBytes)
             throws IllegalArgumentException {
-        byte[] transformedBuffer = getBytes(classBytes);
-        boolean transformed = false;
-        for (javax.persistence.spi.ClassTransformer transformer : persistenceUnitMetadata.getTransformers()) {
-            if (ROOT_LOGGER.isTraceEnabled())
-                ROOT_LOGGER.tracef("rewrite entity class '%s' using transformer '%s' for '%s'", className,
-                        transformer.getClass().getName(),
-                        persistenceUnitMetadata.getScopedPersistenceUnitName());
-            byte[] result;
-            try {
-                // parameter classBeingRedefined is always passed as null
-                // because we won't ever be called to transform already loaded classes.
-                result = transformer.transform(classLoader, className, null, protectionDomain, transformedBuffer);
-            } catch (IllegalClassFormatException e) {
-                throw JpaLogger.ROOT_LOGGER.invalidClassFormat(e, className);
-            }
-            if (result != null) {
-                transformedBuffer = result;
-                transformed = true;
-            if (ROOT_LOGGER.isTraceEnabled())
-                ROOT_LOGGER.tracef("entity class '%s' was rewritten", className);
-            }
-        }
+        final AccessControlContext accessControlContext =
+                AccessController.doPrivileged(GetAccessControlContextAction.getInstance());
+        PrivilegedAction<ByteBuffer> privilegedAction =
+                new PrivilegedAction<ByteBuffer>() {
+                    // run as security privileged action
+                    @Override
+                    public ByteBuffer run() {
 
-        return transformed ? ByteBuffer.wrap(transformedBuffer) : null;
+                        byte[] transformedBuffer = getBytes(classBytes);
+                        boolean transformed = false;
+                        for (javax.persistence.spi.ClassTransformer transformer : persistenceUnitMetadata.getTransformers()) {
+                            if (ROOT_LOGGER.isTraceEnabled())
+                                ROOT_LOGGER.tracef("rewrite entity class '%s' using transformer '%s' for '%s'", className,
+                                        transformer.getClass().getName(),
+                                        persistenceUnitMetadata.getScopedPersistenceUnitName());
+                            byte[] result;
+                            try {
+                                // parameter classBeingRedefined is always passed as null
+                                // because we won't ever be called to transform already loaded classes.
+                                result = transformer.transform(classLoader, className, null, protectionDomain, transformedBuffer);
+                            } catch (IllegalClassFormatException e) {
+                                throw JpaLogger.ROOT_LOGGER.invalidClassFormat(e, className);
+                            }
+                            if (result != null) {
+                                transformedBuffer = result;
+                                transformed = true;
+                                if (ROOT_LOGGER.isTraceEnabled())
+                                    ROOT_LOGGER.tracef("entity class '%s' was rewritten", className);
+                            }
+                        }
+                        return transformed ? ByteBuffer.wrap(transformedBuffer) : null;
+                    }
+                };
+        return WildFlySecurityManager.doChecked(privilegedAction, accessControlContext);
     }
 
     private byte[] getBytes(ByteBuffer classBytes) {
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
@@ -55,6 +55,8 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
     // required: name of the persistent unit scoped to deployment file
     private volatile String scopedName;
 
+    private volatile ArrayList<String> containingModuleName;
+
     // optional: jndi name of non-jta datasource
     private volatile String nonJtaDataSourceName;
 
@@ -133,6 +135,16 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
     @Override
     public String getScopedPersistenceUnitName() {
         return scopedName;
+    }
+
+    @Override
+    public void setContainingModuleName(ArrayList<String> containingModuleName) {
+        this.containingModuleName = containingModuleName;
+    }
+
+    @Override
+    public ArrayList<String> getContainingModuleName() {
+        return containingModuleName;
     }
 
     @Override
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2011, Red Hat, Inc., and individual contributors
+ * Copyright 2021, Red Hat, Inc., and individual contributors
  * as indicated by the @author tags. See the copyright.txt file in the
  * distribution for a full listing of individual contributors.
  *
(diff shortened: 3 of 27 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
```diff
@@ -52,30 +57,40 @@ public JPADelegatingClassFileTransformer(PersistenceUnitMetadata pu) {
     @Override
     public ByteBuffer transform(ClassLoader classLoader, String className, ProtectionDomain protectionDomain, ByteBuffer classBytes)
```
scottmarlow, line 58: Transform() method was changed to wrap as security manager checked code.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -55,6 +55,8 @@
     // required: name of the persistent unit scoped to deployment file
     private volatile String scopedName;
 
+    private volatile ArrayList<String> containingModuleName;
```
scottmarlow, line 58: More tracking is added to track which (sub-)module a persistence unit is defined in.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
```diff
@@ -24,8 +24,10 @@
```
scottmarlow, line 24: Change for sub-deployment components to only depend on persistence units from the same sub-deployment when handling `initialize-in-order`.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
```diff
@@ -283,6 +283,8 @@ private void postParseSteps(
             }
 
             pu.setScopedPersistenceUnitName(scopedPersistenceUnitName);
+
+            pu.setContainingModuleName(getContainingModuleName(deploymentUnit));
```
scottmarlow, line 287: Here we make note of the actual deployment module name containing the persistence unit (e.g. ear name + jar/war sub-deployment name).

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -72,7 +80,11 @@ public WeldJpaInjectionServices(DeploymentUnit deploymentUnit) {
         //now we have the service controller, as this method is only called at runtime the service should
         //always be up
         final PersistenceUnitServiceImpl persistenceUnitService = (PersistenceUnitServiceImpl) serviceController.getValue();
-        return new EntityManagerResourceReferenceFactory(scopedPuName, persistenceUnitService.getEntityManagerFactory(), context, deploymentUnit.getAttachment(JpaAttachments.TRANSACTION_SYNCHRONIZATION_REGISTRY), ContextTransactionManager.getInstance());
+        if (persistenceUnitService.getEntityManagerFactory() != null) {
```
scottmarlow, line 83: The enhancement here is to return a LazyEntityManager if we detect that the persistence unit EntityManagerFactory is still null (meaning it hasn't started yet).

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -89,8 +101,12 @@ public WeldJpaInjectionServices(DeploymentUnit deploymentUnit) {
         //now we have the service controller, as this method is only called at runtime the service should
         //always be up
         final PersistenceUnitServiceImpl persistenceUnitService = (PersistenceUnitServiceImpl) serviceController.getValue();
```
scottmarlow, line 103: The enhancement here is to return a LazyEntityManagerFactory if we detect that the persistence unit EntityManagerFactory is still null (meaning it hasn't started yet).

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -128,4 +144,143 @@ public EntityManagerResourceReferenceFactory(String scopedPuName, EntityManagerF
...
+                    }
+            );
+            try {
+                latch.await();
+            } catch (InterruptedException e) {
+                // Thread was interrupted, which we will preserve in case a higher level operation needs to see it.
+                Thread.currentThread().interrupt();
+                // rather than just returning the current EntityManagerFactory (might be null or not null),
+                // fail with a runtime exception.
+                throw new RuntimeException(e);
```
scottmarlow, line 252: Do others agree with setting the thread as interrupted for the higher level to possible detect and then failing the current thread as well by throwing runtime exception?

scottmarlow:
I'm not sure of what caused the Windows JDK11 failure:
- org.eclipse.microprofile.metrics.tck.inheritance.VisibilitySimplyTimedMethodBeanTest.callSimplyTimedMethodsOnce
- org.wildfly.test.integration.vdx.standalone.InvalidCharactersTestCase.missingLastQuoteStandaloneXml 

But otherwise the test results look good.  I also ran the Jakarta EE 8 Full Platform TCKs (including Injection, CDI, BeanValidation) and all TCK tests passed.

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
scottmarlow: Restoring the interrupt state ensures that any higher level check of thread.isInterrupted() will return true.

Since the latch.await() call was interrupted by a different thread for a good reason, it seems that the best outcome is to fail by throwing an exception.

bstansberry: +1. Back in the day I learned that it was best practice to restore the interrupted state.

jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
bstansberry: Is this needed for this bug?

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -128,4 +144,143 @@ public EntityManagerResourceReferenceFactory(String scopedPuName, EntityManagerF
...
+                EntityManagerFactory entityManagerFactory;
+
+                @Override
+                public EntityManagerFactory getInstance() {
+                    PrivilegedAction<Void> privilegedAction =
+                            new PrivilegedAction<Void>() {
+                                // run as security privileged action
+                                @Override
+                                public Void run() {
+                                    entityManagerFactory = persistenceUnitService.getEntityManagerFactory();
```
bstansberry, line 193: Perhaps use a single class LazyFactory<T> and pass a Callable<T> to its constructor. One Callable impl does ^^^, the other does the object instantiation at L266. My impression is these 2 classes are the same except for the way the ultimate object is obtained.

jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
scottmarlow: Yes, we are now seeing later calls to transform application entity classes than we saw without this change.  When we run the new unit test with the Java security manager, we fail without this change.

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
scottmarlow: Switched to use a single class lazy factory https://github.com/wildfly/wildfly/pull/14539/commits/823f482fdeef5df5a67b782a8ac28f84de99c8dd (can be squashed into other change after review).

scottmarlow:
@bstansberry here are some possible ideas for handling of MSC thread calls into org.jboss.as.weld.services.bootstrap.WeldJpaInjectionServices that block on org.jboss.as.weld.services.bootstrap.WeldJpaInjectionServices$LazyFactory:
- Pass a timeout to the `CountDownLatch#await` call so that the lazy factory injection eventually fails (instead of blocking forever).
- Or add some handling to at least log useful information that helps identify this problem occurring without failing on reaching a timeout condition.

scottmarlow:
@bstansberry https://github.com/wildfly/wildfly/pull/14539/commits/06f866821a992d77f545d0f0482b7aa472653bdd does an assertion check that the calling thread isn't a MSC thread.

scottmarlow:
Linux with security manager (JDK8)  failure in RankedAffinityTestCase doesn't seem related to this change.

jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
bstansberry: Can this be moved to the same package as JPAClassFileTransformerProcessor and the class made package-protected instead of public? If this needs to be done with a privileged block we should lock down the ability for unknown code to create one of these.

bstansberry (changes requested):
Looks pretty good but I have a few comments.

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -128,4 +161,86 @@ public EntityManagerResourceReferenceFactory(String scopedPuName, EntityManagerF
...
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        @Override
+        public ResourceReference<T> createResource() {
+            serviceController.addListener(
+                    new LifecycleListener() {
+
+                        @Override
+                        public void handleEvent(final ServiceController<?> controller, final LifecycleEvent event) {
+                            if (event == LifecycleEvent.UP) {
```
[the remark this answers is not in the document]
scottmarlow: @bstansberry for FAILED, we will throw:
    @Message(id = 57, value = "Persistence unit '%s' failed.")
    IllegalStateException persistenceUnitFailed(String scopedPuName);

For REMOVED, we will throw:
    @Message(id = 58, value = "Persistence unit '%s' removed.")
    IllegalStateException persistenceUnitRemoved(String scopedPuName);

The scoped PuName includes the deployment name and persistent unit name which gives some context.  I assume the failure is unexpected but not really sure of a better description to use that describes what REMOVED/FAILED means to users.

scottmarlow:
The Linux - Elytron - JDK 8 failure in FineHotRodSessionExpirationTestCase doesn't look to be related to this change.

jpa/subsystem/src/main/java/org/jboss/as/jpa/classloader/JPADelegatingClassFileTransformer.java
scottmarlow: @bstansberry done

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
```diff
@@ -133,10 +141,56 @@ private void addPersistenceProviderModuleDependencies(DeploymentPhaseContext pha
...
+                        for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
+                            String moduleName = pu.getContainingModuleName().get(pu.getContainingModuleName().size() - 1);
+                            if (moduleName.equals(moduleMetaData.getFileName())) {
+                                ROOT_LOGGER.tracef("Jakarta EE components in %s will depend on persistence unit %s", moduleName, pu.getScopedPersistenceUnitName());
+                                collectPersistenceUnitsForCurrentDeploymentUnit.add(pu);
+                            }
+                        }
+                    }
+                }
+                if (collectPersistenceUnitsForCurrentDeploymentUnit != null && !collectPersistenceUnitsForCurrentDeploymentUnit.isEmpty()) {
```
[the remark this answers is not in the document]
scottmarlow: done

weld/jpa/src/main/java/org/jboss/as/weld/deployment/processor/JpaDependenciesProvider.java
```diff
@@ -22,43 +22,73 @@
...
+                    if (!SubDeploymentMarker.isSubDeployment(resourceRoot)) {
+                        final PersistenceUnitMetadataHolder holder = resourceRoot.getAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS);
+                        for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
+                            String jpaContainerManaged = pu.getProperties().getProperty(Configuration.JPA_CONTAINER_MANAGED);
+                            boolean deployPU = (jpaContainerManaged == null ? true : Boolean.parseBoolean(jpaContainerManaged));
+                            if (deployPU) {
+                                final ServiceName serviceName = PersistenceUnitServiceImpl.getPUServiceName(pu);
+                                dependencies.add(serviceName);
+                            }
+                        }
```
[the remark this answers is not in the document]
scottmarlow: done

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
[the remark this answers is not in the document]
scottmarlow: done

bstansberry (approved):

PR #14543, opened 2021-08-09
[WFLY-14912] Final step for vault removal. 

bstansberry:
Follow up to #14472 to confirm my proposed test fix does as expected.


darranl:
The two failed jobs look unrelated so I have just triggered them again.  We may be able to take this PR instead of mine if this goes green.

bstansberry:
Closing as #14472 now covers this.

PR #14555, opened 2021-08-18
Exclude the ClusteredMessagingTestCase from running in WildFly Previe…

jamezp:
…w. It requires the default JMS server be present in the management model to add a JMS queue.

No JIRA required


jamezp:
EE 9 run https://ci.wildfly.org/buildConfiguration/WF_WildflyPreviewNightly/268547

rhusar (approved):
LGTM!

PR #14556, opened 2021-08-20
[WFLY-15140]: Upgrade WildFly to JBoss Metadata 14.0.0.Final

ehsavoie:
* Using the jakarta native artefacts from JBoss Metadata
* Upgrade Jboss Metadata version to 14.0.0.Final

Jira: https://issues.redhat.com/browse/WFLY-15140

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -90,7 +90,6 @@
         <version.org.hibernate.validator>7.0.1.Final</version.org.hibernate.validator>
         <version.org.jberet>2.0.1.Final</version.org.jberet>
         <version.org.jboss.activemq.artemis.integration>1.0.4</version.org.jboss.activemq.artemis.integration>
-        <version.org.jboss.metadata>14.0.0.Beta2</version.org.jboss.metadata>
         <version.org.jboss.weld.weld>4.0.2.Final</version.org.jboss.weld.weld>
         <version.org.jboss.weld.weld-api>4.0.SP1</version.org.jboss.weld.weld-api>
         <version.org.wildfly.transaction.client>2.0.0.Final</version.org.wildfly.transaction.client>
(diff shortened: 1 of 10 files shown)


ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/metadata/common/main/module.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bstansberry, line 1: I don't think this file is needed as the other one uses an expression.

ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/metadata/ejb/main/module.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bstansberry, line 1: I don't think this file is needed as the other one uses an expression.

ee-9/feature-pack/pom.xml
```diff
@@ -90,7 +90,7 @@
         <version.org.hibernate.validator>7.0.1.Final</version.org.hibernate.validator>
         <version.org.jberet>2.0.1.Final</version.org.jberet>
         <version.org.jboss.activemq.artemis.integration>1.0.4</version.org.jboss.activemq.artemis.integration>
-        <version.org.jboss.metadata>14.0.0.Beta2</version.org.jboss.metadata>
+        <version.org.jboss.metadata>14.0.0.Final</version.org.jboss.metadata>
```
bstansberry, line 93: How about dropping this property, changing the version in the root pom, and then in this pom remove any jb-metadata dependency info except for the two -jakarta artifacts? In the long run I don't see why regular WF, even an EE 8 based variant, needs to stay on JBM 13.

bstansberry (changes requested):

ee-9/feature-pack/pom.xml
ehsavoie: ok

ee-9/feature-pack/pom.xml
```diff
@@ -1060,24 +1059,11 @@
         <dependency>
```
bstansberry, line 1059: I don't think this is needed any more.

ee-9/feature-pack/pom.xml
```diff
@@ -1090,24 +1076,12 @@
...
```
bstansberry, line 1076: I don't think this is needed any more.

ee-9/feature-pack/pom.xml
```diff
@@ -1120,18 +1094,7 @@
...
```
bstansberry, line 1094: I don't think this is needed any more.

ee-feature-pack/common/pom.xml
```diff
@@ -1009,6 +1009,17 @@
             </exclusions>
         </dependency>
 
+        <dependency>
```
bstansberry, line 1012: Did something fail without this? This module doesn't build anything so should not need transitive deps of anything it declares. Code that consumes this, e.g.https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-feature-pack/pom.xml#L396 is getting this dependency info imported from elsewhere, e.g. https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-feature-pack/pom.xml#L364

bstansberry (changes requested):

bstansberry (approved):

PR #14562, opened 2021-08-20
[WFLY-11566] EJB: make a bean class a superclass to a proxy

tadamski:
This PR restores setting proxy superclass to component class for EJB views. This setting was done as a part of https://issues.redhat.com/browse/WFLY-11566 but have to be reverted because of failing TCK tests. With additional fixes below the TCK 9.1 testsuite passes on my laptop. EE-8 testsuite interop tests (which were removed from 9.1+) pass without csiv2 tests https://wildflycts-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jakartaee-ee8/job/runner/135713/console but I have checked that those tests fail with wildfly:main as well when run with the same method.

@chengfang  could you please review?

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/component/session/SessionBeanObjectViewConfigurator.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/component/session/SessionBeanObjectViewConfigurator.java
@@ -115,7 +115,7 @@ public abstract class SessionBeanObjectViewConfigurator implements ViewConfigura
 
                 if (componentMethod != null) {
                     if(!Modifier.isPublic(componentMethod.getModifiers())) {
-                        throw EjbLogger.ROOT_LOGGER.ejbBusinessMethodMustBePublic(componentMethod);
+                        EjbLogger.ROOT_LOGGER.ejbBusinessMethodMustBePublic(componentMethod);
                     }
 
                     configuration.addViewInterceptor(method, new ImmediateInterceptorFactory(new ComponentDispatcherInterceptor(componentMethod)), InterceptorOrder.View.COMPONENT_DISPATCHER);
--- a/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
@@ -24,6 +24,7 @@
 
 package org.jboss.as.ejb3.logging;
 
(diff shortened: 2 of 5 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/component/session/SessionBeanObjectViewConfigurator.java
```diff
@@ -115,7 +115,7 @@ public void configureDependency(final ServiceBuilder<?> serviceBuilder, final Co
 
                 if (componentMethod != null) {
                     if(!Modifier.isPublic(componentMethod.getModifiers())) {
-                        throw EjbLogger.ROOT_LOGGER.ejbBusinessMethodMustBePublic(componentMethod);
+                        EjbLogger.ROOT_LOGGER.ejbBusinessMethodMustBePublic(componentMethod);
                     }
```
chengfang, line 119: We may want to log it in debug level so users won't be alarmed by this message.

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -408,6 +408,10 @@
     @Message(id = 37, value = "Dynamic stub creation failed for class %s")
     void dynamicStubCreationFailed(String clazz, @Cause Throwable t);
 
+    @LogMessage(level = INFO)
+    @Message(id = 38, value = "Jakarta Enterprise Beans business method %s must be public")
+    void ejbBusinessMethodMustBePublic(final Method method);
+
```
chengfang, line 414: Not sure if it's good to reuse the log id `38`. But if we decide to log it in debug level, we can just directly log it in java class without this entry, so this won't be a problem.

ejb3/src/main/java/org/jboss/as/ejb3/component/session/SessionBeanObjectViewConfigurator.java
tadamski: Fixed

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
tadamski: Fixed. Moved at the end with the new number.

chengfang (approved):

PR #14566, opened 2021-08-21
[WFLY-15044] Avoid Nullpointer (IIOP)

boris-unckel:

### DIFF
--- a/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/rmi/ValueAnalysis.java
+++ b/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/rmi/ValueAnalysis.java
@@ -174,7 +174,7 @@ public class ValueAnalysis extends ContainerAnalysis {
                 wo = null; // Wrong return type
             }
             if (wo != null) { // Right modifiers?
-                int mods = spf.getModifiers();
+                int mods = wo.getModifiers();
                 if (!Modifier.isPrivate(mods))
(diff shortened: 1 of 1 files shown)


iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/rmi/ValueAnalysis.java
```diff
@@ -173,7 +173,7 @@ protected void doAnalyze() throws RMIIIOPViolationException {
                     && wo.getReturnType() != Void.TYPE) {
                 wo = null; // Wrong return type
             }
-            if (wo != null) { // Right modifiers?
+            if (wo != null && spf != null) { // Right modifiers?
```
bstansberry, line 176: @tadamski Please review.

I suspect this isn't the right fix here and that the problem is that L177 should be

int mods = wo.getModifiers();

I don't see why the modifiers of the 'serialPersistentFields' field would be relevant at this point.

tadamski (changes requested):
@bstansberry I agree, spf seems irrelevant in this place and it looks like a bug.
@boris-unckel I believe this PR should be modified according to Brian's suggestion.

boris-unckel:
@tadamski Fixed accordingly. Could you check again and approve, please?

tadamski (approved):

tadamski:
@boris-unckel Approved.

PR #14567, opened 2021-08-21
[WFLY-15044] Avoid Nullpointer (EJB)

boris-unckel:


boris-unckel:
@chengfang Hello Cheng, could you review/approve, please?

ejb3/src/main/java/org/jboss/as/ejb3/remote/LocalEjbReceiver.java
```diff
@@ -310,8 +311,8 @@ public void discardResult() {
         private final boolean allowPassByReference;
 
         CloningExceptionProducer(final ObjectCloner resultCloner, final Exception exception, final boolean allowPassByReference) {
-            this.resultCloner = resultCloner;
-            this.exception = exception;
+            this.resultCloner = checkNotNullParam("resultCloner", resultCloner);
+            this.exception = checkNotNullParam("exception", exception);
             this.allowPassByReference = allowPassByReference;
```
chengfang, line 316: I don't think we need null check here. `resultCloner` is already assigned some value in all callers, and for `exception` it won't be null since it is called from within a catch block; and even if `exception` is null, it will be cloned to null, which is valid.

chengfang (approved):

PR #14568, opened 2021-08-21
[WFLY-15044] Avoid Nullpointer (EE)

boris-unckel:

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/ManagedReferenceMethodInjectionInterceptorFactory.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/ManagedReferenceMethodInjectionInterceptorFactory.java
@@ -99,6 +99,10 @@ final class ManagedReferenceMethodInjectionInterceptorFactory implements Interce
             if (reference == null && optional) {
                 return context.proceed();
             }
+            if (reference == null) {
+                throw EeLogger.ROOT_LOGGER.managedReferenceMethodWasNull(method);
+            }
+
             boolean ok = false;
             try {
                 componentInstance.setInstanceData(valueKey, reference);
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
@@ -276,6 +276,9 @@ public class ModuleJndiBindingProcessor implements DeploymentUnitProcessor {
                     if (registered == null)
                         throw e;
                     BinderService service = (BinderService) registered.getService();
+                    if (service == null || service.getSource() == null) {
+                        throw e;
+                    }
                     if (!service.getSource().equals(bindingConfiguration.getSource()))
                         throw EeLogger.ROOT_LOGGER.conflictingBinding(bindingName, bindingConfiguration.getSource());
                 } catch (CircularDependencyException e) {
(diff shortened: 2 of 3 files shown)


ee/src/main/java/org/jboss/as/ee/component/ManagedReferenceMethodInjectionInterceptorFactory.java
```diff
@@ -105,7 +105,7 @@ public Object processInvocation(final InterceptorContext context) throws Excepti
                 final InvocationType invocationType = context.getPrivateData(InvocationType.class);
                 try {
                     context.putPrivateData(InvocationType.class, InvocationType.DEPENDENCY_INJECTION);
-                    method.invoke(target, reference.getInstance());
+                    method.invoke(target, reference != null ? reference.getInstance() : null);
```
bstansberry, line 108: This and the other change in this class are questionable.  There's an 'optional' field and logic at L99 to deal with cases where this is null and optional==true. And then later on the code will fail if this is null and optional==false.

It's not clear the thing to do here is to no longer fail. More intuitive would be to fail earlier on in a clearer way than an NPE.

boris-unckel: @bstansberry I have found a existing logic/exception for "field" [https://github.com/wildfly/wildfly/blob/main/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java#L1043](https://github.com/wildfly/wildfly/blob/main/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java#L1043). 
Overloading EeLogger

```java
    @Message(id = 131, value = "ManagedReference was null and injection is not optional for injection into method %s")
    RuntimeException managedReferenceWasNull(Method method);
```
Adding after L101:

```java
            if (reference == null) {
                throw EeLogger.ROOT_LOGGER.managedReferenceWasNull(method);
            }
```

Does this solution address your concerns?

ee/src/main/java/org/jboss/as/ee/component/ManagedReferenceMethodInjectionInterceptorFactory.java
```diff
@@ -99,6 +99,10 @@ public Object processInvocation(final InterceptorContext context) throws Excepti
             if (reference == null && optional) {
                 return context.proceed();
             }
+            if (reference == null) {
+                throw EeLogger.ROOT_LOGGER.managedReferenceMethodWasNull(method);
+            }
```
bstansberry, line 104: +1

bstansberry (changes requested):
Thanks @boris-unckel. Change re my last comment looks good. Apologies; I wrote another comment earlier but it seems I never submitted it. I'm doing so now.

ee/src/main/java/org/jboss/as/ee/component/deployers/ResourceReferenceProcessor.java
```diff
@@ -272,6 +272,8 @@ public ManagedReference getReference() {
             final BindingConfiguration bindingConfiguration;
             if (!isEmpty(lookup)) {
                 bindingConfiguration = new BindingConfiguration(name, new LookupInjectionSource(lookup));
+            } else if (isEmpty(value)) {
+                throw EeLogger.ROOT_LOGGER.cannotLoad(value);
```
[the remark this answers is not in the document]
boris-unckel: I think it is detected because of the && condition.:L254=>_both_ fields with value; L256 _both_ fields null.
I'll try the assert and come back - interesting question in general. I can document this with "Won't fix" - any solution should help with error prevention or ease analysis. Some of my managers would do it to decrease the total number of findings (regardless of the sense) - I don't like "Key Performance Indicator" work :-)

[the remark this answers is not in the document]
boris-unckel: @bstansberry `assert !isEmpty(lookup) || value != null;` is not recognized by sonarcloud. I'll remove ResourceReferenceProcessor from this PR and mark it as won't fix.

bstansberry:
Thanks, @boris-unckel

PR #14569, opened 2021-08-21
[WFLY-15044] Avoid Nullpointer (Infinispan)

boris-unckel:

### DIFF
--- a/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
+++ b/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
@@ -1481,6 +1481,9 @@ public class InfinispanSubsystemXMLReader implements XMLElementReader<List<Model
                     break;
                 }
                 default: {
+                    if (address == null) {
+                        throw ParseUtils.missingOneOf(reader, EnumSet.of(XMLElement.ENTRY_TABLE, XMLElement.BUCKET_TABLE));
+                    }
                     this.parseStoreElement(reader, address, operations);
                 }
             }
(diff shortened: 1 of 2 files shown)


clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
```diff
@@ -1481,6 +1481,7 @@ private void parseLegacyJDBCStore(XMLExtendedStreamReader reader, PathAddress ca
                     break;
                 }
                 default: {
+                    address = address != null ? address : cacheAddress;
```
pferraro, line 1484: This is not correct.  If address is null here, then we'd actually want to throw an exception, since either entry-table, bucket-table, or both; must be defined previously (which determines the target address).

pferraro (changes requested):

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemXMLReader.java
boris-unckel: @pferraro Thanks for your feedback. I would suggest [o.w.common.Assert.impossibleSwitchCase](https://github.com/wildfly/wildfly-common/blob/master/src/main/java/org/wildfly/common/Assert.java#L552) or do you prefer another (new) one?

pferraro: In this case, a null address is unexpected, but not "impossible" - and indicates that the xml violates the schema.  However, since we do not formally validate against our schema before parsing, this is a condition that we should handle during parsing. I suggest using:
```java
if (address == null) {
    throw ParseUtils.missingOneOf(reader, EnumSet.of(XMLElement.ENTRY_TABLE, XMLElement.BUCKET_TABLE));
}
```

pferraro (approved):

PR #14570, opened 2021-08-22
[WFLY-15044] Avoid Nullpointer (Connector)

boris-unckel:


tadamski (changes requested):
@boris-unckel The null pointer fix seems OK, but what is the rationale for the logging fix? Are those names different?

boris-unckel:
> The null pointer fix seems OK, but what is the rationale for the logging fix? Are those names different?

@tadamski You can see the reason in Line 112 [https://github.com/wildfly/wildfly/pull/14570/files#diff-5a33fcd55dc7be71481cd1158ae9a91bc14bdeee84481786b2e0e78ef2bb34b1R112](https://github.com/wildfly/wildfly/pull/14570/files#diff-5a33fcd55dc7be71481cd1158ae9a91bc14bdeee84481786b2e0e78ef2bb34b1R112) connectorXmlDescriptor may be null.

tadamski:
@boris-unckel  Ahh, OK. Approving.

tadamski (approved):

PR #14572, opened 2021-08-22
[WFLY-14871] Logic in try/finally blocks hides exceptions (EJB)

boris-unckel:

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/component/interceptors/AdditionalSetupInterceptor.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/component/interceptors/AdditionalSetupInterceptor.java
@@ -47,25 +47,45 @@ public class AdditionalSetupInterceptor implements Interceptor {
 
     @Override
     public Object processInvocation(final InterceptorContext context) throws Exception {
+        Throwable originalException = null;
+        Object retValue = null;
         try {
             for (int i = 0; i < actions.length; ++i) {
                 actions[i].setup(Collections.<String, Object>emptyMap());
             }
-            return context.proceed();
+            retValue = context.proceed();
+        } catch(Throwable ex) {
+            if (ex instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            originalException = ex;
         } finally {
-            Throwable error = null;
+            Throwable suppressed = originalException;
             for (int i = actions.length - 1; i >=0; --i) {
                 SetupAction action = actions[i];
                 try {
                     action.teardown(Collections.<String, Object>emptyMap());
-                } catch (Throwable e) {
-                    error = e;
+                } catch (Throwable t) {
+                    if (suppressed != null) {
+                        suppressed.addSuppressed(t);
+                    } else {
+                        suppressed = t instanceof RuntimeException ? t : new RuntimeException(t);
+                    }
                 }
             }
-            if (error != null) {
-                throw new RuntimeException(error);
+            if (suppressed != null) {
+                if (suppressed instanceof RuntimeException) {
+                    throw (RuntimeException) suppressed;
+                }
+                if (suppressed instanceof Exception) {
+                    throw (Exception) suppressed;
+                }
+                if (suppressed instanceof Error) {
+                    throw (Error) suppressed;
+                }
             }
         }
+        return retValue;
     }
 
     public static InterceptorFactory factory(final List<SetupAction> actions) {


boris-unckel:
@chengfang another one, could you review/approve, please?

ejb3/src/main/java/org/jboss/as/ejb3/component/interceptors/AdditionalSetupInterceptor.java
```diff
@@ -47,25 +47,46 @@ public AdditionalSetupInterceptor(final List<SetupAction> actions) {
...
+                if (suppressed instanceof RuntimeException) {
+                    throw (RuntimeException) suppressed;
+                }
+                if (suppressed instanceof Error) {
+                    throw (Error) suppressed;
+                }
+                if (suppressed instanceof Exception) {
+                    throw (Exception) suppressed;
+                }
+                throw new RuntimeException(suppressed);
```
chengfang, line 86: After the 3 `if`, `throw new RuntimeException()` seems unreachable.

The old version wraps any throwable as RuntimeException and throw it. To be roughly equivalent, I think we only need to wrap `Error` to RuntimeException, and everything else just throw it as is.

boris-unckel: @chengfang thanks for your feedback.

> After the 3 if, throw new RuntimeException() seems unreachable.

I'll remove it.

> The old version wraps any throwable as RuntimeException and throw it. 

The old version wraps any throwable out of the _finally block_ as RuntimeException. In case of a Throwable out of the try block and no Throwable out of the finally block, this try block one would have been thrown without wrapping.
I'd suggest to add this in line 72 `suppressed = t;` as `suppressed = new RuntimeException(t);`

chengfang: How about this, from line 76 - 87:
```
            if (suppressed != null) {
                if (suppressed instanceof RuntimeException) {
                    throw suppressed;
                }
                throw new RuntimeException(suppressed);
            }
```

Whether it's from try block or finally block, if it's a RuntimeException, we throw it as is; else we wrap it as RuntimeException and throw it. This will match the old behavior.

boris-unckel: @chengfang I have changed it that way, please have a look

chengfang (approved):

ejb3/src/main/java/org/jboss/as/ejb3/component/interceptors/AdditionalSetupInterceptor.java
boris-unckel: @chengfang We have been to fast, tests failing now. The method signature has `throws Exception` in it. Exceptions from the try block must not be thrown as RuntimeException but in a native way. I'll change to `suppressed = new RuntimeException(t);` version and keep the separated Error/Exception/Runtime throws.

boris-unckel: @chengfang 
Local tests pass with this version:

```    public Object processInvocation(final InterceptorContext context) throws Exception {
        Throwable originalException = null;
        Object retValue = null;
        try {
            for (int i = 0; i < actions.length; ++i) {
                actions[i].setup(Collections.<String, Object>emptyMap());
            }
            retValue = context.proceed();
        } catch(Throwable ex) {
            if (ex instanceof InterruptedException) {
                Thread.currentThread().interrupt();
            }
            originalException = ex;
        } finally {
            Throwable suppressed = originalException;
            for (int i = actions.length - 1; i >=0; --i) {
                SetupAction action = actions[i];
                try {
                    action.teardown(Collections.<String, Object>emptyMap());
                } catch (Throwable t) {
                    if (suppressed != null) {
                        suppressed.addSuppressed(t);
                    } else {
                        suppressed = t instanceof RuntimeException ? t : new RuntimeException(t);
                    }
                }
            }
            if (suppressed != null) {
                if (suppressed instanceof Exception) {
                    throw (Exception) suppressed;
                }
                if (suppressed instanceof RuntimeException) {
                    throw (RuntimeException) suppressed;
                }
                if (suppressed instanceof Error) {
                    throw (Error) suppressed;
                }
            }
        }
        return retValue;
    }
```

`context.proceed();` can throw a Exception from it's signature, this is the origin in the try block (not wrapped before). It worked because  `processInvocation(final InterceptorContext context) throws Exception` allowed it.

What do you think?

chengfang: Can you put `(suppressed instanceof RuntimeException)` before `(suppressed instanceof Exception)`?

chengfang: The above change looks good to me.

boris-unckel:
@chengfang I have pushed the version from our discussion. Three single unit tests failed (all different), I don't think they are related. Please check again.

chengfang:
I agree these 3 failures are unrelated.

`ReplicatedFailoverTestCase.testBackupFailoverAfterFailback`: [WFLY-14481](https://issues.redhat.com/browse/WFLY-14481) Intermittent failures of ReplicatedFailoverTestCase

`SecurityAuditingTestCase.test` & `SecurityAuditingTestCase.testSingleMethodAnnotationsUser1Template`: [WFLY-11134](https://issues.redhat.com/browse/WFLY-11134) SecurityAuditingTestCase fails intermittently

boris-unckel:
@bstansberry this PR is approved and tests passed / are checked. Please check and merge.

PR #14573, opened 2021-08-22
[WFLY-14871] Logic in try/finally blocks hides exceptions (IIOP)

boris-unckel:

### DIFF
--- a/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/naming/jndi/CNCtx.java
+++ b/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/naming/jndi/CNCtx.java
@@ -398,6 +398,8 @@ public class CNCtx implements javax.naming.Context {
             return url;
         } else {
             InputStream in = null;
+            Throwable originalException = null;
+            String retValue = null;
             try {
                 URL u = new URL(url);
                 in = u.openStream();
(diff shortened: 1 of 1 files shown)


bstansberry:
@tadamski Please review

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/naming/jndi/CNCtx.java
```diff
@@ -406,27 +408,52 @@ private String getStringifiedIor(String url) throws NamingException {
...
+                            suppressed = ne;
+                        }
                     }
-                } catch (IOException e) {
-                    NamingException ne = IIOPLogger.ROOT_LOGGER.invalidURLOrIOR(url);
-                    ne.setRootCause(e);
-                    throw ne;
                 }
+                if (suppressed != null) {
+                    if (suppressed instanceof RuntimeException) {
```
tadamski, line 441: Maybe we could simplify it a bit and just do or and rethrow. Why do we need the cast?

tadamski (commented):
The logic seems good to me. I'm wondering whether the code should be simplified a little bit as described in a comment above.

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/naming/jndi/CNCtx.java
boris-unckel: @tadamski in general this "pattern" is the way I solved all similar cases. In new Line 420 the Throwable may contain the RuntimeException or Error. You can't rethrow a undeclared Throwable, therefor the cast.

boris-unckel: @tadamski Christian Beikov had the same concern at the same point. He refactored it this way: [https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/internal/util/ExceptionHelper.java](https://github.com/hibernate/hibernate-orm/blob/wip/6.0/hibernate-core/src/main/java/org/hibernate/internal/util/ExceptionHelper.java) It works, but it completely undermines the typesafety of throw / Throwables. I don't want to provide it without broader review/acceptance.

tadamski (approved):

PR #14574, opened 2021-08-22
[WFLY-14871] Logic in try/finally blocks hides exceptions (Messaging-AMQ)

boris-unckel:


boris-unckel:
@ehsavoie 

> I think that the 'best' fix is to avoid throwing a StartException in the finally and just trace the exception there.

Thanks for your feedback. I have lost the history of the review, resetting git head hard, sorry. Please check the fixed version, tests passed.

boris-unckel:
Hello @ehsavoie my above comment sounds like your previous work is lost but the opposite is the case - the code is updated accordingly to your review comment. Could you review and approve, please?

ehsavoie (approved):

PR #14578, opened 2021-08-25
[WFLY-14350] Allows DEFAULT_CONTEXT_PROPAGATION to use expressions and sets default value

xjusko:
Fixes https://issues.redhat.com/browse/WFLY-14350

### DIFF
--- a/xts/src/main/java/org/jboss/as/xts/XTSSubsystemAdd.java
+++ b/xts/src/main/java/org/jboss/as/xts/XTSSubsystemAdd.java
@@ -220,7 +220,7 @@ class XTSSubsystemAdd extends AbstractBoottimeAddStepHandler {
             XtsAsLogger.ROOT_LOGGER.debugf("nodeIdentifier=%s%n", coordinatorURL);
         }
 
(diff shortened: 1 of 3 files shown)


darranl:
@bstansberry should changing the use of expressions trigger a version bump?  It feels like we should detect in transformers.

darranl (changes requested):
This will need a model version bump and also transformer support as older secondary host controllers would not recognise the expressions on these resources.

xjusko:
So if I understand this correctly, should I create a new xsd schema version which differs from 3.0 only in allowing expression in DEFAULT_CONTEXT_PROPAGATION?
If yes, then the new xsd schema would be the same as 3.0 and the parser should be able to parse both definitions of DEFAULT_CONTEXT_PROPAGATION based on the version of xsd?

bstansberry:
@tomjenkinson The XTS subsystem was deprecated in Nov 2020 via WFLY-14093. Do we want a change like this? If not let's close the PR and resolve the JIRA.

If we want it Marek would need to bump the management API version and add a new xsd version, plus add a domain transformer. The basic effect of all that would be to properly document in the read-resource-description output and in the xsd that the default value is false. There's no change in how the actual services work as the code was already using false as a default

PR #14581, opened 2021-08-26
[WFLY-15174] Use commons-lang3 only, remove commons-lang(2) uses

langfr:
Jira https://issues.redhat.com/browse/WFLY-15174


testsuite/integration/basic/pom.xml
```diff
@@ -884,7 +884,6 @@
                                 <include>org/wildfly/test/integration/microprofile/**/*Test.java</include>
                             </includes>
                             <excludes>
-                                <!-- requires org.apache.commons.lang in test servlet otherwise passes -->
```
bstansberry, line 887: This should be restored and just changed to reference lang3.

If we keep the exclude element we need the comment explaining why.

testsuite/integration/basic/pom.xml
```diff
@@ -2697,7 +2696,6 @@
                                         <include>org/jboss/as/test/integration/weld/builtin/*TestCase.java</include>
                                     </includes>
                                     <excludes>
-                                        <!-- requires org.apache.commons.lang in test servlet otherwise passes -->
```
bstansberry, line 2700: This should be restored and just changed to reference lang3.

testsuite/integration/basic/pom.xml
```diff
@@ -3332,7 +3330,6 @@
...
```
bstansberry, line 3335: This should be restored and just changed to reference lang3.

bstansberry (changes requested):
Minor point about some pom comments aside, looks good.

bstansberry:
Thanks @langfr

PR #14584, opened 2021-08-27
[WFLY-15001] Second stage of security realm removal from WildFly.

darranl:
https://issues.redhat.com/browse/WFLY-14941
https://issues.redhat.com/browse/WFLY-15001

Depends on https://github.com/wildfly/wildfly-core/pull/4716, however that PR is also dependent on this being merged,


darranl:
New set of experiments:

WFCORE-5526 vs WFLY-15001_II

-DallTests - 
-Dts.bootable - 
-Dts.layers - 
-DallTests -Dsecurity.manager=true -

darranl:
Just submitting some further changes, the affected jobs were "Galleon Linux - JDK 11" and "Linux with security manager - JDK 8"

pom.xml
```diff
@@ -426,6 +426,7 @@
         <version.org.jboss.mod_cluster>1.4.3.Final</version.org.jboss.mod_cluster>
         <version.org.jboss.narayana>5.12.0.Final</version.org.jboss.narayana>
         <version.org.jboss.openjdk-orb>8.1.5.Final</version.org.jboss.openjdk-orb>
+        <version.org.picketbox>5.0.3.Final-redhat-00007</version.org.picketbox>
```
rgdoliveira, line 432: Hi @darranl , I was taking a look in this change and seems that wildfly pom already contained the org.picketbox property.

So now it is duplicated, see both https://github.com/wildfly/wildfly/blob/main/pom.xml#L432 and https://github.com/wildfly/wildfly/blob/main/pom.xml#L478

darranl: We will have a further look in our next round of removals, ideally all PicketBox references get to be removed but we will see.

PR #14590, opened 2021-08-31
[WFCORE-5406][WFLY-14793] Adding required --add-opens & --add-exports clauses

ropalka:
to be able to run on both NOT MODULAR and MODULAR JDK types. The analysis why
these modular jdk params are needed is available at WFCORE-5406 JIRA (round 2).

https://issues.redhat.com/browse/WFLY-14793


ropalka:
@bstansberry please review

bstansberry (approved):

PR #14595, opened 2021-09-01
[WFLY-14980] add default 60s r/o timeout to listeners

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-14980

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
@@ -38,6 +38,8 @@ import static org.wildfly.extension.undertow.HttpListenerResourceDefinition.PROX
 import static org.wildfly.extension.undertow.HttpListenerResourceDefinition.REQUIRE_HOST_HTTP11;
 import static org.wildfly.extension.undertow.HttpsListenerResourceDefinition.SSL_CONTEXT;
 import static org.wildfly.extension.undertow.ListenerResourceDefinition.ALLOW_UNESCAPED_CHARACTERS_IN_URL;
+import static org.wildfly.extension.undertow.ListenerResourceDefinition.READ_TIMEOUT;
+import static org.wildfly.extension.undertow.ListenerResourceDefinition.WRITE_TIMEOUT;
 import static org.wildfly.extension.undertow.ListenerResourceDefinition.RFC6265_COOKIE_VALIDATION;
 import static org.wildfly.extension.undertow.ServletContainerDefinition.DEFAULT_COOKIE_VERSION;
 import static org.wildfly.extension.undertow.ServletContainerDefinition.DISABLE_FILE_WATCH_SERVICE;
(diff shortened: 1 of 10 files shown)


bstansberry:
@fl4via Please review

This is an API change so needs a model version bump unless one's already been done for the release it goes into.

baranowb:
@bstansberry  @fl4via @ropalka  - there is no dedicated section of docs( admin/started guides) with description of params, those seem to be part of model ref. So this does not need specific doc update?

bstansberry:
@baranowb Right; this kind of simple management API change is normally just documented via the model ref, unless there's other existing documentation that needs an update.

fl4via:
> This is wrong: https://github.com/wildfly/wildfly/blob/main/undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtension.java#L82

@baranowb yes, it is in fact. When submitting a PR for upgrading the model version, please make sure that this line is fixed as well. Good catch!

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
```diff
@@ -125,6 +127,29 @@ private static void registerTransformersWildFly18(ResourceTransformationDescript
                 .getAttributeBuilder()
                     .setValueConverter(AttributeConverter.DEFAULT_VALUE, CONNECTION_IDLE_TIMEOUT)
                 .end();
+
+        final ResourceTransformationDescriptionBuilder serverBuilder = subsystemBuilder.addChildResource(UndertowExtension.SERVER_PATH);
+
+        final AttributeTransformationDescriptionBuilder http = serverBuilder.addChildResource(UndertowExtension.HTTP_LISTENER_PATH).getAttributeBuilder()
+                .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, WRITE_TIMEOUT)
```
fl4via, line 134: I need to check this with @bstansberry  or someone else that has more knowledge on transformers than I do.
I think that, because we are just adding a default value to an already existing attribute, we don't add any rejects here. I think we just need to:
- either add the default value when transforming to older versions, so that the absence of a read/write-timeout in current WF model means a read/write-timeout value set to 60 in previous versions
- or we just omit the value if it is undefined (meaning it will have the default value of 60) when transforming to older versions, so that the transformed model is clean of that
I would bet on the first option, but I could be wrong. As I said, we need some feedback from transformers pros :-)

fl4via (changes requested):
As @darranl  said in the other PR, let's create a separate Jira for the model upgrade, and submit a separate PR. The other two Jiras will be linked to the model upgrade one.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
bstansberry: This is a quick, perhaps off-topic comment before I go into meetings.

https://docs.wildfly.org/25/Extending_WildFly.html#common-transformation-use-cases covers some of the cases; perhaps this is 4.8.3?

The basic requirement is that servers running on the legacy version need to end up with the same runtime behavior as those running the current version, or the transformer logic should result in the op setting the config to whatever the legacy server can't handle being rejected.

pferraro: The use of DiscardAttributeChecker.DEFAULT_VALUE here looks correct to me.

ropalka:
This PR can be closed, it is superseded by https://github.com/wildfly/wildfly/pull/15069

baranowb:
REplaced by: https://github.com/wildfly/wildfly/pull/15069

PR #14602, opened 2021-09-02
[WFLY-15194] Update jsoup version

jasondlee:
https://issues.redhat.com/browse/WFLY-15194


jasondlee:
@bstansberry Guillaume has confirmed on the HV side that this change is good to commit.

bstansberry (approved):

PR #14607, opened 2021-09-03
WFLY-15192 Reduce the usage of TimeHandle in ejb timer service; and g…

chengfang:
…eneral code cleanup

https://issues.redhat.com/browse/WFLY-15192

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerImpl.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerImpl.java
@@ -24,7 +24,6 @@ package org.jboss.as.ejb3.timerservice;
 import java.io.Serializable;
 import java.util.Date;
 import java.util.concurrent.Semaphore;
-
 import javax.ejb.EJBException;
 import javax.ejb.ScheduleExpression;
 import javax.ejb.Timer;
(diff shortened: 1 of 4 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerImpl.java
```diff
@@ -555,31 +530,22 @@ public void unlock() {
...
-            return false;
-        }
-        if (obj instanceof TimerImpl == false) {
-            return false;
-        }
-        TimerImpl otherTimer = (TimerImpl) obj;
-        return this.handle.equals(otherTimer.getTimerHandle());
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (this.getClass() != o.getClass()) return false;
```
boris-unckel, line 536: The class is not final and you're switching from instanceof to getClass. This might cause trouble for derived classes.

chengfang: this is to make sure `this` and `other` are of the exactly the same type. We don't want to compare an instance of super type against a derived type. With `instanceof`, if `this` is `TimerImpl` and `other` is `SubTimerImpl`, `other instanceof TimerImpl` would eval to true and continue, which is not what we want.

boris-unckel: thanks for the explanation

bstansberry (approved):
This LGTM. The only use of the removed TimerImpl.getTimerHandle() method was in CalendarTimer's unneeded override of TimerImpl.equals. The new TimerService.equals implwill produce the same results as what the old one did without requiring a handle.

PR #14613, opened 2021-09-04
[WFLY-15208] Compare Objects with correct type or method (jaxrs)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15208

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/deployment/JaxrsIntegrationProcessor.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/deployment/JaxrsIntegrationProcessor.java
@@ -559,7 +559,7 @@ public class JaxrsIntegrationProcessor implements DeploymentUnitProcessor {
         if ("".equals(value.trim())) {
             return false;
         }
-        return !value.equals(attribute.getDefaultValue());
+        return !value.equals(attribute.getDefaultValue() != null ? attribute.getDefaultValue().asString() : null);
     }
 
     /**


boris-unckel:
@jamezp Could you trigger a retest, please? And have a look at it for review?

jaxrs/src/main/java/org/jboss/as/jaxrs/deployment/JaxrsIntegrationProcessor.java
```diff
@@ -559,7 +559,7 @@ private boolean isTransmittable(AttributeDefinition attribute, ModelNode modelNo
         if ("".equals(value.trim())) {
             return false;
         }
-        return !value.equals(attribute.getDefaultValue());
+        return !value.equals(attribute.getDefaultValue() != null ? attribute.getDefaultValue().asString() : null);
```
jamezp, line 562: This could just be:
```
return attribute.getDefaultValue() != null && !value.equals(attribute.getDefaultValue().asString())
```

That's a bit of a nit though I guess :)

boris-unckel: Yes, after reading the smarter solution I wonder why I did the first one. Still a long way for me to go :-)

boris-unckel: @jamezp I wanted provide the fix but let's check twice:

```
Scenario "attribute.getDefaultValue() returns null"

(1) Current fix of the PR
!value.equals(attribute.getDefaultValue() != null ? attribute.getDefaultValue().asString() : null);
!value.equals(null != null ? attribute.getDefaultValue().asString() : null);
!value.equals(null);
!false;
return true;

(2) Short edition
attribute.getDefaultValue() != null && !value.equals(attribute.getDefaultValue().asString())
null != null && !value.equals(attribute.getDefaultValue().asString())
false && !value.equals(attribute.getDefaultValue().asString())
false && not evaluated anymore
return false;
```

jamezp: Sorry you're right. It should be:
```
return attribute.getDefaultValue() == null || !value.equals(attribute.getDefaultValue().asString())
```

However, that's just a minor nit and really shouldn't hold this PR up.

jamezp (approved):

PR #14615, opened 2021-09-06
[WFLY-15210] Remove MP extensions and subsystem from default configur…

yersan:
…ation for the Mixed Domain tests

Jira issue: https://issues.redhat.com/browse/WFLY-15210

The initial idea described in the Jira was to detect whether we are running a product as the Domain Controller for the mixed domain tests, however, since right now this is only used to bypass a difference about shipping MP extensions in the default config, I found easier to remove them if they exist, allowing the code to work also for the product.


bstansberry:
Thanks, @yersan

PR #14631, opened 2021-09-09
WFLY-15198 Avoid unneeded TimerImpl instances during db refresh

chengfang:
and some general code cleanup.

https://issues.redhat.com/browse/WFLY-15198 

The most important changes are:
 * DatabaseTimerPersistence.java line 557, add 2 params to `timerFromResult` method: (timerId, timerState). These 2 values are in most cases already retrieved from the `ResultSet` by the caller already. So just pass them in, instead of calling `getString()` inside the method again, to avoid getting the same column multiple times.
 * DatabaseTimerPersistence.java line 835, optimize the logic to reconcile potential difference between in-memory and db timer state, and delay the call to `timerFromResult` to avoid unneeded `TimerImpl` instantiation. First check in-memory timer state, then check db timer state, and only if there is a mismatch, then call `timerFromResult`.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -22,6 +22,8 @@
 
 package org.jboss.as.ejb3.timerservice.persistence.database;
 
+import static org.jboss.as.ejb3.timerservice.TimerServiceImpl.safeClose;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.Closeable;
@@ -43,7 +45,6 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.Date;
-import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -54,7 +55,6 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
-
 import javax.sql.DataSource;
 import javax.transaction.HeuristicMixedException;
 import javax.transaction.HeuristicRollbackException;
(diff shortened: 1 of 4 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -847,37 +823,40 @@ public void run() {
...
-                                            boolean validMemoryTimer = oldTimer != null && !valid.contains(oldTimer.getState());
-                                            // if timers memory - db are in non intersect subsets of valid/invalid states. we put them in sync
-                                            if (validMemoryTimer && validDBTimer) {
-                                                synchronized (DatabaseTimerPersistence.this) {
-                                                    knownTimerIds.get(timedObjectId).add(holder.timer.getId());
-                                                    listener.timerSync(oldTimer, holder.timer);
+                                        TimerImpl oldTimer = timerService.getTimer(id);
+                                        // if it is already in memory but it is not in sync we have a problem
+                                        // remove and add -> the probable cause is db glitch
+                                        boolean validMemoryTimer = oldTimer != null && !TimerState.CREATED_ACTIVE_IN_TIMEOUT_RETRY_TIMEOUT.contains(oldTimer.getState());
```
bstansberry, line 841: Is this the right name for this var? It seems more like 'invalidMemoryTimer'. I know it was already the name of the var, so changing it is not necessary to merge this; I'm just pointing it out because if it's not the best name it made the pretty straightforward logic here seem a kind of mysterious.

bstansberry (commented):
This LGTM. I made a comment about a var name; if you want to change that, that's fine, and if not that's fine too.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
chengfang: yeah, I felt the same. I've just renamed it to `invalidMemoryTimer` and hope it makes it a bit clearer.

bstansberry (approved):

PR #14633, opened 2021-09-09
[WFLY-15218] Drop Agroal model transformers prior to 2.0.0

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15218


yersan:
This PR still needs to be modified, I've realized that dropping the transformer XML file drops the necessary subsystem tests using expressions

PR #14634, opened 2021-09-09
[WFLY-15032] Use jboss-invocation-jakarta version for WildFly Preview

ropalka:
https://issues.redhat.com/browse/WFLY-15032


ropalka:
@bstansberry please review

PR #14646, opened 2021-09-12
[WFLY-15266] Move WildFly Preview to a native jakarta namespace varia…

bstansberry:
…nt of the ee subsytem integration module

https://issues.redhat.com/browse/WFLY-15266

Builds on #14642

PR #14652, opened 2021-09-14
[WFLY-15280] Move WildFly Preview to a native jakarta namespace variant of the ee-security subsytem integration module

bstansberry:
https://issues.redhat.com/browse/WFLY-15280

Builds on #14646


fjuma (approved):
Thanks @bstansberry!

PR #14663, opened 2021-09-15
WFLY-15291 Upgrade jberet-core from 2.0.1.Final to 2.0.2.Final in ee-…

chengfang:
…9 preview

https://issues.redhat.com/browse/WFLY-15291

In jberet-core 2.0.2.Final, the reference to javax transaction api has been removed, and so jberet-core.jar need not be transformed any more.

diff between jberet-core 2.0.1.Final and 2.0.2.Final: https://github.com/jberet/jsr352/compare/2.0.1.Final...2.0.2.Final


bstansberry:
:)

PR #14620, opened 2021-09-07
[WFLY-14987] Use Elytron SSLContext from MP Reactive Messaging Kafka Connector

kabir:
https://issues.redhat.com/browse/WFLY-14987 /  https://github.com/wildfly/wildfly-proposals/pull/411

Requires: #14618 (we currently contain the WFLY-14798 commits which are NOT part of this PR)
This has previously been reviewed in #14475

#14618, #14619 and #14620 (this PR) split up the original #14475 which had all the related work bundled together

PR #14619, opened 2021-09-07
[WFLY-14932] MP Reactive Messaging Kafka user API

kabir:
https://issues.redhat.com/browse/WFLY-14932 /  https://github.com/wildfly/wildfly-proposals/pull/402.

Requires: #14618 (we currently contain the WFLY-14798 commits which are NOT part of this PR)
This has previously been reviewed in #14475

#14618, #14619 (this PR) and #14620 split up the original #14475 which had all the related work bundled together

PR #14618, opened 2021-09-07
[WFLY-14798] Upgrade to MP Reactive Messaging 2.0

kabir:
https://issues.redhat.com/browse/WFLY-14798 /  https://github.com/wildfly/wildfly-proposals/pull/395

This has previously been reviewed in https://github.com/wildfly/wildfly/pull/14475. The only change is the _[WFLY-14798] [WFLY-15014] Upgrade the MP RM TCK to 2.0.1-RC1 _ commit which allows us to reenable some disabled TCK tests. The MP RM API stays the same until there is a final 2.0.1 release.

#14618 (this PR), #14619 and #14620 split up the original #14475 which had all the related work bundled together

PR #14668, opened 2021-09-16
[WFLY-15299] Kafka user API module should be visible to users

kabir:
https://issues.redhat.com/browse/WFLY-15299
Follow-up on #14620

Some more commits following up on #14619 and  #14618


microprofile/galleon-common/src/main/resources/modules/system/layers/base/io/smallrye/reactive/messaging/connector/kafka/api/main/module.xml
```diff
@@ -22,10 +22,6 @@
   -->
 
 <module xmlns="urn:jboss:module:1.8" name="io.smallrye.reactive.messaging.connector.kafka.api">
```
boris-unckel, line 24: This should be module:1.9 as actual version.

PR #14675, opened 2021-09-17
[WFLY-15307] AppClient o.j.a.a.s.parsing.AppClientXml readServerElement_18

boris-unckel:
AppClient o.j.a.a.s.parsing.AppClientXml readServerElement_18 not called

Fixes https://issues.redhat.com/browse/WFLY-15307

### DIFF
--- a/appclient/src/main/java/org/jboss/as/appclient/subsystem/parsing/AppClientXml.java
+++ b/appclient/src/main/java/org/jboss/as/appclient/subsystem/parsing/AppClientXml.java
@@ -123,7 +123,7 @@ public class AppClientXml extends CommonXml {
                 // Instead of having to list the remaining versions we just check it is actually a valid version.
                 for (Namespace current : Namespace.domainValues()) {
                     if (readerNS.equals(current)) {
-                        readServerElement_1_1(readerNS, reader, address, operationList);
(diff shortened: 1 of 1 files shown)


appclient/src/main/java/org/jboss/as/appclient/subsystem/parsing/AppClientXml.java
```diff
@@ -119,6 +119,9 @@ public void readElement(final XMLExtendedStreamReader reader, final List<ModelNo
             case DOMAIN_17_0:
                 readServerElement_1_1(readerNS, reader, address, operationList);
                 break;
+            case DOMAIN_18_0:
```
darranl, line 122: We don't need a specific case for 18 adding, we just need to call the new method in the for loop below which catches any version not listed above - i.e. all versions after 17.

darranl (changes requested):

appclient/src/main/java/org/jboss/as/appclient/subsystem/parsing/AppClientXml.java
boris-unckel: Fixed - I'll push after CI kill or completion

darranl (approved):
Thank you @boris-unckel

PR #14676, opened 2021-09-17
[WFLY-15243] Drop Transactions model transformers prior to 6.0.0

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15243


pferraro:
@mmusgrov Please review.

mmusgrov (approved):
I tested the PR locally and the changes look okay with the caveat that I have very little experience with the transformers.

rhusar (changes requested):
This is missing EAP_7_4_0 transformers tests and thus is removing code which shoundn't be removed.

yersan:
Hello @rhusar WF23 (base EAP 7.4) was shipped with Transaction model version 6.0.0. The idea is to drop all the transformers below WF23. As such, my understanding of the transaction subsystem is we have to drop all transformers, the current transaction model version is 6.0.0 since WF23/EAP74, which makes there is nothing to transform so far.

Am I missing something here?

bstansberry:
Yep; WF 23 used model version 6.0.0 which is the same as current.

rhusar:
@bstansberry @yersan While that's true, it's not a valid argument in this context. 

First of all, while the model is the same, the code handling it is different. That's why we have transformers tests in place to verify that the code that has since evolved in main branch - now WildFly 25 - is still yielding the same result as we expect on legacy EAP 7.4. slaves. Think about an example when a new item is inadvertently added to an Enum - while that would work on main, it would fail on EAP 7.4 - that's why we need to have those tests to detect regressions. Moreover, we are trying to create a sustainable software here, so eventually, those tests will have to be added so `TransactionSubsystemTestCase#testTransformersFull(...)` was deleted unnecessarily - which wouldn't have been if the EAP 7.4 tests were added.

yersan:
I see what do you mean @rhusar, although none of the extensions available on WildFly had added transformers against 7.4.0, so my understanding is the gap in the test cases was already there independently of dropping them. Does it sound correct to you?
My understanding was we were adding transformer tests if only if we have an extension management model upgrade.

From the feedback, it sounds like we have to add transformers always on each new product release for each extension or at least until the most immediate previous product release uses the same extension model version.

rhusar:
> I see what do you mean @rhusar, although none of the extensions available on WildFly had added transformers against 7.4.0

That's wrong. To find out tests against a controller version you can just do a search for the `ModelTestControllerVersion`; in this case specifically `org.jboss.as.model.test.ModelTestControllerVersion#EAP_7_4_0` – there are 5 test classes. 

> so my understanding is the gap in the test cases was already there independently of dropping them

Yes, that's what why I requested changes on the PR. Whoever does any new development in the subsystem, they need to add back all the code you have deleted which should not have been deleted.

> From the feedback, it sounds like we have to add transformers always on each new product release for each extension or at least until the most immediate previous product release uses the same extension model version.

We have historically done this, when a realease is done, new version is added to `ModelTestControllerVersion` and trackers were created to add the tests.

PR #14677, opened 2021-09-17
[WFLY-15246] Drop Weld model transformers prior to 4.0.0

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15246


yersan:
Hello @manovotn , failed test for Linux - JD11 is unrelated, would you mind reviewing this PR when you have a chance? Thanks!

manovotn:
Hello @yersan, I will take a look probably later today or tomorrow.

BTW does this mean that transformers are no longer present in WFLY? Or is this just dropping anything prior to current version and if we introduce a new config option, we will need to re-introduce this? I am asking because we will indeed need to introduce one such option for next major (for CDI in EE 10) :)

yersan:
Hi @manovotn , transformers are still useful for WildFly if you need to transform to WF 23 and later. Notice WF23 was the base server to build EAP 7.4

If you need to introduce a new config option that requires a transformation to make Weld management model work on a mixed domain supporting WF23 or later, then you will need to recover the basic structure of the files deleted on this PR, and start over a new transformer chain from current to 4.0.0. Any management model change prior to 4.0.0 doesn't need to take into account, because 4.0.0 was the management model version for WF 23.

Does it clear your doubts?

manovotn:
> Does it clear your doubts?

Yes, it basically says, I will have to reintroduce the very same thing as soon as WFLY starts accepting EE 10 changes.

I am talking about functionality described in https://issues.redhat.com/browse/WELD-2680
And the code that I had already drafted is here - https://github.com/wildfly/wildfly/compare/main...manovotn:weld2677?expand=1

That is mildly unfortunate but I don't think I can contribute my changes yet :-(

manovotn (approved):

yersan:
> That is mildly unfortunate

Yes, as soon as this gets merged it will conflict with your WIP, sorry @manovotn for the inconvenience. The advantage of emerging the changes now is we will reduce the time required to run the testsuite, and we need them merged before the next EAP version release. Keeping the transformers is otherwise harmless.

>  but I don't think I can contribute my changes yet :-(

Yes, so far we are producing EE 9.

PR #14678, opened 2021-09-17
[WFLY-15245] Drop Web Services model transformers prior to 3.0.0

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15245


yersan:
Hi @jimma , could you reviewing this when you have a chance? thanks

jimma:
@yersan  should we downgrade the current version to 2.0.0 : https://github.com/wildfly/wildfly/blob/main/webservices/server-integration/src/main/java/org/jboss/as/webservices/dmr/WSExtension.java#L72?

yersan:
Hi @jimma I don't think you should downgrade the version if there were already releases using 3.0.0. I'm not sure what would be the consequences but downgrade the model version looks counterintuitive to me

jimma:
@yersan  This is an internal version and I am just curious why this internal version doesn't consolidate with the schema version. This is the commit changes this version with the 2.0 transformer :https://github.com/wildfly/wildfly/commit/b8bf02299ac4255abe2eb573bf5e8e74d6ac6c1f

yersan:
@jimma Sometimes we could have model changes that don't necessarily require a schema update. b8bf022 is an example where the model changed from nonability to resolve expressions to the ability to resolve them. This ability is not represented in the schema, so, bump the model but let the schema lag behind sounds fine to me.

In any case, since we have already versions released using 3.0.0, we should not downgrade the current extension model version to align with the schema, you will be bitten later when you need to move forward from 2.0.0 to 3.0.0 due to any change in the future because you could have two different releases (maybe supported releases) using the same model version when those model versions are different.

jimma:
@yersan  Thanks for the update. It's good to merge this PR.

pferraro (approved):

PR #14699, opened 2021-09-22
WFLY-15306 Clear cached timer info from in-memory timers when using D…

chengfang:
…atabaseTimerPersistence

https://issues.redhat.com/browse/WFLY-15306


chengfang:
`org.jboss.as.test.multinode.ejb.timer.database.DatabaseTimerServiceRefreshTestCase#testTimerProgrammaticRefresh` failed, because the test assumes the timer info is obtained from in-memory cache, but with the code change in this PR, timer info can be erased from in-memory and so info will be retrieved from the database, and hence it returns the up-to-date value (reflecting the fact that this timer has been cancelled by node2, and so getting timer info from node1 will also return null). We should modify the test logic.

chengfang:
@tadamski  please review.

tadamski (approved):
LGTM

PR #14700, opened 2021-09-23
[WFLY-15336] Correct the list of unused modules for LayersTestCase

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15336

javax.transaction.api is no longer provisioned when we are creating the servlet distribution by using galleon layers.


bstansberry:
Thanks so much, @yersan :)

PR #14702, opened 2021-09-23
WFLY-13859 Clustering TS: Support testing against different Infinispan Server distributions

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-13859


rhusar:
Unfortunately the Infinispan server doesn't maintain parsing compatibility, so we need to be able to override server configuration as well.

diegolovison:
Looks good to me.

pferraro (approved):

PR #14707, opened 2021-09-24
[WFLY-15337] Manual Model microprofile tests fail due to missing stan…

yersan:
…dalone-microprofile.xml

Jira issue: https://issues.redhat.com/browse/WFLY-15337

Right now tests are failing when we are using the ee-galleon-pack distribution, which does not contain MP stuff. 

Instead of doing a considerable refactor of the manualmode pom, here we are disabling the tests when we are not testing a full server distribution.


yersan:
-- Alt Distribution
https://ci.wildfly.org/viewLog.html?buildId=274314&buildTypeId=WF_MasterLinuxAltDistJdk8

bstansberry (approved):
Thanks, @yersan!

PR #14708, opened 2021-09-24
[WFLY-15344] Fix wildfly.build.output.dir for elytron-oidc-client test module

OndrejKotek:
https://issues.redhat.com/browse/WFLY-15344

### DIFF
--- a/testsuite/integration/elytron-oidc-client/pom.xml
+++ b/testsuite/integration/elytron-oidc-client/pom.xml
@@ -46,7 +46,6 @@
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}/..</jbossas.project.dir>
         <surefire.forked.process.timeout>3600</surefire.forked.process.timeout>
-        <wildfly.build.output.dir>build/target/${server.output.dir.prefix}-${server.output.dir.version}</wildfly.build.output.dir>
         <!-- properties to enable plugins shared by various bootable jar executions -->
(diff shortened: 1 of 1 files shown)


testsuite/integration/elytron-oidc-client/pom.xml
```diff
@@ -46,7 +46,6 @@
         <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
         <jbossas.project.dir>${jbossas.ts.dir}/..</jbossas.project.dir>
         <surefire.forked.process.timeout>3600</surefire.forked.process.timeout>
-        <wildfly.build.output.dir>build/target/${server.output.dir.prefix}-${server.output.dir.version}</wildfly.build.output.dir>
```
yersan, line 49: @OndrejKotek This module is only available via galleon-pack, which means, it is only provisioned at the server built on build/dist.

At the moment of writing, this module is not available to other server distributions, for example, the distribution built from the ee-feature-pack module (ee-build/ee-dist). 

If your goal is to allow to supply a build output directory externally to test a custom server, then we need to ensure that these tests are not executed on the WildFly test suite when we are using ee-galleon-pack (ee-build/ee-dist), otherwise, the expectation of this Jira seems invalid.

Is your goal with this change to be able to supply an external directory built differently from what galleon-pack/ee-feature-pack modules do?

yersan (changes requested):

OndrejKotek:
After a discussion with Yeray, closing this one for now, waiting for the clarification of the OIDC client feature placement.

PR #14719, opened 2021-09-27
[WFLY-15318] Fix OcspTestCase failures with SE17

Ashpan:
https://issues.redhat.com/browse/WFLY-15318

### DIFF
--- a/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/TestingOcspServer.java
+++ b/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/TestingOcspServer.java
@@ -18,6 +18,7 @@
 
 package org.wildfly.test.integration.elytron.certs.ocsp;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.mockserver.model.HttpRequest.request;
 import static org.mockserver.model.HttpResponse.response;
 
@@ -42,6 +43,7 @@ import org.junit.Assert;
 import org.mockserver.integration.ClientAndServer;
 import org.mockserver.matchers.Times;
 import org.mockserver.model.Header;
+import org.mockserver.model.HttpRequest;
 import org.mockserver.model.HttpResponse;
 import org.mockserver.model.NottableString;
 import org.wildfly.common.iteration.ByteIterator;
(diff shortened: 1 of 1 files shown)


testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/OcspTestBase.java
```diff
@@ -349,7 +349,9 @@ public static void beforeTest() throws Exception {
     }
 
     public static void startOcspServer() throws Exception {
-        ocspServer = new TestingOcspServer(OCSP_PORT);
+        int version = Integer.parseInt(System.getProperty("java.version").split("\\.", 2)[0]);
```
fjuma, line 352: We shouldn't need the JDK version.

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/TestingOcspServer.java
```diff
@@ -123,12 +126,21 @@ public void start() throws Exception {
         server = new ClientAndServer(port);
         server.when(
                 request()
-                        .withMethod("POST")
-                        .withPath("/ocsp"),
+                        .withMethod(this.method)
```
fjuma, line 129: Instead of configuring the expected method based on the JDK version, it's possible to configure the mock server with multiple expectations, e.g.,

```
server.when(request().withMethod("POST")...)...respond();
server.when(request().withMethod("GET")...)...respond();
```
That way, we'll be able to handle the correct method regardless of the JDK version without having to worry about detecting the version ourselves.

fjuma (changes requested):

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/OcspTestBase.java
Ashpan: Fixed!

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/TestingOcspServer.java
Ashpan: Fixed!

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/certs/ocsp/TestingOcspServer.java
```diff
@@ -126,32 +128,48 @@ public void start() throws Exception {
...
+                .respond(request -> getFullHttpRequest(request, servlet));
+        server.when(
+                request()
+                        .withMethod("GET")
+                        .withPath("/ocsp/.*"),
+                Times.unlimited())
+                .respond(request -> getFullHttpRequest(request, servlet));
+    }
+
+    public HttpResponse getFullHttpRequest(HttpRequest request, HttpOcspServlet servlet){
```
fjuma, line 140: Very minor comment but would be good to rename this to getHttpResponse.

fjuma (approved):
Thanks very much @Ashpan!

PR #14725, opened 2021-09-28
[WFLY-15319] Ignore JSFDeploymentProcessorTestCase on SE > 8

fjuma:
https://issues.redhat.com/browse/WFLY-15319


fjuma:
@bstansberry While updating this to use `assumeJDKVersionBefore` and adding a check to see if the security manager is enabled, I realized that I needed to update this test to switch to manual deployment in order for the test to get skipped properly when the security manager is enabled.

bstansberry (approved):

PR #14728, opened 2021-09-28
[WFLY-15256] Add documentation for the Elytron OIDC Client subsystem

fjuma:
https://issues.redhat.com/browse/WFLY-15256

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
@@ -1,0 +1,151 @@
+[[Elytron_OIDC_Client]]
+= Elytron OpenID Connect Client Subsystem Configuration
+
+The ability to secure applications using https://openid.net/specs/openid-connect-core-1_0.html[OpenID Connect] is
+provided by the _elytron-oidc-client_ subsystem.
+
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
```diff
@@ -0,0 +1,150 @@
...
+        <provider-url>http://localhost:8180/auth/realms/demo</provider-url>
+        <ssl-required>external</ssl-required>
+        <credential name="secret" secret="0aa31d98-e0aa-404c-b6e0-e771dba1e798" />
+    </secure-deployment
+</subsystem>
+----
+
+The `secure-deployment` resource allows you to provide configuration for a specific deployment. In
+the example above, the `secure-deployment` resource is providing the configuration that should be used
+for the `APP_NAME.war` deployment.
```
bstansberry, line 97: This should be the runtime-name of the deployment, right? That is for cases where the runtime-name differs from the value of the deployment resource address / management name, the matching would be based on the runtime-name as that is what things like DUPs should deal with.

fjuma: Yes, thanks for catching that. I've updated this.

bstansberry (approved):

PR #14729, opened 2021-09-28
[WFLY-15343] More efficient/cleaner use of Maps

zodac-personal:
https://issues.redhat.com/browse/WFLY-15343

PR for the 'webservices' module

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSIntegrationProcessorJAXWS_JMS.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSIntegrationProcessorJAXWS_JMS.java
@@ -27,6 +27,7 @@ import static org.jboss.as.webservices.util.ASHelper.getAnnotations;
 import static org.jboss.as.webservices.util.DotNames.WEB_SERVICE_ANNOTATION;
 import static org.jboss.as.webservices.util.WSAttachmentKeys.JMS_ENDPOINT_METADATA_KEY;
 
+import com.arjuna.ats.internal.jdbc.drivers.modifiers.list;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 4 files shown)


webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSIntegrationProcessorJAXWS_JMS.java
```diff
@@ -92,37 +92,40 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
-                        QName serviceName = new QName(tns, service);
-                        QName portName = new QName(tns, port);
-                        String soapAddress = parser.filterSoapAddress(serviceName, portName, SOAPAddressWSDLParser.SOAP_OVER_JMS_NS);
-                        if (soapAddress != null) {
-                            ClassInfo webServiceClassInfo = (ClassInfo) ai.target();
-                            String beanClassName = webServiceClassInfo.name().toString();
-                            //service name ?
-                            list.add(new JMSEndpointMetaData(beanClassName, port, beanClassName, wsdlLocation, soapAddress));
-                        }
+        List<JMSEndpointMetaData> list = new LinkedList<>();
```
boris-unckel, line 95: May be you can safe some resource here:
L94:  `List<JMSEndpointMetaData> list = Collections.emptyList();`
L95: keep the guard: `if (!map.isEmpty()) {`
L96: `list = new LinkedList<>();`

zodac: I was thinking the same (existing early with an empty list if the map is empty), but the list gets passed into the constructor of `JMSEndpointsMetaData`, and that will replace any empty list with `Collections.emptyList()`. So I assume any unnecessary list will be garbage-collected away pretty quickly.

But I'm happy to change it if you feel it's worthwhile (saves other readers from needing to check how the list is used I guess).

boris-unckel: @zodac The previous implementation had the guard and the receiver of the list `JMSEndpointsMetaData` checks for null and emptiness. I assume this is intentionally and avoiding unnecessary object creation/processing seems useful here. `Collections.emptyList()` returns a static object - no (unnecessary!) garbage-collection will happen.

zodac: Sounds fair. I'll extract the loop logic to another method to make it clearer to return early with an empty list instead.

darranl:
@zodac Can you please confirm if you are still working on this PR?

As per the comment from Brian this will need squashing back to a single commit as the change does not justify multiple commits, also we are still missing the confirmation of the MIT licence terms.

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSIntegrationProcessorJAXWS_JMS.java
```diff
@@ -82,48 +83,56 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
-                        QName serviceName = new QName(tns, service);
-                        QName portName = new QName(tns, port);
-                        String soapAddress = parser.filterSoapAddress(serviceName, portName, SOAPAddressWSDLParser.SOAP_OVER_JMS_NS);
-                        if (soapAddress != null) {
-                            ClassInfo webServiceClassInfo = (ClassInfo) ai.target();
-                            String beanClassName = webServiceClassInfo.name().toString();
-                            //service name ?
-                            list.add(new JMSEndpointMetaData(beanClassName, port, beanClassName, wsdlLocation, soapAddress));
-                        }
+        final List<JMSEndpointMetaData> list = extractSOAPOverJMSBindings(unit, map);
```
darranl, line 91: Maybe it should have always been a read only list but at this point the list returned could now be read only OR modifiable.

darranl:
I am going to close this PR as it has been dormant for a few months now, if you would like to continue please reopen.

PR #14723, opened 2021-09-28
WFLY-15342 EJB Timers out of sync within cluster causing inexistent t…

chengfang:
…imer to be triggered

https://issues.redhat.com/browse/WFLY-15342

PR #14736, opened 2021-09-29
WFLY-15372 Remove the use of TimerHandle in ejb timer management oper…

chengfang:
…ations

https://issues.redhat.com/browse/WFLY-15372

This PR removes the use of `TimerHandle` from ejb timer management operations. It also removes the `timerHandle` field from `TimerImpl` class and modifies its `getHandle()` method to return a new instance of `TimerHandleImpl` for every call, since it's not worthwhile to save it as an instance variable.

~~Any conflict with https://github.com/wildfly/wildfly/pull/14723 should be manually resolved.~~ (conflict resolved)


bstansberry:
@chengfang Indeed there's a conflict. :)

bstansberry (approved):

PR #14743, opened 2021-10-01
[WFLY-15396] Override core to roll back to Jandex 2.3.1

bstansberry:
https://issues.redhat.com/browse/WFLY-15396

Unfortunately this reverts the https://issues.redhat.com/browse/WFLY-15278 fix.


jmesnil (approved):
Corresponding downgrade in WildFly Core with https://github.com/wildfly/wildfly-core/pull/4780/

PR #14744, opened 2021-10-01
WFLY-15391 Update observability module with a dependency on wildfly-elytron to use individual modules

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-15391


bstansberry:
@jasondlee Please review

jasondlee (approved):

jasondlee:
It looks good, but, just to make sure I understand, this is just narrowing down the scope of the dependency?

parsharma:
Yes, @jasondlee. Instead of using the whole elytron dependency, we can use an individual module.

yersan (commented):
This is reducing the needed dependencies at build time, and indeed can be done for runtime as well. The `org.wildfly.extension.opentelemetry-api`JBoss Modules module can also be narrowed down and use `org.wildfly.security.manager` instead of `org.wildfly.security.elytron`. `org.wildfly.extension.opentelemetry` looks fine.

jasondlee (approved):

PR #14762, opened 2021-10-06
[WFLY-15407] Convert "Hacking on WildFly" to the standard WF docs

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15407

The parent doc is Hacking_On_WildFly.adoc, the introduction for the reader is in the first document 
1)_hacking/github_setup.adoc
2) _hacking/contributing.adoc[]
3) _hacking/pullrequest_standards.adoc[]

The PR has a different order for the files. It would be great to receive both feedback regarding good English and fixes for markdown issues and/or best practise.

### DIFF
--- a/docs/src/main/asciidoc/Hacking_On_WildFly.adoc
+++ b/docs/src/main/asciidoc/Hacking_On_WildFly.adoc
@@ -1,0 +1,38 @@
+[[Hacking_On_WildFly]]
+= Hacking on WildFly
+WildFly developer team;
+:revnumber: {version}
+:revdate: {localdate}
+:toc: macro
+:toclevels: 2
+:toc-title: Hacking on WildFly
+:doctype: book
+:icons: font
+:source-highlighter: coderay
+:wildflyVersion: 26
+:sectlinks:
+:idseparator: -
+:include-depth: 2
+
+ifdef::env-github[:outfilesuffix: .adoc]
+
+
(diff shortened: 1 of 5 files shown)


yersan (commented):
Hello @boris-unckel I've added some comments.

I would also add this new documentation would deserve an entry on the main documentation file: https://github.com/wildfly/wildfly/blob/dc5a12c4d7bb515bdad24267fd9d87bf66a3de25/docs/src/main/asciidoc/index.adoc

At least under the "More Resources" section and maybe also on wildfly.org

yersan (commented):
Changes look good to me, thanks @boris-unckel, I've just added some minor points I found when testing it

About if finally, we are going to link it on the index, let's wait from @bstansberry comment about it.

bstansberry (commented):
Thanks so much @boris-unckel! I have feedback but the general content is very nice!

bstansberry:
@boris-unckel 

+1 to adding to index.adoc

Re the README, good suggestion. There is no 'latest'. We could work out how to create such a thing, but I think that should be a separate task. And we wouldn't want to update the README until the doc it points to is published, which won't be for a while, as we publish docs with the major releases.

boris-unckel:
@bstansberry I have added the link to the index.adoc. Ready to fly :-)

bstansberry (approved):

PR #14770, opened 2021-10-07
WFLY-15429 Generalize org.wildfly.clustering.ejb.BeanContext for reuse by multiple EJB component clustering services

pferraro:
https://issues.redhat.com/browse/WFLY-15429


rachmatowicz (approved):
Looks good to me, allowing configuration of different bean types using the same underlying base class.

PR #14773, opened 2021-10-07
[WFLY-12610] adding properties to setup transaction recovery and operation to run recovery

ochaloup:
https://issues.redhat.com/browse/WFLY-12610

The point of this PR is giving a way to execute Narayana recovery processing from jboss cli without a need to open socket + to make recovery configuration changes without a need to restart JVM + do not suspend the recovery manager when WildFly is suspended (e.g. WFLY suspend happens on call of the graceful shutdown).


ochaloup:
I need to fix the two failing tests but the concept of the change should be working.

mmusgrov (approved):
The functionality looks very good indeed.

I had a number of minor non-functional comments/queries (which, apart from my comment about `TransactionSubsystem61Parser.java`, you may ignore) so I am approving the PR.

bstansberry:
Superseded by #15382

PR #14776, opened 2021-10-08
WFLY-15456 Do not register timer service resource for stateful sessio…

chengfang:
…n bean

https://issues.redhat.com/browse/WFLY-15456


tadamski (approved):
LGTM

PR #14780, opened 2021-10-10
[WFLY-15469] Add missing keyword "abstract" (connector)

boris-unckel:
AbstractComplexSubsystemTestCase

Fixes https://issues.redhat.com/browse/WFLY-15469


tadamski (approved):
@boris-unckel sure, approved

PR #14782, opened 2021-10-11
Create SECURITY.md

zidingz:
A simple instruction for security researchers.


bstansberry:
Thanks @zidingz -- we added one not too long ago at https://github.com/wildfly/wildfly/blob/main/SECURITY.md. It took a while to get the mail list we wanted to use set up.

PR #14789, opened 2021-10-13
[WFLY-15452] - allow specifying AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-15452

____

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
@@ -5,10 +5,15 @@
 
 package org.wildfly.extension.undertow;
 
+import java.util.EnumSet;
+import java.util.Map;
+
+import org.jboss.as.controller.Feature;
 import org.jboss.as.controller.PersistentResourceXMLDescription;
 import org.jboss.as.controller.PersistentSubsystemSchema;
 import org.jboss.as.controller.SubsystemSchema;
 import org.jboss.as.controller.xml.VersionedNamespace;
+import org.jboss.as.version.Stability;
 import org.jboss.staxmapper.IntVersion;
 
 /**
@@ -35,10 +40,10 @@ public enum UndertowSubsystemSchema implements PersistentSubsystemSchema<Underto
     VERSION_12_0(12),   // WildFly 23-26.1, EAP 7.4
     VERSION_13_0(13),   // WildFly 27       N.B. There were no schema changes between 12.0 and 13.0!
     VERSION_14_0(14),   // WildFly 28
-    VERSION_15_0(15),   // WildFly 31-present
+    VERSION_14_0_PREVIEW(14, 0, Stability.PREVIEW)   // WildFly 32-present
     ;
-    static final UndertowSubsystemSchema CURRENT = VERSION_15_0;
 
+    static final Map<Stability, UndertowSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_14_0, VERSION_14_0_PREVIEW));
     private final VersionedNamespace<IntVersion, UndertowSubsystemSchema> namespace;
 
     UndertowSubsystemSchema(int major) {
@@ -50,7 +55,15 @@ public enum UndertowSubsystemSchema implements PersistentSubsystemSchema<Underto
     }
 
     UndertowSubsystemSchema(IntVersion version) {
-        this.namespace = SubsystemSchema.createLegacySubsystemURN(UndertowExtension.SUBSYSTEM_NAME, version);
+        this(version, Stability.DEFAULT);
+    }
+
+    UndertowSubsystemSchema(final int major, final int minor, final Stability stability) {
+        this(new IntVersion(major, minor), stability);
+    }
+
+    UndertowSubsystemSchema(final IntVersion version, final Stability stability) {
+        this.namespace = SubsystemSchema.createLegacySubsystemURN(UndertowExtension.SUBSYSTEM_NAME, stability, version);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


kabir (dismissed):
@baranowb I think you need a transformers tests (apologies if there is one and I missed it).

Also, you need to bump UndertowExtension.CURRENT_MODEL_VERSION, since 11.0.0 has already been in a release and you're changing the model

undertow/src/main/resources/schema/wildfly-undertow_12_1.xsd
```diff
@@ -0,0 +1,1078 @@
...
+                        <xs:documentation>
+                            <![CDATA[
+                                                If this listener is supporting non-SSL requests, and a request is received for which a matching <security-constraint> requires SSL transport,
+                                                undertow will automatically redirect the request to the socket binding port specified here.
+                                               ]]>
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:attribute>
+                <xs:attribute name="max-ajp-packet-size" type="xs:int"/>
+                <xs:attribute name="allowed-request-attributes-pattern" use="optional" type="xs:string"/>
```
fl4via, line 244: please, add a CDATA documenting it here as well

undertow/src/main/java/org/wildfly/extension/undertow/Namespace.java
```diff
@@ -48,12 +48,13 @@
...
     UNDERTOW_11_0("urn:jboss:domain:undertow:11.0"),
     UNDERTOW_12_0("urn:jboss:domain:undertow:12.0"),
+    UNDERTOW_12_1("urn:jboss:domain:undertow:12.1"),
     ;
 
     /**
      * The current namespace version.
      */
-    public static final Namespace CURRENT = UNDERTOW_12_0;
+    public static final Namespace CURRENT = UNDERTOW_12_1;
```
fl4via, line 57: Historically we have been incrementing only the major version number for quite a while. It is intersting that you brought up this, though, because I see that Undertow has 1.1, 1.2 and 3.1. @bstansberry do we have any policies in WF as to when should we upgrade the major model version versus the minor one?

fl4via (changes requested):
Have you checked if there is need to update the community documentation in WF as well (in docs dir)?

darranl (commented):
For changes like this it is clearer both now and in the future to separate the version bump into it's own separate commit, when the two are combined like this it is not possible to use the GitHub UI or git diff tools later to identify the changes made to the schema.

darranl:
Added "missing_reqs" as the requirements for a feature request have not been met.

Also related to my comment about the schema version etc.. I see you have a second PR also duplicating the changes - this is another reason to use a separate commit for the bump - both RFEs can then continue from the same commit reducing the scope for conflict.

baranowb:
> Have you checked if there is need to update the community documentation in WF as well (in docs dir)?

Yes. AFAIR there is no documentation that covers atttribs now, its all generated from model and descriptor files as "model reference.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowTransformers.java
```diff
@@ -155,6 +166,12 @@ private static void registerTransformers_EAP_7_2_0(ResourceTransformationDescrip
                     .setDiscard(DiscardAttributeChecker.DEFAULT_VALUE, PRESERVE_PATH_ON_FORWARD)
                     .addRejectCheck(RejectAttributeChecker.DEFINED, PRESERVE_PATH_ON_FORWARD)
                 .end();
+        final ResourceTransformationDescriptionBuilder serverBuilder = subsystemBuilder.addChildResource(UndertowExtension.SERVER_PATH);
+
+        final AttributeTransformationDescriptionBuilder ajp = serverBuilder.addChildResource(UndertowExtension.AJP_LISTENER_PATH).getAttributeBuilder()
```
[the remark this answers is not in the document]
baranowb: Im almost sure it had to be done, buy it might have been due to wrong current model. I will double check.

undertow/src/main/java/org/wildfly/extension/undertow/Namespace.java
bstansberry: @fl4via @baranowb We should always do a major bump in WildFly main / WildFly Core main. That eliminates varying interpretations of what's 'minor' and leaves minor versions available for use in other streams if necessary.

I've never gotten a sense that people are trying to get semantic meaning from our API/schema versions so there's not much downside to just doing majors.

baranowb: @bstansberry roger.

fl4via:
> > Have you checked if there is need to update the community documentation in WF as well (in docs dir)?
> 
> Yes. AFAIR there is no documentation that covers atttribs now, its all generated from model and descriptor files as "model reference.

In fact, you're right.

fl4via (approved):

bstansberry (approved):

fl4via (approved):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
```diff
@@ -61,4 +75,4 @@ public VersionedNamespace<IntVersion, UndertowSubsystemSchema> getNamespace() {
     public PersistentResourceXMLDescription getXMLDescription() {
         return UndertowPersistentResourceXMLDescriptionFactory.INSTANCE.apply(this);
```
PrarthonaPaul, line 76: I am not sure if this is done already, but it would be good to add the new attributes to the XML description based on the stability level. You can check this using the `.since(versionNumber)` function. 
Something like this: https://github.com/wildfly/wildfly-core/blob/main/elytron/src/main/java/org/wildfly/extension/elytron/ElytronSubsystemSchema.java#L195-L197

PrarthonaPaul (commented):
Thanks @baranowb 
I have added a small comment. But other than that, the changes look good to me! 

Also, I noticed the schema changes related to this RFE is added to the subsystem bump commit. If you are going to cherry-pick the commit to the other RFE, it would be good to move that change to the RFE commit.

darranl:
If making any changes please use an additional commit and do not change the existing commits.

bstansberry:
This was merged as part of #18002.

Thanks @baranowb @kabir @darranl @PrarthonaPaul @fl4via

PR #14794, opened 2021-10-13
WFLY-15494 Duplicate dependencies in module.xml

clara0:
https://issues.redhat.com/browse/WFLY-15494


clara0:
@bstansberry Yes, I agree to the MIT License terms.

PR #14805, opened 2021-10-17
[WFLY-15512] Sync like parent method (ejb)

boris-unckel:
Overrides should match their parent class methods in synchronization

Fixes https://issues.redhat.com/browse/WFLY-15512


boris-unckel:
@chengfang Could you review and approve, please?

PR #14809, opened 2021-10-17
[WFLY-15517] "ThreadLocal" variables clean up (jpa)

boris-unckel:
"ThreadLocal" variables should be cleaned up when no longer used (jpa)

Fixes https://issues.redhat.com/browse/WFLY-15517
as part of https://issues.redhat.com/browse/WFLY-15513


stuartwdouglas:
This will result in a lot more list allocations for no real gain.

PR #14810, opened 2021-10-17
[WFLY-15518] Test preparation errors shall cause failure (connector)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15518


tadamski (changes requested):
@boris-unckel Looking at the test, it seems to me as well that it currently doesn't check anything. No exceptions are thrown because there are no files that were actually supposed to be tested. The particular test that you modified (I agree with the change btw) fails because there is no directory with this specific configuration. IMO the test requires rewriting from the scratch to work.

tadamski (approved):
@boris-unckel thanks for refactoring the whole test; I'm looking at the JCA failure that you mentioned

tadamski:
I have fixed JBJCA-1435. We would be able to remove @ignore after IJ 1.5.3 is released and merged.

PR #14816, opened 2021-10-18
WFLY-15499 Honor sybase as database value, and remove the unused fiel…

chengfang:
…d databaseDialects in DatabaseTimerPersistence class

https://issues.redhat.com/browse/WFLY-15499


tadamski (approved):
LGTM

PR #14822, opened 2021-10-20
[WFLY-15540] Move WildFly Preview to a native Jakarta namespace varia…

bstansberry:
…nt of the jpa/spi module

https://issues.redhat.com/browse/WFLY-15540


yersan (approved):

scottmarlow:
Question, does the jipijapa-spi-jakarta artifact need to be built before the Hibernate 6 integration code is compiled?  The Hibernate 6 integration code will probably be under wildfly/jpa/hibernate6.

scottmarlow (approved):

PR #14823, opened 2021-10-20
[WFLY-15491] Move WildFly Preview to a native Jakarta namespace variant of the JSF subsystem module

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15491

It also upgrades the JSF Mojarra implementation to the latest available implementation for JSF 3.0.

Notice license files are still not being processed as expected due to a different issue, see https://issues.redhat.com/browse/WFLY-15536

@bstansberry With these changes in place, I still can see the `wildfly-jsf-jakarta` artifact getting transformed, I've tried to find the reason, but I'm not sure how I can debug it, would you mind shedding some light here and getting us to know how we can debug and find the root cause?

@fjuma just for your info, keep in mind we are upgrading the JSF spec version here, so I would expect some changes in the documentation and in the JSF subsystem too, because this subsystem is prepared to work with different JSF implementation versions, so it seems there would be some additional work on the JSF subsystem outside of this Jira.

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -519,6 +519,14 @@
                 </exclusion>
                 <exclusion>
                     <groupId>${project.groupId}</groupId>
+                    <artifactId>wildfly-jsf</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>${project.groupId}</groupId>
+                    <artifactId>wildfly-jsf-injection</artifactId>
+                </exclusion>
+                <exclusion>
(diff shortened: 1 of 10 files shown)


bstansberry (commented):
This LGTM in terms of the structure of the change for the transformed modules. @fjuma can you have a look? Note the MyFaces upgrade in WildFly Preview.

fjuma (approved):
Thanks very much for working on this @yersan! This looks good to me, including the MyFaces upgrade.

PR #14824, opened 2021-10-20
[WFLY-15536] License files need to be filtered by the maven resource …

yersan:
…plugin

Jira issue: https://issues.redhat.com/browse/WFLY-15536

I wonder if we should move this filtering to the license plugin instead, that would prevent wildfly / wildfly core take care of it and, maybe potentially avoid filtering resources unnecessarily.

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -204,6 +204,7 @@
                             <resources>
                                 <resource>
                                     <directory>${mp.galleon.common.resources.directory}</directory>
+                                    <filtering>true</filtering>
                                     <excludes>
                                         <exclude>layers/standalone/jaxrs</exclude>
                                     </excludes>
@@ -223,6 +224,7 @@
                             <resources>
                                 <resource>
                                     <directory>${oidc.galleon.common.resources.directory}</directory>
+                                    <filtering>true</filtering>
                                 </resource>
                             </resources>
                             <overwrite>true</overwrite>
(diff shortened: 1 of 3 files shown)


yersan:
@bstansberry First of all, let me describe what are the license configuration files doing, please correct me if I am making a mistake in my interpretation. This is a license configuration file: https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml

They are there to overwrite (or even supply when the license information does not exist?) when necessary some properties related to the license information that can be fetched from a maven artifact that is under dependency. For example, properties like `name` and `url`. I think the intention is to reduce the available set of licenses, sometimes the name and url are different for two artifacts but they are talking about the same license.

This is an issue because if we don't filter the GA maven properties used on the license configuration files, then the entry for such dependency will be doing nothing there because the maven license plugin won't be able to match the license included in an artifact with the configuration we want for such artifact.

I suspect it has gone unnoticed because the license entries inside the relevant artifacts, I mean the artifacts that are using GA as maven properties, are the expected ones. Indeed, if you prefer do not to filter these resources, I think we could go ahead and remove them from the license configuration file. I think we have been adding entries and entries there just because seems logical, but my understanding is the available set of maven dependencies are the artifacts used to extract a license, and if the license entry is in a license configuration file, their properties will be replaced, if there is no entry, then it will use the entry available on the artifact which being processed.

PR #14827, opened 2021-10-20
WFLY-15552 contention due to the synchronized block in TimerServiceIm…

chengfang:
…pl.getTimers()

https://issues.redhat.com/browse/WFLY-15552

* replacing `HashMap` with `ConcurrentHashMap` for `timers`
* replacing `HashSet` with `ArrayList` inside `getTimers()` method as the impl type to hold the result. The method return type is still `Collection`.
* use `ConcurrentMap.putIfAbsent(...)` inside `registerTimerResource()` method.
* Inside method `getTimers()`, the weakly consistent view of the underlying timers cache should be okay to produce the method return result. With the previously used synchronized block, it was always consistent.
* Previously inside `registerTimerResource()` and ` unregisterTimerResource()`, adding/removing from timers cache and registering/unregistering with management resources are in the same synchronized block. In this PR, the resource management part is no longer in the same critical section.  Registering/unregistering timer resources should not block other access to timers.
* Use `EnumSet` instead of `HashSet` to hold timer states `EXPIRED`, `CANCELED`.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
@@ -39,6 +39,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


chengfang:
@tadamski please review.

tadamski (approved):
@chengfang LGTM

PR #14832, opened 2021-10-24
[WFLY-15582] Remove unneeded code from legacy security subsystem; prune…

bstansberry:
… pom and module dependencies

https://issues.redhat.com/browse/WFLY-15582

First bit of work for WFLY-15566, which is to restore the legacy security extension to WildFly Preview, as there's no reason not to make its current model-only incarnation available there.

But first I want to reduce the module dependencies to the needed minimum. Which led to a general code deletion and tightening exercise. That seems worthwhile on its own and need not block for the WF Preview work, so here it is.


bstansberry:
@fjuma Please review.

fjuma (approved):
Thanks @bstansberry!

PR #14833, opened 2021-10-24
[WFLY-15465][WFLY-15536] Properly control maven-resource-plugin filtering

bstansberry:
https://issues.redhat.com/browse/WFLY-15465
https://issues.redhat.com/browse/WFLY-15536

Also a commit to remove a risky bit of apparently unused filtering cruft in the jaxrs subsystem pom.

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -236,8 +236,18 @@
                                 <resource>
                                     <directory>${mp.galleon.common.resources.directory}</directory>
                                     <excludes>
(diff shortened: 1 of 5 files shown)


yersan (approved):
LGTM

I'm assuming we have added the module.xml and license filtering only for the needed directories. I mean, for example, wildfly-core license files don't need to be filtered so we are not adding the license directory when we are unpacking the wildfly-core resources, `ee-8-api/src/main/resources` although contains license configuration file and modules we are not filtering them because at the moment they don't need to be filtered and so on. I assume this is intentional and not an oversight.

PR #14835, opened 2021-10-25
WFLY-15562 Simplify logging and validation in various createTimer met…

chengfang:
…hods in TimerServiceImpl

https://issues.redhat.com/browse/WFLY-15562

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
@@ -244,6 +244,9 @@ public class TimerServiceImpl implements TimerService, Service<TimerService> {
     public Timer createCalendarTimer(ScheduleExpression schedule, TimerConfig timerConfig)
             throws IllegalArgumentException, IllegalStateException, EJBException {
         assertTimerServiceState();
+        if (schedule == null) {
+            throw EJB3_TIMER_LOGGER.invalidTimerParameter("schedule", null);
+        }
         Serializable info = timerConfig == null ? null : timerConfig.getInfo();
         boolean persistent = timerConfig == null || timerConfig.isPersistent();
         return this.createCalendarTimer(schedule, info, persistent, null);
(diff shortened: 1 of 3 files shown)


tadamski (approved):
LGTM

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
```diff
@@ -257,13 +260,13 @@ public Timer createIntervalTimer(Date initialExpiration, long intervalDuration,
             throws IllegalArgumentException, IllegalStateException, EJBException {
         assertTimerServiceState();
         if (initialExpiration == null) {
-            throw EJB3_TIMER_LOGGER.initialExpirationIsNullCreatingTimer();
+            throw EJB3_TIMER_LOGGER.invalidTimerParameter("initialExpiration", null);
```
boris-unckel, line 263: You might consider [org.wildfly.common.Assert.checkNotNullParam](https://github.com/wildfly/wildfly-common/blob/1.5.4.Final/src/main/java/org/wildfly/common/Assert.java#L47) for these cases.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
```diff
@@ -257,13 +260,13 @@ public Timer createIntervalTimer(Date initialExpiration, long intervalDuration,
...
         }
         if (initialExpiration.getTime() < 0) {
```
boris-unckel, line 265: You might consider [org.wildfly.common.Assert.checkMinimumParameter](https://github.com/wildfly/wildfly-common/blob/1.5.4.Final/src/main/java/org/wildfly/common/Assert.java#L305) for these cases.

chengfang:
While reusing WildFly common Assert is good, using ejb's own error logging allows for more direct handling and targeted message. So in this case, I'm leaning towards ejb's own.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java
bstansberry: The EJB3_TIMER_LOGGER message provides more context and is internationalized.

PR #14840, opened 2021-10-26
[WFLY-15574] Upgrade com.squareup.okhttp3 3.14.9

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-15574

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -307,7 +307,7 @@
         <version.com.ibm.async.asyncutil>0.1.0</version.com.ibm.async.asyncutil>
         <version.com.microsoft.azure>8.6.5</version.com.microsoft.azure>
         <version.com.nimbus.jose-jwt>8.11</version.com.nimbus.jose-jwt>
-        <version.com.squareup.okhttp3>3.9.0</version.com.squareup.okhttp3>
+        <version.com.squareup.okhttp3>3.14.9</version.com.squareup.okhttp3>
         <version.com.squareup.okio>1.17.5</version.com.squareup.okio>
         <version.com.sun.activation.jakarta.activation>1.2.2</version.com.sun.activation.jakarta.activation>
         <version.com.sun.faces>2.3.14.SP04</version.com.sun.faces>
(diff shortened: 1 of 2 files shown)


rhusar (changes requested):
If we can upgrade this will need to remove this workaround as well - https://github.com/wildfly/wildfly/blob/main/testsuite/integration/clustering/pom.xml#L50-L51

jasondlee (approved):
@rhusar's requests aside, this looks fine to me.

rhusar:
I didn't get to updating my review before the merge, the changes I requested are good.

However, @jasondlee @boris-unckel I am wondering why this specific version for upgrade? If I am looking at the right places, we are consuming jaeger 1.6 which already uses v4.9.0? 

https://github.com/jaegertracing/jaeger-client-java/blob/v1.6.0/build.gradle#L28
https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp

boris-unckel:
@rhusar The 3.x series stays Java based, the 4.x starts with Kotlin. The update was focusing on "latest minor" in terms of bugfixing. I don't know the Kotlin dependencies / consequences.

rhusar:
@boris-unckel OK I see the logic (I was a bit surprised when I was using older version of the library with infinispan server driver library that it was failing silently (just timing out) so that's why I needed that upgrade in there).

rhusar:
@bstansberry BTW do we have a specific policy against consuming "non-Java" dependencies, e.g. libraries written in Kotlin? TBH I am not aware if we consume anything non-Java at the moment.

bstansberry:
@rhusar Before doing anything like that we'd need to discuss with folks like @rgdoliveira.

rgdoliveira:
@bstansberry @rhusar (cc @dosoudil @dbkreling) PNC has now an environment with Kotlin, so I believe it is possible to productize a Kotlin project, but we haven't tried yet as we didn't have any projects like this before.

PR #14846, opened 2021-10-27
WFLY-15573 State transfer can fail due to ConcurrentModificationException if distributed web session attributes are concurrently modified

pferraro:
https://issues.redhat.com/browse/WFLY-15573


rhusar (approved):
LGTM

PR #14855, opened 2021-10-28
[WFLY-14967] Move WildFly Preview to a native Jakarta namespace varia…

boris-unckel:
…nt of the jpa/eclipselink module
Fixes https://issues.redhat.com/browse/WFLY-14967

Adds jakarta.json.api as optional dependency:
https://issues.redhat.com/browse/WFLY-9566

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -511,6 +511,10 @@
                 </exclusion>
                 <exclusion>
                     <groupId>${project.groupId}</groupId>
+                    <artifactId>jipijapa-eclipselink</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>${project.groupId}</groupId>
                     <artifactId>wildfly-bean-validation</artifactId>
                 </exclusion>
                 <exclusion>
@@ -1043,6 +1047,12 @@
         <dependency>
             <groupId>${ee.maven.groupId}</groupId>
             <artifactId>jipijapa-spi-jakarta</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${ee.maven.groupId}</groupId>
+            <artifactId>jipijapa-eclipselink-jakarta</artifactId>
             <scope>provided</scope>
         </dependency>
 
--- a/ee-9/pom.xml
+++ b/ee-9/pom.xml
@@ -71,6 +71,7 @@
         <!--<version.jakarta.xml.bind.jakarta-xml-bind-api>3.0.0</version.jakarta.xml.bind.jakarta-xml-bind-api>-->
         <version.org.apache.myfaces.core>3.0.1</version.org.apache.myfaces.core>
         <version.org.eclipse.yasson>2.0.1</version.org.eclipse.yasson>
+        <version.org.eclipselink.version>3.0.2</version.org.eclipselink.version>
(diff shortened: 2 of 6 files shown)


scottmarlow (approved):

yersan (commented):
Hi @scottmarlow , I've noticed two things that I don't understand at all on the changes on this PR, so this is just for your info:

1. This PR is not introducing any code change on jpa/jipijapa-eclipselink, however, it is redefining its module on ee-9 adding additional module dependencies, not sure why, but it sounds weird to me.
2. If the module is the same, instead of re-defining it, I think we should try to re-use it moving it outside of the ee-feature-pack/pruned and letting the maven-resource-plugin deal with it by adding the -jakarta suffix when neccesary.

ee-9/pom.xml
```diff
@@ -71,6 +71,7 @@
         <!--<version.jakarta.xml.bind.jakarta-xml-bind-api>3.0.0</version.jakarta.xml.bind.jakarta-xml-bind-api>-->
         <version.org.apache.myfaces.core>3.0.1</version.org.apache.myfaces.core>
         <version.org.eclipse.yasson>2.0.1</version.org.eclipse.yasson>
+        <version.org.eclipselink.version>3.0.2</version.org.eclipselink.version>
```
bstansberry, line 74: This property has no effect. To change the version used in the ee-9 modules you'd need to add an entry for this artifact into the dependencyManagement section of this pom.

boris-unckel:
@bstansberry @scottmarlow thanks for your feedback, based on that I have started with the following work items:

1. I have created https://github.com/wildfly/wildfly/pull/14885 which moves eclipselink from prune to common without deprecated dependencies.
2. If it's merged I would prepare a separate PR for WFLY-9566 which adds "javax.json.api" as optional dependency.
3. After the WFLY-9566 merge I would rebase / update WFLY-14967 using the expression syntax.
4. If there is need and agreement I can prepare a new Jira/PR to use eclipselink 3.0.2 for the WildFly Preview.

darranl:
FYI this PR also presently has merge conflicts.

darranl:
As this is an active area being developed there is a risk of further merge conflicts but is it possible to get these current ones addressed so we can see if we can get it reviewed and merged?

boris-unckel:
> As this is an active area being developed there is a risk of further merge conflicts but is it possible to get these current ones addressed so we can see if we can get it reviewed and merged?

Hello @darranl thanks for keeping an eye on this one. As the preparation steps are not ready yet, it makes no sense to rebase right now. This would be step 3 mentioned here: https://github.com/wildfly/wildfly/pull/14855#issuecomment-962580936 Step 2 is here https://github.com/wildfly/wildfly/pull/15042

ee-9/feature-pack/pom.xml
```diff
@@ -1046,6 +1050,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>${ee.maven.groupId}</groupId>
+            <artifactId>jipijapa-eclipselink-jakarta</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
yersan, line 1057: Hi @boris-unckel , just in case you missed it on the wildfly google groups, notice that we have changed the structures of the wildfly preview FP and this dependency and its license file now should be placed in `ee-9/common` maven module.

boris-unckel:
Superseded by https://github.com/wildfly/wildfly/pull/15070

PR #14861, opened 2021-10-29
WFLY-15569 Database persistent auto timer are created twice

chengfang:
https://issues.redhat.com/browse/WFLY-15569

Added a create-auto-timer property sql statement to create an auto timer only if no equally-configured auto timer exists in the database table.

Tested with timer database of H2, PostgreSQL & MySQL. It should also work with Oracle (where the issue was reported) and other major DBMS.


chengfang:
@tadamski please review.

tadamski (commented):
@chengfang IIUC this change is supposed to synchronize automatic timers using the database. I have few questions regarding the implementation:
* IIUC automatic timers weren't added to database because they were automatically recreated on each redeployment. Won't the record in the database prevent recreation of the timer after server restart?
* Would this mechanism interfere with distributed timers synchronized by clustering? Would those mechanisms be complementary?
* TBH I'm not sure if I understand the auto timer SQL correctly. Why do you use the dummy table in the statement instead of just adding the record if another one is not present?

chengfang:
@tadamski  persistent auto timers are saved to database (or other persistence store), and when redeployed, any auto timers are matched against existing auto timers in db, and if there is a match, the existing auto timer is used. New auto timers are created only if there is no matching existing auto timers from persistence.

This enhancement should work well with database-based distributed timers, and is part of it. 
 
The dummy table used in the auto timer query is there just to meet the syntax requirement of all supported database products, and the dummy table practically contains no content. The basic idea of that jdbc statement is to create a new auto timer only if the same auto timer does not exist. It is complicated by the need to combine both the check and create parts into one single statement, while also making it work in all supported database products.

darranl:
@chengfang / @tadamski FYI we have a merge conflict on this one - is this a discussion to progress so we can get this reviewed?

darranl:
Thanks, I have removed the label so sounds like we just need the review discussion to continue with @tadamski

tadamski (approved):
I just wanted to clarify my doubts but after the explanation those changes look good to me.

PR #14862, opened 2021-10-30
WFLY-14559 Add four new RESTEasy context parameters

ronsigal:
https://issues.redhat.com/browse/WFLY-14559


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsParamHandler.java
```diff
@@ -105,10 +105,18 @@ private void updateServerConfig(OperationContext context, String attributeName,
...
+            config.setResteasyMatchCacheSize(value);
         } else if (JaxrsConstants.RESTEASY_MEDIA_TYPE_MAPPINGS.equals(attributeName)) {
             config.setResteasyMediaTypeMappings(value);
         } else if (JaxrsConstants.RESTEASY_MEDIA_TYPE_PARAM_MAPPING.equals(attributeName)) {
             config.setResteasyMediaTypeParamMapping(value);
+        } else if (JaxrsConstants.RESTEASY_PATCH_FILTER_DISABLED.equals(attributeName)) {
+            config.setResteasyPatchfilterDisabled(value);
+        } else if (JaxrsConstants.RESTEASY_PATCH_FILTER_LEGACY.equals(attributeName)) {
+            config.setResteasyPatchfilterLegacy(value);
         } else if (JaxrsConstants.RESTEASY_PREFER_JACKSON_OVER_JSONB.equals(attributeName)) {
```
chengfang, line 120: why not use switch on enum instead of the long series of string if-else comparison? The enum should be much faster.

ronsigal: Hi @chengfang,

Good question. Like
```
       switch (attributeName) {
            case JaxrsElement.JAXRS_2_0_REQUEST_MATCHING.getLocalName():
                
        }
```
or even
```
        switch (attributeName) {
            case JaxrsConstants.JAXRS_2_0_REQUEST_MATCHING:
                ...
        }
```
Ok, I can change that.

-Ron

chengfang: I guess there are already Enum defined for every xml element and attribute names, so we can switch on enum instead of string.

ronsigal: If you prefer. Wouldn't it be faster to switch on Strings instead of calling getLocalName()?

chengfang: For the `attributeName` passed to this method, we can get the corresponding enum value. I guess it's already defined in `JaxrsElement` or similar enum types.  Then we can switch on the enum value of the attribute.

```java
switch(attributeNameEnum) {
  case JaxrsElement.JAXRS_2_0_REQUEST_MATCHING:
```

Will something like that work in this case?

ronsigal: Oh, I see. Looks good to me.

ronsigal: Done

chengfang (approved):

ronsigal:
Thank you, @chengfang.

bstansberry:
Superseded by #17742

PR #14867, opened 2021-11-01
[WFLY-15605] Native Jakarta variant of MP RSO

kabir:
https://issues.redhat.com/browse/WFLY-15605


ee-9/source-transform/microprofile/reactive-streams-operators-smallrye/extension/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+        <artifactId>wildfly-ee-9-source-transform-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>26.0.0.Beta1-SNAPSHOT</version>
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>wildfly-microprofile-reactive-streams-operators-jakarta</artifactId>
```
bstansberry, line 38: @kabir I don't think this module is necessary. Per https://docs.google.com/spreadsheets/d/1TekfyRb2UBCLqsPQ83WTg8XOw9qwuHVSh-xk2VEdFvk/edit#gid=1390360814, the artifact produced by the regular one isn't get transformed, so it seems to be fine as-is.

bstansberry (approved):

PR #14869, opened 2021-11-02
[WFLY-15475] Fix "Weld is not initialized yet"

jasondlee:
https://issues.redhat.com/browse/WFLY-15475

Expose otel-api module only if CDI is available
Add test to verify fix

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetrySetupTask.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetrySetupTask.java
@@ -11,7 +11,6 @@ import org.jboss.as.test.shared.ServerReload;
 import org.jboss.dmr.ModelNode;
 
 public class OpenTelemetrySetupTask implements ServerSetupTask {
-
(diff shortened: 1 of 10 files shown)


testsuite/integration/basic/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetrySetupTask.java
```diff
@@ -11,30 +11,38 @@
 import org.jboss.dmr.ModelNode;
 
 public class OpenTelemetrySetupTask implements ServerSetupTask {
+    private static boolean initialized = false;
+    private static int count = 2; // Ugly hack to prevent multiple reloads
```
bstansberry, line 15: If this is just to avoid reloads, it adds complexity and fragility. Test classes should clean up after themselves so the next test is starting from a known config, and not count on some other test class to do it at some later point.

jasondlee: It is and it does, but if you're ok with the reloads, I'll yank this like I'm starting a mower.

jasondlee: Pushed.

bstansberry (approved):

PR #14886, opened 2021-11-08
[WFLY-15640]: Upgrade WildFly Galleon plugins to 5.2.5.Final.

ehsavoie:
* Upgrading the galleon plugins to 5.2.5.Final
* fixing the feature pack to define the version of the cli artefact
with the classifier 'client'.

Jira: https://issues.redhat.com/browse/WFLY-15640

### DIFF
--- a/servlet-feature-pack/common/pom.xml
+++ b/servlet-feature-pack/common/pom.xml
@@ -300,6 +300,18 @@
         </dependency>
 
         <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-cli</artifactId>
+            <classifier>client</classifier>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
             <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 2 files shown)


servlet-feature-pack/common/pom.xml
```diff
@@ -299,6 +299,18 @@
...
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-cli</artifactId>
+            <classifier>client</classifier>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
bstansberry, line 312: @ehsavoie @jfdenise I'm a bit concerned this was necessary. Will the 5.2.5 plugin work properly with old feature packs that didn't do this?

If this is just a case of updating the FP to follow current best practices I'm not concerned.

jfdenise: @bstansberry the change in 5.2.5 impacts provisioning time. Older feature-packs depend on older versions of the plugin so, during provisioning the older version is used. New WildFly feature-packs that depend on 5.2.5 need this fix.

bstansberry: Thanks, Jeff.

PR #14890, opened 2021-11-09
[WFLY-10862] Log warning when MP Health reports DOWN

xstefank:
https://issues.redhat.com/browse/WFLY-10862

### DIFF
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
@@ -166,7 +166,13 @@ public class MicroProfileHealthReporter {
         builder.add("status", status.toString());
         builder.add("checks", checkResults);
 
-        return new SmallRyeHealth(builder.build());
+        JsonObject build = builder.build();
+
+        if (status.equals(HealthCheckResponse.Status.DOWN)) {
+            MicroProfileHealthLogger.LOGGER.healthDownStatus(build.toString());
+        }
+
+        return new SmallRyeHealth(build);
     }
 
     private HealthCheckResponse.Status processChecks(Map<HealthCheck, ClassLoader> checks, JsonArrayBuilder results, HealthCheckResponse.Status status) {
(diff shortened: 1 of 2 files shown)


microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
```diff
@@ -166,7 +166,13 @@ private SmallRyeHealth getHealth(Map<HealthCheck, ClassLoader> serverChecks, Map
         builder.add("status", status.toString());
         builder.add("checks", checkResults);
 
-        return new SmallRyeHealth(builder.build());
+        JsonObject build = builder.build();
+
+        if (status.equals(HealthCheckResponse.Status.DOWN)) {
+            MicroProfileHealthLogger.LOGGER.healthDownStatus(build.toString());
```
darranl, line 172: Is this a one off log or does this get logged periodically whilst down?

xstefank: every time the health reports DOWN

xstefank:
test failures seems unrelated

xstefank:
@darranl @bstansberry ping on this one please

darranl:
@jmesnil / @jamezp Do either of you have any views on the message being logged more than once?

In the past we would try and only log a warning once for a specific event and avoid spamming the logs.  But I don't know if due to the nature of health and how logs are handled in cloud use cases if a repeated log is more appropriate?

jamezp:
I've got no real strong opinion, but I will say if we only log it once and it comes up and goes back down we wouldn't know. I can understand while it's down it will be spammy though.

darranl:
So if it came up and went down again I think that would justify a second log - we would not want to loose that it has gone again.

If no strong objections maybe we merge as is and revisit if we need to reduce the number of messages logged.

jamezp:
That sounds reasonable to me @darranl

xstefank:
@darranl we have this behavior also in Quarkus and integrated in SR health (WFLY just doesn't use it). It is important to keep it for every call if you want to match container restarts to the actual health-related responses. I see a use case if you are trying to determine if the pod was taken down because of something internal in the container or something else.

darranl:
Lets run CI once more and get this merged

bstansberry:
Thanks, @xstefank

PR #14891, opened 2021-11-09
WFLY-15634 Upgrade wildfly-transaction-client from 1.1.14.Final to 2.…

chengfang:
…0.0.Final

https://issues.redhat.com/browse/WFLY-15634

To make wildfly-transaction-client (1.1.14->2.0.0) and wildfly-transaction-client-jakarta both on version 2.0.0.Final.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,7 +500,7 @@
         <version.org.wildfly.http-client>1.1.8.Final</version.org.wildfly.http-client>
         <version.org.wildfly.naming-client>1.0.14.Final</version.org.wildfly.naming-client>
         <version.org.wildfly.security.elytron>1.17.0.Final</version.org.wildfly.security.elytron>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -500,7 +500,7 @@
         <version.org.wildfly.http-client>1.1.8.Final</version.org.wildfly.http-client>
         <version.org.wildfly.naming-client>1.0.14.Final</version.org.wildfly.naming-client>
         <version.org.wildfly.security.elytron>1.17.0.Final</version.org.wildfly.security.elytron>
-        <version.org.wildfly.transaction.client>1.1.14.Final</version.org.wildfly.transaction.client>
+        <version.org.wildfly.transaction.client>2.0.0.Final</version.org.wildfly.transaction.client>
```
bstansberry, line 503: This can be removed now; it's only needed if you're overriding the value from the root pom.

bstansberry (approved):

PR #14895, opened 2021-11-12
[WFLY-15652] load InitialContextFactory class within doPrivileged block

istudens:
https://issues.redhat.com/browse/WFLY-15652

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/InitialContext.java
+++ b/naming/src/main/java/org/jboss/as/naming/InitialContext.java
@@ -22,6 +22,9 @@
 
 package org.jboss.as.naming;
 
+import java.lang.reflect.Constructor;
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Hashtable;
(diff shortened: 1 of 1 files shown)


istudens:
@darranl could you review this please? see a comment in the jira. thanks

naming/src/main/java/org/jboss/as/naming/InitialContext.java
```diff
@@ -112,8 +115,17 @@ protected Context getDefaultInitCtx() throws NamingException {
             } else {
                 final ClassLoader classLoader = WildFlySecurityManager.getCurrentContextClassLoaderPrivileged();
                 try {
-                    final Class<?> factoryClass = Class.forName(factoryClassName, true, classLoader);
-                    defaultInitCtx = ((javax.naming.spi.InitialContextFactory)factoryClass.newInstance()).getInitialContext(myProps);
+                    final Constructor<javax.naming.spi.InitialContextFactory> factoryClassConstructor;
+                    if (WildFlySecurityManager.isChecking()) {
+                        factoryClassConstructor = AccessController.doPrivileged((PrivilegedExceptionAction<Constructor<javax.naming.spi.InitialContextFactory>>) () -> {
+                            final Class<?> factoryClass = Class.forName(factoryClassName, false, classLoader);
```
darranl, line 121: It is not enough to request changes on this PR but IMO I find it better to split all of the common code out, in this case it could just be a static method.

Where code like this is duplicated there is a risk in the future that the two get out of sync so bugs only occur in one case or the other.

darranl:
Regarding the actual changes, I have some concerns that this may not be appropriate as this now means the caller can load any classes without their own protection domain being included in the permission check - I think the existing behaviour that the callers protections domain is needed for loading classes specified in the properties is correct.

PR #14907, opened 2021-11-17
[WFLY-15673] Upgrade jboss-openjdk-orb to 8.1.6.Final

tadamski:
Upgrade in order to prepare for Jakarta transformation PR + fix of WFLY-15345


tadamski:
This upgrade introduced a bug. I will fix it and open another one.

PR #14910, opened 2021-11-18
WFLY-15583 Adjust sql statements during initialization of DatabaseTim…

chengfang:
…erPersistence

https://issues.redhat.com/browse/WFLY-15583


bstansberry (approved):
LGTM.

Note that the JIRA says "a database type, once determined, is *unlikely* to change during the life time of a DatabaseTimerPersistence", but AFAICT the 'database' field *can't* change unless the service is started and stopped again, in which case sql gets repopulated. Which is good.

PR #14928, opened 2021-11-22
[WFLY-15719] Disabling ScriptAssertTestCase on JDK15+ because JavaScript engine have been removed since JDK15 including

ropalka:
https://issues.redhat.com/browse/WFLY-15719

Disabling ScriptAssertTestCase on JDK15+ because JavaScript engine have been removed since JDK version 15

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/beanvalidation/hibernate/scriptassert/ScriptAssertTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/beanvalidation/hibernate/scriptassert/ScriptAssertTestCase.java
@@ -31,6 +31,7 @@ import javax.validation.Validator;
 
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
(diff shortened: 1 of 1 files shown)


darranl:
Do we need something somewhere to track lost and missing functionality in case an alternative is needed?

ropalka:
How was it handled in Hibernate @scottmarlow @gunnarmorling that Java script engine is gone since JDK15?

gunnarmorling:
Hey @ropalka, the best person to ask about this would be @gsmet. I haven't been involved with Hibernate Validator for several years. That said, you can add the Nashorn dependency explicitly now that it has been removed from the JDK.

gsmet:
Yes, that would be me :).

In HV, we use groovy-jsr223 for this purpose: https://github.com/hibernate/hibernate-validator/blob/main/engine/pom.xml#L132-L136 . But it might require some work to make it available in WildFly.

ropalka:
According to provided information I believe excluding it on JDK15+ is sufficient @bstansberry WDYT?

darranl (approved):
I think lets proceed on this one.

PR #14930, opened 2021-11-22
WFLY-15713 Remove redundant null checks in TimerHandleImpl constructor.

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-15713

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
@@ -2284,29 +2284,29 @@ public interface EjbLogger extends BasicLogger {
 //    @Message(id = 334, value = "Could not deserialize info in timer ")
 //    RuntimeException failToDeserializeInfoInTimer(@Cause Throwable e);
 
-    /**
-     * Creates an exception indicating the Id cannot be null
-     *
-     * @return an {@link IllegalArgumentException} for the error.
-     */
-    @Message(id = 335, value = "Id cannot be null")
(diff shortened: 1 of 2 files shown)


RanabirChakraborty:
@chengfang can you please review once :)

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -2284,30 +2284,6 @@
...
-
-    /**
-     * Creates an exception indicating the timer service cannot be null
-     *
-     * @return an {@link IllegalArgumentException} for the error.
-     */
-    @Message(id = 337, value = "Timer service cannot be null")
-    IllegalArgumentException timerServiceIsNull();
-
     /**
```
chengfang, line 2287: We should just comment out these methods to make it easy to see their history and not to accidentally reuse these ids.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerHandleImpl.java
```diff
@@ -69,16 +69,6 @@
      * @throws IllegalArgumentException If either of the passed parameters is null
      */
     public TimerHandleImpl(final String id, final String timedObjectId, final TimerServiceImpl service) throws IllegalArgumentException {
```
chengfang, line 71: Can you also remove the IllegalArgumentException from method signature, and javadoc?

RanabirChakraborty:
Thanks! did the changes.

chengfang (approved):

PR #14933, opened 2021-11-23
[WFLY-15449] Move WildFly Preview to native jakarta Taglibs

ropalka:
https://issues.redhat.com/browse/WFLY-15449

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -428,7 +428,7 @@
                                 <exclude>com.fasterxml.jackson.jaxrs:jackson-jaxrs-base\z</exclude>
                                 <exclude>com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider\z</exclude>
                                 <exclude>com.fasterxml.jackson.module:jackson-module-jaxb-annotations\z</exclude>
-								<!-- end JPMS or ISGi metadata excludes -->
+                                <!-- end JPMS or OSGi metadata excludes -->
                             </jakarta-transform-excluded-artifacts>
                         </configuration>
                     </execution>
@@ -736,24 +736,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.taglibs</groupId>
-            <artifactId>taglibs-standard-spec</artifactId>
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.taglibs</groupId>
-            <artifactId>taglibs-standard-impl</artifactId>
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.taglibs</groupId>
-            <artifactId>taglibs-standard-compat</artifactId>
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
             <groupId>jakarta.json.bind</groupId>
             <artifactId>jakarta.json.bind-api</artifactId>
             <scope>provided</scope>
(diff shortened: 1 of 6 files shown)


ropalka:
@fl4via please review

darranl:
@fl4via / @fjuma Would it be possible to please get some reviews on this PR?

darranl:
Sorry I see the duplicate ping from @ropalka I have a few PRs open at the moment.

fjuma (approved):

ee-9/feature-pack/pom.xml
```diff
@@ -428,7 +428,10 @@
                                 <exclude>com.fasterxml.jackson.jaxrs:jackson-jaxrs-base\z</exclude>
                                 <exclude>com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider\z</exclude>
                                 <exclude>com.fasterxml.jackson.module:jackson-module-jaxb-annotations\z</exclude>
-								<!-- end JPMS or ISGi metadata excludes -->
+                                <!-- end JPMS or OSGi metadata excludes -->
+                                <exclude>org.apache.taglibs:taglibs-standard-compat\z</exclude>
+                                <exclude>org.apache.taglibs:taglibs-standard-impl\z</exclude>
+                                <exclude>org.apache.taglibs:taglibs-standard-spec\z</exclude>
```
bstansberry, line 434: @ropalka This shouldn't be needed. I think the reason you're seeing these still getting transformed is because these 3 are listed as deps in ee-feature-pack/common/pom.xml and and in the ee-feature-pack-common-licenses.xml in that module. But I think having those there is wrong, as they are already included in servlet-feature-pack/ee-8-api. Having them in ee-feature-pack/common as well is duplication.

We don't bring servlet-feature-pack/ee-8-api stuff into this feature-pack so removing it from ee-feature-pack/common should make it disappear from the build of this f-p processing.

If that didn't work for some reason, then instead of adding the above lines the thing to do would be to add excludes to the org.wildfly:wildfly-ee-feature-pack-common dependency element in this pom. We should only use the exclude lines here for things we want incorporated in the f-p but know they don't need to be transformed.

ee-9/feature-pack/src/main/resources/modules/system/layers/base/jakarta/servlet/jstl/api/main/module.xml
```diff
@@ -25,21 +25,17 @@
...
         <module name="java.logging"/>
         <module name="java.naming"/>
         <module name="java.sql"/>
-        <!-- org.xml.sax -->
-        <!--WFLY-14219 Remove deprecated <module name="javax.api"/> -->
-        <module name="javax.servlet.api"/>
-        <module name="javax.servlet.jsp.api"/>
-        <module name="org.apache.xalan"/>
+        <module name="jakarta.servlet.api"/>
+        <module name="jakarta.servlet.jsp.api"/>
```
bstansberry, line 38: For now please use the javax.* alias modules for these 2 deps. I'm trying to keep the use of the javax aliases consistent until something happens that forces a change or we decide to convert everything to jakarta.

bstansberry (changes requested):

ee-9/feature-pack/src/main/resources/modules/system/layers/base/jakarta/servlet/jstl/api/main/module.xml
ropalka: Fixed @bstansberry .

ee-9/feature-pack/pom.xml
ropalka: Fixed @bstansberry - thanks for the feedback.

bstansberry (approved):

fl4via (approved):

PR #14936, opened 2021-11-23
[WFLY-15732] / [WFLY-15356] Remove use of security-plugins from Undertow subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-15732


darranl:
Hello @fl4via could you please review this one, hopefully helps with less code to transform in the Undertow subsystem as well as this removes some Jakarta API implementations.

@pferraro Could you please take a look at the clustering code I have deleted and let me know if I gone far enough / too far?

There will be some further clean up later when it comes to making application security domain resources optional, but for now I just want to eliminate the final pieces of code still calling the legacy APIs and report an error if they are called.

ropalka (approved):
Since this is about code removal I was able to review this PR.

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/sso/UndertowSecuritySerializationContextInitializer.java
```diff
@@ -1,44 +0,0 @@
...
-import org.wildfly.clustering.marshalling.protostream.AbstractSerializationContextInitializer;
-
-/**
- * @author Paul Ferraro
- */
-@MetaInfServices(SerializationContextInitializer.class)
-public class UndertowSecuritySerializationContextInitializer extends AbstractSerializationContextInitializer {
-
-    public UndertowSecuritySerializationContextInitializer() {
-        super("io.undertow.security.api.proto");
```
pferraro, line 37: The file `clustering/web/undertow/src/main/resources/io.undertow.security.api.proto` should also be removed.

clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/sso/UndertowSingleSignOnManagementProvider.java
```diff
@@ -1,59 +0,0 @@
...
-import org.wildfly.clustering.web.container.HostSingleSignOnManagementProvider;
-import org.wildfly.clustering.web.sso.DistributableSSOManagementProvider;
-import org.wildfly.clustering.web.sso.LegacySSOManagementProviderFactory;
-import org.wildfly.extension.undertow.session.SessionManagementProviderFactory;
-
-/**
- * {@link SessionManagementProviderFactory} for Undertow using either the {@link DistributableSSOManagementProvider} for the given host, the default provider, or a legacy provider.
- * @author Paul Ferraro
- */
-@MetaInfServices(HostSingleSignOnManagementProvider.class)
```
pferraro, line 42: While the distributed HostSingleSignOnManagementProvider was removed (by deleting this class and its supporting classes), the org.wildfly.extension.undertow.HostSingleSignOnServiceHandler class will still configure an in-memory single sign-on provider.
Should this not also be removed?

pferraro (changes requested):
Is the intention to deprecate the resource defined by HostSingleSignOnDefinition?
This resource is still functional, and this PR only seems to remove clustered behavior (the local implementation is still intact).

PR #14938, opened 2021-11-24
[WFLY-15739] Set java.security.manager system property value to "allow" explicitly

ropalka:
https://issues.redhat.com/browse/WFLY-15739

Set java.security.manager system property value to "allow" in order to be able to install security manager on JDK 18+.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -202,7 +202,7 @@
         <!-- Surefire args -->
         <surefire.extra.args></surefire.extra.args>
         <surefire.jpda.args></surefire.jpda.args>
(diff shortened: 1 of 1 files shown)


darranl:
Do we need to consider our start up scripts for this as well?

ropalka:
Yes, I am working on these scripts @darranl . I am also not sure if this proposal PR will work on JDKs prior to JDK17.

ropalka:
Please consider this PR experimental @darranl  - I'm curious about the CI results.

ropalka:
JDK8 supports only empty string "" and "default" special string values. So I expect JDK8 to fail.

ropalka:
You can remove the "hold" label @darranl . It is ready for official review.
Setting -Djava.security.manager=allow works also on JDKs prior to JDK17.

PR #14941, opened 2021-11-24
[WFLY-15734] / [WFLY-15356] Remove security-plugins dependency from EJB3.

darranl:
https://issues.redhat.com/browse/WFLY-15734

This PR is just removing some code which should no longer be in use for legacy security integration so we can delete the legacy security modules and get the PicketBox dependency removed.


darranl:
@tadamski / @chengfang Would it be possible to please get a review on this one?

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -3250,4 +3250,8 @@
     @LogMessage(level = WARN)
     @Message(id = 529, value = "Failed to retrieve info from database for timer: %s")
     void failedToRetrieveTimerInfo(final TimerImpl timer, @Cause Exception e);
+
+    @Message(id = 530, value = "The deployment is configured to use legacy security which is no longer supported.")
+    IllegalStateException legacySecurityUnsupported();
+
```
chengfang, line 3256: is it possible to include some context info like the security domain name used?

darranl: Thanks @chengfang I will have a look if I can add this.

darranl:
@chengfang I have added the name of the security domain to the log message, if you are Ok I think I can get this through tomorrow.

chengfang (approved):

PR #14942, opened 2021-11-24
[WFLY-15733] / [WFLY-15356] Remove security-plugins dependency from messaging.

darranl:
https://issues.redhat.com/browse/WFLY-15733

### DIFF
--- a/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/apache/activemq/artemis/main/module.xml
+++ b/ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/apache/activemq/artemis/main/module.xml
@@ -63,7 +63,6 @@
         <!--this reverse dependency is here so integration classes in the AS code base can be instantiated by Artemis-->
         <module name="org.wildfly.extension.messaging-activemq" services="import"/>
         <module name="org.picketbox" optional="true"/>
-        <module name="org.jboss.as.security-plugins"/>
         <!-- this optional dependency is required to be able to use this module from a jms-bridge to connect to a remote
              WildFly server [AS7-6549] -->
         <module name="org.wildfly.naming-client" optional="true"  services="import"/>
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
@@ -21,6 +21,8 @@
  */
 package org.wildfly.extension.messaging.activemq;
 
+import static org.wildfly.extension.messaging.activemq.logging.MessagingLogger.ROOT_LOGGER;
+
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PATH;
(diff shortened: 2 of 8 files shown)


darranl:
@ehsavoie could you please give this one a review?

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
```diff
@@ -329,18 +329,12 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
                 ServiceName elytronDomainCapability = context.getCapabilityServiceName(ELYTRON_DOMAIN_CAPABILITY, elytronSecurityDomainModel.asString(), SecurityDomain.class);
                 elytronSecurityDomain = Optional.of(serviceBuilder.requires(elytronDomainCapability));
             } else {
-                // Add legacy security
-                String domain = SECURITY_DOMAIN.resolveModelAttribute(context, model).asString();
-                securityDomainContext = Optional.of(serviceBuilder.requires(SECURITY_DOMAIN_SERVICE.append(domain)));
-                // WFLY-6652 / WFLY-10292 this dependency ensures that Artemis will be able to destroy any queues created on behalf of a
-                // pooled-connection-factory client during server stop
-                serviceBuilder.requires(SECURITY_BOOTSTRAP_SERVICE);
+                throw ROOT_LOGGER.legacySecurityUnsupported();
```
ehsavoie, line 337: I think this is 'wrong' as it would require an elytron domain to be always configured. I think we should incorporate https://github.com/wildfly/wildfly/pull/14790 in there.

ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/apache/activemq/artemis/main/module.xml
```diff
@@ -63,7 +63,6 @@
         <!--this reverse dependency is here so integration classes in the AS code base can be instantiated by Artemis-->
         <module name="org.wildfly.extension.messaging-activemq" services="import"/>
         <module name="org.picketbox" optional="true"/>
```
ehsavoie, line 65: Why do we leave picketbox there ?

ehsavoie (changes requested):

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
darranl: +1 it is wrong, but no more wrong than the current codebase i.e. in WildFly 26 Beta if we configure to reach this section the service would not be available so I wanted to throw an error rather than quietly ignore and accidentally leave unsecured.

But yes I think we need to get your PR in and then this one can be rebased and likely eliminate this error.

ee-9/feature-pack/src/main/resources/modules/system/layers/base/org/apache/activemq/artemis/main/module.xml
darranl: The PicketBox dependency will be removed in a follow up PR.

ehsavoie (approved):

PR #14945, opened 2021-11-24
[WFLY-15356] Remove the wildfly-security-plugins module

darranl:
https://issues.redhat.com/browse/WFLY-15356

Submitting as a draft as the individual sub tasks are being reviewed independently.  Once they are all approved this one is ready.

PR #14948, opened 2021-11-24
[WFLY-15747] / [WFLY-13889] Remove PicketBox dependency from iiop-openjdk.

darranl:
Raising as a DRAFT as this depends on the following PR:
    https://github.com/wildfly/wildfly/pull/14945

This removes the dependency on PicketBox and removes the legacy code which was using the PicketBox APIs.


darranl:
@tadamski would it be possible to please review the last commit in relation to iiop-openjdk:
    https://github.com/wildfly/wildfly/pull/14948/commits/3c01a7a867736a92260d380986c52f9cab4b3f71

darranl:
Hi @tadamski if we can please get a review on the last commit on this one I think we can get this merged, this will also mean PicketBox will be out of the way for the Jakarta EE tranformation.

tadamski (changes requested):
hi @darranl LGTM but I believe SECURITY_CLIENT and  SECURITY_IDENTITY would have to be removed from IIOPInitializer as well.

darranl:
Superseded by https://github.com/wildfly/wildfly/pull/15136

PR #14952, opened 2021-11-25
WFLY-15525 JBeret: Make it possible to limit number of records retrieved via a JDBC store

TomasHofman:
https://issues.redhat.com/browse/WFLY-15525
https://issues.redhat.com/browse/EAP7-1821

### DIFF
--- a/batch-jberet/pom.xml
+++ b/batch-jberet/pom.xml
@@ -76,6 +76,11 @@
             <artifactId>jboss-msc</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>wildfly-weld-spi</artifactId>
         </dependency>
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchDeploymentDescriptorParser_1_0.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchDeploymentDescriptorParser_1_0.java
@@ -53,6 +53,7 @@ public class BatchDeploymentDescriptorParser_1_0 implements XMLStreamConstants, 
         String dataSourceName = null;
         String jobExecutorName = null;
         Boolean restartJobsOnResume = null;
+        Integer executionRecordsLimit = null;
         boolean empty = true;
 
         while (reader.hasNext() && reader.nextTag() != END_ELEMENT) {
@@ -61,6 +62,8 @@ public class BatchDeploymentDescriptorParser_1_0 implements XMLStreamConstants, 
 
             // Process the job repository
             if (element == Element.JOB_REPOSITORY) {
+                executionRecordsLimit = parseExecutionRecordsLimit(reader);
+
                 // Only one repository can be defined
                 if (jobRepository != null || jobRepositoryName != null) {
                     BatchLogger.LOGGER.multipleJobRepositoriesFound();
@@ -104,11 +107,19 @@ public class BatchDeploymentDescriptorParser_1_0 implements XMLStreamConstants, 
             BatchLogger.LOGGER.debugf("An empty batch element in the deployment descriptor was found for %s.", deploymentUnit.getName());
             return null;
         }
-        return new BatchEnvironmentMetaData(jobRepository, jobRepositoryName, dataSourceName, jobExecutorName, restartJobsOnResume);
+        return new BatchEnvironmentMetaData(jobRepository, jobRepositoryName, dataSourceName, jobExecutorName,
+                restartJobsOnResume, executionRecordsLimit);
     }
 
     String parseJdbcJobRepository(final XMLExtendedStreamReader reader) throws XMLStreamException {
         throw ParseUtils.unexpectedElement(reader);
+    }
+
+    Integer parseExecutionRecordsLimit(final XMLExtendedStreamReader reader) throws XMLStreamException {
+        if (reader.getAttributeCount() > 0) {
+            throw ParseUtils.unexpectedAttribute(reader, 0);
+        }
+        return null;
     }
 
     static String readRequiredAttribute(final XMLExtendedStreamReader reader, final Attribute attribute) throws XMLStreamException {
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/job/repository/CommonAttributes.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/job/repository/CommonAttributes.java
@@ -1,3 +1,25 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
 package org.wildfly.extension.batch.jberet.job.repository;
 
 import org.jboss.as.controller.SimpleAttributeDefinition;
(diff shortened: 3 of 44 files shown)


chengfang:
`org.jboss.as.test.integration.batch.repository.JdbcRepositoryExecutionsLimitTestCase.` failed in some CI jobs, which looks related. @TomasHofman

chengfang (approved):

TomasHofman:
The remaining test failures in "Linux with sec manager" do not look related to this PR.

TomasHofman:
PR rebased.

TomasHofman:
There are still some reactive messaging tests consistently failing in the security manager job. I'm going to investigate locally, but I don't see how they are relevant to changes in the batch subsystem.

TomasHofman:
I able to reproduce the sec manager failures also locally, both with changes in this PR and without these changes. I was trying with following commit:

```
commit 210c43ea7b0fa77287668d9b9ba679daf91ce395 (HEAD -> main, upstream/main)
Merge: f4e7afd2c7 73f7ed236d
Author: Brian Stansberry <brian.stansberry@redhat.com>
Date:   Tue Jan 18 17:30:56 2022 -0600

    Merge pull request #15070 from boris-unckel/WFLY-14967_ee9_eclipselink_v2
    
    [WFLY-14967] Move WildFly Preview to a native Jakarta namespace varia…
```

So I would conclude that the failures are not related to this change.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchDeploymentDescriptorParser_1_0.java
```diff
@@ -61,6 +62,8 @@ public BatchEnvironmentMetaData parse(final XMLExtendedStreamReader reader, fina
 
             // Process the job repository
             if (element == Element.JOB_REPOSITORY) {
+                executionRecordsLimit = parseExecutionRecordsLimit(reader);
```
darranl, line 65: What is the reason for changing parsing in the 1.0 parser?

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/job/repository/CommonAttributes.java
```diff
@@ -0,0 +1,19 @@
+package org.wildfly.extension.batch.jberet.job.repository;
```
darranl, line 1: This is missing a copyright header.

darranl (changes requested):
I have added a couple of comments but I am also not seeing transformers or transformer testing.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchDeploymentDescriptorParser_1_0.java
TomasHofman: I made this change in order to make the `BatchDeploymentDescriptorParser_1_0` more extendable, to simplify the new `BatchDeploymentDescriptorParser_3_0` class. I would have to copy paste the whole `parse()` method otherwise. This way I just need to override the `parseExecutionRecordsLimit()` method in version 3 parser, and the rest of the code still lives in the original classes only. In version 1 the method doesn't accept any attributes and just returns `null`, so it doesn't add any new features to version 1.

The same approach was used in version 2 with `parseJdbcJobRepository()` method, so it seemed like a good idea to follow the approach.

darranl: Sorry yes I see you throw an error if called - please disregard my comment.

TomasHofman:
@darranl I added new commit with transformers and tests for them. Would you check that please? I'm not very well versed in transformers and there is not much docs on them. The transformer tests download some productized artifacts from EAP 7.4, is that OK? I think other subsystems do that too, but I'm not sure if it works for community members who don't have EAP repos in their `settings.xml`.

@chengfang when doing transformers for schema version 3, I also added transformer for version 2, where the `security-domain` attribute was added. It's probably not very important at this point, since version 2 schema is from EAP 7.0 times. Let me know if I should rather remove this.

chengfang:
From WFLY-15217 (https://issues.redhat.com/browse/WFLY-15217), batch transformer was dropped as part of [WFLY-15061](https://issues.redhat.com/browse/WFLY-15061) Reduce mixed domain support to WF 23 and later.

darranl:
This is the branch to support EAP 7.x so this is the branch that needs transformers at the time the subsystem is expanded - there is no porting activity.

TomasHofman:
OK, I updated the last commit to only include transformers for the latest schema change.

batch-jberet/pom.xml
```diff
@@ -75,6 +75,17 @@
...
             <artifactId>jboss-msc</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.picketbox</groupId>
+            <artifactId>picketbox</artifactId>
```
darranl, line 85: No new PicketBox dependencies can be added to the server.

darranl (dismissed):
It looks like a PicketBox dependency has crept in, that will need to be removed.

batch-jberet/pom.xml
TomasHofman: Ouch, that was wrongly resolved merge conflict. Should be fixed now.

batch-jberet/src/test/java/org/wildfly/extension/batch/jberet/JBeretSubsystemParsingTestCase.java
```diff
@@ -151,6 +164,42 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
...
+        KernelServicesBuilder builder = createKernelServicesBuilder(createAdditionalInitialization());
+        LegacyKernelServicesInitializer kernelServicesInitializer = builder.createLegacyKernelServicesBuilder(createAdditionalInitialization(), controllerVersion, subsystemModelVersion)
+                .addMavenResourceURL("org.wildfly.core:wildfly-threads:" + controllerVersion.getCoreVersion())
+                .dontPersistXml();
+        try {
+            BatchSubsystemExtension.class.getClassLoader().loadClass("javax" + ".batch.operations.JobStartException");
+            kernelServicesInitializer.addMavenResourceURL("org.jboss.eap:wildfly-batch-jberet:" + controllerVersion.getMavenGavVersion());
+        } catch (ClassNotFoundException e) {
+            kernelServicesInitializer.addMavenResourceURL("org.wildfly:wildfly-batch-jberet-jakarta:26.0.0.Final");
+        }
```
TomasHofman, line 194: I'm bit unsure about this piece.

The batch subsystem module (and this test class with it) is transformed into Jakarta-version subsystem with renamed package names (wildfly-batch-jberet -> wildfly-batch-jberet-jakarta).

In the original subsystem, I need to do 

```kernelServicesInitializer.addMavenResourceURL("org.jboss.eap:wildfly-batch-jberet:" + controllerVersion.getMavenGavVersion())```

when testing that the config can be transformed to previous EAP version.

This stops working when the test is run in the transformed "-jakarta" module, because the `org.jboss.eap:wildfly-batch-jberet` code would still use javax API, while it needs to use jakarta API.

So when this test is run in the transformed module, I need to include jakarta version of the subsystem instead:
```
kernelServicesInitializer.addMavenResourceURL("org.wildfly:wildfly-batch-jberet-jakarta:26.0.0.Final");
```
As JBoss EAP doesn't productize the jakarta version modules, I'm using "org.wildfly" groupId here.

Also, the only way that occurs to me how to detect that the test is run in "-jakarta" module is to try to load some javax.batch API class, and if that fails, I'm in the jakarta world.

Does that look alright? Is there better way to do this?

**Comment by TomasHofman** [comment]
All tests are passing after rebasing. Requested changes were done (picketbox dependency removed, transformer tests added).

**Comment by chengfang** [comment]
@TomasHofman  fyi, there is conflict caused by my other PR.

**Comment by TomasHofman** [comment]
@chengfang thanks for info. I updated this PR to use 1.3.12.Final, as you suggest. The CI will fail though until the jberet version is released.

File: batch-jberet/src/test/java/org/wildfly/extension/batch/jberet/JBeretSubsystemParsingTestCase.java (hunk 4 above)
Comment (bstansberry) [reply batch-jberet/src/test/java/org/wildfly/extension/batch/jberet/JBeretSubsystemParsingTestCase.java:194]: @TomasHofman what you have is ok; another way is to do 

if (JobStartException.class.getName().startsWith("jakarta.") { // i'm in jakarta land }

If this is working I'd say leave it as is though.

Comment (TomasHofman) [reply batch-jberet/src/test/java/org/wildfly/extension/batch/jberet/JBeretSubsystemParsingTestCase.java:194]: @bstansberry thanks for confirming! I'm keeping the current version then :+1:

**Comment by chengfang** [comment]
> @chengfang thanks for info. I updated this PR to use 1.3.12.Final, as you suggest. The CI will fail though until the jberet version is released.

jberet-core 1.3.12.Final and 2.0.4.Final (for ee-9/pom.xml) have been released. Ideally we should upgrade the jberet-core version for ee-9/pom.xml to 2.0.4.Final, to be consistent across 2 variants (I should've made it clear in my previous comment). @TomasHofman

**Comment by chengfang** [comment]
@TomasHofman test runs failed with the following error:
```
Caused by: org.jboss.arquillian.container.spi.client.container.LifecycleException: The java process starting the managed server exited unexpectedly with code [1]
```
Could it be caused by some test changes introduced? I've restarted a few test runs but still see the same errors.

TomasHofman:
I'm aware, I'm not sure yet. I will investigate it more tomorrow. I'm might have done something wrong during rebasing...

bstansberry (approved):

bstansberry:
Thanks @TomasHofman

PR #14953, opened 2021-11-25
[WFLY-15754] Upgrade to SmallRye  Reactive Messaging 3.13

kabir:
https://issues.redhat.com/browse/WFLY-15754
https://issues.redhat.com/browse/WFLY-15780

### DIFF
--- a/microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/deployment/ReactiveMessagingDependencyProcessor.java
+++ b/microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/deployment/ReactiveMessagingDependencyProcessor.java
@@ -22,6 +22,7 @@
 
 package org.wildfly.extension.microprofile.reactive.messaging.deployment;
 
+import static org.jboss.as.weld.Capabilities.WELD_CAPABILITY_NAME;
 import static org.wildfly.microprofile.reactive.messaging.common.ReactiveMessagingAttachments.IS_REACTIVE_MESSAGING_DEPLOYMENT;
 
 import java.security.PrivilegedAction;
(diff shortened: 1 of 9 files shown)


microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/deployment/ReactiveMessagingDependencyProcessor.java
```diff
@@ -184,4 +192,13 @@ private boolean allowExperimentalAnnotations() throws DeploymentUnitProcessingEx
         }
         return experimental;
     }
+
+    @SuppressWarnings("deprecation")
+    private void ignorePrecalulatedJandex(DeploymentUnit deploymentUnit, String... modules) {
+        final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
+        if (support.hasCapability(WELD_CAPABILITY_NAME)) {
+            WeldCapability weld = support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class).get();
```
bstansberry, line 200: That's a clever way to avoid dealing with the NoSuchCapabilityException!

bstansberry:
@manovotn FYI re the last commit, which is a workaround for an issue where we bytecode transform artifacts for EE 9+ / WildFly Preview, but that results in the jandex.idx in the artifact being out of sync.

bstansberry (approved):

microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/deployment/ReactiveMessagingDependencyProcessor.java
kabir: Copied and pasted from JSF :-D

manovotn:
> @manovotn FYI re the last commit, which is a workaround for an issue where we bytecode transform artifacts for EE 9+ / WildFly Preview, but that results in the jandex.idx in the artifact being out of sync.

What can I say... can't wait for preview to full switch to `jakarta` namespace :-)

PR #14959, opened 2021-11-25
[WFLY-15758] / [WFLY-13889] Remove the appclient dependency on PicketBox.

darranl:
Submitting as a draft as this requires the following PR to be approved:
    https://github.com/wildfly/wildfly/pull/14945

Only the last commit in this PR needs to be reviewed, the others are being reviewed elsewhere.


darranl:
@bstansberry would you be able to give the last commit a review on this one so we can merge?  appclient itself by now should only use the dependency if legacy security is configured but that is now disabled.

bstansberry (approved):

PR #14968, opened 2021-11-26
[WFLY-15328] Split most source out of the ee-9/feature-pack module

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15328


ee-9/common-microprofile/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ee-9-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>26.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>wildfly-preview-feature-pack-common-microprofile</artifactId>
```
yersan, line 39: This probably would bet better if we name. it differently. Instead of `common-microprofile`, maybe `common-full`. 

This module is here just to supply the transformed modules that don't belong to the WildFly EE galleon pack and only belong to WildFly Full Galleon Pack. We could have other modules besides the MicroProfile ones.

ee-9/galleon-content-microprofile/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ee-9-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>26.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>wildfly-preview-feature-pack-content-microprofile</artifactId>
```
yersan, line 39: This probably would be better if we name it differently. Instead of `content-microprofile`, maybe `content-full`.

This module is here just to supply the Galleon Full content that overrides the Galleon EE content. We could have content that is not MicroProfile specific, for example, at the moment of writing, we have here OIDC stuff which is not supplied by default by the EE Galleon Pack and it is not MicroProfile Specific.

ee-9/common-microprofile/pom.xml
bstansberry: I think the name is ok for now. There's no obviously better one that comes to mind and I don't think we should block on trying to come up with one right now, since this is an internal detail.

ee-9/galleon-content-microprofile/pom.xml
bstansberry: Same as above; seems ok for now.

Note that IIRC the oidc stuff will end up in wildfly-ee at some point.

bstansberry (approved):
Great work, @yersan.

ee-9/galleon-content-microprofile/pom.xml
yersan: That is my impression too, I'm fine if we leave it as it is. Indeed, I think I was getting the relationships incorrectly. After discussions on https://issues.redhat.com/browse/WFLY-15752 I understand the general intention is:

wildfly-ee galleon pack -> known as "full", from "WildFly Full"
wildfly galleon pack-> known as microprofile, from "WildFly Micropofile"

bstansberry: Yes, but TBH the word 'full' for wildfly-ee doesn't really fit any more either, since wildfly-servlet isn't really relevant any more and wildfly-ee isn't the "full distribution" of WF any more either, it's just a subset.  That 'full' word in things like standalone-full.xml comes from distinguishing the full EE platform vs the web profile. (But wildfly-servlet isn't EE web profile compliant either!) 

Your content-full idea isn't a bad idea; only negative is that the word 'full' now has such a confused meaning. Otherwise that would be a great name.

bstansberry:
@yersan As discussed in https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/6HGMHVTINH74UIFSAXMFQAX5LVVCMNRN/ we're trying to wrap up changes by this Thursday.  Since the big benefit from this comes when we no longer have any artifacts being transformed by Galleon, and we won't be at that point for WF 26, in our release planning discussion this AM the consensus was this should wait for WF 27, trying to get it merged basically as the first change after we update the pom versions following the WF 26 tag.

ee-9/galleon-content-microprofile/pom.xml
yersan: Ok, got it, I'll resolve the conflicts to get it prepared

darranl:
@bstansberry I am going to bundle up this one assuming your review still stands - if you can please give a double check and I can get this one in.

PR #14978, opened 2021-11-29
[WFLY-15767] Update OpenTelemetry to 1.9.1

jasondlee:
https://issues.redhat.com/browse/WFLY-15767

### DIFF
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
@@ -32,7 +32,6 @@
         <artifact name="${io.opentelemetry:opentelemetry-exporter-jaeger}"/>
         <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-common}"/>
         <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-trace}"/>
-        <artifact name="${io.opentelemetry:opentelemetry-proto}"/>
     </resources>
 
     <dependencies>
--- a/ee-9/source-transform/observability/opentelemetry/pom.xml
+++ b/ee-9/source-transform/observability/opentelemetry/pom.xml
@@ -90,10 +90,6 @@
         </dependency>
         <dependency>
             <groupId>io.opentelemetry</groupId>
-            <artifactId>opentelemetry-proto</artifactId>
-        </dependency>
-        <dependency>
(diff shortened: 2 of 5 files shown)


ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/trace/main/module.xml
```diff
@@ -32,7 +32,6 @@
         <artifact name="${io.opentelemetry:opentelemetry-exporter-jaeger}"/>
         <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-common}"/>
         <artifact name="${io.opentelemetry:opentelemetry-exporter-otlp-trace}"/>
-        <artifact name="${io.opentelemetry:opentelemetry-proto}"/>
```
bstansberry, line 35: Are we going to replace this with something from https://github.com/open-telemetry/opentelemetry-proto-java? The OTEL v 1.7 release notes say something about that.

jasondlee: It does, but we don't need it (no idea why it was added, tbh). I started the server and ran manual tests with my otel demo app. Traces were recorded and exported to Jaeger as expected, and all tests in basic/ and manualmode/ are passing with no errors/complaints in the logs, so I don't think we need it all.

bstansberry: @jasondlee Can users somehow trigger use of this 'Jaeger gRPC exporter' thing that this lib seems to back, i.e. via the APIs we expose? Or is that something our subsystem would have to do for them?

jasondlee: The export of traces is done automatically by otel itself, based on some config params. It pushes the data to the external collecter (Jaeger, Otel Collector, etc) without any interaction from the user. Is that what you're asking?

bstansberry: @jasondlee coming back to this; sorry for the slow reply

What I was trying to get at was a distinction between what we test vs what a user can do (without obvious hacking). If users could do things without obvious hacking, then we have to think harder about taking those things away. But if there's no API we expose to application code that lets that code control this, and our subsystem has a clear configuration model and that provides no way to control this, then there's no concern.

PR #14985, opened 2021-12-01
[WFLY-15775]:Add xerces module dependency to webservice deployment

jimma:
https://issues.redhat.com/browse/WFLY-15775

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSDependenciesProcessor.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSDependenciesProcessor.java
@@ -47,6 +47,7 @@ public final class WSDependenciesProcessor implements DeploymentUnitProcessor {
             ModuleIdentifier.create("javax.xml.soap.api"),
             ModuleIdentifier.create("javax.xml.ws.api")
     };
+    public static final ModuleIdentifier XERCES_IMPL = ModuleIdentifier.create("org.apache.xerces");
 
     private final boolean addJBossWSDependencies;
 
@@ -65,5 +66,8 @@ public final class WSDependenciesProcessor implements DeploymentUnitProcessor {
         for(ModuleIdentifier api : JAVAEE_APIS) {
             moduleSpec.addSystemDependency(new ModuleDependency(moduleLoader, api, false, false, true, false));
         }
+        //After jboss modules 2.0, the xercers module is not added
+        moduleSpec.addSystemDependency(new ModuleDependency(moduleLoader, XERCES_IMPL, true, false, true, false));
+
     }
 }


webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSDependenciesProcessor.java
```diff
@@ -65,5 +66,8 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
         for(ModuleIdentifier api : JAVAEE_APIS) {
             moduleSpec.addSystemDependency(new ModuleDependency(moduleLoader, api, false, false, true, false));
         }
+        //After jboss modules 2.0, the xercers module is not added
+        moduleSpec.addSystemDependency(new ModuleDependency(moduleLoader, XERCES_IMPL, true, false, true, false));
```
jamezp, line 70: My only concern here is that this adds Xerces to a deployment. AIUI this means that any deployment which is processed with the processor cannot use the JDK JAXP or other implementation.

bstansberry: @jimma  @ropalka What behavior difference does this introduce vs what we had in WF 25 and earlier? That is, were some things possible in terms of user control over the JAXP impl that now won't be?

I'm interested in how things would differ in two cases:

1) Users who are using webservices.

2) Users who are not.

If, in WF 25 and earlier, trying to use some different JAXP with a webservices app just wouldn't work because CXF needs xerces, then adding this dep to webservices apps doesn't prevent something that used to work from still working *for those apps*.

For non-WS apps, if this DUP adding this dep is a breaking something that formerly worked, they can remove WS from the config or disable it from a deployment descriptor. That's not ideal, but is a valid workaround.

jimma: @jamezp @bstansberry  From what I debug and the code reading, adding this dependency to ws deployment is the same behavior as WF 25  and earlier.  In WF25, the xerces is the default global dependency and Serviceloader can load xerces without explicitly declare this dependency.  I remembered the xerces under /WEB-INF/lib  can be loaded to override the global default jaxp implementation in WF25 because of the classloader hierarchy.  And this is still the case after this adding xerces dependency to ws deployment change.  @ropalka , please correct me if I am wrong.

ropalka: This proposal PR @jimma  goes against all our effort: https://issues.redhat.com/browse/EAP7-1752 .
We want and need to upgrade JAXP API versions to 1.5 or 1.6 respectively (version depends on JDK version).

To quick answer your questions @bstansberry :
 * Q: Were some things possible in terms of user control over the JAXP impl that now won't be?
 * A: https://issues.redhat.com/browse/MODULES-407?focusedCommentId=19335517&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19335517
 
 * Q: 1. Users who are using webservices
 * A: This PR aligns current WF upstream with WF 25 and earlier. I.e. no change in behavior.
 
 * Q:  2. Users who are not.
 * A: Either JDK JAXP implementation will be used (if none is specified in DU dependencies)
        or configured JDK JAXP implementation will be used if DU dependencies reference it. This is change in behavior.

The purpose of EAP7-1752 in short is: 
 * eliminate Xalan & Xerces dependencies where possible (ideally everywhere)
 * allow WildFly users to use JAXP configuration in "standardized way"
 * migrate WildFly to most recent JAXP APIs and implementations

bstansberry: Thanks @ropalka; that's very helpful.

AIUI from what @jimma has told me, the problem with what we are trying to do with JAXP 1.5/1.6 support is that it breaks our webservices stack, because CXF doesn't work properly without xerces. And it might be a while before that is corrected.

Breaking our webservices stack isn't an acceptable outcome, and not doing that is a higher priority than bringing JAXP 1.5/1.6 support.

The good news is it does seem that because of your MODULES-407 work some solution is possible that allows for JAXP 1.5/1.6 support for those uninterested in webservices. They either remove WS from the server config or for deployments that don't need WS they exclude that subsystem via jboss-deployment-structure.xml.

If there's some better way to achieve the outcome that this PR achieves, let's get it recorded. But if it's not acceptable to @jimma and is not achievable in the next few days (say by my start of day next Wednesday) then I think we should go with this, as resolving the webservices problem is a WF 26 blocker. We can always improve more in later releases.

ropalka: AFAIK only 2 standalone jaxws tests are broken plus one WS-RM test that is generating incorrect SOAP envelope.
I do plan to have a look to those 2 standalone jaxws tests ASAP (didn't have time yet).
Are there some other tests that are failing @jimma ? The TCK almost completely passed.

ropalka: I agree with you @bstansberry that if we don't resolve these TCK failures in next few days this PR is the way to go.

ropalka (approved):
Jakarta EE 8 Glassfish RI is using xerces jar as JAXP provider so it is not possible to pass reverse tests with JDK JAXP implementation on WildFly side. Here's my +1 for this PR.

PR #14986, opened 2021-12-01
[WFLY-15776] Remove management identity's security domain configuration from domain mode

skyllarr:
https://issues.redhat.com/browse/WFLY-15776

### DIFF
--- a/servlet-feature-pack/galleon-common/src/main/resources/feature_groups/host-management-interfaces.xml
+++ b/servlet-feature-pack/galleon-common/src/main/resources/feature_groups/host-management-interfaces.xml
@@ -2,11 +2,6 @@
 <feature-group-spec name="host-management-interfaces" xmlns="urn:jboss:galleon:feature-group:1.0">
     <!-- TODO Temporary fork to remove security realms. -->
 
-    <feature spec="core-service.management.access.identity">
-        <param name="security-domain" value="ManagementDomain"/>
-    </feature>
-
-
     <feature spec="host.core-service.management.management-interface.http-interface">
         <param name="interface" value="management"/>
         <param name="port" value="${jboss.management.http.port:9990}"/>


bstansberry:
@Skyllarr In the scripts we've been shipping since Elytron came in that illustrate how to convert a config to use Elytron security instead of legacy, we include the addition of this identity-realm:

https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/common/src/main/resources/content/docs/examples/enable-elytron.cli#L14

I figure there was a reason we had that, so I'd like to understand better why we did and why it seems to be wrong to have it.

I think that file is unnecessary now, but that's a different topic.

Skyllarr:
@bstansberry Thank you, I did not know this was in the enable-elytron script. 

This identity mapping configures "ManagementDomain" to be the domain to obtain the identity for the authorization decisions. So IIUC even if different domain is configured for authentication to the management-interface, the "ManagementDomain" will be used for authorization operations. This can result in the whoami operation to return anonymous identity even if a user authenticated with specific identity. 

I did not think this should be default behaviour. But since it was in the enable-elytron script as well, I need to reconsider this.

darranl:
I think we will be safe removing this identity mapping.

I think as you have seen from debugging for management we default to using the security domain that was used to secure the management interface so we don't really need the domain to be specified here.

Where this mapping does become useful is where we have requests coming in from multiple sources, potentially different management interfaces, directly from application and from JMX and possibly a few others.

At the core of Elytron is the notion that an identity can flow between security domains and as a result of this flow the identity is "transformed" to meet the new context.  i.e. a user of an application may be mapped to roles specific to the application but once that identity flows to the security domain for management they would have a different set of roles for management.  

This is the time we would want to use that element as at that point we are specifying the common security domain all identities can map to.

We are going back about six years so I think we may have had a point in the codebase where we didn't automatically use the identity from the interface so this was mandatory to switch to Eltytron, I think it was something around that which meant it was added to the script.

@Skyllarr did you want to have one more check and I can approve?  Also can you please raise the issue to remove that script?

Skyllarr:
@darranl  Thanks a lot for in depth explanation. It makes sense to remove this mapping then. Users can add it when they need to eg. when they have multiple managment interfaces as you mentioned. 

Btw it is added in [wildfly-core](https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/galleon-common/src/main/resources/layers/standalone/management/layer-spec.xml#L9) management layer as well. So I'll remove it from there also. 

I raised [issue](https://issues.redhat.com/browse/WFLY-15789) to remove the script.

darranl (approved):
Approved, this is only needed for limited advanced use cases - by default we will assume the domain of the management endpoint.

PR #14994, opened 2021-12-03
[WFLY-5754] Use specific description for reply value types in TimerAttributeDefinition and LegacySupport

petrberan:
https://issues.redhat.com/browse/WFLY-5754

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
@@ -1,0 +1,68 @@
+package org.jboss.as.test.integration.ejb.timerservice.attribute;
+
+import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
+import org.jboss.as.controller.descriptions.StandardResourceDescriptionResolver;
+import org.jboss.as.ejb3.subsystem.deployment.TimerAttributeDefinition;
+
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.junit.Test;
+
+import java.util.Locale;
+
+import static org.jboss.as.controller.SimpleAttributeDefinitionBuilder.create;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DESCRIPTION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NILLABLE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REPLY_PROPERTIES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.TYPE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UNIT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE_TYPE;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class TimerAttributeDefinitionTestCase {
+
+    @Test
+    public void testTimerAttributeDefinition(){
+        SimpleOperationDefinitionBuilder builder = new SimpleOperationDefinitionBuilder("operation",
+                new StandardResourceDescriptionResolver("simple", TimerAttributeDefinitionTestCase.class.getName(), Thread.currentThread().getContextClassLoader()));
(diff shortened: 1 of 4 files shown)


petrberan:
Contains changes needed for https://github.com/wildfly/wildfly-core/pull/4851 to pass

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DESCRIPTION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.NILLABLE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REPLY_PROPERTIES;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.TYPE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.UNIT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.VALUE_TYPE;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class TimerAttributeDefinitionTestCase {
```
bstansberry, line 23: @petrberan This test should be in the ejb3 maven module.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    public void testTimerAttributeDefinition(){
+        SimpleOperationDefinitionBuilder builder = new SimpleOperationDefinitionBuilder("operation",
+                new StandardResourceDescriptionResolver("simple", TimerAttributeDefinitionTestCase.class.getName(), Thread.currentThread().getContextClassLoader()));
```
bstansberry, line 28: This should be using the ejb3 subsystem ResourceDescriptionResolver so it can validate that necessary i18n strings are available in the production code. Once this test is in the ejb3 subsystem module it should be possible to use that via EJB3Extension.getResourceDescriptionResolver.

TimerAttributeDefinitionTestCase.properties should not be needed since the real properties can be used.

bstansberry (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
petrberan: Thank you, moved

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/timerservice/attribute/TimerAttributeDefinitionTestCase.java
petrberan: I'm having 2 problems with this change. Could I perhaps ask you for an assistence in this matter?

I'm able to get the [LocalDescriptions.properties](https://github.com/wildfly/wildfly/blob/main/ejb3/src/main/resources/org/jboss/as/ejb3/subsystem/LocalDescriptions.properties) locale bundle, which is fine. But I'm not able to extract descriptions with only `timer` prefix (such as [timer.calendar-timer](https://github.com/wildfly/wildfly/blob/f132f24cabad4d5089eb96f9f1a6ce9b8e95e6a6/ejb3/src/main/resources/org/jboss/as/ejb3/subsystem/LocalDescriptions.properties#L67)) since the OperationDefinitionBuilder requires some prefix even before the `timer` prefix itself (would need `<prefix>.timer.calendar-timer`). Do you know of any way of getting descriptions with only timer prefix?

Second problem being that since this issue is about Operation reply value description, sooner or later I find myself on [this line](https://github.com/wildfly/wildfly-core/blob/d500934273e89a1def8b4d718687e6768484efba/controller/src/main/java/org/jboss/as/controller/descriptions/StandardResourceDescriptionResolver.java#L222), that is looking for key with `reply` in it. There is no key with reply in LocalDescriptions.properties, meaning that I would need to add `timer.reply` to the bundle, effectively bypassing the reason why I would use the EJB3Extension as suggested over my own description bundle.

darranl (changes requested):
I think now it would be better to remove the LegacySupport changes under the security subsystem as that has now been removed from WildFly.

PR #14997, opened 2021-12-03
[WFLY-15286] RESTEasy Upgrade to 5.0

jamezp:
Upgrades base RESTEasy to 5.0.1.Final. It introduces the new RESTEasy MicroProfile dependency which consists of the MicroProfile REST Client and MicroProfile Config sources.

This also removes RESTEasy Spring and a minor update of renaming a couple test servlet names.

## [WFLY-15286](https://issues.redhat.com/browse/WFLY-15286)
This is the first commit and upgrades to RESTEasy 5  (Jakarta RESTful Web Services 2.1) .

## [WFLY-15705](https://issues.redhat.com/browse/WFLY-15705)
This is the third and fourth commits. The third commit removes RESTEasy Spring from WildFly. This was done after the upgrade which did introduce the new RESTEasy Spring dependencies in case we need to revert it.

The fourth commit here just changes a couple servlet names in some test `web.xml` files. This is needed, but it seemed odd they servlets were named "spring" and had nothing to do with Spring.


darranl:
@jamezp we probably need to catch up on this one, I know we talked about but that feels like it was last year ;-)

jamezp:
@darranl Sounds good to me. Ping me whenever. It may or may not be required depending on the direction we go with Jakarta EE.

jamezp:
I'm closing this as it's now out of date.

PR #15000, opened 2021-12-06
WFLY-15792 Simplify getTimeZone in CalendarBasedTimeout and remove re…

chengfang:
…dundant null checks

https://issues.redhat.com/browse/WFLY-15792

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/schedule/CalendarBasedTimeout.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/schedule/CalendarBasedTimeout.java
@@ -23,7 +23,6 @@ package org.jboss.as.ejb3.timerservice.schedule;
 
 import static org.jboss.as.ejb3.logging.EjbLogger.EJB3_TIMER_LOGGER;
 
-import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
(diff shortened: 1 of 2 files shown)


darranl:
@tadamski Are you able to add a review to this one?

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/schedule/CalendarBasedTimeout.java
```diff
@@ -539,47 +530,6 @@ private boolean isDayOfMonthWildcard() {
         return this.scheduleExpression.getDayOfMonth().equals("*");
     }
 
-    private void nullCheckScheduleAttributes(ScheduleExpression schedule) {
```
tadamski, line 542: @chengfang Are we certain that those are not null? Are all those parts of code reduntant?

chengfang: These 7 fields all have non-null defaults: https://github.com/jakartaee/enterprise-beans/blob/master/api/src/main/java/jakarta/ejb/ScheduleExpression.java#L533-L540

So unless the user intentionally set them to null, they will not be null.

tadamski: ok, thanks for claryfing

tadamski (approved):

PR #14614, opened 2021-09-06
[WFCORE-5343]: Preparing YAML Support

ehsavoie:
* Adding metadata for the underlying types for a list attrbite definition.

Thanks for submitting your Pull Request!

Please delete this text, and add a link to the Jira issue solved by this PR.

Remember to use the Jira issue ID in the PR title, and any commits.

PR #15006, opened 2021-12-07
[WFCORE-5343]: Preparing YAML Support

ehsavoie:
* Adding metadata for the underlying types for a list attribute definition.

Jira: https://issues.redhat.com/browse/WFCORE-5343

Precedes #https://github.com/wildfly/wildfly/pull/14614


bstansberry:
We've been chatting, and I don't think this is necessary as the new abstract ListAttributeDefinition cannot be abstract; it must have a default impl.

The ListAttributeDefinition subclasses modified here don't need to be modified to allow YAML override support:

1) TimerAttributeDefinition is a runtime-only attribute.
2) The LegacySupport ones are part of the model-only legacy security subsystem, which we only support in admin-only mode to facilitate config migration. YAML overrides is out of scope for that subsystem.

PR #15024, opened 2021-12-12
[WFLY-15826] Unable to deploy WAR with only real in login-config

whittlec:
- Add unit test
- Swap the order of the equals expression so it avoids the NPE
- Remove throws declaration in private method so as to build

https://issues.redhat.com/browse/WFLY-15826

### DIFF
--- a/elytron-oidc-client/subsystem/src/test/java/org/wildfly/extension/elytron/oidc/OidcActivationProcessorTest.java
+++ b/elytron-oidc-client/subsystem/src/test/java/org/wildfly/extension/elytron/oidc/OidcActivationProcessorTest.java
@@ -1,0 +1,223 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


elytron-oidc-client/subsystem/src/test/java/org/wildfly/extension/elytron/oidc/OidcActivationProcessorTest.java
```diff
@@ -0,0 +1,205 @@
+package org.wildfly.extension.elytron.oidc;
```
fjuma, line 1: Just a very minor comment. This file is missing a copyright header. This can be copied/pasted from any other java file.

fjuma (commented):
@whittlec Thanks very much for working on this! I've just added a small comment.

fjuma:
@whittlec Thanks for adding that! If you could also squash your commits into one and include "[WFLY-15826]" in your commit message, that would be great (e.g., see the commit message format [here](https://github.com/wildfly/wildfly/commit/2da7ffa97963d6d252ef080cce90780edace4e6b)).

darranl:
@fjuma I have just triggered CI, I think we are ready for a final review.

fjuma (approved):
Thanks @whittlec!

bstansberry:
Thanks, @whittlec!

I believe this is your first WildFly PR. Before we merge the first one we have a question:

whittlec:
Yes, I'm quite happy to agree to the MIT license terms.

bstansberry:
Thanks @whittlec!

PR #15026, opened 2021-12-13
[WFLY-15830] Update HostExcludesTestCase configuration to work with WF27

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-15830

Hello @bstansberry / @darranl 

We have removed some extensions during the WF26 development cycle, the `HostExcludesTestCase` tracks those forcing us to re-create an additional enum in the test case itself when we move forward to the next release. 

This commit will be necessary after creating the WF26 release, so I'm adding it here in case you want to cherry-pick it as soon as you open the main branch to WF27 development, then you can immediately close this PR. Otherwise, I'll rebase this and convert it to a real PR once we get the main branch open to WF27 development.


yersan:
@darranl Ok, this still needs the main branch open for WF27 since the test modified here needs to run and pass against WF27. I can convert it to a PR (if that somehow helps, for example, anyone could kick off the Jobs once we have main on WF27 and merge it afterward).

In any case, notice that the current Job report is not valid and this should be merged once the tests have passed on a WF27 branch.

darranl:
@yersan I have just moved main to 27.0.0.Beta1

PR #15028, opened 2021-12-14
[WFLY-15793] Add a ts.elytron-oidc-client profile to testsuite/integration/pom.xml

Ashpan:
https://issues.redhat.com/browse/WFLY-15793

### DIFF
--- a/testsuite/integration/pom.xml
+++ b/testsuite/integration/pom.xml
@@ -271,6 +271,15 @@
             </modules>
         </profile>
 
+        <!-- -Dts.elytron-oidc-client-->
+        <profile>
+            <id>ts.integ.group.elytron-oidc-client</id>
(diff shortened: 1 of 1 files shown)


testsuite/integration/pom.xml
```diff
@@ -271,6 +271,15 @@
             </modules>
         </profile>
 
+        <!-- -Dts.elytron-oidc-client-->
+        <profile>
+            <id>ts.elytron-oidc-client</id>
```
fjuma, line 276: Just a minor comment, may be good to use `ts.integ.group.elytron-oidc-client` here just to be consistent with the other ids.

Ashpan: Changed

fjuma (approved):
Thanks @Ashpan!

PR #15030, opened 2021-12-15
 [WFLY-15832] Ability to create ConfigSource root directory in MP Config subsystem.

kabir:
…mallrye

Enhance tests to illustrate that if a directory for a config-source or a
config-source-root does not exist, the config is still valid and will
not give errors.

https://issues.redhat.com/browse/WFLY-15832 / https://github.com/wildfly/wildfly-proposals/pull/450
Needs a core release with https://github.com/wildfly/wildfly-core/pull/4899

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Config_SmallRye.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Config_SmallRye.adoc
@@ -68,7 +68,7 @@ This results in the XML configuration:
 
 [source,xml,options="nowrap"]
 ----
-<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:1.0">
+<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:2.0">
     <config-source name="file-props">
         <dir path="/etc/config/numbers-app"/>
     </config-source>
@@ -95,6 +95,82 @@ int numMax; // <2>
 This corresponds to the layout used by https://docs.openshift.com/enterprise/3.2/dev_guide/configmaps.html[OpenShift ConfigMaps].
 The `dir` value corresponds to the `mountPath` in the ConfigMap definition in OpenShift or Kubernetes.
 
+==== ConfigSources from Root directory
+You can also point to a root directory by adjusting the examples in the preceding section to include
+`root=true` when defining them. Top level directories within this root directory each become an
+individual `ConfigSource` reading from a directory similar to what we saw earlier in
+<<configsource-from-directory, ConfigSource from Directory>>. Any directories below the top-level
+directories are ignored. Also, any files in the root directory are ignored; only files in the top
+level directories within the root directory will be used for the configuration.
+
+This is especially useful when running on OpenShift where constructs such as `ConfigMap` and
+`ServiceBinding` instances get mapped under
+a common known location. For example if there are two `ConfigMap` instances (for this example,
+we will call them `map-a`, and `map-b`) used by your application pod, they will
+each get mapped under `/etc/config`. So you would have `/etc/config/map-a` and `/etc/config/map-b`
+directories.
+
+Each of these directories will have files where the file name is the name of the property and the file
+content is the value of the property, like we saw earlier.
+
+We can thus simply run the following CLI command to pick up all these child directories as
+a `ConfigSource` each:
+[source,options="nowrap"]
+----
+/subsystem=microprofile-config-smallrye/config-source=config-map-root:add(dir={path=/etc/config, root=true})
+----
+
+This results in the XML configuration:
+
+[source,xml,options="nowrap"]
+----
+<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:2.0">
+    <config-source name="config-map-root">
+        <dir path="/etc/config" root="true"/>
+    </config-source>
+</subsystem>
+----
+Assuming the `/etc/config` directory contains the `map-a` and `map-b` directories we are using
+for this example, the above is analogous to doing:
+----
+/subsystem=microprofile-config-smallrye/config-source=file-props:add(dir={path=/etc/config/map-a})
+/subsystem=microprofile-config-smallrye/config-source=file-props:add(dir={path=/etc/config/map-b})
+----
+Specifying the root directory rather than each individual directory removes the need to know the
+exact names of each entry under the common parent directory (this is especially useful in some
+OpenShift scenarios where the names of these directories are auto-generated).
+
+The situation where two `ConfigSource` entries under the same root both contain the same property
+should be avoided. However, to make this situation deterministic, the directories representing each
+`ConfigSource` are sorted by their name according to standard Java sorting rules before doing the
+lookup of values. To make this more concrete, if we have the following entries:
+* `/etc/config/map-a/name` contains `kabir`
+* `/etc/config/map-b/name` contains `jeff`
+
+Since `map-a` will come before `map-b` after sorting, in the following scenario `kabir` (coming from
+`map-a`) will be injected for the following `username` field:
+
+[source,java,options="nowrap"]
+----
+@Inject
+@ConfigProperty(name = "name")
+String username;
+----
+
+
+You may override this default sorting by including a file called `config_ordinal` in a directory. The
+ordinal specified in that file will be used for config values coming from that directory. Building
+on our previous example, if we had:
+* `/etc/config/map-a/config_ordinal` contains `120`
+* `/etc/config/map-b/config_ordinal` contains `140`
+
+Since now `map-b` has a higher ordinal (`140`) than `map-a` (`120`), we will instead inject the value
+`jeff` for the earlier `username` field.
+
+If there is no `config_ordinal` file in a top-level directory under the root directory, the
+ordinal used when specifying the `ConfigSource` will be used for that directory.
+
+
 === ConfigSource from Class
 
 You can create a specific type of `ConfigSource` implementation by creating a `config-source` resource
@@ -111,7 +187,7 @@ This results in the XML configuration:
 
 [source,xml,options="nowrap"]
 ----
-<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:1.0">
+<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:2.0">
     <config-source name="my-config-source">
         <class name="org.example.MyConfigSource" module="org.example"/>
     </config-source>
@@ -136,7 +212,7 @@ This results in the XML configuration:
 
 [source,xml,options="nowrap"]
 ----
-<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:1.0">
+<subsystem xmlns="urn:wildfly:microprofile-config-smallrye:2.0">
     <config-source-provider name="my-config-source-provider">
          <class name="org.example.MyConfigSourceProvider" module="org.example"/>
     </config-source-provider>
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootRegistrationService.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootRegistrationService.java
@@ -1,0 +1,124 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.microprofile.config.smallrye;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.TreeMap;
+import java.util.function.Supplier;
+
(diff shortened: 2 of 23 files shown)


docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Config_SmallRye.adoc
```diff
@@ -145,6 +145,54 @@ This results in the XML configuration:
...
+You can create a `ConfigSourceProvider` for a root directory by creating a `config-source-root` resource pointing to a
+root directory. Top level directories within this root directory each become an individual `ConfigSource` reading from a
+directory similar to what we saw earlier in <<configsource-from-directory, ConfigSource from Directory>>.
+
+This is especially useful when running on OpenShift where `ConfigMap` and `ServiceBinding` instances get mapped under
+a common known location. For example if there are two `ConfigMap` instances (for this example, we will call them
+`map-a`, and `map-b`) used by your application pod, they will
+each get mapped under `/etc/config`. So you would have `/etc/config/map-a` and `/etc/config/map-b` directories. Each of
+these directories will have files where the file name is the name of the property and the file content is the value of
+the property, like we saw earlier.
```
jmesnil, line 158: should we mentioned that any directories below the top-level directories are ignored?

iirc some implementation of mp config allowed to have deep-nested file representation, eg
```
$ cat a/b/c/foo
bar
```

would be equivalent to the prop `a.b.c.foo=bar`

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Config_SmallRye.adoc
```diff
@@ -145,6 +145,54 @@ This results in the XML configuration:
...
+/subsystem=microprofile-config-smallrye/config-source=file-props:add(dir={path=/etc/config/map-a})
+/subsystem=microprofile-config-smallrye/config-source=file-props:add(dir={path=/etc/config/map-b})
+----
+Using `config-source-root` removes the need to know the exact names of each entry under the common parent directory.
+
+The situation where two `ConfigSource` entries under the same root both contain the same property should be avoided. However, to make this situation deterministic, the directories representing each `ConfigSource` are sorted by their name according to standard Java sorting rules before doing the lookup of values. To make this more concrete, if we have the following entries:
+* `/etc/config/map-a/num.max` contains `100`
+* `/etc/config/map-b/num.max` contains `200`
+
+Since `map-a` will come before `map-b` after sorting, in the following scenario `200` (coming from `map-a`) will be injected
```
jmesnil, line 186: Your example is incorrect:
> Since `map-a` will come before `map-b` after sorting, in the following scenario `200` (coming from `map-a`) will be injected

should be

> Since `map-a` will come before `map-b` after sorting, in the following scenario `100` (coming from `map-a`) will be injected

this could be avoided by explicitly setting the ordinal in each top-level dir according to https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#manually-defining-the-ordinal-of-a-built-in-configsource


* `/etc/config/map-a/config_ordinal -> 101`
* `/etc/config/map-b/config_ordinal -> 102`
-> will always look at map-b first so that `num.max` is resolved to `200`

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootDefinition.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Resource definition to add a base directory as a config source. The contained top-level directories become
+ * dir type config sources.
+ *
+ * @author Kabir Khan
+ */
+class ConfigSourceRootDefinition extends PersistentResourceDefinition {
+
+    static AttributeDefinition ORDINAL = SimpleAttributeDefinitionBuilder.create("ordinal", ModelType.INT)
+            .setDefaultValue(new ModelNode(100))
```
jmesnil, line 53: You are using the existing default ordinal version so no issue here.

However I wonder if we should maybe change it to avoid any clash with the microprofile-config.properties (that is also set to 100) and our existing config source definition.

Maybe putting it below 100 would make sure that a ConfigSourceDir would be use before such a ConfigSourceRoot? wdyt?

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootRegistrationService.java
```diff
@@ -0,0 +1,119 @@
...
+
+        @Override
+        public Iterable<ConfigSource> getConfigSources(ClassLoader forClassLoader) {
+            TreeMap<String, ConfigSource> map = new TreeMap<>();
+            File[] files = rootDirectory.listFiles();
+            if (files != null) {
+                // If the directory exist, this will be an empty list
+                for (File file : rootDirectory.listFiles()) {
+                    if (file.isDirectory()) {
+                        map.put(file.getName(), new FileSystemConfigSource(file, ordinal));
```
jmesnil, line 111: Should we log all the top-level dir that would be used as config sources.

I think that'd be valuable info to see something in the log like:

```
* INFO Using /etc/config/rhosak (ordinal=100) as ConfigSource
* INFO Using /etc/config/mydb-secret (ordinal=102) as ConfigSource
```

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemParser_2_0.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * The name space used for the {@code subsystem} element
+     */
+    public static final String NAMESPACE = "urn:wildfly:microprofile-config-smallrye:2.0";
+
+    private static final PersistentResourceXMLDescription xmlDescription;
+
+    static {
+        xmlDescription = builder(MicroProfileConfigExtension.SUBSYSTEM_PATH, NAMESPACE)
+                .addChild(builder(MicroProfileConfigExtension.CONFIG_SOURCE_PATH)
```
jmesnil, line 40: not that a big deal bit but that might be better to write first the config source provider and then all the kind of config sources

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootDefinition.java
```diff
@@ -0,0 +1,104 @@
...
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+
+/**
+ * Resource definition to add a base directory as a config source. The contained top-level directories become
+ * dir type config sources.
+ *
+ * @author Kabir Khan
+ */
+class ConfigSourceRootDefinition extends PersistentResourceDefinition {
```
jmesnil, line 50: Did you think about incorporating this in the existing ConfigSourceDefintion with a new `root` attribute that would active this feature in couple with the `dir` attribute?

`<config-source="foo" dir="/etc/config/my-secret">` -> create a  config source for the `/etc/config/my-secret`
`<config-source="bar" dir="/etc/config" root="true>` -> create a  config source for each top-level dirs in `/etc/config`

microprofile/config-smallrye/src/main/resources/schema/wildfly-microprofile-config-smallrye_2_0.xsd
```diff
@@ -0,0 +1,82 @@
...
+                                        <xs:attribute name="name" type="xs:string" use="required" />
+                                        <xs:attribute name="value" type="xs:string" use="required" />
+                                    </xs:complexType>
+                                </xs:element>
+                            </xs:sequence>
+                            <xs:element name="class" type="classType" minOccurs="0" maxOccurs="1" />
+                            <xs:element name="dir" minOccurs="0" maxOccurs="1">
+                                <xs:complexType>
+                                    <xs:attribute name="path" type="xs:string" use="required" />
+                                    <xs:attribute name="relative-to" type="xs:string" />
```
jmesnil, line 51: actually, regarding my comment above, the `root` attribute would be part of this complex type

microprofile/config-smallrye/src/test/java/org/wildfly/extension/microprofile/config/smallrye/Subsystem_1_0_ParsingTestCase.java
```diff
@@ -23,30 +23,16 @@
...
-import org.jboss.as.model.test.ModelTestUtils;
 import org.jboss.as.subsystem.test.AbstractSubsystemBaseTest;
 import org.jboss.as.subsystem.test.AdditionalInitialization;
-import org.jboss.as.subsystem.test.KernelServices;
-import org.jboss.as.subsystem.test.KernelServicesBuilder;
-import org.jboss.dmr.ModelNode;
-import org.junit.Test;
 
-public class Subsystem_1_0_ParsingTestCase extends AbstractSubsystemBaseTest {
+public class MPConfigSubsystem_1_0_ParsingTestCase extends AbstractSubsystemBaseTest {
```
jmesnil, line 33: Code does not compile, the file must also be renamed to `MPConfigSubsystem_1_0_ParsingTestCase.java`

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source/from_dir/SetupTask.java
```diff
@@ -50,6 +50,7 @@
 
     private static final String ADDR_A = "/subsystem=microprofile-config-smallrye/config-source=propsA";
     private static final String ADDR_B = "/subsystem=microprofile-config-smallrye/config-source=propsB";
+    private static final String ADDR_NON_EXISTANT = "/subsystem=microprofile-config-smallrye/config-source=not-there";
```
jmesnil, line 53: should it be `ADDR_NON_EXISTENT`?

jmesnil (changes requested):
The request change is to make the code compile :) 

Otherwise, my comment are suggestions but the management model that is in the PR is fine too.

jmesnil:
this PR can be retested after #15035(that incorporates wildfly/wildfly-core#4899 is merged)

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootDefinition.java
kabir: @jmesnil I am refactoring as suggested to use a `root` nested attribute in the `dir` attribute of the original ConfigSourceDefinition. 

But your point got me thinking, and IMHO this ordinal should be **higher** than the microprofile-config.properties one so it fits in with the pattern of something outside the deployment being able to **override** the deployment values. 

I suggest 150, so that system properties and env vars can still override values set in the subsystem. Do you agree?

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceRootRegistrationService.java
kabir: Done

jmesnil (approved):
LGTM

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source/from_root_dir/ConfigSourceRootTestCase.java
```diff
@@ -0,0 +1,89 @@
...
+                .addClasses(TestApplication.class, TestApplication.Resource.class, AbstractMicroProfileConfigTestCase.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        return war;
+    }
+
+    @ArquillianResource
+    private URL url;
+
+    @Test
+    public void testGetWithConfigProperties() throws Exception {
```
mchoma, line 76: So my understanding is this is only test for this feature. What I am going to do tomorrow is to understand if all hard requirements [1] are covered by this test. But I see 9 points in hard requirements, whereas I see 6 tests here. So probably something will be missing?

Maybe some comments could help me understand which test is testing which use case.

[1] https://github.com/wildfly/wildfly-proposals/pull/450/files#diff-9f1354c78479d2107750d078d4f846420ac82216e6eda99d34a0ad36ea2d948fR66

mchoma: 1. Do we have test that root=false is default? Do we have test that if root=false is configured, resolving of nested directories is not working?
Basically I am looking for covering this hard requirement `The default value of the `root` nested attribute is `false`, giving the legacy behaviour`
2. I do not see nested dirs in setup task so I believe those requirements are not tested:
Basically I am looking for covering this hard requirement  `Only top-level directories are considered, directories nested within those are ignored`
3. I do not see files created in root directory
Basically I am looking for covering this hard requirement `Files contained directly within the root directory are ignored and are not used for config`

kabir: @mchoma I believe I have added what you are asking for now (good catch)

I've addressed 1) in ConfigSourceFromDirTestCase. It now adds one of the config sources with the default value for `root` and the other explicitly as `root=false`. In both directories I create a nested directory containing a property file each, and verify those properties are not considered.

For 2 + 3) I have modified ConfigSourceRootTestCase. 
For 2) I added a file in a directory nested under one of the directories under the root dir.
For 3) I added a file directly under the root directory. 
Neither of these files get considered as properties.

microprofile/config-smallrye/src/test/java/org/wildfly/extension/microprofile/config/smallrye/MPConfigSubsystemParsingTestCase.java
```diff
@@ -96,15 +102,52 @@ private void testRejectingTransformers(ModelTestControllerVersion controllerVers
...
+            // Since the transformers are chained the rejection of the 'root' nested attribute installed
+            // by the 'my-config-source-with-root' config source is tested in later legacy controller versions.
+            int remove = -1;
+            for (int i = 0 ; i < ops.size() ; i++) {
+                if (ops.get(i).get("address").asString().contains("my-config-source-with-root")) {
+                    remove = i;
+                }
+            }
+            ops.remove(remove);
+        }
```
bstansberry, line 118: No need to change this now, but if you ever want to clean this out go ahead as there is no longer a requirement to test transformation to prior to WF 23 / EAP 7.4.

bstansberry:
Thanks @kabir

PR #15031, opened 2021-12-16
[PREVIEW] Upgrade to Hibernate Search 6.1

yrodiere:
* [WFLY-15842](https://issues.redhat.com/browse/WFLY-15842): Add (private) module for Elasticsearch's low-level REST client
* [WFLY-15841](https://issues.redhat.com/browse/WFLY-15841): Add (private) module for com.carrotsearch.hppc
* [WFLY-15840](https://issues.redhat.com/browse/WFLY-15840): Upgrade to Apache Avro 1.11.0
* [WFLY-15839](https://issues.redhat.com/browse/WFLY-15839): Upgrade to Apache Lucene 8.11
* [WFLY-15838](https://issues.redhat.com/browse/WFLY-15838): Upgrade to Hibernate Search 6.1

@scottmarlow Opening this as draft to show you what we'll need to do in WildFly 27 on the Hibernate Search side of things, on top of the work you're doing to migrate to Hibernate ORM 6.

For now this works with ORM 5.6, but moving to ORM 6 should only be a matter of changing some Hibernate Search Maven artifact names, once we publish them (in early January).


rursprung:
i saw this by pure chance. with your changes you're now using elasticsearch 7.16.1 (cf. your `pom.xml`). are you aware of the whole license change SNAFU which happened with ES 7.11? as of that release they're no longer Apache 2.0 licensed, leading to issues when using it in some setups (ranging anywhere from "i disagree with their politics" to "my lawyers are uneasy about using this version"). due to this license change a fork has been created: [OpenSearch](https://opensearch.org/).
it seems that Hibernate Search will support it with 6.1, as it's already listed on their page: https://hibernate.org/search/ (and also explicitly in their [compatibility list](https://docs.jboss.org/hibernate/search/6.1/reference/en-US/html_single/#getting-started-compatibility)).

it might be preferable to migrate directly to OpenSearch to stick to a pure Apache 2.0 licensed dependency?

yrodiere:
Thank you for the heads-up. I am unfortunately very aware of the license changes; I'm having to explain this just about every time the topic of the Elasticsearch client comes up :)

Elasticsearch changed its license, that's true. The server changed its license.

The high-level Java REST client also changed its license, because of some bizarre architecture choices that force it to have a dependency on server JARs.

But Hibernate Search uses the low-level Java REST client, and that one is still ASL 2.0. See the source code if you don't believe me: https://github.com/elastic/elasticsearch/blob/master/client/rest/src/main/java/org/elasticsearch/client/RestClient.java#L5-L6.
It's been ASL 2.0 since its inceptions, and Elastic has repeatedly said there are no plans for that to change.

Hibernate Search 6.1 is compatible with OpenSearch, but only an OpenSearch *server*. It does not work with the OpenSearch client JAR, which is binary-incompatible (different packages, classes). Essentially Hibernate Search uses the Elasticsearch low-level REST client (which is basically just an HTTP client) to talk with OpenSearch.

Moving to the OpenSearch low-level REST client is not an option at the moment as it would be a backwards-incompatible change, and Hibernate Search has a compatibility policy that only allows that kind of changes in major releases (e.g. 7.0 or 8.0).

Bottom line: we could change integration tests in this PR to target OpenSearch instead of Elasticsearch, and that would work. But we can't change the client (and we don't need to).

yrodiere:
This PR is now using Hibernate Search 6.1.0.Final. It seems all tests pass except with the security manager enabled, but Hibernate Search has never supported the security manager, even in 5.x, so we can probably exclude Hibernate Search tests when using the security manager.

@scottmarlow how do you want to proceed? Do you want to wait for an upgrade to ORM 6 before we upgrade to Search 6.1, or do you want to upgrade to ORM 5.6 + Search 6.1 as a first step?

scottmarlow:
> 
> @scottmarlow how do you want to proceed? Do you want to wait for an upgrade to ORM 6 before we upgrade to Search 6.1, or do you want to upgrade to ORM 5.6 + Search 6.1 as a first step?

I would like to upgrade to an ORM 6.0 release (perhaps the CR1 when that is released).  Other option could be the already released https://github.com/hibernate/hibernate-orm/releases/tag/6.0.0.Beta3 if CR1 is delayed much.

I did sync https://github.com/scottmarlow/wildfly/tree/orm6_part5 with the latest ORM 6 `main` changes yesterday, so ORM 6.0 CR1 might be better.

scottmarlow:
@yrodiere fyi, just created draft pr https://github.com/wildfly/wildfly/pull/15165.

yrodiere:
Thanks @scottmarlow . I'll update my PR to base it on yours and use Hibernate Search's `-orm6` artifacts as soon as I know where to put my integration tests; I reached out to Brian about that.

yrodiere:
Superseded by #15185

PR #15058, opened 2022-01-03
WFLY-15851 Weld - fix UrlScanner to correctly reference child file instead of parent directory

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-15851

Reported via JIRA; this seems like a small oversight that nobody bumped into up until now.


bstansberry:
Thanks, @manovotn

PR #15040, opened 2021-12-17
[WFLY-15847] Digest mechanism needs sticky sessions

aogburn:
https://issues.redhat.com/browse/WFLY-15847

PR #15062, opened 2022-01-07
[WFLY-15732] / [WFLY-15356] Remove use of security-plugins from Undertow subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-15732

Supersedes:
  https://github.com/wildfly/wildfly/pull/14936
  https://github.com/wildfly/wildfly/pull/15040


darranl:
@pferraro Would it be possible for you to please have a look at the most recent changes, this is based on our discussions last year regarding https://github.com/wildfly/wildfly/pull/14936

You can ignore that it is deprecated in a model version not yet in place as that PR is just under test and will be merged before this one.

https://wildfly.zulipchat.com/#narrow/stream/242838-Jakarta-EE/topic/Undertow.20use.20of.20security-plugins

undertow/src/main/java/org/wildfly/extension/undertow/UndertowRootDefinition.java
```diff
@@ -103,7 +102,7 @@
     protected static final SimpleAttributeDefinition DEFAULT_SECURITY_DOMAIN =
```
pferraro, line 102: Should we deprecate this attribute?

pferraro (changes requested):
There are 2 other classes to be removed:

* org.wildfly.extension.undertow.sso.InMemorySingleSignOnManagerServiceConfigurator
* org.wildfly.extension.undertow.sso.NonDistributableSingleSignOnManagementProvider

undertow/src/main/java/org/wildfly/extension/undertow/UndertowRootDefinition.java
darranl: I think we will be revisiting this attribute as part of another RFE so I don't think we should deprecate within these changes as it does still have an active meaning with an Elytron configuration.

Within the other features we are working on I think does introduce the question "Where should a default security domain be defined?"  IMO we have one subsystem that defines security domains so that should define which is the default instead of all of the other subsystems having their own notion of default.

darranl:
Hi @pferraro I have deleted those two additional classes.

As I mention in my previous comment that attribute may be deprecated as part of another change but the attribute still has a meaning at this point in time.

pferraro (approved):

PR #15072, opened 2022-01-10
[WFLY-16737] Upgrade H2 to 2.1.210

bstansberry:
Use the -ifNotExists option with tests that may have 'remote' callers initialize the H2 db
    
H2 2.1.210 defaults to a random management password so set a real one
    
Work around the fact that H2 2.x no longer treats BINARY cols as VARBINARY and defaults to size of 1.
    
    TableResourceDefinition's ColumnAttribute.DATA defaults to BINARY, which will result in a col wherein H2 will only allow a single byte. I suspect the design assumed a more varbinary semantic.
    
    See https://github.com/h2database/h2database/pull/2427
    
    Perhaps the default value of TableResourceDefinition.ColumnAttribute.DATA should be reconsidered.
    
    An alternative fix would be to configure a 'table=string' resource and specify VARBINARY as the col type for 'data'.

https://issues.redhat.com/browse/WFLY-16737

H2 offers different 'compatibility modes' (see http://www.h2database.com/html/features.html#compatibility). For a while this branch was using MODE=LEGACY but that no longer seems to be necessary to pass the testsuite, so my preference is to default to the H2 default of REGULAR.  But I added an expression to the DS URL to make it easy to override, e.g. to LEGACY.

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/persistence/AbstractDatabasePersistenceWebFailoverTestCase.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/persistence/AbstractDatabasePersistenceWebFailoverTestCase.java
@@ -73,7 +73,7 @@ public abstract class AbstractDatabasePersistenceWebFailoverTestCase extends Abs
     public static class ServerSetupTask extends CLIServerSetupTask {
         public ServerSetupTask() {
             this.builder.node(THREE_NODES)
-                    .setup("/subsystem=datasources/data-source=web-sessions-ds:add(jndi-name=\"java:jboss/datasources/web-sessions-ds\", enabled=true, use-java-context=true, connection-url=\"jdbc:h2:tcp://localhost:%s/./web-sessions\", driver-name=h2", DB_PORT)
+                    .setup("/subsystem=datasources/data-source=web-sessions-ds:add(jndi-name=\"java:jboss/datasources/web-sessions-ds\", enabled=true, use-java-context=true, connection-url=\"jdbc:h2:tcp://localhost:%s/./web-sessions;VARIABLE_BINARY=TRUE\", driver-name=h2", DB_PORT)
                     .setup("/subsystem=infinispan/cache-container=web/invalidation-cache=database-persistence:add")
                     .setup("/subsystem=infinispan/cache-container=web/invalidation-cache=database-persistence/store=jdbc:add(data-source=web-sessions-ds, fetch-state=false, shared=true)")
                     .teardown("/subsystem=infinispan/cache-container=web/invalidation-cache=database-persistence:remove")
--- a/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
@@ -7,7 +7,7 @@
             <param name="use-java-context" value="true"/>
             <param name="jndi-name" value="java:jboss/datasources/ExampleDS"/>
             <param name="data-source" value="ExampleDS"/>
-            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE&quot;"/>
+            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;MODE=LEGACY&quot;"/>
             <param name="driver-name" value="h2"/>
             <param name="user-name" value="sa"/>
             <param name="password" value="sa"/>
--- a/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
@@ -7,7 +7,7 @@
             <param name="use-java-context" value="true"/>
             <param name="jndi-name" value="java:jboss/datasources/ExampleDS"/>
             <param name="data-source" value="ExampleDS"/>
-            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;MODE=LEGACY&quot;"/>
+            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;MODE=${wildfly.h2.compatibility.mode:REGULAR}&quot;"/>
             <param name="driver-name" value="h2"/>
             <param name="user-name" value="sa"/>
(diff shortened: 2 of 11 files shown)


boris-unckel:
@bstansberry You have to set the major version `2` and minor version `1` for H2: 
* https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml#L4
* https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/wrong-empty-name-driver.xml#L4
* https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/wrong-wo-name-driver.xml#L4

(wondering why these are not below ...src/test/resources/....)

Reference is here:
* major: https://github.com/h2database/h2database/blob/master/h2/src/main/org/h2/engine/Constants.java#L86
* minor: https://github.com/h2database/h2database/blob/master/h2/src/main/org/h2/engine/Constants.java#L91

bstansberry:
Thanks, @boris-unckel. That will fix a few.

boris-unckel:
Sorry for the wrong minor, you fixed it already. 2.0.206 is still at minor 0.

boris-unckel:
@beikov Is it correct that we have no chance to use H2 2.0.206 due to this https://github.com/h2database/h2database/issues/3340#issuecomment-1006711337 (fix not released yet) using Hibernate 5.3.x?

boris-unckel:
@bstansberry I have found a solution (hm, workaround). H2 1.x treats `NULL` as `DEFAULT` (which is a sequence here) Enable the 1.x compatibility mode with `MODE=LEGACY` https://www.h2database.com/html/features.html#compatibility `Identity columns have implicit DEFAULT ON NULL clause. It means a NULL value may be specified for this column in INSERT command and it will be treated as DEFAULT.`

```
diff --git a/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml b/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
index 54745faf9f..7cda7ad76f 100644
--- a/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
+++ b/ee-feature-pack/galleon-common/src/main/resources/feature_groups/h2-datasource.xml
@@ -7,7 +7,7 @@
             <param name="use-java-context" value="true"/>
             <param name="jndi-name" value="java:jboss/datasources/ExampleDS"/>
             <param name="data-source" value="ExampleDS"/>
-            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE&quot;"/>
+            <param name="connection-url" value="&quot;jdbc:h2:mem:test;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;MODE=LEGACY&quot;"/>
             <param name="driver-name" value="h2"/>
             <param name="user-name" value="sa"/>
             <param name="password" value="sa"/>

```

bstansberry:
Thanks @boris-unckel I went ahead and merged that into my branch to see what effect it has.

Note that this is an experimental branch so we wouldn't bump H2 like this until folks like @scottmarlow (JPA/Hibernate) have had a chance to evaluate. I don't think I'd want to move on to a version that doesn't pass Hibernate testing.

boris-unckel:
Hibernate has a PR to support H2 version 2.x https://github.com/hibernate/hibernate-orm/pull/4588 It's a Hibernate 6.0 PR with jakarta.* package names.

scottmarlow:
> Note that this is an experimental branch so we wouldn't bump H2 like this until folks like @scottmarlow (JPA/Hibernate) have had a chance to evaluate. I don't think I'd want to move on to a version that doesn't pass Hibernate testing.

+1, once https://github.com/hibernate/hibernate-orm/pull/4588 is merged, I think that should allow hibernate master branch to be used for testing with H2 but I don't think that https://github.com/hibernate/hibernate-orm/pull/4588 is actually testing against H2 from what I can tell by looking at https://github.com/hibernate/hibernate-orm/blob/main/gradle/libraries.gradle#L16:
```
 h2Version = '1.4.197'
```

beikov:
Hi all. I think that with H2 2.1.210, which was released a few hours ago, the Hibernate tests work after my PR https://github.com/hibernate/hibernate-orm/pull/4588. So to get the testsuite with 5.3 working I will need to backport some fixes like ignoring time zone related test failures, which according to the H2 team are expected somehow, and also some "bad" column names which are now keywords that need to be renamed.

We can't bump the version to 2.1.210 yet though because we also use h2gis for testing spatial functionality and that library hasn't provided a version yet which is compatible with the new H2 version. Here is the issue for tracking this: https://github.com/orbisgis/h2gis/issues/1235

As soon as H2GIS 2.0 is out, we can start to evaluate bumping the H2 version on main/6.0 as well, but from what I tested locally, H2 2.1.210 works with Hibernate 6.0-SNAPSHOT and should also work with Hibernate 5.3 after my backports. Is it really necessary for 5.3 to support H2 2.0+ or can we make that a 6.0 only thing?

scottmarlow:
Is it really necessary for 5.3 to support H2 2.0+ or can we make that a 6.0 only thing?

IMO, supporting H2 on Hibernate ORM 6 sounds like a good step forward.

bstansberry:
@darranl let's keep it for now.

bstansberry:
This is now ready to go. @boris-unckel fyi re the last commit.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/persistence/AbstractDatabasePersistenceWebFailoverTestCase.java
```diff
@@ -73,7 +73,7 @@ public static void afterClass() throws Exception {
     public static class ServerSetupTask extends CLIServerSetupTask {
         public ServerSetupTask() {
             this.builder.node(THREE_NODES)
-                    .setup("/subsystem=datasources/data-source=web-sessions-ds:add(jndi-name=\"java:jboss/datasources/web-sessions-ds\", enabled=true, use-java-context=true, connection-url=\"jdbc:h2:tcp://localhost:%s/./web-sessions\", driver-name=h2", DB_PORT)
+                    .setup("/subsystem=datasources/data-source=web-sessions-ds:add(jndi-name=\"java:jboss/datasources/web-sessions-ds\", enabled=true, use-java-context=true, connection-url=\"jdbc:h2:tcp://localhost:%s/./web-sessions;VARIABLE_BINARY=TRUE\", driver-name=h2", DB_PORT)
```
bstansberry, line 76: @pferraro @rhusar FYI re this. See the PR description or first commit's message for more about this.

rhusar: 👍

jamezp (approved):

PR #15086, opened 2022-01-12
[WFLY-15904] Update test depedency nimbus-jose-jwt from 8.11 to 8.23

bstansberry:
This it to trigger an update of transitive dep json-smart to 2.4.5 or later to get it off security scanner radar.

The nimbus-jose-jwt project has a number of 9.x releases but I went for the latest release in the same major version, just to improve the odds the upgrade would be suitable.

https://issues.redhat.com/browse/WFLY-15904

Note that this dep is only used in tests; this change has no impact on production code.

@fjuma If this passes CI, please review.


fjuma (approved):
Thanks @bstansberry!

PR #15096, opened 2022-01-13
[WFLY-15918] Use java.transaction.xa JPMS module together with java.sql module where necessary

ropalka:
https://issues.redhat.com/browse/WFLY-15918


ropalka:
In upcomming JBoss Modules 2.0.2.Final upgrade all java.transaction.xa exported packages will be removed from java.sql module to be in alignment with Java's JPMS definitions. Thus it is necessary to properly reference java.transaction.xa instead of java.sql.

ropalka:
@fl4via or @bstansberry please review

PR #14887, opened 2021-11-08
[WFLY-15524]: Upgrade netty from 4.1.68 to 4.1.70.

ehsavoie:
* Upgrading netty to 4.1.70.
* Removing netty-all and using the proper netty artefacts in pom.xml.
* Removing the io.netty jboss module and creating io.netty.netty-*
  modules for each artefact.
* Adding a task to manage the native parts.

Jira: https://issues.redhat.com/projects/WFLY/issues/WFLY-15524
Requires: https://github.com/wildfly/wildfly/pull/14886

PR #15122, opened 2022-01-17
[26.x][WFLY-15524]: Upgrade netty from 4.1.68 to 4.1.72. 

ehsavoie:
* Upgrading netty to 4.1.72.
* Removing netty-all and using the proper netty artefacts in pom.xml.
* Removing the io.netty jboss module and creating io.netty.netty-*
modules for each artefact.
* Adding a task to manage the native parts.
* Adding the missing io.netty.netty-handler dependency to the artemis journal module.

Jira: https://issues.redhat.com/browse/WFLY-15935
Jira: https://issues.redhat.com/browse//WFLY-15524

Main PRs:
https://github.com/wildfly/wildfly/pull/14887
https://github.com/wildfly/wildfly/pull/15117


jmesnil:
@ehsavoie could you please mention in the PR description the upstream PRs for the `main` branch?

ehsavoie:
@jmesnil Done.
Also I'm not sure the CI failures are related to that PR

jmesnil:
@ehhttps://ci.wildfly.org/buildConfiguration/WF_PullRequest26x_WindowsJdk11/290926?showRootCauses=false&expandBuildChangesSection=true&expandBuildTestsSection=true might be related

OIDC failures are a known regression that will be fixed by an upcoming WildFly Core release

jmesnil (commented):
@ehsavoie can you explain why the replacement of netty-all required the creation of all these JBOss modules?

Couldn't they all be provided by `io.netty`? 
As the PR stands, that seems to be a significant changes for 26.0.1.Final?

ehsavoie:
@jmesnil  The initial goal was to reduce the size and to have modules only depending of what they required instead of depending on everything as some parts are only used by some subsystems.
I agree that it it 'big' for 26.0.1 but it is cleaner in the long run.

jmesnil (changes requested):
We can't remove `io.netty` module in a micro release of WildFly.
Brian told me that the reason it was removed was that we could not keep this module and exposing a different set of resources.
Adding artifacts until we are able to provide an identical `io.netty` would required new devs and we are past the freeze for WildFly 26.0.1.Final.

This PR will have to wait after this release to be merged.

jmesnil:
@boris-unckel  my understanding is that CVE-2021-43797  is not relevant for WildFly as we don't use Netty for HTTP. @ehsavoie is that correct?

boris-unckel:
@jmesnil @ehsavoie @bstansberry thank you all to clarify the impact.

bstansberry:
Thanks, @ehsavoie

PR #15126, opened 2022-01-18
[WFLY-15943]: Can't create an external connection factory using a discovery group with socket binding.

ehsavoie:
* Fixing the way the service name is created.

Jira: https://issues.redhat.com/browse/WFLY-15943

### DIFF
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/ExternalConnectionFactoryAdd.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/ExternalConnectionFactoryAdd.java
@@ -82,7 +82,7 @@ public class ExternalConnectionFactoryAdd extends AbstractAddStepHandler {
         List<String> connectorNames = Common.CONNECTORS.unwrap(context, model);
         ServiceBuilder<?> builder = context.getServiceTarget()
                 .addService(serviceName)
-                .addAliases(MessagingServices.getActiveMQServiceName(name));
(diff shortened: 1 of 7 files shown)


messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/ExternalConnectionFactoryAdd.java
```diff
@@ -82,7 +82,7 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
         List<String> connectorNames = Common.CONNECTORS.unwrap(context, model);
         ServiceBuilder<?> builder = context.getServiceTarget()
                 .addService(serviceName)
-                .addAliases(MessagingServices.getActiveMQServiceName(name));
+                .addAliases(JMSServices.getConnectionFactoryBaseServiceName(MessagingServices.getActiveMQServiceName("")).append(name));
```
jmesnil, line 85: what would be the difference between ^^ and `JMSServices.getConnectionFactoryBaseServiceName(MessagingServices.getActiveMQServiceName(name))` ?

ehsavoie: it is different because name is appended at the end of the service name: *jboss.messaging-activemq.jms.cf.name* while your change will create *jboss.messaging-activemq.name.jms.cf*

bstansberry: A new no-arg MessagingServices.getActiveMQServiceName() would be clearer.

PR #15136, opened 2022-01-22
[WFLY-15747] / [WFLY-13889] Remove PicketBox dependency from iiop-openjdk.

darranl:
https://issues.redhat.com/browse/WFLY-15747

Supersedes https://github.com/wildfly/wildfly/pull/14948


darranl:
Hi @tadamski this follow up on https://github.com/wildfly/wildfly/pull/14948 as I needed to submit as a new PR if you could please review the last commit I have applied the changes you requested.

I have also left it that if somehow the values for legacy security are set we report an error and fail, I want to avoid ignoring legacy security and accidentally leaving something unsecured.

darranl:
Looks like it now has some related CI failures to follow up.

tadamski (approved):
I think it is good now

PR #15140, opened 2022-01-25
[WFLY-15964] OpenTelemetry ratio-based sampler should not accept value bigger than 1.0.

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-15964
OpenTelemetry ratio-based sampler should not accept value bigger than 1.0.


bstansberry:
@jasondlee Please review

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
```diff
@@ -146,7 +146,7 @@
             .setAttributeGroup(GROUP_SAMPLER)
             .setValidator((parameterName, value) -> {
                 if (value.isDefined() && value.getType() != ModelType.EXPRESSION) {
-                    long val = value.asLong();
+                    double val = value.asDouble();
```
jasondlee, line 149: Ha. That's an embarrassing mistake. :) Looks good to me. Good catch!

jasondlee (approved):

PR #15146, opened 2022-01-25
WFLY-15963 Combine ORDINALS and ORDINAL_TO_WEEK_NUMBER_MAPPING fields…

chengfang:
… in ejb3 DayOfMonth class

cleanup (improve the use of toLowerCase and equalsIgnoreCase; remove unnecessary null check)

https://issues.redhat.com/browse/WFLY-15963

some notes about changes of code cleanup:

* the regex `Pattern "\\s+"` is moved to the class-level to avoid compiling the pattern for every invocation of the method.
* `isDayOfWeekBased` method is renamed to `splitDayOfWeekBased` and the return type is changed from `boolean` to `String[]`. This method splits the input string to do the checking, and the caller of this method also needs to split the data. To avoid splitting it twice, this method is changed to return the split result: a non-null value is equivalent to previous true return value; a null is equivalent to the previous false return value.


chengfang:
@tadamski please review.

tadamski (approved):
LGTM

PR #15147, opened 2022-01-26
[WFLY-15969] Upgrade ASM to 9.2 in order to support JDK18

ropalka:
https://issues.redhat.com/browse/WFLY-15969


bstansberry:
@jimma Please review.

We dropped the Hibernate transformer module so there's no JPA/Hibernate use any more.

If @kabir wants to approve, great, but it's not required as the smallrye-config releases we use depend on ASM 9.2.

kabir (approved):

jimma:
@ropalka @bstansberry  This looks good to me.

bstansberry:
Thanks, @ropalka

PR #14958, opened 2021-11-25
[WFLY-14842] / [WFLY-13889] Remove JCA dependency on PicketBox.

darranl:
Submitting as a draft as the following PR needs to be approved first:
    https://github.com/wildfly/wildfly/pull/14945

https://issues.redhat.com/browse/WFLY-14842

PR #15156, opened 2022-01-27
[WFLY-14842] / [WFLY-13889] Remove JCA dependency on PicketBox.

darranl:
https://issues.redhat.com/browse/WFLY-14842

Supersedes https://github.com/wildfly/wildfly/pull/14958


darranl:
Hi @tadamski we now have CI passing, if you could please review the final commit we can get this in.

This is now the last component using PicketBox.

tadamski (approved):
LGTM

PR #15157, opened 2022-01-27
[WFLY-15978] Enable Elytron domain security for EAR deployed connectors

reinhapa:
https://issues.redhat.com/browse/WFLY-15978


darranl (dismissed):
As this is a single change not connected to anything else can you please rebase to remove the merge commit.

Also as you are doing this can you please add the Jira ID to the commit message.

reinhapa:
@darranl I will do this with the next commit as I removed all legacy checks in the connector space to always use Elytron security...

tadamski (approved):
After discussion with Patrick I think this PR is good. There is no need to distinguish between elytron security domain and legacy one and because of that we can treat each security domain configured directly in IronJacamar as Elytron one.

@fjuma @darranl does this makes sense from security perspective?

fjuma (approved):
Thanks @reinhapa!

bstansberry:
Thanks, @reinhapa!

PR #15159, opened 2022-01-28
[WFLY-15609] added new http handler for transaction timeout cleanup

istudens:
Issue: https://issues.redhat.com/browse/WFLY-15609

### DIFF
--- a/ee-9/source-transform/undertow/pom.xml
+++ b/ee-9/source-transform/undertow/pom.xml
@@ -272,5 +272,9 @@
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron-ssl</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.transaction</groupId>
+            <artifactId>wildfly-transaction-client</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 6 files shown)


istudens:
@fl4via could you review this fix please? 
It brings a new dependency on `org.wildfly.transaction.client` module due to `ContextTransactionManager` usage. Tests are passing as well as the reproducer, but I am not completely sure about this new dependency and its potential influence. I have to mark it `optional` since it won't compile the `wildfly-servlet-galleon-pack` otherwise.

darranl:
Should we really be adding further dependencies to the servlet feature pack?

As we work more with feature packs and layers this feels like the kind of handler that should be dynamically added based on the presence of other layers allowing the overall set of handlers in the resulting servlet container to be defined by the provisioned server.

istudens:
@darranl Is there any example of such an approach you mentioned in WildFly which I can look at?

xstefank:
retest this please

xstefank:
@istudens @darranl @fl4via can you take a look at this, please? Getting stale.

istudens:
I will open discussion regarding this issue.

fl4via:
@istudens I am also unsure about the addition of a new dependency for this. I would go with this solution if it is the only way to do fix the bug.
Have you started discussion about this somewhere? I cannot find it?

PR #15166, opened 2022-01-31
WFLY-15988 Use new set instead of add in BatchThreadPoolDescriptionRe…

chengfang:
…solver, and use array instead of list in ChainedContextHandle

https://issues.redhat.com/browse/WFLY-15988


pferraro (approved):

bstansberry:
Thanks, @chengfang

PR #15174, opened 2022-02-01
[WFLY-15992] Add a testsuite structure for tests specific to WildFly …

bstansberry:
…Preview

https://issues.redhat.com/browse/WFLY-15992

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -508,6 +508,7 @@
             <modules>
                 <module>integration</module>
                 <module>layers</module>
+                <module>preview</module>
                 <module>scripts</module>
             </modules>
         </profile>
--- a/testsuite/preview/basic/pom.xml
+++ b/testsuite/preview/basic/pom.xml
@@ -1,0 +1,540 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  Set these VM properties in your IDE debugger
+
+  -Djava.util.logging.manager=org.jboss.logmanager.LogManager
+  -Djboss.home=${workspace_loc:jboss-as-build}/target/jboss-as-8.0.0.Alpha4-SNAPSHOT
+  -DallowConnectingToRunningServer=true
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ts-preview</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>27.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <!-- ********************************************************************************** -->
+    <!-- ******************************** Basic Integration ******************************* -->
+    <!-- ********************************************************************************** -->
+    <artifactId>wildfly-ts-preview-basic</artifactId>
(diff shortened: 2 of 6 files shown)


bstansberry:
@yrodiere FYI

bstansberry:
I've moved this out of draft state as @yrodiere was able to base on it in #15185 and he hasn't told me this is garbage. :)

yrodiere:
Indeed this works great, thank you @bstansberry . However, would you consider merging https://github.com/bstansberry/wildfly/pull/60 in your branch before merging into main? I don't think we can use the `testsuite/preview` module without these dependency overrides.

bstansberry:
@yrodiere Duplicating the properties concerns me as they will for sure fall out of sync. I can see a couple solutions:

1) Add the org.wildfly:wildfly-ee-9-parent pom as an import scope dependency in testsuite/preview. For that to work though, ee-9/pom.xml's dependency management will need to have explicit dependency elements for all the deps where it is overriding the standard wildfly version.  Right now it does that for many/most of them (partly because for a while I thought it was necessary) but for others it just overrides the version property.

2) Create a new testsuite module under ee-9* and move this stuff there. The downside there is lot of things in testsuite/pom.xml would need to be duplicated.


*BTW at some point we should rename 'ee-9' to 'preview' as Preview is a long term thing, while 'ee-9' was just the initial name used before what Preview was and how it would be long term became apparent. That change will break a lot of PRs though so better to defer that to when the amount of activity for Preview calms down.

yrodiere:
> 1. Add the org.wildfly:wildfly-ee-9-parent pom as an import scope dependency in testsuite/preview. For that to work though, ee-9/pom.xml's dependency management will need to have explicit dependency elements for all the deps where it is overriding the standard wildfly version

I tried something like that when facing a similar problem in Hibernate Search a while ago, and I remember I couldn't get it to work because for some reason, dependency management in the parent POM was taking precedence over the imported POM. Maybe I did something wrong though; it's worth a try.

But in Hibernate Search, it didn't work and I ended up simply adding another level of indirection for version properties; see below.

> 2\. Create a new testsuite module under ee-9* and move this stuff there. The downside there is lot of things in testsuite/pom.xml would need to be duplicated.

Sure, that would work. Up to you to decide what you hate duplicating most :)

----

A third solution would be to define the Jakarta versions through yet another set of properties in the top-level pom (note the `.preview` suffix):

```xml
        <version.jakarta.ejb.jakarta-ejb-api.preview>4.0.0</version.jakarta.ejb.jakarta-ejb-api.preview>
        <version.jakarta.enterprise.preview>3.0.0</version.jakarta.enterprise>
        <version.jakarta.inject.jakarta.inject-api.preview>2.0.0</version.jakarta.inject.jakarta.inject-api.preview>
        <version.jakarta.persistence.preview>3.0.0</version.jakarta.persistence.preview>
        <version.jakarta.transaction.jakarta-transaction-api.preview>2.0.0</version.jakarta.transaction.jakarta-transaction-api.preview>
```

Then do the overrides in both `ee-9/pom.xml` and `testsuite/preview/pom.xml`, like this:

```xml
        <version.jakarta.ejb.jakarta-ejb-api>${version.jakarta.ejb.jakarta-ejb-api.preview}</version.jakarta.ejb.jakarta-ejb-api>
        <version.jakarta.enterprise>${version.jakarta.enterprise.preview}</version.jakarta.enterprise>
        <version.jakarta.inject.jakarta.inject-api>${version.jakarta.inject.jakarta.inject-api.preview}</version.jakarta.inject.jakarta.inject-api>
        <version.jakarta.persistence>${version.jakarta.persistence.preview}</version.jakarta.persistence>
        <version.jakarta.transaction.jakarta-transaction-api>${version.jakarta.transaction.jakarta-transaction-api.preview}</version.jakarta.transaction.jakarta-transaction-api>
```

Then the modules will always use the same version of the "preview" dependencies; the only risk is for either module to omit some properties overrides, but if that happens, hopefully the testsuite would fail (since we would be using `javax` packages either in the preview modules or in the testsuite).

PR #15175, opened 2022-02-01
WFLY-15603 Move WildFly Preview to a native Jakarta namespace variant of the MicroProfile OpenAPI subsystem module

pferraro:
https://issues.redhat.com/browse/WFLY-15603

Updates WF preview to a native jakarta version of io.smallrye:smallrye-open-api


pferraro:
Blech - let's do this one step at a time.

PR #15176, opened 2022-02-01
WFLY-15606 Move WildFly Preview to a native jakarta namespace variant of smallrye-open-api

pferraro:
https://issues.redhat.com/browse/WFLY-15606

PR #15150, opened 2022-01-26
WFLY-15556 Move WildFly Preview to a native Jakarta namespace variant of the wildfly-mod_cluster-undertow module

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-15556

Requires
https://github.com/wildfly/wildfly/pull/14966

PR #15180, opened 2022-02-02
[WFLY-15601] Move WildFly Preview to a native Jakarta namespace variant of the MicroProfile Health subsystem module

xstefank:
…nt of the MicroProfile Health subsystem module

https://issues.redhat.com/browse/WFLY-15601

### DIFF
--- a/ee-9/source-transform/microprofile/pom.xml
+++ b/ee-9/source-transform/microprofile/pom.xml
@@ -43,6 +43,7 @@
 
     <modules>
         <module>fault-tolerance-smallrye</module>
+        <module>health-smallrye</module>
         <module>openapi-smallrye</module>
         <module>reactive-streams-operators-smallrye/cdi-provider</module>
     </modules>
--- a/ee-9/common-microprofile/pom.xml
+++ b/ee-9/common-microprofile/pom.xml
@@ -60,6 +60,11 @@
         </dependency>
 
         <dependency>
+            <groupId>${full.maven.groupId}</groupId>
+            <artifactId>wildfly-microprofile-health-smallrye-jakarta</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>jose-jwt</artifactId>
             <exclusions>
--- a/ee-9/common-microprofile/src/main/resources/license/preview-feature-pack-common-microprofile-licenses.xml
+++ b/ee-9/common-microprofile/src/main/resources/license/preview-feature-pack-common-microprofile-licenses.xml
@@ -45,5 +45,16 @@
                 </license>
             </licenses>
         </dependency>
+        <dependency>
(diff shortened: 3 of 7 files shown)


ee-9/common-microprofile/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/health-smallrye/main/module.xml
```diff
@@ -0,0 +1,55 @@
...
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.wildfly.extension.microprofile.health-smallrye">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.wildfly:wildfly-microprofile-health-smallrye-jakarta}"/>
```
bstansberry, line 32: This file can be removed and a maven filter expression can be added to the existing file, a la https://github.com/wildfly/wildfly/pull/15186/files#diff-676b4930e63bfb3406b21496abd703b2d864fa286d1d8a54023552fa60d2e90aR32

This file is the same as that one except for the -jakarta suffix and using jakarta.* dependencies. But in WF Preview using jakarta.* deps is not needed because the javax.* modules are just aliases to the jakarta.* ones.

ee-9/source-transform/pom.xml
```diff
@@ -59,6 +59,7 @@
         <module>jsf/subsystem</module>
         <module>mail</module>
         <module>microprofile/reactive-streams-operators-smallrye/cdi-provider</module>
+        <module>microprofile/health-smallrye</module>
```
bstansberry, line 62: This *was* ok but a couple other PRs have https://github.com/wildfly/wildfly/pull/15176/files#diff-1375aa050fa94d63cc4058a8fbdd07c42da879866e881c4e4ec2ed1f0c4983daR512 in it which will introduce conflict. I'll likely end up merging one of those today.

bstansberry (changes requested):

bstansberry:
Note that I'll be merging #15150 and #15186 soon, so in case you are spending your Friday evening fixing conflicts, FYI that I'll likely add some more in a little while. ;)

bstansberry (approved):

ee-9/source-transform/microprofile/pom.xml
```diff
@@ -45,6 +45,7 @@
         <module>fault-tolerance-smallrye</module>
         <module>openapi-smallrye</module>
         <module>reactive-streams-operators-smallrye/cdi-provider</module>
+        <module>health-smallrye</module>
```
rhusar, line 48: Just a nitpick but this one is out of order, the general idea is to keep items in poms sorted alphabetically.

rhusar (approved):
LGTM

bstansberry:
Thanks, @xstefank

PR #15165, opened 2022-01-31
Move WildFly Preview to a native jakarta namespace variant of Hibernate ORM 6.0.0.Final

scottmarlow:
https://issues.redhat.com/browse/WFLY-15440
https://issues.redhat.com/browse/WFLY-15694
https://issues.redhat.com/browse/WFLY-16178

Signed-off-by: Scott Marlow <smarlow@redhat.com>

PR #15185, opened 2022-02-03
WFLY-15838 + WFLY-15839 + WFLY-15840 + WFLY-15841 + WFLY-15842 Upgrade to Hibernate Search 6 (EE 9 + ORM 6)

yrodiere:
* [WFLY-15842](https://issues.redhat.com/browse/WFLY-15842): Add (private) module for Elasticsearch's low-level REST client
* [WFLY-15841](https://issues.redhat.com/browse/WFLY-15841): Add (private) module for com.carrotsearch.hppc
* [WFLY-15840](https://issues.redhat.com/browse/WFLY-15840): Upgrade to Apache Avro 1.11.0
* [WFLY-15839](https://issues.redhat.com/browse/WFLY-15839): Upgrade to Apache Lucene 8.11
* [WFLY-15838](https://issues.redhat.com/browse/WFLY-15838): Upgrade to Hibernate Search 6.1

~Based on #15174 ([WFLY-15992](https://issues.redhat.com/browse/WFLY-15992)) and #15165  ([WFLY-15440](https://issues.redhat.com/browse/WFLY-15440)), which need to be merged first.~ => Done

Supersedes #15031.


yrodiere:
As far as I can see the failure in Linux - JDK 11 is the one about compat tests, which Scott mentioned in #15174. I didn't look at the other failures.

I ran the tests related to Hibernate Search locally though, and I found one failure. It's caused by a bug in Hibernate ORM 6: https://github.com/hibernate/hibernate-orm/pull/4777

scottmarlow:
@yrodiere fyi, https://github.com/scottmarlow/wildfly/tree/orm6_fixup2 has more updates to pass more tests (especially if you use the latest ORM 6.0 snapshot build).

yrodiere:
@scottmarlow Thanks for the heads-up, I just rebased. Let's see how it goes...

scottmarlow:
@yrodiere I think that ORM 6 (in JPA/Persistence compatible mode) needs queries to include a `SELECT alias`.  I updated some other tests from:
```
query = em.createQuery("from Employee e where e.id=:id");
```

To:
```
query = em.createQuery("select e from Employee e where e.id=:id");
```

And from:
```
query = em.createQuery("from Employee");
```

To:
```
query = em.createQuery("select e from Employee e");
```

And from:
```
        {@NamedQuery(name = "allCustomers", query = "from Customer"),
                @NamedQuery(name = "customerById", query = "from Customer c where c.id=:id")})
 public class Customer {
```

To:
```
       {@NamedQuery(name = "allCustomers", query = "select c from Customer c"),
               @NamedQuery(name = "customerById", query = "select c from Customer c where c.id=:id")})
 public class Customer {
```

scottmarlow:
If you see failures due to the CDI Bean Manager not being available, that is likely caused by a change to not enable CDI by default.  

https://github.com/eclipse-ee4j/jakartaee-tck/pull/874/ mentions you can workaround that a few ways like specifying that beans are Application scoped (and a few other ways mentioned in the pr).
@ApplicationScoped

yrodiere:
Thanks @scottmarlow , I rebased on your PR with your fixes, and also fixed CDI bean detection in Hibernate Search tests. Tests pass locally, let's see about CI...

scottmarlow:
@yrodiere I pushed more changes to https://github.com/wildfly/wildfly/pull/15165 including switching to ORM 6.0.0.Final.

yrodiere:
@scottmarlow Thanks for the heads-up, I rebased on your branch.

scottmarlow:
It looks like all of the test failures are for microprofile + clustering, I don't see any Hibernate test failures which is great!

yrodiere:
Ok, this should remove the last remaining failure related to Hibernate Search (5, in this case).

Do I need to do something about the gazilion failures related to microprofile + clustering? Or is this something that already fails on your PR or the main branch?

scottmarlow:
> Ok, this should remove the last remaining failure related to Hibernate Search (5, in this case).
> 
> Do I need to do something about the gazilion failures related to microprofile + clustering? Or is this something that already fails on your PR or the main branch?

No, you can ignore those failures as I don't believe they should be related to the Hibernate Search upgrade.

yrodiere:
Ok. It seems there are even fewer failures after rebasing on main; none are related to Hibernate Search.

This PR is ready as far as I'm concerned. Since your PR will have test failures until Hibernate Search gets upgraded, maybe you should integrate this PR into your own? So that we'll eventually merge the ORM 6 + Search 6 upgrades simultaneously.

scottmarlow:
> Ok. It seems there are even fewer failures after rebasing on main; none are related to Hibernate Search.
> 
> This PR is ready as far as I'm concerned. Since your PR will have test failures until Hibernate Search gets upgraded, maybe you should integrate this PR into your own? So that we'll eventually merge the ORM 6 + Search 6 upgrades simultaneously.

@bstansberry @darranl do you have a preference for whether https://github.com/wildfly/wildfly/pull/15165 + https://github.com/wildfly/wildfly/pull/15185 are merged as one or separate pull requests?

https://github.com/wildfly/wildfly/pull/15165 does not fix one of the Hibernate Search test failures but that will be addressed by https://github.com/wildfly/wildfly/pull/15185.

scottmarlow:
@yrodiere https://github.com/wildfly/wildfly/pull/15165/commits/108f01332cb82bca82d6d3db6819942e3fc287b2#diff-74bfacd78cff4d8815bf930dda24f8402d5b0d2741ae39de85a355ff243da292 adds org.junit.Ignore to disable the  testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/hibernate/search/HibernateSearchJPATestCase.java Search test (see comment `Ignore until we upgrade to Hibernate Search 6.1 via WFLY-15838`).

After https://github.com/wildfly/wildfly/pull/15165 is merged and you sync up, could you please remove the @Ignore from testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/hibernate/search/HibernateSearchJPATestCase.java since your pull request will allow that test to work again.

Thanks

yrodiere:
@scottmarlow Thanks for the heads-up. I rebased on your branch and removed the `@Ignore` annotation.

yrodiere:
Hi @scottmarlow, I see your PR (#15165) was merged, so I rebased this one on main; it's ready for review. Please let me know if I need to change anything!

The only test failures before my rebase seemed unrelated (something about JMS and something about the WildFly client); I don't expect anything different after the rebase, since I didn't change anything. But let's see.

ee-9/common/src/main/resources/modules/system/layers/base/org/hibernate/search/backend/elasticsearch/main/module.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2021, Red Hat, Inc., and individual contributors
```
scottmarlow, line 5: Please update to 2022.

ee-9/common/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2013, Red Hat, Inc., and individual contributors
```
scottmarlow, line 5: Please update all of the added (new) modules to have 2022 as date.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernatePersistenceProviderAdaptor.java
```diff
@@ -92,9 +92,6 @@ public void addProviderProperties(Map properties, PersistenceUnitMetadata pu) {
 
         // Enable JPA Compliance mode
         putPropertyIfAbsent( pu, properties, AvailableSettings.JPA_COMPLIANCE, true);
-
-        // Search hint
-        putPropertyIfAbsent( pu, properties,"hibernate.search.index_uninverting_allowed", "true" );
```
scottmarlow, line 97: :+1: as per https://docs.jboss.org/hibernate/search/6.0/migration/html_single/

> hibernate.search.index_uninverting_allowed
> 
>     Index uninverting was deprecated in Hibernate Search 5 due to poor performance and is no longer allowed. All index fields that you want to sort on must be [marked as sortable](https://docs.jboss.org/hibernate/search/6.0/migration/html_single/#sortablefield).

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/HibernateSearchProcessor.java
```diff
@@ -77,42 +83,67 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
-                moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, defaultSearchModule, false, true, true, false));
-                ROOT_LOGGER.debugf("deployment %s contains %s annotation, added %s dependency", deploymentUnit.getName(), SEARCH_INDEXED_ANNOTATION_NAME, defaultSearchModule);
+                moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, MODULE_MAPPER_ORM_DEFAULT,
+                        false, true, true, false));
+                ROOT_LOGGER.debugf("deployment %s contains %s annotation, added %s dependency", deploymentUnit.getName(),
+                        ANNOTATION_INDEXED_NAME, MODULE_MAPPER_ORM_DEFAULT);
+            }
+            else {
+                // Hibernate Search 5.x support
+                // TODO Remove this when we switch to EE 10 and remove Hibernate Search 5.x modules
```
scottmarlow, line 127: Should there be a single tracking WFLY jira for the various TODOs?

yrodiere: Sure. I created https://issues.redhat.com/browse/WFLY-16362 and referenced it in every TODO.

yrodiere:
Thanks @scottmarlow. I addressed all your comments (copyright years, JIRA ticket for TODOs). Please let me know if there's anything else.

scottmarlow (approved):

bstansberry:
@yrodiere The Linux Security Manager CI job failures are due to https://issues.redhat.com/browse/WFCORE-5908. Why this PR has surfaced this bug needs investigation though. The failure indicates an earlier test (org.jboss.as.test.integration.ejb.mdb.vaultedproperties.MDBWithVaultedPropertiesTestCase) is not properly cleaning up after itself, but we only see that when testing this PR.

testsuite/integration/basic/pom.xml
```diff
@@ -3418,6 +3418,44 @@
...
+                                    <excludes>
+                                        <!-- Hibernate Search is not compatible with the security manager. Never has been. -->
+                                        <exclude>org/jboss/as/test/integration/hibernate/search/**</exclude>
+                                    </excludes>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>basic-integration-default-web.surefire</id>
+                                <configuration>
+                                    <excludes>
```
bstansberry, line 3447: I believe adding 'combine.children="append"' will fix the CI failure. This profile is superseding the entire list of excludes from this execution instead of just adding to it.

bstansberry (changes requested):

testsuite/integration/basic/pom.xml
bstansberry: Never mind; that doesn't work.

There are only 3 tests involved here so it's better to drop the pom stuff and add @BeforeClass methods that call org.jboss.as.test.shared.util.AssumeTestGroupUtil.assumeSecurityManagerDisabled()

yrodiere: > There are only 3 tests involved here so it's better to drop the pom stuff and add @BeforeClass methods that call org.jboss.as.test.shared.util.AssumeTestGroupUtil.assumeSecurityManagerDisabled()

Done; thanks for the hint, I was looking for something like this :)

yrodiere:
Thanks @bstansberry , I applied the changes you requested and the previously failing tests now pass. I think we're good to merge?

bstansberry (commented):
@yrodiere Looks good, but I want to confirm that the jboss-api settings in the various module.xml files are as expected. Most are not set, making the module public API.

(Note that even if things are public in WF, EAP can use jboss-api="unsupported" downstream if it doesn't want to support direct use of a module.)

yrodiere:
@bstansberry I confirm jboss-api settings are as intended.

To clarify the intent:

* The following modules are completely internal and not expected to appear in any public API, thus they are declared as private:
  * `com.carrotsearch.hppc`
  * `org.apache.avro`
* The following modules can be useful to users of Hibernate Search, but direct use is not expected to be supported in EAP. **We could declare them as public in WildFly and unsupported in EAP, if you prefer**.
  * `org.apache.lucene`
  * `org.elasticsearch.client.rest-client`
  * `com.google.code.gson`
* The following modules contain public Hibernate Search APIs that are absolutely necessary to Hibernate Search users, and thus they are public.
  * `org.hibernate.search.engine`
  * `org.hibernate.search.mapper.orm`
  * `org.hibernate.search.mapper.pojo-base`
* The backend modules also contain public Hibernate Search APIs that are absolutely necessary to Hibernate Search users, but they include APIs that involve types from private/unsupported modules (see above), however; those methods won't be usable unless users explicitly import unsupported/private modules (`org.apache.lucene` or `org.elasticsearch.client.rest-client`). This is as intended. See for example `org.jboss.as.test.integration.hibernate.search.backend.lucene.extension.HibernateSearchLuceneDependencyTestCase` or ``org.jboss.as.test.integration.hibernate.search.backend.elasticsearch.extension.HibernateSearchElasticsearchGsonDependencyTestCase`.`
  * `org.hibernate.search.backend.elasticsearch`
  * `org.hibernate.search.backend.lucene`
* Some artifacts don't even have a JBoss module because they are considered incubating or should not be used in production environments. ***We could add modules declared as public or private in WildFly and unsupported in EAP, if you prefer****. See for example `org.jboss.as.test.integration.hibernate.search.v5migrationhelper.simple.HibernateSearchV5MigrationHelperTestCase` or `org.jboss.as.test.integration.hibernate.search.coordination.HibernateSearchOutboxPollingTestCase`.
  * `hibernate-search-mapper-orm-coordination-outbox-polling`
  * `hibernate-search-v5migrationhelper-engine`
  * `hibernate-search-v5migrationhelper-orm`

bstansberry (approved):
@yrodiere Thanks so much for such a thorough answer! If you don't mind I'll copy that into a comment in WFLY-15838 as it's excellent design documentation and will remain more visible there.

yrodiere:
No problem, thanks for merging :)

PR #15191, opened 2022-02-04
WFLY-15541 Eliminate need for source transformation of wildfly-clustering-ejb-infinispan

pferraro:
https://issues.redhat.com/browse/WFLY-15541

This module only depended on javax.ejb.api for an exception.  This is easy enough to workaround, and thus avoid the need to source transform this module.


bstansberry:

chengfang (approved):

PR #15192, opened 2022-02-05
[WFLY-16005] Reexport org.jboss.ironjacamar.api module from org.jboss.as.connector module

ropalka:
https://issues.redhat.com/browse/WFLY-16005

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/workmanager/LongRunningThreadsCheckTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/workmanager/LongRunningThreadsCheckTestCase.java
@@ -140,7 +140,7 @@ public class LongRunningThreadsCheckTestCase {
         ja.addPackage(MultipleConnectionFactory1.class.getPackage()).addClasses(LongRunningThreadsCheckTestCase.class,
                 JcaMgmtServerSetupTask.class, JcaMgmtBase.class, JcaTestsUtil.class);
         ja.addPackage(AbstractMgmtTestBase.class.getPackage());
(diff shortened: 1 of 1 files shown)


bstansberry:
@ropalka I don't understand why a JPMS related change is relevant to an application (i.e. WF) running in classpath mode doing things that don't involve access to JDK modules.

I think the fix here is to have to test deployment add a dependency on the necessary modules. The org.jboss.as.connector module is not a public module and what it exposes shouldn't be modified to accommodate test cases poking into internals.  There are other subsystem modules that depend on org.jboss.as.connector, but in an ideal world whatever they use would be factored out into a separate module, or in some other way what they can access would be limited. In an ideal world the only things exposed by an extension module would be things needed by the management kernel to install it as an extension.

ropalka:
> @ropalka I don't understand why a JPMS related change is relevant to an application (i.e. WF) running in classpath mode doing things that don't involve access to JDK modules.

I'll try to explain @bstansberry . It is relevant because this test is executed in JBoss Modules environment which enforces the boundaries. Before the JDK change ( https://github.com/openjdk/jdk18/commit/a5080effc7ec7e260e84e3169c36c5217f18d231 ) the generated byte code used Object.equals(Object) via InterfaceMethodRef and since java.base is visible by default to all modules it worked. After the change in the compiler the generate byte code is changed to org.jboss.jca.core.api.workmanager.StatisticsExecutor.equals(Object) InterfaceMethodRef and therefore JBoss Modules boundaries will be applied.

Of course it is also possible to fix that the other way i.e. by adding the org.jboss.ironjacamar.api module into test deployment war. Let me know which you prefer more. You can see my decision in the last comment of the JIRA why I decided for the current solution which I think is the correct one.

ropalka:
I updated the test as you suggested @bstansberry .

bstansberry (approved):
Thanks, @ropalka

PR #15209, opened 2022-02-09
[WFLY-15263] Adjust SecurityAuditingTestCase to use an Elytron configuration

fjuma:
https://issues.redhat.com/browse/WFLY-15263

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/auditing/SecurityAuditingTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/auditing/SecurityAuditingTestCase.java
@@ -22,10 +22,10 @@
 
 package org.jboss.as.test.integration.security.auditing;
 
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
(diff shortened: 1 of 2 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/auditing/SecurityAuditingTestCase.java
```diff
@@ -69,50 +72,58 @@
...
 @ServerSetup({SecurityAuditingTestCase.SecurityAuditingTestCaseSetup.class})
-@Ignore("[WFLY-15263] Update for lack of legacy security.")
 public class SecurityAuditingTestCase extends AnnSBTest {
 
     private static final Logger log = Logger.getLogger(testClass());
 
     private static File auditLog = new File(System.getProperty("jboss.home", null), "standalone" + File.separator + "log" + File.separator + "audit.log");
 
-    static class SecurityAuditingTestCaseSetup extends AbstractMgmtServerSetupTask {
+    private static final boolean USE_ELYTRON = System.getProperty("elytron") != null;
```
bstansberry, line 81: No need to update this PR just for this, but I assume this line is cruft.

fjuma: Yeah, good catch, that line isn't needed here. I've removed it.

bstansberry (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/security/auditing/SecurityAuditingTestCase.java
bstansberry: Thanks. I'm glad I asked as something I did later led to https://github.com/wildfly/wildfly/pull/15211. :)

PR #15219, opened 2022-02-11
[WFLY-16036] Use default security domain for EJBs when no default specified

skyllarr:
https://issues.redhat.com/browse/WFLY-16036

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
@@ -80,6 +80,7 @@ import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
 import org.wildfly.extension.requestcontroller.ControlPoint;
+import org.wildfly.security.auth.principal.AnonymousPrincipal;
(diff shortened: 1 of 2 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
```diff
@@ -274,7 +275,7 @@ public Principal getCallerPrincipal() {
         }
 
         // TODO Should this return a Principal when security is not activated.
```
fjuma, line 277: Just minor, but we can remove this TODO comment now. :)

fjuma (approved):
Thanks @Skyllarr!

darranl (approved):

bstansberry:
Thanks, @Skyllarr

ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponent.java
rhusar: Looks like this got merged with the TODO still in place :-P

Skyllarr: @rhusar I submitted PR that removes the TODO comment https://github.com/wildfly/wildfly/pull/15238 :tada:

PR #15221, opened 2022-02-14
WFLY-16040 Upgrade jberet-core from 1.3.10.Final to 1.3.10.SP1

chengfang:
https://issues.redhat.com/browse/WFLY-16040

This PR is to get the bug fix for [JBERET-537](https://issues.redhat.com/browse/JBERET-537) into WildFly before porting to downstream. It competes with https://github.com/wildfly/wildfly/pull/14952, which will upgrade jberet-core to a newer version (jberet-core 1.3.11.Final). It will be good for this PR to be considered first.


bstansberry:
Thanks, @chengfang

PR #15222, opened 2022-02-14
[WFLY-16033]: Reading a bridge with include-runtime throws WFLYCTL0216.

ehsavoie:
When a bridge deployment failed, Wildfly wasn't notified and still  provided the resource.
Now it will fail the add operation.

Jira: https://issues.redhat.com/browse/WFLY-16033

### DIFF
--- a/messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/logging/MessagingLogger.java
+++ b/messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/logging/MessagingLogger.java
@@ -882,4 +882,7 @@ public interface MessagingLogger extends BasicLogger {
     @Message(id = 105, value = "The %s %s is configured to use socket-binding %s, but this socket binding doesn't have the multicast-address or a multicast-port attributes configured.")
     OperationFailedException socketBindingMulticastNotSet(String resourceType, String resourceName, String socketBindingName);
 
+    @Message(id = 106, value = "The bridge %s didn't deploy.")
+    OperationFailedException failedBridgeDeployment(String bridgeName);
+
 }
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AbstractActiveMQComponentControlHandler.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AbstractActiveMQComponentControlHandler.java
@@ -43,7 +43,6 @@ import org.jboss.as.controller.SimpleAttributeDefinition;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.ResourceDescriptionResolver;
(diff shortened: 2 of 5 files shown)


messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/logging/MessagingLogger.java
```diff
@@ -882,4 +882,7 @@
     @Message(id = 105, value = "The %s %s is configured to use socket-binding %s, but this socket binding doesn't have the multicast-address or a multicast-port attributes configured.")
     OperationFailedException socketBindingMulticastNotSet(String resourceType, String resourceName, String socketBindingName);
 
+    @Message(id = 106, value = "The bridge %s didn't deploy.")
+    OperationFailedException failedBridgeDeployment(String bridgeName);
```
bstansberry, line 886: This doesn't sound like an OFE. An OFE indicates a user mistake, not a server fault.

bstansberry: AbstractActiveMQComponentControlHandler.getActiveMQComponentControl needs to throw a different exception.

A Stage.RUNTIME failure adding a resource should not result in removing the resource unless the operation is configured to rollback-on-runtime-failure. So throwing a runtime exception if the Artemis stuff can't be deployed is correct. If the code that calls this method isn't failing but is ignoring a problem, it's fine to have it throw a RuntimeException, but not an OFE, unless somehow you know what the user mistake was (if so, the op should fail in Stage.MODEL).

So, the add can succeed (e.g. at boot, which is rollback-on-runtime-failure=false), but the Artemis control is not there. That means the call to AbstractActiveMQComponentControlHandler.getActiveMQComponentControl might fail. But that failure shouldn't be ControllerLogger.ROOT_LOGGER.managementResourceNotFound(address), because the management resource does exist.

ehsavoie: Thanks for the clarification.

ehsavoie: @bstansberry I think there is an issue with the current code as the rollback will occur  even with rollback-on-runtime-failure=false with a RuntimeException

https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L1163 will always return ROLLBACK when accessed from https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractOperationContext.java#L1086 which is where we land if we throw an Exception in the performRuntime
I don't see how the add can succeed, but I might have misunderstood your comment

bstansberry: @ehsavoie if you change the exception back to OFE we can get this one merged.

bstansberry:
Thanks @ehsavoie

PR #15233, opened 2022-02-16
[WFLY-16050] Improve error message for duplicate components

jasondlee:
https://issues.redhat.com/browse/WFLY-16050

Add new error message for duplicate services (Exceptions thrown are caught and handled back up the chain, so logging is likely the best we can do here without major changes)

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
@@ -22,11 +22,15 @@
 
 package org.jboss.as.ee.component.deployers;
 
+import static org.jboss.as.ee.component.Attachments.COMPONENT_REGISTRY;
+import static org.jboss.as.ee.component.Attachments.EE_MODULE_CONFIGURATION;
+import static org.jboss.as.ee.logging.EeLogger.ROOT_LOGGER;
+import static org.jboss.as.server.deployment.Attachments.MODULE;
+
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.jboss.as.ee.logging.EeLogger;
 import org.jboss.as.ee.component.Attachments;
 import org.jboss.as.ee.component.BasicComponent;
 import org.jboss.as.ee.component.BasicComponentCreateService;
@@ -46,6 +50,7 @@ import org.jboss.as.ee.component.InjectionSource;
 import org.jboss.as.ee.component.InterceptorDescription;
 import org.jboss.as.ee.component.ViewConfiguration;
 import org.jboss.as.ee.component.ViewService;
+import org.jboss.as.ee.logging.EeLogger;
 import org.jboss.as.ee.metadata.MetadataCompleteMarker;
 import org.jboss.as.naming.ManagedReferenceFactory;
 import org.jboss.as.naming.ServiceBasedNamingStore;
@@ -68,11 +73,6 @@ import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 
-import static org.jboss.as.ee.logging.EeLogger.ROOT_LOGGER;
-import static org.jboss.as.ee.component.Attachments.COMPONENT_REGISTRY;
-import static org.jboss.as.ee.component.Attachments.EE_MODULE_CONFIGURATION;
-import static org.jboss.as.server.deployment.Attachments.MODULE;
-
 /**
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  * @author Eduardo Martins
@@ -80,6 +80,26 @@ import static org.jboss.as.server.deployment.Attachments.MODULE;
 public final class ComponentInstallProcessor implements DeploymentUnitProcessor {
 
     private static final ServiceName JNDI_BINDINGS_SERVICE = ServiceName.of("JndiBindingsService");
+
+    private static final List<String> SPEC_COMPONENTS = List.of(
+            "BeanManager",
+            "DefaultContextService",
+            "DefaultDataSource",
+            "DefaultJMSConnectionFactory",
+            "DefaultManagedExecutorService",
+            "DefaultManagedScheduledExecutorService",
+            "DefaultManagedThreadFactory",
+            "EJBContext",
+            "HandleDelegate",
+            "InAppClientContainer",
+            "InstanceName",
+            "ModuleName",
+            "ORB",
+            "TimerService",
+            "TransactionSynchronizationRegistry",
+            "Validator",
+            "ValidatorFactory"
+    );
 
     public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
         final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
@@ -193,7 +213,13 @@ public final class ComponentInstallProcessor implements DeploymentUnitProcessor 
                 ServiceBuilder<ManagedReferenceFactory> serviceBuilder = serviceTarget.addService(bindInfo.getBinderServiceName(), service);
                 bindingConfiguration.getSource().getResourceValue(resolutionContext, serviceBuilder, phaseContext, service.getManagedObjectInjector());
                 serviceBuilder.addDependency(bindInfo.getParentContextServiceName(), ServiceBasedNamingStore.class, service.getNamingStoreInjector());
-                serviceBuilder.install();
+                try {
+                    serviceBuilder.install();
+                } catch (DuplicateServiceException e) {
+                    handleDuplicateService(configuration, bindInfo.getAbsoluteJndiName());
+                    throw e;
+                }
+
             }
         }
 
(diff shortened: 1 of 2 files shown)


jasondlee:
@bstansberry I'm changing really old code here. Definitely need some careful scrutiny.

bstansberry:
@jasondlee I don't think this would help much. AIUI the information being logged is already being logged, just in a different format. This might be a bit more helpful format, but OTOH it's more logging, so not a big gain. Maybe I'm wrong. What would the full (new + existing) logging look like?

Can you expand the WFLY-16050 description to better explain the problem being fixed? My impression is it was about duplicate registration of java:module/Validator, which is due to an EE spec requirement that we bind that. If so shouldn't the improved error message discuss that?

jasondlee:
Yup. That's fair. I think the problem we're going to run into is the `Exception` is caught, then wrapped and rethrown as `DeploymentUnitProcessingException`. Eventually, *that* will get caught in `DeploymentUnitPhaseService` and wrapped/rethrown as  a `StartException`, which is logged in `ServiceControllerImpl.startFailed()`.

The wrapping and rethrowing seems to be pretty generic and meant to handle a wide variety of failures. We could log more verbosely at the point of failure, but it is the `StartException` -- miles away from the point of failure -- that finally does the logging we're seeing. I'm not sure what we could do about that, or how wise special-case-handling might be at that point. 

You're (likely) more familiar with the code paths in question, and I realize you're probably not super familiar with the exact issue, but am I missing something?

bstansberry:
I don't think you should try and do anything about the top level stack trace (i.e. excluding Caused by stuff) of the StartException, or do anything about catching failures in DeploymentUnitPhaseService. What you have here seems like the right place to catch the failure and improve what we communicate.

Adding a log message and then logging the exception as we do now is a valid solution; I'm just not sure whether what this change would log adds enough value. An example of the logging would help to clarify that.

Thanks for the update to the JIRA description. That does spell out that this is about spec defined bindings, which are limited in number, so that does open up the option of specifically handling those duplicate bindings.

jasondlee:
So should the logging just be made more explicit? Explaining that it's possibly due to a collision with a spec-mandated bean? Alternatively, we generate a list of spec mandate bean names and check against that, logging appropriately? Maintaining such a list is likely to be error-prone, but I guess the specs don't change that often (which can either make it less likely that a new bean will be introduced, or make it so that it happens so infrequently that we forget to update the list :).

bstansberry:
My instinct is checking an explicit list would be fine.

I didn't go through the EE platform spec with a fine-toothed comb but IIRC Validator was the only one I saw!

Maybe individual API specs mandate more of these things but I suspect it's just what's in the EE platform spec.

bstansberry:
Hmm, there may be more of these than I thought:

https://gist.github.com/bstansberry/ecb504c4212f061c2deb64a00430b5e7

That gist includes a lot of things not directly relevant as it is all BinderService services, not just java:module or java:comp.

bstansberry:
There's 15 which isn't too bad.

BeanManager
DefaultContextService
DefaultDataSource
DefaultJMSConnectionFactory
DefaultManagedExecutorService
DefaultManagedScheduledExecutorService
DefaultManagedThreadFactory
HandleDelegate
InAppClientContainer
InstanceName
ModuleName
ORB
TransactionSynchronizationRegistry
Validator
ValidatorFactory

bstansberry:
and maybe EJBContext and TimerService

jasondlee:
OK. Where are you finding these? I'd like to give it extra eyes :)

bstansberry:
@jasondlee They are in gist I linked. I just focused on the java.comp and java.module ones.

jasondlee:
@bstansberry Pushed changes with PR (and chat) feeback

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
```diff
@@ -193,7 +213,15 @@ protected void deployComponent(final DeploymentPhaseContext phaseContext, final
                 ServiceBuilder<ManagedReferenceFactory> serviceBuilder = serviceTarget.addService(bindInfo.getBinderServiceName(), service);
                 bindingConfiguration.getSource().getResourceValue(resolutionContext, serviceBuilder, phaseContext, service.getManagedObjectInjector());
                 serviceBuilder.addDependency(bindInfo.getParentContextServiceName(), ServiceBasedNamingStore.class, service.getNamingStoreInjector());
-                serviceBuilder.install();
+                try {
+                    serviceBuilder.install();
+                } catch (Exception e) {
+                    if (e instanceof DuplicateServiceException) {
+                        handleDuplicateService(configuration, bindingName);
```
bstansberry, line 220: I believe 'bindingName' is just part of the JNDI name, although I could be wrong. Logging the absolute JNDI name is better. I can be gotten from bindInfo.getAbsoluteJndiName().

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
```diff
@@ -68,9 +64,13 @@
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 
-import static org.jboss.as.ee.logging.EeLogger.ROOT_LOGGER;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
```
bstansberry, line 69: Since you foolishly reordered imports you get to fix the existing incorrect order. ;)

Our code style is:

import static blah.blah  (with no wildcards)

import java.blah
import javax.blah

import other.blah

bstansberry (changes requested):

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
```diff
@@ -193,7 +213,15 @@ protected void deployComponent(final DeploymentPhaseContext phaseContext, final
                 ServiceBuilder<ManagedReferenceFactory> serviceBuilder = serviceTarget.addService(bindInfo.getBinderServiceName(), service);
                 bindingConfiguration.getSource().getResourceValue(resolutionContext, serviceBuilder, phaseContext, service.getManagedObjectInjector());
                 serviceBuilder.addDependency(bindInfo.getParentContextServiceName(), ServiceBasedNamingStore.class, service.getNamingStoreInjector());
-                serviceBuilder.install();
+                try {
+                    serviceBuilder.install();
+                } catch (Exception e) {
+                    if (e instanceof DuplicateServiceException) {
```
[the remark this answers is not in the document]
jasondlee: Yup. You're right. I moved that code and didn't clean it up.

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
jasondlee: I blame IDEA for that. I'll fix this, then try to figure out how to make IDEA behave.

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
jasondlee: Changed.

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -1210,4 +1207,10 @@
 
     @Message(id = 132, value = "ManagedReference was null and injection is not optional for injection into method %s")
     RuntimeException managedReferenceMethodWasNull(Method method);
+
+    @LogMessage(level = ERROR)
+    @Message(id = 133, value = "A service with the name '%s' has already been installed as '%s' in accordance with Jakarta EE " +
```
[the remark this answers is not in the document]
jasondlee: I have updated the message and changed the method name.

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
jasondlee: Fixed.

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -61,6 +46,18 @@
...
+import javax.xml.stream.Location;
+import javax.xml.stream.XMLStreamException;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Collection;
+import java.util.Set;
+import java.util.concurrent.RejectedExecutionException;
+
+import static org.jboss.logging.Logger.Level.*;
```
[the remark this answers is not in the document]
jasondlee: Fixed

ee/src/main/java/org/jboss/as/ee/component/deployers/ComponentInstallProcessor.java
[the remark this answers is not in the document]
jasondlee: Fixed

bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #15237, opened 2022-02-17
[WFLY-16052] Remove testsuite use of org.hibernate.stat.Statistics.ge…

bstansberry:
…tSecondLevelCacheStatistics

https://issues.redhat.com/browse/WFLY-16052

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/secondlevelcache/JPA2LCTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/secondlevelcache/JPA2LCTestCase.java
@@ -43,7 +43,6 @@ import javax.sql.DataSource;
 import java.sql.Connection;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -57,8 +56,7 @@ public class JPA2LCTestCase {
 
     private static final String ARCHIVE_NAME = "jpa_SecondLevelCacheTestCase";
 
-    // cache region name prefix, use getCacheRegionName() method to get the value!
-    private static String CACHE_REGION_NAME = null;
+    private static final String CACHE_REGION_NAME = JPA2LCTestCase.class.getPackage().getName() + '.';;
 
     @Deployment
     public static Archive<?> deploy() {
@@ -89,24 +87,6 @@ public class JPA2LCTestCase {
     protected <T> T rawLookup(String name, Class<T> interfaceType) throws NamingException {
         return interfaceType.cast(iniCtx.lookup(name));
     }
-
-    // Cache region name depends on the internal entity cache naming convention:
-    // "fully application scoped persistence unit name" + "the entity class full name"
-    // first part could be rewritten by property "hibernate.cache.region_prefix"
-    // This method returns prefix + package name, the entity name needs to be appended
-    public String getCacheRegionName() throws Exception {
-
-        if (CACHE_REGION_NAME == null) {
-            SFSB2LC sfsb = lookup("SFSB2LC", SFSB2LC.class);
-            String prefix = sfsb.getCacheRegionName();
-
-            assertNotNull("'hibernate.cache.region_prefix' is null.", prefix);
-            CACHE_REGION_NAME = prefix + '.' + this.getClass().getPackage().getName() + '.';
-        }
-
-        return CACHE_REGION_NAME;
-    }
-
 
     @Test
     @InSequence(1)
@@ -177,7 +157,7 @@ public class JPA2LCTestCase {
     @InSequence(3)
     public void testEntityCacheSameSession() throws Exception {
         SFSB2LC sfsb = lookup("SFSB2LC", SFSB2LC.class);
-        String message = sfsb.sameSessionCheck(getCacheRegionName());
+        String message = sfsb.sameSessionCheck(CACHE_REGION_NAME);
         if (!message.equals("OK")) {
             fail(message);
         }
@@ -194,7 +174,7 @@ public class JPA2LCTestCase {
     public void testEntityCacheSecondSession() throws Exception {
 
         SFSB2LC sfsb = lookup("SFSB2LC", SFSB2LC.class);
-        String message = sfsb.secondSessionCheck(getCacheRegionName());
+        String message = sfsb.secondSessionCheck(CACHE_REGION_NAME);
 
         if (!message.equals("OK")) {
             fail(message);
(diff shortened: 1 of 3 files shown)


bstansberry:
@scottmarlow Please review.

rhusar (changes requested):
Nope, this needs more changes. Currently fails with:


&amp#27;&amp#27;21:45:45,347 ERROR [org.jboss.as.ejb3.invocation] (default task-1) WFLYEJB0034: Jakarta Enterprise Beans Invocation failed on component StatefulBean for method public abstract void org.jboss.as.test.clustering.cluster.ejb.xpc.bean.Stateful.echo(java.lang.String): jakarta.ejb.EJBException: java.lang.IllegalArgumentException: Unknown cache region : stateful.war#mypc.org.jboss.as.test.clustering.cluster.ejb.xpc.bean.Employee

scottmarlow:
> Nope, this needs more changes. Currently fails with:
> 
> &amp#27;&amp#27;21:45:45,347 ERROR [org.jboss.as.ejb3.invocation] (default task-1) WFLYEJB0034: Jakarta Enterprise Beans Invocation failed on component StatefulBean for method public abstract void org.jboss.as.test.clustering.cluster.ejb.xpc.bean.Stateful.echo(java.lang.String): jakarta.ejb.EJBException: java.lang.IllegalArgumentException: Unknown cache region : [stateful.war#mypc.org.jboss.as.test.clustering.cluster.ejb.xpc.bean.Employee](https://github.com/wildfly/wildfly/pull/15237#pullrequestreview-886185326)

Hmm, that is odd, it sounds like the Hibernate ORM 5.3 Statistics.getSecondLevelCacheStatistics(String regionName) is expecting the fully scoped region name but I'm guessing Statistics.getDomainDataRegionStatistics(String regionName) doesn't.

In ORM 5.3, I see the following for StatisticsImpl:
```
	@Override
	public CacheRegionStatisticsImpl getSecondLevelCacheStatistics(String regionName) {
		if ( sessionFactory == null ) {
			return null;
		}
		return getCacheRegionStatistics( sessionFactory.getCache().unqualifyRegionName( regionName ) );
	}

	public CacheRegionStatisticsImpl getDomainDataRegionStatistics(String regionName) {
		if ( sessionFactory == null ) {
			return null;
		}

		return l2CacheStatsMap.getOrCompute(
				regionName,
				s -> {
					final Region region = sessionFactory.getCache().getRegion( s );

					if ( region == null ) {
						throw new IllegalArgumentException( "Unknown cache region : " + s );
					}

					if ( region instanceof QueryResultsRegion ) {
						throw new IllegalArgumentException(
								"Region name [" + s + "] referred to a query result region, not a domain data region"
						);
					}

					return new CacheRegionStatisticsImpl( region );
				}
		);
	}
```

for ORM 6.0 (main branch), I see:
```
	public CacheRegionStatisticsImpl getDomainDataRegionStatistics(String regionName) {
		return l2CacheStatsMap.getOrCompute(
				regionName,
				this::instantiateCacheRegionStatistics
		);
	}

```

It looks like org.hibernate.cache.spi.CacheImplementor.unqualifyRegionName() was deprecated in ORM 5.3 for removal along with Statistics#getSecondLevelCacheStatistics(), which I think means we will need to change the cache region name that we pass in to  ORM 6.0 Statistics#getDomainDataRegionStatistics() to not include the (now internal only) application name, so instead of `stats.getDomainDataRegionStatistics(CACHE_REGION_NAME + "Employee")` we should just have `stats.getDomainDataRegionStatistics("Employee")`.

bstansberry:
@scottmarlow @rhusar Hopefully fixed now.

bstansberry:
Note that https://github.com/wildfly/wildfly/pull/15237/files#diff-8af7fc2a4bbddac885f35c1afebb6b7bed71417d98ecf6029515ead9a7fd6648L100-R99 is kind of meaningless, as that code was only used in tests that have been ignored (which is why the test didn't fail with problems like the clustering one.) I fixed it though just so it won't fail from that if ever unignored.

scottmarlow:
FYI, the original code `stats.getDomainDataRegionStatistics(CACHE_REGION_NAME + "Employee")` reflects that each Hibernate second level cache is identified by a unique identifier that is comprised of the application deployment name (e.g. myapplication.war) with the persistence unit name concatenated (with a `#` separator in between).  The original Hibernate ORM implementation expected the full internal cache region name to be passed in which made sense as there could be collisions between different persistence units.  

Now that ORM 6 has removed the getSecondLevelCacheStatistics(String cacheRegionName) method which accepted the full name that is used internally to find the right second level cache, we will rely on the shorter name.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/xpc/bean/StatefulBean.java
```diff
@@ -166,9 +168,12 @@ private void logStats(String methodName) {
         session.getSessionFactory().getStatistics().setStatisticsEnabled(true);
         session.getSessionFactory().getStatistics().logSummary();
         String[] entityRegionNames = session.getSessionFactory().getStatistics().getSecondLevelCacheRegionNames();
-        for (String name : entityRegionNames) {
+        String legacyPrefix = session.getSessionFactory().getSessionFactoryOptions().getCacheRegionPrefix() + ".";
+        for (String qualifiedName : entityRegionNames) {
+            log.trace("qualified cache entity region name = " + qualifiedName);
+            String name = qualifiedName.startsWith(legacyPrefix) ? qualifiedName.substring(legacyPrefix.length()) : qualifiedName;
```
bstansberry, line 174: @scottmarlow I did this startsWith check to make the test robust, i.e. it works when Hibernate 5.3 getSecondLevelCacheRegionNames includes the prefix, but also works if a later version does not.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/secondlevelcache/JPA2LCTestCase.java
```diff
@@ -97,11 +96,7 @@
     public String getCacheRegionName() throws Exception {
 
         if (CACHE_REGION_NAME == null) {
-            SFSB2LC sfsb = lookup("SFSB2LC", SFSB2LC.class);
-            String prefix = sfsb.getCacheRegionName();
-
-            assertNotNull("'hibernate.cache.region_prefix' is null.", prefix);
-            CACHE_REGION_NAME = prefix + '.' + this.getClass().getPackage().getName() + '.';
+            CACHE_REGION_NAME = this.getClass().getPackage().getName() + '.';
```
scottmarlow, line 99: We could make `CACHE_REGION_NAME` final and statically assign the value.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/xpc/bean/StatefulBean.java
scottmarlow: :+1:

scottmarlow (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/secondlevelcache/JPA2LCTestCase.java
bstansberry: @scottmarlow Thanks; leaving it the way I had it would confuse future maintainers, wondering why a simple thing was done in a complicated way.

I made CACHE_REGION_NAME a constant and got rid of the getCacheRegionName method; callers just read the constant.

scottmarlow (approved):

scottmarlow:
Thanks Brian!

PR #15239, opened 2022-02-17
Drop legacy -Dts.clust switch

rhusar:


rhusar:
@pferraro After 8 years I think it's time to drop this "compatibility" bits.

pferraro (approved):

bstansberry:
Thanks, @rhusar

PR #15231, opened 2022-02-15
[WFLY-16049] Update the OIDC tests so that they can run with Keycloak 17.0.0 and later

fjuma:
https://issues.redhat.com/browse/WFLY-16049

PR #15245, opened 2022-02-17
[26.x][WFLY-16049] Update the OIDC tests so that they can run with Keycloak 17.0.0 and later

fjuma:
https://issues.redhat.com/browse/WFLY-16049
Upstream: https://github.com/wildfly/wildfly/pull/15231


bstansberry:
Thanks @fjuma

PR #15249, opened 2022-02-21
WFLY-16060 README.md contains broken link to contributing doc, and incorrect java versioin

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-16060

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Building
 
 Prerequisites:
 
-* JDK 8 or newer - check `java -version`
+* JDK 11 or newer - check `java -version`
 * Maven 3.6.0 or newer - check `mvn -v`
 * On *nix systems, make sure that the maximum number of open files for the user running the build is at least 4096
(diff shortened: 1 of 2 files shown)


chengfang:
another line in readme needs change:
`2. Make sure Xmx in eclipse.ini is at least 1280M, and it's using Java 8`

chengfang:
Though it could be a separate issue, it may be easier just to also update `contributing.adoc` to reflect current java version requirement:

`Building WildFly requires Java 8 or newer. Make sure you have JAVA_HOME set to point to the JDK8 installation. Build uses Maven 3.`

`* OpenJDK 8 (eight!) You can use any distribution in example Eclipse`

parsharma:
@chengfang I have updated the files, In case there are more places where we need to update the java version, I think we should log another Jira Issue for it.

Thanks

chengfang (approved):

bstansberry:
Thanks, @parsharma

PR #15250, opened 2022-02-21
[WFLY-16067] Use ServiceBuilder.requires() instead of ServiceBuilder.addDependencies()

ropalka:
https://issues.redhat.com/browse/WFLY-16067


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MdbDeliveryDependenciesProcessor.java
```diff
@@ -78,15 +78,14 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                     if (mdbDescription.isClusteredSingleton()) {
```
chengfang, line 78: Any reason the above `addDependency` at line 76 (and another one at line 98-99) is not replaced?

ropalka: Yes, other addDependency() methods have different Java method signature @chengfang

chengfang (approved):

PR #15252, opened 2022-02-21
EE10 experiment

bstansberry:
just for CI of some stuff

there will be some JSF and JPA failures due to 2 known issues

https://hibernate.atlassian.net/browse/HHH-15084
https://issues.redhat.com/browse/WFLY-16061

I never tested the EL upgrade locally so I have no idea what impact that will bring.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/modules/export/AlphaExtension.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/modules/export/AlphaExtension.java
@@ -29,7 +29,7 @@ import javax.enterprise.inject.spi.Extension;
 public class AlphaExtension implements Extension {
 
     public void registerBravo(@Observes BeforeBeanDiscovery event, BeanManager manager) {
-        event.addAnnotatedType(manager.createAnnotatedType(BravoBean.class));
-        event.addAnnotatedType(manager.createAnnotatedType(CharlieBean.class));
+        event.addAnnotatedType(manager.createAnnotatedType(BravoBean.class), null);
+        event.addAnnotatedType(manager.createAnnotatedType(CharlieBean.class), null);
     }
 }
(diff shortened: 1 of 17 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/modules/export/AlphaExtension.java
```diff
@@ -29,7 +29,7 @@
 public class AlphaExtension implements Extension {
 
     public void registerBravo(@Observes BeforeBeanDiscovery event, BeanManager manager) {
-        event.addAnnotatedType(manager.createAnnotatedType(BravoBean.class));
-        event.addAnnotatedType(manager.createAnnotatedType(CharlieBean.class));
+        event.addAnnotatedType(manager.createAnnotatedType(BravoBean.class), null);
```
manovotn, line 32: Just a side note, I'd recommend avoiding the use of `null` as an identifier. It "just works" because Weld happens to deal with it but this isn't specified anywhere and might change.
You can do something as simple as `BravoBean.class.getName()`

bstansberry: 👍  Thanks.

manovotn:
BTW looking at the failures on this PR, it seems there are some dependencies using removed deprecated APIs from CDI.
Namely `BeanManager#createInjectionTarget(AnnotatedType)` and `BeforeBeanDiscovery#addAnnotatedType(AnnotatedType)`

The error from JSF seems to be only because of source transformation from older version, since new versions have this fixed ([at least according to GH repo](https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/flow/FlowDiscoveryCDIExtension.java)). I know WFLY used some jboss fork version so I am not sure how we handle this for EE 9 and 10.

The other error is coming from Hibernate and is something I probably overlooked back when I was sending a PR for similar thing in hibernte validator. I've sent a PR - https://github.com/hibernate/hibernate-orm/pull/4829

bstansberry:
@manovotn re the JSF problem that is tracked via https://issues.redhat.com/browse/WFLY-16061. @fjuma merged the fix into jboss/mojarra; just needs a release and an upgrade.

manovotn:
Yea, I've had the same PR pending there for half a year now https://github.com/jboss/mojarra/pull/90 :-D
Guess I can close it now.

bstansberry:
I'm closing this as @manovotn is carrying this forward via #15254

PR #15254, opened 2022-02-22
WFLY-16100, WFLY-16056 EE 10 changes for CDI; adaptation of failing tests

manovotn:
This is built on top of Brian's changes in https://github.com/wildfly/wildfly/pull/15252 as a PoC.

It adds some of missing CDI bits that are fairly easy to implement in WFP.
I created a PR out of it simply to get a CI run on all the tests and show the scope of needed changes. I was initially trying to do this locally but the runs take a considerable amount of time and CPU, so I decided to let it run here to be able to get back to it in an async manner :-)


manovotn:
Ok, there is about 250 tests that are broken by this change. I'll look into them in the coming days; most should be easily fixable by either adding a non-empty beans.xml or by annotating anything that's supposed to be a CDI bean.

manovotn:
The JDK 11 failure of `org.jboss.as.test.integration.weld.deployment.BeanDiscoveryTest` is related - I've changed that test to expect Weld 5 behavior in case of deployments with empty `beans.xml`.
We could `@Ignore` it with a comment for the time being, but I am pretty sure nobody will remember to flip that back on later.
Alternatively, I could switch it back to original behavior and exclude it from EE 9 but I am -1 for that as it will again show the same problem at a later stage.

Meanwhile, I am adding more TCK setup changes and getting green results locally. I'll push my changes shortly.

manovotn:
I've ignored the two test methods that are causing mayhem for JDK 11 testing.
I've also modified more MP TCK modules that spotted failures in and force-pushed it here. I was able to get a pass locally but I'll let the CI tell me what else I missed.

manovotn:
@bstansberry I think this PR should be ready for review. It is a lot of files, but the separate commits have individual changes which might be easier to review.

TL;DR is:
* Switched default empty beans.xml behavior to `annotated` discovery mode
* Introduced subsystem configuration allowing users to switch to previous behavior (mode `all`)
* Fixed WFLY tests that were crashing because of this change
  *  NOTE: I noticed a lot of exluded tests for EE 9 testing; these possibly still have this issue, but it should be trivial to fix once they are enabled again
* Lots of MP TCKs were failing so I've altered several POMs to make use of the new Weld config and run in legacy mode which makes them pass
  * I am planning to look into those failures and contribute them upstream, but this temporary fix should do the trick for WFLY
* There are 14 failing tests because of Hibernate and/or JSF impl using now removed CDI API methods; all of these tests are excluded in this PR and can be re-enabled once we get the dependencies updated

bstansberry:
@manovotn I sent you https://github.com/manovotn/wildfly/pull/1 re the transformer stuff. Other than that everything here looks good, so if that change looks good to you, this should be good, except I'd like a WFLY-xxxxx mention in any commit messages and in the PR title. That could be WFLY-15687 or a separate one for this "initial" work. A separate issue will help us better track incremental progress.

manovotn:
@bstansberry I have created https://issues.redhat.com/browse/WFLY-16100 as a subtask of the original issue.
All commits in this PR are now attributed to either WFLY-16100 or WFLY-16056 (test fixes) and the title has been adapted as well.

bstansberry (approved):
Excellent! Thanks @manovotn -- our first big step in WF itself toward EE 10!

PR #15270, opened 2022-03-01
26.x [WFLY-15832] Ability to create ConfigSource root directory in MP Config

kabir:
subsystem.

Enhance tests to illustrate that if a directory for a config-source or a
config-source-root does not exist, the config is still valid and will
not give errors.

https://issues.redhat.com/browse/WFLY-15832


bstansberry:
Thanks @kabir

PR #15271, opened 2022-03-02
[WFLY-15680] Upgrade Jakarta Activation to 2.1 in WildFly Preview

yersan:
- Upgrade Jakarta Activation
- Integrate Eclipse Angus Activation

Jira issue: https://issues.redhat.com/browse/WFLY-15680

### DIFF
--- a/ee-9/common/pom.xml
+++ b/ee-9/common/pom.xml
@@ -218,6 +218,17 @@
         </dependency>
 
         <dependency>
+            <groupId>org.eclipse.angus</groupId>
+            <artifactId>angus-activation</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.el</artifactId>
             <exclusions>
--- a/ee-9/common/src/main/resources/license/preview-feature-pack-common-licenses.xml
+++ b/ee-9/common/src/main/resources/license/preview-feature-pack-common-licenses.xml
@@ -46,6 +46,17 @@
       </licenses>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.angus</groupId>
+      <artifactId>angus-activation</artifactId>
+      <licenses>
+        <license>
+          <name>Eclipse Distribution License, Version 1.0</name>
+          <url>https://repository.jboss.org/licenses/edl-1.0.txt</url>
+          <distribution>repo</distribution>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
       <groupId>org.jboss</groupId>
       <artifactId>jboss-ejb-client-jakarta</artifactId>
       <licenses>
--- a/ee-9/common/src/main/resources/modules/system/layers/base/org/eclipse/angus/activation/main/module.xml
+++ b/ee-9/common/src/main/resources/modules/system/layers/base/org/eclipse/angus/activation/main/module.xml
@@ -2,7 +2,7 @@
 
 <!--
   ~ JBoss, Home of Professional Open Source.
-  ~ Copyright 2020, Red Hat, Inc., and individual contributors
+  ~ Copyright 2022, Red Hat, Inc., and individual contributors
   ~ as indicated by the @author tags. See the copyright.txt file in the
   ~ distribution for a full listing of individual contributors.
   ~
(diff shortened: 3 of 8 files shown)


ee-9/ee-9-api/src/main/resources/modules/system/layers/base/jakarta/activation/api/main/module.xml
```diff
@@ -25,7 +25,8 @@
 <module xmlns="urn:jboss:module:1.9" name="jakarta.activation.api">
 
     <resources>
-        <artifact name="${com.sun.activation:jakarta.activation}"/>
+        <artifact name="${jakarta.activation:jakarta.activation-api}"/>
+        <artifact name="${org.eclipse.angus:angus-activation}"/>
```
bstansberry, line 29: Is putting API and impl in the same module the right thing to do?

I know in the past the answer was 'yes', but I'm wondering whether the fact they've now been split at Eclipse justifies a rethink.

yersan: Probably not, because it leaves us with less flexibility.

@bstansberry I thought it was necessary to continue exporting to the callers the implementation which was already available under the `jakarta.activation.api` module. And the alternative to not add two artifacts to this module was to create a new module for the implementation, make it a dependency of `jakarta.activation.api` and export it, so the callers who depend on `jakarta.activation.api` don't get problems and can continue using the implementation classes from the caller classloader.

But that's conceptually wrong, the callers should use the API exposed by `jakarta.activation.api` and there is no need to export or reuse the implementation in the caller. The responsibility to handle the implementation classloader can be kept in the `jakarta.activation.api` module if we make the implementation a dependency of this module. There is no need to export the implementation. And I believe that callers should not be using classes from the implementation that are outside of the API contract.

So, yes, probably, this should be split into two modules.

bstansberry: @yersan OK, lets try that. It would be good to do the split soon so if it causes problems we have time to notice and react. If we release an EE 10 server without a split we are somewhat constrained from doing the split later.

ee-9/ee-9-api/src/main/resources/modules/system/layers/base/jakarta/activation/api/main/module.xml
```diff
@@ -25,17 +25,12 @@
...
 
     <dependencies>
         <module name="java.logging" />
-        <module name="javax.api" />
-        <module name="javax.mail.api" optional="true">
-            <imports><include path="META-INF"/></imports>
-        </module>
-        <module name="com.sun.xml.messaging.saaj" optional="true"/>
-        <module name="org.jboss.ws.native.jbossws-native-core" optional="true"/>
-        <module name="org.apache.cxf" optional="true"/>
```
bstansberry, line 39: @jimma Please review these last 3 lines.  The org.jboss.ws.native.jbossws-native-core module doesn't exist, so removing that one for sure won't break anything. :)

bstansberry: @yersan The WF Preview ts failures make me suspect this needs some other optional deps so it can find implementations.

yersan: Right @bstansberry , I've added the implementation to the modules that export it to CXF, let's see how it goes.

About the removals I did on this module, I still need to understand the original intention of having them there, but I believe they are unnecessary at least for the API, because the Jarkarta Activation API 2.1 doesn't require them.
It seems they could be providers for the Jakarta Activation SPI, I hope @jimma can shed some light there.

About the Jakarta Mail one, I haven't found the reason why it was there yet; this PR is still in progress. Any lights are welcome :)

Let's see what CI says.

jimma: @yersan @bstansberry  The "com.sun.xml.messaging.saaj" and "jbossws-native-core" dependencies are added for jaxrpc. We should remove them all. Please see [AS7-2550](https://issues.redhat.com/browse/AS7-2550).
For "org.apache.cxf" dependency, I found this is added for [AS7-2836](https://issues.redhat.com/browse/AS7-2836) and it's quite old change. But I think "org.apache.cxf" can be safely removed too. The activiation api should not depend any thing from CXF.

bstansberry (approved):

PR #15274, opened 2022-03-03
WFLY-16111 Increase timeout and count of attempts for starting container in integration tests for OIDC

kstekovi:
new property for count of startup attempts and timeout for one attempt with default vales and possibility change them if needed.

https://issues.redhat.com/browse/WFLY-16111


kstekovi:
Hi @fjuma could you please check this pull request? We are facing random fails due to container doesn't start in default time limit.

fjuma (approved):

PR #15276, opened 2022-03-03
WFLY-16112 Batch JobOperatorService should look for only active job n…

chengfang:
…ames to stop during suspend

https://issues.redhat.com/browse/WFLY-16112

This fix is needed to address a customer issue (https://issues.redhat.com/browse/JBEAP-23244).


ivassile:
@bstansberry We have to include this fix in EAP 7.4.4 [1] and retag. Can you please merge it? Thanks!

[1] https://issues.redhat.com/browse/JBEAP-23323

PR #15288, opened 2022-03-06
[WFLY-16120] Upgrade Agroal to 1.15

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16120


bstansberry:
Thanks @boris-unckel

PR #15294, opened 2022-03-07
[26.x] [WFLY-15964] OpenTelemetry ratio-based sampler should not accept value bigger than 1.0

jasondlee:
https://issues.redhat.com/browse/WFLY-15964

Port fix from main


bstansberry:
Thanks @jasondlee

PR #15295, opened 2022-03-07
[26.x] [WFLY-15813] Update io.grpc version

jasondlee:
https://issues.redhat.com/browse/WFLY-15813

Port update from main


bstansberry:
Thanks, @jasondlee

PR #15306, opened 2022-03-09
[WFLY-15816]: Remove snakeyaml module from WildFly.

ehsavoie:
* removing snakeyaml from WildFly since it is defined in WildFly Core

Jira: https://issues.redhat.com/browse/WFLY-15816


bstansberry:
Thanks, @ehsavoie

PR #15308, opened 2022-03-10
 WFLY-15114 Refactor MP Health integration tests to avoid duplications

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-15114

### DIFF
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationReadyTestBase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationReadyTestBase.java
@@ -99,5 +99,4 @@ public abstract class MicroProfileHealthApplicationReadyTestBase {
 
         checkGlobalOutcome(managementClient, "check-ready", false, null);
     }
-
 }
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationStartupTestBase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationStartupTestBase.java
@@ -55,7 +55,7 @@ public abstract class MicroProfileHealthApplicationStartupTestBase {
     @Deployment(name = "MicroProfileHealthApplicationStartupTestBaseSetup")
     public static Archive<?> deploySetup() {
         WebArchive war = ShrinkWrap.create(WebArchive.class, "MicroProfileHealthApplicationStartupTestBaseSetup.war")
-            .addClass(MicroProfileHealthApplicationStartupSetupTask.class);
+                .addClass(MicroProfileHealthApplicationStartupSetupTask.class);
         return war;
     }
 
@@ -86,7 +86,7 @@ public abstract class MicroProfileHealthApplicationStartupTestBase {
     @InSequence(2)
     @OperateOnDeployment("MicroProfileHealthApplicationStartupTestBase")
     public void testApplicationStartupAfterDeployment(@ArquillianResource URL url) throws Exception {
-            checkGlobalOutcome(managementClient, "check-started", true, "successfulStartupProbe");
+        checkGlobalOutcome(managementClient, "check-started", true, "successfulStartupProbe");
     }
 
     @Test
@@ -97,5 +97,4 @@ public abstract class MicroProfileHealthApplicationStartupTestBase {
 
         checkGlobalOutcome(managementClient, "check-started", false, null);
     }
-
 }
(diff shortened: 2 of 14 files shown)


bstansberry:
@xstefank Please review

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationReadyTestBase.java
```diff
@@ -100,4 +112,57 @@ public void testHealthCheckAfterUndeployment() throws Exception {
...
+                            return;
+                        }
+                    }
+                    fail("Probe named " + probeName + " not found in " + content);
+                }
+            }
+        }
+    }
+
+    static void testHttpOperation(ModelNode response, boolean mustBeUP, String probeName){
```
xstefank, line 147: This is a management operation, not HTTP

xstefank (changes requested):
@parsharma there are also `*Startup*` tests in the same package that can also be refactored.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationReadyTestBase.java
parsharma: Changes done. Thanks for letting me know.

xstefank (changes requested):
Do you need to repeat testHttp and testManagement method in every base class? Would it be better to move this to some utility class or common parent of base classes so it doesn't need to be duplicated?

parsharma:
@xstefank yes, correct. I have removed all repeated methods please take a look. Thanks for the review comment.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationStartupTestBase.java
```diff
@@ -42,6 +42,7 @@
 import java.io.IOException;
 import java.net.URL;
 
+
 /**
```
xstefank, line 46: these whitespace changes are unnecessary but I'll leave this to the merger to decide if this needs to be fixed

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationWithoutStartupOperationTestCase.java
```diff
@@ -26,10 +26,9 @@
 import org.jboss.dmr.ModelNode;
 
 import java.io.IOException;
+import static org.wildfly.test.integration.microprofile.health.MicroProfileHealthApplicationReadyTestBase.testManagementOperation;
```
xstefank, line 29: personally, I think it would be better to move it into a separate class because referencing static method from *Ready* class in *Startup* or *Liveness* might be confusing.

xstefank (changes requested):

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthApplicationWithoutStartupOperationTestCase.java
parsharma: Yeah, I have created one class Utility.java. Please suggest if you have any other names.

xstefank: Maybe better would be `MicroProfileHealthUtils`

parsharma: Okay, done.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthUtils.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018, Red Hat, Inc., and individual contributors
```
xstefank, line 3: this should be 2022

xstefank (approved):
Approving, just the year in the license is wrong.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthUtils.java
parsharma: Done

bstansberry:
Thanks, @parsharma

PR #15314, opened 2022-03-11
[26.x] WFLY-15569 Database persistent auto timer are created twice

chengfang:
https://issues.redhat.com/browse/WFLY-15569

Port the fix from main branch to 26.x branch.


bstansberry:
Thanks, @chengfang

PR #15317, opened 2022-03-11
WFLY-16142 Upgrade WildFly Common from 1.5.2.Final to 1.6.0.Final

chengfang:
https://issues.redhat.com/browse/WFLY-16142


bstansberry (approved):
Note to self -- the reason this property exists is for establishing the version used by wildfly-galleon-maven-plugin, where the version from wildfly-core dependency management isn't picked up.

PR #15328, opened 2022-03-16
[WFLY-16158]: Deployment submodel should be runtime

ehsavoie:
* setting runtime on deployment submodels for Messaging, Webservices,
  Resource adapters, datasources, jpa jaxrs, jpa, ejb3, undertow

Jira: https://issues.redhat.com/browse/WFLY-16158


ehsavoie:
Well, some of the deployment submodel were registered as read-only :)
A workaround which I consider hacky ;) is to use the isFeature instead.

yersan:
@bstansberry / @ehsavoie Was there a final conclusion about this?

bstansberry:
@ehsavoie Re the point #3 concern from https://issues.redhat.com/browse/WFCORE-3432 I don't think we need to worry about it at this point. We don't guarantee bug-for-bug compatibility and we're making major changes now so we're not doing releases right now where we avoid such changes.

A test is failing with this but once that's sorted I'll merge it if you ping me when CI passes.

ehsavoie:
@bstansberry ping :)

bstansberry:
Thanks @ehsavoie

PR #15341, opened 2022-03-18
[WFLY-16164] Removing JDR dependency on Xalan and switching to JDK JAXP

ropalka:
https://issues.redhat.com/browse/WFLY-16164


bstansberry:
Thanks, @ropalka

PR #15301, opened 2022-03-08
[WFLY-16127] Upgrade jboss-iiop-client to 1.0.2.Final

tadamski:
https://issues.redhat.com/browse/WFLY-16127

PR because of Jakarta migration

PR #14880, opened 2021-11-04
WFLY-15028 Move WildFly Preview to a native jakarta namespace variant…

chengfang:
… of the JBoss EJB Client jar

[WFLY-15028](https://issues.redhat.com/browse/WFLY-15028) Move WildFly Preview to a native jakarta namespace variant of the JBoss EJB Client jar

[WFLY-15626](https://issues.redhat.com/browse/WFLY-15626) Upgrade jboss-ejb-client from 4.0.43.Final to 4.0.44.Final

PR #15344, opened 2022-03-18
[WFLY-16127] iiop-client: Jakarta module configuration

tadamski:
This PR is a follow on https://github.com/wildfly/wildfly/pull/15301 and contains iiop-client module configuration fix.

### DIFF
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/iiop-client/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/iiop-client/main/module.xml
@@ -27,7 +27,7 @@
     </properties>
 
     <resources>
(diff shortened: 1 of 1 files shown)


chengfang:
There may be other files that need to be updated, if haven't been updated already, like adding license to the feature-pack license file, excluding jboss-iiop-client in ee-9/feature-pack build pom.  Similar to https://github.com/wildfly/wildfly/pull/14880/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8 , which was a while ago and something may have changed since then, but overall should still be relevant.

tadamski:
@chengfang thanks
I have created https://github.com/wildfly/wildfly/pull/15349 to deal with those config issues

PR #15373, opened 2022-03-24
[WFLY-16198] Restriction of XML External Entity Reference (XXE)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16198


boris-unckel:
@bstansberry Tests passed, could you review and approve, please? I would provide a 26.x backport if you agree, what do you think?

ropalka (approved):

PR #15374, opened 2022-03-24
WFLY-16123 Replace CompositeIterable/Iterator with impl in wildfly-common

pferraro:
https://issues.redhat.com/browse/WFLY-16123


rhusar (approved):
👍 LGTM

PR #15386, opened 2022-03-31
[WFLY-16225] Add .gitleaks.toml file with allowlist

skyllarr:
https://issues.redhat.com/browse/WFLY-16225


bstansberry:
Thanks @Skyllarr

PR #15397, opened 2022-04-04
WFLY-16207 Application caches throw marshalling exceptions when using deployment classes for keys or values

pferraro:
https://issues.redhat.com/browse/WFLY-16207


testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/infinispan/remote/RemoteInfinispanCacheTestCase.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Deployment(name = DEPLOYMENT_2, managed = false, testable = false)
+    @TargetsContainer(NODE_2)
+    public static Archive<?> deployment2() {
+        return createDeployment();
+    }
+
+    private static Archive<?> createDeployment() {
+        return createDeployment(CreatedCacheBean.class);
```
rhusar, line 52: s/CreatedCacheBean/RemoteCacheBean

rhusar (changes requested):

rhusar:
@pferraro The latest failure is:

Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("javax.management.MBeanPermission" "org.infinispan.client.hotrod.impl.ClientStatistics#-[org.wildfly.clustering.infinispan:cache=org.jboss.as.test.clustering.cluster.infinispan.bean.remote.RemoteCacheBean,name=remote,type=HotRodClient]" "registerMBean")" in code source "(vfs:/content/RemoteCacheBean.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.RemoteCacheBean.war" from Service Module Loader")

pferraro:
Adding hold label until we work out how this will play with Infinispan query, etc.

clustering/infinispan/embedded/spi/src/main/java/org/wildfly/clustering/infinispan/marshall/DefaultEncoderRegistry.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    private Transcoder findTranscoder(MediaType fromType, MediaType toType) {
+        return this.transcoderCache.computeIfAbsent(fromType, mt -> new ConcurrentHashMap<>(4)).computeIfAbsent(toType, mt -> this.transcoders.stream().filter(t -> t.supportsConversion(fromType, toType)).findFirst().orElse(null));
+    }
+
+    @Deprecated
+    @Override
+    public org.infinispan.commons.dataconversion.Encoder getEncoder(Class<? extends org.infinispan.commons.dataconversion.Encoder> encoderClass, short encoderId) {
```
rhusar, line 105: Looks like a remnant that can just be an import here and other places (org.infinispan.commons.dataconversion.Encoder) unless it's intentional...

rhusar (approved):
LGTM now.

clustering/infinispan/embedded/spi/src/main/java/org/wildfly/clustering/infinispan/marshall/DefaultEncoderRegistry.java
pferraro: This was intentional.  The Encoder interface and the methods of EncoderRegistry that use it are deprecated - so the FQCN was used to reduce the likelihood of merge conflicts when these methods are removed.

PR #15401, opened 2022-04-04
WFLY-12459 - Track processing time of in flight requests in Undertow

jasondlee:
https://issues.redhat.com/browse/WFLY-12459

Add support to WFLY for UNDERTOW-1593

Requires https://github.com/undertow-io/undertow/pull/1294

### DIFF
--- a/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/active_request/ActiveRequestTrackerHandlerTest.java
+++ b/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/active_request/ActiveRequestTrackerHandlerTest.java
@@ -1,0 +1,138 @@
+package org.jboss.as.test.integration.web.handlers.active_request;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.IOException;
+import java.net.URL;
+
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.arquillian.api.ServerSetupTask;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.test.shared.ServerReload;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup(ActiveRequestTrackerHandlerTest.ActiveRequestTrackingSetupTask.class)
+public class ActiveRequestTrackerHandlerTest {
+
+    private static final String WEB_XML
+            = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+            + "<web-app xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns=\"http://java.sun.com/xml/ns/javaee\"\n"
+            + "         xsi:schemaLocation=\"http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd\"\n"
+            + "         metadata-complete=\"false\" version=\"3.0\">\n"
+            + "    <servlet-mapping>\n"
+            + "        <servlet-name>javax.ws.rs.core.Application</servlet-name>\n"
+            + "        <url-pattern>/*</url-pattern>\n"
+            + "    </servlet-mapping>"
+            + "</web-app>";
+
+    @Deployment
+    public static WebArchive deploy() {
+        return ShrinkWrap.create(WebArchive.class, "active-request-tracking.war")
+                .addClasses(DelayServlet.class);
+    }
+
+    @ArquillianResource
+    protected URL url;
+
+    @Test
+    public void testActiveRequest() throws Exception {
+        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+            ManagementClient managementClient = new ManagementClient(TestSuiteEnvironment.getModelControllerClient(),
+                    TestSuiteEnvironment.getServerAddress(), TestSuiteEnvironment.getServerPort(), "remote+http");
+
+
+            new Thread(() -> makeRequest(httpClient)).start();
+            // This test is somewhat time-sensitive in that an in-flight request is needed to verify that the handler
+            // is correctly tracking and returning active requests. To that end, the test submits a request in a
+            // Thread (above) to let the request be made in a background thread while the main thread attempts to
+            // get the active request information. There is a short sleep here to force/encourage the JVM to allow the
+            // background thread to start to initiate the request. That thread has a short delay in it as well (see
+            // servlet) to give this thread time to retrieve the data.
+            Thread.sleep(500);
+
+            ModelNode response = managementClient.getControllerClient().execute(
+                    Operations.createOperation("list-active-requests",
+                            Operations.createAddress(
+                                    "subsystem", "undertow",
+                                    "server", "default-server",
+                                    "host", "default-host")
+            ));
+
+            assertEquals(new ModelNode("success"), response.get("outcome"));
+            ModelNode result = response.get("result");
+            assertEquals(ModelType.LIST, result.getType());
+            assertEquals(1, result.asList().size());
+        }
+    }
+
+    private void makeRequest(CloseableHttpClient httpClient) {
+        try {
+            httpClient.execute(new HttpGet(url.toExternalForm() + "test"));
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static class ActiveRequestTrackingSetupTask implements ServerSetupTask {
+        public static final ModelNode subsystemAddress = Operations.createAddress(
+                "subsystem", "undertow");
+        public static final ModelNode hostAddress = Operations.createAddress(
+                "subsystem", "undertow",
+                "server", "default-server",
+                "host", "default-host");
+
+        @Override
+        public void setup(final ManagementClient managementClient, final String containerId) throws Exception {
+            // /subsystem=undertow/server=default-server/host=default-host:write-attribute(name="active-request-tracking-enabled", value="true")
+            execute(managementClient,
+                    Operations.createWriteAttributeOperation(subsystemAddress, "statistics-enabled", true),
+                    true);
+            execute(managementClient,
+                    Operations.createWriteAttributeOperation(hostAddress, "active-request-tracking-enabled", true),
+                    true);
+            ServerReload.reloadIfRequired(managementClient);
+        }
+
+        @Override
+        public void tearDown(final ManagementClient managementClient, final String containerId) throws Exception {
+            execute(managementClient,
+                    Operations.createWriteAttributeOperation(subsystemAddress, "statistics-enabled", false),
+                    true);
+            execute(managementClient,
+                    Operations.createWriteAttributeOperation(hostAddress, "active-request-tracking-enabled", false),
+                    true);
+            ServerReload.reloadIfRequired(managementClient);
+        }
+
+        private ModelNode execute(final ManagementClient managementClient,
+                                  final ModelNode op,
+                                  final boolean expectSuccess) throws IOException {
+            ModelNode response = managementClient.getControllerClient().execute(op);
+            final String outcome = response.get("outcome").asString();
+            if (expectSuccess) {
+                assertEquals(response.toString(), "success", outcome);
+                return response.get("result");
+            } else {
+                assertEquals("failed", outcome);
+                return response.get("failure-description");
+            }
+        }
+    }
+
+}
--- a/undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
@@ -1,0 +1,47 @@
+package org.wildfly.extension.undertow;
+
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import io.undertow.predicate.Predicate;
+import io.undertow.server.handlers.ActiveRequestTrackerHandler;
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+
+public class ActiveRequestTrackerService implements Service<ActiveRequestTrackerHandler> {
+    private final Consumer<ActiveRequestTrackerService> serviceConsumer;
+    private final Supplier<Host> host;
+    private final Predicate predicate;
+    private ActiveRequestTrackerHandler handler;
+
+    public ActiveRequestTrackerService(Consumer<ActiveRequestTrackerService> serviceConsumer,
+                                       Supplier<Host> host,
+                                       final Predicate predicate) {
+        this.serviceConsumer = serviceConsumer;
+        this.host = host;
+        this.predicate = predicate;
+    }
+
+    @Override
+    public void start(StartContext startContext) throws StartException {
+        final Host host = this.host.get();
+        host.setActiveRequestTrackerHandler(httpHandler -> {
+            handler = new ActiveRequestTrackerHandler(httpHandler, predicate);
+            return handler;
+        });
+
+        serviceConsumer.accept(this);
+    }
+
+    @Override
+    public void stop(StopContext stopContext) {
+
+    }
+
+    @Override
+    public ActiveRequestTrackerHandler getValue() throws IllegalStateException, IllegalArgumentException {
+        return handler;
+    }
+}
--- a/undertow/src/main/java/org/wildfly/extension/undertow/Host.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/Host.java
@@ -42,6 +42,7 @@ import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 
 import io.undertow.Handlers;
@@ -94,6 +95,7 @@ public class Host implements Service<Host>, FilterLocation {
     private final int defaultResponseCode;
     private volatile HttpHandler rootHandler = null;
     private volatile AccessLogService  accessLogService;
+    private volatile UnaryOperator<HttpHandler> activeRequestTrackerHandler;
     private volatile GateHandlerWrapper gateHandlerWrapper;
     private volatile Function<HttpHandler, HttpHandler> accessLogHttpHandler;
 
@@ -188,6 +190,7 @@ public class Host implements Service<Host>, FilterLocation {
         AccessLogService logService = accessLogService;
         HttpHandler rootHandler = pathHandler;
         final Function<HttpHandler, HttpHandler> accessLogHttpHandler = this.accessLogHttpHandler;
+        final UnaryOperator<HttpHandler> activeRequestTrackerHandler = this.activeRequestTrackerHandler;
 
         ArrayList<UndertowFilter> filters = new ArrayList<>(this.filters);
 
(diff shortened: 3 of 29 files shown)


undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd
```diff
@@ -316,6 +316,7 @@
         </xs:attribute>
         <xs:attribute name="disable-console-redirect" use="optional" type="xs:boolean" default="false"/>
         <xs:attribute name="queue-requests-on-start" type="xs:boolean" default="true"/>
+        <xs:attribute name="active-request-tracking-enabled" type="xs:boolean" default="false"/>
```
pferraro, line 319: We don't generally modify legacy (i.e. already released) schemas, unless there was a bug in the originally release version.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
```diff
@@ -0,0 +1,112 @@
...
+            .setRestartAllServices()
+            .build();
+    private static final SimpleOperationDefinition LIST_ACTIVE_REQUESTS =
+            new SimpleOperationDefinitionBuilder("list-active-requests",
+                    UndertowExtension.getResolver("active-request-tracker"))
+            .setReplyType(ModelType.LONG)
+            .setRuntimeOnly()
+            .build();
+
+    static final Collection<SimpleAttributeDefinition> ATTRIBUTES = Arrays.asList(PREDICATE);
```
pferraro, line 47: Is there a specific reason to capture the AttributeDefinition subtype?  If you instead use a Collection<AttributeDefinition>, you can avoid the awkward casting below.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
```diff
@@ -0,0 +1,112 @@
...
+    static final ActiveRequestTrackerDefinition INSTANCE = new ActiveRequestTrackerDefinition();
```
pferraro, line 48: Is this static final necessary? A ResourceDefinition is a temporary object, and such a reference prevents it from being garbage collected after resource registration completes.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.undertow.predicate.Predicate;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.handlers.ActiveRequestTrackerHandler;
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+
+public class ActiveRequestTrackerService implements Service<ActiveRequestTrackerService> {
```
pferraro, line 15: Exposing the MSC Service API via the service value is an anti-pattern, i.e. a consumer should not have access to the start(...)/stop(...) methods.
Since this service seems to provide a ActiveRequestTrackerHandler, perhaps you mean to use Service<ActiveRequestTrackerHandler> instead?

undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd
jasondlee: Right. Sorry. That was noise left from an experiment. I'll clean it up.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
jasondlee: I believe the answer to both of this question and the one above is that I was following an example in existing code. :)  I can certainly clean these up, though, if that's not the recommended practice.

undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd
jasondlee: @pferraro Do we need wildfly-undertow_14_0.xsd?

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
jasondlee: TBH, I did here what I saw in AccessLogService: undertow/src/main/java/org/wildfly/extension/undertow/AccessLogService.java

undertow/src/main/resources/schema/wildfly-undertow_10_0.xsd
pferraro: 13.0 is new as of WF27, so we don't need a new schema.  The same commit also incremented the model version, so there is no need to bump that either.  Interesting, this commit: https://github.com/wildfly/wildfly/commit/a1f9ad78f6c960e824a2ca007a40593706c2d060
... appears to have incremented the model version from 11.0.0 -> 13.0.0, so it seems there was never a 12.0.0 model.  That commit should never have been merged.
See: https://issues.redhat.com/browse/WFLY-15850

pferraro: FYI, I've filed https://issues.redhat.com/browse/WFLY-15850 to fix this before WF27 is released.
I don't mean that you need to address this here, but just be aware that whatever transformers you are introducing as part of this commit will need to be adjusted accordingly once WFLY-15850 is fixed.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
```diff
@@ -0,0 +1,112 @@
...
+public class ActiveRequestTrackerDefinition extends PersistentResourceDefinition {
+    static final RuntimeCapability<Void> ACTIVE_REQUEST_TRACKING_CAPABILITY =
+            RuntimeCapability.Builder.of(Capabilities.CAPABILITY_ACTIVE_REQUEST_TRACKING, true,
+                            ActiveRequestTrackerService.class)
+                    .setDynamicNameMapper(DynamicNameMappers.GRAND_PARENT)
+                    .build();
+    protected static final SimpleAttributeDefinition PREDICATE = new SimpleAttributeDefinitionBuilder(Constants.PREDICATE, ModelType.STRING, true)
+            .setAllowExpression(true)
+            .setValidator(PredicateValidator.INSTANCE)
+            .setRestartAllServices()
```
pferraro, line 38: I doubt that this attribute would require restarting *all* services. It should be sufficient to use Flag.RESTART_RESOURCE_SERVICES.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
```diff
@@ -0,0 +1,112 @@
...
+            .build();
+    private static final SimpleOperationDefinition LIST_ACTIVE_REQUESTS =
+            new SimpleOperationDefinitionBuilder("list-active-requests",
+                    UndertowExtension.getResolver("active-request-tracker"))
+            .setReplyType(ModelType.LONG)
```
pferraro, line 43: The reply type does not appear to be consistent with the actual operation.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerRemove.java
```diff
@@ -0,0 +1,20 @@
+package org.wildfly.extension.undertow;
+
+import org.jboss.as.controller.AbstractRemoveStepHandler;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.dmr.ModelNode;
+import org.jboss.msc.service.ServiceName;
+
+public class ActiveRequestTrackerRemove extends AbstractRemoveStepHandler {
```
pferraro, line 9: This operation handler is missing rollback behavior.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerAdd.java
```diff
@@ -0,0 +1,46 @@
...
+import io.undertow.predicate.Predicate;
+import io.undertow.predicate.Predicates;
+import org.jboss.as.controller.AbstractAddStepHandler;
+import org.jboss.as.controller.CapabilityServiceBuilder;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.dmr.ModelNode;
+
+public class ActiveRequestTrackerAdd extends AbstractAddStepHandler {
```
pferraro, line 15: This operation handle is missing rollback behavior.

undertow/src/main/java/org/wildfly/extension/undertow/Host.java
```diff
@@ -94,6 +94,7 @@
     private final int defaultResponseCode;
     private volatile HttpHandler rootHandler = null;
     private volatile AccessLogService  accessLogService;
+    private volatile Function<HttpHandler, HttpHandler> activeRequestTrackerHandler;
```
pferraro, line 97: UnaryOperator<HttpHandler> ?

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
jasondlee: Fixed locally.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
jasondlee: Fixed locally.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
jasondlee: Fixed locally.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
jasondlee: Fixed locally.

undertow/src/main/java/org/wildfly/extension/undertow/HostAdd.java
```diff
@@ -60,7 +63,9 @@
     static final HostAdd INSTANCE = new HostAdd();
 
     private HostAdd() {
-        super(HostDefinition.ALIAS, HostDefinition.DEFAULT_WEB_MODULE, HostDefinition.DEFAULT_RESPONSE_CODE, HostDefinition.DISABLE_CONSOLE_REDIRECT, HostDefinition.QUEUE_REQUESTS_ON_START);
+        super(HostDefinition.ALIAS, HostDefinition.DEFAULT_WEB_MODULE, HostDefinition.DEFAULT_RESPONSE_CODE,
```
pferraro, line 66: `super(HostDefinition.ATTRIBUTES);`

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerRemove.java
jasondlee: I'm not sure what you're asking for here. Point to an example?

undertow/src/main/java/org/wildfly/extension/undertow/Host.java
jasondlee: Can you clarify?

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
pferraro: Perhaps it only needs to restart the services of the parent resource?

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
```diff
@@ -0,0 +1,112 @@
...
+import org.jboss.as.controller.capability.DynamicNameMappers;
+import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
+import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceRegistry;
+
+public class ActiveRequestTrackerDefinition extends PersistentResourceDefinition {
```
pferraro, line 29: I am generally skeptical of single attribute child resources.  Is this child resource really warranted?  or might it be better represented as an attribute of its parent resource?

undertow/src/main/java/org/wildfly/extension/undertow/HostAdd.java
jasondlee: I just added the new attribute to the existing list, but I can modify HostDefinition.ATTRIBUTES and use that.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerRemove.java
pferraro: If this operation is part of a batch, say operation 2 of 5, and a failure occurs in operation 3, 4 or 5, then the AbstractAddStepHandler.rollbackRuntime(...) method will be invoked for the purpose of undoing the work that was done in performRuntime(...).  Similarly, the AbstractRemoveStepHandler invokes its recoverServices(...) method for the purpose of undoing the work that was done in its performRemove(...) method.

Check out the generic AddOperationHandler and RemoveOperationHandler classes used by the clustering subsystems for an example:
https://github.com/wildfly/wildfly/blob/main/clustering/common/src/main/java/org/jboss/as/clustering/controller/AddStepHandler.java
https://github.com/wildfly/wildfly/blob/main/clustering/common/src/main/java/org/jboss/as/clustering/controller/RemoveStepHandler.java

N.B. The clustering subsystems encapsulate resource runtime behavior within a separate ResourceServiceHandler abstraction, since, generally speaking, the rollback logic of the add operation is the same as the runtime logic of the remove operation, and visa versa.

undertow/src/main/java/org/wildfly/extension/undertow/Host.java
pferraro: Generally, we would use java.util.function.UnaryOperator to represent decorator functions.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
pferraro: See https://issues.redhat.com/browse/WFLY-8512

undertow/src/main/java/org/wildfly/extension/undertow/HostAdd.java
```diff
@@ -106,16 +140,19 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
...
         }
         final Supplier<Server> sSupplier = csb.requiresCapability(Capabilities.CAPABILITY_SERVER, Server.class, serverName);
         final Supplier<UndertowService> usSupplier = csb.requiresCapability(Capabilities.CAPABILITY_UNDERTOW, UndertowService.class);
         final Supplier<ControlledProcessStateService> cpssSupplier = csb.requires(ControlledProcessStateService.SERVICE_NAME);
-        final Supplier<SuspendController> scSupplier = csb.requires(context.getCapabilityServiceName(Capabilities.REF_SUSPEND_CONTROLLER, SuspendController.class));
-        csb.setInstance(new Host(hostConsumer, sSupplier, usSupplier, cpssSupplier, scSupplier, name, aliases == null ? new LinkedList<>(): aliases, defaultWebModule, defaultResponseCode, queueRequestsOnStart));
+        final Supplier<SuspendController> scSupplier =
+                csb.requires(context.getCapabilityServiceName(Capabilities.REF_SUSPEND_CONTROLLER, SuspendController.class));
+        csb.setInstance(new Host(hostConsumer, sSupplier, usSupplier, cpssSupplier, scSupplier, name,
+                aliases == null ? new LinkedList<>(): aliases, defaultWebModule, defaultResponseCode, queueRequestsOnStart));
```
pferraro, line 155: nit: Generally, we don't reformat code when making functional changes.

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/active_request/ActiveRequestTrackerHandlerTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private void makeRequest(CloseableHttpClient httpClient) {
+        try {
+            httpClient.execute(new HttpGet(url.toExternalForm() + "test"));
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static class ActiveRequestTrackingSetupTask implements ServerSetupTask {
```
pferraro, line 93: Pro tip: leverage CLIServerSetupTask for this.
e.g. https://github.com/wildfly/wildfly/blob/main/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/remote/InfinispanServerSetupTask.java

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerDefinition.java
pferraro: After looking at this change in more detail, it seems that this feature has 2 bits of configuration, a boolean flag on the host, and a single property child resource.
So, a user would either:
* Keep the flag disabled, and not specify an active-request-tracker resource
* Enable the flag, and specify an active-request-tracker resource

That would suggest that the boolean flag is superfluous, and the feature can enabled or disabled depending on whether the child resource is defined or not.

Secondly, is there any possibility that this resource would ever have additional attributes?  If not, it seems to me that this is better represented by a single "active-request-tracker" (or similar) attribute of type ModelType.String defined within the Host resource.  WDYT?

undertow/src/main/java/org/wildfly/extension/undertow/UndertowService.java
```diff
@@ -129,6 +129,10 @@ public static ServiceName accessLogServiceName(final String server, final String
         return virtualHostName(server, virtualHost).append(Constants.ACCESS_LOG);
     }
 
+    public static ServiceName activeRequestTrackingServiceName(final String server, final String virtualHost) {
+        return virtualHostName(server, virtualHost).append(Constants.ACTIVE_REQUEST_TRACKER);
+    }
+
```
pferraro, line 135: Since you have a capability defined for this service, we should prefer that for generating the service name.

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
```diff
@@ -89,12 +89,20 @@
             .setDefaultValue(ModelNode.TRUE)
             .setAllowExpression(true)
             .build();
+    static final SimpleAttributeDefinition ACTIVE_REQUEST_TRACKING_ENABLED = new SimpleAttributeDefinitionBuilder(Constants.ACTIVE_REQUEST_TRACKING_ENABLED, ModelType.BOOLEAN, true)
+            .setRestartAllServices()
+            .setDefaultValue(ModelNode.FALSE)
+            .setAllowExpression(true)
+            .build();
```
pferraro, line 96: This new attribute requires a transformer to handle the case where this attribute is defined and the model needs to be transformed for a host using an older version of the model.

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
```diff
@@ -89,12 +89,20 @@
...
 
     static final HostDefinition INSTANCE = new HostDefinition();
-    private static final Collection<AttributeDefinition> ATTRIBUTES = Collections.unmodifiableCollection(Arrays.asList(ALIAS, DEFAULT_WEB_MODULE, DEFAULT_RESPONSE_CODE, DISABLE_CONSOLE_REDIRECT, QUEUE_REQUESTS_ON_START));
+    private static final Collection<AttributeDefinition> ATTRIBUTES = Collections.unmodifiableCollection(
+            Arrays.asList(ALIAS, DEFAULT_WEB_MODULE, DEFAULT_RESPONSE_CODE, DISABLE_CONSOLE_REDIRECT,
+                    QUEUE_REQUESTS_ON_START, ACTIVE_REQUEST_TRACKING_ENABLED));
     private static final List<? extends PersistentResourceDefinition> CHILDREN = Collections.unmodifiableList(Arrays.asList(
             LocationDefinition.INSTANCE,
             AccessLogDefinition.INSTANCE,
+            ActiveRequestTrackerDefinition.INSTANCE,
```
pferraro, line 105: This new resource requires a transformer to handle the case where this resource is defined and the model needs to be transformed for a host using an older version of the model, which has no such resource registered.

undertow/src/main/java/org/wildfly/extension/undertow/ActiveRequestTrackerService.java
jasondlee: I think I have this cleaned up as requested locally. I'll ping you when the PR is updated.

undertow/src/main/java/org/wildfly/extension/undertow/Host.java
jasondlee: Updated.

bstansberry:
@jasondlee  Flavia says #/18473 supersedes this, so I'm going to close it. If that's wrong somehow, please coordinate with Flavia.

PR #15408, opened 2022-04-06
WFLY-16245 org.wildfly.clustering.ejb.client is missing dependency on org.wildfly.clustering.marshalling.spi

pferraro:
https://issues.redhat.com/browse/WFLY-16245


rhusar:
@rachmatowicz Please review.

darranl:
@pferraro Do we have an upstream PR for this one?

@rachmatowicz if we can please get a review on this one I have a small window now to include in WildFly 26.1.0

darranl (approved):
Approving as already accepted upstream,

PR #15417, opened 2022-04-12
[WFLY-14266] Connector report directory: don't make resource adapter …

tadamski:
…deployer dependent on resource adapter subsystem in appclient

This issue fixes a bug introduced by the RFE above.


scottmarlow:
Jakarta EE 8 Platform TCK results show that tests previously failing are now passing.

PR #15418, opened 2022-04-12
[WFLY-14266] Connector report directory: don't make resource adapter …

tadamski:
…deployer dependent on resource adapter subsystem in appclient

This issue fixes a bug introduced by the RFE above.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
+++ b/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
@@ -81,7 +81,10 @@ import org.jboss.msc.service.ServiceTarget;
  */
 public class ParsedRaDeploymentProcessor implements DeploymentUnitProcessor {
 
-    public ParsedRaDeploymentProcessor() {
+    private final boolean appclient;
+
+    public ParsedRaDeploymentProcessor(final boolean appclient) {
+        this.appclient = appclient;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
```diff
@@ -208,8 +211,11 @@ public static ServiceBuilder<ResourceAdapterDeployment> process(final ConnectorX
                 .addDependency(ConnectorServices.MANAGEMENT_REPOSITORY_SERVICE, ManagementRepository.class, raDeploymentService.getManagementRepositoryInjector())
                 .addDependency(ConnectorServices.RESOURCE_ADAPTER_REGISTRY_SERVICE, ResourceAdapterDeploymentRegistry.class, raDeploymentService.getRegistryInjector())
                 .addDependency(support.getCapabilityServiceName(ConnectorServices.TRANSACTION_INTEGRATION_CAPABILITY_NAME), TransactionIntegration.class, raDeploymentService.getTxIntegrationInjector())
-                .addDependency(ConnectorServices.CONNECTOR_CONFIG_SERVICE, JcaSubsystemConfiguration.class, raDeploymentService.getConfigInjector())
-                .addDependency(ConnectorServices.RESOURCEADAPTERS_SUBSYSTEM_SERVICE, ResourceAdaptersSubsystemService.class, raDeploymentService.getResourceAdaptersSubsystem());
+                .addDependency(ConnectorServices.CONNECTOR_CONFIG_SERVICE, JcaSubsystemConfiguration.class, raDeploymentService.getConfigInjector());
+
+            if (!appclient) {
```
scottmarlow, line 216: Is there a reason why the RESOURCEADAPTERS_SUBSYSTEM_SERVICE is not available in the Application Client Container?  But is available when not running the Application Client Container?

I'm trying to understand under what conditions we know that the RESOURCEADAPTERS_SUBSYSTEM_SERVICE should be expected to be available.

darranl: If this is in a deployment unit processor we may be able to more reliably check using a capability check - we did similar as we were removing legacy security so we can tell if the subsystem is present.

bstansberry: +1 to this discussion. I'm not going to block merging this but we should follow up on this ASAP. AIUI the basic issue here is we have a case where the jca subsystem is in the config but the resource-adapters subsystem is not. That's the case in our std appclient.xml but the same thing could happen in other contexts.

PR #15419, opened 2022-04-12
[WFLY-16251] Add an initial migrate operation for picketlink-federation that just migrates the default (empty) configuration to keycloak-saml

fjuma:
https://issues.redhat.com/browse/WFLY-16251

### DIFF
--- a/picketlink/src/main/java/org/wildfly/extension/picketlink/federation/MigrateOperation.java
+++ b/picketlink/src/main/java/org/wildfly/extension/picketlink/federation/MigrateOperation.java
@@ -1,0 +1,338 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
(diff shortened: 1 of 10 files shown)


picketlink/src/main/java/org/wildfly/extension/picketlink/federation/MigrateOperation.java
```diff
@@ -0,0 +1,339 @@
...
+            newAddOp.get(OP_ADDR).set(newAddress);
+
+            PathAddress address = PathAddress.pathAddress(newAddress);
+
+            if (address.size() > 1) {
+                // non-empty subsystem configuration, fail for now
+                throw ROOT_LOGGER.cannotMigrateNonEmptyConfiguration();
+            }
+
+            if (newAddOp.isDefined()) {
```
bstansberry, line 266: Why wouldn't it be defined at this point?

bstansberry (approved):
My question re why newAddOp would be undefined is a nit; this LGTM.

picketlink/src/main/java/org/wildfly/extension/picketlink/federation/MigrateOperation.java
fjuma: Good catch, fixed.

PR #15423, opened 2022-04-13
Jakarta JSON Upgrades for Jakarta EE 10

jamezp:
* https://issues.redhat.com/browse/WFLY-16263
* https://issues.redhat.com/browse/WFLY-16264
* https://issues.redhat.com/browse/WFLY-16200
* https://issues.redhat.com/browse/WFLY-16262

Upgrade Jakarta JSON Processing to 2.1.0 and migrate to using Eclipse Parsson. Also upgrade Jakarta JSON Binding to 3.0.0 and Yasson to 3.0.0-RC1.

### DIFF
--- a/ee-9/common/src/main/resources/modules/system/layers/base/org/eclipse/parsson/main/module.xml
+++ b/ee-9/common/src/main/resources/modules/system/layers/base/org/eclipse/parsson/main/module.xml
@@ -1,16 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!--
-  ~ JBoss, Home of Professional Open Source.
-  ~
-  ~ Copyright 2020 Red Hat, Inc., and individual contributors
-  ~ as indicated by the @author tags.
+  ~ Copyright 2022 Red Hat, Inc.
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
   ~ You may obtain a copy of the License at
   ~
-  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~   http://www.apache.org/licenses/LICENSE-2.0
   ~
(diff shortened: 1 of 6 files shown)


ee-9/common/src/main/resources/modules/system/layers/base/org/eclipse/parsson/main/module.xml
```diff
@@ -0,0 +1,34 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<!-- TODO Consider dropping direct dep on core-galleon-pack and instead using
+     a source dep, eliminating the need to re-provision this just to get a
+     different version. -->
```
bstansberry, line 20: I don't think this is relevant to this file. IIRC I wrote it somewhere. I doubt it's relevant there any more either. ;)

jamezp: Yeah. I almost removed it too. Then though what if it means something to core :) I can remove it though.

bstansberry: Well. I'm wrong. It's relevant to this file (assuming you'll add this module in core) but I guess isn't relevant to the org.glassfish.jakarta.json one.

If this module.xml file and the alias one are going into core with the same content, I believe both of these can be removed once the core upgrade with them in it is integrated. The TODO I was talking about is done.

bstansberry (approved):

bstansberry:
Thanks @jamezp

PR #15428, opened 2022-04-14
[WFLY-16266]:Upgrade saaj to 1.5.3

jimma:
https://issues.redhat.com/browse/WFLY-16266


bstansberry:
@jimma Is this safe for a bug fix / micro release?

jimma:
@bstansberry It's safe and this is kind of late upgrade. CXF 3.4.5 has saaj 1.5.3:  https://github.com/apache/cxf/blob/cxf-3.4.5/parent/pom.xml#L191, so it's good to have this upgrade in 26.x too.

bstansberry:
Thanks, @jimma

PR #15433, opened 2022-04-18
WFLY-16272 Fix IllegalArgumentException while using an expression in use-java-context

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-16272

@tadamski  please review

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Util.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Util.java
@@ -46,7 +46,7 @@ public class Util {
      */
     public static String getJndiName(final OperationContext context, final ModelNode modelNode) throws OperationFailedException {
         final String rawJndiName = JNDI_NAME.resolveModelAttribute(context, modelNode).asString();
-        return cleanJndiName(rawJndiName, modelNode.hasDefined(USE_JAVA_CONTEXT.getName()) && modelNode.get(USE_JAVA_CONTEXT.getName()).asBoolean());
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Util.java
```diff
@@ -46,7 +46,7 @@ public class Util {
      */
     public static String getJndiName(final OperationContext context, final ModelNode modelNode) throws OperationFailedException {
         final String rawJndiName = JNDI_NAME.resolveModelAttribute(context, modelNode).asString();
-        return cleanJndiName(rawJndiName, modelNode.hasDefined(USE_JAVA_CONTEXT.getName()) && modelNode.get(USE_JAVA_CONTEXT.getName()).asBoolean());
+        return cleanJndiName(rawJndiName, modelNode.hasDefined(USE_JAVA_CONTEXT.getName()) && USE_JAVA_CONTEXT.resolveModelAttribute(context, modelNode).asBoolean());
```
bstansberry, line 49: The modelNode.hasDefined(USE_JAVA_CONTEXT.getName()) is wrong too. USE_JAVA_CONTEXT has a default value of 'true', so that call is disabling the use of that default value.

soul2zimate:
Thanks, I have removed that piece of code.

soul2zimate:
Hi @bstansberry  @tadamski May I get your approval on this ? so I can backport this to EAP.

bstansberry (approved):
This is fine; the code executing the runtime mgmt ops reads the model to determine this setting and the PR in this logic is the correct way to read the model.

PR #15435, opened 2022-04-18
WFLY-16256 - Tracer Instance Leak in WildFlyClientTracingRegistrarProvider

jasondlee:
https://issues.redhat.com/browse/WFLY-16256

Add scope modifier to ensure bean instance/context is released

### DIFF
--- a/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
+++ b/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
@@ -21,16 +21,17 @@
  */
 package org.wildfly.microprofile.opentracing.smallrye;
 
-import io.opentracing.Tracer;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
-
+import javax.enterprise.context.RequestScoped;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.BeforeShutdown;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ProcessAnnotatedType;
+
+import io.opentracing.Tracer;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
 public class TracingCDIExtension implements Extension {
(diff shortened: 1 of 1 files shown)


microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
```diff
@@ -42,9 +43,10 @@ public static void registerApplicationTracer(ClassLoader classLoader, Tracer tra
     }
 
     public void registerTracerBean(@Observes AfterBeanDiscovery abd) {
-        abd.addBean().addTransitiveTypeClosure(Tracer.class).produceWith(i -> {
-            return TRACERS.get(WildFlySecurityManager.getCurrentContextClassLoaderPrivileged());
-        });
+        abd.addBean()
+                .scope(RequestScoped.class)
```
manovotn, line 47: I _think_ this change is wrong.
The `@Dependent` scope (default) seems deliberate here so that the impl can delegate to underlying map with CL as key. So long as you have known CL, you should always return the same instance of `Tracer`.
These tracer instances are registered via `TracingDeploymentProcessor` so I'd expect that the DUP is called once per deployment and adds only one tracer.
Furthermore, `Tracer` should be a single instance per running app which is something request scope doesn't fit.

Something else is likely holding on to these instances - this is usually some dynamic resolution (such as `Instance<>`) and never releases dependent instances. I'd say this needs to come from somewhere inside `WildFlyClientTracingRegistrarProvider` that's used behind the scenes in reproducer. I can try to look around a bit although I am not as familiar with OT integration code.

manovotn: I think I found the actual cause, it is the following line - https://github.com/wildfly/wildfly/blob/main/microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/WildFlyClientTracingRegistrarProvider.java#L54

It will redo the dynamic lookup for `Tracer` every single time and since it is just `CDI.current()` (and not an explicit `Instance`) it will use `WeldProvider` as the `Instance` and store the result there. `@Dependent` beans are by definition new instance on each lookup which  the current impl bypasses but the lookup is unaware of it.
This also explains why this PR "fixes" the problem as well - by using standard scope, the lookup will search for the bean in the active context instead of always assuming new instance. However, it is IMO not the correct fix.

Long story short, I will send a PR for it in an hour or so but I'd appreciate if someone else double checked the heap to verify I am not missing something else (@jasondlee perhaps? :))

manovotn:
This PR can be closed (which I don't have permissions for) as it is replaced by https://github.com/wildfly/wildfly/pull/15438

microprofile/opentracing-smallrye/src/main/java/org/wildfly/microprofile/opentracing/smallrye/TracingCDIExtension.java
jasondlee: That was all super helpful. :) I'll take a look at your PR and dispose of this one. :)

PR #15451, opened 2022-04-21
[WFLY-16285] Upgrade commons-lang3 3.12.0

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16285

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -333,7 +333,7 @@
         <version.commons-collections>3.2.2</version.commons-collections>
         <version.commons-digester>1.8.1</version.commons-digester>
         <version.commons-io>2.11.0</version.commons-io>
-        <version.commons-lang3>3.11</version.commons-lang3>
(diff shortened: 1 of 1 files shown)


jimma:
@boris-unckel  This looks good to me. Thanks !

darranl:
FYI at the time of this PR I don't believe we had a dependency on commons-lag3 as it had been removed by:
  https://github.com/wildfly/wildfly/commit/929b4d81e8f6bac9c2441f8da0c270d6ea4193b2

It has been added back again:
  https://github.com/wildfly/wildfly/blob/main/pom.xml#L3595-L3599

But only as a dependency for the Jakarta EE 9 variant of the web testsuite:
  https://github.com/wildfly/wildfly/blob/main/testsuite/preview/source-transform/web/pom.xml#L741-L745

I will trigger CI again.

bstansberry:
Thanks @boris-unckel

PR #15456, opened 2022-04-21
[WFLY-16290] DatabaseTimerPersistence loadSqlProperties should be cal…

bmaxwell:
https://issues.redhat.com/browse/WFLY-16290

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -130,6 +130,20 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service<Datab
     private MarshallingConfiguration configuration;
     private RefreshTask refreshTask;
 
+    /** database values */
+    private static final String POSTGRES = "postgres";
+    private static final String POSTGRESQL = "postgresql";
+    private static final String MYSQL = "mysql";
+    private static final String MARIADB = "mariadb";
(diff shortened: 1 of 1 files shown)


chengfang:
`investigateDialect` and `loadSqlProperties` methods currently have circular dependency: `investigateDialect` needs to adjust certain property in `sql` properties; and `loadSqlProperties` needs to check `database` attribute which may be set in `investigateDialect`.

`database` attribute is optional in ejb3 subsystem configuration, though most users would explicitly set it. When not set in configuration, its value is determined in `investigateDialect`.  In the proposed fix,  when `loadSqlProperties` is called first, and this method may access `database`, which is not initialized yet.

How about we don't change the order of `investigateDialect` and `loadSqlProperties`; instead, we move the 2 calls of `adjustCreateAutoTimerStatement` (root of NPE) out of `identifyDialect`, and into the first half of `loadSqlProperties`?

Inside `loadSqlProperties`, we can safely check `database` value, if it's `oracle` or `db2`, call the appropriate `adjustCreateAutoTimerStatement(...)'. `loadSqlProperties` is also a logically better place to adjust certain sql properties.

bmaxwell:
@chengfang  I have updated the PR

chengfang:
There are compilation failures due to some removed imports.

bmaxwell:
@chengfang  yes sorry, I was trying to switch to VSCode and it seems to remove stuff for reasons I do not understand :(

chengfang (approved):

bstansberry:
Thanks @bmaxwell

rikcarve:
Any chance this get's back-ported to 26.1?

PR #15458, opened 2022-04-21
WFLY-16291 WARN if invalid 'database' specified for EJB Database Time…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-16291

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
@@ -3257,4 +3257,8 @@ public interface EjbLogger extends BasicLogger {
     @LogMessage(level = WARN)
     @Message(id = 531, value = "No client mappings registry provider found for %s; using legacy provider based on static configuration")
     void legacyClientMappingsRegistryProviderInUse(String name);
+
+    @LogMessage(level = WARN)
+    @Message(id = 532, value = "Database detected from configuration is: '%s'. If this is incorrect, please specify the correct database.")
+    void unknownDatabaseName(String name);
 }
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -337,6 +337,8 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service<Datab
                 unified = MSSQL;
             } else if (name.contains(SYBASE) || name.contains(JCONNECT)) {
                 unified = SYBASE;
+            } else {
+                EjbLogger.EJB3_TIMER_LOGGER.unknownDatabaseName(name);
             }
          }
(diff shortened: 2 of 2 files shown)


chengfang:
in case of an unrecognized `database` configuration value, there are already debug logs from line 279-281:

```
EjbLogger.EJB3_TIMER_LOGGER.debugf("Database dialect '%s' read from configuration, adjusting it to match the final database valid value.", database);
database = identifyDialect(database);
EjbLogger.EJB3_TIMER_LOGGER.debugf("New Database dialect is '%s'.", database);
```

RanabirChakraborty:
Hello @chengfang yes there are already debug logs present but currently, there is no WARN logger there.
The users are nowadays adding different kinds of database names and also telling the support team to check for them. Because of the invalid database names, it's failing and also not showing any logger when the server starts, So I think adding WARN logger will be a good requirement here.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -337,6 +337,8 @@ private String identifyDialect(String name) {
                 unified = MSSQL;
             } else if (name.contains(SYBASE) || name.contains(JCONNECT)) {
                 unified = SYBASE;
+            } else {
+                EjbLogger.ROOT_LOGGER.unknownDatabaseName(name);
```
chengfang, line 341: `EJB3_TIMER_LOGGER` should be used here.

RanabirChakraborty: Thanks, @chengfang did the changes.

chengfang (approved):

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -3257,4 +3257,8 @@ public interface EjbLogger extends BasicLogger {
     @LogMessage(level = WARN)
     @Message(id = 531, value = "No client mappings registry provider found for %s; using legacy provider based on static configuration")
     void legacyClientMappingsRegistryProviderInUse(String name);
+
+    @LogMessage(level = WARN)
+    @Message(id = 532, value = "Database detected from configuration is: '%s' , If this is incorrect, please specify the correct database.")
```
rhusar, line 3262: This needs fixing: there is an extra space before the colon and "If" is incorrectly capitalized.

rhusar (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
RanabirChakraborty: Before the colon can't able to see any extra space, I guess it will be the comma. Changed it accordingly.

bstansberry: Actually it should either be a semi-colon and small i or a full-stop/period with a capital I.

RanabirChakraborty: Thanks @bstansberry changed it accordingly.

bstansberry (approved):

PR #15470, opened 2022-04-25
[WFLY-16304] Upgrade Apache james to 0.8.7

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16304

PR #15471, opened 2022-04-25
[26.x][WFLY-16304] Upgrade Apache james to 0.8.7

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16304
Upstream https://github.com/wildfly/wildfly/pull/15470


boris-unckel:
@jamezp And here the 26.x version of it.

jamezp (approved):

PR #15472, opened 2022-04-26
[WFLY-16305] Upgrade cryptacular to 1.2.5

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16305


jimma:
@boris-unckel  This looks good to me .

PR #15464, opened 2022-04-22
WFLY-16100, WFLY-16056 to add build compatible extension support

starksm64:
Signed-off-by: Scott M Stark <starksm64@gmail.com>
More work on https://issues.redhat.com/browse/WFLY-16100 that adds missing build compatible extension integration and dependencies. There is a problem with scanning deployments with BuildCompatibleExtensions as seen in this CDI test failure:

org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanWithLookup.SyntheticBeanWithLookupTest
```
org.jboss.weld.exceptions.AmbiguousResolutionException: WELD-001335: Ambiguous dependencies for type MyDependentBean with qualifiers 
 Possible dependencies: 
  - Managed Bean [class org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanWithLookup.MyDependentBean] with qualifiers [@Any @Default],
  - Managed Bean [class org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanWithLookup.MyDependentBean] with qualifiers [@Any @Default]
	at org.jboss.weld.core@5.0.0.CR2//org.jboss.weld.bean.builtin.InstanceImpl.checkBeanResolved(InstanceImpl.java:254)
	at org.jboss.weld.core@5.0.0.CR2//org.jboss.weld.bean.builtin.InstanceImpl.get(InstanceImpl.java:113)
	at deployment.SyntheticBeanWithLookupTest355982c2ff53323b8c1b8eb3ad186685e835f45.war//org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanWithLookup.MyPojoCreator.create(MyPojoCreator.java:16)
	at deployment.SyntheticBeanWithLookupTest355982c2ff53323b8c1b8eb3ad186685e835f45.war//org.jboss.cdi.tck.tests.build.compatible.extensions.syntheticBeanWithLookup.MyPojoCreator.create(MyPojoCreator.java:9)
	at org.jboss.weld.core@5.0.0.CR2//org.jboss.weld.lite.extension.translator.LiteExtensionTranslator.lambda$synthesis$0(LiteExtensionTranslator.java:157)
	at org.jboss.weld.core@5.0.0.CR2//org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.create(BeanConfiguratorImpl.java:372)
[...]

PR #15474, opened 2022-04-26
WFLY-15687 Add support for CDI 4 Build compatible extensions

manovotn:
This PR superseeds https://github.com/wildfly/wildfly/pull/15464. I have squashed Scott's commits into one and added mine on top.

JIRA link - https://issues.redhat.com/browse/WFLY-15687

There are still some remaining TCK failures, namely 12 of them.

> [ERROR] Tests run: 1851, Failures: 12, Errors: 0, Skipped: 74

10 of these are JSF related (old artifact using now removed deprecated APIs) - fixed by updating JSF impl artifact.

The remaining two are what concerns us:

* `SyntheticBeanWithLookupTest`- ~~this one is (most likely) in WFLY's integration code but I am failing to see how to properly fix that. The last commit in this PR adds a very nasty hack which makes the test pass, but I am anything but happy with it :-/~~
  * As captured in the comments below, this will now be addressed by https://issues.redhat.com/browse/WELD-2718
* `StartupShutdownTest` - this one needs to be fixed on Weld's side, see https://issues.redhat.com/browse/WELD-2717, there is already a pending PR

CC @starksm64


starksm64:
@manovotn Ok, great, I'll see if can get our Mojarra fork updated.

manovotn:
> @manovotn Ok, great, I'll see if can get our Mojarra fork updated.

@starksm64 That fork is long updated now. I first sent https://github.com/jboss/mojarra/pull/90 which was forgotten and then later Brian did https://github.com/jboss/mojarra/pull/92 which got merged.
But I think there was no release since.

jasondlee:
I'm working on making a release of jsf-impl now. There's some weirdness with the versions I need to figure out, but I'm actively working on this.

manovotn:
I've spent some "quality" tie debugging the problem with `SyntheticBeanWithLookupTest` and I think I've found a way to workaround this in a cleaner fashion directly in Weld.
Following JIRA captures it along with PR - https://issues.redhat.com/browse/WELD-2718

Therefore, I'll remove the hacky approach from this PR and we can probably turn the draft into actual PR.
Once Weld 5 Final is out, WFLY can update and with that should have a passing CDI TCK.

jasondlee:
The JSF update now has a PR: https://github.com/wildfly/wildfly/pull/15487

manovotn:
Nice @jasondlee, thanks!
Once we merge both PRs, I could bring back WFLY testing on Weld PRs (if there is a nightly WFP build).

bstansberry:
@manovotn The JSF upgrade is merged. We produce a WFP wildfly-preview-latest-SNAPSHOT.zip nightly from the  https://ci.wildfly.org/viewType.html?buildTypeId=WF_WildflyPreviewNightly job.

bstansberry (changes requested):
@manovotn 

Please add entries for the 2 new deps to:

https://github.com/wildfly/wildfly/blob/main/ee-9/common/src/main/resources/license/preview-feature-pack-common-licenses.xml

https://github.com/wildfly/wildfly/blob/main/ee-9/ee-9-api/src/main/resources/license/preview-feature-pack-ee-9-api-licenses.xml

Note that the elements in those files are ordered by maven GA.

Other than that, LGTM.

manovotn:
> @manovotn
> 
> Please add entries for the 2 new deps to:
> 
> https://github.com/wildfly/wildfly/blob/main/ee-9/common/src/main/resources/license/preview-feature-pack-common-licenses.xml
> 
> https://github.com/wildfly/wildfly/blob/main/ee-9/ee-9-api/src/main/resources/license/preview-feature-pack-ee-9-api-licenses.xml
> 
> Note that the elements in those files are ordered by maven GA.
> 
> Other than that, LGTM.

@bstansberry updated

manovotn:
I have also released Weld 5 Final today, so I'll update this PR with it

bstansberry:
Those jobs failed as well, including with SE 17.

The dependencyManagement we use when running the MP 5 TCKs with WFP is out of date though, so I'm going to get that cleaned up and then I'll see if that makes the problem go away.

manovotn:
> @manovotn @kabir The WildFly Preview job failed with a java.lang.VerifyError in the MP Reactive Messaging TCK

I am afraid I have no idea why. I am not aware of any change in Weld between CR2 and Final that could cause it. Besides, if it were in Weld, I assume we'd see failures sooner and/or all over the place.
Did the TCK config change recently?

bstansberry:
@manovotn I'm confused as well. I thought it might be due to WFLY-16337 (i.e. the changes here surfaced an existing problem) but a test branch combining this and a fix for that had the same issue:

https://github.com/wildfly/wildfly/pull/15516

I'll discuss with @kabir as he's our RSO expert. AIUI these failures are on the client side. The client-side classpath is tricky.

We've been running PRs against WFP+SE17 for a long time, plus nightlies. So something is odd here.

scottmarlow:
> VerifyError

https://www.eclipse.org/lists/cdi-dev/msg00586.html is reporting the same VerifyError

bstansberry:
@manovotn I've been trying simple component upgrades (i.e. not the rest of this PR) and this problem appears with this commit:

https://github.com/weld/core/commit/1f0b5c63739df5d94f1ac21a4f835a93e16141cd

To reproduce in ee-9/pom.xml I update the CDI API to 4.0.0, weld-api to 5.0.CR2, and weld-core to a SNAPSHOT built from that sha. Build the server and then

mvn install -Dts.ee9 -pl testsuite/integration/microprofile-tck/reactive-streams-operators

If I use weld 5.0.0.CR2 instead of that snapshot the problem doesn't appear.

manovotn:
@bstansberry the commit you linked is a POM change in Weld's SE artifacts and those aren't used in WFLY at all from what I know. Also performing a build with a revert on this commit doesn't solve anything.

If I were to guess, you meant https://github.com/weld/core/commit/18af30b7ae367ca2dac107d22b64c5c596c7bb6d?
That one seems to be the turning point; if I revert that, I get the TCK to pass.

That's weird, the issue was linked primarily to self interception and if I am not mistaken, the class being proxied is [this one](https://github.com/eclipse/microprofile-reactive-streams-operators/blob/master/tck/src/main/java/org/eclipse/microprofile/reactive/streams/operators/tck/arquillian/ReactiveStreamsCdiTck.java) and that seems just fine. At least we know what's causing it, but I am still trying to grasp why.

starksm64:
Based on:
https://stackoverflow.com/questions/42707610/use-of-stackmap-frames-and-how-it-helps-in-byte-code-verification

And the code in this error msg:
org/glassfish/jms/injection/RequestedJMSContextManager$Proxy$_$$_WeldClientProxy.toString()Ljava/lang/String;
@4: ifne
Reason:
Expected stackmap frame at this location.
Bytecode:
0000000: 2ab4 0018 9a00 082a b700 5bb0 b800 2e4c
0000010: 2ab4 0034 b600 3ac0 001c 59b6 005c 2b59
0000020: c600 0cc0 0030 b600 3da7 0004 57a7 0013
0000030: 2b59 c600 0cc0 0030 b600 3da7 0004 57bf
0000040: 5aa5 0004 b02a c000 28b0
Exception Handler Table:
bci [16, 48] => handler: 48

Which maps to:
Begin method:
2a: __ALoad0
b4: GetField(18-66)
9a: __IfNe(82a)
2a: __ALoad0
b7: InvokeSpecial(5b-50)
b0: __AReturn
b8: InvokeStatic(2e4c)
4c: __AStore1
2a: __ALoad0
b4: GetField(34-4a)
b6: InvokeVirtual(3a-40)
c0: CheckCast(1c59)
59: Dup
b6: InvokeVirtual(5c2b)
2b: __ALoad1
59: Dup
c6: __IfNonNull(c-40)
c0: CheckCast(30-4a)
b6: InvokeVirtual(3d-59)
a7: Goto(457)
57: Pop
a7: Goto(132b)
2b: __ALoad1
59: Dup
c6: __IfNonNull(c-40)
c0: CheckCast(30-4a)
b6: InvokeVirtual(3d-59)
a7: Goto(457)
57: Pop
bf: AThrow
5a: DupX1
a5: __IfACmpEq(4-50)
b0: __AReturn
2a: __ALoad0
c0: CheckCast(28-50)
b0: __AReturn
End method:

It seems the verifier is expecting a stackmap entry for the Ifne
instruction jump that is not there. You have to see the full class file to
be able to see the stackmap table. Maybe that protected stuff is requiring
a new stackmap entry.

bstansberry:
Sorry for the bad link @manovotn :( 

Yes, https://github.com/weld/core/commit/18af30b7ae367ca2dac107d22b64c5c596c7bb6d is the one I meant

manovotn:
So, the https://github.com/weld/core/commit/18af30b7ae367ca2dac107d22b64c5c596c7bb6d is definitely the cause.
It seems to happen when there is a hierarchy of at least two classes, where the super class is in a different package and has `protected` methods which are then subject to proxying by Weld.

The error, however, is completely off because those proxy files don't even have a stackmap in the first place! (you can see that if you inspect them with `javap -v -p`)
jboss-classfilewriter apparently creates version 50.0 files (JDK 6) which can, by JLS definition, lack the stackmap in which case JVM does type inference. So the `VerifyError` is actually just misleading and the cause if likely somewhere else. This is also confirmed by the fact that the error points to `toString` method which hasn't changed in any way and the generated bytecode for this method is identical between CR2 and Final.
That said, I was unable to determine the actual cause so far.

All I know is that from the offending commit, the change is that `protected` methods on proxy are now invoked via `invokevirtual`/`invokeinterface` instead of via method handler invocation (reflection). This should, in theory, be valid but instead blows up :-/

bstansberry:
In case it's useful, ReactiveStreamsCdiTck is a javax.* namespace class, which means it's getting bytecode transformed when installed into WFP.

kabir:
It might be interesting to see the difference in bytecode between the original and transformed class. @ropalka (I think you worked on this) - is there a hook for @manovotn to be able to dump the transformed class bytes?

manovotn:
@kabir I managed to create a reproducer yesterday and have bytecode ([weld can dump all created proxies via an option](https://docs.jboss.org/weld/reference/latest/en-US/html_single/#_debugging_generated_bytecode)).
Still didn't help though; I suspect that there is a catch somewhere in classfilewriter in combination with bytecode verification but it is more of a guess than anything else.

I will roll back the problematic commit and issue a new Weld release.

bstansberry:
Thanks @manovotn!

FYI, the Windows - JDK 11 job is going to show some weird failures that we're investigating and that have popped up elsewhere, so not related to this PR. So don't worry about them. I kicked off a re-run of that job.

bstansberry (approved):

bstansberry:
Thanks, @manovotn & @starksm64 !

starksm64:
I did not make that comment 8 hours ago, so where did it come from? It seems like it would have been a comment made on May 5 where I then commented about the bytecode in the stack trace. It certainly is not relevant now.

manovotn:
> I did not make that comment 8 hours ago,

Yea, that's what confused me. Probably a glitch in the Matrix :)

PR #15475, opened 2022-04-26
[WFLY-16307] Upgrade io.smallrye.config to 2.10.0

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16307


boris-unckel:
@kabir This is the patch level update only. I have seen 2.10 is released, too. Shall we upgrade to 2.10 in main?
Second question: The 2.9.1 has not been backported to 26.x. Is this related to the MicroProfile version? Or can we go up to 2.9.2/2.10?

kabir:
@boris-unckel We can upgrade to 2.10 in both places. The next MP version (aligned with jakarta.x) will be in the 3.x series of SmallRye Config. Thank you for looking into this :-)

boris-unckel:
I have to fix the commit message.

bstansberry:
Thanks, @boris-unckel

PR #15485, opened 2022-04-27
[WFLY-16299] Migrate to Jakarta EE variant of the Elytron subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-16299

### DIFF
--- a/ee-9/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
+++ b/ee-9/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2015, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
+  ~ distribution for a full listing of individual contributors.
+  ~
+  ~ This is free software; you can redistribute it and/or modify it
+  ~ under the terms of the GNU Lesser General Public License as
+  ~ published by the Free Software Foundation; either version 2.1 of
+  ~ the License, or (at your option) any later version.
+  ~
+  ~ This software is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


ee-9/common/src/main/resources/modules/system/layers/base/org/wildfly/extension/elytron/main/module.xml
```diff
@@ -0,0 +1,59 @@
...
+
+    <dependencies>
+        <module name="java.logging"/>
+        <module name="java.naming"/>
+        <module name="java.security.jgss"/>
+        <module name="java.security.sasl"/>
+        <module name="java.sql"/>
+        <module name="java.xml"/>
+        <module name="javax.security.jacc.api"/>
+        <module name="javax.security.auth.message.api"/>
```
jamezp, line 47: Should these be `jakarta.*`?

yersan: So far, we are not changing them. There are aliases for these ones at ee-9 module

darranl: +1 I am trying to not change this module too much so rely on the aliases.  

This module is just to swap the suffix of the artifact name.  I couldn't do the same we do with other modules to use a property for the suffix as if I added that to WildFly Core the WildFly builds would fail immediately so if I went for cloning the module.xml for now it meant the core PR could be merged without breaking anything.

yersan: Is by any chance the old JBoss Module module leaked into the ee-9 build? I mean, I can see how the dependencies are being excluded, but I don't see how the old Jboss Module wildfly-elytron-integration module is excluded. The exclusions are excluding dependencies, but the old module is a resource.

If ee-9 can be built, it should be fine, I have not built your topic branches locally, so just wondering how it will work.

yersan: Ahh, wait, forget it, it is just an overwrite on full. Now I got it. The module is being overwritten, it is not leaked into the ee-9 build.

fjuma (approved):

yersan (approved):

PR #15488, opened 2022-04-27
WFLY-16313 MP OpenAPI subsystem should not register OpenAPI endpoint for non-JAX-RS WAR unless spec requires it

pferraro:
https://issues.redhat.com/browse/WFLY-16313


bstansberry (approved):

bstansberry:
Thanks @pferraro

PR #15490, opened 2022-04-27
[WFLY-16089] Upgrade Weld version for EE8 (3.1.9)

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16089
A rare case where 26.x was patched first: https://github.com/wildfly/wildfly/commit/92d02eeaafc6212325e1055dfa41cc9b8a64a66c

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -481,7 +481,7 @@
         <version.org.jboss.spec.javax.xml.rpc.jboss-jaxrpc-api_1.1_spec>2.0.0.Final</version.org.jboss.spec.javax.xml.rpc.jboss-jaxrpc-api_1.1_spec>
         <version.org.jboss.spec.javax.xml.soap.jboss-saaj-api_1.4_spec>1.0.2.Final</version.org.jboss.spec.javax.xml.soap.jboss-saaj-api_1.4_spec>
         <version.org.jboss.spec.javax.xml.ws.jboss-jaxws-api_2.3_spec>2.0.0.Final</version.org.jboss.spec.javax.xml.ws.jboss-jaxws-api_2.3_spec>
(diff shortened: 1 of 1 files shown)


boris-unckel:
@manovotn I only checked main before I created the duplicated Jira, sorry. I think now it's better with your original Jira and link to the commit. Could you review and approve, please?

manovotn:
@boris-unckel I didn't update `main` because as far as I know, this branch is targeting EE 10 so the version of WEEld for EE 8 doesn't really matter. Am I right @bstansberry?

boris-unckel:
@manovotn Looking forward to the published versions you're right. I'm looking at it with the chance to have the latest version of JavaSE 11 and JakartaEE 8 in great shape. In my company and project this might be the best bridge before we can achieve the Java-Package jakarta.* namespace. Therefor I'm working a lot to have WF26 and WF27 beeing patched before the big-bang "JakartaEE 10 goes main" happens.

darranl (changes requested):
Thank you for the PR, I am just going to place on hold for now - I am just going to check with Brian but we are getting ready to work on the big changes to drop Jakarta EE 8 support from main (we already know there will be no future Jakarta EE 8 releases from main) so I think we are ready to stop accepting changes that affect EE 8 only on main.

bstansberry:
@boris-unckel @darranl @manovotn Since this passes CI and AFAIK doesn't interfere with anything I don't have any objection. I thought about doing it when we upgraded 26.x just due to the general upstream first policy we have, but I was lazy and got distracted.

manovotn:
Sure, it doesn't interfere.

> I thought about doing it when we upgraded 26.x just due to the general upstream first policy we have, but I was lazy and got distracted.

You can always blame it on my laziness that I only sent the PR for 26.x branch ;-)

PR #15496, opened 2022-04-28
[WFLY-16323] Upgrade net.bytebuddy to 1.12.9

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16323


scottmarlow:
Switching to ByteBuddy 1.12.9 makes sense since that is what Hibernate ORM 6.0.1.Final is using.

scottmarlow (approved):

bstansberry:
Thanks @boris-unckel

PR #15498, opened 2022-04-28
[WFLY-16325] Upgrade smallrye-opentracing 2.1.0

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16325


boris-unckel:
Won't fix. Stuck in MP compatibility: The TCK has no version compliant with smallrye-opentracing 2.1.0 transitive contrib dependencies.

PR #15500, opened 2022-04-29
[WFLY-16327] Upgrade Apache Avro to 1.10.2

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16327

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,7 +385,7 @@
         <version.net.shibboleth.utilities.java-support>7.3.0</version.net.shibboleth.utilities.java-support>
         <version.org.apache.activemq.artemis>2.19.1</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>1.0.2</version.org.apache.activemq.artemis.native>
-        <version.org.apache.avro>1.7.6</version.org.apache.avro>
+        <version.org.apache.avro>1.10.2</version.org.apache.avro>
         <version.org.apache.cxf>3.5.1</version.org.apache.cxf>
         <version.org.apache.cxf.xjcplugins>3.3.1</version.org.apache.cxf.xjcplugins>
         <version.org.apache.ds>2.0.0-M24</version.org.apache.ds>


boris-unckel:
@scottmarlow All links are in the Jira, including the alignment with upstream Hibernate. Could you review and approve, please?

yrodiere (approved):

scottmarlow (approved):

scottmarlow (approved):

bstansberry:
@yrodiere Following the merge of #15185, std WildFly is on 1.11.0 in main. We could move the 1.11.0 version property to ee-9/pom.xml to have it only take effect in WFP, and then make this change in std WF. Simply adding ```<version.org.apache.avro>1.11.0</version.org.apache.avro>``` does is enough to get WFP using 1.11.0. Or, if 1.11.0 is fine with the current std WF we could just close this PR as out of date.

Either way #15501 is ok as main being 1.11.0 meets the upstream first requirement for moving 26.x to 1.10.2.

We'll be converting std WF to use the WFP dependency set sometime in the next month.

yrodiere:
> Simply adding <version.org.apache.avro>1.11.0</version.org.apache.avro> is enough to get WFP using 1.11.0.

Note that I also had to make some changes to the Maven exclusions and the JBoss module itself in order to upgrade to Avro 1.11, so changing the version is not *just* about updating the Maven property: https://github.com/wildfly/wildfly/pull/15185/commits/96765275f022d77cc374adc5202e05315a4d8787 . The changes might also work with Avro 1.10, though; I don't know.

> Or, if 1.11.0 is fine with the current std WF we could just close this PR as out of date.

I don't know if it's fine; I didn't test that in Hibernate Search upstream, and it seems we don't have any tests involving Hibernate Search 5 + Avro in WildFly.

However, closing this PR would still be my preference.

I think the intent is to move to Hibernate ORM 6 + Search 6 in the next release of WF based off the main branch, right? If so, the problem (if any) will disappear as we remove Hibernate Search 5 from WildFly.

> Either way https://github.com/wildfly/wildfly/pull/15501 is ok as main being 1.11.0 meets the upstream first requirement for moving 26.x to 1.10.2.

Right, @boris-unckel 's work was necessary whatever we do with this PR: we needed the upgrade to Avro 1.10 in the 26.x branch.

PR #15501, opened 2022-04-29
[26.x][WFLY-16327] Upgrade Apache Avro to 1.10.2

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16327
Upstream https://github.com/wildfly/wildfly/pull/15500


boris-unckel:
@scottmarlow And here the 26.x version of it. Could you review and approve, please?

scottmarlow:
It looks like Hibernate Search `5.10` branch (https://github.com/hibernate/hibernate-search/blob/5.10/pom.xml#L207) is already using Avro 1.10.2 so it probably makes sense to upgrade to Avro `1.10.2` here as well.

@boris-unckel are you upgrading Avro for use with Hibernate Search?  Or for a different reason?

CC @yrodiere

boris-unckel:
@scottmarlow Yes, that is where I looked it up. I have updated the Jira ticket.

yrodiere (approved):

PR #15508, opened 2022-04-29
[WFLY-16334] Upgrade Apache Thrift 0.16.0

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16334


boris-unckel:
@bstansberry @bstansberry The changelog of thrift is not very readable, I agree. A local test run was fine, but for dynamic use in tracing that must not be a good test coverage to have integration only.

boris-unckel:
PR withdrawn: The aligned version of io.jaegertracing:jaeger-thrift is needed here.

PR #15522, opened 2022-05-02
WFLY-16126 [allowlist/blocklist] Problematic Language usage deprecation and replacement in messaging

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-16126


darranl (changes requested):
This can't be handled as a global find and replace across the whole subsystem.  

Although we may be updating the language used in later versions of the model we can not start introducing incompatible changes like this in older configurations.

Firstly a change like this is going to need to check if a sybsystem bump is needed.
Then depending on how the subsystem handles it's versioning as these vary there will be a need to work out how to update the latest schema and current model without impacting the support for the older versions of the configuration.
Regarding model level changes there will be a question if we want to go straight with the replacement attribute names or deprecate and use an alternative to encourage migration to keep existing CLI scripts functioning.
Then if the names used in the model have changed transformers are going to be needed to convert to the original names when working with an older secondary host controller.

ehsavoie:
Superseded by https://github.com/wildfly/wildfly/pull/15603

PR #15524, opened 2022-05-03
POC of client-side jakarta namespace transformation of testsuite/doma…

bstansberry:
…in deployments


bstansberry:
@darranl I've fixed the test failures I was seeing with earlier iterations of this. As this is just a POC and you're aware of its branch, I'm going to go ahead and close it now.

PR #15529, opened 2022-05-05
[WFLY-16349] Upgrade to jakarta.* versions of MP reactive messaging

kabir:
https://issues.redhat.com/browse/WFLY-16349


bstansberry:
@kabir This will conceptually conflict with 2 open PRs:

https://github.com/wildfly/wildfly/pull/15515
https://github.com/wildfly/wildfly/pull/15528

The first one might need some work as it might be doing more dependencyMgmt stuff than is strictly necessary. But anyway we should align on how/where we manage these TCK versions.

The latter is partly about getting rid of use of jakarta.el:jakarta-el-api, which we won't use at runtime (we use a spec fork, which now has a release) so we shouldn't use it at compile or test time. So we shouldn't introduce another use of it here.

testsuite/integration/microprofile-tck/reactive-messaging/pom.xml
```diff
@@ -154,7 +157,27 @@
                 <ts.microprofile-tck-provisioning.fp.artifactId>wildfly-preview-feature-pack</ts.microprofile-tck-provisioning.fp.artifactId>
                 <!-- Enable bootable jar packaging -->
                 <ts.bootable-jar-microprofile-tck-packaging.phase>process-test-resources</ts.bootable-jar-microprofile-tck-packaging.phase>
+
+                <version.jakarta.enterprise>4.0.0-RC2</version.jakarta.enterprise>
```
bstansberry, line 161: Please use 4.0.0 as it's what the production code uses.

testsuite/integration/microprofile-tck/reactive-messaging/pom.xml
```diff
@@ -154,7 +157,27 @@
...
+                <version.org.eclipse.microprofile.reactive-messaging.api>${version.eclipse.mp.rm.for.jakarta}</version.org.eclipse.microprofile.reactive-messaging.api>
+                <version.jakarta.el.jakarta-el-api>5.0.0-RC1</version.jakarta.el.jakarta-el-api>
```
bstansberry, line 163: Please change this to 
```
<version.org.jboss.spec.jakarta.el.jboss-el-api_5.0_spec>4.0.0.Alpha1</version.org.jboss.spec.jakarta.el.jboss-el-api_5.0_spec>
```

testsuite/integration/microprofile-tck/reactive-messaging/pom.xml
```diff
@@ -154,7 +157,27 @@
...
+                    <artifactId>jakarta.el-api</artifactId>
+                    <version>${version.jakarta.el.jakarta-el-api}</version>
+                    <exclusions>
+                        <exclusion>
+                            <artifactId>*</artifactId>
+                            <groupId>*</groupId>
+                        </exclusion>
+                    </exclusions>
+                </dependency>
+            </dependencies>
```
bstansberry, line 179: Please change to

```


                    <dependency>
                        <groupId>org.jboss.spec.jakarta.el</groupId>
                        <artifactId>jboss-el-api_5.0_spec</artifactId>
                        <version>${version.org.jboss.spec.jakarta.el.jboss-el-api_5.0_spec}</version>
                        <exclusions>
                            <exclusion>
                                <groupId>*</groupId>
                                <artifactId>*</artifactId>
                            </exclusion>
                        </exclusions>
                    </dependency>
```

testsuite/integration/microprofile-tck/reactive-messaging/pom.xml
```diff
@@ -170,8 +193,25 @@
                 <version.jakarta.enterprise>2.0.2</version.jakarta.enterprise>
                 <!--<version.jakarta.inject.jakarta.inject-api>2.0.0</version.jakarta.inject.jakarta.inject-api>-->
 
+                <version.jakarta.enterprise>4.0.0-RC2</version.jakarta.enterprise>
+                <version.org.eclipse.microprofile.reactive-messaging.api>${version.eclipse.mp.rm.for.jakarta}</version.org.eclipse.microprofile.reactive-messaging.api>
+                <version.jakarta.el.jakarta-el-api>5.0.0-RC1</version.jakarta.el.jakarta-el-api>
```
bstansberry, line 198: Same comments as above apply to this profile.

bstansberry (changes requested):
@kabir 

I need to make some further tweaks to #15515 and I don't want this blocked by those. I've made comments below that will get this to the correct GAVs. When this is merged I'll adapt the #15515 stuff to account for this changeset.

kabir:
@bstansberry I think it should be in order now. I tried to do this before finishing on Friday but that el-api dependency made it a bit more complicated than first thought.

bstansberry (approved):
Thanks @kabir 

On #15515 I'll deal with the CDI API exclude of the jakarta.el dep in a higher level pom.

PR #15532, opened 2022-05-05
[WFLY-16339] Add testsuite/domain to the ts.ee9 profile

darranl:
https://issues.redhat.com/browse/WFLY-16339


bstansberry:
Thanks, @darranl

PR #15540, opened 2022-05-09
[WFLY-16356] Get web testsuite running transformed for Jakarta EE 9+

darranl:
https://issues.redhat.com/browse/WFLY-16356


testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/formauth/FormAuthUnitTestCase.java
```diff
@@ -108,6 +98,8 @@ public static WebArchive deployment() {
         war.addAsWebResource(tccl.getResource(resourcesLocation + "restricted/error.html"), "restricted/error.html");
         war.addAsWebResource(tccl.getResource(resourcesLocation + "restricted/login.html"), "restricted/login.html");
 
+        war.as(ZipExporter.class).exportTo(new File("/home/darranl/tmp/Test.war"), true);
```
bstansberry, line 101: I was investigating why this test was failing on lots of CI jobs and it never occurred to me that the path in the failure is clearly wrong for CI. :)

bstansberry (changes requested):

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/formauth/FormAuthUnitTestCase.java
darranl: Oops

darranl:
Sorry about that one, was debugging a bizarre test error - CI running again.

testsuite/preview/source-transform/web/src/test/config/arq/arquillian-bootable.xml
```diff
@@ -0,0 +1,26 @@
...
+    xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
+
+    <defaultProtocol type="jmx-as7" />
+
+    <container qualifier="jboss" default="true">
+        <configuration>
+            <property name="installDir">${install.dir}</property>
+            <property name="jarFile">${bootable.jar}</property>
+
+            <property name="javaVmArguments">${server.jvm.args}</property>
```
bstansberry, line 12: The 'upstream' variant of this includes ' -Djava.security.policy==${basedir}/target/test-classes/security.policy'. Is that cruft there? And the security.policy file it points to as well?

testsuite/preview/source-transform/web/src/test/config/arq/arquillian.xml
```diff
@@ -0,0 +1,25 @@
...
+            <property name="serverConfig">${jboss.server.config.file.name:standalone.xml}</property>
+            <property name="jbossArguments">${jboss.args}</property>
+            <!-- -Djboss.inst is not necessarily needed, only in case the test case neeeds path to the instance it runs in.
+                 In the future, Arquillian should capable of injecting it into @ArquillianResource File or such. -->
+            <property name="allowConnectingToRunningServer">true</property>
+            <property name="managementAddress">${node0:127.0.0.1}</property>
+            <property name="managementPort">${as.managementPort:9990}</property>
+
+            <!-- AS7-4070 -->
+            <property name="waitForPorts">${as.debug.port:8787} ${as.managementPort:9990}</property>
```
bstansberry, line 20: The 'upstream' file didn't have this line. Adding it is fine but looking at this got me wondering whether we are waiting 8 seconds with every server launch for a debug socket that typically isn't open. And if so can we find a smarter way?

bstansberry (commented):
@darranl This LGTM. I have a couple 'review comments' but they are really tangent points for possible later work.

I'll let this sit for an hour or so to give you a chance to notice those; then I'll merge. And you should feel free to merge.

darranl:
@bstansberry I just copied "a set of arquillian" configurations instead of the "the web" configurations - I will copy in the web ones to be safe.

bstansberry (approved):

PR #15545, opened 2022-05-10
[WFLY-16369] Remove unused commons-cli version property

fjuma:
https://issues.redhat.com/browse/WFLY-16369


bstansberry (approved):

darranl (changes requested):
I have just added a PR for Jakarta EE which uses the commons-lang3 property so it us just commons-cli unused now.

The reason we need it back is tests have been getting away with using an ApacheDS shaded jar when they are not even using ApacheDS.

bstansberry:
Thanks, @fjuma

PR #15552, opened 2022-05-12
[WFLY-16374] Default configuration for domain uses BASIC as a http-authentication-factory mechanism

ivassile:
Issue: https://issues.redhat.com/browse/WFLY-16374


bstansberry (changes requested):
@ivassile The files in the legacy-configs dir are meant to match as much as possible the file we shipped in the named release, so we don't want to 'correct' them. We use them to check we can still process the old configs.

bstansberry (approved):

PR #15556, opened 2022-05-13
Wfly 15978

tadamski:
https://issues.redhat.com/browse/WFLY-15978

This relies on: https://github.com/ironjacamar/ironjacamar/pull/754


tadamski:
hi @darranl could you please take a look whether such refactor makes sense? The goal is to be able to migrate EARs with IronJacamar configuration files to Elytron.

connector/src/main/java/org/jboss/as/connector/metadata/api/resourceadapter/ActivationSecurityUtil.java
```diff
@@ -0,0 +1,93 @@
...
+            return false;
+        // security uses elytron
+        if (security.isElytronEnabled())
+            return false;
+        // check if legacy domain is non-null
+        final String domain = security.resolveSecurityDomain();
+        return domain != null && domain.trim().length() > 0;
+    }
+
+    private static boolean isLegacySecurityRequired(org.jboss.jca.common.api.metadata.resourceadapter.WorkManagerSecurity config) {
```
darranl, line 80: Can this type be in the imports above?

darranl: Ignore that, I see a class with the same name is in this package.

darranl:
@tadamski sorry I missed the original request to review - are these changes still needed as we already have https://github.com/wildfly/wildfly/pull/15157 merged?

tadamski:
@darranl I think we can close it; I still need to do some cleaning here but I'm working on it in another branches currently

PR #15561, opened 2022-05-16
[WFLY-15431] Add elytron-oidc-client to the EE feature pack

fjuma:
https://issues.redhat.com/browse/WFLY-15431
https://issues.redhat.com/browse/WFLY-16435


yersan (commented):
I'm not asking for changes, just to highlight to other reviewers there is a difference in how we compose the full galleon pack vs ee galleon pack

In the case of the full galleon pack, it inherits the ee galleon pack, so I would expect to remove from full the copy of the OIDC resources since it could be redundant. It could be harmless, but it looks like it is the correct approach.

Another point to look at is whether it could be necessary to change the property used for the Group and version attributes on wildfly-elytron-oidc-* artifacts from `${full.maven.groupId}` to `${ee.maven.groupId}`

fjuma:
@yersan Those are good points, I can make those changes.

fjuma:
@yersan I've made the updates that were mentioned.

yersan (changes requested):
@fjuma would you mind also correcting the HostExludeTestCase removing this:

https://github.com/wildfly/wildfly/blob/cae473fc7b8ceaa3abf650809f37b142deb8747a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java#L238

This test should fail when we are running the test suite by using the ee-galleon-pack. The point is since now we are going to have this module available from the ee distribution, we also need to have the host-exclusions for it. That test verifies them.

You can run this Job with your topic branch to verify that test after you prepare it for the ee-galleon-pack distribution :

https://ci.wildfly.org/viewType.html?buildTypeId=WF_MasterLinuxAltDistJdk8

You could use this maven goal to verify the test on your local environment:

`mvn clean test -Dtest=HostExcludesTestCase -Dtestsuite.default.build.project.prefix=ee-`

fjuma:
Thanks @yersan, I've removed that line and also started a custom job [here](https://ci.wildfly.org/buildConfiguration/WF_MasterLinuxAltDistJdk8/306784?buildTab=overview&showRootCauses=false).

yersan:
FYI @fjuma , I started a new one since we have to use JDK11:

https://ci.wildfly.org/viewLog.html?buildId=306947

yersan (changes requested):
Hi @fjuma , as the integration test revealed, you still need to add some changes to correct the missing host exclusions after moving the elytron-oidc-client

Test failures revealed that also `org.wildfly.extension.opentelemetry` has the same issue; lack of exclusion.

You should add these exclusions here: https://github.com/wildfly/wildfly/blob/779e809e490ed1549fc945fa9e1d63da36272a0a/ee-feature-pack/galleon-common/src/main/resources/feature_groups/domain-host-excludes.xml#L8 It's an ugly file and we should add them alphabetically for each WildFly exclusion. And I think until WildFly24.0 because those extensions were added on WildFly 25:https://github.com/wildfly/wildfly/blob/aa4d583ccfde697f5959c649ac85112639079989/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java#L191-L193

yersan:
https://ci.wildfly.org/viewLog.html?buildId=307822

yersan (approved):
Thanks, @fjuma , CI job confirms it was fixed, LGTM

darranl (approved):

PR #15563, opened 2022-05-18
WFLY-16383 mod_cluster: Do not register contexts when in suspend mode

TomasHofman:
This change prevents mod_cluster from registering web contexts when the
container is started in suspended mode. The result is, that when the
contexts remain unregistered, reverse proxy will keep serving 404
responses (rather than 503) on those contexts.

There has been previous attempts to achieve this behavior (WFLY-14121,
WFLY-13074).

https://issues.redhat.com/browse/WFLY-16383

### DIFF
--- a/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
+++ b/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
@@ -118,14 +118,17 @@ public class UndertowEventHandlerAdapterService implements UndertowEventListener
     }
 
     private synchronized void onStart(Context context) {
-        ContainerEventHandler handler = this.configuration.getContainerEventHandler();
-
-        handler.add(context);
-
         State state = this.configuration.getSuspendController().getState();
 
-        // TODO break into onDeploymentAdd once implemented in Undertow
-        if(state == State.RUNNING) {
+        if (state == State.RUNNING) {
+            // Following handler calls result in context registration on the proxy side, which in turn results in
+            // the proxy starting to serve 503 responses for this context. We want to avoid context registration unless
+            // the server is in RUNNING state.
+            // During normal server startup, this block is skipped and the add() & start() handler calls are done later
+            // in the `#resume()` method.
+            ContainerEventHandler handler = this.configuration.getContainerEventHandler();
+            handler.add(context);
+            // TODO break into onDeploymentAdd once implemented in Undertow
             handler.start(context);
         }
 
@@ -202,8 +205,16 @@ public class UndertowEventHandlerAdapterService implements UndertowEventListener
 
     @Override
     public void resume() {
+        // This is run just before container enters RUNNING state - contexts need to be registered.
+        ContainerEventHandler handler = this.configuration.getContainerEventHandler();
         for (Context context : this.contexts) {
-            this.configuration.getContainerEventHandler().start(context);
+            handler.add(context);
+            handler.start(context);
+        }
+        // Proxy will not consider this node to be ready, until it receives a STATUS request - send it now to minimize
+        // time when the proxy serves 503.
+        for (Engine engine : this.server.getEngines()) {
+            this.configuration.getContainerEventHandler().status(engine);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


TomasHofman:
@rhusar @pferraro / @soul2zimate @spyrkob would you like to review this?

I found these previous PRs that are related.

https://github.com/jbossas/jboss-eap7/pull/3411
https://github.com/jbossas/jboss-eap7/pull/4159

From what I understand from reading the customer case and the previous tickets, the desired behavior is such:

* When no nodes serving given web context are present, proxy should respond with 404.
* When all present nodes serving given context are suspended, *proxy should still respond with 404*, meaning that the web context should not be registered on the proxy. Once it is registered, proxy is going respond with 503, until a node announces itself as active. This is what the previous PRs were trying to fix, and it works only partially now (you still get 503 sometimes, even though the container is suspended).

The reason why it only works partially, is that the `handler.add(context)` call is called always in the `UndertowEventHandlerAdapterService#onStart()`:

```
 private synchronized void onStart(Context context) {
        ContainerEventHandler handler = this.configuration.getContainerEventHandler();

        handler.add(context);

        State state = this.configuration.getSuspendController().getState();
        if(state == State.RUNNING) {
            handler.start(context);
        }
}
```

 The `handler.add(context)` call results in DISABLE-APP & STOP-APP requests being sent to a proxy, and these requests cause context registration. Whether these MCMP requests are in fact sent, depends on whether proxy connections are established at the moment or not, so it's a timing issue.

My suggestion is to move the `handler.add(context)` call into the `if(state == State.RUNNING)` condition block, where, in effect, it would be called only when an app is deployed on already active server. For the normal startup/reload situation, the `handler.add(context)` call would be added to the `resume()` method (this method is called just before the container enters RUNNING state, i.e. not in suspend mode):

```
    public void resume() {
        ContainerEventHandler handler = this.configuration.getContainerEventHandler();
        for (Context context : this.contexts) {
            handler.add(context);
            handler.start(context);
        }
        ...
    }
```

TomasHofman:
@pferraro I would say you are objecting to the other point then you quoted. Let me rephrase.

Current behavior:

1. No node serves given deployment => context is not registered on proxy => proxy returns 404;
2. At least one node serves given deployment, but is not ready to respond to requests (startup not finished, in error, suspended before shutdown, ...) => context is registered, but node is considered in ERROR state => proxy returns 503;
3. At least one node serves given deployment and is ready to respond => context is registered, node is considered READY => proxy serves the app.

This to me is in line with what you say about deployment-awareness, unavailable URL is distinguishable from unknown URL, and this would stay.

What this PR changes, is that when a node is started in suspend mode, i.e. we know it will never be able to respond to requests until it is restarted, then it wouldn't register its deployments to the proxy.

pferraro:
A server that was started, and then suspended should be in the same LB state as a server started in suspended mode.  A suspended server is meant to be distinguishable from a worker that is not started.  A suspended server should quickly made available by an ENABLE-APP command.  It seems to me that this proposed change would break that.

pferraro:
@TomasHofman We *must* establish the state of the server with the LB on startup, regardless of whether the server is suspended or not, since it may have previous crashed, in which case we must make sure to disable any contexts.

rhusar (changes requested):
👎  for the changes.

1. The proxy state should be the same regardless whether the server was started in suspended mode or got into the suspend mode eventually.
2. The `404` is a state when the resource does not exist however in the suspended situation the context does exist but not receiving requests in which case the `503` is the correct response. Should a different handling be required, the onus is on the proxy configuration how it handles the situation.
3. The purpose of the suspended state is to resume operation as quickly as possible while the proposed changes would require an additional add operation for the proxy.

rhusar:
@TomasHofman A different resulting behaviour depending on proxy discovery configuration is definitely a bug (due to handling of container events). Please open an issue for that.

BTW regarding the previous issues you linked, if I am remembering correctly, the WFLY-13074 dealt with the initial suspend mode support (when a simple call to suspend was just unhandled by the mod_cluster subsystem) and WFLY-14121 dealt with a subsequent fix around starting or restarting suspended.

TomasHofman:
Follow up ticket: https://issues.redhat.com/browse/WFLY-16416

TomasHofman:
Closing this PR, incorrect behavior.

PR #15570, opened 2022-05-19
Fix for WFLY-16398, Add new tests to cover Galleon update

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-16398

The test is only active on unix. In order to enable the test you must add `-Dts.galleon`

The test can be run alone in testsuite/galleon/update, `mvn clean install`

The base version is set to 26.1.1.Final. Use  `-Dwildfly.test.galleon.update.base.version=<a wildfly version>` to override it.

A custom run of the nightly job "Galleon - Linux - JDK11" has been run successfully: https://ci.wildfly.org/buildConfiguration/WF_GalleonLinuxJdk11/305949
The test job has updated 26.1.0.Final to the latest 27.0.0.Alpha2-SNAPSHOT found in the local cache.

The official nightly could be updated to enable this test.

The test steps are:
* Download/unzip galleon CLI tool
* Setup the galleon CLI local maven cache
* Install the base server then update to the latest SNAPSHOT a complete installation and a trimmed one for wildfly-ee, wildfly and wildfly-preview.


testsuite/galleon/update/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+                            <execution>
+                                <id>enforce-property</id>
+                                <goals>
+                                    <goal>enforce</goal>
+                                </goals>
+                                <configuration>
+                                    <rules>
+                                        <requireProperty>
+                                            <property>wildfly.test.galleon.update.base.version</property>
+                                            <message>You must set wildfly.test.galleon.update.base.version property to a WildFly version that will be used to install prior to update to the latest SNAPSHOT!</message>
```
bstansberry, line 44: Setting this via a maven property in the root pom seems better. That can be overridden from the command line, but it provides a single place to update this when the 'default' value changes, instead of having to edit various CI jobs.

jfdenise: @bstansberry , thank-you. I added a default value, 26.1.1.Final for now.

bstansberry (approved):

PR #15575, opened 2022-05-24
[WFLY-16415] Organize dependency management for simplified use of different dependency sets

bstansberry:
Stop using the root pom for dependency management.

https://issues.redhat.com/browse/WFLY-16415


jamezp (approved):
This looks good to me. I like the clean up and adding of required dependencies too that were previously brought in transitively.

PR #15582, opened 2022-05-26
[WFLY-16412] Failed deployment leaves classloader leaks through undertow references

aogburn:
https://issues.redhat.com/browse/WFLY-16412

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java
@@ -63,6 +63,7 @@ import io.undertow.servlet.handlers.ServletPathMatches;
 import io.undertow.servlet.util.ImmediateInstanceFactory;
 import io.undertow.websockets.extensions.PerMessageDeflateHandshake;
 import io.undertow.websockets.jsr.ServerWebSocketContainer;
+import io.undertow.websockets.jsr.UndertowContainerProvider;
 import io.undertow.websockets.jsr.WebSocketDeploymentInfo;
 
 import org.apache.jasper.deploy.JspPropertyGroup;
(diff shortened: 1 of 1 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java
```diff
@@ -427,6 +427,12 @@ public synchronized void start(final StartContext startContext) throws StartExce
 
     @Override
     public synchronized void stop(final StopContext stopContext) {
+        // Remove the server activity
+        final ServerActivity activity = serverActivity.get();
+        if(activity != null) {
+            suspendController.get().unRegisterActivity(activity);
+        }
+        UndertowContainerProvider.removeContainer(module.getClassLoader());
```
aogburn, line 435: I think we'd want to actually call io.undertow.websockets.jsr.SecurityActions.removeContainer instead but this is protected:
https://github.com/undertow-io/undertow/blob/master/websockets-jsr/src/main/java/io/undertow/websockets/jsr/SecurityActions.java#L40

So do we change that from protected?  Or add the security manager handling to UndertowDeploymentInfoService?

bstansberry: It shouldn't matter. An MSC thread calling stop shouldn't have anything but container classes on the call stack.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java
```diff
@@ -949,19 +954,6 @@ public void resume() {
...
-
-                    @Override
-                    public void contextDestroyed(ServletContextEvent sce) {
-                        final ServerActivity activity = serverActivity.get();
-                        if(activity != null) {
-                            suspendController.get().unRegisterActivity(activity);
-                        }
-                    }
-                };
-                d.addListener(new ListenerInfo(sl.getClass(), new ImmediateInstanceFactory<EventListener>(sl)));
```
bstansberry, line 964: I don't know enough about this code to know if removing this is ok, but if there are concerns from those who do I think you can achieve your objective by leaving this here and replacing serverActivity.get() both here and in the stop method with serverActivity.getAndSet(null).

bmaxwell: @aogburn / @fl4via  can you check this one, it looks like the fix has been stalled for several months

fl4via: yes, it is okay to remove this block.

fl4via (changes requested):
it is just pending the security manager handling

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentInfoService.java
aogburn: Thanks @fl4via, added it to UndertowDeploymentInfoService

fl4via (approved):

PR #15581, opened 2022-05-26
[WFLY-15434][WFLY-15458][WFLY-15460][WFLY-14709]:Move WildFly Preview…

jimma:
… to a native jakarta namespace variant of CXF, WSS4j,XmlSec,Opensaml and Glassfish JAXB3

https://issues.redhat.com/browse/WFLY-14709
https://issues.redhat.com/browse/WFLY-15434
https://issues.redhat.com/browse/WFLY-15458
https://issues.redhat.com/browse/WFLY-15460

PR #15587, opened 2022-05-27
[WFLY-16367] Add a transformed smoke test suite.

darranl:
https://issues.redhat.com/browse/WFLY-16367

This will really need rebasing on top of https://github.com/wildfly/wildfly/pull/15581 - I have just stolen some of Jim's dependencies to get the client side working here.

Also I haven't spent time cleaning up the other executions in case we want to use them.

The dependency list is hopefully clean starting with an empty list and adding dependencies as identified as needed some some old dependencies built up in smoke may have been removed.


bstansberry:
@darranl #15581 is merged.

darranl:
@bstansberry I think the other CI jobs are running, I have rebased this one - let me know if I have accidentally pulled anything back in I should not have.

FYI I have also been able to drop the system dependency.

testsuite/preview/pom.xml
```diff
@@ -50,12 +50,274 @@
...
+            </dependency>
+            
+            <!-- Temporary CXF References -->
+            <dependency>
+                <groupId>org.jboss.ws.cxf</groupId>
+                <artifactId>jbossws-cxf-client-ee9</artifactId>
+                <version>${preview.version.org.jboss.ws.cxf}</version>
+                <scope>system</scope>
+                <systemPath>${wildfly.transformed.repo.dir}/org/jboss/ws/cxf/jbossws-cxf-client/${preview.version.org.jboss.ws.cxf}/jbossws-cxf-client-${preview.version.org.jboss.ws.cxf}.jar</systemPath>
+            </dependency>
```
bstansberry: @darranl I think L158-165 is the only thing that should be added to dependencyManagement in this pom. L52  already results in importing the wildfly-ee9-parent bom, and then all the other entries are duplicates because that bom declares them.

bstansberry (changes requested):

darranl:
@bstansberry FYI I have needed to retain the CXF definitions in testsuite/preview/pom.xml without this I was seeing tests still running without the version override in place so I was getting a mix of Jakarta EE 8 and 9.

testsuite/preview/pom.xml
```diff
@@ -61,6 +61,103 @@
...
+                    <exclusion>
+                        <groupId>javax.xml.bind</groupId>
+                        <artifactId>jaxb-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.jvnet.staxex</groupId>
+                        <artifactId>stax-ex</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
```
bstansberry, line 160: @darranl If you move the dependencyManagement entries for these out of the root pom and into ee-feature-pack/pom.xml that should eliminate the need to declare these here.

darranl:
@bstansberry I have included your PR in my topic branch now so have removed the dependency overrides.

bstansberry (approved):

bstansberry:
Thanks @darranl

PR #15589, opened 2022-05-27
[WFLY-14769] Access checking for txn namespace lookup

tadamski:
https://issues.redhat.com/browse/WFLY-14769

I have added a wrapper to WildFlyNamingContext which enables installation of aroundLookup interceptors. This functionality is used to perform access check for objects from "txn" namespace.

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
@@ -1,0 +1,89 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2023, Red Hat Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.ejb3.subsystem;
+
+import org.jboss.as.ejb3.component.allowedmethods.AllowedMethodsInformation;
+import org.jboss.as.ejb3.component.allowedmethods.MethodType;;
+
+import org.jboss.as.naming.InitialContext;
+import org.jboss.as.naming.LookupInterceptor;
+import org.jboss.msc.Service;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
+
+import javax.naming.InvalidNameException;
+import javax.naming.Name;
+import javax.naming.NamingException;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+public class TransactionNamespaceAccessControlService implements Service {
+    private static final String REMOTE_USER_TRANSACTION = "RemoteUserTransaction";
+    private static final String LOCAL_USER_TRANSACTION = "LocalUserTransaction";
+
+    public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("ejb3", "TransactionNamespaceAccessControlService");
+
+    private LookupInterceptor userTransactionLookupInterceptor = new LookupInterceptor() {
+
+        @Override
+        public void aroundLookup(Name name) throws NamingException {
+            if (requiresCheckAllowed(name)) {
+                try {
+                    AllowedMethodsInformation.checkAllowed(MethodType.GET_USER_TRANSACTION);
+                } catch (IllegalStateException e) {
+                    throw new NamingException(e.getMessage());
+                }
+            }
+        }
+
+        private boolean requiresCheckAllowed(Name origName) throws InvalidNameException {
+            if (origName.isEmpty() || !origName.get(0).startsWith("txn:")) {
+                return false;
+            }
+            final Name reparsedName = (Name) origName.clone();
+            final String first = reparsedName.get(0);
+            final int idx = first.indexOf(':');
+            final String segment = first.substring(idx + 1);
+            reparsedName.remove(0);
+            if (segment.length() > 0 || (origName.size() > 1 && origName.get(1).length() > 0)) {
+                reparsedName.add(0, segment);
+            }
+            if (reparsedName.toString().equals(REMOTE_USER_TRANSACTION) || reparsedName.toString().equals(LOCAL_USER_TRANSACTION)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+    };
+
+    @Override
+    public void start(StartContext context) {
+        InitialContext.getInitialContextFactory().addInterceptor(userTransactionLookupInterceptor);
+    }
+
+    @Override
+    public void stop(StopContext context) {
+        InitialContext.getInitialContextFactory().removeInterceptor(userTransactionLookupInterceptor);
+    }
+}
--- a/naming/src/main/java/org/jboss/as/naming/LookupInterceptor.java
+++ b/naming/src/main/java/org/jboss/as/naming/LookupInterceptor.java
@@ -1,0 +1,43 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2023, Red Hat Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.as.naming;
+
+import javax.naming.Name;
+import javax.naming.NamingException;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+
+/**
+ * A naming interceptor applied before the name is being resolved by the InitialContext
+ */
+public interface LookupInterceptor {
+
+    /**
+     * Handles the naming invocation.
+     *
+     * @param name the name being resolved
+     * @throws NamingException if an invocation error occurs
+     */
+    public void aroundLookup(Name name) throws NamingException;
+}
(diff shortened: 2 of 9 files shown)


tadamski:
@emmartins Could you please review whether this fix makes sense from naming perspective?

bmaxwell:
@emmartins can you review this PR? It has missed making it into a couple of CPs

moulalis:
@tadamski CI failed with this PR, could you please check?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
```diff
@@ -0,0 +1,126 @@
...
+import org.jboss.msc.value.InjectedValue;
+
+import javax.naming.InvalidNameException;
+import javax.naming.Name;
+import javax.naming.NamingException;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+public class TransactionNamespaceAccessControlService implements Service<Void> {
```
darranl, line 42: Does this need to be public?  It looks like it is accessed from the same package only.

naming/src/main/java/org/jboss/as/naming/WildFlyRootContextWrapper.java
```diff
@@ -0,0 +1,346 @@
...
+import javax.naming.directory.SearchResult;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+
+public class WildFlyRootContextWrapper implements DirContext {
```
darranl, line 48: This is another one that looks like it is only used from the same package so could have visibility reduced?

emmartins:
@tadamski does this logic runs for every JNDI lookup?

darranl:
This appears to be implemented by intercepting the client side of the invocation not the server side.  

Is it possible for a deployment to create an InitialContext that bypasses this check?
If so does it matter?  i.e. Is this just a helper or do we need to enforce this strictly?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
tadamski: fixed

tadamski:
> @tadamski does this logic runs for every JNDI lookup?

Yes, every lookup from within a server is wrapped so I was able to plug verification interceptors.



> This appears to be implemented by intercepting the client side of the invocation not the server side.
> 
> Is it possible for a deployment to create an InitialContext that bypasses this check? If so does it matter? i.e. Is this just a helper or do we need to enforce this strictly?

I'm not certain that hack that allows that doesn't exist, but the mechanism of context factory creation is the same as it was before (without a wrapper) so imho no new vulnerability should be introduced.



> @tadamski some of the test failures in this appear to be legitimate (org.jboss.as.test.integration.naming.connector.JMXConnectorTestCase, org.jboss.as.test.integration.naming.rmi.RmiContextLookupTestCase). They started to fail also in EAP testsuite after this was backported.

Fixed.

TomasHofman:
@emmartins @darranl can you conclude the review? It affects whether we can include this downstream.

TomasHofman:
@tadamski well, the test failure in the security manager job looks maybe legitimate, it also fails downstream. Could you check it up?

tadamski:
@TomasHofman fixed

soul2zimate:
All checks have passed. @emmartins @darranl can you review this again and merge?

soul2zimate:
Hi @emmartins  @darranl,  May I ask your review on this? This is required by downstream CP.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
```diff
@@ -0,0 +1,126 @@
...
+            ReparsedName reparsedName = reparse(name);
+            if (reparsedName.getUrlScheme().equals(TRANSACTION_NAMESPACE) &&
+                    (reparsedName.getName().toString().equals(REMOTE_USER_TRANSACTION) || reparsedName.getName().toString().equals(LOCAL_USER_TRANSACTION))) {
+                try {
+                    AllowedMethodsInformation.checkAllowed(MethodType.GET_USER_TRANSACTION);
+                } catch (IllegalStateException e) {
+                    throw new NamingException(e.getMessage());
+                }
+            }
+        }
```
bstansberry, line 65: This is executed on the request path so the algorithm needs to be highly optimized with minimal object allocations. ReparsedName exists to pass a bit of state from a private method so the calling method can use that state in if checks. Instead this could be 'if (requiresCheckAllowed(name)) {' with requiresCheckAllowed(Name name) doing an optimized check.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
```diff
@@ -0,0 +1,126 @@
...
+
+        private ReparsedName reparse(final Name origName) throws InvalidNameException {
+            final Name name = (Name) origName.clone();
```
bstansberry, line 68: As noted above, this can just return a boolean. I believe cloning can be deferred until we've confirmed origName.get(0).startsWith("txn:").

naming/src/main/java/org/jboss/as/naming/InServerInitialContextFactoryService.java
```diff
@@ -0,0 +1,81 @@
...
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+public final class InServerInitialContextFactoryService implements InitialContextFactory, Service<InitialContextFactory> {
```
bstansberry, line 41: This needs to implement the org.jboss.msc.Service interface, not the deprecated org.jboss.msc.service.Service interface.

naming/src/main/java/org/jboss/as/naming/InServerInitialContextFactoryService.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public InitialContextFactory getValue() throws IllegalStateException, IllegalArgumentException {
+        return this;
+    }
+
+    @Override
+    public void start(StartContext startContext) throws StartException {
+        InitialContext.setInitialContextFactory(this);
```
bstansberry, line 62: This has the effect that InitialContext has 1 delegate for a period of time and then this switches it when MSC gets around to starting this service. That's fragile.

Couldn't InitialContext expose a (package-protected) static getter instead of a setter? So it creates an InitialContextFactory like this and exposes it via the static getter. This service just makes that object available as an MSC service, for injection into TransactionNamespaceAccessControlService.

The value type for this service would be a new interface "InServerInitialContextFactory" that has the add/removeInterceptor methods.

naming/src/main/java/org/jboss/as/naming/LookupInterceptor.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.naming.NamingException;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+
+public interface LookupInterceptor {
+
+    public void aroundLookup(Name name) throws NamingException;
+}
```
bstansberry, line 34: The class and method should have javadoc as this is an API used outside this subsystem.

bstansberry (changes requested):

naming/src/main/java/org/jboss/as/naming/InServerInitialContextFactoryService.java
```diff
@@ -0,0 +1,81 @@
...
+import org.jboss.msc.service.StopContext;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+public final class InServerInitialContextFactoryService implements InitialContextFactory, Service<InitialContextFactory> {
+
+    public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("InServerInitialContextFactoryService");
+
+    private List<LookupInterceptor> interceptors = new ArrayList<>();
```
bstansberry, line 45: This needs to be a concurrent collection.

bstansberry (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
```diff
@@ -0,0 +1,126 @@
...
+import org.jboss.msc.value.InjectedValue;
+
+import javax.naming.InvalidNameException;
+import javax.naming.Name;
+import javax.naming.NamingException;
+
+/**
+ * @author <a href="mailto:tadamski@redhat.com">Tomasz Adamski</a>
+ */
+class TransactionNamespaceAccessControlService implements Service<Void> {
```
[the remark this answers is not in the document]
tadamski: fixed

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
tadamski: fixed

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
tadamski: fixed

naming/src/main/java/org/jboss/as/naming/LookupInterceptor.java
tadamski: fixed

naming/src/main/java/org/jboss/as/naming/InServerInitialContextFactoryService.java
tadamski: fixed

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/TransactionNamespaceAccessControlService.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    };
+
+    @Override
+    public void start(StartContext context) throws StartException {
+        try {
+            InitialContext.getInitialContextFactory().addInterceptor(userTransactionLookupInterceptor);
+        } catch (Throwable t) {
+            System.out.println("BAJOBONGO");
+            t.printStackTrace();
```
pferraro, line 87:
```java
throw new StartException(t);

**Review by pferraro — CHANGES_REQUESTED** [review]

File: naming/src/main/java/org/jboss/as/naming/WildFlyRootContextWrapper.java [hunk 11]
```diff
@@ -0,0 +1,346 @@
...
+
+    @Override
+    public String getNameInNamespace() throws NamingException {
+        return delegate.getNameInNamespace();
+    }
+
+    private final void aroundLookup(Name name) throws NamingException {
+        for (LookupInterceptor interceptor : interceptors) {
+            interceptor.aroundLookup(name);
+        }
```
pferraro, line 344: This is prone to ConcurrentModificationExceptions.
Since the list implementation is a synchronized collection wrapper, iteration is only thread-safe within a synchronized block.
If updates are sufficiently infrequent (and if I understand this PR correct, this seems to be the case), I would instead opt for a COW-based concurrent data structure, e.g. CopyOnWriteArrayList.

pferraro (changes requested):

naming/src/main/java/org/jboss/as/naming/WildFlyRootContextWrapper.java
tadamski: Agree, this is a bug. The list itself is mean to be modified during the server startup in most (if not all scenarios), so CopyOnWriteArrayList seems like a good solution. I have updated the PR.

parsharma:
@darranl @emmartins could you please review this again?

Thanks

parsharma:
Hi @emmartins @darranl, May I ask for your review on this? Downstream CP requires this.
Thanks

gaol:
@bstansberry @darranl The PR is waiting for approve and merge for over 3 months. It is quite complete fix, would you please approve and merge if no more comments ? There is a downstream issue depending on this and got postponed for several CPs, thank you very much !

pferraro (changes requested):
This approach seems very hacky to me.
Have you considered decorating the org.wildfly.naming.client.NamingContextFactory implementation from wildfly-transaction-client?
That way you can add your access check to before delegating to the existing implementation, which is already specific to the "txn:" namespace, without needing to interfere with the handling of other namespaces, and without touching the naming subsystem.

tadamski:
> This approach seems very hacky to me. Have you considered decorating the org.wildfly.naming.client.NamingContextFactory implementation from wildfly-transaction-client? That way you can add your access check to before delegating to the existing implementation, which is already specific to the "txn:" namespace, without needing to interfere with the handling of other namespaces, and without touching the naming subsystem.

@pferraro Agree, the solution that you proposed is much better, I have already checked and it works. I'm closing this PR and will open alternative one.

PR #15599, opened 2022-05-31
WFLY-16379 Support expanded CDI batch artifact loading

chengfang:
https://issues.redhat.com/browse/WFLY-16379

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
@@ -22,6 +22,8 @@
 
 package org.wildfly.extension.batch.jberet.deployment;
 
+import java.lang.reflect.Method;
+import java.security.PrivilegedExceptionAction;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -44,6 +46,7 @@ import org.jboss.weld.context.RequestContext;
 import org.jboss.weld.context.unbound.UnboundLiteral;
 import org.jboss.weld.manager.BeanManagerImpl;
 import org.wildfly.extension.batch.jberet._private.BatchLogger;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * ArtifactFactory for Jakarta EE runtime environment.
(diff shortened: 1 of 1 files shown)


batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
```diff
@@ -141,19 +144,34 @@ private BeanManagerImpl getBeanManager() {
...
+            BatchLogger.LOGGER.tracef("Found bean: %s", result);
+            return (Bean<?>) result;
+        } catch (Exception e) {
+            BatchLogger.LOGGER.tracef("Looking up bean reference for '%s'", ref);
+            final Set<Bean<?>> beans = beanManager.getBeans(ref);
+            final Bean<?> bean = beanManager.resolve(beans);
+            if (bean != null) {
+                BatchLogger.LOGGER.tracef("Found bean '%s' for reference '%s'", bean, ref);
+            } else {
+                BatchLogger.LOGGER.tracef("No bean found for reference '%s;'", ref);
```
boris-unckel, line 171: NIT: Is the semicolon intended: '%s;' ?

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
```diff
@@ -141,19 +144,34 @@ private BeanManagerImpl getBeanManager() {
         return beanManager == null ? null : BeanManagerProxy.unwrap(beanManager);
     }
 
-    private static Bean<?> getBean(final String ref, final BeanManager beanManager) {
+    private static Bean<?> getBean(final String ref, final BeanManager beanManager, final ClassLoader classLoader) {
         if (beanManager == null) {
```
boris-unckel, line 148: You might consider to add a `|| classLoader == null` because https://github.com/jberet/jsr352/blob/2.1.0.Final/jberet-core/src/main/java/org/jberet/creation/AbstractArtifactFactory.java#L100 would cause a NPE.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
chengfang: That was there before my change, but I'll take a note and clean them up next time.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/ArtifactFactoryService.java
chengfang: This method is made a bit complicated by implementing this workaround. Basically what it does is first try to get a CDI bean with a `ref` string name. If no result, then tries to look up a CDI bean with a type that is loaded from the `classLoader`. So if the `classLoader` is null, which will never happen, this method should still perform the first part, which is get the bean by the `ref` name. So no need to check null here.

bstansberry (approved):
This looks ok, and I want to see the impact on TCK.

PR #15610, opened 2022-06-04
[WFLY-16439] Remove testsuite use of javax.faces.bean annotations

bstansberry:
https://issues.redhat.com/browse/WFLY-16439


jasondlee (approved):
Looks good to me!

PR #15613, opened 2022-06-06
WFLY-16436 [primary/secondary] Problematic Language usage deprecation and replacement in *.xml files/*

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-16436

 Step 1:
-  Renamed the xml files from host-[master/slave].xml to host-[primary/secondary].xml
-  This needs core-upgrade

Related Issue: https://issues.redhat.com/browse/WFCORE-5897
PR: https://github.com/wildfly/wildfly-core/pull/5108


yersan (changes requested):
Great work @parsharma , thanks. Could you also include here the required changes to replace `jboss.test.host.slave.jvmhome` by `jboss.test.host.secondary.jvmhome` as part of the properties replacement commit?

testsuite/mixed-domain/src/test/java/org/jboss/as/test/integration/domain/mixed/WildcardReadsTestCase.java
```diff
@@ -350,7 +350,7 @@ public void testWildcardHostRootQuery() {
         assertNotNull(resp.toString(), slaveResult);
 
         // Now limit the result to secondary hosts
-        op.get(WHERE, MASTER).set(false);
```
yersan, line 353: Hi @parsharma , this change is not expected on this commit since it is not related to the property renaming, the cause of this change is the renaming of the default hostnames.

In any case, this is the source of the errors you are still getting on the integration Jobs. It's a new case you need to handle differently on the test case.

This operation is querying the mixed domain host resources by filtering the results using "primary = false" where condition. As part of problematic language usage work, we have already renamed the attribute "master" on the "host" resource to "primary", see https://issues.redhat.com/browse/WFCORE-5844. That means this operation in this scenario no longer works as the test case expects because the legacy host, which is part of the mixed domain, uses the attribute "master". The host attribute "primary" doesn't exist on the legacy host.

You have to adapt this test case to handle property this situation. The simplest solution is to execute two operations, one using "primary = false" where condition, and the other using "master = false" where condition.
You should assert that "primary = false" returns empty, and "master = false" returns 1.

A more elegant solution would be to create a composite operation and assert the same.

This where condition is used in other parts of the test, so maybe you can get a common code extracted so it can be invoked from the other tests done on this class.

yersan (changes requested):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -388,7 +388,7 @@ protected void registerAddOperation(ManagementResourceRegistration registration,
         // Sort the add params alphabetically to match what DefaultResourceAddDescriptionProvider does
         SortedSet<AttributeDefinition> sorted = new TreeSet<>(Comparator.comparing(AttributeDefinition::getName));
         for (AttributeDefinition ad : add_attributes) {
-            if (!ad.getImmutableFlags().contains(AttributeAccess.Flag.STORAGE_RUNTIME)) {
+            if (!ad.getFlags().contains(AttributeAccess.Flag.STORAGE_RUNTIME)) {
```
yersan, line 391: This change should disappear after rebasing on top of main branch (WFLY-16825)

yersan (approved):
LGTM, only waiting for Windows Integration Job results. Link in the WildFly Core PR counterpart.

@parsharma, would you mind adding the Jira number to all the commit messages? thanks

parsharma:
@yersan added the JIRA number. Thanks

yersan:
@parsharma You have not pushed up the Jira numbers in the commit message, anyway since this PR is going to be replaced with the one that bumps wildfly core in wildfly full, I'll do that when picking up your commits.

PR #15623, opened 2022-06-08
[DRAFT] standalone-full.xml and standalone-full-ha.xml expressed with layers

jfdenise:
This PR brings commits from https://github.com/wildfly/wildfly/pull/15609 that covers standalone.xml and standalone-ha.xml

The standalone-full.xml and standalone-full-ha.xml are expressed with layers in the 3 feature-packs.

Issues:

We are missing layers:

* jsr77 subsystem
* embedded broker: Tracked by [WFLY-13798](https://issues.redhat.com/browse/WFLY-13798)
* embedded-broker ha (equivalent to feature-group `messaging-activemq-ha`)

Should we add layer for the missing ones?


jfdenise:
@bstansberry , this is the first step of layerization of standalone-full and standalone-full-ha to identify missing pieces.

bstansberry:
@jfdenise Thanks. The jsr77 subsystem is going to be removed from our standard configs and will only be usable in domain.xml to allow management of legacy version host controllers. No need for a layer for that use case. See #15619.

jfdenise:
Closing this one as https://github.com/wildfly/wildfly/pull/15609 contains full/full-ha default configs.

PR #15633, opened 2022-06-09
[WFLY-16415] Exclude community versions of transitive dependencies of…

dbkreling:
… components

Community versions of some transitive dependencies are not being properly resolved to their
productized versions, which cause the productized compilation of the server to fail due to multiple
versions of the same GAVs. The community versions of those dependencies need to be excluded.

### DIFF
--- a/connector/pom.xml
+++ b/connector/pom.xml
@@ -93,10 +93,38 @@
         <dependency>
             <groupId>org.jboss.ironjacamar</groupId>
             <artifactId>ironjacamar-core-api</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jboss.spec.javax.transaction</groupId>
+                    <artifactId>jboss-transaction-api_1.3_spec</artifactId>
+                </exclusion>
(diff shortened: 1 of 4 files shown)


bstansberry (changes requested):
@dbkreling These exclusions should be placed in the dependencyManagement entries in ee-feature-pack/pom.xml. Except for undertow-servlet, which should be in the WildFly Core root pom.

What you did here in undertow/pom.xml is fine though as a temporary workaround.

Bonus points for not going with multiple exclusion entries but instead excluding *:*. I want to get to the point where 99% of dependencyManagement entries exclude *:*. Transitive deps are a bad concept in a codebase as complex as WildFly. But it's ok to stick with multiple entries for this fix, particularly if moving to *:* causes build problems where other modules are missing explicit dependency declarations and rely on a transitive dep.

dbkreling:
This PR was reworked and the changes suggested were submitted via https://github.com/wildfly/wildfly/pull/15641.
Closing this PR

PR #15641, opened 2022-06-10
[WFLY-16415] Exclude community versions of transitive dependencies

dbkreling:
Community versions of some transitive dependencies are not being properly resolved to their
productized versions, which cause the productized compilation of the server to fail due to multiple
versions of the same GAVs. The community versions of those dependencies need to be excluded.


bstansberry (approved):

bstansberry:
Thanks @dbkreling

PR #15642, opened 2022-06-10
Upgrade Jakarta Connectors spec to 2.1 + TCK fix

tadamski:

If this PR is not for the 'main' branch you must add a link to the equivalent change in 'main'.

Remember to use the Jira issue ID in the PR title and any commits.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/jca/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/jca/Constants.java
@@ -50,7 +50,7 @@ public class Constants {
     static final String TX = "TX";
     static final String NON_TX = "NonTX";
 
-    static final Boolean ELYTRON_MANAGED_SECURITY = Boolean.FALSE;
+    static final Boolean ELYTRON_MANAGED_SECURITY = Boolean.TRUE;
 
     static final String ELYTRON_ENABLED_NAME = "elytron-enabled";
 
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
@@ -80,7 +80,7 @@ public class JcaExtension implements Extension {
 
     public static final String SUBSYSTEM_NAME = "jca";
 
-    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(5, 0, 0);
+    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(5, 1, 0);
 
     private static final String RESOURCE_NAME = JcaExtension.class.getPackage().getName() + ".LocalDescriptions";
 
@@ -113,6 +113,7 @@ public class JcaExtension implements Extension {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_3_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_4_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_5_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_6_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
     }
 
     static final class ConnectorSubsystemParser implements XMLStreamConstants, XMLElementReader<List<ModelNode>>,
@@ -325,6 +326,7 @@ public class JcaExtension implements Extension {
             while (reader.hasNext() && reader.nextTag() != END_ELEMENT) {
 
                 switch (Namespace.forUri(reader.getNamespaceURI())) {
+                    case JCA_6_0:
                     case JCA_5_0:
                     case JCA_4_0:
                     case JCA_3_0:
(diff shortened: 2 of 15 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/jca/Constants.java
```diff
@@ -50,7 +50,7 @@ public class Constants {
     static final String TX = "TX";
     static final String NON_TX = "NonTX";
 
-    static final Boolean ELYTRON_MANAGED_SECURITY = Boolean.FALSE;
+    static final Boolean ELYTRON_MANAGED_SECURITY = Boolean.TRUE;
```
bstansberry, line 53: This requires a management API and xsd version bump. (XSD because the current one documents the default behavior as false.)

It needs a transformer that specifically sets the undefined value to true, so the legacy servers run with the settings the domain config indicates.

The parsers for legacy xsd namespaces should also set 'false' if the value is not configured.

bstansberry (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
```diff
@@ -524,11 +528,16 @@ private void parseWorkManager(final XMLExtendedStreamReader reader, final ModelN
...
                         switch (readerNS) {
-                            case JCA_5_0: {
+                            case JCA_5_1: {
                                 String value = rawElementText(reader);
                                 JcaWorkManagerDefinition.WmParameters.ELYTRON_ENABLED.getAttribute().parseAndSetParameter(value, workManagerOperation, reader);
                                 break;
                             }
+                            case JCA_5_0: {
+                                String value = rawElementText(reader);
+                                JcaWorkManagerDefinition.WmParameters.ELYTRON_ENABLED_5_0.getAttribute().parseAndSetParameter(value, workManagerOperation, reader);
```
bstansberry, line 538: This won't end up writing 'false' to the operation unless the xml says 'false'. The AttributeDefinition's default value doesn't factor into parseAndSetParameter.

To get behavior where not explicitly configuring elytron-enabled results in writing false to the operation if an old namespace is used, at the end of the method if it's an old namespace check if workManagerOperation.hasDefined(caWorkManagerDefinition.WmParameters.ELYTRON_ENABLED.getAttribute()) and if not add it set to false.

BTW, if you want you can drop support for namespaces prior to 5.0.

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
```diff
@@ -630,7 +641,14 @@ private void parseDistributedWorkManager(final XMLExtendedStreamReader reader, f
                     }
                     case ELYTRON_ENABLED: {
                         switch (readerNS) {
-                            case JCA_5_0: {
+                            case JCA_5_0:
+                            {
+                                String value = rawElementText(reader);
+                                ((SimpleAttributeDefinition) JcaDistributedWorkManagerDefinition.DWmParameters.ELYTRON_ENABLED_5_0.getAttribute()).parseAndSetParameter(value, distributedWorkManagerOperation, reader);
+                                break;
```
bstansberry, line 648: Same comment I made in parseWorkManager applies here as well.

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
```diff
@@ -80,7 +80,7 @@ public class JcaExtension implements Extension {
 
     public static final String SUBSYSTEM_NAME = "jca";
 
-    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(5, 0, 0);
+    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(5, 1, 0);
```
bstansberry, line 83: Please use 6.0 for this and the XSD. We try to reserve minor versions for use downstream in case that's needed. The version numbers don't really have any semantic meaning.

bstansberry (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
tadamski: fixed 

I will push the namespace drop in the next PR

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
tadamski: fixed

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
tadamski: fixed

bstansberry: @tadamski This is still 5.1.

https://github.com/tadamski/wildfly/pull/4 changes that to 6.0 and also adds a mixed domain transformer and parser handling for old namespaces such that they produce the same ops that they did before (matching what the xsd says they'll do.)

tadamski: oops, sorry, fixed

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
```diff
@@ -630,9 +644,19 @@ private void parseDistributedWorkManager(final XMLExtendedStreamReader reader, f
...
+                                ((SimpleAttributeDefinition) JcaDistributedWorkManagerDefinition.DWmParameters.ELYTRON_ENABLED.getAttribute()).parseAndSetParameter(value, distributedWorkManagerOperation, reader);
+                                break;
+                            }
+                            case JCA_5_0:
+                            {
                                 String value = rawElementText(reader);
                                 ((SimpleAttributeDefinition) JcaDistributedWorkManagerDefinition.DWmParameters.ELYTRON_ENABLED.getAttribute()).parseAndSetParameter(value, distributedWorkManagerOperation, reader);
+                                if (!distributedWorkManagerOperation.hasDefined(JcaDistributedWorkManagerDefinition.DWmParameters.ELYTRON_ENABLED.getAttribute().getName())) {
+                                    distributedWorkManagerOperation.get(JcaDistributedWorkManagerDefinition.DWmParameters.ELYTRON_ENABLED.getAttribute().getName()).set(new ModelNode(false));
+                                }
```
bstansberry, line 659: Doing this test here doesn't account for the scenario where this element isn't present, which would be the typical way of configuring to use the default. Handling at the end of processing the entire [distributed-]workmanager element accounts for that; e.g. https://github.com/tadamski/wildfly/pull/4/commits/698ff398e98cbe68c31a0ce414df942439e1ecff#diff-e210f5446d28f4290a2dbc810897187203061441e0af68a8d403ad87afa7f6baR557

bstansberry: @tadamski Assuming CI's ok for this branch, I'm going to go ahead and merge this. We can use a separate blocker JIRA to track domain transformation and legacy parsing aspects, which don't relate to EE 10 compliance.

PR #15653, opened 2022-06-14
[WFLY-16481]:EndpointService throws NPE when publish ws endpint with …

jimma:
…EndpintPublisherImpl

https://issues.redhat.com/browse/WFLY-16481


bstansberry (approved):
@jimma This is ok but can you file a low priority JIRA to track getting rid of the need for this? If there isn't one already. I know this has been a general topic for a long time, so maybe there's no solution.

bstansberry:
@jimma Sorry; I missed your ping, and my previous comment was unclear.

I meant a WFLY to figure out how to get rid of the need for the direct service name manipulation in https://github.com/wildfly/wildfly/pull/15653/files#diff-51048489fc695b4afb5dc4ab0f6930b055aeb720c1becd75f61408d0e6fcbb0aR260-R265

It's not super high priority.

PR #15656, opened 2022-06-14
WFLY-16485 Use @Deprecated(forRemoval=true) where appropriate

pferraro:
https://issues.redhat.com/browse/WFLY-16485

### DIFF
--- a/clustering/service/src/main/java/org/wildfly/clustering/service/ServiceNameRegistry.java
+++ b/clustering/service/src/main/java/org/wildfly/clustering/service/ServiceNameRegistry.java
@@ -28,7 +28,7 @@ import org.jboss.msc.service.ServiceName;
  * Registry of services names for a set of requirements.
  * @author Paul Ferraro
  */
-@Deprecated
+@Deprecated(forRemoval = true)
 public interface ServiceNameRegistry<R extends Requirement> {
     /**
      * Returns the service name for the specified requirement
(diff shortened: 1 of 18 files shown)


clustering/service/src/main/java/org/wildfly/clustering/service/ServiceNameRegistry.java
```diff
@@ -28,7 +28,7 @@
  * Registry of services names for a set of requirements.
  * @author Paul Ferraro
  */
-@Deprecated
+@Deprecated(forRemoval = true)
```
rhusar, line 31: Tangential to the `forRemoval` – it is strongly recommended that the reason for deprecating a program element be explained in the documentation, using the @deprecated javadoc tag. The documentation should also suggest and link to a recommended replacement API, if applicable. A replacement API often has subtly different semantics, so such issues should be discussed as well.

Perhaps the ` * @deprecated To be removed without replacement` would suffice.

rhusar (approved):
LGTM (just a suggestion to add missing @deprecated javadoc while at it)

clustering/service/src/main/java/org/wildfly/clustering/service/ServiceNameRegistry.java
rhusar: Same applies to the enum members below.

rhusar (approved):
Looks great now, thanks!

PR #15660, opened 2022-06-15
Update to jakarta common annotations 2.1.1 to pick up the deprecated …

starksm64:
…@ManagedBean

Needed for EE 10 signature tests


bstansberry (approved):

bstansberry:
A few of the CI jobs failed because it looks like the artifact wasn't available when they tried to pull it. Others are running ok so I restarted those.

starksm64:
Actually, the update was incorrectly applied to the preview.version.jakarta.activation.jakarta.activation-api property. I updated the PR to set preview.version.jakarta.annotation.jakarta-annotation-api to 2.1.1.

starksm64:
It is strange that the dependency tree check passed with the bad activation version.

PR #15661, opened 2022-06-15
[WFLY-16492] Use 6.0.0 as the new ModelVersion; add domain transforma…

bstansberry:
…tion and legacy XSD parsing compatibility

https://issues.redhat.com/browse/WFLY-16492


tadamski (approved):
LGTM

PR #15662, opened 2022-06-15
[WFLY-16491] checked for null and took appropriate action

rsearls:
https://issues.redhat.com/browse/WFLY-16491 wfly-27  CapabilityServiceSupport is null


rsearls:
This fix is the same as [WFLY-16481].  PR https://github.com/wildfly/wildfly/pull/15653
With drawing

PR #15663, opened 2022-06-16
Simple documentation change to add a section link to the headers. Thi…

jamezp:
…s allows for easily copying a header link when you hover over the header.

Example output:
image

The link on the left of the header can be used to easily copy the link to the header.


bstansberry:
Excellent! Thanks @jamezp

PR #15667, opened 2022-06-16
[WFLY-16468] The microprofile-opentracing layer is redundant

jasondlee:
https://issues.redhat.com/browse/WFLY-16468

Combine "open-tracing" and "microprofile-opentracing" into the latter
Remove the "open-tracing" layer
Adjust references to the "open-tracing" layer to point to new combined layer

### DIFF
--- a/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
+++ b/docs/src/main/asciidoc/_galleon/Galleon_layers.adoc
@@ -363,7 +363,7 @@ link:#gal.microprofile-health[microprofile-health] (optional) +
 link:#gal.microprofile-jwt[microprofile-jwt] (optional) +
 link:#gal.microprofile-metrics[microprofile-metrics] (optional) +
 link:#gal.microprofile-openapi[microprofile-openapi] (optional) +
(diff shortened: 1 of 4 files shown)


bstansberry (approved):

jasondlee:
@jfdenise Would you mind taking a look at this to make sure we've not missed anything relevant?

jfdenise:
@jasondlee , I am fine with the removal. We are not directly dealing with the open-tracing layer in the cloud context or bootable JAR context. I just found occurrence of it in some tests based on WF26.

jfdenise (approved):

jasondlee:
> @jasondlee , I am fine with the removal. We are not directly dealing with the open-tracing layer in the cloud context or bootable JAR context. I just found occurrence of it in some tests based on WF26.

Which tests were they? I searched and didn't see anything that looked relevant, but I may not have used the right keywords.

jfdenise:
Some cloud tests: https://github.com/wildfly/wildfly-s2i/blob/wf-26.0/wildfly-modules/tests/features/s2i.feature#L176, not related to WF.

bstansberry:
Thanks, @jasondlee

PR #15674, opened 2022-06-17
[WFLY-15435]:Move WildFly Preview to a native jakarta namespace varia…

jimma:
…nt of JBoss WS

https://issues.redhat.com/browse/WFLY-15435


pom.xml
```diff
@@ -210,7 +210,7 @@
              to non-snapshot public maven repos -->
         <maven.repository.protocol>https</maven.repository.protocol>
         <!-- The full remote maven repo URL; can be overridden via -D for special use cases -->
-        <maven.repository.url>${maven.repository.protocol}://repository.jboss.org/nexus/content/groups/public/</maven.repository.url>
+        <maven.repository.url>${maven.repository.protocol}://repository.jboss.org/nexus/content/groups/developer/</maven.repository.url>
```
jamezp, line 213: Is this just for testing? We can't make this change otherwise.

jamezp (changes requested):
Overall this looks okay. It can't be merged with the change to the repository and there does seem to be transformation happening with `jbossws-cxf-jaspi` due to missing security upgrades.

pom.xml
jimma: This is something for test I didn't remove it.  I already fixed this . Thanks @jamezp

jimma:
> Overall this looks okay. It can't be merged with the change to the repository and there does seem to be transformation happening with `jbossws-cxf-jaspi` due to missing security upgrades.

The jbossws-cxf-jaspi is some legacy thing I don't plan to migrate to jakarta namespace and remove in the later PR after we all agree on this removal.

ee-9/pom.xml
```diff
@@ -2784,6 +2784,90 @@
...
+                    </exclusion>
+                </exclusions>
+            </dependency>
+
+            <dependency>
+                <groupId>org.jboss.ws.cxf</groupId>
+                <artifactId>jbossws-cxf-resources</artifactId>
+                <version>${preview.version.org.jboss.ws.cxf}</version>
+                <classifier>wildfly2700</classifier>
+            </dependency>
```
yersan, line 2822: Should not this artifact exclude all the transitive dependencies?

We are now following such a pattern.

jamezp: That's a good catch Yeray. I'd agree here.

jamezp:
@jimma We should remove the `jbossws-cxf-jaspi` module from preview.

jimma:
> @jimma We should remove the `jbossws-cxf-jaspi` module from preview.

@jamezp  Done. Given jbossws-cxf-jaspi doesn't work in WildFly standard either and I remove it from ee-feature-patck too.

jamezp (approved):
This looks good to me. @darranl Any objection to removing the jbossws-cxf-jaspi from the current EE 8 variant too? Given big-bang is likely not far off and it doesn't work anyway, I think it's fine.

asoldano (approved):

PR #15680, opened 2022-06-21
WFLY-16489 Clustering TS: Include node name in clustering console log messages

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-16489

Motivation for the change explained on Jira.


pferraro (approved):
Thanks so much for this.  This is one of those things that I've silently tolerated for years...

PR #15686, opened 2022-06-21
Resolve CNFE in upstream weld in-container tests.

pferraro:
Also fixes TCK regression in org.jboss.cdi.tck.tests.implementation.enterprise.lifecycle.EnterpriseBeanLifecycleTest.


manovotn:
@pferraro this doesn't solve the problem for WFLY preview because there is a separate `module.xml` file for weld core module.
Preview variant is located [here](https://github.com/wildfly/wildfly/blob/main/ee-9/common/src/main/resources/modules/system/layers/base/org/jboss/weld/core/main/module.xml).
Adding the same change there fixes the problem I was seeing in Weld TS.

manovotn (approved):
Current state of the PR fixes all of the problems I was seeing.

PR #15702, opened 2022-06-26
[WFLY-16540] Adding missing JPMS open privileges required by JCETestCase

ropalka:
https://issues.redhat.com/browse/WFLY-16540

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -41,6 +41,7 @@
         <bootable-jar-copy-module-files.phase>none</bootable-jar-copy-module-files.phase>
         <!-- Disable the default surefire test execution. -->
         <surefire.default-test.phase>none</surefire.default-test.phase>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -175,6 +175,8 @@
             --add-opens=java.base/java.security=ALL-UNNAMED
             --add-opens=java.base/java.util=ALL-UNNAMED
             --add-opens=java.base/java.util.concurrent=ALL-UNNAMED
+            --add-opens=java.base/javax.crypto=ALL-UNNAMED
+            --add-opens=java.base/sun.security.validator=ALL-UNNAMED
```
bstansberry, line 179: If these are only needed by this one test then we need to organize things so these 2 added add-opens are only applied in that testsuite module, and ideally only for a surefire execution that runs that test. But for now please just limit the addition to testsuite/integration/basic.

bstansberry (changes requested):

pom.xml
ropalka: Done.

bstansberry (approved):

bstansberry:
Thanks @ropalka!

PR #15706, opened 2022-06-27
[WFLY-16545] Register the @Provider, @Path and @ApplicationPath annot…

jamezp:
…ations as bean defining annotations.

https://issues.redhat.com/browse/WFLY-16545

Note the some what of a revert back to an empty `beans.xml` for these tests as `@ApplicationScoped` was added to the `CDIBean`. This was the requirement for the `bean-discovery-mode="all"`. However, by defining the scope on the bean that resolves the bean discovery issue.

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanDefiningAnnotationProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanDefiningAnnotationProcessor.java
@@ -34,6 +34,10 @@ public class BeanDefiningAnnotationProcessor implements DeploymentUnitProcessor 
 
     private static final DotName VIEW_SCOPED_NAME = DotName.createSimple("javax.faces.view.ViewScoped");
     private static final DotName FLOW_SCOPED_NAME = DotName.createSimple("javax.faces.flow.FlowScoped");
+    // Jakarta REST annotations
+    private static final DotName PROVIDER = DotName.createSimple("javax.ws.rs.ext.Provider");
+    private static final DotName APPLICATION_PATH = DotName.createSimple("javax.ws.rs.ApplicationPath");
+    private static final DotName PATH = DotName.createSimple("javax.ws.rs.Path");
 
     @Override
     public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
@@ -52,6 +56,15 @@ public class BeanDefiningAnnotationProcessor implements DeploymentUnitProcessor 
         addAnnotation(deploymentUnit, new AnnotationType(TransactionScoped.class));
         addAnnotation(deploymentUnit, new AnnotationType(VIEW_SCOPED_NAME, true));
         addAnnotation(deploymentUnit, new AnnotationType(FLOW_SCOPED_NAME, true));
+        // Per section 11.2.3 of the Jakarta REST 3.1 specification:
+        // In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource
+        // classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or
+        // use application scope.
+        // Currently, these are not specified as @Stereotype annotations with a default scope. In a later spec this may
+        // happen, in which case these can be removed.
+        addAnnotation(deploymentUnit, new AnnotationType(PROVIDER, false));
+        addAnnotation(deploymentUnit, new AnnotationType(APPLICATION_PATH, false));
+        addAnnotation(deploymentUnit, new AnnotationType(PATH, false));
 
         for (AnnotationType annotationType : CdiAnnotations.BEAN_DEFINING_META_ANNOTATIONS) {
             addAnnotations(deploymentUnit, getAnnotationsAnnotatedWith(index, annotationType.getName()));
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CDIApplicationPathIntegrationTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CDIApplicationPathIntegrationTestCase.java
@@ -31,7 +31,7 @@ import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.as.test.integration.common.HttpRequest;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -52,7 +52,7 @@ public class CDIApplicationPathIntegrationTestCase {
         WebArchive war = ShrinkWrap.create(WebArchive.class,"jaxrsapp.war");
         war.addPackage(HttpRequest.class.getPackage());
         war.addClasses(CDIApplicationPathIntegrationTestCase.class, CDIBean.class, CDIPathApplication.class, CDIResource.class);
-        war.add(new StringAsset("<beans bean-discovery-mode=\"all\"></beans>"), "WEB-INF/beans.xml");
+        war.add(EmptyAsset.INSTANCE, "WEB-INF/beans.xml");
         return war;
     }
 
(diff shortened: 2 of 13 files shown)


jamezp:
These failures are valid. I've not looked into yet as maybe we don't want to end up doing this as it does cause issues with deployments that were previously not processed as beans. This is the reason it's now a draft.

manovotn (commented):
Looking at `Linux - JDK 11` CI run, there are failures in `CoordinatorTestCase`. These are because `org.wildfly.test.extension.rts.common.WorkRestATResource` is marked as `final` and that will fail if it needs to create any subclass (which is always the case for normal scoped beans such as application scoped, or if you need to intercept/decorate these beans).
And I think all of these resources are going to be `@ApplicationScoped`, right? That's what's done in `org.jboss.resteasy.cdiResteasyCdiExtension` IIRC.
Does jaxrs spec say anything about these resources being final?

In the `OpenAPIMultiModuleDeploymentIndexTestCase` I am not sure what's the problem. It seems like the EJB is not recognized but nothing has changes WRT that.This might actually be the case you described where the deployment wasn't a bean archive previously (this test uses EAR with WAR and EJB jar)?

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanDefiningAnnotationProcessor.java
```diff
@@ -34,6 +34,10 @@ public class BeanDefiningAnnotationProcessor implements DeploymentUnitProcessor
 
     private static final DotName VIEW_SCOPED_NAME = DotName.createSimple("javax.faces.view.ViewScoped");
     private static final DotName FLOW_SCOPED_NAME = DotName.createSimple("javax.faces.flow.FlowScoped");
+    // Jakarta REST annotations
+    private static final DotName PROVIDER = DotName.createSimple("javax.ws.rs.ext.Provider");
```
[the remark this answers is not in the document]
jamezp: We're not quite there yet. We're close though :)

jamezp:
@manovotn Interestingly enough, no it doesn't say anything about whether a resource can be final. There is an [example](https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1.html#dynamic_binding) of a `@Provider` which does show a final class. However, that seems wrong especially given a provider is required to be a CDI bean if CDI is available.

The default scope for resources is `@RequestScoped`, but that's not a huge issue :) And yes the `org.jboss.resteasy.cdiResteasyCdiExtension` assigns a scope to resources, providers and applications (`jakata.ws.rs.core.Application`).

manovotn:
> However, that seems wrong especially given a provider is required to be a CDI bean if CDI is available.

+1, those classes are not proxyable according to [CDI spec](https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#unproxyable)
For clarity, client proxy is required for any built-in bean scope apart from `@Dependent` and `@Singleton`

jamezp:
@manovotn This has some information on why this was not done before https://issues.redhat.com/browse/WFLY-2859. Well done, but reverted it seems.

manovotn:
@jamezp right, that's exactly what we talked about - that some archives that were not bean archive now will be which can in turn lead to some side effects.
But truth be told, I don't fully understand how that leads to unproxyable exception? That is, unless you absolutely have to allow `final` class `Provider` and other things which just doesn't work with CDI by definition.

jamezp:
@manovotn I definitely agree with you. Especially since the REST spec states that if CDI is available that applications, providers and resources MUST be CDI beans. I guess really it's an argument of should we mark the deployment as a CDI deployment if it contains any of those.

manovotn:
Right, this should be really brought up to the spec. If you create an issue, I can pitch in with CDI perspective in case it's needed.

Also note that there is a possible scenario in which you can have an archive with REST bits that should become beans but there can be `beans.xml` that explicitly states `none` discovery mode. But I guess you'd have fallback for that already.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanDefiningAnnotationProcessor.java
```diff
@@ -52,6 +56,15 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
         addAnnotation(deploymentUnit, new AnnotationType(FLOW_SCOPED_NAME, true));
+        // Per section 11.2.3 of the Jakarta REST 3.1 specification:
+        // In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource
+        // classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or
+        // use application scope.
+        // Currently, these are not specified as @Stereotype annotations with a default scope. In a later spec this may
+        // happen, in which case these can be removed.
+        addAnnotation(deploymentUnit, new AnnotationType(PROVIDER, false));
+        addAnnotation(deploymentUnit, new AnnotationType(APPLICATION_PATH, false));
+        addAnnotation(deploymentUnit, new AnnotationType(PATH, false));
```
bstansberry, line 67: @jamezp I think these names (or perhaps the AnnotationType objects), and any others handled this way in this DUP, that are not required by the CDI spec itself should be passed in by the relevant subsystem using a new method in WeldCapability.

manovotn: Would that happen early enough though? Weld needs to know the full set of bean defining annotations *before* it analyses any archive in order to determine if the archive is a bean archive in the first place.

jamezp: I could add a DUP which executes between `PARSE_CDI_BEAN_DEFINING_ANNOTATIONS` and `PARSE_WELD_DEPLOYMENT`. There would be room for 128 different processors between those so that does seem reasonable.

I'll update the PR locally and do some testing.

jamezp: @bstansberry @manovotn We could do something like this https://github.com/jamezp/wildfly/commit/986a777f224b40ed52d82c594e495f1aaa6fa6b1. It passed the single tests I added. I'll run the full test suite too and see how it does. It needs some cleanup, but that was a quick test. Using a DUP requires subsystems to have a dependency on `org.jboss.as.weld` which didn't seem right. This keeps the dependency on `org.jboss.as.weld.common`.

bstansberry: @manovotn -- a subsystem attribute like James did is what I had in mind. The values get provided to the weld subsystem object before any DUP runs.

@jamezp I don't see the need for a new capability interface; the existing WeldCapability should be fine.

manovotn: > a subsystem attribute like James did is what I had in mind. The values get provided to the weld subsystem object before any DUP runs.

Got it, makes sense.

> the existing WeldCapability should be fine.

+1, let's reuse that

jamezp: Ah, right. It's a capability runtime API so it's not created and added very early. I'll update the PR as I think that makes the most sense.

jamezp:
I unfortunately don't think this is going to work. [Section 3.1.2](https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1.html#resource_class_constructor) of the specification states:

> A public constructor MAY include parameters annotated with one of the following: `@Context`, `@HeaderParam`, `@CookieParam`, `@MatrixParam`, `@QueryParam` or `@PathParam`.

There is no way I'm aware of in CDI to support cases like this.

manovotn:
> I unfortunately don't think this is going to work. [Section 3.1.2](https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1.html#resource_class_constructor) of the specification states:
> 
> > A public constructor MAY include parameters annotated with one of the following: `@Context`, `@HeaderParam`, `@CookieParam`, `@MatrixParam`, `@QueryParam` or `@PathParam`.
> 
> There is no way I'm aware of in CDI to support cases like this.

Don't you already support that via wrapping of `InjectionTarget` in `ResteasyCdiExtension`? That logic could stay in place but the discovery removes the need for manual registration of beans. Or am I missing something?

jamezp:
Given a Zulip conversation with @manovotn about clarification we likely need in the REST spec, I'm going to close this for now. I'll leave the JIRA open for the time being. There _might_ be some workarounds for this, but the constructor parameter injection is a real issue that could break users.

PR #15717, opened 2022-06-29
[WFLY-16387] Test to demonstrate being able to call the platform mbea…

kabir:
…n server

with RBAC enabled. We need to piggy-back on application security.

Test for https://issues.redhat.com/browse/WFLY-16387. WildFly provides all the functionality, but the user needs to do some setup

‼️ Although this is linked to a feature request, it just adds a test for functionality that already exists. So it can be merged once approved

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/jmx/rbac/AbstractJmxAccessFromDeploymentWithRbacTest.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/jmx/rbac/AbstractJmxAccessFromDeploymentWithRbacTest.java
@@ -26,8 +26,8 @@ import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
-import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.as.arquillian.container.ManagementClient;
 import org.jboss.as.controller.PathAddress;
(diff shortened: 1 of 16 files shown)


darranl:
@fjuma FYI

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/jmx/rbac/AbstractJmxAccessFromDeploymentWithRbacTest.java
```diff
@@ -0,0 +1,256 @@
...
+                operations.add(addSuperUserInclude);
+            }
+
+            // /core-service=management/access=identity:add(security-domain=ManagementDomain)
+            addr = PathAddress.pathAddress(CORE_SERVICE, MANAGEMENT).append(ACCESS, IDENTITY);
+            ModelNode addAccessIdentity = Util.createAddOperation(addr);
+            addAccessIdentity.get("security-domain").set("ManagementDomain");
+            operations.add(addAccessIdentity);
+
+            // /subsystem=undertow/application-security-domain=other:write-attribute(name=http-authentication-factory,value=application-http-authentication)
```
fjuma, line 181: I don't think this is necessary, the default `application-security-domain` configuration should work here.

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/jmx/rbac/AbstractJmxAccessFromDeploymentWithRbacTest.java
```diff
@@ -0,0 +1,256 @@
...
+            addr = PathAddress.pathAddress(SUBSYSTEM, "undertow").append("application-security-domain", "other");
+            operations.add(Util.getWriteAttributeOperation(addr, "http-authentication-factory", "application-http-authentication"));
+
+            // /subsystem=undertow/application-security-domain=other:undefine-attribute(name=security-domain)
```
fjuma, line 185: Same here, I think this can be removed since the default `application-security-domain` configuration should work I think.

fjuma (approved):

PR #15720, opened 2022-06-29
[WFLY-16560] Remove the wildfly-servlet feature pack and the servlet-…

bstansberry:
…build and servlet-dist modules

https://issues.redhat.com/browse/WFLY-16560

This will fail in testsuite/layers but while I sort that I'll let CI have a look at it.


darranl:
Looks like CI largely went well, just the layers test case which needs some adjustment.

bstansberry:
@darranl LayersTestCase passes for me locally now so this should be good to go.

I added a bunch of TODO comments to LayersTestCase for things that need fixing or investigation with how we provision things with a full feature pack not just wildfly-servlet. Those are existing issues revealed by improving the test though, so they are out of scope for WFLY-16560.  A number of them already have fixes in progress.

PR #15721, opened 2022-06-29
[WFLY-16527] Faces 4 fails if CDI isn't enabled for the deployment

jasondlee:
https://issues.redhat.com/browse/WFLY-16527

Add check to require CDI for Faces deployments
Add method to WeldCapability to force Weld initialization
Call new method to force CDI init for Faces deployments

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFComponentProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFComponentProcessor.java
@@ -21,6 +21,18 @@
  */
 package org.jboss.as.jsf.deployment;
 
+import java.io.InputStream;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamConstants;
+import javax.xml.stream.XMLStreamReader;
+
 import org.jboss.as.ee.component.ComponentDescription;
 import org.jboss.as.ee.component.EEApplicationClasses;
 import org.jboss.as.ee.component.EEModuleDescription;
@@ -46,21 +58,6 @@ import org.jboss.metadata.parser.util.NoopXMLResolver;
 import org.jboss.metadata.web.spec.WebMetaData;
 import org.jboss.modules.Module;
 import org.jboss.vfs.VirtualFile;
-
-import javax.xml.stream.XMLInputFactory;
-import javax.xml.stream.XMLStreamConstants;
-import javax.xml.stream.XMLStreamReader;
-import java.io.InputStream;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import org.jboss.as.controller.capability.CapabilityServiceSupport;
-import static org.jboss.as.weld.Capabilities.WELD_CAPABILITY_NAME;
-import org.jboss.as.weld.WeldCapability;
 
 /**
  * Sets up Jakarta Server Faces managed beans as components using information in the annotations and
@@ -240,7 +237,7 @@ public class JSFComponentProcessor implements DeploymentUnitProcessor {
         for (String managedBean : managedBeanClasses) {
             installManagedBeanComponent(managedBean, moduleDescription, module, deploymentUnit, applicationClassesDescription);
         }
-        // process all the other elements elegible for injection in the Jakarta Server Faces spec
+        // process all the other elements eligible for injection in the Jakarta Server Faces spec
         processJSFArtifactsForInjection(deploymentUnit, managedBeanClasses);
     }
 
@@ -254,9 +251,6 @@ public class JSFComponentProcessor implements DeploymentUnitProcessor {
      * classes to use for injection.
      */
     private void processJSFArtifactsForInjection(final DeploymentUnit deploymentUnit, final Set<String> managedBeanClasses) {
-        final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
-        boolean isCDI = support.hasCapability(WELD_CAPABILITY_NAME) &&
-                support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class).get().isPartOfWeldDeployment(deploymentUnit);
         final EEApplicationClasses applicationClassesDescription = deploymentUnit.getAttachment(org.jboss.as.ee.component.Attachments.EE_APPLICATION_CLASSES_DESCRIPTION);
         final EEModuleDescription moduleDescription = deploymentUnit.getAttachment(org.jboss.as.ee.component.Attachments.EE_MODULE_DESCRIPTION);
         final Module module = deploymentUnit.getAttachment(Attachments.MODULE);
@@ -294,7 +288,7 @@ public class JSFComponentProcessor implements DeploymentUnitProcessor {
                                         if (child.getTag() == JsfTag.MANAGED_BEAN_CLASS) {
                                             installManagedBeanComponent(className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
                                         } else {
-                                            installJsfArtifactComponent(child.getTag().getTagName(), className, isCDI, moduleDescription, module, deploymentUnit, applicationClassesDescription);
+                                            installJsfArtifactComponent(child.getTag().getTagName(), className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
                                         }
                                     }
                                 } else if (child != null) {
@@ -406,21 +400,8 @@ public class JSFComponentProcessor implements DeploymentUnitProcessor {
         install(JsfTag.MANAGED_BEAN.getTagName(), className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
     }
 
-    private void installJsfArtifactComponent(String type, String className, boolean isCDI, final EEModuleDescription moduleDescription,
+    private void installJsfArtifactComponent(String type, String className, final EEModuleDescription moduleDescription,
             final Module module, final DeploymentUnit deploymentUnit, final EEApplicationClasses applicationClassesDescription) {
-        try {
-            final Class<?> componentClass = module.getClassLoader().loadClass(className);
-            if (!isCDI) {
-                // if not Jakarta Contexts and Dependency Injection the default constructor is compulsory to inject the artifact
-                componentClass.getConstructor();
-            }
-        } catch (ClassNotFoundException e) {
-            JSFLogger.ROOT_LOGGER.managedBeanLoadFail(className);
-            return;
-        } catch (NoSuchMethodException e) {
-            JSFLogger.ROOT_LOGGER.jsfArtifactNoDefaultConstructor(type, className);
-            return;
-        }
         install(type, className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
     }
 
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFResourceDefinition.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFResourceDefinition.java
@@ -18,19 +18,22 @@
  */
 package org.jboss.as.jsf.subsystem;
 
+import java.util.Arrays;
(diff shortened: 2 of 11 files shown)


weld/subsystem/src/main/java/org/jboss/as/weld/WeldCapabilityImpl.java
```diff
@@ -67,11 +67,15 @@ public ServiceBuilder<?> addBeanManagerService(final DeploymentUnit unit, final
...
+        return forceWeldDeployment || WeldDeploymentMarker.isPartOfWeldDeployment(unit);
     }
 
     public boolean isWeldDeployment(final DeploymentUnit unit) {
-        return WeldDeploymentMarker.isWeldDeployment(unit);
+        return forceWeldDeployment || WeldDeploymentMarker.isWeldDeployment(unit);
+    }
+
+    public void markAsWeldDeployment() {
+        this.forceWeldDeployment = true;
```
bstansberry, line 78: This should call WeldDeploymentMarker.mark

bstansberry: I guess this way can work too, but it seems cleaner to just use the existing WeldDeploymentMarker whose purpose is to track this state.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFComponentProcessor.java
```diff
@@ -253,10 +254,15 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
      * using the facesConfigElement tree it is known when a tag is one of the
      * classes to use for injection.
      */
-    private void processJSFArtifactsForInjection(final DeploymentUnit deploymentUnit, final Set<String> managedBeanClasses) {
+    private void processJSFArtifactsForInjection(final DeploymentUnit deploymentUnit, final Set<String> managedBeanClasses)
+            throws DeploymentUnitProcessingException {
         final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
-        boolean isCDI = support.hasCapability(WELD_CAPABILITY_NAME) &&
-                support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class).get().isPartOfWeldDeployment(deploymentUnit);
+        if (!support.hasCapability(WELD_CAPABILITY_NAME)) {
```
bstansberry, line 260: The JSFResourceDefinition should declare an 'org.wildfly.faces' capability that requires WELD_CAPABILITY_NAME. Convert the JSFResourceDefinition c'tor to pass a SimpleResourceDefinition.Parameters object to super(...) and pass that new capability into that object.

Once you have done that there is no longer any possibility that WeldCapability will not be present when a jsf subsystem DUP runs.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFComponentProcessor.java
```diff
@@ -294,7 +300,7 @@ private void processJSFArtifactsForInjection(final DeploymentUnit deploymentUnit
                                         if (child.getTag() == JsfTag.MANAGED_BEAN_CLASS) {
                                             installManagedBeanComponent(className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
                                         } else {
-                                            installJsfArtifactComponent(child.getTag().getTagName(), className, isCDI, moduleDescription, module, deploymentUnit, applicationClassesDescription);
+                                            installJsfArtifactComponent(child.getTag().getTagName(), className, moduleDescription, module, deploymentUnit, applicationClassesDescription);
```
bstansberry, line 303: Yep; isCDI isn't needed any longer. Any other jsf subsystem code that is conditionally checking for CDI can stop too once JSFResourceDefinition declare an 'org.wildfly.faces' capability that requires WELD_CAPABILITY_NAME.

bstansberry (changes requested):
See my previous comments for requested changes.

jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFResourceDefinition.java
```diff
@@ -59,11 +65,27 @@ public class JSFResourceDefinition extends PersistentResourceDefinition {
...
+    private MailServerDefinition(final PathElement path, AttributeDefinition[] attributes) {
+        super(new Parameters(path,
+                MailExtension.getResourceDescriptionResolver(MailSubsystemModel.MAIL_SESSION, MailSubsystemModel.SERVER_TYPE))
+                .setAddHandler(new MailServerAdd(attributes))
+                .setRemoveHandler(new MailServerRemove(attributes))
+                .setCapabilities(SERVER_CAPABILITY)
+        );
+        this.attributes = Arrays.asList(attributes);
+    }
+     */
```
jamezp, line 78: I assume this was the example and can be removed :)

jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFResourceDefinition.java
```diff
@@ -59,11 +65,27 @@ public class JSFResourceDefinition extends PersistentResourceDefinition {
...
-                ReloadRequiredRemoveStepHandler.INSTANCE);
+        super(new SimpleResourceDefinition.Parameters(JSFExtension.PATH_SUBSYSTEM, JSFExtension.getResourceDescriptionResolver())
+                .setAddHandler(JSFSubsystemAdd.INSTANCE)
+                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
+                .addCapabilities(FACES_CAPABILITY)
+        );
+//        super(JSFExtension.PATH_SUBSYSTEM,
+//                JSFExtension.getResourceDescriptionResolver(),
+//                JSFSubsystemAdd.INSTANCE,
+//                ReloadRequiredRemoveStepHandler.INSTANCE);
```
jamezp, line 88: This can likely be removed too.

jamezp (commented):
Just two minor comments, but otherwise looks good to me.

jamezp:
Actually, the test failures seem to be related.

jsf/subsystem/src/main/java/org/jboss/as/jsf/subsystem/JSFResourceDefinition.java
jasondlee: Good grief. Yes, it can. :P

bstansberry (changes requested):
@jasondlee The functional change looks good.

Can you remove this line please? 

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/pom.xml#L3538

I believe with -Dts.ee9 that test should now pass, and if it doesn't that may mean something more is needed.

My guess is passing AdditionalInitialization.MANAGEMENT instead of null in this line will fix the subsystem test:

https://github.com/wildfly/wildfly/blob/main/jsf/subsystem/src/test/java/org/jboss/as/jsf/subsystem/JSFSubsystemTestCase.java#L58

You've added a subsystem requirement for WeldCapability. For that to work you need to provide the capability, unless the test controller starts in admin-only mode. The subsystem test framework has hooks for providing capabilities, but in this case starting in admin-only is sufficient to check what the test checks, so doing that is the simplest solution. Passing AdditionalInitialization.MANAGEMENT does that.

If you make those changes and it passes CI, please don't wait for my approval to get this merged, as I may not check back for some days.

jasondlee:
I've updated the POM and the test change suggestion worked. I'll push the changes, then find someone to merge when it's green. Thanks!

bstansberry:
@jasondlee Once this is merged and a 'main' with this in it is merged into the big-bang branch, can you please send up a PR to big-bang removing this line?

https://github.com/wildfly/wildfly/blob/big-bang/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/testurl/ForbiddenUrlTestCase.java#L55

Thanks!

jasondlee:
> @jasondlee Once this is merged and a 'main' with this in it is merged into the big-bang branch, can you please send up a PR to big-bang removing this line?

Prepped and ready to go: https://github.com/wildfly/wildfly/pull/15725

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/testurl/ForbiddenUrlTestCase.java
```diff
@@ -98,13 +100,32 @@ public void testAllowedUrl() throws Exception {
...
+            HttpUriRequest getVarRequest = new HttpGet(url.toExternalForm() + jakartafiUrl(FORBIDDEN_URL));
             try (CloseableHttpResponse getVarResponse = client.execute(getVarRequest)) {
                 int statusCode = getVarResponse.getStatusLine().getStatusCode();
-                Assert.assertEquals("Status code should be " + FORBIDDEN_STATUS_CODE + ", but is " + statusCode + ". For more information see JBEAP-18842. ", FORBIDDEN_STATUS_CODE, statusCode);
+                Assert.assertEquals("Status code should be " + FORBIDDEN_STATUS_CODE + ", but is " + statusCode +
+                        ". For more information see JBEAP-18842. ", FORBIDDEN_STATUS_CODE, statusCode);
             }
         }
     }
+
```
jasondlee, line 113: @jamezp @bstansberry, This hack is a bit ugly, but, other than continuing to ignore the test, I'm not sure how best to go about it. That said, the test does now pass in main and WFP as expected.  Also, IDEA cleaned up some formatting and import order that I didn't notice before pushing. I don't mind the clean up, but if you'd like less noise, I can reset and make a quieter change for this. :)

jamezp: I assume `javax.faces.resource` has special meeting in which case I can't think of a better way to handle it.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/testurl/ForbiddenUrlTestCase.java
```diff
@@ -98,13 +100,32 @@ public void testAllowedUrl() throws Exception {
...
+     * EE 9 changes the namespace from javax.* to jakarta.*. Running these tests under EE 9 or later, as WildFly Preview does, will
+     * cause the tests to fail, since they're using the old namespace in strings. This method will alter the URLs if it can detect
+     * that the tests are running under EE 9 or later. Once WildFly switches completely to EE 10, this hack can be removed, and the
+     * strings updated appropriately.
+     * @param url
+     * @return
+     */
+    @Deprecated
+    private String jakartafiUrl(String url) {
+        if (System.getProperty("ts.ee9") != null) {
```
bstansberry, line 125: Instead of checking the active profile, which will break for other profiles or in big-bang, please inspect the FQCN of a Faces API class. If it starts with 'jakarta' then the test is in a 'jakarta' environment.

jasondlee: This is a client test (@RunAsClient), so the namespace is not being transformed, giving me a javax.faces.context.FacesContext in the test method. Class.forName("jakarta.faces.context.FacesContext") throws a CNFE. Am I missing a trick?

bstansberry: Ah, ok, sorry; then testing for the profile is ok. Please use AssumeTestGroupUtil.isWildFlyPreview() though.

#15725 can back this out, as the class already uses the jakarta.* URL in the big-bang branch. That's just minor housekeeping though as the method will still work, it just won't do anything.

bstansberry:
@darranl If you want to merge this as-is on Monday, go ahead. My last comment above is a minor thing.

fjuma (approved):
@darranl Sorry for the delayed response, it was a holiday here on Friday. This looks good to me.

darranl:
No problem, thank you @fjuma I remembered just as I pinged you

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/testurl/ForbiddenUrlTestCase.java
jasondlee: It is. It's a mean to access bundled resources in META-INF, iirc.

jasondlee:
> > @jasondlee Once this is merged and a 'main' with this in it is merged into the big-bang branch, can you please send up a PR to big-bang removing this line?
> 
> Prepped and ready to go: #15725

I'll update this pending PR to remove the WFP check once the change here is merged.

PR #15722, opened 2022-06-29
[DO NOT MERGE] Using for CI testing only

jamezp:
Testing with CI with removing transforming of the feature-pack. This also has main merged into.


ee-9/build/pom.xml
```diff
@@ -88,7 +88,6 @@
                             <plugin-options>
                                 <jboss-maven-dist/>
                                 <jboss-jakarta-transform-artifacts>false</jboss-jakarta-transform-artifacts>
```
jfdenise, line 90: Yo ucan remove all these occurences, you disabled transformation when building the feature-pack, so jakarta-transformation is disabled by default.

PR #15726, opened 2022-06-30
[WFLY-16415] Exclude transitive openjdk-orb from wildfly-messaging

dbkreling:
Build fails with dependency convergence error for the community version of
'org.jboss.openjdk-orb:openjdk-orb' coming from module @wildfly-messaging,
from wildfly-messaging-activemq-subsystem.
Excluding it.


darranl (approved):
This is just a test scope dependency and CI passed so I am going to approve.

ehsavoie (approved):

PR #15735, opened 2022-07-01
WFLY-16581 Convert testsuite/integration/microprofile to the jakarta.* namespace

pferraro:
https://issues.redhat.com/browse/WFLY-16581


kabir (commented):
@darranl the only thing that stands out to me is the removal of the opentracing tests. I realise this subsystem is likely to go away, but until the subsystem is actually removed it seems better to keep the tests (if possible). @jasondlee wdyt?

pferraro:
@kabir The opentracing tests to which you refer are duplicates, the existing testsuite in main contains javax and jakarta variants of the same tests (see: https://github.com/wildfly/wildfly/tree/main/testsuite/integration/microprofile/src/test).
This PR removes the javax variants of these tests, and moves the jakarta variants to the src/main/java directory.

kabir:
@pferraro @darranl in that case no objections :-)

PR #15737, opened 2022-07-01
WFLY-16582 Remove unused org.jboss.ejb.remote.protocol module

chengfang:
https://issues.redhat.com/browse/WFLY-16582


tadamski (approved):
I don't recall seeing this module used anywhere. I think we can remove it.

PR #15738, opened 2022-07-02
[WFLY-14375] Upgrade Apache DS to version AM26

rmartinc:
Issue: https://issues.redhat.com/browse/WFLY-14375

This is the upgrade for apache-ds but in the big-bang branch. It's the same than #15712 but adding two things:

* Some leftovers in the manualmode are deleted (utility files for tests that were deleted before). These files used apache-ds and are not needed in the manualmode folder now. So better remove them.
* I have moved the opens from main testsuite folder to basic and elytron only. Apache-ds is just used in those two test directories.

But as commented this is not urgent. If you prefer to just wait until all the big-bang is in main, it's perfectly OK to me.

Thanks!


darranl:
Just running some more CI, this looks close.

darranl (approved):
Adding an approval to merge once we have clean CI.

PR #15739, opened 2022-07-04
[WFLY-16562][WFLY-16571][WFLY-16585][WFLY-16589] Fix some issues found when updating LayersTestCase

bstansberry:
Fix various issues discovered while doing #15720

Builds on #15720 and #15737

https://issues.redhat.com/browse/WFLY-16562
https://issues.redhat.com/browse/WFLY-16571
https://issues.redhat.com/browse/WFLY-16585
https://issues.redhat.com/browse/WFLY-16589


darranl:
Just retriggered Linux with security manager, the other jobs can be ignored.

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
```diff
@@ -137,37 +259,18 @@ public static void cleanUp() {
...
-        LayersTest.test(servletRoot, new HashSet<>(Arrays.asList(NOT_REFERENCED)),
-                new HashSet<>(Arrays.asList(NOT_USED)));
-    }
-
     @Test
     public void test() throws Exception {
-        // TODO, no more testing than provisioning and execution of layers for now.
-        String root = System.getProperty("layers.install.root");
-        LayersTest.testExecution(root);
+        LayersTest.test(root, new HashSet<>(Arrays.asList(NOT_REFERENCED)),
```
yersan, line 267: It looks all fine but I'm just wondering whether we can use this test for testing the layers.

As far as I remember, this test was created to verify the provisioned modules of each Galleon Layer comparing it with a reference installation provisioned without using Galleon Layers. We only had the servlet distribution as an installation provisioned from the feature pack without using Galleon layers, so, I'm just wondering if it is still valid to use this test.

@jfdenise WDYT?

darranl:
CI looks good, once we have had a couple of reviews this can be merged,

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
yersan: Ahh, sorry for the noise @jfdenise , I think now I got it, we are now provisioning without using layers the reference distribution in the pom from the ee-wildfly galleon pack: [testsuite/layers/pom.xml:L727](https://github.com/wildfly/wildfly/pull/15739/files#diff-05808b6024ac902fa0d4993b009f02f632e0d0005747fff8d094e3f547925036R727)

That clears up my doubt.

yersan (approved):

fjuma (approved):

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
bstansberry: @jfdenise @yersan Adding more layers in order to provision the default configs via layers should definitely decrease the diff. A number of the TODOs I added in LayersTestCase relate to missing layers.

If we use layers for the default configs, test-standalone-reference will still provision everything though, right? That is our standard dist should not be 'slimmed'. The LayersTestCase comparison of test-all-layers (slimmed) to test-standalone-reference (not slimmed) is really useful, even though it can easily fall out of date. It flagged up a bunch of things to investigate when I moved it away from being oriented toward servlet.

jfdenise: @bstansberry , that becomes a user choice, if user sets inherit-packages to false, then the default configuration would just contain the set of needed modules. In my test changes I set inherit-packs to false when generating the reference that is compared to the all-layers (for this kind of configuration).
The standard dist would not set inherit-packages to false, so dist would contain all.

PR #15746, opened 2022-07-06
WFLY-16594 Fix JGroups stack properties key in LocalDescriptions.properties

clara0:
https://issues.redhat.com/browse/WFLY-16594


darranl:
@clara0 how was this issue identified?  Management model testing us normally quite good at reporting missing strings.

clara0:
@darranl I notice that sometimes properties files will end up with duplicate keys, so I wrote a program to check for this issue and found this along the way.

darranl:
@clara0 for this one it is probably worth double checking it is even used as this may be an old string that can just be removed.

pferraro (approved):

PR #15755, opened 2022-07-08
[WFLY-16604] Upgrade Batavia to 1.0.15.Final. With this upgrade there…

jamezp:
… is no need to exclude transformed source files that are included in the source directory.

https://issues.redhat.com/browse/WFLY-16604


bstansberry (approved):

bstansberry:
Thanks @jamezp !

PR #15760, opened 2022-07-10
[WFLY-16624][WFLY-16634] Prevent HibernateSearchElasticsearch*TestCase failures if the environment isn't suitable

bstansberry:
https://issues.redhat.com/browse/WFLY-16624
https://issues.redhat.com/browse/WFLY-16634

Builds on #15759


bstansberry:
Yay! Finally works.

Note that the change to AssumeTestGroupUtil was because ElasticsearchServerSetupObserver uses it and ends up running in the server. Eager initalizing the EMPTY_xAR artifacts resulted in needing to include Shrinkwrap in the deployment which was silly.  I tried changing ElasticsearchServerSetupObserver to a ServerSetupTask but that resulted in different issues.

bstansberry:
I put the hold back on this as I'm going to incorporate https://issues.redhat.com/browse/WFLY-16634 into the same fix.

jamezp (approved):

yrodiere:
@bstansberry Well I certainly didn't expect those problems when I introduced these tests... Thanks for taking care of it!

PR #15761, opened 2022-07-10
[WFLY-16620] Move 'no-longer-Preview' tests from testsuite/preview/basic to testsuite/integration/basic

bstansberry:
https://issues.redhat.com/browse/WFLY-16620

Builds on the https://issues.redhat.com/browse/WFLY-16621 and https://issues.redhat.com/browse/WFLY-16624 fixes incorporated in https://github.com/wildfly/wildfly/pull/15760 against main.


ehsavoie (approved):

darranl:
@bstansberry I think the security manager failures look related?

bstansberry:
@yrodiere FYI.  This has the Hibernate Search test changes from #15760; the new thing here is the move of the tests out of testsuite/preview/basic into testsuite/integration/basic. That's part of of our work this week to move all of the testsuite testing of standard WF to use the 'current' (EE 10, etc) dependencies.

yrodiere:
@bstansberry Makes sense, thanks for the ping. Sorry I wasn't more responsive, I was offline a lot these past two weeks for various reasons. I'm back though, so don't hesitate if you need my help for other changes related to Hibernate Search!

PR #15763, opened 2022-07-10
[WFLY-16625] LayersTestCase should execute jpa-distributed-provisioni…

bstansberry:
…ng-test and all-layers-jpa-distributed-provisioning-full in the ts.ee9 profile

I also added a new, currently unused, property to use to enable executions that should only run with WildFly Preview

Also cleaned out some pom cruft.

https://issues.redhat.com/browse/WFLY-16625


bstansberry:
Hmm; doesn't work. It did locally but it was late so I might have dreamed it!

bstansberry:
Should be good now.

yersan (approved):
LGTM

PR #15769, opened 2022-07-12
[WFLY-16614]:Run testsuite/integration/legacy against WildFly Preview.

ehsavoie:
* Adding the testsuite integration legacy tests as part of the ee9 testsuite.

Jira: https://issues.redhat.com/browse/WFLY-16614


bstansberry:
@ehsavoie I don't think this causes the TS to run against WFP. It just causes it to be run against std WF even though -Dts.ee9 is set.

bstansberry:
@ehsavoie You'll need to change the target branch to big-bang, as https://issues.redhat.com/browse/WFLY-16499 has only been applied in that branch. That's why the test fails with no server=default in the config.

ehsavoie:
Setting PR against big-bang

PR #15767, opened 2022-07-12
[WFLY-16609] Convert testsuite/integration/domain to the jakarta.* na…

yersan:
…mespace

Jira issue: https://issues.redhat.com/browse/WFLY-16609

PR #15773, opened 2022-07-13
[WFLY-16609] Enable the testsuite/integration/domain for all relevant…

yersan:
… Jobs and not only for Wildfly Preview - Linux - JDK 17

Jira issue: https://issues.redhat.com/browse/WFLY-16609

It follows up on https://github.com/wildfly/wildfly/pull/15767

### DIFF
--- a/testsuite/domain/pom.xml
+++ b/testsuite/domain/pom.xml
@@ -50,9 +50,6 @@
 
         <jbossas.ts.dir>${basedir}/..</jbossas.ts.dir>
 
-        <!-- Disable the default surefire test execution. -->
-        <surefire.default-test.phase>none</surefire.default-test.phase>
-
         <!-- Use the WildFly Preview dependencies until standard WildFly moves to EE 10 -->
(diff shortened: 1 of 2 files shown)


yersan:
With the current changes, we are only running the tests for "WildFly Preview - Linux - JDK 17".

The wildfly preview feature pack is built always, so the goal here should be to run this testsuite on all the relevant Jobs by using the WildFly Preview. I think I got confused by the fact that I was assuming WildFly Preview was only being built when using `-Dts.ee9`

testsuite/domain/pom.xml
```diff
@@ -305,7 +302,7 @@
             <id>ee9.profile</id>
             <activation>
                 <property>
-                    <name>ts.ee9</name>
+                    <name>DISABLED.ts.ee9</name>
```
yersan, line 305: I'm also going to revert this since based on our CI configuration, by using ee9 we will be testing wildfly preview with JDK 17, so it makes sense to me to leave it by now.

bstansberry (approved):

bstansberry:
Thanks @yersan

PR #15629, opened 2022-06-08
WFLY-16452 Create a layer for mod_cluster subsystem

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-16452

PR #15780, opened 2022-07-14
WFLY-16452 Create a layer for mod_cluster subsystem

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-16452

Required by https://github.com/wildfly/wildfly/pull/15609

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/modcluster/StartWorkersInSuspendedModeTestCase.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/modcluster/StartWorkersInSuspendedModeTestCase.java
@@ -21,24 +21,6 @@
  */
 package org.jboss.as.test.clustering.cluster.modcluster;
 
-import org.jboss.arquillian.container.test.api.Deployment;
-import org.jboss.arquillian.container.test.api.RunAsClient;
-import org.jboss.arquillian.container.test.api.TargetsContainer;
-import org.jboss.arquillian.junit.Arquillian;
-import org.jboss.as.arquillian.api.ServerSetup;
-import org.jboss.as.controller.client.ModelControllerClient;
-import org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase;
-import org.jboss.as.test.shared.CLIServerSetupTask;
-import org.jboss.as.test.shared.ServerReload;
-import org.jboss.as.test.shared.TestSuiteEnvironment;
-import org.jboss.dmr.ModelNode;
-import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.asset.StringAsset;
-import org.jboss.shrinkwrap.api.spec.WebArchive;
-import org.junit.Assert;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-
 import static org.jboss.as.controller.client.helpers.ClientConstants.ADDRESS;
 import static org.jboss.as.controller.client.helpers.ClientConstants.INCLUDE_RUNTIME;
 import static org.jboss.as.controller.client.helpers.ClientConstants.OUTCOME;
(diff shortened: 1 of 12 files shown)


testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/modcluster/StartWorkersInSuspendedModeTestCase.java
```diff
@@ -109,17 +116,28 @@ private void assertWorkerNodeContextIsStopped() throws Exception {
...
-               .node(NODE_1)
-               .setup("/subsystem=modcluster/proxy=default:write-attribute(name=advertise,value=false)")
-               .setup("/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=proxy1:add(host=localhost,port=8590)")
-               .setup("/subsystem=modcluster/proxy=default:list-add(name=proxies,value=proxy1)")
-               .teardown("/subsystem=modcluster/proxy=default:list-remove(name=proxies,value=proxy1)")
-               .teardown("/socket-binding-group=standard-sockets/remote-destination-outbound-socket-binding=proxy1:remove()")
-               .teardown("/subsystem=modcluster/proxy=default:write-attribute(name=advertise,value=true)")
-            ;
+            super(createContainerSetConfigurationBuilder()
+                    .addContainer(NODE_1, createContainerConfigurationBuilder()
```
pferraro, line 122: There's a super constructor for a single container configuration, e.g.
super(NODE_1, createContainerConfigurationBuilder()...build());

rhusar:
Please review @bstansberry @pferraro @jfdenise, thanks!

I had to open a new PR, but to continue discussion from https://github.com/wildfly/wildfly/pull/15629#issuecomment-1160384237

> @rhusar , I am not sure I am the right person to do this review, pinging @bstansberry. Nevertheless I have some questions:

Thanks, pinged.

> * If a user provision the mod_cluster layer (that provision just the subset of JBoss modules needed for mod_cluster+web-server), then apply a CLI script to add ajp, is the set of provisioned JBoss modules enough or does ajp support requires more JBoss modules to be installed?

Good point – while there is a dependency on the integration module, specifically org.wildfly.mod_cluster.undertow, I didn't realize it's at the moment only an optional dependency. That is probably a remnant of when we had modules to support both JBoss Web and Undertow. The dependency in the context of galleon provisioning is now mandatory as the subsystem is useless without it; fixing that in the first commit. Thanks!

> * It seems that more default configurations would be impacted by this change.

Not sure exactly what you mean, but here's the situation that I need to have 2nd pair of eyes on: the non-ha profiles don't currently add mod_cluster subsystem and the available undertow listeners is `default` and `https` (even though there is an unused `ajp` socket binding...). The ha profiles include mod_cluster subsystem and the default listener is `ajp`. So they way this layer is setup, is that it defaults to `defalt` http listener, but in the `unsecured-basic-ha-profile.xml` where the `ajp` listener is added, the mod_cluster susbystem is reconfigured to use `ajp` as well. There doesn't seem to be a way to have the layer be dynamic, i.e. if `ajp` listener is defined, default to it, otherwise default to `default` listener. Also, I would like to avoid situation of having two separate layers like `mod_cluster-http` and `mod_cluster-ajp`. The current solution I think is plenty, but if people have opinions let me know! Thanks.

bstansberry:
> ... but here's the situation that I need to have 2nd pair of eyes on: the non-ha profiles don't currently add mod_cluster subsystem and the available undertow listeners is `default` and `https` (even though there is an unused `ajp` socket binding...). The ha profiles include mod_cluster subsystem and the default listener is `ajp`. So they way this layer is setup, is that it defaults to `defalt` http listener, but in the `unsecured-basic-ha-profile.xml` where the `ajp` listener is added, the mod_cluster susbystem is reconfigured to use `ajp` as well. There doesn't seem to be a way to have the layer be dynamic, i.e. if `ajp` listener is defined, default to it, otherwise default to `default` listener. Also, I would like to avoid situation of having two separate layers like `mod_cluster-http` and `mod_cluster-ajp`. The current solution I think is plenty, but if people have opinions let me know! Thanks.

@rhusar Besides wanting to deal with the fact that the non-ha profiles don't provision an ajp listener, is there any reason we would want a mod_cluster layer that uses the http listener?

We don't provide the mod_cluster subsystem in standalone.xml/standalone-full.xml, so there is no need to adapt the layer to meet the needs of those config files. We want to be able to provision standalone.xml/standalone-full.xml using layers, but mod_cluster would not be one of those layers, and if the user than added mod_cluster as an additional layer I don't see anything wrong with adjusting the config by adding the ajp listener. So the mod_cluster layer could include the addition of that listener.

People who don't want AJP are free to use the CLI etc to tailor the config to drop the listener and switch to http instead.

I don't think altering `unsecured-basic-ha-profile.xml` will let us reach the goal of being able to provision our standard standalone configs using layers.

bstansberry (approved):
LGTM!

pferraro (approved):

PR #15781, opened 2022-07-14
[WFLY-16639] Use the correct embedded broker config in WFP's standalo…

bstansberry:
…ne-full-ha.xml

Fixes an incorrect implementation of the temp restoration of the embedded broker to the WFP configs (WFLY-16499).

Also removes a messaging socket binding config from standalone.xml that isn't used when the embedded broker is in the config.

Also restores execution of ClusteredMessagingTestCase to ts.surefire.clustering.fullha, under the assumption it was disabled due to WFLY-16499 or my flawed impl that fix.

https://issues.redhat.com/browse/WFLY-16639


pferraro:
FYI, ClusteredJPA2LCTestCase failure is unrelated and addressed by https://github.com/wildfly/wildfly/pull/15782

pferraro (approved):

PR #15792, opened 2022-07-15
[WFLY-14313] Connector: More informative log after datasource test fa…

tadamski:
…ilure

https://issues.redhat.com/browse/WFLY-14313

After the fix the message contains a list of exception causes when there is a problem during connection establishment:

`
{
    "outcome" => "failed",
    "failure-description" => "WFLYJCA0040: failed to invoke operation: WFLYJCA0047: Connection is not valid javax.resource.ResourceException: IJ031084: Unable to create connection IJ031084: Unable to create conn
ection  FATAL: password authentication failed for user \"tomek\"",
    "rolled-back" => true
}
`

@hpehl could you please review?

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/common/pool/PoolOperations.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/common/pool/PoolOperations.java
@@ -100,7 +100,7 @@ public abstract class PoolOperations implements OperationStepHandler {
                             }
 
                         } catch (Exception e) {
-                            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToInvokeOperation(e.getLocalizedMessage()));
+                            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToInvokeOperation(concatenateExceptionCauses(e)));
                         }
                         if (operationResult != null) {
                             context.getResult().set(operationResult);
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/common/pool/PoolOperations.java
```diff
@@ -112,6 +112,17 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
         }
     }
 
+    private String concatenateExceptionCauses(Exception e) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(e.getMessage());
+        Throwable t = e.getCause();
+        while (t != null) {
+            sb.append(" ").append(t.getMessage());
```
bstansberry, line 120: @tadamski WDYT about using toString() here, and perhaps at L117? That adds the exception class name.

I've used getMessage() a lot over the years but then noticed when I looked at logs I didn't understand the problem because the message writer kind of assumed there'd be more info to provide context.

Also, perhaps instead of sb.append(" ") go with sb.append(" Caused by: ")?

jbliznak: originally `e.getLocalizedMessage()` was used, is it ok to change that to simpler `e.getMessage()` or `e.toString()`? IOW are we ok we are losing i18n here?

tadamski: @bstansberry Seems good to me, I have fixed the code to use toString method
@jbliznak Looking at the JDK code, it seems that to String is used getLocalizedMessage under the hood anyway so we shouldn't lose the localization

bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #15796, opened 2022-07-16
[WFLT-16600] Convert the secman testsuite to the jakarta.* namespace.

darranl:
https://issues.redhat.com/browse/WFLY-16600


bstansberry (approved):
I started the https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Removing.20-Dts.2Eee9.20functionality thread after seeing the 2nd commit here. But that shouldn't block merging this, as you used a separate commit for it so it's easy to revert if we want to.

PR #15800, opened 2022-07-18
WFLY-16613: Run testsuite/integration/rbac against WildFly Preview

moulalis:
Issue: https://issues.redhat.com/browse/WFLY-16613

### DIFF
--- a/testsuite/integration/rbac/pom.xml
+++ b/testsuite/integration/rbac/pom.xml
@@ -156,6 +156,29 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>ee9.profile</id>
+            <activation>
+                <property>
+                    <name>ts.ee9</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


jamezp:
I'm closing this as it's not quite what we're looking for.

jamezp:
I was wrong :) I guess this is what we were looking for.

testsuite/integration/rbac/pom.xml
```diff
@@ -156,6 +156,29 @@
...
+                            <!--Re-enable the default surefire execution. -->
+                            <execution>
+                                <id>default-test</id>
+                                <phase>test</phase>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
```
bstansberry, line 181: It turns out this bit wasn't needed, and is the reason this PR fails in CI. In other testsuite modules, usually the default-test surefire execution is wanted, but is turned off with -Dts.ee9 until someone proves it works and adds a profile like this. But in testsuite/integration/rbac we don't default-test to ever run. It runs a surefire execution with a different name.

Turns out the change to the other file was all that was needed to turn this on. That got picked up in #15812 so I'm going to close this one. Thanks for looking into it @moulalis!

PR #15804, opened 2022-07-18
[WFLY-16599] Convert the elytron-oidc-client testsuite to the jakarta* namespace.

darranl:
https://issues.redhat.com/browse/WFLY-16599

This testsuite does not contain any deployment code of it's own so only needed pointing to the Jakarta variant of shared.


fjuma (approved):
Thanks @darranl!

PR #15807, opened 2022-07-18
[WFLY-15685] Jakarta Concurrency 3.0 in WildFly Preview

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-15685
Important Note: There are multiple FIXMEs and TODOs, for instance to extract followups, or simply add new DUP phase values to WFLY Core, but I believe we should deal with those later, once all reviewers agree this is the way to go and everything else is good...


ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -1215,4 +1215,7 @@ public interface EeLogger extends BasicLogger {
             "specifications. The conflicting class is %s. Solutions include providing an alternate name for the component " +
             "or renaming the class.")
     void duplicateJndiBindingFound(String componentName, String jndiName, Class clazz);
+
+    @Message(id = 134, value = "Multiple usage of ContextServiceDefinition.ALL_REMAINING")
```
bstansberry, line 1219: s/usage/uses

bstansberry:
As for duplicating the ee subsystem source, I can live with that for now, but we need to follow up with a refactor of ee-concurrency into a separate maven module, so we can limit the duplication to those parts.

bstansberry:
BTW, I've been looking at the 3rd commit (the main one) but haven't finished.

emmartins:
> All the stuff in the transaction subsystem should be moved to the ee subsystem. Old stuff to ee, new to ee-9/source-transform/ee. And then there is no reason for duplicating the tx subsystem code.
> 
> The only reason the org.jboss.as.txn.ee.concurrency package is in the tx subsystem is to use TransactionManagerService.INTERNAL_SERVICE_NAME to inject a TransactionManager. That should not be necessary.
> 
> In the ee subsystem code:
> 
> 1. Check for capability "org.wildfly.transactions.global-default-local-provider". If not present the transaction subsystem is not present, and skip the rest.
> 2. Don't inject the TM. Just add a non-injection requirement for the "org.wildfly.transactions.global-default-local-provider" capability. This ensures the tm stuff is ready before your service.
> 3. Instead of using injection of the TM, directly use org.wildfly.transaction.client.ContextTransactionManager.getInstance(). That's what TransactionManagerService.getValue() provides, so doing it directly is the same as injecting it.
> 
> Note that the above is the way access to the TM is meant to be handled; it's not a hack to avoid forking. See comment above TransactionManagerService.SERVICE_NAME.

Hi @bstansberry , to avoid making this even more complex I created a separated JIRA and PR for this task at https://github.com/wildfly/wildfly/pull/15811

Once that PR is approved and merged I will rebase and update this PR.

PS: I believe you can still continue to review the 3rd commit on this one.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/deployers/EEConcurrentContextProcessor.java
```diff
@@ -103,7 +107,15 @@ private void setupConcurrentContext(ConcurrentContext concurrentContext, String
...
-
+        // add factories for deployment provided thread context providers
+        final ServiceLoader<ThreadContextProvider> threadContextProviderServiceLoader = ServiceLoader.load(ThreadContextProvider.class, moduleClassLoader);
+        final Iterator<ThreadContextProvider> threadContextProviderIterator = threadContextProviderServiceLoader.iterator();
+        int count = 0;
+        while (threadContextProviderIterator.hasNext()) {
+            final ThreadContextProvider threadContextProvider = threadContextProviderIterator.next();
+            concurrentContext.addFactory(new ThreadContextProviderContextHandleFactory(threadContextProvider, count));
+            count++;
+        }
```
jamezp, line 118: This iteration should likely be wrapped in a privileged action if the security manager is present. It looks like it uses the modules class loader which could be a deployment.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/handle/EE10ContextHandleFactory.java
```diff
@@ -0,0 +1,63 @@
...
+
+import jakarta.enterprise.concurrent.ContextService;
+
+import java.util.Map;
+
+/**
+ * The EE10 ContextHandleFactory, which should replace the legacy one once all impls are migrated.
+ * @author emmartins
+ */
+public interface EE10ContextHandleFactory extends ContextHandleFactory {
```
jamezp, line 28: We should come up with another name for this and avoid "EE10".

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ContextServiceDefinitionDescriptorProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+        final String name = metaData.getName();
+        if (name == null || name.isEmpty()) {
+            throw ROOT_LOGGER.elementAttributeMissing("<context-service>", "name");
+        }
+        final ContextServiceTypesConfiguration contextServiceTypesConfiguration = new ContextServiceTypesConfiguration.Builder()
+                .setCleared(metaData.getCleared())
+                .setPropagated(metaData.getPropagated())
+                .setUnchanged(metaData.getUnchanged())
+                .build();
+        // TODO properties
```
jamezp, line 59: What does this mean?

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ManagedExecutorDefinitionInjectionSource.java
```diff
@@ -0,0 +1,215 @@
...
+
+        try {
+            // install the resource service
+            final ServiceName resourceServiceName = ManagedExecutorServiceResourceDefinition.CAPABILITY.getCapabilityServiceName(resourceName);
+            final ServiceBuilder resourceServiceBuilder = phaseContext.getServiceTarget().addService(resourceServiceName);
+            final Supplier<ManagedExecutorHungTasksPeriodicTerminationService> hungTasksPeriodicTerminationService = resourceServiceBuilder.requires(ConcurrentServiceNames.HUNG_TASK_PERIODIC_TERMINATION_SERVICE_NAME);
+            final ManagedExecutorServiceService resourceService = new ManagedExecutorServiceService(resourceName, resourceJndiName, hungTaskThreshold, hungTaskTerminationPeriod, longRunningTasks, maxAsync, maxAsync, keepAliveTime, keepAliveTimeUnit, threadLifeTime, queueLength, rejectPolicy, threadPriority, hungTasksPeriodicTerminationService);
+            resourceServiceBuilder.setInstance(resourceService);
+            final String contextServiceRef = this.contextServiceRef == null || this.contextServiceRef.isEmpty() ? "java:comp/DefaultContextService" : this.contextServiceRef;
```
jamezp, line 85: This could be right as I'm not real familiar with how this works in WildFly. However, the `ManagedExecutorDefinition.context()` default seems like it should be used. It does seem to be "java:comp/DefaultContextService", but it just seems a bit fragile to hard-code it.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ManagedScheduledExecutorDefinitionInjectionSource.java
```diff
@@ -0,0 +1,206 @@
...
+
+        try {
+            // install the resource service
+            final ServiceName resourceServiceName = ManagedScheduledExecutorServiceResourceDefinition.CAPABILITY.getCapabilityServiceName(resourceName);
+            final ServiceBuilder resourceServiceBuilder = phaseContext.getServiceTarget().addService(resourceServiceName);
+            final Supplier<ManagedExecutorHungTasksPeriodicTerminationService> hungTasksPeriodicTerminationService = resourceServiceBuilder.requires(ConcurrentServiceNames.HUNG_TASK_PERIODIC_TERMINATION_SERVICE_NAME);
+            final ManagedScheduledExecutorServiceService resourceService = new ManagedScheduledExecutorServiceService(resourceName, resourceJndiName, hungTaskThreshold, hungTaskTerminationPeriod, longRunningTasks, maxAsync, keepAliveTime, keepAliveTimeUnit, threadLifeTime, rejectPolicy, threadPriority, hungTasksPeriodicTerminationService);
+            resourceServiceBuilder.setInstance(resourceService);
+            final String contextServiceRef = this.contextServiceRef == null || this.contextServiceRef.isEmpty() ? "java:comp/DefaultContextService" : this.contextServiceRef;
```
jamezp, line 84: Same comment here as the previous one.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ManagedThreadFactoryDefinitionInjectionSource.java
```diff
@@ -0,0 +1,114 @@
...
+    public void getResourceValue(final ResolutionContext context, final ServiceBuilder<?> serviceBuilder, final DeploymentPhaseContext phaseContext, final Injector<ManagedReferenceFactory> injector) throws DeploymentUnitProcessingException {
+        final String resourceName = uniqueName(context);
+        final String resourceJndiName = "java:jboss/ee/concurrency/definition/managedScheduledExecutor/"+resourceName;
+        try {
+            // install the resource service
+            final ServiceName resourceServiceName = ManagedThreadFactoryResourceDefinition.CAPABILITY.getCapabilityServiceName(resourceName);
+            final ServiceBuilder resourceServiceBuilder = phaseContext.getServiceTarget().addService(resourceServiceName);
+            final ManagedThreadFactoryService resourceService = new ManagedThreadFactoryService(resourceName, resourceJndiName, priority);
+            resourceServiceBuilder.setInstance(resourceService);
+            final String contextServiceRef = this.contextServiceRef == null || this.contextServiceRef.isEmpty() ? "java:comp/DefaultContextService" : this.contextServiceRef;
```
jamezp, line 61: Same here too :)

jamezp (commented):
This is just a partial review of the "[adds support for the new major features of EE Concurrency 3.0:](https://github.com/wildfly/wildfly/pull/15807/commits/bba111c81b8a06578330e93baeef3e9003d19326)". I will do the others later today.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ContextServiceDefinitionDescriptorProcessor.java
emmartins: XML elements have generic properties, I will need to followup on this wrt propagating such properties to the injectionsource and beyond (if makes any sense)

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ManagedExecutorDefinitionInjectionSource.java
emmartins: Not sure I am following, you would prefer to grab the JNDI address for the platform's default ContextService from somewhere else?

jamezp: Sorry. I just mean the `ManagedExecutorDefinition` annotation has defaults and it just feels a little cleaner to use the default rather than the hard-coded version. The hard-coded string does match the default so it's not critical, but if the default were to ever change we could see potential issues.

jamezp:
> As for duplicating the ee subsystem source, I can live with that for now, but we need to follow up with a refactor of ee-concurrency into a separate maven module, so we can limit the duplication to those parts.

My concern here is that we would have to fix everything twice. Once in `ee` and once in `ee-9/source-transform/ee`.

bstansberry:
@emmartins https://issues.redhat.com/browse/WFLY-16657 is merged / resolved. Thanks!

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
emmartins: fixed

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/resource/definition/ManagedExecutorDefinitionInjectionSource.java
emmartins: Changed to the const we had in the deployer responsible for binding it.

emmartins:
@bstansberry @jamezp PR reworked on top of [WFLY-16657](https://issues.redhat.com/browse/WFLY-16657), and also addressing your previous concerns. I also left a few "TODO *FOLLOW UP*", identifying the needed followups (on new JIRAs), if this is approved as it is.

ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/concurrent/ContextServiceImpl.java
```diff
@@ -22,33 +22,34 @@
...
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Map;
 
-import org.glassfish.enterprise.concurrent.spi.ContextSetupProvider;
-import org.glassfish.enterprise.concurrent.spi.TransactionSetupProvider;
-import org.wildfly.security.manager.WildFlySecurityManager;
+import static org.jboss.as.ee.logging.EeLogger.ROOT_LOGGER;
+import static org.wildfly.common.Assert.checkArrayBounds;
+import static org.wildfly.common.Assert.checkNotNullParam;
```
bstansberry, line 35: I'm not asking you to fix this, but the old order is correct per our codestyle:

import static

import java
import javax

import other

bstansberry (approved):
@emmartins Looks good. Please add the JIRA to the commit messages. If you push and the only changes are to the commit messages, please note that in a PR comment in case we don't want to wait for CI.

emmartins:
@bstansberry JIRA key added to all commits, no other changes done

PR #15811, opened 2022-07-19
[WFLY-16657] moves EE Concurrency Transaction Resources to EE Subsystem

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-16657


emmartins:
Please note, all was moved except the TransactionSetupProvider service, which I fully removed, that existed only to plug the TransactionSetupProvider on EE Concurrency services.

emmartins:
retest this please

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
```diff
@@ -206,17 +206,21 @@ public interface EeLogger extends BasicLogger {
...
-    //void failedToSuspendTransaction(@Cause Throwable cause);
+    @LogMessage(level = WARN)
+    @Message(id = 17, value = "Failed to suspend transaction.")
+    void failedToSuspendTransaction(@Cause Throwable cause);
 
-    //@Message(id = 18, value = "System error while checking for transaction leak in EE Concurrent invocation.")
-    //void systemErrorWhileCheckingForTransactionLeak(@Cause Throwable cause);
+    @LogMessage(level = WARN)
+    @Message(id = 18, value = "System error while checking for transaction leak in EE Concurrent invocation.")
+    void systemErrorWhileCheckingForTransactionLeak(@Cause Throwable cause);
```
jamezp, line 223: I'm not sure on the solution here, but the logged ID will be different which we've tried to avoid in the past. This is just more of an FYI than anything.

servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/ee/main/module.xml
```diff
@@ -44,6 +44,7 @@
         <module name="javax.el.api"/>
         <module name="javax.interceptor.api"/>
         <module name="javax.security.jacc.api"/>
+        <module name="javax.transaction.api"/>
```
jamezp, line 47: Should this also be optional? Not sure it can, but just curious.

bstansberry: It should be optional. The need for it is guarded by a capability check:

https://github.com/wildfly/wildfly/pull/15811/files/53fcd7d2a7a2b16922e240ce26042d26a7a22caf#diff-e73f8c6aa955e801fe07ee3f2e04df349d0b7b7dc3548ad5fe26a3806dd50f95R64

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
emmartins: Those messages were created and used by the factory we are moving back to EE subsystem so I assume we can reactive these again?

servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/ee/main/module.xml
emmartins: fixed

emmartins:
Squashed RTS-jakarta dependency changes by @jamezp, thank you.

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
bstansberry: 👍

bstansberry (approved):

ee/src/main/java/org/jboss/as/ee/logging/EeLogger.java
jamezp: Yes, they can definitely be enabled. This was just more of a general comment as the idea of the ID is to not change them so they can be tracked to the message. That said with the Wildscribe sutff I guess this is less of an issue.

jamezp (approved):

bstansberry:
Thanks @emmartins!

PR #15820, opened 2022-07-20
Only include the testsuite/test-feature-pack-preview module if -Dskip…

jamezp:
….preview is not set.


bstansberry:
Thanks @jamezp

PR #15821, opened 2022-07-20
[WFLY-16660] Further isolate legacy dependencies from current ones

bstansberry:
https://issues.redhat.com/browse/WFLY-16660


pom.xml
```diff
@@ -665,3920 +650,530 @@
...
-                <artifactId>xercesImpl</artifactId>
-                <version>${version.org.apache.xerces}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>xml-apis</groupId>
-                        <artifactId>xml-apis</artifactId>
-                    </exclusion>
-                </exclusions>
+                <groupId>net.bytebuddy</groupId>
+                <artifactId>byte-buddy</artifactId>
```
yersan, line 1099: This is not a test dependency, why it is left in the main pom?

pom.xml
```diff
@@ -665,3920 +650,530 @@
...
             </dependency>
 
             <dependency>
-                <groupId>org.awaitility</groupId>
-                <artifactId>awaitility</artifactId>
-                <version>${version.org.awaitility.awaitility}</version>
-                <scope>test</scope>
+                <groupId>org.hamcrest</groupId>
+                <artifactId>hamcrest-all</artifactId>
+                <version>${version.org.hamcrest}</version>
```
yersan, line 1170: Missing test scope?... there are some of them with the same missing scope

pom.xml
```diff
@@ -4619,13 +1232,48 @@
...
             </dependency>
 
+            <!--
+            TODO used both in testsuite/integration/legacy-ejb-client and in legacy code.
+            Once legacy code use is gone, convert to <scope>test</scope>
+            -->
+            <dependency>
+                <groupId>org.jboss.narayana.jts</groupId>
+                <artifactId>narayana-jts-idlj</artifactId>
+                <version>${version.org.jboss.narayana}</version>
```
yersan, line 1262: Is this undefined on the relevant bom because it is a server and test dependency?

pom.xml
bstansberry: Thanks, that should be moved.

pom.xml
bstansberry: Thanks I'll add it to this one.

Some of the others are used for compiling non src/test/java code. Some of the ARQ-related ones, I'm not sure if they are or not but in at least 1 case I added the test scope and the build broke, so I stopped doing that. I'm sure it can be added to some though, if someone has time to determine which.

Note that testsuite/shared has its code in src/main/java so any seemingly 'test' dep it has can't have test scope. That was the ARQ problem I hit.

yersan: > Note that testsuite/shared has its code in src/main/java 

ahh, ok, I missed that detail

pom.xml
bstansberry: Fixed.

pom.xml
bstansberry: I fixed the hamcrest-all one.

pom.xml
```diff
@@ -665,3920 +650,525 @@
...
             </dependency>
+
             <dependency>
-                <groupId>${ee.maven.groupId}</groupId>
-                <artifactId>wildfly-clustering-web-container</artifactId>
-                <version>${ee.maven.version}</version>
+                <groupId>${full.maven.groupId}</groupId>
+                <artifactId>wildfly-feature-pack-galleon-content</artifactId>
+                <version>${full.maven.version}</version>
+                <type>zip/</type>
```
yersan, line 1003: There are some of these, notice additional / in `zip/`.

I've just checked these errors are also in main, which means these dependencies are not being used at all

yersan (approved):
Found minor detail declaring pom types, but since it passed before, it should be fine and can be addressed later. LGTM

pom.xml
bstansberry: We should clean those up. But not in this PR. ;)

I don't think they are used. The use case for them is the way WF Core produces them and WF uses the zip. For these local ones we just directly use the local source though. If we every moved the FP build to a separate project they'd be useful.

PR #15828, opened 2022-07-21
WFLY-16666 Do not use component class as superclass for local home interface proxy

chengfang:
https://issues.redhat.com/browse/WFLY-16666


tadamski (approved):
LGTM

PR #15829, opened 2022-07-21
WFLY-16671 Avoid string conversion of cache keys when determining segment

pferraro:
https://issues.redhat.com/browse/WFLY-16671


rhusar (approved):
LGTM

PR #15833, opened 2022-07-21
[WFLY-16652] Move standard WildFly build to EE 10

bstansberry:
https://issues.redhat.com/browse/WFLY-16652
https://issues.redhat.com/browse/WFLY-16651

Builds on #15821 so also resolves https://issues.redhat.com/browse/WFLY-16660

High level overview of what this does:

1) Creates a new 'boms' maven module tree and moves the dependencyManagement stuff out of the root pom and the various xxx-feature-pack-parent poms. See https://issues.redhat.com/browse/WFLY-16660 for an explanation of the various boms.

2) Updates various modules to use the bom that's appropriate for their purpose. For example, batch-jberet produces an artifact not used in the server now, and needs 'legacy' (i.e. javax.* deps) so it uses the wildfly-legacy-ee-bom. OTOH ee-9/source-transform/batch-jberet uses wildfly-standard-ee-bom (which has jakarta.* deps.)

The tightening up of the boms from 1) and 2) surfaced a number of maven dependency issues (e.g. a module not using -jakarta artifact when it should) which I cleaned up.

3) Move the ee-9/ee-9-api module to ee-feature-pack/ee-10-api. Switch the wildfly-ee feature pack to use that instead of ee-feature-pack/ee-8-api

4) Move all but a tiny bit of the contents of ee-9/common and ee-9/common-microprofile to the analogous source in ee-feature-pack/* and galleon-pack. Mostly this amounts to updating the module.xml files to reflect what WFP had been using.

5) Switch the testsuite back to testing std WF by default instead of WFP.

6) Misc cleanups, cruft code deletion etc that I found along the way.


bstansberry:
There are 5 failures I'm seeing plus some known intermittents and maybe one other that I haven't investigated.

HostExcludesTestCase has a simple fix.
ReactiveMessagingChannelsTestCase has 4 failures. Kabir is helping me diagnose what's going on.

yersan:
About the HostExcludesTestCase, we would need to recover this: https://github.com/wildfly/wildfly/pull/15773/files#diff-d4696c308808a57f331264e1d20639f0ef8a33968635726f1d64a3ecb14e93d5L208-L218

jamezp (approved):
This looks good to me.

bstansberry:
I squashed the HostExcludesTestCase fix into https://github.com/wildfly/wildfly/pull/15833/commits/2593db89f4ba84d9916a8abe3008309d551e4c3d as that was the commit that broke the test.

boms/legacy-ee/pom.xml
```diff
@@ -0,0 +1,5533 @@
...
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>27.0.0.Alpha4-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>wildfly-legacy-ee-bom</artifactId>
+    <description>
+		Bill of materials for maven modules that provide traditional capabilities like Jakarta EE
+		and that compile against legacy dependencies
```
darranl, line 44: By legacy do we mean pre-EE9?

bstansberry: Yes. Pre-EE9 and since overridden (as opposed to other stuff that we used pre-EE9 and still use.)

The term is a bit overloaded, as in the source we also use 'legacy' for old subsystems where we now only have admin-only shells for mixed domain.

I have a vague idea that after we clean these boms out by converting the normal source to jakarta.*, the concept may still be useful at some point, in a sense more akin to the mixed domain usage. That is we may still support old versions of stuff (say an EE 10 variant of some spec if we decided to move our standard config to some later variant of that spec.) But we don't need to do that, or we could just drop the 'legacy' boms after we clean them out and then resurrect them later if there is a concrete need.

darranl:
The two test failures are in the same test module, however doing an allTests build locally it all passed for me without issue.

PR #15835, opened 2022-07-22
WFLY-16669: Upgrade HAL to 3.6.2.Final

hpehl:
https://issues.redhat.com/browse/WFLY-16669

Contains only one fix compared to HAL 3.6.1.Final:
- [HAL-1800](https://issues.redhat.com/browse/HAL-1800): Provide backward compatibility after migration from master/slave → primary/secondary


bstansberry:
Thanks @hpehl

PR #15840, opened 2022-07-23
[WFLY-16694] Instead of skipping a test-feature-pack-preview build if…

bstansberry:
… not needed, only build it if needed

https://issues.redhat.com/browse/WFLY-16694


bstansberry:
I converted this to draft as it will break big-bang, since it currently uses WFP in testing outside the ts.ee9 or ts.bootable.ee9 context. I'll merge it into my WFLY-16652 branch, which removes that usage.

I'm leaving it open just to let CI do its thing since it already started.

bstansberry:
This will come into main via big-bang, so I'm closing this.

PR #15857, opened 2022-07-26
[WFLY-16684] Add legacy bom dep mgmt entries for javax.* stuff curren…

bstansberry:
…tly defined in the WF Core bom

https://issues.redhat.com/browse/WFLY-16684


bstansberry:
@yersan FYI. As you work on core if you find anything missing, this PR illustrates the approach to adding the needed deps in full.

PR #15863, opened 2022-07-29
[WFLY-16718] Override the version of all the Jakarta package namespac…

yersan:
…e Elytron dependencies brought in by wildfly-core

Jira issue: https://issues.redhat.com/browse/WFLY-16718

We have not overridden all the Elytron dependencies. That makes the `org.wildfly.security.elytron-base` module have a mix of 1.x and 2.x dependencies.

This will be a temporal solution until we drive Elytron 2.x completely from wildfly-core.

Uses the following PRs:
#15861
#15868

### DIFF
--- a/boms/standard-ee/pom.xml
+++ b/boms/standard-ee/pom.xml
@@ -40,7 +40,7 @@
 
     <artifactId>wildfly-standard-ee-bom</artifactId>
     <description>
-		Build of materials that can be used to build standard WildFly maven modules 
+		Build of materials that can be used to build standard WildFly maven modules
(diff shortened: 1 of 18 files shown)


bstansberry:
Thanks @yersan. I don't know if this conflicts with https://github.com/wildfly/wildfly/pull/15857 / https://github.com/wildfly/wildfly/pull/15858 but I recommend getting those merged first. I'm not available to fix any issues with those. Plus it seems like this should be tested on top of those.

bstansberry (approved):

bstansberry:
Thanks @yersan!

PR #15865, opened 2022-07-30
WFLY-16514 Explicit transaction-timeout of a previous transaction fro…

chengfang:
…m the same thread used instead of default transaction-timeout

https://issues.redhat.com/browse/WFLY-16514

See https://github.com/wildfly/wildfly/pull/15866 for the same fix in main branch.


bstansberry (approved):

bstansberry:
Thanks @chengfang

PR #15866, opened 2022-07-30
WFLY-16514 Explicit transaction-timeout of a previous transaction fro…

chengfang:
…m the same thread used instead of default transaction-timeout

https://issues.redhat.com/browse/WFLY-16514

See https://github.com/wildfly/wildfly/pull/15865 for the same fix in 26.x branch.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionTimeoutLeakServlet.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionTimeoutLeakServlet.java
@@ -1,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2022, Red Hat Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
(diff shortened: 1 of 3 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/transaction/TransactionTimeoutLeakServlet.java
```diff
@@ -0,0 +1,63 @@
...
+                final int timeoutValue = Integer.parseInt(second);
+                if (timeoutValue > 0) {
+                    tm.setTransactionTimeout(timeoutValue);
+                }
+            }
+
+            XAResource xaer = new TestXAResource();
+            tm.begin();
+            tm.getTransaction().enlistResource(xaer);
+            int effectiveTimeout = xaer.getTransactionTimeout();
```
mmusgrov, line 54: This is a cunning way to test the actual timeout being set by the system, nice.

mmusgrov (approved):
Looks good @chengfang thanks very much for taking the issue - I didn't know about org.jboss.as.server.deployment.SetupAction so this has been enlightening for me.

PR #15870, opened 2022-08-01
[WFLY-16677] Turn off builds of legacy javax.* production modules; re…

bstansberry:
…move dependencies upon them.

Clean out additional boms/legacy-xx dependencyManagement entries for items no longer used once these builds are disabled.

Note that this leaves likely unused legacy.version.xx properties behind in the root pom; cleaning those up would be follow-up work.

Note also that completion of #15863 will allow for removal of further legacy entries.


clustering/infinispan/extension/pom.xml
```diff
@@ -86,10 +104,6 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>wildfly-clustering-server-service</artifactId>
         </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>wildfly-transactions</artifactId>
-        </dependency>
         <dependency>
             <groupId>org.wildfly.transaction</groupId>
             <artifactId>wildfly-transaction-client</artifactId>
```
yersan, line 109: There is a -jakarta variant for wildfly-transaction-client, I guess we could use it here instead.

yersan (approved):
Independent of my previous comment, giving a symbolic approval. We can refine any pending  dependencies later, all CI are in green, so it should not prevent merging this

clustering/infinispan/extension/pom.xml
bstansberry: @yersan Thanks. I picked that up in my v2 version of this -- #15871.

bstansberry:
Thanks for the review, @yersan. I'm going to close this one as there are now 3 further variants derived from this, all of which would be used instead of this one.

PR #15871, opened 2022-08-02
[WFLY-16677][WFLY-16733] Version 2 of build cleanups

bstansberry:
Turn off builds of legacy javax.* production modules; remove dependencies upon them.

Clean out additional boms/legacy-xx dependencyManagement entries for items no longer used once these builds are disabled.

More aggressive variant of #15870. It differs from that in starting with a fix for WFLY-16733 which in turn allows dropping most of the EE 8 API artifacts from boms/legacy-ee. It also drops the various org.wildfly.security artifacts from the legacy boms as thinking about it more I don't see the need to keep them.


clustering/el/pom.xml
```diff
@@ -45,7 +45,8 @@
     </description>
 
     <modules>
-        <module>glassfish</module>
+		<!-- TODO uncomment when converted to jakarta.*
+        <module>glassfish</module>-->
```
pferraro, line 49: FYI, the glassfish module was only used for javax.el.* support.  jakarta.el.* support is exclusive to the expressly module.
I prepare a PR to this effect to today.

clustering/infinispan/embedded/spi/pom.xml
```diff
@@ -46,8 +46,8 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.jboss.spec.javax.annotation</groupId>
-            <artifactId>jboss-annotations-api_1.3_spec</artifactId>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
```
pferraro, line 50: This will break.  Infinispan 13's protostream module requires the javax.annotation package during build time.

pferraro (changes requested):

clustering/el/pom.xml
bstansberry: Yeah; I just used the same code comment everywhere. Partly to make it easier to git grep for them later.

pferraro: Gotcha.

bstansberry:
Thanks @pferraro for the review.  I've put the 'hold' label on this one for now, as the impact of development (i.e. IDE behavior) of dropping all these modules and deps from the build system needs further exploration. #15873 and #15875 preserve those builds. Once we have more time to think about it, it would be easy to follow up #15873 with a commit that turns of building them by default.

bstansberry:
Closing as the #15873 alternative has been merged.

PR #15872, opened 2022-08-02
[WFLY-16737] Upgrade H2 to 2.1.210

bstansberry:
This is a variant of #15072 but with the commits related to turning on MODE=LEGACY removed.

My preference is to go with #15072, but since this PR variant already exists I'm leaving it open in case the consensus is to go with this one.

https://issues.redhat.com/browse/WFLY-16737


bstansberry:
@darranl I left the Critical label on this so it stays on the radar screen, but put the 'hold' label on it as my preference is to go with #15072. That said, #15072 shares the same commit as this one, so merging this doesn't preclude merging the other later if we want to think about it more.

PR #15873, opened 2022-08-02
[WFLY-16677][WFLY-16733] Fix spec-api module; move building 'ee 8' modules to a profile

bstansberry:
This is a less aggressive approach to WFLY-16677 vs #15871. This does not disable building any modules; it just moves the ee 8 ones into a profile, which is enabled by default but could be disabled using -Dskip.ee8

This approach preserves the dependency info for use in IDEs.

Note that the last commit that sets up the profile could be omitted, in which case the PR amounts to some code cleanup mostly to not rely upon the ee 8 modules as dependencies in other modules.


microprofile/reactive-streams-operators-smallrye/cdi-provider/pom.xml
```diff
@@ -40,6 +40,18 @@
     <artifactId>wildfly-microprofile-reactive-streams-operators-cdi-provider</artifactId>
     <name>WildFly: MicroProfile Reactive Streams Operators CDI Provider</name>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>wildfly-legacy-ee-bom</artifactId>
```
yersan, line 47: Shouldn't this be wildfly-legacy-expansion-bom instead?

yersan: In any case, since PR Jobs are in green, it means this module doesn't need any expansion dependencies, so it should not prevent merging this, so LGTM

yersan (approved):

darranl (approved):

darranl:
I am labelling this as ready to merge, it preserves IDE debugging / code completion etc... Importantly however it does allow the build to speed up by skipping the redundant compilation steps.

darranl:
@bstansberry once merged do we want to add this option to a subset of CI jobs so just a few do the full build.

bstansberry:
> @bstansberry once merged do we want to add this option to a subset of CI jobs so just a few do the full build.

Yes, for sure. I think at most 1 PR job should do the full build, and maybe 1 nightly.

PR #15875, opened 2022-08-02
[WFLY-16677][WFLY-16733] Fix spec-api module; move building 'ee 8' modules to a profile

bstansberry:
This is the same as #15873 but without the final commit that sets up a profile. This isn't my preference but if we decide we don't want the profile, or at least not yet, this does the most important code cleanup parts.


microprofile/reactive-streams-operators-smallrye/cdi-provider/pom.xml
```diff
@@ -40,6 +40,18 @@
     <artifactId>wildfly-microprofile-reactive-streams-operators-cdi-provider</artifactId>
     <name>WildFly: MicroProfile Reactive Streams Operators CDI Provider</name>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>wildfly-legacy-ee-bom</artifactId>
```
yersan, line 47: As a minor doubt, I think this should be wildfly-legacy-expansion-bom instead

yersan (approved):
LGTM, I guess the one with the profile could compile faster if the developer needs it, otherwise, I cannot see more advantages about one instead the other

bstansberry:
I put the 'hold' label on this and removed 'Critical' as #15873 has the 'ready-for-merge' label. So no point tracking 2 PRs. This will auto-close when #15873 is merged.

PR #15882, opened 2022-08-03
Upgrade legacy jastow from 2.0.10.Final to 2.0.11.Final

ivassile:
Issue: https://issues.redhat.com/browse/WFLY-16745

Upgrade legacy jastow from 2.0.10.Final to 2.0.11.Final
Tag: https://github.com/undertow-io/jastow/releases/tag/2.0.11.Final
Diff: https://github.com/undertow-io/jastow/compare/2.0.10.Final...2.0.11.Final
Hash: a37ef6b8eb79550144f33cc8a2e6008f4b129661


bstansberry:
I'm confident this will conflict with #15881, which is critical. So this should wait until that is in.

These legacy version upgrades aren't really necessary as we'll never produce a server using the artifacts that compile against these releases.

Lerm:
Probably target for this pull request should by branch '26.x' instead of 'main' -- so it could get to 26.1.2 (if it would be released). WildFly since 25.0.0 has a serious regression with JSP compilation (UNDERTOW-204) and this PR fix it by updating Jastow version. Main (27.x) version is now updated to Jastow 2.2.2 (which includes the same fix as 2.0.11 for 2.0.x version) and this PR fixes this regression in 26.1.x branch.

bstansberry:
Thank you for the input, @Lerm -- I added the 'evalFor26.1' to https://issues.redhat.com/browse/WFLY-16745 to ensure that we consider this upgrade for inclusion in WildFly 26.1.2.

Lerm:
> Thank you for the input, @Lerm -- I added the 'evalFor26.1' to https://issues.redhat.com/browse/WFLY-16745 to ensure that we consider this upgrade for inclusion in WildFly 26.1.2.

@bstansberry, I've created a separate PR for 26.x branch for such update as https://github.com/wildfly/wildfly/pull/15943 - it may be useful for 26.x integration (as current PR is targeted to 'main' branch instead of 26.x). Please disregard this new PR if this change is already in work and I've missed something.

bstansberry:
Thank you @Lerm -- that PR is much appreciated. I labeled WFLY-16745 to track getting the update into 26.1.2 but we hadn't gotten around to doing it yet.

PR #15892, opened 2022-08-05
WFLY-16752 Make it possible for Hibernate ORM Dialects to be able to access org.postgresql.jdbc and com.oracle.ojdbc JDBC driver classes

scottmarlow:
https://issues.redhat.com/browse/WFLY-16752


ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
```diff
@@ -56,5 +56,7 @@
         <module name="net.bytebuddy"/>
         <module name="java.xml"/>
         <module name="com.h2database.h2" optional="true"/>
+        <module name="com.oracle.ojdbc" optional="true"/>
```
bstansberry, line 59: A comment here would be good, e.g.

```
<!-- If the Oracle JDBC driver classes are accessible, Hibernate will use them to tailor OracleDialect behavior -->
```

I assume that statement is accurate; if not we should adapt it. It's deliberately vague to help avoid it falling out of date.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
```diff
@@ -56,5 +56,7 @@
...
+        <module name="org.postgresql.jdbc" optional="true"/>
```
bstansberry, line 60:
```
<!-- If the Postgres JDBC driver classes are accessible, Hibernate will use them to tailor PostgresDialect and CockroachDialect behavior -->
```

bstansberry (changes requested):

bstansberry:
@jfdenise Please review. I'm mostly interested in your thoughts on whether doing this at all is a good idea, in case you think not!

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
scottmarlow: :+1: for leaving a comment there as that also gives a hint to future maintainers as to why it is there.

bstansberry: Thanks. And yes, maintainability is the main purpose I had in mind. Any optional module dependency begs the question of why it's optional and what conditions would make it required.

jfdenise (approved):

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #15895, opened 2022-08-05
[WFLY-16696] Update okhttp to newer version

jasondlee:
Version update

https://issues.redhat.com/browse/WFLY-16696

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -313,7 +313,8 @@
         <version.com.ibm.async.asyncutil>0.1.0</version.com.ibm.async.asyncutil>
         <version.com.microsoft.azure>8.6.6</version.com.microsoft.azure>
         <version.com.nimbus.jose-jwt>9.23</version.com.nimbus.jose-jwt>
-        <version.com.squareup.okhttp3>3.14.9</version.com.squareup.okhttp3>
+        <legacy.version.com.squareup.okhttp>3.14.9</legacy.version.com.squareup.okhttp>
+        <version.com.squareup.okhttp>4.10.0</version.com.squareup.okhttp>
         <version.com.squareup.okio>1.17.5</version.com.squareup.okio>
         <legacy.version.com.sun.activation.jakarta.activation>1.2.2</legacy.version.com.sun.activation.jakarta.activation>
         <legacy.version.com.sun.faces>2.3.17.SP01</legacy.version.com.sun.faces>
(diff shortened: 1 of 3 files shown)


bstansberry:
Looks like this is incompatible with Infinispan.

pferraro:
N.B. okhttp is required by infinispan's rest client to infinispan-server, which is only used in our integration tests.  This is not a runtime dependency of WildFly.  This means we can keep the testsuite/integration/clustering dependency at 3.x, while allowing the rest of WF to use 4.x.  Our long term plan is to switch to a containerized infinispan-server for the testsuite which will eliminate the need for okhttp in the testsuite (which is only used to trigger shutdown of infinispan-server).

jasondlee:
> Looks like this is incompatible with Infinispan.

I was thinking we had found a conflict months ago when we tried a component upgrade, but I couldn't remember where, and local tests all passed. Thank you, Jenkins. :)

I'll ping you tomorrow and we can discuss what to do here.

pom.xml
```diff
@@ -313,7 +313,8 @@
         <version.com.ibm.async.asyncutil>0.1.0</version.com.ibm.async.asyncutil>
         <version.com.microsoft.azure>8.6.6</version.com.microsoft.azure>
         <version.com.nimbus.jose-jwt>9.23</version.com.nimbus.jose-jwt>
-        <version.com.squareup.okhttp3>3.14.9</version.com.squareup.okhttp3>
+        <legacy.version.com.squareup.okhttp3>3.14.9</legacy.version.com.squareup.okhttp3>
+        <version.com.squareup.okhttp3>4.9.2</version.com.squareup.okhttp3>
```
pferraro, line 317: Can we rename this property to "version.com.squareup.okhttp"?

pferraro (changes requested):

pom.xml
jasondlee: That shouldn't be a problem, but that means "okhttp" points to v3 and "okhttp3" points to v4, which is kinda confusing. Not sure why that property was named that way originally, but, when in Rome... :)

jasondlee: (Deleting earlier comment)

Can I read this mean "okhttp3" -> "okhttp", and leaving the "legacy" variant (though renamed as well)?

pferraro: I would use "version.com.squareup.okhttp" and "legacy.version.com.squareup.okhttp"

jasondlee: OK. In progress now...

pferraro (approved):

rhusar (approved):
LGTM (n.b. we might be able to get rid of this workaround once infinispan upgrades or we migrate to using containerized testing against infinispan-server).

PR #15665, opened 2022-06-16
[WFLY-16440] Upgrade WildFly HTTP Client to 1.1.12.Final

fl4via:
Signed-off-by: Flavia Rainone <frainone@redhat.com>

Jira: https://issues.redhat.com/browse/WFLY-16440
Main PR: #15664 

        Release Notes - WildFly EJB HTTP Client - Version 1.1.12.Final
                                                        
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-73'>WEJBHTTP-73</a>] -         WildflyClientInputStream.read() and close() fail to throw ioException after waiting on lock
</li>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-80'>WEJBHTTP-80</a>] -         Digest authentication creates URI with port -1 when using standard ports
</li>
</ul>
        
<h2>        Task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-75'>WEJBHTTP-75</a>] -         Problematic Language usage deprecation and replacement
</li>
</ul>
                                                                                                            
<h2>        Component Upgrade
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-82'>WEJBHTTP-82</a>] -         Upgrade jakarta.ejb-api from 4.0.0 to 4.0.1
</li>
</ul>
    
<h2>        Enhancement
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-76'>WEJBHTTP-76</a>] -         Remove unused class org.wildfly.httpclient.ejb.ModuleIdentifier
</li>
</ul>

PR #15908, opened 2022-08-10
[WFLY-16741] Upgrade WildFly HTTP Client to 1.1.13.Final

fl4via:
Jira: https://issues.redhat.com/browse/WFLY-16741

Signed-off-by: Flavia Rainone <frainone@redhat.com>


        Release Notes - WildFly EJB HTTP Client - Version 1.1.13.Final
                                                        
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WEJBHTTP-83'>WEJBHTTP-83</a>] -         WF Http EJB Client does not delete ContextData that has been removed on the server side
</li>
</ul>

Main PR: #15909

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,7 +500,7 @@
         <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
         <version.org.wildfly.core>18.1.1.Final</version.org.wildfly.core>
         <version.org.wildfly.extras.creaper>1.6.2</version.org.wildfly.extras.creaper>
-        <version.org.wildfly.http-client>1.1.11.Final</version.org.wildfly.http-client>
+        <version.org.wildfly.http-client>1.1.13.Final</version.org.wildfly.http-client>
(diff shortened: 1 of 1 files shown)


bstansberry (approved):

bstansberry:
@fl4via #15665 wasn't merged basically because we weren't spending energy on 26.x for most of the summer. Both it and this are ok with me if you feel they are safe for the last micro we will do for our SE 8 / EE 8 server.

However, it seems there is some problem with this update. The major failures on the SE 8 CI jobs are recurring.

I was planning to ignore #15665 in favor of this, but given those SE 8 CI failures, I went ahead and did a retest on the PR (to get more up-to-date results) and unless you say otherwise I'll merge that one if it passes. This one would then need to be updated.

bstansberry:
@fl4via I think we should stick with 1.1.13. This didn't pass CI, and I'd prefer to make no further changes to 26.x if possible so we can tag Wednesday.

PR #15912, opened 2022-08-11
WFLY-16786 Upgrade smallrye-open-api to 3.0.0-CR3

pferraro:
https://issues.redhat.com/browse/WFLY-16786


testsuite/integration/microprofile-tck/openapi/pom.xml
```diff
@@ -172,7 +172,7 @@
             <id>bootablejar.ee9.profile</id>
             <activation>
                 <property>
-                    <name>ts.bootable.ee9</name>
+                    <name>DISABLED.ts.bootable.ee9</name>
```
bstansberry, line 175: Why disable these? We still want to test WFP, which I expect will start having greater differences from std WF soon.

We disabled some of these during the big bang but that seemed more of a case of getting rid of one moving part while everything was in flux.

bstansberry: Same question applies re the 3rd commit.

pferraro: Good point.  These changes had been in my branch since before the big-bang, but they are no longer necessary.  I'll update accordingly.

ee-9/source-transform/microprofile/openapi-smallrye/pom.xml
```diff
@@ -1,235 +0,0 @@
...
-        <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-microprofile-jakarta</artifactId>
-        <!--
-        Maintain separation between the artifact id and the version to help prevent
-        merge conflicts between commits changing the GA and those changing the V.
-        -->
-        <version>27.0.0.Alpha5-SNAPSHOT</version>
-    </parent>
-
-    <artifactId>wildfly-microprofile-openapi-smallrye-jakarta</artifactId>
```
rhusar, line 36: With the `wildfly-microprofile-openapi-smallrye-jakarta` artifact being gone, perhaps the license should be removed as well https://github.com/pferraro/wildfly/blob/openapi/ee-9/common-microprofile/src/main/resources/license/preview-feature-pack-common-microprofile-licenses.xml#L37-L47

rhusar (changes requested):

ee-9/source-transform/microprofile/openapi-smallrye/pom.xml
pferraro: Oops - missed that one.

bstansberry:
Thanks, @pferraro

PR #15913, opened 2022-08-11
[WFLY-16787] Do not use jboss-invocation and jboss-interceptors-api_1…

yersan:
….2_spec dependencies that come from WildFly Core

Jira issue: https://issues.redhat.com/browse/WFLY-16787


bstansberry (approved):

bstansberry:
Thanks, @yersan

PR #15914, opened 2022-08-11
[WFLY-16788] Move org.wildfly.security:wildfly-elytron-jaspi from tes…

yersan:
…tsuite/integration to the legacy modules

Jira issue: https://issues.redhat.com/browse/WFLY-16788

### DIFF
--- a/testsuite/integration/pom.xml
+++ b/testsuite/integration/pom.xml
@@ -535,10 +535,6 @@
         </dependency>
         <dependency>
             <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron-jaspi</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron-permission</artifactId>
         </dependency>
         <dependency>


testsuite/integration/legacy-ejb-client/pom.xml
```diff
@@ -64,6 +64,11 @@
             <artifactId>jakarta.interceptor-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-jaspi</artifactId>
+            <scope>test</scope>
+        </dependency>
```
bstansberry, line 71: Please remove. My testing verifies this dep is not needed.

testsuite/integration/legacy/pom.xml
```diff
@@ -53,6 +53,11 @@
             <artifactId>jboss-jms-api_2.0_spec</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-jaspi</artifactId>
+            <scope>test</scope>
+        </dependency>
```
bstansberry, line 60: Please remove. My testing verifies this dep is not needed.

bstansberry (changes requested):
@yersan Looks like all that's needed is removal of the dep from testsuite/integration/pom.xml; the child modules don't need it.

Except... I think it can be removed from boms/legacy-ee as well, since the only use of it is in testsuite/integration/pom.xml.

yersan:
Thanks for the review @bstansberry , I've removed both dependencies

bstansberry (approved):

bstansberry:
Thanks @yersan

PR #15837, opened 2022-07-22
WFLY-16607 Application deployment fails with EJB (WFLYEJB0275: Could …

chengfang:
…not resolve corresponding ejbCreate or @Init method for home interface)

https://issues.redhat.com/browse/WFLY-16607

Port the fix from main branch https://github.com/wildfly/wildfly/pull/15756

PR #15751, opened 2022-07-07
[26.x] WFLY-16448 Duplicate key in LocalDescriptions.properties file in iiop-openjdk

clara0:
Same issue as https://github.com/wildfly/wildfly/pull/15611
https://issues.redhat.com/browse/WFLY-16448

PR #15747, opened 2022-07-06
[WFLY-16595] backport webservices changes replacing UsersRolesLoginMo…

rsearls:
…dule

https://issues.redhat.com/browse/WFLY-16595 backport webservices 
      (security) replacement changes for UsersRolesLoginModule

PR #15583, opened 2022-05-26
WFLY-16414 Upgrade jberet-core (javax.batch impl) from 1.3.12.Final t…

chengfang:
…o 1.3.13.Final

https://issues.redhat.com/browse/WFLY-16414

PR #15918, opened 2022-08-14
[26.x] Pulls

bstansberry:
#15583 
#15595 
#15655 
#15747 
#15751 
#15837 
#15865


bstansberry:
All have been merged but #15747, which still has open discussion.

PR #15920, opened 2022-08-15
Remove duplicate dependency on jboss-invocation causing build warnings.

rhusar:
Introduced over the weekend via https://github.com/wildfly/wildfly/pull/15913 (FYI @yersan)

### DIFF
--- a/ee/pom.xml
+++ b/ee/pom.xml
@@ -127,10 +127,6 @@
             <artifactId>wildfly-request-controller</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss.invocation</groupId>
-            <artifactId>jboss-invocation</artifactId>
-        </dependency>
-        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>wildfly-naming</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

yersan (changes requested):
Thanks, @rhusar, I came into adding the dependencies when I was testing building wildly full with https://github.com/wildfly/wildfly-core/pull/5171 and I added those dependencies thinking that they were not there.

It seems the order matters here, so, could you remove the other duplicate on [L129-L132](https://github.com/wildfly/wildfly/blob/main/ee/pom.xml#L129-L132)

The same for the ejb3 module.

rhusar:
@yersan Good point, agreed, and done. Although – the `ejb3/pom.xml` dependencies are *very* out of order.

yersan:
Ok, my interpretation of the order issue is that `org.jboss.invocation:jboss-invocation-jakarta` is brought in as transitive dependency by `org.wildfly.core:wildfly-server`, so it will place common classes before in the classpath, making the compilation to fail when used in conjunction with https://github.com/wildfly/wildfly-core/pull/5171

Both solutions should work, either exclude `org.jboss.invocation:jboss-invocation-jakarta` from `org.wildfly.core:wildfly-server` or place `org.jboss.invocation:jboss-invocation` at the beginning, as you did now.

yersan (approved):

bstansberry:
Thanks @rhusar

PR #15953, opened 2022-08-18
[WFLY-16847] Upgrade Eclipse Yasson to 3.0.1

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16847

PR #15954, opened 2022-08-18
[26.x][WFLY-16847] Upgrade Eclipse Yasson to 1.0.11/2.0.4

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16847

Upstream https://github.com/wildfly/wildfly/pull/15953


boris-unckel:
Please find all links to release notes, git tags, github-diff in the Jira.

jamezp (approved):
Even those these contain an enhancement for records, I think it makes sense to do these. Without it there is no support for records.

bstansberry (approved):

PR #15956, opened 2022-08-18
[WFLY-16836] Drop source transformation of the EE security subsystem and remove the Jakarta EE 8 variant.

darranl:
https://issues.redhat.com/browse/WFLY-16836

Hopefully this has captured all of the moving pieces to move from the Jakarta EE 8 variant.

This does also include the commit from https://github.com/wildfly/wildfly/pull/15950 as otherwise this sequence of changes will become hard to coordinate.


ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/glassfish/soteria/main/module.xml
```diff
@@ -28,6 +28,7 @@
     </properties>
 
     <resources>
+        <artifact name="${org.glassfish.soteria:soteria.spi.bean.decorator.weld}"/>
         <artifact name="${org.glassfish.soteria:jakarta.security.enterprise}"/>
```
rhusar, line 32: I would keep these ordered alphabetically.

ee-security/pom.xml
```diff
@@ -54,11 +54,15 @@
...
 
+        <!-- Jakarta-namespace specific deps -->
+
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>wildfly-ee</artifactId>
+            <artifactId>wildfly-ee-jakarta</artifactId>
         </dependency>
 
+        <!-- Other deps consistent with the javax.* ee-security module -->
```
rhusar, line 64: This comment is kinda out of date now ;-)

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/glassfish/soteria/main/module.xml
darranl: @rhusar The reason I put this first was to ensure this decorator was picked up first.

ee-security/pom.xml
darranl: Removed

PR #15957, opened 2022-08-18
WFLY-16253 Undertow subsystem model for WF27 should be 12.0.0, not 13.0.0.

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-16253

TLDR
* WF 26 is using model version 11 = https://github.com/wildfly/wildfly/blob/26.0.0.Final/undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtension.java#L82
* So WF 27 should be using version 12 (and not 13.)

Issue introduced in https://github.com/wildfly/wildfly/commit/a1f9ad78f6c960e824a2ca007a40593706c2d060#diff-b60dca0e646d43bf2678414de075f93e39a0550ab0ad7564b774e8e8615f5340R82


fl4via (approved):

bstansberry:
Thanks, @rhusar!

PR #15961, opened 2022-08-19
[WFLY-16854] Update dependencies for Opentelemetry

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16854

### DIFF
--- a/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
+++ b/testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
@@ -114,6 +114,10 @@ public class LayersTestCase {
         "org.wildfly.security.jakarta.client.resteasy",
         "org.wildfly.security.jakarta.client.webservices",
         "org.jboss.resteasy.microprofile.config",
+        // TODO see gRPC proposal
+        "io.grpc",
+        // TODO analysis
+        "com.google.protobuf",
     };
     // Packages that are not referenced from the module graph but needed.
     // This is the expected set of un-referenced modules found when scanning
@@ -129,14 +133,12 @@ public class LayersTestCase {
         "org.wildfly.extension.microprofile.reactive-streams-operators-smallrye",
         "org.wildfly.reactive.mutiny.reactive-streams-operators.cdi-provider",
         "io.jaegertracing",
-        "io.grpc",
         "io.smallrye.health",
         "io.smallrye.openapi",
         "io.vertx.client",
         "org.eclipse.microprofile.health.api",
         "org.eclipse.microprofile.openapi.api",
         "com.fasterxml.jackson.dataformat.jackson-dataformat-yaml",
-        "com.google.protobuf",
         "org.jboss.resteasy.resteasy-client-microprofile",
         "io.netty.netty-codec-dns",
         "io.netty.netty-codec-http2",
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -462,6 +462,12 @@
                 <version>${version.com.fasterxml.jackson}</version>
             </dependency>
 
+	    <dependency>
+                <groupId>com.fasterxml.jackson.jr</groupId>
+                <artifactId>jackson-jr-objects</artifactId>
+                <version>${version.com.fasterxml.jackson.jr.jackson-jr-objects}</version>
+            </dependency>
+
             <dependency>
                 <groupId>com.fasterxml.woodstox</groupId>
                 <artifactId>woodstox-core</artifactId>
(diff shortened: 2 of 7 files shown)


boris-unckel:
Please find all links to release notes, git tags, github-diff in the Jira.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/com/fasterxml/jackson/jr/jackson-jr-objects/module.xml
```diff
@@ -0,0 +1,33 @@
...
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<module xmlns="urn:jboss:module:1.9" name="com.fasterxml.jackson.jr.jackson-jr-objects">
```
darranl, line 24: For any new module being added a decision needs to be made as to if we consider it public or private API.

However (this is an open question, not a request to make a change yet) - as this module is only depended on by a single module flagged as being private API I wonder if for now the artefact should be contained within the only module that actually makes use of it?

This may be more of a historic thing as well, we do tend to default to adding new modules but where there is only ever a 1:1 relationship do we really need the proliferation of modules.

darranl (changes requested):

ee-feature-pack/common/src/main/resources/modules/system/layers/base/com/fasterxml/jackson/jr/jackson-jr-objects/module.xml
jasondlee: I tend to agree: If this is added ONLY for consumption by one module, I'd just add it as an artifact there and not create Yet Another Module until it's needed. If the dependency goes away, with the set up in this PR, we're likely going to end up with an orphaned module that someone will eventually find and clean up. ;)

boris-unckel: @darranl Thanks for your feedback. Looking at other potential dependencies it's only jackson.jr itself aside from opentelemetry. Integrating it into "io.opentelemetry.trace" is a good solution. I'm waiting for Jason's feedback and provide a consolidated PR afterwards.

jasondlee (changes requested):

boris-unckel:
The errors in the CI are same as local, my first thought was they were unrelated. I have to check.

boris-unckel:
@darranl @jasondlee I have updated the PR addressing the module issue. The okhttp library has updated transitive dependencies, too - Kotlin. 
My local test was done with "-DallTest" but org.jboss.as.test.layers.LayersTestCase did not pop up, neither OK nor failed. What do you recommend to include such tests locally, preferred in a single run with the current -DallTest set?
Please have a look at the current version of the PR, I'm going to provide an update afterwards together with the LayersTestCase fix.

jasondlee:
> @darranl @jasondlee I have updated the PR addressing the module issue. The okhttp library has updated transitive dependencies, too - Kotlin. 

@boris-unckel, are those hard Kotlin dependencies? The library won't work without them?

boris-unckel:
> @boris-unckel, are those hard Kotlin dependencies? The library won't work without them?

@jasondlee Yes, when running the tests ClassNotFoundExceptions raised, this was the point to introduce them.

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
```diff
@@ -114,6 +114,10 @@ public class LayersTestCase {
         "org.wildfly.security.jakarta.client.resteasy",
         "org.wildfly.security.jakarta.client.webservices",
         "org.jboss.resteasy.microprofile.config",
+        // TODO see gRPC proposal
+        "io.grpc",
```
bstansberry, line 118: @hpehl FYI about this PR.

It removes the need for io.grpc from WF, but for WF 27 Beta I'll leave the module in place to not break your grpc feature pack. But for WF 27 Final I'd like to remove it and your FP would need to install it.

bstansberry:
I put the 'hold' label on this because there are 3 PRs open in this same general topic area and I'm working to create a combo PR. #16109 fixes the merge conflict in this, but I don't expect to merge that independently.

testsuite/layers/src/test/java/org/jboss/as/test/layers/LayersTestCase.java
hpehl: @bstansberry thanks for letting me know!

PR #15962, opened 2022-08-19
WFLY-16734 jboss-client.jar not working with Wildfly 26.1.1

chengfang:
https://issues.redhat.com/browse/WFLY-16734

It was reported in the above jira that jboss-client.jar worked in 26.0.0 but failed in 26.1.1 for this user's application. In 26.0.0, jboss-client.jar includes everything in wildfly-elytron.jar, which is an uber jar including many of its dependencies and submodules. In 26.1.1, some components like jboss-ejb-client and wildfly-naming-client split their dependency on wildfly-elytron into its more modular individual submodules. Therefore, when building jboss-client.jar, these individual elytron-* submodules are bundled in jboss-client.jar, which omits some artifacts unreachable via transitive dependency.

In order to maintain some kind of compatibility with 26.0.0, this PR proposes to add some additional elytron-* submodules to jboss-client.jar. Even though those additional artifacts are not required at each component client level (e.g., jboss-ejb-client, or wildfly-naming-client), they are needed to support a wide range of non-maven-based applications with various security mechanisms.

The following artifacts are not added to jboss-client.jar in this PR, as they appear not applicable on the client side, but I may be wrong:

* elytron-audit
* elytron-jacc
* elytron-oidc
* wildfly-elytron-jose-util
* wildfly-elytron-jose-jwk
* elytron-\*-deprecated
* elytron-\*-http-\*


fjuma (commented):
@chengfang Thanks for working on this! To make sure that we aren't accidentally missing anything here, I'm wondering if for WildFly 26.1.2, it would be safer to just add a dependency on `org.wildfly.security:wildfly-elytron` to mimic the behaviour that was present in WildFly 26.0.0.Final? What do you think? That would preserve compatibility. 

We can then working on fixing this properly in the main branch. As a starting point, I've added some comments about dependencies that I think can be removed.

bstansberry:
@chengfang @fjuma I pushed up https://github.com/wildfly/wildfly/pull/15972 in case that's the way you decide to go.

I'd like to be able to merge a fix for this, either #15972 or an approved version of this, by the end of the morning US time tomorrow (Aug 23).

bstansberry:
@chengfang We went ahead with #15972. If something delays 26.1.2 and you prefer a more fine-grained approach like this, please reopen and we can discuss.

chengfang:
@bstansberry @fjuma  using wildfly-elytron dependency looks good. Thanks for getting this done for 26.x.

PR #15958, opened 2022-08-18
[WFLY-16851] Upgrade groovy-all to 4.0.4

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-16851

PR #15965, opened 2022-08-22
WFLY-16862 Hibernate Search engine module should have optional dependencies to other modules

yrodiere:
https://issues.redhat.com/browse/WFLY-16862

I don't think this warrants a dedicated integration test since we're fixing a very exotic use case here.

I also noticed a module had a name that was not aligned with the corresponding Java module, so I took the liberty to fix that. That module was introduced in WF 27, so there's no backward incompatibility to fear.

### DIFF
--- a/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
+++ b/ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
@@ -33,5 +33,11 @@
 
     <dependencies>
         <module name="org.jboss.logging"/>
+
+        <!-- Optional dependencies to import services (used for resolution of bean references in particular). -->
+        <module name="org.hibernate.search.backend.lucene" optional="true" services="import"/>
+        <module name="org.hibernate.search.backend.elasticsearch" optional="true" services="import"/>
+        <module name="org.hibernate.search.mapper.pojo-base" optional="true" services="import"/>
+        <module name="org.hibernate.search.mapper.orm" optional="true" services="import"/>
     </dependencies>
 </module>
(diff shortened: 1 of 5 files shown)


bstansberry:
This is one where we'll need to be sure the Galleon behavior is what we want when a slimmed server is provisioned.

The analogous deps were optional in 26 and earlier as well so there's nothing new here; my comment here is more along the lines of this PR being a good time to re-check the Galleon stuff.

yrodiere:
As far as I can tell, there are failing tests in the "Galleon Linux - JDK 11" build, but those look like unrelated, flaky tests. At least two of them is also failing in other PRs that were merged: https://github.com/wildfly/wildfly/pull/15968 , https://github.com/wildfly/wildfly/pull/15958

> my comment here is more along the lines of this PR being a good time to re-check the Galleon stuff.

Was that meant for me specifically, or a more general comment? If I'm supposed to add a Galleon-specific tests, I'm not sure where. Could you please point me to existing tests that provision a WF server using Galleon?

yrodiere:
> > my comment here is more along the lines of this PR being a good time to re-check the Galleon stuff.
> 
> Was that meant for me specifically, or a more general comment? If I'm supposed to add a Galleon-specific tests, I'm not sure where. Could you please point me to existing tests that provision a WF server using Galleon?

Hey @bstansberry, I played around a bit with `LayersTestCase` (which I assume is "Galleon stuff"?). I can confirm that Hibernate Search modules (and their dependencies) do *not* get provisioned automatically by Galleon, probably because no other module depends on them.

We have some automation in `org.jboss.as.jpa.processor.HibernateSearchProcessor` to add dependencies to Hibernate Search when configuration properties related to Hibernate Search are detected; however, I'm not sure that's taken into account by Galleon. And in any case, such configuration properties are probably not set in `LayersTestCase`. So, users wanting Hibernate Search in their Galleon-provisioned servers will probably have to add an explicit dependency to Hibernate Search (both `org.hibernate.search.mapper.orm` and `org.hibernate.search.backend.lucene`/`org.hibernate.search.backend.elasticsearch`) to their deployment.

I updated `LayersTestCase` to reflect that. As far as I can tell, the problem is not new.

Does that address your concerns? Can we merge this PR and address the problem as part of a separate JIRA issue, since it seems unrelated to my changes?

bstansberry (approved):

bstansberry:
Thanks @yrodiere!

bstansberry:
@yrodiere The JIRA: https://issues.redhat.com/browse/WFLY-17004

yrodiere:
Thanks @bstansberry!

PR #15972, opened 2022-08-23
[WFLY-16734] Add the wildfly-elytron dependency to ensure all classes…

bstansberry:
… present in WF 26.0.0 are included

https://issues.redhat.com/browse/WFLY-16734

This implements the possible approach discussed at https://github.com/wildfly/wildfly/pull/15962#pullrequestreview-1080891670. I want to get CI for this so it can be merged tomorrow if @chengfang / @fjuma decide to use this approach.


fjuma (approved):
Thanks @bstansberry! It looks like @chengfang is on PTO today. I think we should go with this approach for WildFly 26.1.2.Final.

PR #15977, opened 2022-08-24
WFLY-16874 Mark Lucene, Elasticsearch REST client, GSON as public modules

yrodiere:
https://issues.redhat.com/browse/WFLY-16874


yrodiere:
Hey @bstansberry , do you think we can merge this? That's what we discussed some time ago, about some modules being marked as private in WildFly that really should be public, because they include useful features, even though the features probably cannot be supported by vendors (see related PRs in EAP).

bstansberry (approved):

bstansberry:
Thanks @yrodiere!

PR #15978, opened 2022-08-24
WFLY-14419 Add tests for naming subsystem for attributes that allow expression

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-14419

### DIFF
--- a/naming/src/test/java/org/jboss/as/naming/subsystem/NamingSubsystemTestCase.java
+++ b/naming/src/test/java/org/jboss/as/naming/subsystem/NamingSubsystemTestCase.java
@@ -21,6 +21,7 @@
 */
 package org.jboss.as.naming.subsystem;
 
+import static org.junit.Assert.assertEquals;
 import java.io.IOException;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


naming/src/test/java/org/jboss/as/naming/subsystem/NamingSubsystemTestCase.java
```diff
@@ -123,6 +125,34 @@ public void testCompositeBindingUpdate() throws Exception {
...
     }
 
+    @Test
+    public void testExpressionInAttributeValue() throws Exception{
+        final String subsystemXml = readResource("subsystem_expression.xml");
+        final KernelServices ks = createKernelServicesBuilder(createAdditionalInitialization()).setSubsystemXml(subsystemXml).build();
+        final ModelNode naming = ks.readWholeModel().get("subsystem", getMainSubsystemName());
+         ModelNode attribute = naming.get(NamingSubsystemModel.BINDING).get("java:global/a");
+
+        final String value = attribute.get(NamingSubsystemModel.VALUE).resolve().asString();
```
bstansberry, line 135: Calling ModelNode.resolve is more of a test of JBoss DMR, which is tested there.

To confirm that the subsystem code is handling the expressions correctly, instead create ModelNode that represent a /subsystem=naming:read-resource(recursive=true,resolve-expressions=true) operation and use KernelService.executeOperation to run it. The result should have resolved values and you can directly check them.

bstansberry (changes requested):

bstansberry (commented):
Minor point about import order aside, this looks good.

bstansberry (approved):

PR #15999, opened 2022-09-01
[WFLY-10912] CodecSessionConfig#findSessionId() causes an incorrect JSESSIONID Set-Cookie header

petrberan:
https://issues.redhat.com/browse/WFLY-10912


bstansberry:
@petrberan This breaks some clustering tests. @pferraro FYI

petrberan:
Thanks @bstansberry Noticed that already on Fri, managed to fix some over the weekend, but others are failing instead. Will look into that

pferraro (changes requested):
As mentioned in the associated jira, it is not the responsibility of the SessionConfig implementations to validate the incoming session ID.  WFLY-10912 must be fixed within Undertow itself, which should never include a JSESSIONID cookie in the response unless the request references a valid session.

pferraro:
Superseded by https://github.com/wildfly/wildfly/pull/16019

PR #16001, opened 2022-09-02
WFLY-16935: add EnterpiseDB to recongized dialects, throw exception i…

michpetrov:
…f dialect cannot be determined

Issue: [WFLY-16935](https://issues.redhat.com/browse/WFLY-16935)

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -143,6 +143,7 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service<Datab
     private static final String MSSQL = "mssql";
     private static final String SYBASE = "sybase";
     private static final String JCONNECT = "jconnect";
+    private static final String ENTERPRISEDB = "enterprisedb";
 
     /** Names for the different SQL commands stored in the properties*/
     private static final String CREATE_TABLE = "create-table";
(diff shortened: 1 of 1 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -306,6 +309,10 @@ private void investigateDialect() {
             database = identifyDialect(database);
             EjbLogger.EJB3_TIMER_LOGGER.debugf("New Database dialect is '%s'.", database);
         }
+
+        if (database == null) {
+            throw EjbLogger.EJB3_TIMER_LOGGER.failedToFindDialect();
+        }
     }
```
chengfang, line 316: With a null database value, the timer persistence should still be able to function. So it should not throw exception here. Logging a warning like in line 303 should suffice. We can move line 302-304 to here to handle this case for all cases.

chengfang: Around line 245, we can add a null check to prevent NPE from `database` field:

```java
  if(database != null) {
        switch (database) {
            case DB2:
                adjustCreateAutoTimerStatement("FROM SYSIBM.SysDummy1 ");
            break;
            case ORACLE:
                adjustCreateAutoTimerStatement("FROM DUAL ");
            break;
        }
  }
```

michpetrov: I see, I figured it wouldn't work correctly if e.g. the `adjustCreateAutoTimerStatement` wasn't called. I've refactored the PR.

chengfang (approved):

bstansberry:
Thanks @michpetrov

PR #16005, opened 2022-09-03
[WFLY-16939] Update Ironjacamar to 1.5.8.Final

vsalbaba:
https://issues.redhat.com/browse/WFLY-16939

Upgrade to Ironjacamar 1.5.8.Final which includes fix for mysql 8 Reauth plugin [JBJCA-1450](https://issues.redhat.com/browse/JBJCA-1450)


tadamski (approved):
@vsalbaba If the IJ fix managed to deal with MySQL 8 validation bug then I agree we should do the upgrade

bstansberry:
Thanks @vsalbaba

PR #16007, opened 2022-09-05
[26.x] WFLY-16417 Add MP certification page generation script

xstefank:
https://issues.redhat.com/browse/WFLY-16417


xstefank:
@bstansberry we need this in 26.x in case we will have more MP 4.1 certifications.

PR #16010, opened 2022-09-05
[WFLY-16460] Workaround MP JWT TCK regression caused by AES default key size change since JDK19

ropalka:
https://issues.redhat.com/browse/WFLY-16460

### DIFF
--- a/testsuite/integration/microprofile-tck/jwt/pom.xml
+++ b/testsuite/integration/microprofile-tck/jwt/pom.xml
@@ -213,6 +213,7 @@
                     <systemPropertyVariables>
                         <microprofile.jvm.args>${microprofile.jvm.args}</microprofile.jvm.args>
                     </systemPropertyVariables>
+                    <argLine>-Djdk.security.defaultKeySize=AES:128</argLine>
                 </configuration>
             </plugin>
         </plugins>


testsuite/integration/microprofile-tck/jwt/pom.xml
```diff
@@ -213,6 +213,7 @@
                     <systemPropertyVariables>
                         <microprofile.jvm.args>${microprofile.jvm.args}</microprofile.jvm.args>
                     </systemPropertyVariables>
+                    <argLine>-Djdk.security.defaultKeySize=AES:128</argLine>
```
bstansberry, line 216: Should this be done in a profile?

ropalka: You mean something like JDK19 specific profile @bstansberry  ?
IMO it is not necessary. The default AES key size changed from 128 to 256
in JDK19 and this is just a workaround for testsuite until it will be properly fixed in JWT TCK.

darranl:
For this issue do we know the original motivation?  I am assuming they are referring to this:
    https://apps.nsa.gov/iaarchive/programs/iad-initiatives/cnsa-suite.cfm

Has any investigation been undertaken yet to trigger the step "until it will be properly fixed in MP JWT"?  In this context are you referring to the specification, the TCK or the SmallRye JWT implementation or possibly our own integration?

ropalka:
It makes sense it might be driven by that NSA requirement.
I was referring to Ricardo's comment in https://issues.redhat.com/browse/WFLY-16460 @darranl .

fjuma (approved):
@ropalka The workaround looks ok but do we have an issue created to track removing the workaround once there's a version of `microprofile-jwt-auth-tck` with the fix?

ropalka:
You are correct @fjuma . Here's the follow up JIRA: https://issues.redhat.com/browse/WFLY-16975

fjuma:
Thanks @ropalka!

bstansberry:
Thanks @ropalka

PR #16023, opened 2022-09-07
Ensure we allow any ExpressionFactoryWrappers to execute before JspAp…

bstansberry:
…plicationContext provides any ELContext


bstansberry:
This is superseded by #16024, which is the same but just has the JIRA in the commit message.

I'm leaving this open for now just to let CI complete.

jamezp:
Other PR merged.

PR #16027, opened 2022-09-07
WFLY-16960 Change single quote to double quote in docs cli sample com…

chengfang:
…mands

https://issues.redhat.com/browse/WFLY-16960


bstansberry:
Thanks @chengfang

shipsing:
@chengfang, don't we need to use double quotes for database and partition names as both the parameters represent a string? Asking this question for better understanding about the usage of quotes.

Thanks

chengfang:
@shipsing CLI attribute types (string, number, etc) are automatically handled without need for quoting. Double quote can be used to escape some special characters, or when it's no certain.

PR #16035, opened 2022-09-08
[WFLY-16969][WFLY-16970][WFLY-16972] MSC Deprecated API Usages Cleanup

ropalka:
https://issues.redhat.com/browse/WFLY-16969
https://issues.redhat.com/browse/WFLY-16970
https://issues.redhat.com/browse/WFLY-16972

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
+++ b/naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
@@ -21,17 +21,17 @@
  */
 package org.jboss.as.naming;
 
-import org.jboss.msc.value.Value;
+import java.util.function.Supplier;
 
 /**
- * A ManagedReference that simply holds a value'
+ * A ManagedReference that simply holds a value
  *
  * @author Stuart Douglas
  */
 public class ValueManagedReference implements ManagedReference {
-    private final Value<Object> value;
+    private final Supplier<Object> value;
 
-    public ValueManagedReference(Value<Object> value) {
+    public ValueManagedReference(final Supplier<Object> value) {
         this.value = value;
     }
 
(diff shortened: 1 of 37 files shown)


naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
```diff
@@ -21,17 +21,17 @@
...
- * A ManagedReference that simply holds a value'
+ * A ManagedReference that simply holds a value
  *
  * @author Stuart Douglas
  */
 public class ValueManagedReference implements ManagedReference {
-    private final Value<Object> value;
+    private final Supplier<Object> value;
 
-    public ValueManagedReference(Value<Object> value) {
```
bstansberry, line 34: This constructor should be retained and deprecated as external subsystems may be using it. The number of internal subsystems that use it shows that that is a possibility.

I think there should also be a simple 'ValueManagedReference(Object value)' constructor that pretty much all of the uses touched in this PR call instead of creating a Supplier lambda. For those cases where this PR passes in '() -> theValue' that eliminates a bunch of lambdas with just a single one in this class. And for cases where this PR passes in '() -> foo.doSomething()' it restores the current semantic where 'foo.doSomething()' is callled once, vs the PRs semantic where it will be called every time getInstance() is called.

bstansberry (changes requested):
I've left a comment re the ValueManagedReference part. 

I think the DiscoveryService part should be a separate PR as someone else should review it.

naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
ropalka: Thank you @bstansberry for very valuable feedback. Splitting into multiple PRs.

ropalka:
Superseded by:
 * https://github.com/wildfly/wildfly/pull/16071
 * https://github.com/wildfly/wildfly/pull/16066

PR #16048, opened 2022-09-11
WFLY-16978 Restore parser support for legacy JGroups subsystem schemas

pferraro:
https://issues.redhat.com/browse/WFLY-16978


bstansberry:
Thanks, @pferraro

PR #16050, opened 2022-09-12
[WFLY-16984] Refactoring Batch Subsystem to use new MSC value API

ropalka:
https://issues.redhat.com/browse/WFLY-16984

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/BatchConfigurationService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/BatchConfigurationService.java
@@ -22,20 +22,34 @@ import org.jboss.msc.service.Service;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
-import org.jboss.msc.value.InjectedValue;
 import org.wildfly.security.auth.server.SecurityDomain;
+
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 /**
  * A default batch configuration service.
  *
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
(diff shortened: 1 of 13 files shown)


chengfang:
Some service classes are changed to implement MSC 2 Service API, while some others don't. Can you detail what prevents the use of the new Service in those classes? https://issues.redhat.com/browse/WFLY-16930 (Batch subsystem should maintain its own cache of services) could be issue for BatchConfigurationService, and JobOperatorService.

ropalka:
Sure @chengfang , the difference between using org.jboss.msc.Service and org.jboss.msc.service.Service are the following:

Services implementing new API org.jboss.msc.service will throw UnsupportedOperationException when calling the following deprecated methods:
   - ServiceController.[getValue()](https://github.com/jboss-msc/jboss-msc/blob/1.5/src/main/java/org/jboss/msc/service/ServiceControllerImpl.java#L1067)
   - ServiceController.[awaitValue()](https://github.com/jboss-msc/jboss-msc/blob/1.5/src/main/java/org/jboss/msc/service/ServiceControllerImpl.java#L1075)
   - ServiceController.[awaitValue(long,TimeUnit)](https://github.com/jboss-msc/jboss-msc/blob/1.5/src/main/java/org/jboss/msc/service/ServiceControllerImpl.java#L1098)
   - ServiceController.[getService()](https://github.com/jboss-msc/jboss-msc/blob/1.5/src/main/java/org/jboss/msc/service/ServiceControllerImpl.java#L1131)

By implementing deprected org.jboss.msc.service.Service interface aforementioned deprecated methods backward compatibility is maintained.

I didn't refactor these services because my main purpose for this PR was fixing just dependencies declaration.

ropalka:
These ServiceController methods were deprecated because they are exposing MSC runtime internals to API users.

chengfang (approved):
LGTM, thanks @ropalka !

bstansberry:
Thanks @ropalka

PR #16052, opened 2022-09-12
WFLY-16961 wildfly-cli-client.jar should be in org.wildfly:wildfly-cl…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-16961

### DIFF
--- a/client/cli/pom.xml
+++ b/client/cli/pom.xml
@@ -1,0 +1,168 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source.
+  ~ Copyright 2015, Red Hat, Inc., and individual contributors
+  ~ as indicated by the @author tags. See the copyright.txt file in the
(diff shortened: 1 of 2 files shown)


RanabirChakraborty:
With this PR we have tried to move wildfly-cli-client.jar from wildlfy-core to parent wildfly. Though the existence of wildfly-cli-client.jar is still there in wildfly-core. I haven't touched the core as it needs some changes to be done via - https://issues.redhat.com/browse/WFCORE-6038
If someone can review this PR, that'll be great.

bstansberry:
@RanabirChakraborty I'm not in favor of this approach, as it duplicates the dependency information for very little benefit, particularly to WildFly users. It also would requiring duplication of the 'embedding.jar.jpms....' properties in full, which is not good.

There are good reasons why this jar is created where it is, so any move to duplicate it or move it needs to have a benefit greater than the cost.

bstansberry:
I'm going to close this one as I don't think we should do this.

PR #16053, opened 2022-09-12
WFLY-16979 Restore parser support for legacy Infinispan subsystem schemas

pferraro:
https://issues.redhat.com/browse/WFLY-16979


bstansberry:
Thanks @pferraro

PR #16054, opened 2022-09-12
[WFLY-16228] Don't duplicate Jackson jakarta-rs deps between common-ee and standard-ee

bstansberry:
https://issues.redhat.com/browse/WFLY-16228

Follow up to #16042 to remove some duplicated dependencyManagement


jamezp (approved):
Nice catch. I definitely missed that.

PR #16057, opened 2022-09-13
WFLY-16980 Restore parser support for legacy mod_cluster subsystem schemas

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-16980


kabir (approved):

bstansberry:
Thanks @rhusar

mod_cluster/extension/src/main/java/org/wildfly/extension/mod_cluster/ModClusterSchema.java
```diff
@@ -32,14 +32,12 @@
  */
 public enum ModClusterSchema implements Schema<ModClusterSchema> {
 
-/*  Unsupported schema versions - for reference only
     MODCLUSTER_1_0(1, 0), // AS 7.0
     MODCLUSTER_1_1(1, 1), // EAP 6.0-6.2
```
pferraro, line 36: N.B. We do not need to support these versions.

PR #16025, opened 2022-09-07
[WFLY-16738] Replace the overridden dependencies that should be driven by WildFly Core

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-16738

Notice: This task doesn't include the required changes to move back to Core the Undertow dependencies. [WFLY-16987](https://issues.redhat.com/browse/WFLY-16987) will take care of Undertow.

PR #16063, opened 2022-09-14
[WFLY-16738] Replace the overridden dependencies that should be driven by WildFly Core

yersan:
This PR follows up on https://github.com/wildfly/wildfly/pull/16025.

It removes from WildFly the JBoss Modules and licenses for dependencies that will be fully driven by the WildFly Core dependency management and add to the WildFly EE and preview Galleon packs the resources that come from ee-10-api module incorporated by WildFly Core.

The license entries removed here have been added on https://github.com/wildfly/wildfly-core/pull/5204. Technically this PR doesn't need a WildFly Core release first, but the licenses in WildFly full won't be working as expected until we #5204 is merged and we bump in full the core version. 

Notice none of these changes include any related work for Undertow dependencies. Undertow stuff will be treated on a different Jira

Jira issue: https://issues.redhat.com/browse/WFLY-16738


bstansberry (approved):

bstansberry:
Thanks @yersan. I merged the core PR. I'm fine with not waiting for a core upgrade here as we know one is coming next week before we do any further WildFly release.

PR #16066, opened 2022-09-14
[WFLY-16972] Rewriting DiscoveryService to use new MSC value API

ropalka:
https://issues.redhat.com/browse/WFLY-16972

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
@@ -22,15 +22,21 @@
 
 package org.jboss.as.ejb3.deployment.processors;
 
-import org.jboss.msc.inject.Injector;
-import org.jboss.msc.service.Service;
+import org.jboss.as.ejb3.remote.AssociationService;
+import org.jboss.as.ejb3.remote.RemotingProfileService;
+import org.jboss.msc.Service;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
-import org.jboss.msc.value.InjectedSetValue;
 import org.wildfly.discovery.Discovery;
 import org.wildfly.discovery.spi.DiscoveryProvider;
+import org.wildfly.discovery.impl.StaticDiscoveryProvider;
+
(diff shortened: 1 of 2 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
```diff
@@ -22,45 +22,89 @@
...
+        discoveryConsumer.accept(null);
+        synchronized (providers) {
+            if (remotingDP != null) {
+                providers.remove(remotingDP);
+                remotingDP = null;
+            }
+            if (assocationDP != null) {
+                providers.remove(assocationDP);
+                assocationDP = null;
+            }
```
chengfang, line 96: do we also need to remove any extra discovery providers added via line 105?

chengfang:
@rachmatowicz fyi

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
ropalka: Looking at the code I don't think so @chengfang - these are registered via DiscoveryRegistrationProcessor (lines 73 & 74)
DiscoveryService can be restarted any time so everything done in start() method must be undone in stop() method.

ropalka: To be more precise @chengfang this PR doesn't change anything from functional point of view.
It is just one refactoring step that only refactors dependency configuration staff to use a new MSC value API.
Notice that both external methods (in DiscoveryRegistrationProcessor code):
 * HttpDiscoveryConfigurator.configure() 
 * RemoteEJBDiscoveryConfigurator.configure()

are just injecting new provider into DiscoveryService.
This is what was happening before this PR and this what will be happening after this PR (if merged).

ropalka:
Any update WRT this PR @chengfang ?

chengfang (approved):

bstansberry:
@chengfang This isn't in the 27 Beta1 tag. Please apply either the 28.x or 27.x label to this and update the Fix Version on the JIRA to reflect whether this should go into the Oct 12/13 27 Final, or wait for 28. Whatever you decide is best is ok with me. Thanks!

bstansberry:
Thanks @ropalka

PR #16069, opened 2022-09-15
[WFLY-17007]:  Upgrade jboss metadata to 15.2.0.Beta2

ehsavoie:
* Fixing some XML descriptors.
* Updating jboss-metadata to 15.2.0.Beta2

Jira: https://issues.redhat.com/browse/WFLY-17007

Signed-off-by: Emmanuel Hugonnet <ehugonne@redhat.com>

PR #16070, opened 2022-09-15
[WFLY-17007] Upgrade JBoss Metadata to 15.2.0.Beta2

bstansberry:
Signed-off-by: Brian Stansberry <brian.stansberry@redhat.com>

https://issues.redhat.com/browse/WFLY-17007

This is really just to test a new job I added to the CI set. #16069 is the real PR.


bstansberry:
@ehsavoie @chengfang I see there is ongoing discussion on #16069, which is good. Do either of you see any issue with merging this (just the component upgrade) while that discussion continues?  If the upgrade itself is ok (i.e. better than 15.2.0.Beta1) I'd like to get it in to get it exposed to CI runs and TCK runs.

chengfang:
@bstansberry that sounds good to me.

PR #16071, opened 2022-09-15
[WFLY-16969] ValueManagedReference must use java.util.function.Supplier instead of org.jboss.msc.value.Value

ropalka:
https://issues.redhat.com/browse/WFLY-16969

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
+++ b/naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
@@ -21,18 +21,37 @@
  */
 package org.jboss.as.naming;
 
+import java.util.function.Supplier;
+
 import org.jboss.msc.value.Value;
 
 /**
- * A ManagedReference that simply holds a value'
+ * A ManagedReference that simply holds a value.
  *
  * @author Stuart Douglas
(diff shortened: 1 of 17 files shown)


naming/src/main/java/org/jboss/as/naming/ValueManagedReference.java
```diff
@@ -24,14 +24,31 @@
...
-    public ValueManagedReference(Value<Object> value) {
+    /**
+     * Construct a new instance.
+     *
+     * @param value the value to wrap
+     * @deprecated use {#{@link ValueManagedReference(Object)}} instead. This constructor will be removed in the future.
+     */
+    @Deprecated
+    public ValueManagedReference(final Value<Object> value) {
+        this(value != null ? value.getValue() : null);
```
bstansberry, line 43: This changes the semantic by reading the value at this point instead of in getInstance().

I envisioned the 'real' constructor would take and cache a supplier, which getInstance would use, and this and the public ValueManagedReference(final Object value) constructors would create a supplier.

ropalka: Now I see what you had in mind @bstansberry - PR updated accordingly.

bstansberry (approved):

bstansberry:
Thanks @ropalka!

PR #16074, opened 2022-09-16
[WFLY-17014] Make Galleon provisioning content zips unique per release

bstansberry:
This change does this by adding the maven module pom to the zip.

A reasonable alternative when we have a bit more time would be to convert these modules jar packaging.

Signed-off-by: Brian Stansberry <brian.stansberry@redhat.com>

https://issues.redhat.com/browse/WFLY-17014

Related PR: https://github.com/wildfly/wildfly-core/pull/5211 
A WF Core release with that PR merged is not required before this can be merged.

### DIFF
--- a/ee-9/common-microprofile/assembly.xml
+++ b/ee-9/common-microprofile/assembly.xml
@@ -12,4 +12,10 @@
             <outputDirectory/>
         </fileSet>
     </fileSets>
+    <files>
+        <file>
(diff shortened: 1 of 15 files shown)


jfdenise (approved):

yersan (approved):
LGTM. The Galleon Linux - JDK 11  failure is due to a flaky test which is being tracked under WFLY-10233

PR #16080, opened 2022-09-19
[WFLY-14387] Resource adapters subsystem does not accept expression for wm-security attribute

petrberan:
https://issues.redhat.com/browse/WFLY-14387

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
@@ -311,6 +311,7 @@ public class Constants {
             .setAllowExpression(true)
             .setRequired(false)
             .setDefaultValue(ModelNode.FALSE)
+            .setXmlName(ResourceAdapterParser.Attribute.ENABLED.getLocalName())
             .setRestartAllServices()
             .build();
 
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
@@ -376,7 +376,24 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
                 case START_ELEMENT: {
                     switch (WorkManager.Tag.forName(reader.getLocalName())) {
                         case SECURITY: {
-                            WM_SECURITY.parseAndSetParameter("true", operation, reader);
+                            int attributeSize = reader.getAttributeCount();
+                            if (attributeSize > 0) {
+                                for (int i = 0; i < attributeSize; i++) {
+                                    Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+                                    switch (attribute) {
+                                        case ENABLED: {
+                                            String value = reader.getAttributeValue(i);
+                                            WM_SECURITY.parseAndSetParameter(value, operation, reader);
+                                            break;
+                                        }
+                                        default:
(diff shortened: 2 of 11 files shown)


darranl (changes requested):
@petrberan This PR appears to be making two changes, adding an enabled attribute AND enabling the use of expressions?

We can't add new attributes to schemas already released in a .Final version of WildFly.

petrberan:
@darranl These changes don't really enable the use of expressions, since the security attribute has expressions [allowed](https://github.com/wildfly/wildfly/blob/main/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java#L311) for a long time. Based on Brian's comment in the JIRA issue, I created a new attribute for persisting the expression, however the code is written the way that there should be no regression as the case of not using an expression stays the same and the attribute is used only in case of using an expression

darranl:
@petrberan Ok so if this is just about adding a new attribute that needs to be in a version of the schema and management model NOT released in a prior .Final version of WildFly.  

Once we have released a specific version of model / schema we don't make subsequent changes to it adding or removing attributes.

I am trying to look at the Jira issues but I don't see where the recommendation is to add a new attribute?

petrberan:
@darranl So the changes here could possibly apply for schema v7 (and perhaps v6.2), but not to v2 - v6, did I get that right?

The comment I was reffering to is [this](https://issues.redhat.com/browse/WFLY-14387?focusedCommentId=15732475) one, which is also refferencing [this](https://issues.redhat.com/browse/WFLY-14389?focusedCommentId=15732403) one in a similar issue

darranl:
@petrberan So firstly regarding versions. if a new schema / model is used it should be version 7 for WildFly 27 / 8 - we do major model / schema version bumps for major application server bumps - the reason we have some minor bumps is because the 26.x stream continued whilst development for WildFly 27 was unstable so we could still release application server versions.

I see the underlying issue you are working on here is that the "enabled" status is indicated by the presence of an element and you now need to add an attribute which can contain an expression.

When working on a version bump I recommend doing it as two separate commits as this makes it possible to review the changes in GitHub, if the bump and change is a single commit it is not possible to review using GitHub.

So the first commit would be adding the new schema / bumping the model in the subsystem, adjusting subsystem tests, possibly starting no op transformers for now.  Unfortunately the pattern does change subsystem by subsystem so worth reviewing a prior bump in the history for the changes that were needed (a second benefit of a distinct commit is we can refer back to it in the future).

Once working on adding the new attribute the piece of code to be very familiar with if how this switch statement works:

https://github.com/wildfly/wildfly/blob/main/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java#L376-L396

In general you are likely to need new case statements for versions 5.0, 6.0 and 6.1 which all call the parseWorkSecurityManager_5_0 method https://github.com/wildfly/wildfly/blob/main/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java#L497-L597 you would then create a copy of this method parseWorkSecurityManager_7_0 where you can add support for the new attribute you want to add.

For the two original methods you will need to make sure they set the model value to true as they previously would have based on the element being present, for your new method you can now switch to the parsing of the new attribute.

A final piece of work would then be to handle this in the transformers for when a later version is managing an older secondary host controller - as that older secondary will be missing the ability to persist this new attribute IMO if this attribute is set to an expression it should be rejected by the transformer.  If directly set to true or false the transformer should be able to handle that - but maybe leave transformers till the end once the model and parsing is all in place.

Hopefully this gives more info.

petrberan:
Superseded by https://github.com/wildfly/wildfly/pull/16474

PR #16082, opened 2022-09-19
[WFLY-16238] Don't export traces by default

jasondlee:
https://issues.redhat.com/browse/WFLY-16238

If no sender endpoint or agent is configured, configure a no-op sender.

### DIFF
--- a/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingExtensionLogger.java
+++ b/microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingExtensionLogger.java
@@ -22,17 +22,17 @@
 
 package org.wildfly.extension.microprofile.opentracing;
 
-import org.jboss.logging.BasicLogger;
-import org.jboss.logging.Logger;
-import org.jboss.logging.annotations.LogMessage;
-import org.jboss.logging.annotations.Message;
-import org.jboss.logging.annotations.MessageLogger;
-
 import static org.jboss.logging.Logger.Level.DEBUG;
 import static org.jboss.logging.Logger.Level.ERROR;
 import static org.jboss.logging.Logger.Level.INFO;
+import static org.jboss.logging.Logger.Level.WARN;
 
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.Logger;
 import org.jboss.logging.annotations.Cause;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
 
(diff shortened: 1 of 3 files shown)


microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingExtensionLogger.java
```diff
@@ -81,4 +80,7 @@ public interface TracingExtensionLogger extends BasicLogger {
     OperationFailedException seeDownstream();
     */
 
+    @LogMessage(level = DEBUG)
+    @Message(id = 12, value="No Jaeger endpoint configured. Installing a no-op sender")
```
bstansberry, line 84: Perhaps INFO? Or is this something that logs more than once?

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/spi/sender/WildFlySenderFactory.java
```diff
@@ -30,7 +32,12 @@ public class WildFlySenderFactory implements SenderFactory {
 
     @Override
     public Sender getSender(Configuration.SenderConfiguration configuration) {
-        return new WildFlySender(delegate.getSender(configuration));
+        if (configuration.getEndpoint() == null) {
```
bstansberry, line 35: You need to check configuration.getAgentHost() and configuration.getAgentPort() as well. ThriftSenderFactory uses those to create a UdpSender is there is no endpoint is configured. Those are configured via the sender-binding attribute in the subsystem model.

Tangent: SENDER_BINDING should not allow expressions. It's a capability reference and those shouldn't allow expressions.

microprofile/opentracing-extension/src/main/java/org/wildfly/extension/microprofile/opentracing/TracingExtensionLogger.java
bstansberry: s/endpoint/endpoint or sender-binding/g

bstansberry (changes requested):

jasondlee:
> **Dependency Tree Analyzer Output:**
> 
> New Dependencies:
> 
>     * org.jetbrains.kotlin:kotlin-stdlib-common:jar:1.7.10:compile
> 
>     * org.jetbrains.kotlin:kotlin-stdlib:jar:1.7.10:compile
> 
> 
> Major Version Changes:
> 
>     * com.squareup.okio:okio:jar:1.17.5:compile -> 2.8.0
> 
> 
> CC @wildfly/prod

These changes have been moved to #[16089](https://github.com/wildfly/wildfly/pull/16089)

bstansberry (approved):

PR #16067, opened 2022-09-15
[WFLY-16988]:Remove xalan dependency

jimma:
https://issues.redhat.com/browse/WFLY-16988

PR #16102, opened 2022-09-23
[WFLY-16988] Limit the JPMS setting needed by AbstactUnmarshallingFil…

bstansberry:
…terTestCase to testsuite/integration/basic

https://issues.redhat.com/browse/WFLY-16988

Follow up to #16067 to adjust something I overlooked in code review. Plus adds some missing documentation to better explain how testsuite JPMS config is meant to work.


ropalka (approved):
Thanks for documenting the properties @bstansberry

PR #16106, opened 2022-09-26
[WFLY-17041] Eliminating usage of org.jboss.msc.value.Value interface in ViewDescription

ropalka:
https://issues.redhat.com/browse/WFLY-17041

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
@@ -30,6 +30,7 @@ import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import org.jboss.as.ee.component.interceptors.ComponentDispatcherInterceptor;
 import org.jboss.as.ee.component.interceptors.InterceptorOrder;
@@ -45,8 +46,6 @@ import org.jboss.invocation.Interceptors;
 import org.jboss.invocation.proxy.MethodIdentifier;
 import org.jboss.invocation.proxy.ProxyFactory;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.value.Value;
-import org.jboss.msc.value.ImmediateValue;
 
 import static java.lang.reflect.Modifier.ABSTRACT;
 import static java.lang.reflect.Modifier.PUBLIC;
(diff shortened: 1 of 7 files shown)


ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
```diff
@@ -179,12 +179,25 @@ public Deque<ViewConfigurator> getConfigurators() {
...
      *
      * @param serviceName     The view service name
-     * @param viewClassLoader
+     * @param viewClassLoader The view class loader
+     * @param appclient       appclient environment
+     * @deprecated use {@link ViewDescription#createInjectionSource(ServiceName, Supplier, boolean)} instead. This method will be removed in the future.
      */
     protected InjectionSource createInjectionSource(final ServiceName serviceName, Value<ClassLoader> viewClassLoader, boolean appclient) {
         return new ViewBindingInjectionSource(serviceName);
     }
```
chengfang, line 188: any reason this method is not removed, or the Value not converted to Supplier? It's overloaded with the next one. A quick check show no usage of this protected method.

ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
```diff
@@ -179,12 +179,25 @@ public Deque<ViewConfigurator> getConfigurators() {
...
+    /**
+     * Create the injection source
+     *
+     * @param serviceName     The view service name
+     * @param viewClassLoader The view class loader
+     * @param appclient       appclient environment
+     */
+    protected InjectionSource createInjectionSource(final ServiceName serviceName, Supplier<ClassLoader> viewClassLoader, boolean appclient) {
+        return new ViewBindingInjectionSource(serviceName);
+    }
```
chengfang, line 199: this method only uses the first param and ignoring the other 2 params. Can we just remove the 2 params?

ejb3/src/main/java/org/jboss/as/ejb3/component/EJBViewDescription.java
```diff
@@ -36,7 +36,8 @@
 import org.jboss.invocation.proxy.ProxyFactory;
 import org.jboss.metadata.ejb.spec.MethodInterfaceType;
 import org.jboss.msc.service.ServiceName;
-import org.jboss.msc.value.Value;
+
+import java.util.function.Supplier;
```
chengfang, line 41: can we move the java.* import to the top, which I think is the convention in WF?

ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
ropalka: I removed this BC method. You are right it is not necessary. Only code in WildFly used it.

ee/src/main/java/org/jboss/as/ee/component/ViewDescription.java
ropalka: We cannot. It is template method overwritten by [EJBViewDescription](https://github.com/wildfly/wildfly/blob/main/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBViewDescription.java#L105).

ejb3/src/main/java/org/jboss/as/ejb3/component/EJBViewDescription.java
ropalka: Fixed

chengfang (approved):

PR #16107, opened 2022-09-26
[WFLY-17043] /subsystem=iiop, realm needs to deprecated and rejected at runtime

darranl:
https://issues.redhat.com/browse/WFLY-17043

And model bump:
https://issues.redhat.com/browse/WFLY-17044


tadamski (approved):
LGTM - I did similiar thing with security-domain https://github.com/wildfly/wildfly/pull/16145 and I was wondering whether I should throw the attribute away from parser; but maybe I will refactor my PR on top of this one and do the same thing ( == fail on boot and not on parse)

bstansberry:
Thanks @darranl

darranl:
@tadamski +1 is needs to be fail in stage Runtime as we still need to support mixed domain mode where these attributes will be read in the primary host controller to manage older secondaries.

PR #16108, opened 2022-09-26
[WFLY-17046] Deprecate and reject remaining security-realm references in the Undertow subsystem.

darranl:
https://issues.redhat.com/browse/WFLY-17046


darranl:
@bstansberry FYI this is not really a blocker until WildFly 28 but the Undertow subsystem appears to contain an unused model bump so may as well use it so we don't need to revert or waste it.

bstansberry (changes requested):
Needs a minor cleanup in LocalDescriptions.properties; otherwise looks good

darranl:
Thanks @bstansberry should be fixed now

bstansberry (approved):

bstansberry:
Thanks @darranl

PR #16111, opened 2022-09-26
[WFLY-16854] [WFLY-17023] [WFLY-17024] Update dependencies for Opentelemetry and MicroProfile OpenTracing

bstansberry:
https://issues.redhat.com/browse/WFLY-16854
https://issues.redhat.com/browse/WFLY-17023
https://issues.redhat.com/browse/WFLY-17024

This combines these 3 PRs:

#15961
#16089
#16090

It uses the separate module kotlin artifact packaging approach from #16089 instead of the approach in #15961. I did that because it was easiest to pull things together that way. @jasondlee in the discussion at https://github.com/wildfly/wildfly/pull/15961#discussion_r950076537 it sounded like the #15961 approach was preferable. If so, please follow up with a PR moving to that.

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -2052,6 +2052,30 @@
             </dependency>
 
             <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-stdlib</artifactId>
+                <version>${version.org.jetbrains.kotlin}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+
+            <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-stdlib-common</artifactId>
+                <version>${version.org.jetbrains.kotlin}</version>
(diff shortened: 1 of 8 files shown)


bstansberry:
Note that these updates break the Cloud Tests. @kabir FYI.

I suspect that has nothing to do with OTel/OT etc, and may be more about dependency management in the Cloud Tests.

jasondlee (approved):
I think this approach is fine, Brian. My comment about the new module was, iirc, mainly about a new module for jackson-jr-objects, which is unlikely to be needed somewhere else. The Kotlin libs will most likely need to be reused elsewhere as time rolls on.

bstansberry:
@jasondlee @boris-unckel I added a commit that I hope will address the Cloud Tests failure. kotlin-stdlib-jdk8 depends on kotlin-stdlib-jdk7, and the Cloud Tests failure is due to a CNFE for a class in that jar.

kabir:
@bstansberry it looks like you got it passing

bstansberry:
@kabir Yes, I fixed it by taking the liberty of filing and merging https://github.com/wildfly-extras/wildfly-cloud-tests/pull/31.

/me requests forgiveness; asking permission is too much hassle ;)

bstansberry:
Thanks @boris-unckel and @jasondlee

PR #16112, opened 2022-09-26
[WFLY-17045] Move RESTEasy Spring to WildFly Preview as there is not …

jamezp:
…a final version of Spring 6.

https://issues.redhat.com/browse/WFLY-17045


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
```diff
@@ -85,9 +85,6 @@ public void registerAdditionalRuntimePackages(ManagementResourceRegistration res
                     RuntimePackageDependency.optional(RESTEASY_MULTIPART.getName()),
                     RuntimePackageDependency.optional(RESTEASY_CRYPTO.getName()),
                     RuntimePackageDependency.optional(JACKSON_DATATYPE_JDK8.getName()),
-                    RuntimePackageDependency.optional(JACKSON_DATATYPE_JSR310.getName()),
-                    // The following ones are optional dependencies located in org.jboss.as.jaxrs module.xml
-                    // To be provisioned, they need to be explicitly added as optional packages.
-                    RuntimePackageDependency.optional("org.jboss.resteasy.resteasy-spring"));
```
jamezp, line 91: I don't think this was required, but I could use some validation from someone that understands Galleon more than I do. From my understanding the module will be configured in WildFly Preview because it's defined there and there is a task that runs to copy the actual libraries into the modules directory.

pom.xml
```diff
@@ -513,7 +513,7 @@
         <legacy.version.org.jboss.resteasy>4.7.4.Final</legacy.version.org.jboss.resteasy>
         <version.org.jboss.resteasy>6.2.0.Final</version.org.jboss.resteasy>
         <version.org.jboss.resteasy.microprofile>2.0.0.Final</version.org.jboss.resteasy.microprofile>
-        <version.org.jboss.resteasy.spring>3.0.0.Beta1</version.org.jboss.resteasy.spring>
+        <preview.version.org.jboss.resteasy.spring>3.0.0.Beta1</preview.version.org.jboss.resteasy.spring>
```
bstansberry, line 516: See https://github.com/wildfly/wildfly/blob/b4744ce4db091031136b23a56006b45ca18a3cf0/pom.xml#L593-L598

bstansberry (approved):

bstansberry:
Thanks @jamezp

PR #16114, opened 2022-09-27
WFLY-17047 Weld Probe removal

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-17047

~~This PR does not contain Weld version update yet, just changes related to Probe removal~~
EDIT: Weld 5.1.0 is now syncing with Central and this PR will contains the updated version as well.


manovotn:
Since there was no other than Alpha release of WFLY, I have decided it would be cleaner to change weld subsystem 5.0 to not accept devel mode property at all.
However, this draft is currently failing in `WeldSubsystemTestCase` because older subsystem versions can have this property. 
Any hints on what's the proper way to handle this @bstansberry? I tried looking into `WeldTransformers`, but that seems to be for the opposite case (where you have newer config and need to adapt it to older) - but I am probably just missing something, right?

weld/subsystem/src/main/java/org/jboss/as/weld/WeldResourceDefinition.java
```diff
@@ -77,6 +77,7 @@ class WeldResourceDefinition extends PersistentResourceDefinition {
             .setRestartAllServices()
             .build();
 
+    // no longer used since Weld 5
     static final SimpleAttributeDefinition DEVELOPMENT_MODE_ATTRIBUTE =
             new SimpleAttributeDefinitionBuilder(DEVELOPMENT_MODE_ATTRIBUTE_NAME, ModelType.BOOLEAN, true)
             .setAllowExpression(true)
```
bstansberry, line 83: This builder needs a call added:

.setDeprecated(ModelVersion.create(5, 0))

bstansberry: Also in the LocalDescriptions.properties file you'll need to add a 'wed.development-mode.deprecated' property. The message for that is what gets logged if this deprecated attribute is used. Something like "Development mode is no longer supported on servers running the current version."

weld/subsystem/src/main/java/org/jboss/as/weld/WeldResourceDefinition.java
```diff
@@ -91,7 +92,7 @@ class WeldResourceDefinition extends PersistentResourceDefinition {
             .setRestartAllServices()
             .build();
 
-    private static final AttributeDefinition[] ATTRIBUTES = new AttributeDefinition[] { REQUIRE_BEAN_DESCRIPTOR_ATTRIBUTE, LEGACY_EMPTY_BEANS_XML_TREATMENT_ATTRIBUTE, NON_PORTABLE_MODE_ATTRIBUTE, DEVELOPMENT_MODE_ATTRIBUTE, THREAD_POOL_SIZE_ATTRIBUTE };
+    private static final AttributeDefinition[] ATTRIBUTES = new AttributeDefinition[] { REQUIRE_BEAN_DESCRIPTOR_ATTRIBUTE, LEGACY_EMPTY_BEANS_XML_TREATMENT_ATTRIBUTE, NON_PORTABLE_MODE_ATTRIBUTE, THREAD_POOL_SIZE_ATTRIBUTE };
```
bstansberry, line 95: This attribute can't be dropped as it's needed to support mixed version domain with secondary hosts running earlier versions. Also before removing we need to deprecate it and warn (see .setDeprecated... bit above) to give users a chance to remove it from any CLI scripts etc.  We don't have to retain the behavior; we just need to keep the attribute and make it a no-op other than the warn.

darranl:
I have had to do what Brian is suggesting here where I have been disabling the security-realm attribute across subsystems e.g.

https://github.com/wildfly/wildfly/pull/16107

In my case I also report an error at runtime if set but that is probably a step too far for this attribute.

weld/subsystem/src/main/java/org/jboss/as/weld/WeldResourceDefinition.java
manovotn: @bstansberry I have updated the PR accordingly.

manovotn:
What about the test failures I mentioned above? I am not quite grasping how are tests such as [this](https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/test/java/org/jboss/as/weld/WeldSubsystemTestCase.java#L74-L77) supposed to behave when an attribute is being removed from newer version.

manovotn:
I'll get back to this later today.
Looking at @darranl PR, I think I understand what you meant. Latest version of schema still needs to support dev mode attribute, but it will be a no-op.

manovotn:
The PR is now changed to remove Probe as a dependency but it retains Weld devel mode configuration option as it was, only deprecates it.
Is this what you meant @bstansberry @darranl ?

bstansberry (approved):
This LGTM @manovotn 

As @darranl said it's possible to throw an OperationFailedException if this is set to true, e.g. at https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/WeldSubsystemAdd.java#L88, but I think that's overkill. (That code only executes in a server that's running in normal mode.) Doing that would interfere with people using a particular config in different dev environments, without adding a ton of value vs the WARN that gets logged.

bstansberry:
Thanks @manovotn!

PR #16131, opened 2022-10-03
[WFLY-17069] Unlock preview features from underlying JDKs by default

ropalka:
https://issues.redhat.com/browse/WFLY-17069

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -169,6 +169,7 @@
             for the maven module that includes the test.
          -->
         <modular.jdk.args>
+            --enable-preview
             --add-exports=java.desktop/sun.awt=ALL-UNNAMED
             --add-exports=java.naming/com.sun.jndi.ldap=ALL-UNNAMED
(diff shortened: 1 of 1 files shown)


bstansberry:
This will need some discussion. https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/JDK.20Preview.20features

bstansberry:
Once we decide to include a change like this, I think it should be enabled via a profile.

That said, the property you're editing here is used client-side.

yersan:
It is related to https://github.com/wildfly/wildfly-core/pull/5229

bstansberry:
@ropalka I'm going to close this and https://github.com/wildfly/wildfly-core/pull/5229 as the JGroups problem has been corrected. If you want to continue to pursue the topic as a general thing to think about, let's do it on the wildfly-dev mail list.

bstansberry:
Thanks again @ropalka for identifying and raising this problem!

PR #16132, opened 2022-10-04
[WFLY-17051] Add a circular dependency on the JSON Binding API to the…

jamezp:
… implementation, Yasson. This gets around possible service loader issues when the security manager is used. It instead uses the RI which is Yasson by adding the implementation dependency on the API.

https://issues.redhat.com/browse/WFLY-17051
https://issues.redhat.com/browse/WFLY-11365

This is some what of a hack and I'd like to get @bstansberry's approval before this is merged. This works by skipping the `ServiceLoader` finding any implementations and allows for the hard-coded RI to be tried which gets around the issues with the security manager not being able to read the file.

### DIFF
--- a/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/json/bind/api/main/module.xml
+++ b/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/json/bind/api/main/module.xml
@@ -28,5 +28,10 @@
 
     <dependencies>
         <module name="javax.json.api"/>
+        <!-- Circular dependency on the implementation to ensure that:
+             1. The implementation is loaded
+             2. It is done last via the hard-coded class name in the spec API
+         -->
+        <module name="org.eclipse.yasson"/>
     </dependencies>
 </module>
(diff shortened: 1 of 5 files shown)


ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/json/bind/api/main/module.xml
```diff
@@ -27,6 +27,8 @@
     </resources>
 
     <dependencies>
-        <module name="javax.json.api"/>
+        <module name="jakarta.json.api"/>
+        <!-- Circular dependency on the implementation -->
```
bstansberry, line 31: This could use a bit more explanation, to save maintainers having to git blame.

ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/json/bind/api/main/module.xml
```diff
@@ -27,6 +27,8 @@
     </resources>
 
     <dependencies>
-        <module name="javax.json.api"/>
+        <module name="jakarta.json.api"/>
```
bstansberry, line 30: Please don't make this kind of change in unrelated PRs. It's just going to lead to randomness in how we handle this.

jamezp: Oops, I didn't mean to commit that. FWIW this does seem like something we should change at some point as we have to locate and read two files to load modules this way.

bstansberry (approved):

ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/json/bind/api/main/module.xml
bstansberry: Yes. There have been conversations about it; we'll do it at some point. Probably soon.

https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Modules.20javax.20dependencies

PR #16137, opened 2022-10-05
[WFLY-17065] Remove the exporting of the services from the circular i…

jamezp:
…mplementation dependency on the jakarta.xml.bind.api module. This allows the fallback to the hard-coded implementation and works around issues not being able to read files when the security manager is enabled.

Also re-enabled 2 tests that previously had to be ignored due to EE 8/EE 10 conflicts.

https://issues.redhat.com/browse/WFLY-17065

### DIFF
--- a/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/xml/bind/api/main/module.xml
+++ b/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/xml/bind/api/main/module.xml
@@ -35,7 +35,15 @@
         <module name="jakarta.activation.api" export="true"/>
         <module name="org.eclipse.angus.activation" services="export" export="true"/>
         <module name="javax.xml.stream.api"/>
-        <!-- Export the implementation and services as the TCCL is typically the CL of a deployment -->
-        <module name="org.glassfish.jaxb" services="export" export="true"/>
+        <!-- Export the implementation as the TCCL is typically the CL of a deployment. We do not want to export the
+             services as that would trigger the loading through the service loader. This can be an issue and require
+             special permissions if the security manager is present. Not exporting the services falls back to a default
+             implementation provided by the org.glassfish.jaxb module.
+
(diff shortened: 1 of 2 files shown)


ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/xml/bind/api/main/module.xml
```diff
@@ -35,7 +35,11 @@
...
         <module name="org.eclipse.angus.activation" services="export" export="true"/>
         <module name="javax.xml.stream.api"/>
-        <!-- Export the implementation and services as the TCCL is typically the CL of a deployment -->
-        <module name="org.glassfish.jaxb" services="export" export="true"/>
+        <!-- Export the implementation as the TCCL is typically the CL of a deployment. We do not want to export the
+             services as that would trigger the loading through the service loader. This can be an issue and require
+             special permissions if the security manager is present. Not exporting the services falls back to a default
+             implementation provided by the org.glassfish.jaxb module.
+         -->
+        <module name="org.glassfish.jaxb" export="true"/>
```
bstansberry, line 43: tl;dr; I think this is ok, but it does remove exporting one service type besides the JAXB API one. Which is fine IMHO; I'm just noting it here for the record.

org.glassfish.jaxb:jaxb-xjc:4.0.1 has a META-INF/services/com.sun.tools.xjc.Plugin. That class is in the same jar and is an abstract class, not an interface. The file lists these impls:

com.sun.tools.xjc.addon.code_injector.PluginImpl
com.sun.tools.xjc.addon.locator.SourceLocationAddOn
com.sun.tools.xjc.addon.sync.SynchronizedMethodAddOn
com.sun.tools.xjc.addon.at_generated.PluginImpl
com.sun.tools.xjc.addon.episode.PluginImpl
com.sun.tools.xjc.addon.accessors.PluginImpl

If we actually did want to support making those available somehow, doing it via the jakarta.xml.bind.api module is not the way to do it.

jamezp:
Locally ran TCK passes too:
```
Test results: passed: 24,626
```

PR #16142, opened 2022-10-06
[WFLY-17050] Clean up module dependencies for Micrometer

jasondlee:
https://issues.redhat.com/browse/WFLY-17050

PR feedback
Remove metrics dep from Micrometer module
Copyright header updates
Add shared capability to prevent metrics and micrometer from both being provisioned 
Override the health layer to remove the metrics dep 
Modify LayersTestCase to allow for WFP-specific data


ee-9/galleon-content/src/main/resources/layers/standalone/observability/layer-spec.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:1.0" name="observability">
+    <dependencies>
+        <layer name="health" optional="true" />
```
bstansberry, line 4: Shouldn't micrometer be here?

Originally the 'observability' concept (not the layer) meant logging, metrics, health, distributed tracing. We created a layer around that concept. We didn't put logging in it, because it was always present anyway. (Which isn't conceptually valid and something to reconsider for WF 28.) And then tracing didn't make the layer because early on we got feedback that MP OT added runtime overhead even if not really sending any spans. (Which we should reexamine for WFP 28, i.e. whether we should put OTel in WFP's 'observability' layer.) Which left us with a layer for providing metrics in some form and one for health.

So I think with Micrometer in WFP, it takes the place of WF Metrics in providing the metrics part of observability.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemAdd.java
```diff
@@ -57,40 +55,37 @@ public class MicrometerSubsystemAdd extends AbstractBoottimeAddStepHandler {
...
+        MicrometerRegistryService.install(context, securityEnabled);
+        MicrometerCollectorService.install(context);
+        MicrometerContextService.install(context, securityEnabled);
+
+        context.addStep(new AbstractDeploymentChainStep() {
+            @Override
+            public void execute(DeploymentProcessorTarget processorTarget) {
+                processorTarget.addDeploymentProcessor(SUBSYSTEM_NAME, DEPENDENCIES, 0x1910,
+                        new MicrometerDependencyProcessor());
+                processorTarget.addDeploymentProcessor(SUBSYSTEM_NAME, POST_MODULE, POST_MODULE_METRICS - 1, // ???
```
bstansberry, line 73: IIRC you added a Phase constant for this in WF Core, which will replace this when integrated in full WF. Is this the same value as what that will be?

jasondlee: Yes, that constant will replace this, but it doesn't look like Core has been updated yet, so I can't use the constant.

ee-9/galleon-content/src/main/resources/layers/standalone/observability/layer-spec.xml
jasondlee: Updated.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerCdiExtension.java
```diff
@@ -0,0 +1,38 @@
...
+public class MicrometerCdiExtension implements Extension {
+    private final MeterRegistry registry;
+
+    public MicrometerCdiExtension(MeterRegistry meterRegistry) {
+        registry = meterRegistry;
+    }
+
+
+    public void registerMicrometerBeans(@Observes AfterBeanDiscovery abd) {
+        abd.addBean().addTransitiveTypeClosure(MeterRegistry.class).produceWith(i -> registry);
```
manovotn, line 35: This won't fix the issue you are seeing but you'll probably want to add a scope here as well - `@Singleton` seems like a good fit. Otherwise you get `@Dependent` by default.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemDefinition.java
```diff
@@ -37,17 +36,13 @@
...
-
     static final String CLIENT_FACTORY_CAPABILITY = "org.wildfly.management.model-controller-client-factory";
     static final String HTTP_EXTENSIBILITY_CAPABILITY = "org.wildfly.management.http.extensible";
     static final String MANAGEMENT_EXECUTOR = "org.wildfly.management.executor";
-    static final String METRICS_SCAN_CAPABILITY = "org.wildfly.extension.metrics.scan";
-    static final String MICROMETER_HTTP_SECURITY_CAPABILITY = MICROMETER_MODULE + ".http-context.security-enabled";
     static final String PROCESS_STATE_NOTIFIER = "org.wildfly.management.process-state-notifier";
-    static final String METRICS_HTTP_CONTEXT_CAPABILITY = "org.wildfly.extension.metrics.http-context";
 
     private static final RuntimeCapability<Void> MICROMETER_COLLECTOR_RUNTIME_CAPABILITY =
```
bstansberry, line 47: This capability is never registered, i.e. in the current L92.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemDeploymentProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+                // Contexts and Dependency Injection, there's no integration needed
+                MICROMETER_LOGGER.noCdiDeployment();
+                return;
+            }
+            WildFlyRegistry registry = (WildFlyRegistry) deploymentUnit.getServiceRegistry()
+                    .getService(MICROMETER_REGISTRY_RUNTIME_CAPABILITY.getCapabilityServiceName()).getValue();
+
+            weldCapability.registerExtensionInstance(new MicrometerCdiExtension(registry), deploymentUnit);
+        } catch (CapabilityServiceSupport.NoSuchCapabilityException e) {
+            //We should not be here since the subsystem depends on weld capability. Just in case ...
```
bstansberry, line 83: This comment isn't actually true. I don't see any capability in MicrometerSubsystemDefinition that depends on the 'org.wildfly.weld' capability. 

Is the CDI dep actually required, or could this just ignore the failure or do some info or debug logging? IOW w/o CDI I think the JVM and management API metrics would still work, and would be exported in the Micrometer format, which differs from WF Metrics format. So the subsystem could be useful even without CDI.

If CDI really is required, there needs to be a capability registered by this subsystem that requires it; otherwise the management API won't prevent anyone from building a config w/ Micrometer and w/o CDI or removing CDI from a config with Micrometer.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemAdd.java
```diff
@@ -57,40 +54,37 @@ public class MicrometerSubsystemAdd extends AbstractBoottimeAddStepHandler {
...
-                public void execute(DeploymentProcessorTarget processorTarget) {
-                    processorTarget.addDeploymentProcessor(SUBSYSTEM_NAME, DEPENDENCIES, 0x1910,
-                            new MicrometerDependencyProcessor());
-                    processorTarget.addDeploymentProcessor(SUBSYSTEM_NAME, POST_MODULE, POST_MODULE_METRICS - 1, // ???
-                            new MicrometerSubsystemDeploymentProcessor(exposeAnySubsystem, exposedSubsystems));
-                }
-            }, RUNTIME);
+        context.addStep((operationContext, modelNode) -> {
+            ServiceController<?> serviceController = context.getServiceRegistry(false).getService(MICROMETER_COLLECTOR);
+            MicrometerCollector micrometerCollector = MicrometerCollector.class.cast(serviceController.getValue());
```
bstansberry, line 79: Please file a JIRA to get rid of using the MSC ServiceRegistry this way. MicrometerCollectorService.install can return a Supplier< MicrometerCollector> that gets populated in MicrometerCollectorService.start and cleared in stop.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemDeploymentProcessor.java
jasondlee: You're right. CDI isn't required. Without it, the app just won't have the MeterRegistry exposed as a CDI bean, but the extension should load and run without it just fine. I've modified that logger method to log a message rather than throw an Exception.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemAdd.java
jasondlee: Filed: https://issues.redhat.com/browse/WFLY-17166

bstansberry:
Woohoo!

PR #16145, opened 2022-10-07
[WFLY-15920] IIOP: deprecate legacy security-domain attribute

tadamski:
https://issues.redhat.com/browse/WFLY-15920

### DIFF
--- a/iiop-openjdk/src/main/resources/org/wildfly/iiop/openjdk/LocalDescriptions.properties
+++ b/iiop-openjdk/src/main/resources/org/wildfly/iiop/openjdk/LocalDescriptions.properties
@@ -26,6 +26,7 @@ iiop-openjdk.export-corbaloc=Indicates whether the root context should be export
 # security configuration properties.
 iiop-openjdk.support-ssl=Indicates whether SSL is to be supported (on) or not (off).
 iiop-openjdk.security-domain=The name of the security domain that holds the key and trust stores that will be used to establish SSL connections.
+iiop-openjdk.security-domain.deprecated=Legacy security domains are not supported in a running server. Only usable in a domain configuration to support secondary hosts running earlier versions.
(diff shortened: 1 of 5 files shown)


tadamski:
just noticed we have to halt it because of https://github.com/wildfly/wildfly/pull/16107/files

tadamski:
updated

iiop-openjdk/src/main/resources/org/wildfly/iiop/openjdk/LocalDescriptions.properties
```diff
@@ -26,6 +26,7 @@ iiop-openjdk.export-corbaloc=Indicates whether the root context should be export
 # security configuration properties.
 iiop-openjdk.support-ssl=Indicates whether SSL is to be supported (on) or not (off).
 iiop-openjdk.security-domain=The name of the security domain that holds the key and trust stores that will be used to establish SSL connections.
+iiop-openjdk.security-domain.deprecated=The name of the security domain that holds the key and trust stores that will be used to establish SSL connections.
```
bstansberry, line 29: This message should be a description of the deprecation, similar to what follows an @deprecated in Javadoc.

Perhaps:

"Legacy security domains are not supported in a running server. Only usable in a domain configuration to support secondary hosts running earlier versions"

bstansberry (changes requested):

iiop-openjdk/src/main/resources/org/wildfly/iiop/openjdk/LocalDescriptions.properties
tadamski: @bstansberry thanks, fixed

bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #16154, opened 2022-10-10
WFLY-17119 Distributed session metadata only needs millisecond precision

pferraro:
https://issues.redhat.com/browse/WFLY-17119

### DIFF
--- a/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SimpleSessionAccessMetaData.java
+++ b/clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SimpleSessionAccessMetaData.java
@@ -23,6 +23,7 @@
 package org.wildfly.clustering.web.cache.session;
 
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 
 /**
  * @author Paul Ferraro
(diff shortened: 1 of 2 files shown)


clustering/web/cache/src/main/java/org/wildfly/clustering/web/cache/session/SimpleSessionAccessMetaData.java
```diff
@@ -44,11 +45,14 @@ public Duration getLastAccessDuration() {
 
     @Override
     public void setLastAccessDuration(Duration sinceCreation, Duration lastAccess) {
-        int nano = sinceCreation.getNano();
-        // Only retain millisecond precision
-        this.sinceCreation = (nano % 1_000_000) > 0 ? sinceCreation.withNanos((nano / 1_000_000) + 1) : sinceCreation;
-        // Only retain second precision
-        this.lastAccess = (lastAccess.getNano() > 0) ? Duration.ofSeconds(lastAccess.getSeconds() + 1) : lastAccess;
+        this.sinceCreation = normalize(sinceCreation);
+        this.lastAccess = normalize(lastAccess);
```
bstansberry, line 49: No concerns with potentially dropping as much a 999ms from these values?

pferraro: This is intentional.  The lastAccess metadata value is used to determine expiration eligibility (unlike sinceCreation, it is not reported via any public API), hence the original coarse precision.  However, upon further review, capturing the duration of a typical request (i.e. << 255 ms) in seconds vs milliiseconds has the same replication/persistence payload cost.  Given this, it is better to allow sessions to expire more promptly, and thus free heap space sooner.

bstansberry: Thanks, Paul.

bstansberry (approved):

PR #16155, opened 2022-10-10
[WFLY-17086] Warning message WFLYWELD0052 in ee-security quickstart

skyllarr:
Add weld dependency to module jakarta.security.enterprise.api in order to remove warning message WFLYWELD0052 in ee-security quickstart

issue: https://issues.redhat.com/browse/WFLY-17086

### DIFF
--- a/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/security/enterprise/api/main/module.xml
+++ b/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/security/enterprise/api/main/module.xml
@@ -41,5 +41,11 @@
         <module name="javax.annotation.api" />
         <module name="javax.inject.api" />
         <module name="javax.json.api" />
+        <!--TODO remove the WELD dependencies from this module, see comments on https://github.com/wildfly/wildfly/pull/16155 for more details
+        soteria overrides the beanClass in CDIExtension, which makes weld not load them from appropriate module
+         -->
+        <module name="org.jboss.weld.api"/>
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
     </dependencies>
 </module>


bstansberry:
@Skyllarr @manovotn This seems odd. I don't see these kinds of Weld dependencies in other EE API modules, and from a quick look at the ee-security quickstart it looks like what's being injected related to jakarta.security.enterprise.api are instances of interfaces in the API. Which seems like something that would be done with many other EE APIs. So what makes security different?

Is it possible that the dependencies should be in an implementation module, e.g. org.glassfish.soteria or org.wildfly.security.jakarta.security? Both of those have org.jboss.weld.core and org.jboss.weld.spi deps, but not org.jboss.web.api.

manovotn:
> Is it possible that the dependencies should be in an implementation module, e.g. org.glassfish.soteria or org.wildfly.security.jakarta.security? Both of those have org.jboss.weld.core and org.jboss.weld.spi deps, but not org.jboss.web.api.

+1, this will likely do the trick because I presume it is the impl module that "owns" those beans.

Skyllarr:
Still working on this one. So far niether of mentioned implementation modules resolved the warning. I also tried other modules like undertow and elytron-web but those also did not resolve it

manovotn:
Just guessing but maybe you can try to look at the module dependencies of other Jakarta specs and their respective impl.
Basically, the same situation should exist for any Jakarta EE spec which defines an interface which is then provided as a bean by the implementation. So something like entity or transaction managers for instance.
Alternatively, you can try to debug the WARN code - I assume it is coming from [this](https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/ProxyServicesImpl.java#L137-L170) method.

Skyllarr:
@manovotn Thank you! The ee-security quickstart uses [UsernamePasswordCredential](https://github.com/wildfly/quickstart/blob/main/ee-security/src/main/java/org/jboss/as/quickstarts/ee_security/TestAuthenticationMechanism.java#L56)   and the implementation of that is in the jakarta.security.enterprise-api-3.0.0.jar so this PR is correct as is IMO

manovotn:
If you can't crack it, I can try to take a look at this next week.

manovotn:
@Skyllarr @bstansberry I did my debugging homework and I think I found the problem.

These beans (in the quickstart I saw `IdentityStoreHandler`) are coming from WFLY soteria impl. Namely I found them here - https://github.com/wildfly-security/soteria/blob/1.0.1-jbossorg-x/impl/src/main/java/org/glassfish/soteria/cdi/CdiExtension.java#L207-L218
Please correct me, if that's wrong assumption.

Now, these beans are defined synthetically (i.e. via CDI extension) and their values are manually set. This is standard approach, nothing wrong with it.
However, when Weld needs to create a proxy for classin WFLY, it will looks for bean attribute that we commonly refer to as `bean class` and use that as a reference class for determining which module does it belong to. For standard beans, this is their java class but for synthetic beans, this is by default the extension that defines them.
The soteria impl overrides this and sets the API class, [see this code](https://github.com/wildfly-security/soteria/blob/1.0.1-jbossorg-x/impl/src/main/java/org/glassfish/soteria/cdi/CdiExtension.java#L210).

I haven't tested it, but I think that if you set this to the extension class, it should start working without as expected.
Also note that the [`CDIProducer` class sets `Object` as a default for this](https://github.com/wildfly-security/soteria/blob/1.0.1-jbossorg-x/impl/src/main/java/org/glassfish/soteria/cdi/CdiProducer.java#L42), which doesn't make sense.

manovotn:
@Skyllarr @bstansberry I have verified the above theory locally with updated artifacts and I am not seeing the WARN anymore.
You need a code like this - https://github.com/wildfly-security/wildfly-elytron-ee/compare/3.x...manovotn:wildfly-elytron-ee:3.x?expand=1 and then OFC to update this in WFLY core and later in WFLY.
Feel free to use the code as you like :)

Skyllarr:
@manovotn I see, thanks a lot for explanation! I could not find the problem myself. 

Btw I was comparing the flow with other quickstarts and found similar warning for jakarta.faces in [hibernate quickstart](https://issues.redhat.com/browse/WFLY-17153) so I will look and maybe it is a similar problem

bstansberry:
@jasondlee FYI re @Skyllarr's last comment / https://issues.redhat.com/browse/WFLY-17153

Thanks @manovotn for the research!

FWIW, if you all don't disagree, I don't mind the approach Diana took in this PR as a short term workaround for WF 27 (which we're trying to wrap up), perhaps with a TODO comment to explain it will be removed in WF 28. The PR doesn't export the weld modules, it just makes them visible to the jakarta api module.

manovotn:
> FWIW, if you all don't disagree, I don't mind the approach Diana took in this PR as a short term workaround for WF 27 (which we're trying to wrap up), perhaps with a TODO comment to explain it will be removed in WF 28. The PR doesn't export the weld modules, it just makes them visible to the jakarta api module.

Totally fine by me, I just wanted to nail down the actual cause :)

ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/security/enterprise/api/main/module.xml
```diff
@@ -41,5 +41,11 @@
         <module name="javax.annotation.api" />
         <module name="javax.inject.api" />
         <module name="javax.json.api" />
+        <!--TODO move the WELD dependencies to implementation module instead, see comments on https://github.com/wildfly/wildfly/pull/16155 for more details
```
manovotn, line 44: Nitpicker's note - impl module already has these deps, so once it is solved, these can be simply erased.

Skyllarr: @manovotn I think the org.jboss.weld.api is missing there, but I updated the message to be more clear. Thanks!

Skyllarr:
@bstansberry I've added weld deps with TODO comment here and to jakarta.faces.api https://github.com/wildfly/wildfly/pull/16183  so it can be addressed in WF28 as you suggested

bstansberry:
Thanks @Skyllarr

bstansberry:
Thanks @manovotn -- I'd have been reluctant to merge this without your nailing down the underlying problem.

PR #16159, opened 2022-10-11
[WFLY-17112] Can't load a custom load balancing policy on a pooled

ivassile:
connection factory

Issue: https://issues.redhat.com/browse/WFLY-17112


ehsavoie (changes requested):
I'd rather export the dependency in the journal.
Also there is no test :)

ivassile:
Replaced by https://github.com/wildfly/wildfly/pull/16162

PR #16162, opened 2022-10-11
[WFLY-17112]: Can't load a custom load balancing policy on a pooled connection factory.

ehsavoie:
* Exporting "org.apache.activemq.artemis.addons" so that the class can be found by Artemis classes.
* Adding a test to cover this.

Jira: https://issues.redhat.com/browse/WFLY-17112

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ActivemqServerAddonsTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ActivemqServerAddonsTestCase.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2022 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.as.test.manualmode.messaging;
+
+import static org.jboss.as.test.shared.integration.ejb.security.PermissionUtils.createPermissionsXmlAsset;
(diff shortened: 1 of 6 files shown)


testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ActivemqServerAddonsTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+    public void setup() throws Exception {
+        try ( ManagementClient managementClient = createManagementClient()) {
+            final JavaArchive jar = ShrinkWrap.create(JavaArchive.class, "artemis-lb-policy.jar")
+                    .addClasses(OrderedLoadBalancingPolicy.class);
+            final Path addonsModule = getAddonsPath().resolve("main");
+            Files.createDirectories(addonsModule);
+            Files.writeString(addonsModule.resolve("module.xml"), MODULE_XML, StandardOpenOption.CREATE);
+            try ( OutputStream out = Files.newOutputStream(addonsModule.resolve("artemis-lb-policy.jar"), StandardOpenOption.CREATE)) {
+                jar.as(ZipExporter.class).exportTo(out);
+            }
```
bstansberry, line 108: Please use org.jboss.as.test.module.util.TestModule.

bstansberry (changes requested):

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ActivemqServerAddonsTestCase.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Before
+    public void setup() throws Exception {
+        try ( ManagementClient managementClient = createManagementClient()) {
+            if (container.isStarted(DEFAULT_FULL_JBOSSAS)) {
+                container.stop(DEFAULT_FULL_JBOSSAS);
+            }
+            final JavaArchive jar = module.addResource("artemis-lb-policy.jar");
+            jar.addClasses(OrderedLoadBalancingPolicy.class);
+            System.setProperty("module.path", Paths.get("target" ,"wildfly", "modules").toAbsolutePath().toString());
```
bstansberry, line 94: This is ok but the typical way we deal with this is via module.path in system-property-variables in the pom.

That's not necessarily better, it's just the usual handling that people might look for.

bstansberry (approved):
I had one comment but that isn't going to stop me merging this.

bstansberry:
Thanks @ehsavoie

PR #16165, opened 2022-10-12
[WFLY-17122][WFLY-17124][WFLY-17135]:Drop source transformation for x…

jimma:
…ts, webservices and weld/webservices subsystem

https://issues.redhat.com/browse/WFLY-17122
https://issues.redhat.com/browse/WFLY-17124
https://issues.redhat.com/browse/WFLY-17135


ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/webservices/main/module.xml
```diff
@@ -33,7 +33,7 @@
     </exports>
 
     <resources>
-        <artifact name="\${org.wildfly:wildfly-webservices-server-integration@module.jakarta.suffix@}"/>
+        <artifact name="\${org.wildfly:wildfly-webservices-server-integration}"/>
```
jamezp, line 36: Maybe it's not required, but we should remove this leading `\`.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/weld/webservices/main/module.xml
```diff
@@ -29,7 +29,7 @@
     </properties>
 
     <resources>
-        <artifact name="\${org.wildfly:wildfly-weld-webservices@module.jakarta.suffix@}"/>
+        <artifact name="\${org.wildfly:wildfly-weld-webservices}"/>
```
jamezp, line 32: Same comment as above.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/xts/main/module.xml
```diff
@@ -33,7 +33,7 @@
     </exports>
 
     <resources>
-	<artifact name="\${org.wildfly:wildfly-xts@module.jakarta.suffix@}"/>
+	<artifact name="\${org.wildfly:wildfly-xts}"/>
```
jamezp, line 36: And one more :)

weld/webservices/pom.xml
```diff
@@ -4,28 +4,63 @@
    <modelVersion>4.0.0</modelVersion>
 
    <parent>
+      <!--move back to weld-parent when other weld components moves to jakarta namespace-->
       <groupId>org.wildfly</groupId>
-      <artifactId>wildfly-weld-parent</artifactId>
+      <artifactId>wildfly-parent</artifactId>
```
jamezp, line 9: Is this because Weld has not been migrated yet?

jamezp (changes requested):
Overall I think this looks okay. I made a few comments on the `module.xml` files. I do wonder if we should leave the Weld one alone because the other Weld subsystems have not been converted.

weld/webservices/pom.xml
bstansberry: I think this is because @jimma is wanting to avoid both the parent dependencyManagement (legacy-ee) and this modules (standard-ee) both taking effect.

I've never tried but I'm pretty sure you can't use combine-children=override to prevent this, as that is for children of plugin 'configuration' elements.

bstansberry (changes requested):
@jimma The weld/pom.xml needs to be updated to handle this:

https://github.com/wildfly/wildfly/blob/main/weld/pom.xml#L76-L77

That is, create a 'modules' element outside that profile and add a child for webservices. Then delete those 2 lines.

jimma:
> @jimma The weld/pom.xml needs to be updated to handle this:
> 
> https://github.com/wildfly/wildfly/blob/main/weld/pom.xml#L76-L77
> 
> That is, create a 'modules' element outside that profile and add a child for webservices. Then delete those 2 lines.

Fixed. Thanks @bstansberry .

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/webservices/main/module.xml
jimma: Fixed.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/weld/webservices/main/module.xml
jimma: Fixed.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/xts/main/module.xml
jimma: Fixed.  Thanks for review @jamezp

weld/webservices/pom.xml
jimma: @bstansberry @jamezp  Yes. That's reason. The other weld modules are still import the legacy-ee dependencies, so I have to change the parent name . After the other weld modules are migrated to jakarta namespace, we can change back to `wildfly-weld-parent`.

PR #16166, opened 2022-10-12
[WFLY-16714]: Consider removing the embedded broker from the standard…

ehsavoie:
… WF Preview conf.

* replace the embedded broker in the standalone-full and standalone-full-ha configuration files with a remote one.

Jira: https://issues.redhat.com/browse/WFLY-16714


bstansberry:
@ehsavoie You'll need to remove this unless you want to do something complicated to change the config:

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/legacy/pom.xml#L80-L104

I vote for removing it; I only added it so I could run that testsuite module while we were working on the big bang.

bstansberry (approved):
@jfdenise FYI.

bstansberry:
Thanks @ehsavoie

PR #16172, opened 2022-10-14
[WFLY-17145] Fast maven builds with quick profile

xstefank:
https://issues.redhat.com/browse/WFLY-17145

Totally optional Friday fun idea :)

Allows also to build with mvnd!

My machine:

mvn clean install -DskipTests -> 05:46 min
mvn -Dquickly -> 04:47 min
**mvnd -Dquickly -> 01:50 min** (8 threads)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1972,12 +1972,32 @@
 
     <!-- Profiles -->
     <profiles>
+        <profile>
+            <id>quick-build</id>
+            <activation>
+                <property>
+                    <name>quickly</name>
+                </property>
+            </activation>
+            <properties>
+                <skipTests>true</skipTests>
+                <skipITs>true</skipITs>
+                <maven.javadoc.skip>true</maven.javadoc.skip>
+                <checkstyle.skip>true</checkstyle.skip>
+                <enforcer.skip>true</enforcer.skip>
+                <license.skip>true</license.skip>
+                <maven.test.skip>true</maven.test.skip>
+            </properties>
+            <build>
+                <defaultGoal>clean install</defaultGoal>
+            </build>
+        </profile>
 
         <profile>
             <id>ee8.build.profile</id>
             <activation>
                 <property>
-                    <name>!skip.ee8</name>
+                    <name>!quickly</name>
                 </property>
             </activation>
             <modules>
--- a/ee-9/pom.xml
+++ b/ee-9/pom.xml
@@ -124,7 +124,7 @@
             <id>skip.preview</id>
             <activation>
                 <property>
-                    <name>!skip.preview</name>
+                    <name>!quickly</name>
                 </property>
             </activation>
 
--- a/testsuite/integration/smoke/pom.xml
+++ b/testsuite/integration/smoke/pom.xml
@@ -633,6 +633,33 @@
     </build>
 
     <profiles>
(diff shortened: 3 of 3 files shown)


pom.xml
```diff
@@ -1972,6 +1972,26 @@
 
     <!-- Profiles -->
     <profiles>
+        <profile>
+            <id>quick-build</id>
+            <activation>
+                <property>
+                    <name>quickly</name>
```
bstansberry, line 1979: Instead of !skip.ee8 we could also use !quickly here:

https://github.com/wildfly/wildfly/blob/main/pom.xml#L1980

Building those modules is really pretty pointless now.

Perhaps here as well:

https://github.com/wildfly/wildfly/blob/main/ee-9/pom.xml#L127

Whether building WFP is appropriate for 'quick-build' is debatable. How appropriate it is varies over time, but my instinct is it's not really necessary to build it.

bstansberry: The WFP decision should be informed by data; i.e. how much time it adds.

bstansberry: Re both of the above thoughts, doing that would prevent using those profile on the CI servers.

It's a shame profile activation doesn't support and/or configuration.

xstefank: My machine feels better rested on Monday. I'm getting even better times :).

With both above changes 01:13 (199 modules). WFP included: 01:30 (207 modules), current state 01:29 (236 modules).

Most of the modules are compiled pretty quickly. Most time is spent in the galleon provisioning at the end. WFP is provisioned together with ee-dist and dist at the same time (different thread), so there is not so much difference.

xstefank: @bstansberry so I suppose we can go without WFP for now and if required we will add it back?

bstansberry: @xstefank +1.

TBH I don't know if we are actually using -Dskip.ee8 on CI jobs. Maybe I added the profile but never got around to changing CI to use it. If I had set CI to use it, I think what I asked Yeray to fix for bonus points in https://github.com/wildfly/wildfly/pull/16173#pullrequestreview-1142880733 would have resulted in obvious CI failures.

If we're not using it on CI, breaking it is fine. :)

xstefank:
@bstansberry updated, should be good to go.

bstansberry:
@xstefank can you post on the wildfly-dev mail list about this? Just something simple, 'hey, i had an idea, have a look at https://github.com/wildfly/wildfly/pull/16172'

I love it, but it's not 'get it merged today' urgent, so might as well socialize it a bit.

xstefank:
@bstansberry sure, I planned to tweet about it once merged too. But I can do it before it's merged too.

bstansberry:
@xstefank Thanks. I figured might as well give folks a chance to make any suggestions, etc. It's pretty straightforward though. :)

FYI, I pulled down the Team City configs and grepped for use of skip.ee8 and found none, so replacing it won't break anything. I checked just in case some of the nightly jobs, etc were using it.

bstansberry:
Thanks @xstefank!

rhusar:
+1 Looks good to me. I think by now everyone has a couple of aliases for a quick wildfly build, so having this directly managed by everybody in a `pom.xml` is great!

PR #16183, opened 2022-10-18
[WFLY-17153] Remove warning message WFLYWELD0052 in hibernate quickstart

skyllarr:
https://issues.redhat.com/browse/WFLY-17153

### DIFF
--- a/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/faces/api/main/module.xml
+++ b/ee-feature-pack/ee-10-api/src/main/resources/modules/system/layers/base/jakarta/faces/api/main/module.xml
@@ -42,5 +42,9 @@
         <module name="javax.validation.api" export="true"/>
         <module name="org.glassfish.jakarta.el" export="true"/>
         <module name="javax.websocket.api"/>
+        <!--TODO remove the WELD dependencies from this module, they should be in implementation module only. See comments on https://github.com/wildfly/wildfly/pull/16155 for possible causes-->
+        <module name="org.jboss.weld.api"/>
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
     </dependencies>
 </module>


bstansberry:
@Skyllarr @manovotn @jasondlee 

(Jason, the context here is the discussion on a similar PR starting at [4].)

I wonder if this is a long-standing problem. Mojarra provides a CDI extension and uses it to install bunch of producers.[1] Those use the 'beanClass(EE API class)' idiom, e.g. [2].  This looks much the same as the wildfly-elytron-ee code that's being updated at [3] to eliminate the need for these kinds of module deps in the EE Security API module.

I wonder if it's a long-standing problem because the Mojarra code in question has been there for a long time.

[1] https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/cdi/CdiExtension.java#L177
[2] https://github.com/eclipse-ee4j/mojarra/blob/master/impl/src/main/java/com/sun/faces/cdi/FacesContextProducer.java#L39 
[3] https://github.com/wildfly-security/wildfly-elytron-ee/pull/22
[4] https://github.com/wildfly/wildfly/pull/16155#issuecomment-1280901376

manovotn:
> Those use the 'beanClass(EE API class)' idiom, e.g. [2].

Note that what I described in the other PR is not a hard fact from CDI specification. The spec says next to nothing about `beanClass` and the javadoc doesn't mention synth beans at all[1]. Nonetheless, it is a handy way to determine which CL loads the bean in a modular environment (in case of synth beans anyway).

I think we could/should fix this on Mojarra side as well; I can author the PR tomorrow.

> I wonder if it's a long-standing problem because the Mojarra code in question has been there for a long time.

I am not aware of any Weld changes that would trigger this in the latest versions of WFLY. I wouldn't be surprised if this was around for a while and nobody noticed :)

[1] https://github.com/jakartaee/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/Bean.java#L37-L42

manovotn:
I've created Mojarra issue(https://github.com/eclipse-ee4j/mojarra/issues/5157) along with a PR(https://github.com/eclipse-ee4j/mojarra/pull/5158).
I also noticed one other suspicious usage of `beanClass` in their extension for which I created a separate issue(https://github.com/eclipse-ee4j/mojarra/issues/5159) but without PR since I don't know the spec/impl enough to see what exactly they need to achieve.

I should also correct myself because after some digging in the specification, I now think it is clearer what `beanClass` should stand for in case of syth beans. There is a [section](https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#inter_module_injection) in the specification that states how it works for synth beans and that this is needed for inter-module injection which is IMO akin to what we are seeing here.

jasondlee:
How important are the Mojarra changes for WF 27? We already have one change that we've had them merge, and I am about to ask for a release, but if this needs to be in 27, I'll push for this one as well before asking for a release. If not, I'll get the release and integrate it.

bstansberry:
@jasondlee This PR works around the issue, so unless @manovotn feels otherwise and you're comfortable approving the PR  there is no particular urgency to get a Mojarra fix for WF 27.

In both this and https://github.com/wildfly/wildfly/pull/16155 my main concern has been that we understand the issue before implementing a workaround, and at this point I feel like we do.

I am still curious whether the log WARN occurs with the WildFly 26 quickstarts, as that would further confirm that we understand the reported issue, but that checking that is a nice to have.

jasondlee (approved):

jasondlee:
That works for me. I've approved this, and I'll push for a Mojarra release. We'll get the Mojarra change linked here next time 'round.

manovotn:
> @jasondlee This PR works around the issue, so unless @manovotn feels otherwise and you're comfortable approving the PR there is no particular urgency to get a Mojarra fix for WF 27.

+1, I agree with Brian.

manovotn (approved):

bstansberry:
Thanks @Skyllarr @manovotn and @jasondlee

BalusC:
This is candidate to be reverted by https://github.com/eclipse-ee4j/mojarra/issues/5457. How exactly was it previously failing in WildFly?

PR #16184, opened 2022-10-18
[WFLY-17170] Drop suffix from transformed wildfly-elytron-integration artifact

jbliznak:
https://issues.redhat.com/browse/WFLY-17170

Drop -jakarta suffix from the transformed artifact and use new suffix for original artifact instead, until it is properly migrated to Jakarta namespace.

This PR is blocked by https://github.com/wildfly/wildfly-core/pull/5245 and corresponding update in this repo - so I expect CI will fail for now.

### DIFF
--- a/ee-9/feature-pack/pom.xml
+++ b/ee-9/feature-pack/pom.xml
@@ -522,10 +522,6 @@
                     <artifactId>jboss-invocation</artifactId>
                 </exclusion>
                 <exclusion>
-                    <groupId>org.wildfly.core</groupId>
-                    <artifactId>wildfly-elytron-integration</artifactId>
-                </exclusion>
-                <exclusion>
                     <groupId>org.wildfly.security</groupId>
                     <artifactId>wildfly-elytron-jacc</artifactId>
                 </exclusion>
@@ -573,12 +569,6 @@
             <artifactId>wildfly-core-feature-pack-galleon-common</artifactId>
             <type>pom</type>
             <scope>provided</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.wildfly.core</groupId>
-                    <artifactId>wildfly-elytron-integration</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
 
         <dependency>
--- a/ee-feature-pack/common/pom.xml
+++ b/ee-feature-pack/common/pom.xml
@@ -3174,7 +3174,7 @@
 
         <dependency>
             <groupId>org.wildfly.core</groupId>
-            <artifactId>wildfly-elytron-integration-jakarta</artifactId>
+            <artifactId>wildfly-elytron-integration</artifactId>
             <exclusions>
                 <exclusion>
                     <groupId>*</groupId>
--- a/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
+++ b/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
@@ -4469,7 +4469,7 @@
     </dependency>
     <dependency>
       <groupId>org.wildfly.core</groupId>
-      <artifactId>wildfly-elytron-integration-jakarta</artifactId>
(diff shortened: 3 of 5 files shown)


darranl:
I was hoping we could do this one slightly slower so we don't introduce the need for coordinated PRs.

I was hoping we could enter a phase where WildFly Core for a short while publishes two artefacts with the same content then WildFly can move over, then once WildFly has moved over WildFly Core can come back and remove the redundant artefact.

ee-9/feature-pack/pom.xml
```diff
@@ -523,7 +523,7 @@
                 </exclusion>
                 <exclusion>
                     <groupId>org.wildfly.core</groupId>
-                    <artifactId>wildfly-elytron-integration</artifactId>
+                    <artifactId>wildfly-elytron-integration-legacy-namespace</artifactId>
                 </exclusion>
```
bstansberry, line 527: This can be removed altogether. The https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/pom.xml doesn't declare this dependency.

Almost all these kinds of exclusions in this pom can be removed but that's out of scope for this!

ee-9/feature-pack/pom.xml
```diff
@@ -576,7 +576,7 @@
             <exclusions>
                 <exclusion>
                     <groupId>org.wildfly.core</groupId>
-                    <artifactId>wildfly-elytron-integration</artifactId>
+                    <artifactId>wildfly-elytron-integration-legacy-namespace</artifactId>
```
bstansberry, line 579: This can be removed altogether. The https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/galleon-common/pom.xml doesn't declare this dependency.

jbliznak:
@darranl no problem with your proposal either if you think that would mean less work in the end. The main purpose of my proposal was to make public change soon and then take whatever time we need to bake the actual implementation behind this public name.

If I understand your POV, your proposal would only make sense when the source transformation is already dropped in core otherwise instead of two artifacts (the same content under two GAVs) we would need 3 (or even 4) for some time to make the update in WF seamless. So then it is about the schedule when would dropping of source-transformation in core happen - if it is weeks or more, it could still make sense to proceed with this.

darranl:
Lets see what other reviews also say - they may be happy to work with it just sometimes coordinated PRs can be more painful if we can find a way to split out the sequence.

ee-9/feature-pack/pom.xml
jbliznak: Removed, thanks.

ee-9/feature-pack/pom.xml
jbliznak: Removed, thanks.

bstansberry:
> I was hoping we could do this one slightly slower so we don't introduce the need for coordinated PRs.
> 
> I was hoping we could enter a phase where WildFly Core for a short while publishes two artefacts with the same content then WildFly can move over, then once WildFly has moved over WildFly Core can come back and remove the redundant artefact.

If doing coordinated updates is a hassle for @yersan and publishing the same content in two artifacts is easier, that can work. The only use in full of org.wildfly.core:wildfly-elytron-integration are the two exclusions.

That said, I question whether this will prove easier, particularly since there are already PRs open.

yersan:
> That said, I question whether this will prove easier, particularly since there are already PRs open.

From WildFly Core we have another one open but it seems it doesn't interfere at all. For full, it will depend on how many PRs are in place, but it looks like there are only a few uses.

We can start preparing the integration Jobs, and see how they go.

yersan:
@jbliznak yes, that's the plan; I will supersede this PR moving the commits to the PR that bumps the wildfly-core version. 

From the discussion taken yesterday in this PR, I assumed we wanted this one, @bstansberry FYI in case you have any argument against this.

yersan:
Superseded by https://github.com/wildfly/wildfly/pull/16192

bstansberry:
Closing in favor of #16192. Thanks @jbliznak

PR #16185, opened 2022-10-18
WFLY-17165 TopologyChanged event listeners should not perform blocking operations on non-blocking thread

pferraro:
https://issues.redhat.com/browse/WFLY-17165

This is causing state transfer interruptions, failures, deadlocked threads, and hung servers which interrupt subsequent tests.

To summarize changes made in this PR:
* Simplify access to Infinispan's BlockingManager
* Ensure tasked composed from the CompletionStage returned via @PrePassivate events run on the non-blocking thread by leveraging Infinispan's BlockingManager to execute any blocking operations, including:
  * Cache operations
  * Access to synchronized data structures
  * Task submission to thread pools with blocking task queues
* Reimplement inactive session statistics recording entirely within wildfly-clustering-web-undertow, eliminating the need for a cache listener that performs redundant cache lookups that would block cache entry removal (within wildlfy-clustering-web-infinispan)
* Replace a number of queuing single thread pools with Infinispan's blocking thread pool
* Utilize @Listener(observation) to avoid unnecessary listener invocation
* Simplify registration of generic cache listener


rhusar (approved):
I didn't find any issues. Great stuff.

bstansberry:
Thanks @pferraro

PR #16189, opened 2022-10-19
[WFLY-17167]: LocalTransactionTestCase failure when running under -Dts.layers.

ehsavoie:
* Fixing the NullPointException with external pcf.
* Excluding tests that require a real broker.

Jira: https://issues.redhat.com/browse/WFLY-17167


darranl:
mvn install -Dts.layers is running for me locally with this change.

darranl (approved):
I have just manually checked the build log for the Galleon job for the string "test failures" and it is gone.

Thank you @ehsavoie for these changes, I see we have a test case to reproduce the underlying issue as well.

bstansberry:
Thanks @ehsavoie

PR #16193, opened 2022-10-19
WFLY-17174 Upgrade jakarta.transaction-api from 2.0.0 to 2.0.1

chengfang:
https://issues.redhat.com/browse/WFLY-17174

`legacy.version.jakarta.transaction` has the same value as `version.jakarta.transaction.jakarta-transaction-api`, so this PR also updates both props to keep them consistent.

`legacy.version.jakarta.transaction` should probably be removed, and use `version.org.jboss.spec.javax.transaction.jboss-transaction-api_1.3_spec` instead in `legacy-bom.xml`


bstansberry:
Thanks @chengfang

PR #16197, opened 2022-10-20
[WFLY-17111] WildFly documentation fixes

boris-unckel:
Fixes https://issues.redhat.com/browse/WFLY-17111
2nd part.


bstansberry:
Thanks @boris-unckel

PR #16208, opened 2022-10-26
WFLY-17199 Replace hsqldb with h2 for Elytron OCSP test and drop hsqldb dependency

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-17199


bstansberry:
@fjuma Can this dependency be replaced with H2? Looking at org.xipki.datasource.DataSourceWrapper it looks like it can handle H2.

BTW @soul2zimate I'm not suggesting not merging this. But if we can drop a dependency we should follow up doing that.

fjuma:
@bstansberry Yes, it does look like H2 could be used instead.

soul2zimate:
Yes, you're correct. I have rebased to replace `hsqldb` with `h2` for Elytron OCSP test and drop `hsqldb` dependency. Thanks.

bstansberry (approved):
Excellent! Thanks @soul2zimate

PR #16211, opened 2022-10-26
[WFLY-16755] Update javax.* uses in WS docs

bstansberry:
https://issues.redhat.com/browse/WFLY-16755 (webservices part)

@jimma @rsearls FYI


bstansberry:
The CI jobs fail because I cancelled them to save CI resources. CI doesn't test docs-only changes.

PR #16216, opened 2022-10-26
WFLY-17203 Cascaded eviction handling uses more threads than necessary

pferraro:
https://issues.redhat.com/browse/WFLY-17203

Follows up on fix for https://issues.redhat.com/browse/WFLY-17165


rhusar (changes requested):
This seems to cause the LocalFineSessionPassivationTestCase and LocalCoarseSessionPassivationTestCase fail consistently now.

PR #16219, opened 2022-10-28
[WFLY-17206] Fix NetworkHealthTestCase failures on Windows with IPv4

bstansberry:
https://issues.redhat.com/browse/WFLY-17206

PR CI jobs use -Dipv6 so I ran a custom job without that:

https://ci.wildfly.org/viewLog.html?buildId=333038&buildTypeId=WF_MasterWindowsJdk11

### DIFF
--- a/testsuite/integration/manualmode/pom.xml
+++ b/testsuite/integration/manualmode/pom.xml
@@ -107,6 +107,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-commons</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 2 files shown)


ehsavoie (commented):
Shouldn't we take advantage of this to remove the line test (if(TestSuiteEnvironment.isWindows()) { )

ehsavoie (approved):

bstansberry:
@ehsavoie 

> Shouldn't we take advantage of this to remove the line test (if(TestSuiteEnvironment.isWindows()) { )

On Linux I wanted to test the default values in the AttributeDefinitions, as they are meant to work on Linux and that's what the typical end user would use. So I limited it to just Windows, where we know the AD defaults don't work.

bstansberry:
@ehsavoie Before merging I updated the comment to explain more clearly my thinking on that. We can certainly change it though if the default AD values are properly tested in other ways.

PR #16223, opened 2022-10-28
[WFLY-17209] jboss-client.jar is missing jboss-modules classes which are required by jboss-ejb-client

ivassile:
Issue: https://issues.redhat.com/browse/WFLY-17209

### DIFF
--- a/client/shade/pom.xml
+++ b/client/shade/pom.xml
@@ -81,6 +81,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.logmanager</groupId>
+            <artifactId>jboss-logmanager</artifactId>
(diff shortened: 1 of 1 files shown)


ivassile:
@chengfang Let me know if we have to also add org.jboss.logmanager dependency as suggested by Brad on the jbeap issue.

chengfang:
> 

yes, please also add jboss-logmanager dependency.

ivassile:
@chengfang Thanks for confirming that we also need to add org.jboss.logmanager dependency. Modified the PR.

chengfang (approved):

bstansberry:
Thanks @ivassile

PR #16224, opened 2022-10-28
[WFLY-17210] Integration testsuite docs don't properly render single testcase examples

jessicarod7:
https://issues.redhat.com/browse/WFLY-17210


bstansberry:
Thanks @cam-rod

PR #16225, opened 2022-10-28
[WFLY-17198] Add the org.jboss.resteasy.microprofile.config module de…

jamezp:
…pendency to deployments. User should be able to set some RESTEasy specific configuration options with the MP Config API.

https://issues.redhat.com/browse/WFLY-17198
Signed-off-by: James R. Perkins <jperkins@redhat.com>

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/deployment/JaxrsDependencyProcessor.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/deployment/JaxrsDependencyProcessor.java
@@ -102,7 +102,6 @@ public class JaxrsDependencyProcessor implements DeploymentUnitProcessor {
         addDependency(moduleSpecification, moduleLoader, RESTEASY_CRYPTO, true, false);
         addDependency(moduleSpecification, moduleLoader, JACKSON_DATATYPE_JDK8, true, false);
         addDependency(moduleSpecification, moduleLoader, JACKSON_DATATYPE_JSR310, true, false);
-        addDependency(moduleSpecification, moduleLoader, MP_REST_CLIENT, true, false);
 
         final CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
         if (support.hasCapability(WELD_CAPABILITY_NAME)) {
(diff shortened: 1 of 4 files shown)


microprofile/galleon-common/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/config-smallrye/main/module.xml
```diff
@@ -42,5 +42,10 @@
         <module name="org.jboss.logging"/>
         <module name="org.eclipse.microprofile.config.api"/>
         <module name="javax.enterprise.api" />
+        <!-- Optional dependency which provides RESTEasy the ability to use the MP Config for its configuration options.
+             It also provides some additional MP Config sources. The subsystem itself does not depend on this module.
+             However, it does add the module to deployments which is why it is referenced here for the module graph.
+         -->
```
bstansberry, line 48: This isn't a reason to add something in a module.xml.

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileSubsystemDefinition.java
```diff
@@ -52,6 +54,7 @@ protected MicroProfileSubsystemDefinition(Iterable<ConfigSourceProvider> provide
                 .setAddHandler(new MicroProfileConfigSubsystemAdd(providers, sources))
                 .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
                 .setCapabilities(CONFIG_CAPABILITY)
+                .setAdditionalPackages(RuntimePackageDependency.optional("org.jboss.resteasy.microprofile.config"))
```
bstansberry, line 57: I don't think this will result in it being provisioned if layers are used. RuntimePackageDependency.optional should.

microprofile/galleon-common/src/main/resources/modules/system/layers/base/org/wildfly/extension/microprofile/config-smallrye/main/module.xml
jamezp: Makes sense really. I guess this was mostly for the `LayersTestCase` which isn't a valid reason :)

bstansberry (approved):

bstansberry:
Thanks @jamezp

PR #16230, opened 2022-10-31
Revert "[WFLY-17209] jboss-client.jar is missing jboss-modules classes which are required by jboss-ejb-client"

bstansberry:
Reverts wildfly/wildfly#16223

See https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/JBoss.20Modules.20in.20jboss-client.2Ejar

### DIFF
--- a/client/shade/pom.xml
+++ b/client/shade/pom.xml
@@ -81,11 +81,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.jboss.logmanager</groupId>
-            <artifactId>jboss-logmanager</artifactId>
-        </dependency>
-
-        <dependency>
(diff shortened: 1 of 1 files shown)


jamezp (approved):
Approving as a +1. I do think there is likely a fix required in the EJB Client itself, but having the jboss-logmanager and jboss-modules shared in the JAR is not correct IMO.

bstansberry:
BTW, even if we decide not to do this we'll do #16229.

bstansberry:
https://github.com/wildfly/jboss-ejb-client/pull/575 is the JBoss EJB Client fix.

PR #16235, opened 2022-11-02
[WFLY-17177] Configure surefire plugin's 'reportNameSuffix' when test will be executed multiple times

gaol:
Issue: https://issues.redhat.com/browse/WFLY-17177

If this PR is not for the 'main' branch you must add a link to the equivalent change in 'main'.

Remember to use the Jira issue ID in the PR title and any commits.

This PR tries to add `reportNameSuffix>EXECUTION-ID</reportNameSuffix>` to some execution configuration of `maven-surefire-plugin` to avoid the test report being overridden if the test gets executed multiple times.

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1544,6 +1544,7 @@
                                     <additionalClasspathElements>
                                         <additionalClasspathElement>${project.basedir}/../src/test/resources</additionalClasspathElement>
                                     </additionalClasspathElements>
+                                    <reportNameSuffix>basic-integration-default-web</reportNameSuffix>
                                 </configuration>
                             </execution>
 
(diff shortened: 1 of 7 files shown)


bstansberry (approved):

gaol:
It looks like the `Windows - JDK 11` always fails, not sure if it relates, I will try to check it in a Windows machine this week or next week.

gaol:
I tried to run the tests from a vanilla WildFly main branch, and I got the same errors, so the failure does not relate to this change. 

I checked the build log of  `Windows - JDK 11` for other PRs, which is marked as success, but actually it failed, I think the reason the were marked as success is because the same maven surefire report file was used. This PR actually reveals the problem.

I created https://issues.redhat.com/browse/WFLY-17698 for the test failure, and I plan to add extra commit to this PR to fix it as well, please let me know if it needs a separate PR for it.

gaol:
Added the commit to this PR for https://issues.redhat.com/browse/WFLY-17698

bstansberry:
Thanks @gaol!!!

PR #16242, opened 2022-11-03
[WFLY-17229] Add WildFlySecurityManager dependency to the testsuite/layers module

lvydra:
https://issues.redhat.com/browse/WFLY-17229

### DIFF
--- a/testsuite/layers/pom.xml
+++ b/testsuite/layers/pom.xml
@@ -38,6 +38,11 @@
             <artifactId>wildfly-launcher</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.wildfly.security</groupId>
+            <artifactId>wildfly-elytron-security-manager</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.wildfly</groupId>
             <artifactId>wildfly-testsuite-shared</artifactId>
             <exclusions>


testsuite/layers/pom.xml
```diff
@@ -37,6 +37,10 @@
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-launcher</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-security-manager</artifactId>
+        </dependency>
```
yersan, line 43: This dependency represents the WildFly Core Security manager subsystem, which transitively will incorporate the SecurityManager class into the test scope.

I would say it could be more appropriate to use the following one, which is more specific for this use-case:
```
        <dependency>
            <artifactId>wildfly-elytron-security-manager</artifactId>
            <scope>test</scope>
        </dependency>
```

However, I don't have any strong argument to suggest this change, so I'll defer the review to someone that can point out a better argument about when we should use one and when the other. In general, we should minimize the dependencies between subsystems, so I am more inclined to use the Elytron one, but I have no strong arguments to support this.

In any case, this change fixes the issue we found in wildfly-core.

bstansberry: We should use the exact dependency that gets the job done. We also should not rely on transitive dependencies. And if test scope is appropriate we should use that.

Over the years our poms got quite confusing, which added quite a bit of extra effort to various work we've done these last couple years (getting rid of picketbox, javax->jakarta, etc). We cleaned that up quite a bit as part of the big-bang, so a big +1 to being as precise as possible going forwards.

yersan: Just a short note, adding the `<scope>test</scope>` was just a personal intuitive configuration since we know the dependency is used only for running the tests. Since this module doesn't have non-test-related code, it is not mandatory for it, but I would recommend being as narrow as we can.

darranl: Now also with all of the work we have done regarding layers we really need to minimise inter-subsystem / extension dependencies and where we can use appropriate APIs / SPIs instead of direct dependencies.  I know this is only a test later but if we started to do this in other places we could accidentally be pulling in an extension which is about to be deprecated and removed in the foreseeable future.

lvydra:
Thanks @yersan @bstansberry @darranl, updated.

bstansberry:
Thanks @lvydra

PR #16245, opened 2022-11-03
[WFLY-17231] Update HostExcludesTestCase configuration to work with WF28

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-17231

We have added some extensions during the WF27 development cycle, the `HostExcludesTestCase` tracks those forcing us to re-create an additional enum in the test case itself when we move forward to the next release.

It looks like `org.wildfly.extension.datasources-agroal` was removed in WildFly 14 I think we moved it to WildFly Preview) but we added it back later in WF27 to the standard configuration.


bstansberry (approved):

bstansberry:
@yersan There was still a failure so I sent up #16247 to get CI passing without needing to rush this.

yersan:
Ok, thanks @bstansberry , ignoring it is the best, we need to release wildfly-core first to completely fix this one

bstansberry (approved):

bstansberry:
Thanks @yersan

PR #16250, opened 2022-11-03
[WFLY-17237] Drop source transformation for opentelemetry subsystem

jasondlee:
https://issues.redhat.com/browse/WFLY-17237

Update dependencies for base extension
Remove source transform modules
Rename files
Update imports


boms/legacy-ee/pom.xml
```diff
@@ -243,12 +243,12 @@
...
-                <artifactId>wildfly-opentelemetry-legacy-namespace</artifactId>
+                <artifactId>wildfly-opentelemetry</artifactId>
                 <version>${full.maven.version}</version>
             </dependency>
             <dependency>
                 <groupId>${full.maven.groupId}</groupId>
-                <artifactId>wildfly-opentelemetry-api-legacy-namespace</artifactId>
+                <artifactId>wildfly-opentelemetry-api</artifactId>
                 <version>${full.maven.version}</version>
             </dependency>
```
bstansberry, line 253: Both of these should just be deleted. Anything that needs the two non-legacy artifacts shouldn't be relying on this bom to manage that dep. And I don't think anything is doing that.

observability/opentelemetry-api/pom.xml
```diff
@@ -33,48 +33,63 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>wildfly-standard-ee-bom</artifactId>
+                <version>${ee.maven.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
bstansberry, line 50: The observability/pom.xml has a dependencyManagement import of wildfly-legacy-ee-bom. That can either be deleted, and all the child modules control that, as you do here, or you can change that to wildfly-standard-ee-bom and drop this whole block.

observability/opentelemetry/pom.xml
```diff
@@ -33,152 +33,100 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>wildfly-standard-ee-bom</artifactId>
+                <version>${ee.maven.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
bstansberry, line 50: Same comment that I made in opentelemetry-api/pom.xml applies here.

bstansberry (changes requested):

observability/opentelemetry-api/pom.xml
jasondlee: I tried removing that, but that breaks the micrometer build, and we trying to avoid changing that module as part of this work, though I guess it's safe/clean enough. I'll do that if you're ok with it.

bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #16258, opened 2022-11-04
[WFLY-17254] Add a global client interceptors for EJB deployments that will runAs the current security identity to activate any outflow identities

fjuma:
https://issues.redhat.com/browse/WFLY-17254

This PR updates the EJB3 subsystem so that it always registers appropriate client interceptors automatically for EJB deployments to ensure that we always runAs the current security identity in order to activate any outflow identities.

In particular, the following changes have been made:

* The EJB3 subsystem automatically registers a global `org.jboss.ejb.client.EJBClientInterceptor` for EJB deployments. This interceptor will be used when invoking remote EJBs.
* The EJB3 subsystem automatically registers a global `org.jboss.invocation.Interceptor` for EJB deployments. This interceptor will be used when invoking local EJBs. Note that the only reason this additional interceptor is needed is because global `EJBClientInterceptors` don't get invoked for local EJBs due to [WFLY-9976](https://issues.redhat.com/browse/WFLY-9976).

Depends on https://github.com/wildfly/wildfly-core/pull/5263

### DIFF
--- a/ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
+++ b/ee-9/source-transform/ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
@@ -206,6 +206,7 @@ public class InterceptorOrder {
 
     public static final class Client {
 
+        public static final int SECURITY_IDENTITY = 0x001;
         public static final int TO_STRING = 0x100;
         public static final int NOT_BUSINESS_METHOD_EXCEPTION = 0x110;
         public static final int LOCAL_ASYNC_LOG_SAVE = 0x180;
--- a/ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
@@ -206,6 +206,7 @@ public class InterceptorOrder {
 
     public static final class Client {
 
+        public static final int SECURITY_IDENTITY = 0x001;
         public static final int TO_STRING = 0x100;
         public static final int NOT_BUSINESS_METHOD_EXCEPTION = 0x110;
         public static final int LOCAL_ASYNC_LOG_SAVE = 0x180;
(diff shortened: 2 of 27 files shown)


darranl:
@fjuma Looks great - can you please hack in a temporary commit with a hard coded phase ID?  We should be able to get a CI run and then drop that commit once we are ready.

It may be better in a separate PR but maybe we need to start on some test cases for some identity propagation scenarios - I think this should also pick up programmatic authentication, do you know if that is the case?

fjuma:
Thanks @darranl, I've added a temporary commit with the phase ID changes.

Yes, this should cover the programmatic authentication case as well. I'm going to start on tests cases and will include this.

fjuma:
I have updated this PR with identity propagation test cases that cover the following scenarios:

* servlet to EJB invocations within a single deployment with different security domains
* servlet to EJB invocations across deployments with different security domains
* servlet to EJB to EJB invocations across deployments with different security domains

The above cases cover invocations on remote EJBs and local EJBs and there are test cases that make use of programmatic authentication as well.

The test cases that involve invocations on local EJBs are currently failing due to [WFLY-9976](https://issues.redhat.com/browse/WFLY-9976) (global EJB client interceptors don't currently get invoked for local EJBs).

fjuma:
I have now updated this PR so that invocations on local EJBs work as well. More details are in the PR description above.

tadamski (approved):
LGTM
@fjuma As we discussed, I think having this two interceptors is currently the best way to make this work. I was thinking that maybe sharing some parts of interceptor code in an utility class would be beneficial but taking into consideration that the interceptors code isn't very large it may not be worth the effort after all.

fjuma:
@tadamski The `EJBClientInterceptor#handleInvocation` method and the `Interceptor#processInvocation` have different parameter types and different return values so I didn't see a good way to simplify it further.

darranl (approved):

fjuma:
Thank you both for the reviews. Since CI passed, I'm going to remove the temporary commit that was hardcoding the phase ID changes and I'm going to update this so it's no longer a draft PR.

bstansberry:
Thanks @fjuma, @darranl and @tadamski!

PR #16265, opened 2022-11-07
WFLY-17276 FD_SOCK2 should not configure external address/port if no socket-binding was defined

pferraro:
https://issues.redhat.com/browse/WFLY-17276


bstansberry (approved):

bstansberry:
Thanks @pferraro

PR #16269, opened 2022-11-09
[WFLY-17281]:Set release version to maven compiler plugin

jimma:
https://issues.redhat.com/browse/WFLY-17281

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,8 +150,7 @@
           -->
         <maven.min.version>3.6.0</maven.min.version>
         <!-- Require Java 11 -->
-        <maven.compiler.target>11</maven.compiler.target>
-        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.release>11</maven.compiler.release>
 
         <!-- TarArchiver behavior -->
         <assembly.tarLongFileMode>posix</assembly.tarLongFileMode>


pom.xml
```diff
@@ -150,8 +150,7 @@
           -->
         <maven.min.version>3.6.0</maven.min.version>
         <!-- Require Java 11 -->
-        <maven.compiler.target>11</maven.compiler.target>
-        <maven.compiler.source>11</maven.compiler.source>
```
jamezp, line 154: I think we need to keep these. While they'll be ignored if `<release/>` is used, but we don't explicitly set that on the `maven-compiler-plugin`. However, we do explicitly set the `<target/>` and `<source/>` settings. By default too the `jdk.min.version` is set to `maven.compiler.argument.source` which is set from `maven.compiler.source`.

jamezp (changes requested):

bstansberry:
@jamezp @jimma Is this a case of needing to restore the removed lines, PLUS add the new line in the PR? Or should this just be closed?

Either way, lets get it out of the PR queue. :)

jamezp:
I think we should do it, but I also think we should keep the other properties. I created #16503 as a replacement.

PR #16271, opened 2022-11-09
WFLY-8718: remove xa-datasource-class from JDBC driver definition

michpetrov:
Issue: [WFLY-8718](https://issues.redhat.com/browse/WFLY-8718)

The attribute definition itself was kept in since it used in xa-datasource but it shouldn't be included in the definition of a driver (which already has DRIVER_XA_DATASOURCE_CLASS_NAME). This has no impact on schema definition.


tadamski (changes requested):
@michpetrov could you please also deprecate the related attribute so that we can remove it ultimately?

michpetrov:
@tadamski which one do you want to deprecate? The one I removed is still used in XaDataSourceDefinition where I assume it is necessary, it was just wrongly included in the driver definition.

tadamski:
@michpetrov true, my mistake, approving

tadamski (approved):

PR #16274, opened 2022-11-10
[WFLY-17286] Eliminating unused version.org.jboss.byteman property from clustering TS pom.xml

ropalka:
https://issues.redhat.com/browse/WFLY-17286


rhusar (approved):
LGTM 👍  

...kind of looks like this wasn't even used previously.

bstansberry:
Thanks @ropalka

PR #16276, opened 2022-11-11
[WFLY-17268] Drop source transformation for the weld/subsytem module

bstansberry:
https://issues.redhat.com/browse/WFLY-17268


weld/subsystem/src/main/java/org/jboss/as/weld/deployment/AlternativePropertyReplacingBeansXmlParser.java
```diff
@@ -0,0 +1,48 @@
...
+import org.jboss.metadata.property.PropertyReplacer;
+import org.jboss.weld.xml.BeansXmlParser;
+
+/**
+ * Temporary alternative to {@link PropertyReplacingBeansXmlParser} that can work with Weld 5 API and can therefore
+ * choose between current and legacy behavior when parsing empty beans.xml files.
+ * See also https://github.com/eclipse-ee4j/cdi/issues/500
+ *
+ * This class will no longer be needed once Weld 5 becomes a direct dependency for non-preview WFLY.
+ * The functionality should then be merged into {@link PropertyReplacingBeansXmlParser}
```
bstansberry, line 34: Note that I intend to deal with this in a follow up commit, but I don't want the basic work to block on it.

If I get a chance to do that before this PR is approved, then I'll add that to this PR, otherwise I'll use a separate PR.

It's not a big deal to do; it's just been a busy week so haven't gotten to it.

manovotn:
@bstansberry IIUIC, this means that you can now fully depend on Weld 5/CDI 4, along with build compatible extensions.
In that case, you might also want to take a look at these places:

* https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldPortableExtensionProcessor.java#L61-L63
* https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/discovery/WeldClassFileInfo.java#L130
* https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanArchiveProcessor.java#L89

And on a slightly different note, this piece can also be removed as there is a release that contains the change:
* https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/WeldDeployment.java#L305-L313

All of the above should be trivial changes that I think you could include here as well.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/AlternativePropertyReplacingBeansXmlParser.java
manovotn: Whichever way suits you, but I wouldn't mind it being part of this PR along with the changes I mentioned in my comment so that we get that out of the way and don't forget it.

bstansberry:
Thanks @manovotn -- I had time today so I made all those changes.

manovotn (approved):
Looking good, thanks @bstansberry!

PR #16280, opened 2022-11-14
WFLY-17288: deprecate driver-name attribute in JDBC driver definition…

michpetrov:
…, adjust installed-drivers definitions

Issue: [WFLY-17288](https://issues.redhat.com/browse/WFLY-17288)

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
@@ -32,8 +32,8 @@ import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_DAT
 import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_MAJOR_VERSION;
 import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_MINOR_VERSION;
 import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_MODULE_NAME;
-import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_NAME;
 import static org.jboss.as.connector.subsystems.datasources.Constants.DRIVER_XA_DATASOURCE_CLASS_NAME;
+import static org.jboss.as.connector.subsystems.datasources.Constants.INSTALLED_DRIVER_NAME;
 import static org.jboss.as.connector.subsystems.datasources.Constants.JDBC_COMPLIANT;
 import static org.jboss.as.connector.subsystems.datasources.Constants.MODULE_SLOT;
 import static org.jboss.as.connector.subsystems.datasources.GetDataSourceClassInfoOperationHandler.dsClsInfoNode;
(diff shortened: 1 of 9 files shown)


darranl (changes requested):
This looks like it could need a transformer (and a model bump) if this attribute is moving from required to optional as it looks like it is used on a resource definition?

This would mean when managing older secondary host controllers their resource definitions will require that this is set.

tadamski (changes requested):
I agree that this change is good and necessary, but also agree that the transformer is needed.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
```diff
@@ -59,18 +57,13 @@ public class GetInstalledDriverOperationHandler implements OperationStepHandler
...
-    private final ParametersValidator validator = new ParametersValidator();
-
     private GetInstalledDriverOperationHandler() {
-        validator.registerValidator(DRIVER_NAME.getName(), new StringLengthValidator(1));
     }
 
     @Override
     public void execute(final OperationContext context, final ModelNode operation) throws OperationFailedException {
 
-        validator.validate(operation);
```
bstansberry, line 71: This should not be removed. A ParameterValidator should be used for input check, not ModelNode.require.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
```diff
@@ -59,18 +57,13 @@ public class GetInstalledDriverOperationHandler implements OperationStepHandler
...
-
-        final String name = operation.require(DRIVER_NAME.getName()).asString();
+        final String name = operation.require(INSTALLED_DRIVER_NAME.getName()).asString();
```
bstansberry, line 66: Using ModelNode.require is an anti-pattern, although it did no harm here as the validator already confirmed the parameter.

ModelNode.get should be used.

bstansberry:
@darranl @tadamski -- I agree with @michpetrov about not needing a transformer. Going back to WF 12, the only thing done with the driver-name param in the add op was to validate that if it was set that it didn't differ from the resource name. So not passing a value to a the secondary HC will work fine.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
bstansberry: @michpetrov Adding a validator to an AttributeDefinition doesn't cause it to be used. For it to be used, methods like AD.validateAndSet or AD.validateOperation need to be called. For the typical ADs that are used for resource attributes/add op params/write-attribute values, devs often don't see those calls because the base classes like AbstractAddStepHandler take care of that part.  But this is a custom OSH, so it needs to validate its inputs.

That said, I see now the old ParametersValidator thing was unnecessary. Changing L63 to this should handle it:

```
final String name = INSTALLED_DRIVER_NAME.validateOperation(operation).asString();
```

bstansberry (changes requested):
The GetInstalledDriverOperationHandler validation needs to be restored; otherwise LGTM.

michpetrov:
@bstansberry thanks, fixed and rebased

bstansberry (approved):

bstansberry:
Thanks @michpetrov and @tadamski

idafalco:
In administration console of wildfly28, in Add Datasource i can't add the driver. While in older version of wildfy there was a dropdown menu 'driver name' in which we can select the Driver. The driver association fails and the test connection fails. How should I do?

PR #16275, opened 2022-11-11
[WFLY-16869] Drop source transformation for the Transactions subsystem (and other modules)

jmfinelli:
This PR should be considered after [WFLY-17264](https://issues.redhat.com/browse/WFLY-17264) is merged

[WFLY-17290](https://issues.redhat.com/browse/WFLY-17290)
[WFLY-17291](https://issues.redhat.com/browse/WFLY-17291)
[WFLY-17265](https://issues.redhat.com/browse/WFLY-17265)
[WFLY-17258](https://issues.redhat.com/browse/WFLY-17258)
[WFLY-17095](https://issues.redhat.com/browse/WFLY-17095)
[WFLY-16869](https://issues.redhat.com/browse/WFLY-16869)
[WFLY-17261](https://issues.redhat.com/browse/WFLY-17261)

PR #16287, opened 2022-11-15
[WFLY-17095] Drop source transformation for messaging-activemq subsystem

bstansberry:
https://issues.redhat.com/browse/WFLY-17095

This is a cherry-pick of a commit from #16275. That PR is blocked by WFLY-17264, which in turn is blocked by this one.

@jmfinelli FYI


ehsavoie (changes requested):
A lot of the changes are not related to the transformation but are formatting. Could this be moved to another commit (on the same PR) so we can see the real impact changes as opposed to just formatting ones.

bstansberry:
@ehsavoie Thanks. The formatting changes I believe come from two things:

1) Copying the generated (transformed) source to the primary source tree. It seems the source generation resulted in formatting changes, e.g. replacement of FQCNs with simple class names. This is not something done by Manuel.

2) Copy the pom dependency entries from the source-transform pom to replace the ones in old pom.

Both of these are the most accurate way of doing this as they are a simple move of known-good source to a different location. Trying to do some more complicated port for easier review would be more likely to introduce bugs.

ehsavoie (approved):
I'll fix my remarks in a following PR

bstansberry:
Thanks @jmfinelli

PR #16298, opened 2022-11-18
WFLY-17313 Distributed TimerService fails when cache is configured with jdbc-store

pferraro:
https://issues.redhat.com/browse/WFLY-17313

Adds missing key formatters.


rhusar (approved):
LGTM

PR #16300, opened 2022-11-18
[WFLY-16990]: The broker roles are not displayed in the proper format.

ehsavoie:
* Updating the way roles are displayed to have coherence between json and not json list

Jira: https://issues.redhat.com/browse/WFLY-16990


bstansberry (approved):
Technically this should require and API version bump as it adds runtime-only attributes, which is an API change. But since it only affects read-resource[-description] results and it's unlikely anyone is careful about rejecting output with 'new' attributes, it should be ok. Plus you told me some of the new data was already being returned.

bstansberry:
Thanks @ehsavoie

PR #16302, opened 2022-11-20
[WFLY-17315] Remove uses of deprecated ExtensionContext.isRegisterTra…

bstansberry:
…nsformers()

https://issues.redhat.com/browse/WFLY-17315


bstansberry:
Note that to reduce the diff I tried to leave the meaningful transformer code in the same source file where it previously was (in a new static class), but ServiceLoader loading wouldn't work when I did that, so I had to move it out to a new source file.

Also, I expect a lot of this transformation code is cruft and can just be removed, but I didn't want to think about that for this task.

yersan (approved):
LGTM. At first glance, there is no need to keep the transformers for these "legacy" extensions since they have not changed since WF23.

@bstansberry I'm not sure why there were not removed as part of https://issues.redhat.com/browse/WFLY-15061.
Do you know by any chance if there is any reason why this set was not taken into account for such a cleanup Jira? Otherwise, I can create a Jira to do such a cleanup if it proceeds.

bstansberry:
Thanks @yersan. I agree they could just be removed. I'm pretty sure not doing it was just an oversight.

PR #16307, opened 2022-11-23
WFLY-17297 Replace uses of ServiceName SocketBinding.JBOSS_BINDING_NAME with use of the capability

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-17297

### DIFF
--- a/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPRootDefinition.java
+++ b/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPRootDefinition.java
@@ -22,7 +22,6 @@
 
 package org.wildfly.iiop.openjdk;
 
-import static org.wildfly.iiop.openjdk.Capabilities.IIOP_CAPABILITY;
 import static org.wildfly.iiop.openjdk.Capabilities.LEGACY_SECURITY;
 
 import java.util.ArrayList;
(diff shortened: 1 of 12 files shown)


bstansberry (dismissed):
Hi @soul2zimate 

Extensions should not reference capability instances from other extensions or from the server kernel, as that breaks the code decoupling goal of capabilities. Only the string names of capabilities should be used.

The way both of these subsystems should work would be by using OperationContext.getCapabilityServiceTarget and using CapabilityServiceTarget.addCapability to create a CapabilityServiceBuilder for services that depend on other capabilities (i.e. socket bindings).

The IIOP subsystem has an IIOP_CAPABILITY that may be suitable for this. I don't see one in the RTS subsystem, so we need to create one.

Note also that IIOPRootDefinition SOCKET_BINDING and SSL_SOCKET_BINDING should declare  setCapabilityReference in their builder. Not doing so is a bug.

Same is true for RTSSubsystemDefinition SOCKET_BINDING, HOST and SERVER.

I think we might be better off filing a "properly use capabilities" JIRA for RTS, as the scope of work there goes well beyond WFLY-17297's focus. Perhaps for IIOP as well, as using capabilities properly there should also result in changes in how dependencies for some of the security related things are wired as well.

soul2zimate:
Thanks, I have rebased to use `CapabilityServiceBuilder`

The failed tests in `Linux with security manager - JDK 11` and `Windows - JDK 11` are irrelevant to this change.

soul2zimate:
This blocks https://github.com/wildfly/wildfly-core/pull/5281. Hi @bstansberry, Could you review the latest change ?

darranl:
@bstansberry I am going to dismiss your review as changes made and ask the subsystem developers to review.

@soul2zimate FYI if making changes to multiple subsystems it is easier to get reviews if you submit as separate PRs, to merge this we need both to agree however the subsystems likely could have been merged separatly.

soul2zimate:
I have created some sub-tasks for the origianl issue and opened three different PRs https://github.com/wildfly/wildfly/pull/16708, https://github.com/wildfly/wildfly/pull/16709 https://github.com/wildfly/wildfly/pull/16710 for each subsystem

bstansberry:
Thanks so much for this, and for being patient, @soul2zimate! I'm closing this as AIUI the other 3 PRs replaced this one.

PR #16313, opened 2022-11-28
[WFLY-17328] Remove uses of deprecated AbstractAddStepHandler methods

bstansberry:
https://issues.redhat.com/browse/WFLY-17328

This consists of ensuring that the capabilities provided by a resource are registered via its ResourceDefinition, i.e. via the SimpleResourceDefinition.Parameters object used to construct the ResourceDefinition. If that is done, having the add handler record the capability is redundant.

In most of the cases here, the ResourceDefinition was already recording the capability, so the redundant recording by the add handler could simply be removed.


fjuma (approved):
Thanks @bstansberry!

tadamski (approved):

chengfang (approved):

PR #16315, opened 2022-11-28
WFLY-17330 Remove unused nu.xom, org.dom4j, org.jaxen modules and remove Hibernate ORM 6 dependency on dom4j

scottmarlow:
Signed-off-by: Scott Marlow <smarlow@redhat.com>

https://issues.redhat.com/browse/WFLY-17330


scottmarlow:
Pending test results passing, will make one more change to also remove modules `nu.xom` + `org.jaxen`

scottmarlow:
Looks like a legit failure in org.jboss.as.test.layers.LayersTestCase:

>  Running org.jboss.as.test.layers.LayersTestCase
> [ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 223.362 s <<< FAILURE! - in org.jboss.as.test.layers.LayersTestCase
> [ERROR] test(org.jboss.as.test.layers.LayersTestCase)  Time elapsed: 216.515 s  <<< ERROR!
> java.lang.Exception: ERROR, some modules have not been provisioned: nu.xom,org.dom4j,org.jaxen,
>   at org.jboss.as.test.layers.LayersTest.test(LayersTest.java:133)
>   at org.jboss.as.test.layers.LayersTestCase.test(LayersTestCase.java:317)
>   at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>   at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
>   at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>   at java.base/java.lang.reflect.Method.invoke(Method.java:566)
>   at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)

bstansberry:
@scottmarlow The solution to the test failure is to remove the nu.xom, org.dom4j and org.jaxen modules. That test is written to detect cases where we provide unused modules, so it has correctly detected that removing the org.dom4j dep from org.hibernate has resulted in unused modules. :)

bstansberry:
I suggest updating WFLY-17330 to also note it will result in the removal of the 3 modules.

scottmarlow:
> I suggest updating WFLY-17330 to also note it will result in the removal of the 3 modules.

Thanks, will remove the other unused modules also update the issue.  :)

scottmarlow:
Pushed too soon as testsuite still uses xom

bstansberry:
@scottmarlow Please restore the dependencyManagement entries that were in boms/common-ee but put them in the root pom. For dom4j and jaxen make them 'test' scope. (The xom one is used by testsuite/shared, which is not a test module itself; it's a utility lib, so its compile deps can't be test scope.) Then we don't specify the versions in the various child poms as they are controlled by dependency management.

See the comments before various other external libs that are declared in the root pom for the kind of comment that should precede these 3 as well. Basically an explanation for why these are in the root pom to make it easier to do something better in the future.

When we do https://issues.redhat.com/browse/WFLY-17331 we'll move these kinds of things out of the root pom.

scottmarlow:
> @scottmarlow Please restore the dependencyManagement entries that were in boms/common-ee but put them in the root pom. For dom4j and jaxen make them 'test' scope. (The xom one is used by testsuite/shared, which is not a test module itself; it's a utility lib, so its compile deps can't be test scope.) Then we don't specify the versions in the various child poms as they are controlled by dependency management.
> 
> See the comments before various other external libs that are declared in the root pom for the kind of comment that should precede these 3 as well. Basically an explanation for why these are in the root pom to make it easier to do something better in the future.
> 
> When we do https://issues.redhat.com/browse/WFLY-17331 we'll move these kinds of things out of the root pom.

Thanks for the detailed comment!  I made the requested change.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow!

PR #16320, opened 2022-11-29
[WFLY-17336] Eliminating usage of deprecated org.jboss.msc.inject.MapInjector

ropalka:
https://issues.redhat.com/browse/WFLY-17336

### DIFF
--- a/mail/src/main/java/org/jboss/as/mail/extension/DirectMailSessionService.java
+++ b/mail/src/main/java/org/jboss/as/mail/extension/DirectMailSessionService.java
@@ -40,7 +40,7 @@ public class DirectMailSessionService extends MailSessionService {
     private final SessionProvider provider;
 
     public DirectMailSessionService(final SessionProvider provider) {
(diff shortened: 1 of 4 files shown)


mail/src/main/java/org/jboss/as/mail/extension/DirectMailSessionService.java
```diff
@@ -40,7 +40,7 @@ public class DirectMailSessionService extends MailSessionService {
     private final SessionProvider provider;
 
     public DirectMailSessionService(final SessionProvider provider) {
-        super(null);
+        super(null, null);
```
bstansberry, line 43: This is probably out of scope for this, but this being a subclass is clunky. AFAICT this class and MailSessionService are actually completely distinct impls of Service<Session>.

@yersan fyi

bstansberry (commented):
This LGTM.

bstansberry:
Thanks @ropalka

PR #16329, opened 2022-11-30
WFLY-17347 Fix race conditions in Undertow subsystem tests causing them to fail frequently.

pferraro:
https://issues.redhat.com/browse/WFLY-17347

### DIFF
--- a/undertow/src/test/java/org/wildfly/extension/undertow/AbstractUndertowSubsystemTestCase.java
+++ b/undertow/src/test/java/org/wildfly/extension/undertow/AbstractUndertowSubsystemTestCase.java
@@ -89,7 +89,7 @@ public abstract class AbstractUndertowSubsystemTestCase extends AbstractSubsyste
           return properties;
       }
 
-    public static void testRuntime(KernelServices mainServices, final String virtualHostName, int schemaVersion) throws Exception {
+    public static void testRuntime(KernelServices mainServices, final String virtualHostName, int schemaVersion) throws InterruptedException {
         if (!mainServices.isSuccessfulBoot()) {
             Throwable t = mainServices.getBootError();
             Assert.fail("Boot unsuccessful: " + (t != null ? t.toString() : "no boot error provided"));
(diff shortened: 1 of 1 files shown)


bstansberry:
@pferraro I put the hold label on this because I'm uncomfortable with what seems like a workaround that obscures a serious problem. Barring some other event occurring that has destabilized the service container, the service container should be stable when a management op that modifies it completes. That's quite a fundamental thing for WildFly's use of MSC and if it's not working correctly we need to get to the bottom of that.

@yersan @ropalka and I will be meeting to talk about this on Friday.

pferraro:
@bstansberry That's fine - though these test failures are certainly creating a lot of CI noise.

Here are my observations so far...
Many of the services installed by the Undertow subsystem have a somewhat unique characteristic that the service for a child resource "registers" with its parent service.  e.g. UndertowService has a collection of Servers, which has a collection of Hosts, which has a collection of Locations, which has a collection of filters, etc.
However, rather than an add operation for a child resource triggering a restart of the services created by its parent resource, such that the value provided by the parent's service is immutable, the Undertow subsystem's services are all mutable, that is, rather than the parent collecting all known children, each child service depends on its parent's service, and registers itself with its parent. This means that, when the service of the parent resource is UP, a process querying the parent might find that it has no registered children, if the services of its child resources have not yet started.  These are the types of failures that we are seeing on the CI - and have been for the past 7 months or so.  Indeed, the CI runs for this PR ran into another instance of this issue:
https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk17/339292?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true
... since I neglected to address another situation where a parent verifies that it contains a specific number of children without ensuring that the service of the child resource has started.

Is it possible that the aberrant behavior of these Undertow subsystem tests is due to the fact that it uses a dummy ControlledProcessStateService?  I don't fully understand the role this might play - but it seemed to me an odd thing to stub out.

bstansberry:
> This means that, when the service of the parent resource is UP, a process querying the parent might find that it has no registered children, if the services of its child resources have not yet started. These are the types of failures that we are seeing on the CI - and have been for the past 7 months or so.

This is what I want to get to the bottom of. If "the services of its child resources have not yet started", either 1) the expected service container stability following completion of a management op isn't being achieved, or 2) the child resource services are doing things outside the start/stop lifecycle (e.g. registering with the parent), or 3) the test is messing with services in invalid ways. I'm concerned it's the first case.

Note that we are seeing an increased number of seemingly similar problems in other tests, particularly in WF Core.

Thanks for the tip about ControlledProcessStateService. I haven't looked but perhaps that's relevant to why this particular test module is so problematic. That said, even if CPSS didn't exist, the service container should be stable following completion of the management op.

yersan:
> 1) the expected service container stability following completion of a management op isn't being achieved

I also concluded about this behavior when I took a look at https://issues.redhat.com/browse/WFCORE-6149

pferraro:
I've update this to additionally fix the failure revealed by https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk17/339292?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

bstansberry:
Thanks @pferraro

PR #16331, opened 2022-12-01
[WFLY-17350] Custom mail providers are not loaded

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-17350

The getSession() method was changed to use the ~~"org.jboss.as.mail"~~ "jakarta.mail.api" classloader to prevent any caller module be able to load the session. In our Jakarta EE development cycle, we found "org.jboss.as.naming" was under this category. 

However, changing the TCCL in getSession left the application classloader outside without being able to provide other Mail providers.

Now we change the strategy and add the Mail API providers available on "org.jboss.as.naming" and roll back the TCCL changes.

### DIFF
--- a/servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/naming/main/module.xml
+++ b/servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/naming/main/module.xml
@@ -51,5 +51,14 @@
         <!-- It is used only if org.wildfly.undertow.http-invoker capability is available -->
         <module name="io.undertow.core" optional="true"/>
         <module name="sun.jdk" />
+        <!--
+        The Angus Mail services are required to be able to load the Jakarta Mail Session from this Module context classloader
+        when listing the Mail context stored in the JNDI tree.
+        -->
+        <module name="org.eclipse.angus.mail" services="import" optional="true">
+            <imports>
+                <include path="META-INF"/>
+            </imports>
+        </module>
     </dependencies>
 </module>
--- a/mail/src/main/java/org/jboss/as/mail/extension/SessionProviderFactory.java
+++ b/mail/src/main/java/org/jboss/as/mail/extension/SessionProviderFactory.java
@@ -168,14 +168,7 @@ class SessionProviderFactory {
 
         @Override
         public Session getSession() {
-            final Session session;
-            final ClassLoader current = Thread.currentThread().getContextClassLoader();
-            try {
-                Thread.currentThread().setContextClassLoader(Session.class.getClassLoader());
-                session = Session.getInstance(properties, new ManagedPasswordAuthenticator(sessionConfig));
-            } finally {
-                Thread.currentThread().setContextClassLoader(current);
-            }
+            final Session session = Session.getInstance(properties, new ManagedPasswordAuthenticator(sessionConfig));
             return session;
         }
     }
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/providers/MailSessionCustomProviderTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/providers/MailSessionCustomProviderTestCase.java
@@ -1,0 +1,68 @@
+/*
+ *  JBoss, Home of Professional Open Source.
+ *  Copyright 2022, Red Hat, Inc., and individual contributors
+ *  as indicated by the @author tags. See the copyright.txt file in the
+ *  distribution for a full listing of individual contributors.
+ *
+ *  This is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU Lesser General Public License as
+ *  published by the Free Software Foundation; either version 2.1 of
+ *  the License, or (at your option) any later version.
+ *
+ *  This software is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ *  Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this software; if not, write to the Free
+ *  Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ *  02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.test.integration.mail.providers;
+
+import jakarta.annotation.Resource;
+import jakarta.mail.Provider;
+import jakarta.mail.Session;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+public class MailSessionCustomProviderTestCase {
+
+    @Resource(lookup = "java:jboss/mail/Default")
+    private Session session;
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class)
+                .addAsResource(MailSessionCustomProviderTestCase.class.getPackage(),"javamail.providers", "META-INF/javamail.providers")
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    public void testCustomProvidersInApplication() {
+        Assert.assertNotNull(session);
+
+        Provider[] providers = session.getProviders();
+
+        int found = 0;
+        for (Provider p : providers) {
+            if (p.toString().equals("jakarta.mail.Provider[TRANSPORT,CustomTransport,org.jboss.qa.management.mail.custom.CustomTransport,JBoss QE]")) {
+                found++;
+            }
+            if (p.toString().equals("jakarta.mail.Provider[STORE,CustomStore,org.jboss.qa.management.mail.custom.CustomStore,JBoss QE]")) {
+                found++;
+            }
+        }
+
+        Assert.assertTrue("The expected custom providers cannot be found on the default Mail Session", found == 2);
+    }
+}
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/providers/javamail.providers
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/providers/javamail.providers
@@ -1,0 +1,2 @@
+protocol=CustomTransport; type=transport; class=org.jboss.qa.management.mail.custom.CustomTransport; vendor=JBoss QE;
+protocol=CustomStore; type=store; class=org.jboss.qa.management.mail.custom.CustomStore; vendor=JBoss QE;


yersan:
Let me hold it a bit since I want to add a test case

bstansberry:
@yersan I don't understand the PR description. I see how changing the previous change of the TCCL would have broken loading providers from the deployment. But...

1) AFAICT the TCCL was changed to the jakarta.mail.api module classloader, not to org.jboss.as.mail.
2) I don't understand what the org.jboss.as.naming module has to do with it.

If we end up with this fix, the dependency in org.jboss.as.naming needs to be optional. It also will need an explanatory comment.

yersan:
Hi @bstansberry , yes, my explanation was poor and I wrote the wrong module name.

> AFAICT the TCCL was changed to the jakarta.mail.api module classloader, not to org.jboss.as.mail.

Yes, I mistakenly wrote "org.jboss.as.mail" but was "jakarta.mail.api" which is the module that provides the Session.class


> I don't understand what the org.jboss.as.naming module has to do with it

The naming subsystem is able to list the JNDI tree. The Mail Session object is stored at `java:jboss/mail/Default` and the naming subsystem, by using `/subsystem=naming:jndi-view()` has the ability to list the Mail naming contexts and show its values.

In order to get a Mail Session value when listing the JNDI tree from the naming subsystem, the Jakarta Mail API needs to load the `[StreamProvider](https://github.com/eclipse-ee4j/angus-mail/blob/master/core/src/main/resources/META-INF/services/jakarta.mail.util.StreamProvider)` implementation which is supplied by Angus Mail.

This implementation is loaded from the context class loader, which is `org.jboss.as.naming` JBoss Modules module in this case. This `StreamProvider` usage is new in Jakarta Mail, for this reason, this was not needed before.

To be able to get a Session instance, we need to make the Mail Angus service implementations visible to `org.jboss.as.naming` JBoss Modules module, hence the change for this module.

> It also will need an explanatory comment.

You're right, I should add a comment on the module since this behavior is a bit obscure. Even more obscure because `org.jboss.as.naming` has indeed no dependencies to `org.jboss.as.mail` which is the module that makes access to methods that load the Session.

I don't have experience with naming, I see there are several points where modules are loaded on demand, so I suspect that this module is somehow loaded by the naming subsystem. In any case, that was the old behavior in WF26, so I assume we are fine here. 

Another point with minor doubts on my side is why we have to create an Object Session when `java:jboss/mail/Default` is retrieved, but again this has been the behavior in the past so I have not dug into it and I assumed that is correct and expected, probably the Mail specification can tell us about this point.

servlet-feature-pack/common/src/main/resources/modules/system/layers/base/org/jboss/as/naming/main/module.xml
```diff
@@ -51,5 +51,14 @@
         <!-- It is used only if org.wildfly.undertow.http-invoker capability is available -->
         <module name="io.undertow.core" optional="true"/>
         <module name="sun.jdk" />
+        <!--
+        The Angus Mail services are required to be able to load the Jakarta Mail Session from this Module context classloader
+        when listing the Mail context stored in the JNDI tree.
```
bstansberry, line 56: Is this issue primarily about listing, or is the listing aspect just a detail of a test? IOW is the core problem here that using the provider doesn't work? Or is it just that the list of Provider objects (which afaict are just metadata holders) from Session.getProviders() is inaccurate?

yersan: Hi @bstansberry , maybe showing the error trace I can make it more clear what the issue is. We get this error trace when listing the JNDI view using the naming subsystem:

`[standalone@localhost:9990 /] /subsystem=naming:jndi-view()`

The error is:

```
16:30:28,591 ERROR [org.jboss.as.controller.management-operation] (management-handler-thread - 1) WFLYCTL0013: Operation ("jndi-view") failed - address: ([("subsystem" => "naming")]): java.lang.IllegalStateException: Not provider of jakarta.mail.util.StreamProvider was found
    at jakarta.mail.api@2.1.1.jbossorg-1//jakarta.mail.util.FactoryFinder.find(FactoryFinder.java:64)
    at jakarta.mail.api@2.1.1.jbossorg-1//jakarta.mail.util.StreamProvider.provider(StreamProvider.java:186)
    at jakarta.mail.api@2.1.1.jbossorg-1//jakarta.mail.Session.<init>(Session.java:254)
    at jakarta.mail.api@2.1.1.jbossorg-1//jakarta.mail.Session.getInstance(Session.java:308)
    at org.jboss.as.mail@28.0.0.Beta1-SNAPSHOT//org.jboss.as.mail.extension.SessionProviderFactory$ManagedSession.getSession(SessionProviderFactory.java:171)
    at org.jboss.as.mail@28.0.0.Beta1-SNAPSHOT//org.jboss.as.mail.extension.MailSessionService.getValue(MailSessionService.java:73)
    at org.jboss.as.mail@28.0.0.Beta1-SNAPSHOT//org.jboss.as.mail.extension.MailSessionManagedReferenceFactory.getReference(MailSessionManagedReferenceFactory.java:56)
    at org.jboss.as.mail@28.0.0.Beta1-SNAPSHOT//org.jboss.as.mail.extension.MailSessionManagedReferenceFactory.getInstanceClassName(MailSessionManagedReferenceFactory.java:50)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.ServiceBasedNamingStore.list(ServiceBasedNamingStore.java:189)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.NamingContext.list(NamingContext.java:354)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.NamingContext.list(NamingContext.java:370)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.management.JndiViewOperation.addEntries(JndiViewOperation.java:122)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.management.JndiViewOperation.addEntries(JndiViewOperation.java:136)
    at org.jboss.as.naming@28.0.0.Beta1-SNAPSHOT//org.jboss.as.naming.management.JndiViewOperation$1.execute(JndiViewOperation.java:72)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.AbstractOperationContext.executeStep(AbstractOperationContext.java:1045)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.AbstractOperationContext.processStages(AbstractOperationContext.java:777)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.AbstractOperationContext.executeOperation(AbstractOperationContext.java:466)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.OperationContextImpl.executeOperation(OperationContextImpl.java:1433)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.ModelControllerImpl.internalExecute(ModelControllerImpl.java:449)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.ModelControllerImpl.lambda$executeForResponse$0(ModelControllerImpl.java:260)
    at org.wildfly.security.elytron-base@2.0.0.Final//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:304)
    at org.wildfly.security.elytron-base@2.0.0.Final//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:270)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.ModelControllerImpl.executeForResponse(ModelControllerImpl.java:260)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.ModelControllerImpl.executeOperation(ModelControllerImpl.java:254)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.ModelControllerImpl.execute(ModelControllerImpl.java:237)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler.doExecute(ModelControllerClientOperationHandler.java:241)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:163)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1$1.run(ModelControllerClientOperationHandler.java:159)
    at org.wildfly.security.elytron-base@2.0.0.Final//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:328)
    at org.wildfly.security.elytron-base@2.0.0.Final//org.wildfly.security.auth.server.SecurityIdentity.runAs(SecurityIdentity.java:285)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.AccessAuditContext.doAs(AccessAuditContext.java:254)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.AccessAuditContext.doAs(AccessAuditContext.java:225)
    at org.jboss.as.controller@20.0.0.Beta3//org.jboss.as.controller.remote.ModelControllerClientOperationHandler$ExecuteRequestHandler$1.execute(ModelControllerClientOperationHandler.java:159)
    at org.jboss.as.protocol@20.0.0.Beta3//org.jboss.as.protocol.mgmt.ManagementRequestContextImpl$1.doExecute(ManagementRequestContextImpl.java:70)
    at org.jboss.as.protocol@20.0.0.Beta3//org.jboss.as.protocol.mgmt.ManagementRequestContextImpl$AsyncTaskRunner.run(ManagementRequestContextImpl.java:160)
    at org.jboss.threads@2.4.0.Final//org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
    at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:1990)
    at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1486)
    at org.jboss.threads@2.4.0.Final//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1377)
    at java.base/java.lang.Thread.run(Thread.java:829)
    at org.jboss.threads@2.4.0.Final//org.jboss.threads.JBossThread.run(JBossThread.java:513)
```

The main problem is any caller module that needs to load a Mail Session Object now needs to see the Angus Mail services because in order to get a Session instance, we need to have an implementation of `StreamProvider`, which is provided by Angus Mail.

Firstly, we fixed it by explicitly using the "jakarta.mail.api" module as the TCCL, but that solution left out the other classloaders where the user can add other mail providers.

Now, since WildFly needs to load the session object when the naming subsystem lists the JNDI thread, we have added the Angus Mail services to the `org.jboss.as.naming` module.

Another third solution, which could be more reliable, could be to use a combined class loader when we are getting the session. For example a combination between the caller classloader and the "jakarta.mail.api" classloader.

bstansberry: Thanks. I understood the reported problem; what I wasn't clear about is whether this was a symptom of a much bigger problem. We want to fix the reported problem, of course, and I'm happy to merge this to fix that problem. But jndi-view is a fairly minor feature so I want to understand if there is a bigger issue that affect application behavior, not just the management layer.

yersan: In my opinion, there should not be an issue with the application behavior, since both, the Jakarta Mail API and the Angus Mail implementation, are dependencies visible to the application classpath.

PR #16347, opened 2022-12-06
WFLY-17348 A distributed web application request coinciding with leaving of the primary session owner can cause an NPE

pferraro:
https://issues.redhat.com/browse/WFLY-17348


rhusar (approved):
LGTM

PR #16350, opened 2022-12-06
[WFLY-17365] Failed deployment is not removed from Undertow ServletContainer's deployments map

aogburn:
https://issues.redhat.com/browse/WFLY-17365

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentService.java
@@ -106,6 +106,9 @@ public class UndertowDeploymentService implements Service<UndertowDeploymentServ
                 HttpHandler handler = deploymentManager.start();
                 Deployment deployment = deploymentManager.getDeployment();
                 host.get().registerDeployment(deployment, handler);
+            } catch (Throwable e ) {
+                container.get().getServletContainer().removeDeployment(deploymentInfo);
+                throw e;
             } finally {
                 StartupContext.setInjectionContainer(null);
             }


ropalka (changes requested):
Please move this line to startContext() method.

aogburn:
> Please move this line to startContext() method.

Done.

ropalka (approved):

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentService.java
```diff
@@ -106,6 +106,9 @@ public void startContext() throws ServletException {
                 HttpHandler handler = deploymentManager.start();
                 Deployment deployment = deploymentManager.getDeployment();
                 host.get().registerDeployment(deployment, handler);
+            } catch (Throwable e ) {
```
bstansberry, line 109: @aogburn I don't understand how this compiles in CI but this method can't throw Throwable. This would need to be catch (RuntimeException | ServletException e).

bstansberry (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentService.java
bstansberry: Never mind; my IDE was complaining about this yesterday but seems to have gotten over it. The try block can only throw RuntimeException, ServletException or Error so the compiler allows it.

bstansberry (approved):

bstansberry:
Thanks @aogburn

PR #16353, opened 2022-12-07
WFLY-17371 Deprecate comment attribute of session-cookie resource

pferraro:
https://issues.redhat.com/browse/WFLY-17371

Also introduces enumeration of supported subsystem model versions, reducing redundant ModelVersion instances for resources/attribute deprecations, transformers, tests, etc.

### DIFF
--- a/undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
+++ b/undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
@@ -319,6 +319,7 @@ undertow.setting.session-cookie.http-only=Is cookie http-only
 undertow.setting.session-cookie.secure=Is cookie secure?
 undertow.setting.session-cookie.domain=Cookie domain
 undertow.setting.session-cookie.comment=Cookie comment
+undertow.setting.session-cookie.comment.deprecated=Ignored. Cookie comments were deprecated for removal as of version 6.0 of the Jakarta Servlet specification, per RFC 6265.
 undertow.setting.session-cookie.max-age=Max age of cookie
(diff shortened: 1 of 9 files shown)


undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
```diff
@@ -319,6 +319,7 @@ undertow.setting.session-cookie.http-only=Is cookie http-only
 undertow.setting.session-cookie.secure=Is cookie secure?
 undertow.setting.session-cookie.domain=Cookie domain
 undertow.setting.session-cookie.comment=Cookie comment
+undertow.setting.session-cookie.comment.deprecated=Cookie comments were deprecated for removal in version 6.0 of the Jakarta Servlet specification.
```
rhusar, line 322: Let's be more descriptive here and save users some time and explain (1) the RFC part of this (i.e. RFC 6265) and (2) that IIUC this is not a typical deprecation, since it actually has no effect (`jakarta.servlet.http.Cookie#setComment` is simply a no-op). Something like:

> Cookie comments were deprecated for removal in version 6.0 of the Jakarta Servlet specification as they are no longer required with RFC 6265 and the value is ignored.

rhusar (changes requested):

undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
pferraro: Done.

rhusar (approved):
Looks good to me, thanks.

PR #16372, opened 2022-12-13
[WFLY-17404] Upgrade RESTEasy to 6.2.2.Final

jamezp:
https://issues.redhat.com/browse/WFLY-17404

Tag: https://github.com/resteasy/resteasy/releases/tag/6.2.2.Final
Diff: https://github.com/resteasy/resteasy/compare/6.2.1.Final...6.2.2.Final


asoldano (approved):

jamezp:
https://issues.redhat.com/browse/RESTEASY-3247 needs to be thought over again. When the `org.jboss.resteasy.microprofile.config`is present, it will load the `org.jboss.resteasy.microprofile.config.ConfigConfiguration`. However, if the `microprofile-smallrye-config` subsystem is not present, the implementation cannot be found.

jamezp:
I've pushed a second commit, but I'm not 100% sure this is how we want to handle it. It some ways it does make sense to handle this way. It works and makes sense to allow different configurations per-deployment though.

bstansberry (approved):

bstansberry:
Thanks @jamezp and @asoldano

PR #16380, opened 2022-12-15
[WFLY-17408] Manage the plexus-utils dependency in full WildFly's own…

soul2zimate:
… dependencyManagement

Issue: https://issues.redhat.com/browse/WFLY-17408

related WFCORE PR: https://github.com/wildfly/wildfly-core/pull/5328

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -467,6 +467,7 @@
         <version.org.bitbucket.jose4j>0.8.0</version.org.bitbucket.jose4j>
         <version.org.bytebuddy>1.12.18</version.org.bytebuddy>
         <version.org.cryptacular>1.2.5</version.org.cryptacular>
+        <version.org.codehaus.plexus.plexus-utils>3.5.0</version.org.codehaus.plexus.plexus-utils>
         <version.org.codehaus.woodstox.stax2-api>4.2.1</version.org.codehaus.woodstox.stax2-api>
         <version.org.codehaus.woodstox.woodstox-core>6.4.0</version.org.codehaus.woodstox.woodstox-core>
         <version.org.eclipse.angus.angus-activation>1.0.0</version.org.eclipse.angus.angus-activation>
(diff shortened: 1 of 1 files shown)


boms/common-ee/pom.xml
```diff
@@ -771,6 +771,13 @@
                 <version>${version.com.fasterxml.jackson.jr.jackson-jr-objects}</version>
             </dependency>
 
+            <dependency>
+                <groupId>org.codehaus.plexus</groupId>
+                <artifactId>plexus-utils</artifactId>
+                <version>${version.org.codehaus.plexus.plexus-utils}</version>
+                <scope>test</scope>
+            </dependency>
```
bstansberry, line 779: Please move this into the root pom. This bom should only include production code dependencies. I'd like to create a testbom in full WF like we have in WF Core[1], so until we do that I'm leaving test deps in the root pom and add comments to them to note they are testsuite-only.

[1] https://issues.redhat.com/browse/WFLY-17412

pom.xml
```diff
@@ -599,6 +599,7 @@
         <version.org.reactivestreams>1.0.4</version.org.reactivestreams>
         <version.org.syslog4j>0.9.30</version.org.syslog4j>
         <!-- Test dependency -->
+        <version.org.codehaus.plexus.plexus-utils>3.5.0</version.org.codehaus.plexus.plexus-utils>
```
bstansberry, line 602: Please move this alphabetically with the others. The 'Test dependency' comment is unneeded IMHO. Documentation of why we have a particular dependency is better placed next to the dependencyManagement entry for it.

If you can also move the version.org.keycloak.... property into the correct alphabetical position I'd appreciate it.

bstansberry (changes requested):
Thanks for handling this, @soul2zimate! I have a couple comments.

pom.xml
soul2zimate: fixed, Thanks

boms/common-ee/pom.xml
soul2zimate: fixed, I read from pom.xml there is https://issues.redhat.com/browse/WFLY-17331 already.

soul2zimate:
retest this please

bstansberry:
Thank you, @soul2zimate!

PR #16381, opened 2022-12-15
[WFLY-17256] Drop source transformation for the microprofile-jwt-smal…

bstansberry:
…lrye subsystem

Since this is the last microprofile subsystem to convert, convert the artifactId of the microprofile module and prune the ee-9/source-transform/microprofile tree

https://issues.redhat.com/browse/WFLY-17256


fjuma (approved):
Thanks @bstansberry!

PR #16382, opened 2022-12-15
WFLY-17411 Fix incorrect MSC usage by ModClusterResource

pferraro:
https://issues.redhat.com/browse/WFLY-17411

Summary of changes:
* Modifies runtime operations to use captured service values (based on logic from WFLY-12601), rather than relying on Service.getValue()
* Replaces all runtime Resource classes related to mod_cluster with a single dynamic Resource implementation that consumes captured ModCluster service values
* Removes static references to ModCluster-related resource definitions


pferraro:
Rebased against main.

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterNodeDefinition.java
```diff
@@ -49,306 +65,251 @@
...
-            protected void handleNode(OperationContext context, ModClusterStatus.Node ctx, ModelNode operation) throws OperationFailedException {
-                context.getResult().set(new ModelNode(ctx.getTtl()));
+            public ModelNode execute(ModClusterStatus.Node node) {
+                return new ModelNode(node.getUri().toString());
             }
-        });
-        resourceRegistration.registerReadOnlyAttribute(FLUSH_PACKETS, new AbstractNodeOperation() {
-
+        },
+        ALIASES(Constants.ALIASES, ModelType.LIST, PrimitiveListAttributeDefinition.Builder::new) {
```
rhusar, line 161: This inadvertently results in model type LIST of value type LIST instead of intended STRING.

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterBalancerDefinition.java
```diff
@@ -43,147 +54,111 @@
...
-        resourceRegistration.registerSubModel(ModClusterNodeDefinition.INSTANCE);
-        resourceRegistration.registerSubModel(ModClusterLoadBalancingGroupDefinition.INSTANCE);
-    }
-
-    @Override
-    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
-        resourceRegistration.registerReadOnlyAttribute(MAX_ATTEMPTS, new AbstractBalancerOperation() {
-
+    enum LoadBalancerMetric implements Metric<ModClusterStatus.LoadBalancer> {
+        STICKY_SESSION_COOKIE(Constants.STICKY_SESSION_COOKIE, ModelType.STRING) {
```
rhusar, line 58: So these were previously not annotated as access type `metrics`, but rather `read-only`. I think this is fine, but can we confirm this is the intention, since some of these are  not 'metrics' per se.

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterNodeDefinition.java
```diff
@@ -49,306 +65,251 @@
...
-            protected void handleNode(OperationContext context, ModClusterStatus.Node ctx, ModelNode operation) throws OperationFailedException {
-                context.getResult().set(new ModelNode(ctx.getOpenConnections()));
+            public ModelNode execute(ModClusterStatus.Node node) {
+                return new ModelNode(node.getRequestQueueSize());
             }
-        });
-        resourceRegistration.registerReadOnlyAttribute(PING, new AbstractNodeOperation() {
-
+        },
+        TIMEOUT(Constants.TIMEOUT, ModelType.INT) {
```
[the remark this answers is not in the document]
pferraro: Nice catch.  Fixing...

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterBalancerDefinition.java
pferraro: Good point.  Let's leave that change for another time.  I'll revert this to normal attribute registration.

rhusar: Sounds good, thanks.

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterNodeDefinition.java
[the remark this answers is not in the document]
pferraro: Fixed

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterNodeDefinition.java
pferraro: Fixed.

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterBalancerDefinition.java
pferraro: Fixed.

rhusar (approved):
👍🏼 Looks good now.

I have opened https://issues.redhat.com/browse/WFLY-17568 to track the uncovered metrics issue above.

PR #16387, opened 2022-12-16
Fix the "quickly" build profile to ensure it works in clean environme…

jamezp:
…nts.


pferraro:
Why change the behavior of `-Dquickly` to instead set `maven.test.skip.exec` to `true`?

jamezp:
> Why change the behavior of `-Dquickly` to instead set `maven.test.skip.exec` to `true`?

Yes. I'm going to try other options for this. What I did works, but it's too verbose. I was trying to rush through it and didn't think enough about it :)

PR #16408, opened 2022-12-17
WFLY-17438 Eliminate static references to JSF subsystem ResourceDefinition instances

pferraro:
https://issues.redhat.com/browse/WFLY-17438


pferraro:
Of ffs - I already did this one...

PR #16416, opened 2022-12-19
Small changes to make the -Dquickly build work on a clean repository.…

jamezp:
… Add a job to ensure it continues working.

The property `maven.test.skip` skips the `org.apache.maven.plugins:maven-jar-plugin::test-jar` goal as well as compiling tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2079,7 +2079,6 @@
                 <checkstyle.skip>true</checkstyle.skip>
                 <enforcer.skip>true</enforcer.skip>
                 <license.skip>true</license.skip>
-                <maven.test.skip>true</maven.test.skip>
             </properties>
             <build>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -2079,7 +2079,6 @@
                 <checkstyle.skip>true</checkstyle.skip>
                 <enforcer.skip>true</enforcer.skip>
                 <license.skip>true</license.skip>
-                <maven.test.skip>true</maven.test.skip>
```
xstefank, line 2082: yeah, better to leave this out. For even quicker builds when just server needs to be rebuilt we can specify this ourselves.

xstefank (approved):

pferraro (approved):

PR #16427, opened 2023-01-02
[WFLY-16868] Drop the source-transform variant of the ee subsystem

bstansberry:
Note that the brings the EE 10 Concurrency impl from the source-transform tree.
    
The source-transform/ee module wasn't really doing source transform; it was a complete code fork.

https://issues.redhat.com/browse/WFLY-16868

Requires #16429


bstansberry:
@emmartins Does the EE Concurrency stuff in the ee/src/main tree in this PR look ok?  This PR is basically a replacement of the source in ee/ with what was in ee-9/source-transform/ee, so it should be fine, but I figure if anyone would notice something odd it would be you. :)

emmartins:
I didn't notice anything wrong and the checks here are all good, but perhaps just confirm TCK is still green and if that's the case please go ahead with the merge.

bstansberry:
Thanks @emmartins -- I ran the standalone concurrency TCK and it passed.

PR #16429, opened 2023-01-03
[WFLY-17269] Drop source transformation for the weld/common and weld/…

bstansberry:
…spi modules

https://issues.redhat.com/browse/WFLY-17269


manovotn (approved):
LGTM, I also ran your branch against Weld TS to be sure and found no issues.

PR #16430, opened 2023-01-03
[WFLY-17464] Asciidoctor attribute sourceHighlighter position

boris-unckel:
Fix https://issues.redhat.com/browse/WFLY-17464


bstansberry:
Thanks @boris-unckel

PR #16438, opened 2023-01-05
WFLY-17471 write-attribute operation handler used by clustering subsystems apply wrong values to runtime during rollback

pferraro:
https://issues.redhat.com/browse/WFLY-17471


rhusar (approved):
Nice catch, LGTM.

PR #16439, opened 2023-01-05
[WFLY-17473] Eliminating usage of deprecated org.jboss.msc.service.ValueInjectionService

ropalka:
https://issues.redhat.com/browse/WFLY-17473

PR #16443, opened 2023-01-06
[WFLY-17477] Refactoring - eliminating usage of deprecated org.jboss.msc.service.ValueService

ropalka:
https://issues.redhat.com/browse/WFLY-17477

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/structure/ComponentAggregationProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/structure/ComponentAggregationProcessor.java
@@ -32,8 +32,11 @@ import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.server.deployment.module.ResourceRoot;
-import org.jboss.msc.service.ValueService;
-import org.jboss.msc.value.ImmediateValue;
+import org.jboss.msc.service.Service;
+import org.jboss.msc.service.ServiceName;
+import org.jboss.msc.service.ServiceBuilder;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
 
 import java.util.List;
 import java.util.Map;
@@ -57,9 +60,10 @@ public final class ComponentAggregationProcessor implements DeploymentUnitProces
             return;
         }
 
-        phaseContext.getServiceTarget().addService(ComponentRegistry.serviceName(deploymentUnit), new ValueService<>(new ImmediateValue<Object>(componentRegistry)))
-                .addDependency(moduleDescription.getDefaultClassIntrospectorServiceName(), EEClassIntrospector.class, componentRegistry.getClassIntrospectorInjectedValue())
-                .install();
+        final ServiceName sn = ComponentRegistry.serviceName(deploymentUnit);
+        final ServiceBuilder<?> sb = phaseContext.getServiceTarget().addService(sn);
+        sb.setInstance(new ValueService(componentRegistry));
+        sb.addDependency(moduleDescription.getDefaultClassIntrospectorServiceName(), EEClassIntrospector.class, componentRegistry.getClassIntrospectorInjectedValue()).install();
 
         deploymentUnit.putAttachment(COMPONENT_REGISTRY, componentRegistry);
 
@@ -116,6 +120,25 @@ public final class ComponentAggregationProcessor implements DeploymentUnitProces
             for (final ComponentDescription componentDescription : deploymentUnit.getAttachmentList(org.jboss.as.ee.component.Attachments.ADDITIONAL_RESOLVABLE_COMPONENTS)) {
                 applicationDescription.addComponent(componentDescription, deploymentRoot.getRoot());
             }
+        }
+    }
+
+    private static final class ValueService implements Service<ComponentRegistry> {
+        private final ComponentRegistry value;
+        public ValueService(final ComponentRegistry value) {
+            this.value = value;
+        }
+
+        public void start(final StartContext context) {
+            // noop
+        }
+
+        public void stop(final StopContext context) {
+            // noop
+        }
+
+        public ComponentRegistry getValue() {
+            return value;
         }
     }
 }
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
@@ -153,12 +153,13 @@ import org.jboss.ejb.client.EJBTransportProvider;
 import org.jboss.javax.rmi.RemoteObjectSubstitutionManager;
 import org.jboss.metadata.ejb.spec.EjbJarMetaData;
 import org.jboss.msc.inject.Injector;
+import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
-import org.jboss.msc.service.ValueService;
-import org.jboss.msc.value.ImmediateValue;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StopContext;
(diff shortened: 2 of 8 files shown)


chengfang:
ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJBRemoteInvocationPassByValueWriteHandler.java and 
ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemDefaultPoolWriteHandler.java

for the above 2 classes, looks like the same changes are also included in the other PR #16439 

Can you check and if so limit those changes to one PR?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -612,4 +614,29 @@ private static void addClusteringServices(final OperationContext context, final
             builder.setInstance(new SingletonBarrierService(policy)).setInitialMode(ServiceController.Mode.ON_DEMAND).install();
         }
     }
+
+    public final class ValueService<T> implements Service<T> {
```
chengfang, line 618: this may be `private static` class, if not used elsewhere.

Do you also plan to update this deprecated `Service` class to `org.jboss.msc.Service`?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ExceptionLoggingWriteHandler.java
```diff
@@ -72,9 +75,33 @@ void updateOrCreateDefaultExceptionLoggingEnabledService(final OperationContext
...
             // create and install the service
-            final ValueService<AtomicBoolean> service = new ValueService<>(new ImmediateValue<>(new AtomicBoolean(enabled)));
-            context.getServiceTarget().addService(serviceName, service)
-                    .install();
+            final ServiceBuilder<?> sb = context.getServiceTarget().addService(serviceName);
+            sb.setInstance(new ValueService(() -> new AtomicBoolean(enabled))).install();
+        }
+    }
+
+    public final class ValueService<T> implements Service<T> {
```
chengfang, line 83: this may be private static class, if not used elsewhere.

Do you also plan to update this deprecated Service class to org.jboss.msc.Service?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
ropalka: Good catch @chengfang - I fixed it. There is no plan to migrate to o.j.m.Service now. It will be handled in the future.

ropalka:
> Can you check and if so limit those changes to one PR?
Fixed @chengfang - thanks for identifying this collision!

chengfang (approved):

rhusar:
Adding @pferraro to review the clustering changes.

ee/src/main/java/org/jboss/as/ee/structure/ComponentAggregationProcessor.java
```diff
@@ -118,4 +123,29 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
             }
         }
     }
+
+    private static final class ValueService<T> implements Service<T> {
```
bstansberry, line 127: @ropalka This class seems overly complicated, as 'componentRegistry' is a final var that cannot be null. I think this can just be a no-op start/stop and return componentRegistry from getValue.

ropalka: It was refactored out from JBoss MSC code base so I didn't optimize it further @bstansberry . It is fixed now.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -337,8 +339,8 @@ protected void performBoottime(OperationContext context, ModelNode operation, Re
         this.denyAccessByDefault.set(defaultMissingMethodValue);
 
         final ModelNode defaultStatefulSessionTimeout = EJB3SubsystemRootResourceDefinition.DEFAULT_STATEFUL_BEAN_SESSION_TIMEOUT.resolveModelAttribute(context, model);
-        final ValueService<AtomicLong> defaultStatefulSessionTimeoutService = new ValueService<>(new ImmediateValue<>(
-                defaultStatefulSessionTimeout.isDefined() ? new AtomicLong(defaultStatefulSessionTimeout.asLong()) : DefaultStatefulBeanSessionTimeoutWriteHandler.INITIAL_TIMEOUT_VALUE));
+        final ValueService<AtomicLong> defaultStatefulSessionTimeoutService = new ValueService<>(() ->
+                defaultStatefulSessionTimeout.isDefined() ? new AtomicLong(defaultStatefulSessionTimeout.asLong()) : DefaultStatefulBeanSessionTimeoutWriteHandler.INITIAL_TIMEOUT_VALUE);
```
pferraro, line 343: We still want resolve to resolve the default timeout eagerly, that way we can use a Service that provides a fixed value.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ExceptionLoggingWriteHandler.java
```diff
@@ -72,9 +75,33 @@ void updateOrCreateDefaultExceptionLoggingEnabledService(final OperationContext
             value.set(enabled);
         } else {
             // create and install the service
-            final ValueService<AtomicBoolean> service = new ValueService<>(new ImmediateValue<>(new AtomicBoolean(enabled)));
-            context.getServiceTarget().addService(serviceName, service)
-                    .install();
+            final ServiceBuilder<?> sb = context.getServiceTarget().addService(serviceName);
+            sb.setInstance(new ValueService(() -> new AtomicBoolean(enabled))).install();
```
pferraro, line 79: Similar to the above, we should create the AtomicBoolean eagerly and use a Service that provides a fixed value.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/singleton/service/ValueServiceActivator.java
```diff
@@ -48,7 +50,32 @@ public class ValueServiceActivator implements ServiceActivator {
     public void activate(ServiceActivatorContext context) {
         ServiceBuilder<?> builder = context.getServiceTarget().addService(ServiceName.JBOSS.append("test", "service", "installer"));
         Supplier<SingletonPolicy> policy = builder.requires(ServiceName.parse(SingletonDefaultRequirement.SINGLETON_POLICY.getName()));
-        Consumer<ServiceTarget> installer = target -> policy.get().createSingletonServiceBuilder(SERVICE_NAME, new ValueService<>(new ImmediateValue<>(Boolean.TRUE)), new ValueService<>(new ImmediateValue<>(Boolean.FALSE))).build(context.getServiceTarget()).install();
+        Consumer<ServiceTarget> installer = target -> policy.get().createSingletonServiceBuilder(SERVICE_NAME, new ValueService<Boolean>(() -> Boolean.TRUE), new ValueService<Boolean>(() -> Boolean.FALSE)).build(context.getServiceTarget()).install();
```
pferraro, line 53: Similar comments to the above, but this is just a test class, so I wouldn't worry about it.

pferraro (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
ropalka: Fixed

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/ExceptionLoggingWriteHandler.java
ropalka: Fixed

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/singleton/service/ValueServiceActivator.java
ropalka: Also fixed

pferraro (approved):

PR #16449, opened 2023-01-10
WFLY-17493 Drop support for Undertow subsystem schemas prior to WildFly 10.1

pferraro:
https://issues.redhat.com/browse/WFLY-17493


ropalka (approved):

fl4via (commented):
I see it is already merged, but just so it does not continue marked as pending, this one LGTM :-)

PR #16454, opened 2023-01-12
WFLY-17385 Static and final method verification for bean superclasses

chengfang:
https://issues.redhat.com/browse/WFLY-17385

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/util/EjbValidationsUtil.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/util/EjbValidationsUtil.java
@@ -124,7 +124,7 @@ public final class EjbValidationsUtil {
     public static boolean verifyEjbPublicMethodAreNotFinalNorStatic(Method[] methods, String classname) {
         boolean isEjbCompliant = true;
         for (Method method : methods) {
-            if (Modifier.isPublic(method.getModifiers()) && !EjbValidationsUtil.verifyMethodIsNotFinalNorStatic(method, classname))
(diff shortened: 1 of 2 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/util/EjbValidationsUtil.java
```diff
@@ -124,7 +124,7 @@ public static void verifyEjbClassAndDefaultConstructor(final Constructor<?> ctor
     public static boolean verifyEjbPublicMethodAreNotFinalNorStatic(Method[] methods, String classname) {
         boolean isEjbCompliant = true;
         for (Method method : methods) {
-            if (Modifier.isPublic(method.getModifiers()) && !EjbValidationsUtil.verifyMethodIsNotFinalNorStatic(method, classname))
+            if (Object.class != method.getDeclaringClass() && !EjbValidationsUtil.verifyMethodIsNotFinalNorStatic(method, classname))
                 isEjbCompliant = false;
```
pferraro, line 128: Nit: this method should fail-fast by immediately returning false when a final/static method is detected,

pferraro (approved):
Feel free to ignore or address my comment separately.

ejb3/src/main/java/org/jboss/as/ejb3/util/EjbValidationsUtil.java
chengfang: I think the main purpose of this method is to log a warning for each potentially non-compliant public method. So upon the first match, it will continue to find more.

pferraro: Ok, that makes sense.

bstansberry:
Thanks @chengfang and @pferraro

PR #16466, opened 2023-01-17
[WFLY-17506] Use deployment-transformer-feature-pack jar in the wildf…

bstansberry:
…ly-preview feature pack

https://issues.redhat.com/browse/WFLY-17506

Note that this brings a dependency on a beta. I feel confident I can release a .Final on short notice.

This artifact comes from https://github.com/bstansberry/deployment-transformer-feature-pack. My intent is to change ownership of that repo to the wildfly-extras org.

The beta code in question is the same as what's been in wildfly/wildfly for a long time; I did some git manipulation to merge it and its history into the https://github.com/bstansberry/deployment-transformer-feature-pack repo; for example, see https://github.com/bstansberry/deployment-transformer-feature-pack/commits/0300d1fd0884c3b96bbe046ffc3a0a84a27f634b/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java?browsing_rename_history=true&new_path=deployment-transformer/src/main/java/org/wildfly/ee8to9/transformer/DeploymentTransformer.java&original_branch=main


jamezp (approved):

bstansberry:
Thanks for the review @jamezp

PR #16471, opened 2023-01-19
[WFLY-17412] Create a testbom

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-17412


bstansberry:
Hi @parsharma -- please include 'WFLY-17412' in the commit message and in the title of the PR (which can be edited.) Thanks!

We'll need to discuss with @kabir how to deal with the 'Cloud Tests Trigger/run-test' failure. I assume it's a case of those tests depending on our root pom for their own test dependencies, so there will need to be a migration.

testbom/pom.xml
```diff
@@ -0,0 +1,315 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bstansberry, line 1: Please move this directory into the boms directory.

bstansberry: Please call it boms/standard-test. Thanks!

parsharma: Done, thanks.

parsharma: > Hi @parsharma -- please include 'WFLY-17412' in the commit message and in the title of the PR (which can be edited.) Thanks!
> 
> We'll need to discuss with @kabir how to deal with the 'Cloud Tests Trigger/run-test' failure. I assume it's a case of those tests depending on our root pom for their own test dependencies, so there will need to be a migration.

yes, I have found out that some deps depend on root pom.xml.

bstansberry:
Thanks @parsharma

PR #16474, opened 2023-01-20
[WFLY-14387] Resource adapters subsystem does not accept expression for wm-security attribute

petrberan:
https://issues.redhat.com/browse/WFLY-14387

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
@@ -376,7 +376,6 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
                 case START_ELEMENT: {
                     switch (WorkManager.Tag.forName(reader.getLocalName())) {
                         case SECURITY: {
-                            WM_SECURITY.parseAndSetParameter("true", operation, reader);
                             switch (Namespace.forUri(reader.getNamespaceURI())) {
                                 case RESOURCEADAPTERS_1_0:
                                 case RESOURCEADAPTERS_1_1:
@@ -385,8 +384,13 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
                                 case RESOURCEADAPTERS_4_0:
                                     security = parseWorkManagerSecurity(operation, reader);
                                     break;
-                                default: // If the switch statement contains a default it does not need to be revisited each time a new schema is added unless it actually affects it.
+                                case RESOURCEADAPTERS_5_0:
+                                case RESOURCEADAPTERS_6_0:
+                                case RESOURCEADAPTERS_6_1:
                                     security = parseWorkManagerSecurity_5_0(operation, reader);
+                                    break;
+                                default: // If the switch statement contains a default it does not need to be revisited each time a new schema is added unless it actually affects it.
+                                    security = parseWorkManagerSecurity_7_0(operation, reader);
                             }
                             break;
                         }
@@ -409,6 +413,8 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
         Map<String, String> groupMappings = null;
 
         boolean userMappingEnabled = false;
+
+        WM_SECURITY.parseAndSetParameter("true", operation, reader);
 
         while (reader.hasNext()) {
             switch (reader.nextTag()) {
@@ -496,22 +502,43 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
 
     protected WorkManagerSecurity parseWorkManagerSecurity_5_0(final ModelNode operation, final XMLStreamReader reader) throws XMLStreamException,
             ParserException {
+        WM_SECURITY.parseAndSetParameter("true", operation, reader);
+        return parseWorkManagerSecurityElements(operation, reader);
+    }
+
+    protected WorkManagerSecurity parseWorkManagerSecurity_7_0(final ModelNode operation, final XMLStreamReader reader) throws XMLStreamException,
+            ParserException {
+        for (int i = 0; i < reader.getAttributeCount(); i++) {
+            Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+            switch (attribute) {
+                case ENABLED: {
+                    String value = reader.getAttributeValue(i);
+                    WM_SECURITY.parseAndSetParameter(value, operation, reader);
+                    break;
+                }
+                default:
+                    throw new ParserException(bundle.unexpectedAttribute(reader.getAttributeLocalName(i), reader.getLocalName()));
+            }
+        }
+        return parseWorkManagerSecurityElements(operation, reader);
+    }
+
+    /**
+     * Parses the common elements of Work Manager Security of the Resource Adapter subsystem version 5.0, 6.0, 6.1 and 7.0
+     */
+    private WorkManagerSecurity parseWorkManagerSecurityElements(final ModelNode operation, final XMLStreamReader reader) throws XMLStreamException,
+            ParserException {
         boolean mappingRequired = false;
         String domain = null;
         boolean elytronEnabled = false;
-        String defaultPrincipal = null;
-        List<String> defaultGroups = null;
-        Map<String, String> userMappings = null;
-        Map<String, String> groupMappings = null;
-
         boolean userMappingEnabled = false;
 
         while (reader.hasNext()) {
             switch (reader.nextTag()) {
                 case END_ELEMENT: {
                     if (WorkManager.Tag.forName(reader.getLocalName()) == WorkManager.Tag.SECURITY) {
-                        return new WorkManagerSecurityImpl(mappingRequired, domain, elytronEnabled, defaultPrincipal,
-                                defaultGroups, userMappings, groupMappings);
+                        return new WorkManagerSecurityImpl(mappingRequired, domain, elytronEnabled, null,
+                                null, null, null);
                     } else {
                         if (WorkManagerSecurity.Tag.forName(reader.getLocalName()) == WorkManagerSecurity.Tag.UNKNOWN) {
                             throw new ParserException(bundle.unexpectedEndTag(reader.getLocalName()));
@@ -561,27 +588,18 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
                             break;
                         }
                         case MAP: {
+                            String from = rawAttributeText(reader, WorkManagerSecurity.Attribute.FROM.getLocalName());
+                            if (from == null || from.trim().equals(""))
+                                throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.FROM.getLocalName(), reader.getLocalName()));
+                            String to = rawAttributeText(reader, WorkManagerSecurity.Attribute.TO.getLocalName());
+                            if (to == null || to.trim().equals(""))
+                                throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.TO.getLocalName(), reader.getLocalName()));
+                            ModelNode object = new ModelNode();
+                            WM_SECURITY_MAPPING_FROM.parseAndSetParameter(from, object, reader);
+                            WM_SECURITY_MAPPING_TO.parseAndSetParameter(to, object, reader);
                             if (userMappingEnabled) {
-                                String from = rawAttributeText(reader, WorkManagerSecurity.Attribute.FROM.getLocalName());
-                                if (from == null || from.trim().equals(""))
-                                    throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.FROM.getLocalName(), reader.getLocalName()));
-                                String to = rawAttributeText(reader, WorkManagerSecurity.Attribute.TO.getLocalName());
-                                if (to == null || to.trim().equals(""))
-                                    throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.TO.getLocalName(), reader.getLocalName()));
-                                ModelNode object = new ModelNode();
-                                WM_SECURITY_MAPPING_FROM.parseAndSetParameter(from, object, reader);
-                                WM_SECURITY_MAPPING_TO.parseAndSetParameter(to, object, reader);
                                 operation.get(WM_SECURITY_MAPPING_USERS.getName()).add(object);
                             } else {
-                                String from = rawAttributeText(reader, WorkManagerSecurity.Attribute.FROM.getLocalName());
-                                if (from == null || from.trim().equals(""))
-                                    throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.FROM.getLocalName(), reader.getLocalName()));
-                                String to = rawAttributeText(reader, WorkManagerSecurity.Attribute.TO.getLocalName());
-                                if (to == null || to.trim().equals(""))
-                                    throw new ParserException(bundle.requiredAttributeMissing(WorkManagerSecurity.Attribute.TO.getLocalName(), reader.getLocalName()));
-                                ModelNode object = new ModelNode();
-                                WM_SECURITY_MAPPING_FROM.parseAndSetParameter(from, object, reader);
-                                WM_SECURITY_MAPPING_TO.parseAndSetParameter(to, object, reader);
                                 operation.get(WM_SECURITY_MAPPING_GROUPS.getName()).add(object);
                             }
                             break;
@@ -713,7 +731,11 @@ public class ResourceAdapterParser extends CommonIronJacamarParser {
         /**
          * path
          */
-        PATH("path");
+        PATH("path"),
+        /**
+         * enabled attribute
+         */
+        ENABLED("enabled");
 
         private String name;
 
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
@@ -194,9 +194,10 @@ public final class ResourceAdapterSubsystemParser implements XMLStreamConstants,
         if (transactionSupport == TransactionSupportEnum.XATransaction) {
             isXa = true;
         }
-        if (ra.hasDefined(WM_SECURITY.getName()) && ra.get(WM_SECURITY.getName()).asBoolean()) {
+        if (ra.hasDefined(WM_SECURITY.getName()) && (ra.get(WM_SECURITY.getName()).getType().equals(ModelType.EXPRESSION) || ra.get(WM_SECURITY.getName()).asBoolean())) {
             streamWriter.writeStartElement(Activation.Tag.WORKMANAGER.getLocalName());
             streamWriter.writeStartElement(WorkManager.Tag.SECURITY.getLocalName());
+            WM_SECURITY.marshallAsAttribute(ra, streamWriter);
             WM_SECURITY_MAPPING_REQUIRED.marshallAsElement(ra, streamWriter);
             WM_SECURITY_DOMAIN.marshallAsElement(ra, streamWriter);
             WM_ELYTRON_SECURITY_DOMAIN.marshallAsElement(ra, streamWriter);
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
@@ -478,17 +478,11 @@ public final class ResourceAdapterSubsystemParser implements XMLStreamConstants,
         list.add(subsystem);
 
         try {
-            String localName;
             switch (Namespace.forUri(reader.getNamespaceURI())) {
-                case RESOURCEADAPTERS_1_0:
-                case RESOURCEADAPTERS_1_1:
-                case RESOURCEADAPTERS_2_0:
-                case RESOURCEADAPTERS_3_0:
-                case RESOURCEADAPTERS_4_0:
-                case RESOURCEADAPTERS_5_0:
-                case RESOURCEADAPTERS_6_0:
-                case RESOURCEADAPTERS_6_1:{
-                    localName = reader.getLocalName();
+                case UNKNOWN:
+                    break;
+                default: {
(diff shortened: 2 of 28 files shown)


pferraro:
Why does WFLY-14387 require a new schema version?

petrberan:
The intention is to modify the schema to hold the expression, not disallow it. And as Darran said in the previous PR:

"Ok so if this is just about adding a new attribute that needs to be in a version of the schema and management model NOT released in a prior .Final version of WildFly.

Once we have released a specific version of model / schema we don't make subsequent changes to it adding or removing attributes."

pferraro:
OK - after looking at the existing code, I think I understand now.
It looks like the wm-security attribute does not actually appear in the XML at all, but is currently set to true if the work manager contains security configuration.
And, if I understand correctly, you want to modify the existing XML reader/writer to actually parse/persist the wm-security attribute, but via an alternate name (i.e. "enabled").
So, yes, we will need a new schema version - but the management model itself has no changes.

petrberan:
It "does" appear in a way that either there is a <security> tag meaning the elytron security is on, or there is no such tag meaning that it's off. The goal is to add a new attribute "enabled" which can persist the expression, so the first commit is bumping up the schema version to 7 by simple copy&paste, next commit (I'll push on Monday) will be adding the attribute and following logic as can be seen in the previous PR and the third commit will be adding a transformer

pferraro:
Hmm - why would we need a new attribute?  Why can't we just parse/persist the model value of the existing attribute (assuming it is added to the new schema)?

petrberan:
Which 2 attributes do you have in mind? The problem now is that the expression is not persisted and there is no such attribute that would allow that. There is an element <security>, meaning the elytron security is on, and if there is no such element, the security is considered off. I intended to add the "enabled" attribute for this element in the v7 schema and change the appropriate parsing so the expression can be persisted in this attribute.

petrberan:
Forgot to tag, but feel free to have a look @darranl if the first commit is alright

pferraro:
> I intended to add the "enabled" attribute for this element in the v7 schema and change the appropriate parsing so the expression can be persisted in this attribute.

OK - I think we are in agreement.  I was originally under the impression that you intended to create a new AttributeDefinition for this "enabled" attribute, rather than using the existing attribute definition and expose it to the XML in the new schema via a different name.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
```diff
@@ -194,49 +194,81 @@ private void writeRaElement(XMLExtendedStreamWriter streamWriter, ModelNode ra,
...
+                            WM_SECURITY_MAPPING_FROM.marshallAsAttribute(node, streamWriter);
+                            WM_SECURITY_MAPPING_TO.marshallAsAttribute(node, streamWriter);
+                            streamWriter.writeEndElement();
+                        }
                         streamWriter.writeEndElement();
                     }
                     streamWriter.writeEndElement();
                 }
                 streamWriter.writeEndElement();
+                streamWriter.writeEndElement();
```
pferraro, line 270: This conditional logic is not correct.  All defined attributes should be persisted to XML, regardless of whether or not they might be ignored at runtime.
In general, a subsystem model written to and read from XML should create operations that result in an identical model, with no loss of information.

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
```diff
@@ -194,49 +194,81 @@ private void writeRaElement(XMLExtendedStreamWriter streamWriter, ModelNode ra,
...
+            } else {
+                securityEnabled = WM_SECURITY.getDefaultValue().asBoolean();
+            }
+
+            // Preserves the expression in the XML if an expression was used to turn the WorkManager Security off
+            if (!securityEnabled && isExpression) {
+                streamWriter.writeStartElement(Activation.Tag.WORKMANAGER.getLocalName());
+                streamWriter.writeStartElement(WorkManager.Tag.SECURITY.getLocalName());
+                WM_SECURITY.marshallAsAttribute(ra, streamWriter);
+                streamWriter.writeEndElement();
```
pferraro, line 219: This is not correct.  The security element should be written if any of its attributes are defined, regardless of its value.

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
petrberan: The intention was to add a way to persist the expression without changing the current behavior. At this moment if there is `security` element in the XML the `wm-security` is on, if there is no such element the `wm-security` is off. The `enabled` attribute is optional in case of the expression to prevent any regression. I could set the `enabled` attribute to be required instead of optional or simply use the optional attribute to hold `true` and `false` as well

In both cases, the change should be only in this class

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
petrberan: The security element is written if the `wm-security` is either set to true or is an expression (regarding of its resolved value). Only in case the `wm-security` is false the element is not written at all. The enabled attribute is optional and is used only in case an expression was used to change the settings. The problem I see is if an expression was used the rest of the `wm-security` logic is not persisted and should be persisted in case the expression was used, regardless the actual value.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
pferraro: > The intention was to add a way to persist the expression without changing the current behavior.

A new version of an XSD is not required to retain the same syntax/semantics as an older version. The only requirement is that it must generate operations that create equivalent runtime behavior.

Since this subsystem uses a single parser class for all versions of the schema (which is a good thing!), you'll want to toggle the parsing behavior based on the version of the namespace of the schema used by the XML being parsed.

> At this moment if there is security element in the XML the wm-security is on, if there is no such element the wm-security is off.

This logic should stay in place, but only if the version of the namespace being parsed is less than the version you just created.

> The enabled attribute is optional in case of the expression to prevent any regression.

The enabled attribute is not required to be optional.  Users migrating to WF28 from WF27 will be using an older version of the schema, which will be parsed using the same logic as it was previously, and will be persisted using the new version of the schema via this subsystem's XMLElementWriter.  If the attribute is required, then the writer simply always writes its value to the XML (rather than omitting if it matches the default).

> I could set the enabled attribute to be required instead of optional or simply use the optional attribute to hold true and false as well

If this is the only attribute of the containing XML element, then it makes sense to make it required in the XSD.  However, it does seem odd that this attribute does not reside within an element that nests the configuration that it is meant to enable.  WDYT?

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
```diff
@@ -376,17 +377,63 @@ protected WorkManager parseWorkManager(final ModelNode operation, final XMLStrea
...
-                                    break;
-                                default: // If the switch statement contains a default it does not need to be revisited each time a new schema is added unless it actually affects it.
-                                    security = parseWorkManagerSecurity_5_0(operation, reader);
+                            boolean securityEnabled = false;
+                            int attributeSize = reader.getAttributeCount();
+                            if (attributeSize > 0) {
+                                for (int i = 0; i < attributeSize; i++) {
+                                    Attribute attribute = Attribute.forName(reader.getAttributeLocalName(i));
+                                    switch (attribute) {
+                                        case ENABLED: {
```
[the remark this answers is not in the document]
petrberan: Thanks for noticing, will be fixed in next push

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
```diff
@@ -376,17 +377,63 @@ protected WorkManager parseWorkManager(final ModelNode operation, final XMLStrea
...
+                                            // Checks the attribute and either sets the value directly if the value is
+                                            // bool or resolves it as an expression.
+                                            String value = reader.getAttributeValue(i);
+                                            WM_SECURITY.parseAndSetParameter(value, operation, reader);
+                                            if (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("false")) {
+                                                securityEnabled = Boolean.valueOf(value);
+                                                break;
+                                            }
+                                            if (operation.get(WM_SECURITY.getName()).getType().equals(ModelType.EXPRESSION)) {
+                                                securityEnabled = operation.get(WM_SECURITY.getName()).resolve().asBoolean();
```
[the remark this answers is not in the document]
petrberan: Good point. I was trying to "simulate" the pre-v7 XML behavior in the v7 schema, but if the behavior can differ, then this `securityEnabled` is not necessary anymore.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
petrberan: I believe the problem was me trying to maintain the same behavior in the new schema and not really understanding the relation of `ResourceAdapterSubsystemParser` and `ResourceAdapterParser`. There could be a lot of simplifying if the `ResourceAdapterParser` will always persist the v7 schema and the entire conditional logic with the attribute in `ResourceAdapterSubsystemParser` can be moved into a separate method for the v7 parser. Thanks for the clarification, now I understand it better.

> However, it does seem odd that this attribute does not reside within an element that nests the configuration that it is meant to enable. WDYT?

I was thinking about making it a sub-element as well, but looking at the schema, all `enabled` properties are used as attributes, so I wanted to keep the consistency. Moreover, if the `wm-security` is set off, the entire configuration is erased. I'm afraid if there would be any kind of regression if in case of an expression there would always be one nested element. What's your view on this?

pferraro: > There could be a lot of simplifying if the ResourceAdapterParser will always persist the v7 schema

Correct.  Subsystems only register a single XML writer, which should always persist XML that adheres to the current version of the subsystem schema.

> and the entire conditional logic with the attribute in ResourceAdapterSubsystemParser can be moved into a separate method for the v7 parser.

Yes.

> I was thinking about making it a sub-element as well, but looking at the schema, all enabled properties are used as attributes, so I wanted to keep the consistency.

Actually, what you have in the XSD makes sense.
e.g.
```xml
<security enabled="...">
   <mapping-required>...</mapping-required>
   <!-- ... -->
</security>
```

What confused me is that, looking at the logic in your writer, you were ending the element after writing the WM_SECURITY attribute, but now I see that you intended only to do this when security is disabled, which we already covered on another thread.

> Moreover, if the wm-security is set off, the entire configuration is erased.

Ideally, the writer logic should be driven by what the model defines, independent of its values.  In that way, nothing should ever be "erased".  The writer should preserve all "information" in the model, such that, when re-read by the parser, the subsystem will generate an equivalent model.
Meaning, if the model for a work manager defines additional security resources/attributes, those should be persisted, even if the value of the WM_SECURITY attributes would ignore it at runtime.
This should also simply your XMLElementWriter logic immensely, I think.

> I'm afraid if there would be any kind of regression if in case of an expression there would always be one nested element.

What kind of regression do you anticipate?

petrberan: Changed the behavior as agreed, there is a new parser for the v7 schema and the `enabled` attribute is now required. 

> What kind of regression do you anticipate?

I was afraid of regression when it comes to adding another extra element, but in the end it would be solved in the parser, so bad guess on my part and it should be fine

> Ideally, the writer logic should be driven by what the model defines, independent of its values. In that way, nothing should ever be "erased". The writer should preserve all "information" in the model, such that, when re-read by the parser, the subsystem will generate an equivalent model.
Meaning, if the model for a work manager defines additional security resources/attributes, those should be persisted, even if the value of the WM_SECURITY attributes would ignore it at runtime.
This should also simply your XMLElementWriter logic immensely, I think.

I have a question regarding persisting model attributes. For example, [wm-security-mapping-required](https://docs.wildfly.org/27/wildscribe/deployment/subsystem/resource-adapters/ironjacamar/ironjacamar/resource-adapter/index.html#attr-wm-security-mapping-required) is stored in configuration and it is, as long as it's set and the `wm-security` is true (or an expression is used in this PR), but when the `wm-security` is set to false, the `wm-security-mapping-required` is set back to default value and it's no longer in the config. Is this behavior intended or should any sub-element with non-default value be persisted regardless of the `wm-security` presence in the config?

petrberan: To be precise, let's assume this situation
```
<tag>
    <wm-security enabled="true">
        <mapping-required>true</mapping-required>
    </wm-security>
</tag>
```
and I set the `wm-security` to false, should I get 
```
<tag>
    <wm-security enabled="false">
        <mapping-required>true</mapping-required>
    </wm-security>
</tag>
```
in order to persist the `mapping-required` value or can it be removed like this
```
<tag/>
```
?

petrberan:
@pferraro Added transformers for the resource adapters. I believe there are all necessary changes in this PR. Any insight will be welcome

petrberan:
Hi @pferraro can I ask for a review please?

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
```diff
@@ -194,9 +194,10 @@ private void writeRaElement(XMLExtendedStreamWriter streamWriter, ModelNode ra,
         if (transactionSupport == TransactionSupportEnum.XATransaction) {
             isXa = true;
         }
-        if (ra.hasDefined(WM_SECURITY.getName()) && ra.get(WM_SECURITY.getName()).asBoolean()) {
+        if (ra.hasDefined(WM_SECURITY.getName()) && (ra.get(WM_SECURITY.getName()).getType().equals(ModelType.EXPRESSION) || ra.get(WM_SECURITY.getName()).asBoolean())) {
```
tadamski, line 197: Maybe let's indeed write it in all circumstances to avoid losing any data that for whatever reason was placed in the config.

tadamski (changes requested):
Apart of the comment above the PR looks good to me, if the tests are passing I guess we should be fine.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
petrberan: You mean as having only `(ra.hasDefined(WM_SECURITY.getName())`? Although I'm not against it, that would change the behavior of the subsystem as the configuration is deleted once the `wm-security` attribute is false. I would prefer this issue to deal with the expressions and another issue dealing with the behavior, since this there is downstream dependency for this issue

petrberan: Do we want to change the behavior in this PR or should a next issue for the change be created? @tadamski

bstansberry:
@tadamski Can you respond to @petrberan above?

@pferraro Does this look ok now?

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersExtension.java
```diff
@@ -94,6 +98,7 @@ public void initializeParsers(final ExtensionParsingContext context) {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.RESOURCEADAPTERS_5_0.getUriString(), () -> ResourceAdapterSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.RESOURCEADAPTERS_6_0.getUriString(), () -> ResourceAdapterSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.RESOURCEADAPTERS_6_1.getUriString(), () -> ResourceAdapterSubsystemParser.INSTANCE);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.RESOURCEADAPTERS_7_0.getUriString(), () -> ResourceAdapterSubsystemParser.INSTANCE);
```
pferraro, line 101: Don't use the supplier variant of this method with static parser instances.  This results in unnecessary metaspace bloat.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
```diff
@@ -596,6 +600,121 @@ protected WorkManagerSecurity parseWorkManagerSecurity_5_0(final ModelNode opera
...
+                                operation.get(WM_SECURITY_MAPPING_GROUPS.getName()).add(object);
+                            }
+                            break;
+                        }
+                        default:
+                            throw new ParserException(bundle.unexpectedElement(reader.getLocalName()));
+                    }
+                    break;
+                }
+            }
```
pferraro, line 713: Why duplicate all of this logic?

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
tadamski: @petrberan ok, I think creating another issue is fine

tadamski (approved):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
petrberan: I believe I did this for the same reason why `parseWorkManagerSecurity_5_0` duplicates the `parseWorkManagerSecurity` (and adds the `ELYTRON_SECURITY_DOMAIN`), to be more flexible when it comes to future changes in the 7_0 parse method. Would you prefer if I created separate method for the duplicate parsing code and used it in the 5_0 and 7_0 parse methods??

darranl: The reason early on we started using the duplication of parse methods for newer methods was to make it easier to just delete old versions as they became unsupported rather than having to manually unravel version specific changes within the method.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
```diff
@@ -376,16 +376,20 @@ protected WorkManager parseWorkManager(final ModelNode operation, final XMLStrea
...
                                 case RESOURCEADAPTERS_3_0:
                                 case RESOURCEADAPTERS_4_0:
+                                    WM_SECURITY.parseAndSetParameter("true", operation, reader);
                                     security = parseWorkManagerSecurity(operation, reader);
                                     break;
+                                case RESOURCEADAPTERS_7_0:
+                                    security = parseWorkManagerSecurity_7_0(operation, reader);
+                                    break;
                                 default: // If the switch statement contains a default it does not need to be revisited each time a new schema is added unless it actually affects it.
+                                    WM_SECURITY.parseAndSetParameter("true", operation, reader);
```
darranl, line 392: This switch statement is back to front, the default needs to be the latest parser.

As this stands when version 8 of this schema is added it will revert to calling version 5 of this parse method.

Instead you need a case for 5 and 6 that calls version 5 of the parse method, then default then calls version 7 so it is used for 7 and later until the next time a change is made.

Also adding these parseAndSetParameter calls makes the switch statement more complication, I think those will be better at the top of the parse methods and then obviously not at the top of the method for 7.

darranl (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterParser.java
pferraro: In general, inheritance is a poor way to model versioning...
But given the current state, it would make sense to encapsulate the parsing of the attributes of the security element in a separate method from the parsing of the child elements.  This way your subclass only needs to override the attribute parsing, and eliminate the duplication of the child element parsing.

petrberan:
@darranl @pferraro Can I ask for a review please?

petrberan:
@darranl @pferraro Can I ask for a review please?

darranl:
@pferraro do you have anything else specifically to add?  I added one small comment but TBH not really created by this PR as following the existing pattern.

pferraro (approved):

darranl (approved):

darranl:
@petrberan Almost ready to go with this one, just a remaining merge conflict if you can please take a look / rebase.  Thanks.

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdapterSubsystemParser.java
```diff
@@ -487,7 +488,8 @@ public void readElement(final XMLExtendedStreamReader reader, final List<ModelNo
                 case RESOURCEADAPTERS_4_0:
                 case RESOURCEADAPTERS_5_0:
                 case RESOURCEADAPTERS_6_0:
-                case RESOURCEADAPTERS_6_1:{
+                case RESOURCEADAPTERS_6_1:
+                case RESOURCEADAPTERS_7_0:{
```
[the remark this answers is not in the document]
petrberan: Changed the switch while rebasing

petrberan:
@darranl Should be all fine now, rebased the PR

PR #16477, opened 2023-01-23
WFLY-17552 MDC from 'default task' Thread copied to 'Batch Thread' on…

chengfang:
… creation but not cleared

https://issues.redhat.com/browse/WFLY-17552

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/DiagnosticContextHandle.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/DiagnosticContextHandle.java
@@ -1,0 +1,62 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.wildfly.extension.batch.jberet.deployment;
+
+import java.util.Map;
+
+import org.jboss.logging.MDC;
+import org.jboss.logging.NDC;
+
+/**
+ * A {@code ContextHandle} to propagate and clear mapped diagnostic context (MDC)
+ * and nested diagnostic context (NDC) data used in logging.
+ */
+class DiagnosticContextHandle implements ContextHandle {
+    private final Map<String, Object> mdcMap;
+    private final String ndcTop;
+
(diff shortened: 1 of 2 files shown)


jamezp (commented):
I'm not sure where we put it, but we may want this to be a separate `ContextHandle. It would affect more than just batch the more I think about it. I think we might want to actually do it from the control point.

jamezp:
@chengfang Do you know where this is copied? I don't see it being copied. I do think this is a bug, but we need to find a better place for it. We need to clear NDC too.

chengfang:
Yeah, clearing it inside the finally block is not enough even for batch subsystem.  Not sure if handling it centrally will work for all subsystems.  ejb subsystem already has interceptors for managing MDC NDC in ejb threads. This PR is related to a customer issue so I'll address it as a batch issue for now.

chengfang:
MDC data are in ThreadLocal, and are copied from the parent thread (i.e., undertow thread) to child thread when new threads are created in batch thread pool.

jamezp:
It really feels like there might be multiple places we need to handle it TBH. The request-controller in core, however that is not guaranteed to be set. I need to look at how all the thread pools are created and see what we can do from there. That is the ideal place is within the thread pool if that is possible.

chengfang:
I think the best time for such propagation is at invocation handoff time,  when you have access to both the incoming invocation data and newly created task.

Thread creation time is too early since the parent thread may not the the same as the invocation thread (a thread pool may be pre-filled). Request-controller could be too late and it does not have access to the original invocation thread.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/MappedDiagnosticContextHandle.java
```diff
@@ -0,0 +1,54 @@
...
+        for (Map.Entry<String, Object> e : mdcMap.entrySet()) {
+            MDC.put(e.getKey(), e.getValue());
+        }
+
+        return new Handle() {
+            @Override
+            public void tearDown() {
+                for (String k : mdcMap.keySet()) {
+                    MDC.remove(k);
+                }
```
jamezp, line 50: This will only clear old entries that were already present prior to the thread being created. It won't clear any new entries added.

jamezp: Yes. We should clear it on the thread in general. We actually need to the same for `NDC` as well.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/DiagnosticContextHandle.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public Handle setup() {
+        for (Map.Entry<String, Object> e : mdcMap.entrySet()) {
+            MDC.put(e.getKey(), e.getValue());
+        }
+
+        if (ndcTop != null) {
+            NDC.clear();
```
jamezp, line 50: I'm not sure we should clear this here. I would think we want to inherit it.

jamezp (commented):
Other than the minor comment, this looks good. The one place this won't work is if the user uses their own slf4j or log4j2. However, I don't think there is anything we can do there.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/DiagnosticContextHandle.java
chengfang: the top NDC item (`ndcTop`) is copied from the invoking thread's NDC. If the batch thread is a newly created one, it may inherit NDC from its parent, and so pushing `ndcTop` to NDC without clearing it would duplicate it.

If the batch thread is a reused one from the batch thread pool, it contains no NDC data since it has been cleared in its previous run. 

So with `NDC.clear()` here, it makes both cases consistent.

jamezp: Ah, right. I should know that :)

jamezp (approved):

bstansberry:
Thanks @chengfang  and @jamezp

PR #16481, opened 2023-01-24
[WFLY-17547] EJB: remove dependencies on ejb-client-legacy

tadamski:
https://issues.redhat.com/browse/WFLY-17547


chengfang:
some check failed with:
```
 Child module /home/runner/work/wildfly/wildfly/testsuite/integration/legacy-ejb-client of /home/runner/work/wildfly/wildfly/testsuite/integration/pom.xml does not exist @ 
```

Looks like https://github.com/wildfly/wildfly/blob/main/testsuite/integration/pom.xml#L76 this line need to be removed too.

tadamski:
> Looks like https://github.com/wildfly/wildfly/blob/main/testsuite/integration/pom.xml#L76 this line need to be removed too.

True, tanks. Fixed.

chengfang:
looks like ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml still has a section for jboss-ejb-client-legacy, which should be removed.

pom.xml
```diff
@@ -526,7 +526,6 @@
         <version.org.jboss.common.jboss-common-beans>2.0.1.Final</version.org.jboss.common.jboss-common-beans>
         <legacy.version.org.jboss.ejb-client>4.0.50.Final</legacy.version.org.jboss.ejb-client>
         <version.org.jboss.ejb-client>5.0.2.Final</version.org.jboss.ejb-client>
-        <version.org.jboss.ejb-client-legacy>3.0.3.Final</version.org.jboss.ejb-client-legacy>
```
bstansberry, line 529: @tadamski Please also remove https://github.com/wildfly/wildfly/blob/main/pom.xml#L503 -- 'legacy.version.org.jboss.narayana'. The only use of that property was in testsuite/integration/legacy-ejb-client.pom.xml so this PR makes it cruft.

tadamski: fixed

chengfang (approved):

PR #16487, opened 2023-01-26
WFLY-17567 Transaction logic in WebJPATestCase servlet masks root cause of test failures

pferraro:
This test fails frequently on CI, its problematic transaction handling on failure is masking the root cause.

https://issues.redhat.com/browse/WFLY-17567


scottmarlow (approved):
Thanks Paul!

PR #16489, opened 2023-01-27
[WFLY-17559] Disable PicketBox configuration possibility

tadamski:
https://issues.redhat.com/browse/WFLY-17559

@bstansberry @fjuma @darranl Although the related issue was closed I was wondering whether changes proposed with this PR may be good to implement. 

The thing is that IronJacamar still supports PicketBox configuration - standalone configurations of the projects rely on it. I guess to remove it we would have to replace it with Elytron in IronJacamar code, but I don't think it is feasible. 

As a result, some the PicketBox configuration can be "propagated" to WildFly from IJ. In this PR I'm trying to disable
possibility of such configuration and remove all PicketBox references from the docs.


connector/src/main/resources/org/jboss/as/connector/subsystems/datasources/LocalDescriptions.properties
```diff
@@ -176,7 +176,7 @@ datasources.data-source.enlistment-trace=Defines if WildFly/IronJacamar should r
 
 datasources.data-source.recovery-username=The user name used for recovery
 datasources.data-source.recovery-password=The password used for recovery
-datasources.data-source.recovery-security-domain=The security domain used for recovery, for PicketBox security domains
+datasources.data-source.recovery-security-domain=Deprecated. The security domain used for recovery, for PicketBox security domains
```
bstansberry, line 179: To deprecate an attribute:

1) Add a setDeprecated call to the AttributeDefinition builder
2) Add a 'datasources.data-source.recovery-security-domain.deprecated=xxxx' string to provide details. For example:  'datasources.data-source.recovery-security-domain.deprecated=PicketBox security domains are not supported in servers running the current version of the datasources subsystem. Use 'authentication-context' to configure Elytron security.'

bstansberry (changes requested):
Thanks @tadamski.  I closed WFLY-17559 because it was kind of vague and misleading, and opened https://issues.redhat.com/browse/WFLY-17566 instead. This is is the 'converted to purely admin-only, so they fail in a non-admin-only server' in WFLY-17566, for these attributes.

Using WFLY-17559 for this makes sense though and I re-opened it. WFLY-17566 goes beyond this; it's also about things like the old 'enable' and 'disable' operations and various other things that were deprecated in EAP 7.

connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
```diff
@@ -977,4 +978,7 @@ public interface ConnectorLogger extends BasicLogger {
 
     @Message(id = 130, value = "Report directory %s does not exist")
     OperationFailedException reportDirectoryDoesNotExist(String reportDirectory);
+
+    @Message(id = 131, value = "Cannot deploy the datasource. Only Elytron security is supported.")
```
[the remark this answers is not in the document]
tadamski: fixed

connector/src/main/resources/org/jboss/as/connector/subsystems/datasources/LocalDescriptions.properties
tadamski: thanks. fixed.

tadamski:
After the discussion in WFLY-17042 (elytron-enabled fields have to default to false because we can't change the model now) I decided to refactor this PR: elytron-enabled fields are deprecated and have to effect. Elytron is always used even if elytron-enabled is set to false, security-domains are marked as deprecated and are also ignored.

connector/src/main/resources/org/jboss/as/connector/subsystems/datasources/LocalDescriptions.properties
```diff
@@ -177,7 +179,9 @@ datasources.data-source.enlistment-trace=Defines if WildFly/IronJacamar should r
 datasources.data-source.recovery-username=The user name used for recovery
 datasources.data-source.recovery-password=The password used for recovery
 datasources.data-source.recovery-security-domain=The security domain used for recovery, for PicketBox security domains
+datasources.data-source.recovery-security-domain.deprecated=PicketBox security domains are not supported in servers running the current version of the datasources subsystem. Use 'authentication-context' to configure Elytron security
 datasources.data-source.recovery-elytron-enabled=Enables Elytron security for handling authentication of connections for recovery. The Elytron authentication-context to be used will be current context if no context is specified (see authentication-context).
+datasources.data-source.recovery-elytron-enabled.deprecated=Elytron is enabled by default. Setting this flag to false will cause validation to fail.
```
bstansberry, line 184: @tadamski Why is the semantic for this different? Shouldn't it just be that in a current version server recovery-elytron-enabled is ignored and setting recovery-security-domain fails?

XaDataSourceAdd.performRuntime looks like it's still trying to reason based on the value of recovery-elytron-enabled.

tadamski: @bstansberry hmm, actually this is a docs error; my intention was it to work just as you described

tadamski: fixed

bstansberry (approved):

darranl:
Added ready-for-merge based on Brian's approval and clean CI

bstansberry:
Thanks @tadamski!

PR #16491, opened 2023-01-27
[WFLY-17356] [testsuite] Remove master from NonDefaultRemoteWorkerTestCase and other minor classes in the testsuite

petrberan:
https://issues.redhat.com/browse/WFLY-17356


petrberan:
@yersan The ocsp-responder.xml cannot be changed as the OCSPServer is checking the `master` attribute, but I managed to find other uses of the word master not shown in this issue's description since grep is case sensitive

PR #16493, opened 2023-01-27
[WFLY-17572] Wait for the Keycloak started log message to ensure Keyc…

jamezp:
…loak is fully started.

https://issues.redhat.com/browse/WFLY-17572

### DIFF
--- a/testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/KeycloakContainer.java
+++ b/testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/KeycloakContainer.java
@@ -20,6 +20,8 @@ package org.wildfly.test.integration.elytron.oidc.client;
 
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.containers.wait.strategy.WaitAllStrategy;
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
 
 import java.time.Duration;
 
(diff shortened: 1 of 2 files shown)


fjuma (approved):
Thanks @jamezp!

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/KeycloakContainer.java
```diff
@@ -67,6 +67,7 @@ protected void configure() {
             waitingFor(Wait.forHttp("/").forPort(PORT_HTTP));
             withCommand("start-dev");
         }
+        waitingFor(Wait.forLogMessage(".*Keycloak.*started.*", 1));
```
fjuma, line 70: Just realized, I think this line will need to be moved into the `else` block above since I think the message for Keycloak is different from the message for RH SSO.

jamezp: Ah, that would make sense. I'll make that change and push up the commit again with some added logging since it still seems to be failing on CI, but I can't get it to fail locally.

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/KeycloakContainer.java
```diff
@@ -64,7 +66,10 @@ protected void configure() {
         if (isUsedRHSSOImage()) {
             waitingFor(Wait.forHttp("/auth").forPort(PORT_HTTP));
         }else{
-            waitingFor(Wait.forHttp("/").forPort(PORT_HTTP));
+            final WaitStrategy strategy = new WaitAllStrategy()
+                    .withStrategy(Wait.forHttp("/").forPort(PORT_HTTP))
+                    .withStrategy(Wait.forLogMessage(".*Keycloak.*started.*", 1));
+            waitingFor(strategy);
```
fjuma, line 72: Nice, hopefully this does the trick!

fjuma (approved):
Thanks @jamezp!

bstansberry:
Thanks @jamezp and @fjuma

PR #16501, opened 2023-01-31
[WFLY-17042] JCA: set elytron-enabled to true by default

tadamski:
https://issues.redhat.com/browse/WFLY-17042

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
@@ -66,8 +66,6 @@ import org.jboss.jca.common.api.metadata.ds.XaDataSource;
  */
 public class Constants {
 
-    private static final Boolean ELYTRON_MANAGED_SECURITY = Boolean.FALSE;
-
     public static final String DATASOURCES = "datasources";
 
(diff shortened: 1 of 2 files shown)


darranl:
Where do these constants get used?  Just in Runtime checks or could they influence the model state?

tadamski:
Actually the are used only in attribute definition file and are equivalent to setDefault(ModelNode.TRUE).
Maybe I will change it this way for clarity.

darranl:
If this changes the model I don't think it will be feasible as we need to support older configuration in domain mode.

A model change like this would likely need a version bump but also we would need to consider the meaning when parsing older XML, additionally there would be a transformer requirement but as this attribute has a default value I am not sure we could reliable write transformers.

tadamski:
I guess the only solution then would be to deprecated elytron-enabled attribute and document that is this being ignored. Is it acceptable?

darranl:
I think we do have other code where at runtime we know Elytron is the only solution around, in this case now there is no choice available that runtime just assumes if they need security it is Elytron security.

That would at least keep is compatible for domain mode as well where older hosts can make their own runtime check.

tadamski:
OK. I'm going to close this PR then and update the docs in WFLY-17559 to make sure that is it clear that elytron-enabled setting has no effect.

PR #16510, opened 2023-02-03
[WFLY-17599] validate NamingBindingResourceDefinition.TYPE from allow…

soul2zimate:
…ed values

Issue: https://issues.redhat.com/browse/WFLY-17599

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
+++ b/naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
@@ -35,6 +35,7 @@ import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.access.management.AccessConstraintDefinition;
 import org.jboss.as.controller.operations.validation.EnumValidator;
+import org.jboss.as.controller.operations.validation.StringAllowedValuesValidator;
 import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.Resource;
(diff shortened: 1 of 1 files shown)


naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
```diff
@@ -56,6 +57,8 @@
  */
 public class NamingBindingResourceDefinition extends SimpleResourceDefinition {
 
+    private static final String[] ALLOWED_TYPES = {"char", "java.lang.Character", "byte", "java.lang.Byte", "short", "java.lang.Short", "int", "java.lang.Integer", "long", "java.lang.Long", "float", "java.lang.Float", "double", "java.lang.Double", "boolean", "java.lang.Boolean", "java.lang.String", "java.net.URL"};
```
rhusar, line 60: Any reason why not to use  `java.lang.Class#getName` e.g. `Character.class.getName()`, etc?

soul2zimate: This list is copied from `NamingBindingAdd.coerceToType` method, but yes, we can use that approach as well.

darranl (changes requested):
If this is restricting the list of allowed values based on what was previously allowed I think this will at the very least need a model bump.

As this is more restrictive than previously we likely don't need transformers for the change.

soul2zimate:
Does it ? This only moves the check earlier to `Stage.Model` rather than `Stage.Runtime`.  An invalid value previously fails at `NamingBindingAdd.coerceToType`

bstansberry (approved):
I agree @soul2zimate. The Stage.RUNTIME validation logic used for this attribute is unchanged since 2012, so I can't see how doing the validation in Stage.MODEL can break any existing working use case. Technically, bumping the API version is the right thing to do, but I think it's overkill. The primary thing this does is get the read-resource-description content to reflect the long-standing actual behavior.

bstansberry:
Thanks @soul2zimate

PR #16511, opened 2023-02-03
[WFLY-17600] add missing validators for some allowed values

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-17600


bstansberry (changes requested):
If EnumValidator, StringAllowedValuesValidator, or any other implementation of AllowedValuesValidator is passed to setValidator, then setAllowedValues should not be called. The AllowedValuesValidator.getAllowedValues method provides the needed data, and calling ADBuilder.setAllowedValues adds no benefit and can only lead to a discrepancy between what read-resource-description reports and what the validator actually allows.

IIRC there was some effort put into rationalizing this a bit and making it less of a secret, but I don't recall happened with that.

bstansberry (approved):

bstansberry:
Thanks @soul2zimate and sorry for the delay here.

I'm going to close this in favor of #16666, which is just this with the merge conflict resolved.

PR #16513, opened 2023-02-03
Ignore TwoConnectorsEJBFailoverTestCase

bstansberry:
Ignore the test case pending resolution of WFLY-17605.

@pferraro @rachmatowicz @rhusar @chengfang @tadamski FYI.


rhusar:
I also noticed this hanging in last few weeks in our cluster lab runs. Since the test case has not seen changes since 7/2021 I am suspecting this is not a test issue but easily a legitimate issue. I see first failure in our lab on Jan 21, 2023 so that should help @tadamski @rachmatowicz @chengfang narrow the cause down.

PR #15069, opened 2022-01-08
[WFLY-14980] add default 60s r/o timeout to listeners

ropalka:
https://issues.redhat.com/browse/WFLY-14980

Follows on: https://github.com/wildfly/wildfly/pull/15041
Supersedes: https://github.com/wildfly/wildfly/pull/14595

PR #16514, opened 2023-02-04
Fixup WFLY-14980 which has broken 'main' build.

rhusar:
The https://github.com/wildfly/wildfly/pull/15069 was out of date and broke main!


rhusar:
@bstansberry this fixes the broken build. cc @pferraro

bstansberry:
Thanks @rhusar

PR #16517, opened 2023-02-06
WFLY-17610 Deprecate concurrency-level attribute of infinispan locking resource

pferraro:
https://issues.redhat.com/browse/WFLY-17610


rhusar (approved):
👍🏼 Looks good.

rhusar (changes requested):
Actually, please update the XSD as well, we typically prefix the text in <xs:documentation> with "Deprecated."

rhusar (approved):
Looks good now, thanks. 👍🏼

PR #16518, opened 2023-02-06
[WFLY-17606] JCA: ModifiableResourceAdapter#hashCode,toString,equals …

tadamski:
…method synchronized

https://issues.redhat.com/browse/WFLY-17606

Make ModifiableResourceAdapter (ActivationImpl) resilient to ConcurrentModificationException as suggested in https://issues.redhat.com/browse/JBJCA-1384

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ModifiableResourceAdapter.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ModifiableResourceAdapter.java
@@ -71,5 +71,20 @@ public class ModifiableResourceAdapter extends ActivationImpl {
         this.raXmlDeploymentServiceName = raXmlDeploymentServiceName;
     }
 
+    @Override
+    public synchronized int hashCode() {
+        return super.hashCode();
+    }
+
+    @Override
+    public synchronized boolean equals(Object obj) {
+        return super.equals(obj);
+    }
+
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ModifiableResourceAdapter.java
```diff
@@ -71,5 +71,20 @@ public void setRaXmlDeploymentServiceName(ServiceName raXmlDeploymentServiceName
...
+    @Override
+    public synchronized boolean equals(Object obj) {
+        return super.equals(obj);
+    }
+
+    @Override
+    public synchronized String toString() {
+        return super.toString();
+    }
+
```
pferraro, line 88: I think this is fine as a workaround, but I think the upstream thread-safety issue should still be addressed.  WDYT?

tadamski: @pferraro you mean fixing it in ironjacamar? tbh my line of thinking was that ActivationImpl (the parent class) works on collections that are provided to it so theoretically some other implementations could decide to construct it with synchronized collections. But giving a second look, I see that in ActivationImpl creates ArrayLists itself anyway. As a result I agree, I'm going to synchronize the method in IJ code.

PR #16522, opened 2023-02-09
[WFLY-17357] [git branches] Fix problematic language usage in branches

petrberan:
https://issues.redhat.com/browse/WFLY-17357

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/Class_Loading_in_WildFly.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/Class_Loading_in_WildFly.adoc
@@ -353,7 +353,7 @@ ear deployment is as follows:
 [IMPORTANT]
 
 The xsd for jboss-deployment-structure.xml is available at
-https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd[https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd]
+https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_3.xsd[jboss-deployment-structure-1_3.xsd]
 
 [[accessing-jdk-classes]]
 == Accessing JDK classes
--- a/docs/src/main/asciidoc/_testsuite/WildFly_Testsuite_Harness_Developer_Guide.adoc
+++ b/docs/src/main/asciidoc/_testsuite/WildFly_Testsuite_Harness_Developer_Guide.adoc
@@ -14,8 +14,7 @@ merged here later.
 [[adding-a-new-maven-plugin]]
 == Adding a new maven plugin
 
-The plugin version needs to be specified in jboss-parent. See
-https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml#L65 .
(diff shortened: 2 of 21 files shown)


petrberan:
There are still some "master" branches left, those still use the word "master" instead of "main" or any other variant. Few links were removed due to the fact that said link doesn't work anymore and has no alternative (helloworld-rs for example)

docs/src/main/asciidoc/_developer-guide/Class_Loading_in_WildFly.adoc
```diff
@@ -353,7 +353,7 @@ ear deployment is as follows:
 [IMPORTANT]
 
 The xsd for jboss-deployment-structure.xml is available at
-https://github.com/wildfly/wildfly-core/blob/master/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd[https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd]
+https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd[https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd]
```
yersan, line 356: You can just use the latest one which is: [jboss-deployment-structure-1_3.xsd](https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_3.xsd)

In addition, Can you replace the `[https://github.com/wildfly/wildfly-core/blob/main/server/src/main/resources/schema/jboss-deployment-structure-1_2.xsd]` by `[jboss-deployment-structure-1_2.xsd]` ?

docs/src/main/asciidoc/_testsuite/WildFly_Testsuite_Harness_Developer_Guide.adoc
```diff
@@ -15,7 +15,7 @@ merged here later.
 == Adding a new maven plugin
 
 The plugin version needs to be specified in jboss-parent. See
-https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml#L65 .
+https://github.com/jboss/jboss-parent-pom/blob/main/pom.xml#L65 .
```
yersan, line 18: Again, remove the line number and describe the paragraph a bit differently so the user can understand what the links provide, for example, The plugin version needs to be specified in jboss-parent at the `<properties>` section of the jboss-parent pom.xml file

yersan (changes requested):
thanks, @petrberan. I've added some very minor comments that would be great if are done together with this change.

I've also checked you left this one:

src/main/asciidoc/_developer-guide/ejb3/Jakarta_Enterprise_Beans_over_HTTP.adoc:https://github.com/wildfly/wildfly-http-client/blob/master/docs/src/main/asciidoc/wire-spec-v1.asciidoc

It can use `main` branch instead of `master`

petrberan:
Thanks for looking into this @yersan , I've made the requested changes and managed to find one more problem to fix.

Can you please take a look at changes in Micrometer.adoc as I'm not sure if the code is fine now.

bstansberry (approved):

bstansberry:
Thanks @petrberan

PR #16527, opened 2023-02-14
Fix for WFLY-17630, Upgrade galleon-plugins to 6.3.0.Final

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-17630
Upgrade galleon-plugins and remove the copy of wildfly-config-gen artifact inside the feature-pack. It is now resolved from maven.


bstansberry (approved):

bstansberry:
Thanks @jfdenise

PR #16533, opened 2023-02-14
WFLY-17637 Remove version.jakarta.el.jakarta-el-api property from pom.xml

chengfang:
https://issues.redhat.com/browse/WFLY-17637


bstansberry:
Thanks @chengfang

PR #16536, opened 2023-02-15
[WFLY-14587] JCA subsystem: throw parsing exception if name for workm…

tadamski:
…anager pools is configured

https://issues.redhat.com/browse/WFLY-14587

I believe this is not a model change but a bug fix. The name was never a part of a model was a side effect of delegating parsing to jboss-threads. Otoh, I think that current naming in which thread pool matches workmanager name is pretty clean and I would vote for not changing it.

@ehsavoie FYI

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
@@ -512,14 +512,14 @@ public class JcaExtension implements Extension {
                 Namespace readerNS = Namespace.forUri(reader.getNamespaceURI());
                 switch (element) {
                     case LONG_RUNNING_THREADS: {
-
+                        checkName(reader);
                         org.jboss.as.threads.Namespace ns = org.jboss.as.threads.Namespace.THREADS_1_1;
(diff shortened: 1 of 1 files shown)


rhusar (changes requested):
The change makes `org.jboss.as.test.integration.jca.workmanager.WorkManagerThreadsCheckTestCase` fail.

tadamski:
@rhusar  fixed (I have reverted log removal as it is necessary with CLI usage)

bstansberry (approved):
Nice. I agree this is just a parsing bug fix. Most, maybe all, of our parsers, including this one, fail if they find unexpected attributes or elements; this is just adding a missing check for that.

bstansberry:
Having [long|short]-running-thread-pool be resources instead of ObjectTypeAttributeDefinition attributes is the fundamental flaw here; that's why these things have names. But the effort to correct that would be pretty high, even ignoring the backward compatibility code that would be needed. Between this and [WFLY-14579](https://issues.redhat.com/browse/WFLY-14579) we have reasonable behavior.

bstansberry:
Thanks @tadamski and @rhusar

PR #16538, opened 2023-02-15
[WFLY-17640]: Messaging ActiveMQ server cannot be created in management model because it requires dependency on legacy security domain.

ehsavoie:
* We no longer need to depend on a legacy security capability.
* Updating error message to be more explicit about what is required.

Jira: https://issues.redhat.com/browse/WFLY-17640


bstansberry (changes requested):
We still need the capability check in a Host Controller.

Perhaps:

1) Remove the default value from SECURITY_DOMAIN.
2) Add a setCapabilityReference call to SECURITY_DOMAIN.

Basically in domain.xml you no longer get to skip setting security-domain=other for your profile that manages a legacy HC.

I don't know if security-enabled=true without setting either elytron-domain or security-domain is a valid config. It seems the WildFlySecurityManager is meant to allow it in certain conditions.  If it's not valid, SecurityDomainCheckHandler can just enforce that one or the other is set. Otherwise delete it.

ehsavoie:
If i remove the default value that would break existing behaviour in legacy servers right ?
I'm not sureI fully understand the issue: on a WildFly 28 hte user would be suing either elytron or disable security. On a legacy HC the control would still be in place.

bstansberry:
> On a legacy HC the control would still be in place.

The capability verification needs to happen on the DC and not rely on a secondary HC enforcing it, which might not happen until some point long in the future when that HC connects.

Yes, it would break legacy configs; they would need to explicitly configure this value and not rely upon the default. This would be something the server migration tool would need to handle (i.e. if security enabled and elytron-domain not set, set security-domain to 'other'. @emmartins FYI.

bstansberry (approved):

bstansberry:
@ehsavoie FYI looking at the class you changed led me to file https://issues.redhat.com/browse/WFLY-17860

bstansberry:
Thanks @ehsavoie

messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/logging/MessagingLogger.java
```diff
@@ -884,4 +884,7 @@ public interface MessagingLogger extends BasicLogger {
     @Message(id = 106, value = "The bridge %s didn't deploy.")
     OperationFailedException failedBridgeDeployment(String bridgeName);
 
+    @Message(id = 107, value = "You must define a elytron security doman when security is enabled.")
```
mchoma, line 887: doman -> domain

PR #16539, opened 2023-02-15
WFLY-17641 Only install session management services for WARs

pferraro:
https://issues.redhat.com/browse/WFLY-17641


rhusar (approved):
Looks good.

PR #16541, opened 2023-02-16
WFLY-17561 Move test dependency which is dependent on the WFCORE

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-17561
Related WildFly Core issue: https://issues.redhat.com/browse/WFCORE-6203

### DIFF
--- a/boms/standard-test/pom.xml
+++ b/boms/standard-test/pom.xml
@@ -325,6 +325,94 @@
                 <version>${version.org.syslog4j}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>${version.junit}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>${version.org.mockito}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.byteman</groupId>
+                <artifactId>byteman-submit</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.byteman</groupId>
+                <artifactId>byteman</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.byteman</groupId>
+                <artifactId>byteman-bmunit</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.jboss.byteman</groupId>
+                <artifactId>byteman-install</artifactId>
+                <version>${version.org.jboss.byteman}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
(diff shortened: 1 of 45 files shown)


pom.xml
```diff
@@ -555,8 +555,10 @@
         <version.sun.jaxb>${version.org.glassfish.jaxb}</version.sun.jaxb>
         <version.wsdl4j>1.6.3</version.wsdl4j>
         <version.xml-resolver>1.2</version.xml-resolver>
-        <version.xom>1.3.8</version.xom>
```
yersan, line 558: `version.xom` was degraded here, is there any reason behind it?

parsharma: Sorry, I faced conflicts and at that time it was degraded. I have updated the version

yersan (changes requested):
Hi @parsharma there are still several dependencies at the main pom.xml with test scope. We want to avoid having pure test dependencies also on the main WildFly pom, so all of those dependencies should be moved to the test-bom

parsharma:
Hello @yersan,
Previously, I was not aware that we can use test bom in every module that's why some test maven dependencies were left. I am working on it, I will update the code soon.
Thanks

ee-feature-pack/common/pom.xml
```diff
@@ -51,6 +51,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>${ee.maven.groupId}</groupId>
+                <artifactId>wildfly-standard-test-bom</artifactId>
+                <version>${ee.maven.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
bstansberry, line 60: I don't think should be necessary. There's no testing involved with this module.

parsharma: Right, I have updated the code.
Thanks

yersan (approved):
LGTM

bstansberry:
The one CI job fails due to https://issues.redhat.com/browse/WFLY-17702. I'll fix that.

boms/standard-test/pom.xml
```diff
@@ -1,14 +1,12 @@
...
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
     <parent>
-        <artifactId>wildfly-parent</artifactId>
         <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-parent</artifactId>
         <version>28.0.0.Beta1-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
```
yersan, line 9: We should always avoid these sorts of formatting changes, they are not related to sorting of dependencies and they will introduce unnecessary conflicts with other branches, for example, conversion branches. 

Consistency with other pom.xml files could be also important; you can expect the same pattern on each pom, which can help when you are looking at the files or even using any script to make batch changes for all the poms.

So if we can avoid them, much better, I guess.

bstansberry (approved):

parsharma:
Sure @bstansberry.

bstansberry:
Thanks @parsharma and @yersan !!!!!

PR #16545, opened 2023-02-19
[WFLY-11609] EJB: deprecate default-entity-bean-instance-pool and def…

tadamski:
…ault-entity-bean-optimistic-locking attributes

https://issues.redhat.com/browse/WFLY-11609

iiuc the DEFAULT_ENTITY_BEAN_INSTANCE_POOL runtime logic has no effect, hence removing


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -475,10 +474,6 @@ protected void execute(DeploymentProcessorTarget processorTarget) {
             EJB3SubsystemDefaultPoolWriteHandler.SLSB_POOL.updatePoolService(context, model);
         }
 
-        if (model.hasDefined(DEFAULT_ENTITY_BEAN_INSTANCE_POOL)) {
-            EJB3SubsystemDefaultPoolWriteHandler.ENTITY_BEAN_POOL.updatePoolService(context, model);
-        }
-
         if (model.hasDefined(DEFAULT_SFSB_CACHE)) {
```
chengfang, line 477: below around line 502, any reason the similar is not applied to `DEFAULT_ENTITY_BEAN_OPTIMISTIC_LOCKING`?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
```diff
@@ -100,9 +100,11 @@ public class EJB3SubsystemRootResourceDefinition extends SimpleResourceDefinitio
                     .setAllowExpression(true).build();
     static final SimpleAttributeDefinition DEFAULT_ENTITY_BEAN_INSTANCE_POOL =
             new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.DEFAULT_ENTITY_BEAN_INSTANCE_POOL, ModelType.STRING, true)
+                    .setDeprecated(EJB3Model.VERSION_10_0_0.getVersion())
```
chengfang, line 103: It will be good to mark this attribute as @Deprecated too.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
```diff
@@ -100,9 +100,11 @@ public class EJB3SubsystemRootResourceDefinition extends SimpleResourceDefinitio
...
                     .setAllowExpression(true).build();
     static final SimpleAttributeDefinition DEFAULT_ENTITY_BEAN_OPTIMISTIC_LOCKING =
             new SimpleAttributeDefinitionBuilder(EJB3SubsystemModel.DEFAULT_ENTITY_BEAN_OPTIMISTIC_LOCKING, ModelType.BOOLEAN, true)
+                    .setDeprecated(EJB3Model.VERSION_10_0_0.getVersion())
```
chengfang, line 107: It will be good to mark this attribute as @Deprecated too.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
tadamski: It seems to me that DEFAULT_ENTITY_BEAN_OPTIMISTIC_LOCKING has some effect so I thought it would be better if we remove it the same time those deprecated attributes are.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
tadamski: fixed

chengfang (approved):

bstansberry:
Thanks @tadamski and @chengfang

PR #16548, opened 2023-02-20
[WFLY-17653] Upgrade OpenTelemetry to 1.23.1

jasondlee:
https://issues.redhat.com/browse/WFLY-17653

Bump version
Update archives to match upstream packageing reorgs

### DIFF
--- a/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
+++ b/ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
@@ -828,6 +828,16 @@
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-common</artifactId>
+      <licenses>
+        <license>
+          <name>Apache License 2.0</name>
+          <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+        </license>
+      </licenses>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-exporter-otlp-common</artifactId>
       <licenses>
         <license>
@@ -838,7 +848,7 @@
     </dependency>
     <dependency>
       <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-exporter-otlp-trace</artifactId>
+      <artifactId>opentelemetry-exporter-otlp</artifactId>
       <licenses>
         <license>
           <name>Apache License 2.0</name>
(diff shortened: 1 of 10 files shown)


ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
```diff
@@ -826,6 +826,16 @@
         </license>
       </licenses>
     </dependency>
+    <dependency>
```
jamezp, line 829: Think we're missing the `opentelemetry-api-logs` dependency in here.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/api/main/module.xml
```diff
@@ -25,6 +25,12 @@
     </properties>
     <resources>
         <artifact name="${io.opentelemetry:opentelemetry-api}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-api-logs}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-sdk}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-sdk-logs}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-sdk-metrics}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-sdk-trace}"/>
+        <artifact name="${io.opentelemetry:opentelemetry-sdk-common}"/>
```
jamezp, line 33: Is there a reason these were mostly moved out of `io.opentelemtry.trace`? This is explicitly named with "api" in the module name and these look like implementation details. Am I wrong?

jamezp (commented):
~~Is~~ Not sure how this got there.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/api/main/module.xml
jasondlee: I moved those to make them visible to.. something -- I can't recall off hand; I think it's in the Micrometer PR. To be honest, these should probably be in their own module, though I'm trying to reduce the churn in this PR. If you think it's important enough, I can reorg those modules.

ee-feature-pack/common/src/main/resources/license/ee-feature-pack-common-licenses.xml
jasondlee: Added.

ee-feature-pack/common/src/main/resources/modules/system/layers/base/io/opentelemetry/api/main/module.xml
jamezp: It feels a little weird to put the i the API IMO. I would opt more for a `io.opentelemetry.sdk` module over moving them here. Other opinions are welcome though :)

jamezp (approved):
I think this looks okay. It's a pretty big jump, but it sounds like we need to do it. 

@bstansberry the only minor question I'd have is if it's okay to rename `io.opentelemtry.trace` to `io.opentelemety.export`. It's a private module so it should be fine, but I wasn't sure if we have anything outside of WildFly using it. Maybe @jmesnil has an idea there as it would likely be something cloud related.

rhusar (changes requested):
@jasondlee BTW the CI failures are related and need fixing:

    java.lang.Exception: Some unreferenced modules are un-expected [io.opentelemetry.sdk]

jasondlee:
> @jasondlee BTW the CI failures are related and need fixing:
> 
> ```
> java.lang.Exception: Some unreferenced modules are un-expected [io.opentelemetry.sdk]
> ```

Right. Thanks. That's on my list for first thing Monday morning. :)

bstansberry:
Thanks @jasondlee, @jamezp and @rhusar

PR #16550, opened 2023-02-21
[WFLY-17138] Convert MicroProfile Metrics subsystem to Admin-Only Mode

jasondlee:
https://issues.redhat.com/browse/WFLY-17138

Convert module to legacy module
Add MigrateOperation to remove extension/subsystem from configurations


microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
```diff
@@ -73,19 +71,27 @@ public class MicroProfileMetricsSubsystemDefinition extends PersistentResourceDe
             .setAllowExpression(true)
             .build();
 
-    static final AttributeDefinition[] ATTRIBUTES = { SECURITY_ENABLED, EXPOSED_SUBSYSTEMS, PREFIX };
+    static final AttributeDefinition[] ATTRIBUTES = {SECURITY_ENABLED, EXPOSED_SUBSYSTEMS, PREFIX};
 
     protected MicroProfileMetricsSubsystemDefinition() {
         super(new SimpleResourceDefinition.Parameters(MicroProfileMetricsExtension.SUBSYSTEM_PATH,
                 MicroProfileMetricsExtension.getResourceDescriptionResolver(MicroProfileMetricsExtension.SUBSYSTEM_NAME))
                 .setAddHandler(MicroProfileMetricsSubsystemAdd.INSTANCE)
```
bstansberry, line 79: This can/should be 'new ModelOnlyAddStepHandler(ATTRIBUTES)'. MicroProfileMetricsSubsystemAdd can/should be removed.

testsuite/integration/microprofile-tck/pom.xml
```diff
@@ -200,6 +199,15 @@
...
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>**/metrics/**</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
```
bstansberry, line 210: What's this about? I don't think we should be excluding tests from the other MP TCKs.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsExtension.java
```diff
@@ -74,18 +76,23 @@ static ResourceDescriptionResolver getResourceDescriptionResolver(final boolean
...
-        registration.registerOperationHandler(GenericSubsystemDescribeHandler.DEFINITION, GenericSubsystemDescribeHandler.INSTANCE);
+
+        return Collections.singleton(registration);
     }
 
     @Override
-    public void initializeParsers(ExtensionParsingContext context) {
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, MicroProfileMetricsParser_1_0.NAMESPACE, MicroProfileMetricsParser_1_0::new);
+    public void initializeLegacyParsers(ExtensionParsingContext context) {
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, MicroProfileMetricsParser_1_0.NAMESPACE, new MicroProfileMetricsParser_1_0());
```
bstansberry, line 95: This should stay the way it was. A supplier ends up being created eventually (from a lambda) so moving to the constructor call saves nothing. And the standard idiom is to use a supplier for all but the current parser (and perhaps for it as well.) In some cases that results in a minor perf improvement, hence the standard idiom.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
```diff
@@ -20,12 +20,13 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-package org.wildfly.extension.microprofile.metrics._private;
+package org.wildfly.extension.microprofile.metrics;
```
bstansberry, line 23: There's a reason for this package, although it wasn't really carried through (here or in many other places that have similar packages.)

The module.xml should have:

```
    <exports>
        <exclude path="org/wildfly/extension/microprofile/metrics/_private"/>
    </exports>
```

The gist of it is this interface must be public but the module.xml bit ^^^ makes it inaccessible outside the module.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
```diff
@@ -43,7 +44,7 @@ public interface MicroProfileMetricsLogger extends BasicLogger {
     /**
      * A logger with the category {@code org.wildfly.extension.batch}.
      */
-    MicroProfileMetricsLogger LOGGER = Logger.getMessageLogger(MicroProfileMetricsLogger.class, "org.wildfly.extension.microprofile.metrics.smallrye");
+    MicroProfileMetricsLogger ROOT_LOGGER = Logger.getMessageLogger(MicroProfileMetricsLogger.class, "org.wildfly.extension.microprofile.metrics.smallrye");
```
bstansberry, line 48: Any of the other message methods that are no longer invoked should be commented out to save memory. We leave the source code in place but commented so the message # sequence is visible and message ids don't get re-used.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
```diff
@@ -83,4 +84,12 @@ public interface MicroProfileMetricsLogger extends BasicLogger {
     OperationFailedException seeDownstream();
     */
 
+    @Message(id = 10, value = "The migrate operation can not be performed: the server must be in admin-only mode")
+    OperationFailedException migrateOperationAllowedOnlyInAdminOnly();
+
+    @Message(id = 11, value = "Migration failed, see results for more details.")
+    String migrationFailed();
+
+    @Message(id = 12, value = "Cannot migrate 'secure-server' resource. This resource is not supported by the new elytron-oidc-subsystem yet.")
```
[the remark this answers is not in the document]
bstansberry: I suspect this message is not meant to be here. :)

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
```diff
@@ -83,4 +84,12 @@ public interface MicroProfileMetricsLogger extends BasicLogger {
     OperationFailedException seeDownstream();
     */
 
+    @Message(id = 10, value = "The migrate operation can not be performed: the server must be in admin-only mode")
+    OperationFailedException migrateOperationAllowedOnlyInAdminOnly();
+
+    @Message(id = 11, value = "Migration failed, see results for more details.")
```
bstansberry, line 90: s/,/;/g

bstansberry (changes requested):

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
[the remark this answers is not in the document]
jasondlee: You're right. Copy/paste error. Removed.

microprofile/metrics-smallrye/src/main/resources/org/wildfly/extension/microprofile/metrics/LocalDescriptions.properties
```diff
@@ -5,3 +5,11 @@ microprofile-metrics-smallrye.remove=Remove the subsystem
 microprofile-metrics-smallrye.security-enabled=True if authentication is required to access the HTTP endpoint on the HTTP management interface.\
   This attribute overrides the value specified in the security-enabled attribute of the /subsystem=metrics resource.
 microprofile-metrics-smallrye.exposed-subsystems=The names of the WildFly subsystems that exposes their metrics (or '*' to expose any subsystem metrics).
+
+
+microprofile-metrics-smallrye.describe-migration=Describe all the steps to remove legacy MicroProfile Metrics subsystem resources.
+microprofile-metrics-smallrye.describe-migration.reply.migration-error=
+microprofile-metrics-smallrye.describe-migration.migration-warnings=Any warnings about resources that could not be migrated.
+microprofile-metrics-smallrye.migrate=Remove the legacy MicroProfile metrics subsystem. Returns true if the subsystem migration has been actually performed and false otherwise.
```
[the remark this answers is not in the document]
jasondlee: Removed

microprofile/metrics-smallrye/src/main/resources/org/wildfly/extension/microprofile/metrics/LocalDescriptions.properties
```diff
@@ -5,3 +5,11 @@ microprofile-metrics-smallrye.remove=Remove the subsystem
 microprofile-metrics-smallrye.security-enabled=True if authentication is required to access the HTTP endpoint on the HTTP management interface.\
   This attribute overrides the value specified in the security-enabled attribute of the /subsystem=metrics resource.
 microprofile-metrics-smallrye.exposed-subsystems=The names of the WildFly subsystems that exposes their metrics (or '*' to expose any subsystem metrics).
+
+
+microprofile-metrics-smallrye.describe-migration=Describe all the steps to remove legacy MicroProfile Metrics subsystem resources.
+microprofile-metrics-smallrye.describe-migration.reply.migration-error=
```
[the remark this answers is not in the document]
jasondlee: Added

testsuite/integration/microprofile-tck/pom.xml
jasondlee: @bstansberry, the Fault Tolerance TCK has tests that exercise the _optional_ use of MP Metrics with FT. Without this exclusion, those tests run, and the TCKs fail. I think the exclusion is needed, but I can move it into the FT module if you'd prefer.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsExtension.java
jasondlee: Reverted

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
jasondlee: Done

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsLogger.java
jasondlee: Copy and paste. ;) Fixed.

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MigrateOperation.java
```diff
@@ -0,0 +1,152 @@
...
+    public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+        if (!describe && context.getRunningMode() != RunningMode.ADMIN_ONLY) {
+            throw ROOT_LOGGER.migrateOperationAllowedOnlyInAdminOnly();
+        }
+
+        final PathAddress subsystemsAddress = context.getCurrentAddress().getParent();
+        final Map<PathAddress, ModelNode> migrateOperations = new LinkedHashMap<>();
+
+        context.addStep((operationContext, modelNode) -> {
+            final List<String> warnings = new LinkedList<>();
```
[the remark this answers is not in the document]
jasondlee: Done

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
jasondlee: Done

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
```diff
@@ -73,19 +72,27 @@ public class MicroProfileMetricsSubsystemDefinition extends PersistentResourceDe
...
-                        MICROPROFILE_METRICS_SCAN));
+                .setCapabilities(MICROPROFILE_METRICS_HTTP_SECURITY_CAPABILITY, MICROPROFILE_METRICS_SCAN));
+    }
+
+    @Override
+    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+        for (AttributeDefinition att : ATTRIBUTES) {
+            resourceRegistration.registerReadWriteAttribute(att, null,
+                    new ReloadRequiredWriteAttributeHandler(ATTRIBUTES));
+        }
```
bstansberry, line 90: @jasondlee  FYI -- if you pass ATTRIBUTES as a param to the superclass constructor, you don't need to override this method.

The effect is the same, just a bit less code, so I'm not asking you to change this unless you change this branch for a different reason.

bstansberry (approved):

microprofile/metrics-smallrye/src/main/java/org/wildfly/extension/microprofile/metrics/MicroProfileMetricsSubsystemDefinition.java
jasondlee: Noted. Thanks.

fabiobrz:
> Thanks @jasondlee and @fabiobrz

Thank you @bstansberry - giant steps in here!

PR #16559, opened 2023-02-27
[WFLY-17470] Fix Mail Session cannot be injected when using CDI metho…

yersan:
…d producer

Jira issue: https://issues.redhat.com/browse/WFLY-17470

Weld threads explicitly set the context classloader to null, so they fail when Weld is validating the injection points since the Session providers cannot be loaded from the TCCL.


bstansberry (approved):

bstansberry:
Thanks @yersan

PR #16560, opened 2023-02-27
[WFLY-17663] Documentation should be updated to not mention elytron-tool.jar usage

skyllarr:
https://issues.redhat.com/browse/WFLY-17663


bstansberry:
Thanks @Skyllarr

PR #16570, opened 2023-03-02
[WFLY-17638][WFLY-17664] RESTEasy Upgrades

jamezp:
This PR contains upgrades for RESTEasy and RESTEasy MicroProfile. It also removes unneeded legacy `org.jboss.resteasy:resteasy-client` dependency. There is one extra commit, that can be removed, to remove the legacy-ejb-client module from the build as it no longer exists.

---
# RESTEasy 6.2.3.Final Upgrade

Tag: https://github.com/resteasy/resteasy/releases/tag/6.2.3.Final
Diff: https://github.com/resteasy/resteasy/compare/6.2.2.Final...6.2.3.Final

https://issues.redhat.com/browse/WFLY-17638

---
# RESTEasy MicroProfile 2.1.0.Final Upgrade

Tag: https://github.com/resteasy/resteasy-microprofile/releases/tag/2.1.0.Final
Diff: https://github.com/resteasy/resteasy-microprofile/compare/2.0.0.Final...2.1.0.Final

https://issues.redhat.com/browse/WFLY-17664

### DIFF
--- a/testsuite/integration/pom.xml
+++ b/testsuite/integration/pom.xml
@@ -98,7 +98,6 @@
                 <module>manualmode</module>
                 <module>secman</module>
                 <module>legacy</module>
-                <module>legacy-ejb-client</module>
                 <module>elytron</module>
                 <module>elytron-oidc-client</module>
             </modules>
@@ -230,15 +229,6 @@
             <activation><property><name>ts.legacy</name></property></activation>
             <modules>
                 <module>legacy</module>
-            </modules>
-        </profile>
-
-        <!-- -Dts.legacy.ejb.client -->
-        <profile>
-            <id>ts.integ.group.legacy.ejb.client</id>
-            <activation><property><name>ts.legacy.ejb.client</name></property></activation>
-            <modules>
-                <module>legacy-ejb-client</module>
             </modules>
         </profile>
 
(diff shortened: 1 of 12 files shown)


testsuite/integration/pom.xml
```diff
@@ -437,6 +427,11 @@
 
     <dependencies>
         <!-- Dependencies specific to the integration tests. -->
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <scope>test</scope>
+        </dependency>
```
bstansberry, line 434: Which child modules use this? If it's only a few please declare it in those specifically so we know where it is needed.

As part of the big-bang work we cleaned this one up to try and reduce what's listed to things that are basically used everywhere.

bstansberry (changes requested):
Please add WFLY-17547 to the commit message for the first commit and to the PR title. That's the issue that removed the legacy-ejb-client testsuite. Thanks for cleaning that up!

testsuite/integration/pom.xml
jamezp: TBH, I'm not too sure. I know `testsuite/integration/web` is what failed. I can just add it there and keep testing to see which ones fail. I assume it came in via the `org.jboss.resteasy:resteasy-client` which was previously in here.

jamezp: I forgot to note this was a runtime exception. I can have a look at where better it might fit though.

jamezp:
This passed for me locally, seems to be doing better on CI.

bstansberry:
Thanks @jamezp

PR #16574, opened 2023-03-04
[WFLY-17684] (pt 1) Consolidate elytron-oidc-client and servlet-feature-pack Galleon content

bstansberry:
First part of the https://issues.redhat.com/browse/WFLY-17684 work.

This part involves taking Galleon content that is made part of the wildfly-ee feature pack but is currently not in the ee-feature-pack source tree and moving it into ee-feature-pack.

The elytron-oidc-client source tree is collapsed into a single module, for the subsystem.

I also did a bit of tightening up of dependency exclusions, moving them into the boms where they belong and converting some specific exclusions to global ones.

This builds on #16578, which makes it simpler to do this kind of refactoring without breaking the Dependency Tree Input Build CI job.

It's best to review this one commit at a time.


fjuma (approved):
Thanks @bstansberry!

PR #16579, opened 2023-03-08
[WFLY-17699] fix test for latest IBM changes since ibmjdk-11.0.15.

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-17699

Making this Draft because this requires WFCORE change in https://github.com/wildfly/wildfly-core/pull/5404 and Elytron changes in https://github.com/wildfly-security/wildfly-elytron/pull/1722

Otherwise, `SpnegoMechTestCase` fails on IBM JDK 11.0.15.

cc @Skyllarr

### DIFF
--- a/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/http/SpnegoMechTestCase.java
+++ b/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/http/SpnegoMechTestCase.java
@@ -87,9 +87,11 @@ import org.jboss.as.test.integration.security.common.KDCServerAnnotationProcesso
 import org.jboss.as.test.integration.security.common.Krb5LoginConfiguration;
 import org.jboss.as.test.integration.security.common.Utils;
 import org.jboss.as.test.integration.security.common.servlets.SimpleServlet;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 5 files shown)


bstansberry:
@soul2zimate When the elytron bit is ready (i.e. merged and a release on its way), can this be updated to ignore the problem test case when IBM JDK is used so then we can merge this?  Same question applies to https://github.com/wildfly/wildfly-core/pull/5404.

My impression is to avoid breaking CI these 3 changes all need to happen near simultaneously. But that is not practical. So I'd rather disable the few tests that require that, merge your test changes, then let the upgrade come into core and then core into full. Once the upgrades are in we can unignore the few tests.

soul2zimate:
@bstansberry I don't see any CI jobs running on `IBM JDK`. If we consider them with `IBM JDK`, it would be a problem for them and WFCORE jobs like `Full integration - Linux / Windows`. I add a second commit to ignore the affected test as you suggested.

For Elytron update in WFCORE, IMHO it's fine. We just need wait until Elytron is ready and upgrade it along with https://github.com/wildfly/wildfly-core/pull/5404 in one PR. But sure, if @yersan @Skyllarr prefer to merge the WFCORE change early and ignore affected `KerberosNativeMgmtSaslTestCase` and `KerberosHttpMgmtSaslTestCase`, I can update in core as well.

yersan:
@soul2zimate , Regarding WildFly Core, I don't have any preferences, but if those failures are blocking someone running any other testsuite, ignoring them will be fine until getting the proper fix when all the pieces are ready. 

In case of ignoring them, I suggest using `Assume` and ignoring only the specific JDK (and Version if possible) but avoid ignoring the whole test class since we will lose coverage on the JDKs that are able to run the test without the pending fixes.

I think that was Brian's intention reading his question:

> can this be updated to ignore the problem test case **when IBM JDK is used** so then we can merge this?

soul2zimate:
update to skip `SpnegoMechTestCase` only for IBM JDK.

bstansberry:
Thanks @soul2zimate and @yersan

PR #16617, opened 2023-03-13
[WFLY-17741] Using wildcard (*) in combination with test-connection-in-pool operation on (non)XA datasource

khermano:
Issue: https://issues.redhat.com/browse/WFLY-17741

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/common/pool/PoolOperations.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/common/pool/PoolOperations.java
@@ -289,13 +289,17 @@ public abstract class PoolOperations implements OperationStepHandler {
                     .getDataSources().size());
             if (repository.getDataSources() != null) {
                 for (DataSource ds : repository.getDataSources()) {
-                    if (jndiName.equalsIgnoreCase(ds.getJndiName()) && ds.getPool() != null) {
+                    if (jndiName.equals("*")) {
+                        result.add(ds.getPool());
+                    }
+                    if ((jndiName.equalsIgnoreCase(ds.getJndiName()) && ds.getPool() != null)) {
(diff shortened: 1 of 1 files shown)


bstansberry:
Hi @khermano -- I don't think this will work and if it does it isn't how operations with a wildcard address are meant to work. If you use a wildcard in the address for an operation you should get a complex response that consists of multiple elements in a list, one for each actual resource that matched the wildcard.  Each element must include a section that shows the address of the specific resource, and then one that shows the normal response *for that resource*.  These must work correctly both in standalone mode and domain, and the wildcard must be allowed in any part of the address, not just the final element in the address.

This is quite complex to implement correctly, and adding this to a particular operation is a Feature Request; the absence of it for an operation is not a bug. As noted in https://issues.redhat.com/browse/WFLY-17741 I don't think we should attempt to do this, and if we do I don't any work should be done until a full eng/docs/qe feature team is formed with buy-in as to when it will be delivered.

bstansberry:
See also https://issues.redhat.com/browse/WFLY-17745

khermano:
Hi @bstansberry. Thank you for your review, quick response, and all the information. 
In such a case, I will close the PR and if you agree, I will resolve the issue with a Won't Do resolution? Or should I use something else?

bstansberry:
Hi @khermano -- Won't Do sounds fine.

khermano:
Thank you for the information @bstansberry. :)

khermano:
@bstansberry, in the JIRA description, you mentioned that it'd be a lot of work requiring experienced engineers' involvement. Would it be better to unassign the issue and close this PR so it is easier to pick it up?

bstansberry:
@khermano Yes, unassigning is fine. I'll go ahead and close this PR now.

PR #16618, opened 2023-03-13
[WFLY-17695] Create test case for MP HeaderPropagation

Dkafetzis:
https://issues.redhat.com/browse/WFLY-17695


jamezp (changes requested):
I'm assuming this won't work without a fix for https://issues.redhat.com/browse/WFLY-13397.

jamezp:
I'm replacing this PR with #16620 as it makes sense to include the tests with the fix.

PR #16626, opened 2023-03-16
WFLY-17732 - orphaned JDNI binding on deployment failure

baranowb:
Issue:  https://issues.redhat.com/browse/WFLY-17732

Hold on a merge. This needs investigation, as after rebase, behavior changed when it comes to exported JNDI space.

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
@@ -302,6 +302,10 @@ public class ModuleJndiBindingProcessor implements DeploymentUnitProcessor {
                     service = (BinderService) controller.getService();
                     if (!equals(service.getSource(), bindingConfiguration.getSource())) {
                         throw EeLogger.ROOT_LOGGER.conflictingBinding(bindingName, bindingConfiguration.getSource());
+                    } else {
+                        //IF we are here, source is the same deployment, so single ref. Above service will be removed, so no need to create
+                        //binder release not incrementing/acquiring - it will be removed as well.
+                        return;
                     }
                 }
                 //as these bindings are not child services
--- a/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
@@ -58,6 +58,7 @@ import org.jboss.msc.service.LifecycleListener;
 import org.jboss.msc.service.Service;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceController.Mode;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.msc.service.StartContext;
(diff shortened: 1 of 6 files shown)


baranowb:
This fails due to difference in how exported name space work in current WFLY?

rhusar:
BTW @baranowb this is still failing on CI.

baranowb:
Yeah, I saw, no idea why it passed localy... This is just annoying.

parsharma:
@baranowb could you please check CI is still failing?
Thanks

baranowb:
@parsharma Yeah, Ive seen it. I honestly have no idea, it does not fail locally, nor on anything other than docker image for some reason.

gaol:
@rhusar PR got updated and CI passed, would you please review ? thanks

tadamski:
I'm not very familiar with naming subsystem. I think this is a question to @emmartins

emmartins:
I will review this during this week

emmartins:
I can't replicate the issue with WildFly

ee/src/main/java/org/jboss/as/ee/component/deployers/ModuleJndiBindingProcessor.java
```diff
@@ -301,16 +302,44 @@ protected void addJndiBinding(final EEModuleConfiguration module, final BindingC
                     service = (BinderService) controller.getService();
                     if (!equals(service.getSource(), bindingConfiguration.getSource())) {
                         throw EeLogger.ROOT_LOGGER.conflictingBinding(bindingName, bindingConfiguration.getSource());
+                    } else {
```
emmartins, line 305: no matter no being able to replicate the original issue, and having to re-evaluate the JNDI leftovers after deploy fail, this change is not good, it assumes this is about a bind for same deployment yet this is wrong. There are for instance use cases where the same global JNDI is made by multiple deployments, which we should not fail and (unlike this change) skip the increase of the binding ref counter.

baranowb: are there any tests for that? if thats a real case, Im fine with ditching this. Nothing fail in TS with his change(AFAIR).

baranowb:
hmm. indeed something changed in how deployments are handled, thought scanner might have an update: https://issues.redhat.com/browse/WFCORE-6255 but its still not merged. However I can still see leftover bindings:
"applications" => {"app-2.0.0.war" => {
            "java:app" => {"AppName" => {
                "class-name" => "java.lang.String",
                "value" => "app-2.0.0"
            }},
            "modules" => {"app-2.0.0" => {"java:module" => {
                "InAppClientContainer" => {
                    "class-name" => "java.lang.Boolean",
                    "value" => "false"
                },
                "InstanceName" => {
                    "class-name" => "java.lang.Object",
                    "value" => "?"
                },
                "ModuleName" => {
                    "class-name" => "java.lang.String",
                    "value" => "app-2.0.0"
                },
                "Validator" => {
                    "class-name" => "jakarta.validation.Validator",
                    "value" => "?"
                },
                "ValidatorFactory" => {
                    "class-name" => "org.jboss.as.ee.beanvalidation.LazyValidatorFactory",
                    "value" => "org.jboss.as.ee.beanvalidation.LazyValidatorFactory@53df3034"
                }
            }}}
        }}
    }

with 
[standalone@localhost:9990 /] deployment-info
NAME          RUNTIME-NAME  PERSISTENT ENABLED STATUS 
app-2.0.0.war app-2.0.0.war false      true    FAILED

PR #16634, opened 2023-03-17
WFLY-17763 Upgrade to Hibernate ORM 6.2.0.CR4

scottmarlow:
https://issues.redhat.com/browse/WFLY-17763

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -474,7 +474,7 @@
         <version.org.hamcrest.legacy>1.3</version.org.hamcrest.legacy>
         <version.org.hibernate.commons.annotations>6.0.6.Final</version.org.hibernate.commons.annotations>
         <version.org.hibernate.search>6.1.8.Final</version.org.hibernate.search>
-        <version.org.hibernate>6.2.0.CR3</version.org.hibernate>
+        <version.org.hibernate>6.2.0.CR4</version.org.hibernate>
         <version.org.hibernate.validator>8.0.0.Final</version.org.hibernate.validator>
         <version.org.hornetq>2.4.9.Final</version.org.hornetq>
         <version.org.infinispan>14.0.7.Final</version.org.infinispan>


bstansberry (approved):

bstansberry:
@scottmarlow The Cloud Tests job fails due to https://github.com/wildfly-extras/wildfly-cloud-tests/issues/56, for which I'll send up a PR in a bit.

HHH-16316 would be a possible candidate for a Windup rule to update persistence.xml files, as the change looks to be a matter of repackaging deprecated classes:

org.hibernate.dialect -> org.hibernate.community.dialect

So a rule could either suggest switching to the non-deprecated variant of the dialect, and note that if the unsupported old dialect has been moved to org.hibernate.community.dialect.

bstansberry:
https://github.com/wildfly-extras/wildfly-cloud-tests/pull/57 is to fix the Cloud Tests.

bstansberry:
Thanks @scottmarlow

scottmarlow:
Thanks for addressing the cloud test #57 failure @bstansberry!  

Good point about HHH-16316 ([pr](https://github.com/hibernate/hibernate-orm/pull/6255)), https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/eap8/eap7/hibernate6.windup.xml#L444 does check for removed dialects that are now in the hibernate-community-dialects artifact:

> <dependency>
>     <groupId>org.hibernate.orm</groupId>
>     <artifactId>hibernate-community-dialects</artifactId>
>     <version>6.2.0.CR4</version>
> </dependency>
> 

I think the https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/eap8/eap7/hibernate6.windup.xml#L444 rule could be worded to reflect that the hibernate-community-dialects module is not included (by WildFly) and such should indicate that the persistence.xml is using a removed dialect.

PR #16636, opened 2023-03-20
[WFLY-15178] Split parse and marshalling tests into known groups. Cur…

jamezp:
…rently keeping the legacy tests disabled.

https://issues.redhat.com/browse/WFLY-15178


bstansberry (approved):

jamezp:
Note the jobs previously passed and got re-queued because of CI issues.

PR #16638, opened 2023-03-22
WFLY-17629 JPA/Hibernate 2nd level cache fails to invalidate updated entries when cache-container configured with ProtoStream marshaller

pferraro:
https://issues.redhat.com/browse/WFLY-17629

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jpa/persistence.xml
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jpa/persistence.xml
@@ -2,7 +2,7 @@
 <persistence xmlns="http://java.sun.com/xml/ns/persistence" version="2.0">
     <persistence-unit name="MainPU">
         <provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
-        <jta-data-source>java:jboss/datasources/ExampleDS</jta-data-source>
+        <jta-data-source>java:jboss/datasources/H2</jta-data-source>
         <shared-cache-mode>ALL</shared-cache-mode>
         <properties>
             <property name="hibernate.dialect" value="org.hibernate.dialect.H2Dialect"/>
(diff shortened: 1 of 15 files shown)


jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/CacheKeyImplementationMarshaller.java
```diff
@@ -0,0 +1,96 @@
...
+                    break;
+                case ENTITY_INDEX:
+                    entity = reader.readString();
+                    break;
+                case HASH_CODE_INDEX:
+                    hashCode = OptionalInt.of(reader.readUInt32());
+                    break;
+                default:
+                    reader.skipField(tag);
+            }
```
rhusar, line 71: This switch statement never parses out for tenant (TENANT_INDEX) which is thus always null.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jpa/persistence.xml
```diff
@@ -11,7 +11,7 @@
             <property name="hibernate.hbm2ddl.auto" value="update"/>
             <property name="hibernate.cache.use_second_level_cache" value="true"/>
             <property name="hibernate.cache.use_query_cache" value="true"/>
-            <property name="hibernate.cache.infinispan.entity.cfg" value="entity-replicated"/>
+            <!--property name="hibernate.cache.infinispan.entity.cfg" value="entity-replicated"/-->
```
rhusar, line 14: Consider deleting this altogether instead of commenting out.

rhusar (changes requested):
CacheKeyImplementationMarshaller needs a fix.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/CacheKeyImplementationMarshaller.java
pferraro: Whoops.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jpa/persistence.xml
pferraro: Indeed.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/CacheKeyImplementationMarshaller.java
pferraro: Clearly, these need unit tests...

rhusar (approved):
👍🏼 Looks good to me  now.

PR #16641, opened 2023-03-22
[WFLY-17756] Properly render adoc admonitions in the Github UI

khermano:
Issue: https://issues.redhat.com/browse/WFLY-17756

Unfortunately, the admonition must be in every individual document to be effective.
You can check my branch https://github.com/khermano/wildfly/tree/WFLY-17756_from_main_2, which has admonitions only in the top-level documents.
I was looking at docinfo files, but sadly, I couldn't find a workable solution yet.
For better visibility, I created multiple commits based on the directories.

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -4,6 +4,14 @@
 :icons: font
 :source-highlighter: coderay
 
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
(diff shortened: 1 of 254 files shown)


bstansberry:
Thanks @khermano!  There was a lot more to this than I expected, but it makes sense that each file would need this as the issue is about how github renders the adoc file itself! Much appreciated.

khermano:
Np, you're welcome.

PR #16642, opened 2023-03-22
[WFLY-17771] Remove domain.xml host-exclude entries for releases prio…

bstansberry:
…r to WF 23

https://issues.redhat.com/browse/WFLY-17771


yersan (approved):
LGTM

PR #16644, opened 2023-03-23
[WFLY-17773] Eliminate the transitive BouncyCastle dependency which was pulling in an older version.

darranl:
https://issues.redhat.com/browse/WFLY-17773
https://issues.redhat.com/browse/WFLY-17775

### DIFF
--- a/boms/standard-test/pom.xml
+++ b/boms/standard-test/pom.xml
@@ -62,6 +62,12 @@
                 <artifactId>apacheds-test-framework</artifactId>
                 <version>${version.org.apache.ds}</version>
                 <scope>test</scope>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.bouncycastle</groupId>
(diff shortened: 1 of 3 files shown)


bstansberry (approved):
@darranl LGTM.

A side question is why org.jboss.as.test.integration.logging.syslogserver is in testsuite/shared as it seems like only ts/int/elytron uses it. And TBH I'm not sure ts/int/elytron uses it as I can't find any use of it via simple IDEA usage search techniques. I suspect that package may be cruft, with the stuff from core's testsuite/shared being all that's used these days.

darranl:
TBH I didn't look I was more focused on the dependencies and compilation error, it does however sound like something only applicable to core.

jamezp (approved):

bstansberry:
RE that package, I filed https://issues.redhat.com/browse/WFLY-17790.

bstansberry:
Thanks @darranl and @jamezp

PR #16647, opened 2023-03-24
WFLY-17780 Add @ClientViewScoped into bean defining annotations

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-17780

CC @jasondlee @bstansberry 

This doesn't address the fact that these should likely be provided to Weld via some other means instead of hard-coding them but it makes the sample work.
Although I am fairly surprised there isn't any TCK test with `@ClientWindowScoped` that would be failing?


jasondlee:
I think this is fine. There's probably a bug in Mojarra I should look at, but this should be OK for now.

jasondlee (approved):

bstansberry:
Thanks @manovotn and @jasondlee

PR #16648, opened 2023-03-24
[WFLY-11173] The JPADefinition.DEPLOY_INSTANCE ResourceDefinition is not correct

petrberan:
https://issues.redhat.com/browse/WFLY-11173

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
@@ -70,20 +70,17 @@ public class JPADefinition extends SimpleResourceDefinition {
                     .setDefaultValue(new ModelNode(ExtendedPersistenceInheritance.DEEP.toString()))
                     .build();
 
-    JPADefinition(boolean feature) {
-        super(getParameters(feature));
+    JPADefinition() {
+        super(getParameters());
     }
 
-    private static Parameters getParameters(boolean feature) {
+    private static Parameters getParameters() {
         Parameters result = new Parameters(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME),
                 JPAExtension.getResourceDescriptionResolver())
-                .setFeature(feature).setRuntime(!feature);
-        if (feature) {
-            result = result.setCapabilities(JPA_CAPABILITY);
-        }
-        // TODO WFLY-11173 add/remove handlers don't make sense on the deployment resource
-        result = result.setAddHandler(JPASubSystemAdd.INSTANCE).setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE);
-
+                .setFeature(true)
+                .setCapabilities(JPA_CAPABILITY)
+                .setAddHandler(JPASubSystemAdd.INSTANCE)
+                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE);
         return result;
     }
 
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java
@@ -1,0 +1,42 @@
+package org.jboss.as.jpa.subsystem;
+
+import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.controller.operations.common.Util;
+import org.jboss.as.controller.operations.global.ReadAttributeHandler;
+import org.jboss.as.controller.registry.ManagementResourceRegistration;
+
+import java.util.List;
+
+import static org.jboss.as.jpa.subsystem.JPADefinition.DEFAULT_DATASOURCE;
+import static org.jboss.as.jpa.subsystem.JPADefinition.DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE;
+
+public class JPADeploymentDefinition extends SimpleResourceDefinition {
+
+    private static final PathElement ADDRESS = PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME);
+
+    JPADeploymentDefinition() {
+        super(getParameters());
+    }
+
+    private static Parameters getParameters() {
+        Parameters result = new Parameters(ADDRESS,
+                JPAExtension.getResourceDescriptionResolver())
+                .setFeature(false)
+                .setRuntime();
+        return result;
+    }
+
+    @Override
+    public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
+        for (AttributeDefinition attribute : List.of(DEFAULT_DATASOURCE, DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE)) {
+            resourceRegistration.registerReadOnlyAttribute(attribute, (context, operation) -> {
+                PathAddress subsystemAddress = context.getCurrentAddress().getParent().getParent().append(ADDRESS);
+                context.addStep(Util.getReadAttributeOperation(subsystemAddress, attribute.getName()), ReadAttributeHandler.RESOLVE_INSTANCE, context.getCurrentStage(), true);
+            });
+        }
+    }
+}
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java
@@ -1,0 +1,152 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.as.test.integration.jpa.deploymentsubsystemoperations;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.operations.common.Util;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.jboss.dmr.ModelNode;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.ADD;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILED;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FAILURE_DESCRIPTION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOVE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.WRITE_ATTRIBUTE_OPERATION;
+
+/**
+ * Regression tests for WFLY-11173
+ * Tests proper operations behavior in /deployment/subsystem=jpa case
+ *
+ * @author <a href="mailto:pberan@redhat.com">Petr Beran
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class JPADeploymentSubsystemOperationsTestCase {
+
+    private static ModelControllerClient controllerClient = TestSuiteEnvironment.getModelControllerClient();
+
+    private static final PathAddress DEPLOYMENT_PATH = PathAddress.pathAddress(PathElement.pathElement(DEPLOYMENT, JPADeploymentSubsystemOperationsTestCase.class.getSimpleName() + ".war"));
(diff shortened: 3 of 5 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
```diff
@@ -72,25 +74,41 @@ public class JPADefinition extends SimpleResourceDefinition {
...
     @Override
     public void registerAttributes(ManagementResourceRegistration registration) {
-        registration.registerReadWriteAttribute(DEFAULT_DATASOURCE, null, new ReloadRequiredWriteAttributeHandler(DEFAULT_DATASOURCE));
-        registration.registerReadWriteAttribute(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE, null, new ReloadRequiredWriteAttributeHandler(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE));
+        // WFLY-11173
+        // In case of /deployment=* the attributes cannot be written, only read. If read, the main /subsystem=jpa values are returned.
+        if (deployment) {
+            registration.registerReadOnlyAttribute(DEFAULT_DATASOURCE, (context, operation) -> {
+                PathElement subsystemAddress = PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME);
+                ModelNode modelNode = context.getOriginalRootResource().getChild(subsystemAddress).getModel().get(DEFAULT_DATASOURCE.getName());
```
bstansberry, line 99: This should be

context.readResourceFromRoot(subsystemAddress, false)

The only place where this matters is if this attribute was read as a step in a composite op that wrote the subsystem attribute. When that happens, the read should reflect the change made by the write. 

The getOriginalRootResource() is pretty much just for cases where an OSH needs to modify runtime services and as part of doing that it needs to know the original value of something. For example as part of rolling back a change in an op that failed.

jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADefinition.java
```diff
@@ -72,25 +74,41 @@ public class JPADefinition extends SimpleResourceDefinition {
...
+                context.getResult().set(modelNode);
+            });
+            registration.registerReadOnlyAttribute(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE, (context, operation) -> {
+                PathElement subsystemAddress = PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME);
+                ModelNode modelNode = context.getOriginalRootResource().getChild(subsystemAddress).getModel().get(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE.getName());
```
bstansberry, line 104: This should be

context.readResourceFromRoot(subsystemAddress, false)

bstansberry (changes requested):
@scottmarlow Please review.

pferraro:
Reusing the same class for both the subsystem model and deployment model seems problematic in general.  Why not use a separate class for the deployment resource definition?  They expose a different operations different set of attributes, and those attributes that are common will have different access characteristics.

petrberan:
Hi @bstansberry . What's your opinion on what pferraro asked above?

bstansberry:
@petrberan I think it's very good input. In the end this boolean being passed to the JPADefinition constructor is affecting pretty much all the behavior other than the 2 AttributeDefinitions. It's cleaner and less likely to produce bugs to use two classes that do what's appropriate in a straightforward way. (BTW the registerAdditionalRuntimePackages impl is not relevant to the deployment variant.)

If you want both variants could be static classes in the JPADefinition file so there it's easier to see how the 2 ADs are used for the two cases.

petrberan:
@pferraro @bstansberry Resolved all the requested changes, it allowed me to clean up the JPADefinition a little bit. Noticed an invalid test case in `testReadResourceOperation` (comparing 2 default values) and fixed that as well

petrberan:
@bstansberry  @pferraro Can I ask for a review please?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java
```diff
@@ -0,0 +1,175 @@
...
+        ModelNode writeAttributeModelNode = new ModelNode();
+        writeAttributeModelNode.get(OP).set(WRITE_ATTRIBUTE_OPERATION);
+        writeAttributeModelNode.get(OP_ADDR).set(SUBSYSTEM_PATH.toModelNode());
+        writeAttributeModelNode.get(NAME).set(DEFAULT_EXTENDED_PERSISTENCE_INHERITANCE);
+        writeAttributeModelNode.get(VALUE).set("SHALLOW");
+        controllerClient.execute(writeAttributeModelNode);
+
+        ModelNode subsystemModelNode = new ModelNode();
+        subsystemModelNode.get(OP).set(READ_RESOURCE_OPERATION);
+        subsystemModelNode.get(OP_ADDR).set(SUBSYSTEM_PATH.toModelNode());
```
pferraro, line 104:
```java
Util.getResourceResourceOperation(address);

File: testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java [hunk 4]
```diff
@@ -0,0 +1,175 @@
...
+        ModelNode subsystemResult = controllerClient.execute(subsystemModelNode);
+
+        ModelNode deploymentModelNode = new ModelNode();
+        deploymentModelNode.get(OP).set(READ_RESOURCE_OPERATION);
+        deploymentModelNode.get(OP_ADDR).set(DEPLOYMENT_PATH.append(SUBSYSTEM_PATH).toModelNode());
```
pferraro, line 109:
```java
Util.getResourceResourceOperation(address);

File: testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java [hunk 5]
```diff
@@ -0,0 +1,175 @@
...
+     * Tests the proper behavior for the {@code add} and {@code remove} operations on the JPA subsystem in a deployment.
+     * These operations should fail since they are unsupported in a deployment's subsystem.
+     *
+     * @throws IOException if an error occurs during the {@code ModelNode} execution.
+     */
+    @Test
+    public void testAddRemoveResource() throws IOException {
+        ModelNode removeModelNode = new ModelNode();
+        removeModelNode.get(OP).set(REMOVE);
+        removeModelNode.get(OP_ADDR).set(DEPLOYMENT_PATH.append(SUBSYSTEM_PATH).toModelNode());
```
pferraro, line 130:
```java
Util.createRemoveOperation(address);

File: testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java [hunk 6]
```diff
@@ -0,0 +1,175 @@
...
+        ModelNode removeResult = controllerClient.execute(removeModelNode);
+
+        checkForFailure(removeResult, REMOVE.toUpperCase(), "WFLYCTL0031");
+
+        ModelNode addModelNode = new ModelNode();
+        addModelNode.get(OP).set(ADD);
+        addModelNode.get(OP_ADDR).set(DEPLOYMENT_PATH.append(SUBSYSTEM_PATH).toModelNode());
```
pferraro, line 137:
```java
Util.createAddOperation(address);

File: testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java [hunk 7]
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @throws IOException if an error occurs during the {@code ModelNode} execution.
+     */
+    @Test
+    public void testWriteAttribute() throws IOException {
+        ModelNode writeAttributeModelNode = new ModelNode();
+        writeAttributeModelNode.get(OP).set(WRITE_ATTRIBUTE_OPERATION);
+        writeAttributeModelNode.get(OP_ADDR).set(DEPLOYMENT_PATH.append(SUBSYSTEM_PATH).toModelNode());
+        writeAttributeModelNode.get(NAME).set(DEFAULT_DATASOURCE);
+        writeAttributeModelNode.get(VALUE).set("Foobar");
```
pferraro, line 155:
```java
Util.getWriteAttributeOperation(address, DEFAULT_DATASOURCE, "Foobar");

File: jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java [hunk 8]
```diff
@@ -0,0 +1,40 @@
...
+import static org.jboss.as.jpa.subsystem.JPADefinition.DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE;
+
+public class JPADeploymentDefinition extends SimpleResourceDefinition {
+
+    JPADeploymentDefinition() {
+        super(getParameters());
+    }
+
+    private static Parameters getParameters() {
+        Parameters result = new Parameters(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME),
```
pferraro, line 20: I would recommend creating a static PathElement constant within this class that references the subsystem path, rather than calculating it multiple times.

jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java
```diff
@@ -0,0 +1,40 @@
...
+        resourceRegistration.registerReadOnlyAttribute(DEFAULT_DATASOURCE, (context, operation) -> {
+            PathAddress subsystemAddress = PathAddress.pathAddress(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME));
+            ModelNode modelNode = context.readResourceFromRoot(subsystemAddress, false).getModel().get(DEFAULT_DATASOURCE.getName());
+            context.getResult().set(modelNode);
+        });
+        resourceRegistration.registerReadOnlyAttribute(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE, (context, operation) -> {
+            PathAddress subsystemAddress = PathAddress.pathAddress(PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, JPAExtension.SUBSYSTEM_NAME));
+            ModelNode modelNode = context.readResourceFromRoot(subsystemAddress, false).getModel().get(DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE.getName());
+            context.getResult().set(modelNode);
+        });
```
pferraro, line 38: The logic in these operation handlers will not work in domain mode, since the subsystem resources are not found at the root.
Rather, I would have these handlers delegate to the actual read-attribute operation handler as registered with the subsystem resource, so that they correctly handle operation flags, like expression resolution, etc.

pferraro (changes requested):

darranl:
Hello @petrberan are you still working through this PR as we have some unresolved issues?  We are working to identify the final set of changes to merge into a WildFly 29 Alpha tag.

jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java
petrberan: Can you perhaps explain a bit what needs to be done to tackle this request @pferraro ?

I've been trying to change the code the past week in order to "delegate to the actual read-attribute operation handler as registered with the subsystem resource" and I'm getting either `StackOverflow`, `IllegalArgumentException` or any other exception no matter what I do.

However, if I keep these changes, then it works just fine even in domain mode. For example, when I change the two attributes to non-default value ("ExampleDS" and "SHALLOW") and read the JPA subsystem in `/host=primary/server=server-one/subsystem=jpa:read-resource`, I'm getting

```
"outcome" => "success",
    "result" => {
        "default-datasource" => "ExampleDS",
        "default-extended-persistence-inheritance" => "SHALLOW"
```

The same outcome when I read the attribute in the deployment `/host=primary/server=server-one/deployment=kitchensink.war/subsystem=jpa:read-resource`. When I change the `default-datasource` to an expression `${whatever.test:ExampleDS}` then after resolving attributes in BOTH the deployment and the subsystem, I'm getting either "ExampleDS" or "NonExampleDS" (if the `whatever.test` is indeed "NonExampleDS").

From my point of view, everything works as intended. What am I missing?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java
```diff
@@ -0,0 +1,152 @@
...
+    // Hardcoded to avoid adding the JPA module as a dependency
+    // Same goes for the "jpa" in SUBSYSTEM_PATH
+    private static final String DEFAULT_DATASOURCE = "default-datasource";
+    private static final String DEFAULT_EXTENDED_PERSISTENCE_INHERITANCE = "default-extended-persistence-inheritance";
+
+    @Deployment
+    public static Archive<?> deploy() {
+        WebArchive war = ShrinkWrap.create(WebArchive.class, JPADeploymentSubsystemOperationsTestCase.class.getSimpleName() + ".war");
+        war.addAsResource(JPADeploymentSubsystemOperationsTestCase.class.getPackage(), "persistence.xml", "META-INF/persistence.xml");
+        return war;
```
pferraro, line 77: This deployment is not sufficient to test this issue.
The JPA subsystem does not create /subsystem=jpa deployment resources unless it contains a PersistenceUnit.

jpa/subsystem/src/main/java/org/jboss/as/jpa/subsystem/JPADeploymentDefinition.java
pferraro:
```java
        for (AttributeDefinition attribute : List.of(DEFAULT_DATASOURCE, DEFAULT_EXTENDEDPERSISTENCE_INHERITANCE)) {
            resourceRegistration.registerReadOnlyAttribute(attribute, new OperationStepHandler() {
                @Override
                public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
                    PathAddress subsystemAddress = context.getCurrentAddress().getParent().getParent().append(ADDRESS);
                    context.addStep(Util.getReadAttributeOperation(subsystemAddress, attribute.getName()), ReadAttributeHandler.INSTANCE, context.getCurrentStage(), true);
                }
            });
        }
```

petrberan: Thank you, I'll test it manually to see if it behaves the same in standalone as well as in domain

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/deploymentsubsystemoperations/JPADeploymentSubsystemOperationsTestCase.java
petrberan: Is it really not sufficient? If the deployment subsystem resources are not present the tests fail as the deployment subsystem read-resource operation returns `undefined`. 

As long as there's `persistence-unit` in the persistence.xml, it gets picked by the [PersistenceUnitParseProcessor](https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java#L418) and is marked as a JPA Deployment and thus is handled by the [PersistenceUnitServiceHandler](https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java#L171)

pferraro: @petrberan You are indeed correct! :)
When I tested your branch locally, I had the persistence.xml mistakenly located within /META-INF/persistence.xml, instead of /WEB-INF/classes/META-INF/persistence.xml - which is the only valid location for a war.

I missed the logic with testReadResourceOperation(), which is the only "non-negative" test case - which, as you pointed out, would not succeed I my statement above had any validity...

petrberan:
@pferraro Thanks for helping me. All changes should be implemented, I've changed the `war.addAsResource` to `war.addAsWebInfResource` to make it more clear as well. Should be all good to go

pferraro (approved):

scottmarlow (approved):

darranl:
Thank you for the work to take this to completion @petrberan FYI I am being a bit cautious about interactions between PRs so I am just going to kick off CI again but I will try and get this merged today.

PR #16653, opened 2023-03-27
WFLY-17795 Application Client container is missing Mail, CDI, Dependency Injection, Messaging SPEC API classes

scottmarlow:
https://issues.redhat.com/browse/WFLY-17795


bstansberry (changes requested):
Thanks @scottmarlow.

Please organize the 'extension' elements alphabetically by module name. The order doesn't matter from a technical pov so keeping them alphabetical looks better and prevents people thinking the order does matter somehow.

Edit: Please put the new subsystems in their alphabetical location as well. We keep logging out of order for no-longer-relevant historical reasons, but the rest we put in order.

bstansberry:
@scottmarlow I retriggered the jobs that didn't work; some are running, others are queued.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #16657, opened 2023-03-27
WFLY-17681 Add Micrometer support to MP Fault Tolerance

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-17681

### DIFF
--- a/microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
+++ b/microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
@@ -39,7 +39,7 @@ import org.jboss.as.weld.WeldCapability;
 /**
  * This {@link DeploymentUnitProcessor} registers required CDI portable extension that adds support
  * for MP Fault Tolerance interceptor bindings. Moreover, it specifies which metrics provider to use according to
- * metrics integrations available at runtime.
+ * metrics integrations available at runtime (MP Metrics, Micrometer, or no metrics).
  *
  * @author Radoslav Husar
  */
@@ -66,14 +66,7 @@ public class MicroProfileFaultToleranceDeploymentProcessor implements Deployment
         // Configure which metrics provider to use
         Set<String> registeredSubsystems = deploymentUnit.getAttachment(Attachments.REGISTERED_SUBSYSTEMS);
 
-        MetricsIntegration metricsIntegration;
-
-        if (registeredSubsystems.contains("microprofile-metrics-smallrye")) {
-            metricsIntegration = MetricsIntegration.MICROPROFILE_METRICS;
-        } else {
-            metricsIntegration = MetricsIntegration.NOOP;
-        }
-
+        MetricsIntegration metricsIntegration = registeredSubsystems.contains("micrometer") ? MetricsIntegration.MICROMETER : MetricsIntegration.NOOP;
         ROOT_LOGGER.metricsProvider(metricsIntegration.name());
 
         weldCapability.registerExtensionInstance(new FaultToleranceExtension(metricsIntegration), deploymentUnit);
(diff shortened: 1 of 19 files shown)


microprofile/fault-tolerance-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/faulttolerance/deployment/MicroProfileFaultToleranceDeploymentProcessor.java
```diff
@@ -70,6 +70,8 @@ public void deploy(DeploymentPhaseContext phaseContext) {
 
         if (registeredSubsystems.contains("microprofile-metrics-smallrye")) {
             metricsIntegration = MetricsIntegration.MICROPROFILE_METRICS;
+        } else if (registeredSubsystems.contains("micrometer")) {
+            metricsIntegration = MetricsIntegration.MICROMETER;
```
jasondlee, line 74: I'm not sure you really need to check for MP Metrics here, as it is no longer part of the system, and any older configurations that did have it will be forced to migrate and remove it. Regardless, if you'd like to keep it for safety's sake or whatever, I would put Micrometer first, as you're more likely to find that.

jasondlee (approved):
One small comment which you can take or leave. Other than that, this looks good to me.

microprofile/galleon-common/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
```diff
@@ -40,15 +40,17 @@
         <module name="jakarta.enterprise.api"/>
         <module name="jakarta.enterprise.concurrent.api"/>
 
+        <module name="io.micrometer" optional="true"/>
         <module name="io.smallrye.common.annotation"/>
         <module name="io.smallrye.config"/>
-        <module name="io.smallrye.metrics"/>
+        <module name="io.smallrye.metrics" optional="true"/>
```
bstansberry, line 46: @rhusar @jasondlee @kabir I think the io.smallrye.metrics module can be removed altogether. The only thing that depends on it non-optionally is the org.wildfly.extension.microprofile.metrics-smallrye module and that dependency is a bug that should be removed. Once that is gone there is no scenario where the module will be provisioned in a slimmed server. If the smallrye deps of MicroProfile FT and MicroProfile RM can both function correctly without io.smallrye.metrics, then there's no reason to keep shipping it.

bstansberry:
The testsuite/integration/microprofile-tck/fault-tolerance/src/test/java/org/wildfly/extension/microprofile/faulttolerance/tck/CleanupMetricRegistries.java class seems like cruft now. I expect it probably fails, but doesn't result in TCK failure because the failure is part of undeploy.

microprofile/galleon-common/src/main/resources/modules/system/layers/base/io/smallrye/fault-tolerance/main/module.xml
rhusar: Thanks @bstansberry I have also realized tonight that this should just be gone instead of made optional here. I am pretty sure this is not required. I ll fix this in the morning.

rhusar: @bstansberry @jasondlee @kabir Opened https://issues.redhat.com/browse/WFLY-17806 to track removal and started working on this.

rhusar: PR is open - https://github.com/wildfly/wildfly/pull/16664

rhusar:
> The testsuite/integration/microprofile-tck/fault-tolerance/src/test/java/org/wildfly/extension/microprofile/faulttolerance/tck/CleanupMetricRegistries.java class seems like cruft now. I expect it probably fails, but doesn't result in TCK failure because the failure is part of undeploy.

Fixed in https://github.com/wildfly/wildfly/pull/16664.

jasondlee (approved):

bstansberry:
Thanks @rhusar @jasondlee

PR #16658, opened 2023-03-28
[WFLY-17144_followup] Add Micrometer support to WildFly

jasondlee:
https://issues.redhat.com/browse/WFLY-17144

- Move OTLP-specific config attributes to an attr group for future-proofing 
- Modify xsd to match
- PR feedback

### DIFF
--- a/observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
+++ b/observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
@@ -45,7 +45,7 @@ public class SubsystemParsingTestCase extends AbstractSubsystemBaseTest {
         super(MicrometerExtension.SUBSYSTEM_NAME, new MicrometerExtension());
         this.schema = schema;
         this.testXml = "<subsystem xmlns=\"" + this.schema.getNamespace() +
-                "\" endpoint=\"http://localhost:4318/v1/metrics\"></subsystem>";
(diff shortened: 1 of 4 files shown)


observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -45,7 +45,7 @@ public SubsystemParsingTestCase(MicrometerSubsystemSchema schema) {
         super(MicrometerExtension.SUBSYSTEM_NAME, new MicrometerExtension());
         this.schema = schema;
         this.testXml = "<subsystem xmlns=\"" + this.schema.getNamespace() +
-                "\" endpoint=\"http://localhost:4318/v1/metrics\"></subsystem>";
+                "\"><otlp-registry endpoint=\"http://localhost:4318/v1/metrics\"/></subsystem>";
```
bstansberry, line 48: This is a pre-existing flaw, but this test case needs beefing up to ensure proper testing of the basic management model behavior.

1) There should be coverage of a test xml that includes all attributes. The exposed-subsystems attribute is not covered. (Note that the this.testXml var here can be beefed up and used; you don't need another xml chunk.)
2) For any attribute that supports expressions, the xml value should be an expression of the form '${test.<attribute_name>:<adefaultvalue>}'.  The exact name of the expression property doesn't matter; the idea is to include an expression, but with a default value so the test isn't required to actually set the property.

bstansberry (approved):

jasondlee:
@bstansberry I'm working on some more testing for .Final, so I'll throw this one the stack as well.

PR #16659, opened 2023-03-28
[WFLY-15349] Use batch statement when creating ejb timer table in DatabaseTimerPersistence

petrberan:
https://issues.redhat.com/browse/WFLY-15349

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -398,14 +398,11 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service {
                 try {
                     String createTable = sql.getProperty(CREATE_TABLE);
                     String[] statements = createTable.split(";");
+                    statement = connection.createStatement();
                     for (final String sql : statements) {
-                        try {
-                            statement = connection.createStatement();
-                            statement.executeUpdate(sql);
-                        } finally {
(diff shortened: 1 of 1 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -398,13 +398,14 @@ private void checkDatabase() {
...
-                        } finally {
-                            safeClose(statement);
+                    try {
+                        statement = connection.createStatement();
+                        for (final String sql : statements) {
+                            statement.addBatch(sql);
                         }
+                        statement.executeBatch();
+                    } finally {
+                        safeClose(statement);
```
chengfang, line 408: The `statement` is closed twice: above at line 408, and below line 419. I think we can move the declaration of `statement` into line 400, and remove the close in line 419.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -398,13 +398,14 @@ private void checkDatabase() {
                 try {
```
chengfang, line 398: the nested try block (line 398 and line 401) seems unnecessary. We can combine them into one.

petrberan:
Thank you @chengfang 

Should be resolved now

chengfang (approved):

bstansberry:
Thanks @petrberan and @chengfang

PR #16667, opened 2023-03-29
WFLY-17807 Invalid @Schedule expression results in infinite loop

chengfang:
https://issues.redhat.com/browse/WFLY-17807


tadamski (approved):
LGTM

PR #16661, opened 2023-03-28
[WFLY-17802] Upgrade ironjacamar to 3.0.1.Final

tadamski:
https://issues.redhat.com/browse/WFLY-17802

PR #16670, opened 2023-03-30
[WFLY-17180][JBJCA-1402] added validation-query-timeout attribute to …

istudens:
…datasources

IronJacamar upgrade to 3.0.1.Final + WFLY-17180 fix.

Issue: 
https://issues.redhat.com/browse/WFLY-17802
https://issues.redhat.com/browse/WFLY-17180

This PR supersedes https://github.com/wildfly/wildfly/pull/16661.


tadamski (approved):
LGTM

bstansberry:
@tadamski @istudens We should not need to integrate a new feature in order to do a micro version bump of a component.

Components also shouldn't be adding features in micros.

PR #16679, opened 2023-03-31
[WFLY-17739] Remove uses of deprecated AbstractRemoveStepHandler methods

bstansberry:
https://issues.redhat.com/browse/WFLY-17739


bstansberry:
A bunch of ci.wildfly.org jobs blew up for environmental reasons, but the two that didn't had what look like related failures, which I'll need to sort.

pferraro (approved):

PR #16683, opened 2023-03-31
WFLY-17827 Eliminate extra write operation on session creation when cache is non-transactional

pferraro:
https://issues.redhat.com/browse/WFLY-17827


rhusar (approved):
👍🏼 LGTM

PR #16685, opened 2023-03-31
[WFLY-17810] Add more testing for the Micrometer extension

jasondlee:
https://issues.redhat.com/browse/WFLY-17810

- Replace Basic*TestCase with one that integrates w/otel collector
- Add OpenTelemetryCollectorContainer to wrap otel collector with Testcontainers
- Code clean up in main modules for issues found in testing
- Fortify subsystem parsing tests
- Add expression handling tests to subsystem partsing tests


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/metrics/WildFlyMetric.java
```diff
@@ -84,6 +84,7 @@ private ModelNode readAttributeValue(PathAddress address, String attributeName)
             // [WFLY-11933] if the value can not be read if the management resource is not accessible due to RBAC,
             // it is logged it at a lower level.
             if (error.contains("WFLYCTL0216")) {
+                System.err.println("***** blargh");
```
rhusar, line 87: Forgotten debug statement on System.err. :-)

rhusar (changes requested):

rhusar:
Also note that the changed test failed on windows, galleon and bootable-jar so it will need adjusting (skipping?) for those configs/platforms.

jasondlee:
> Also note that the changed test failed on windows, galleon and bootable-jar so it will need adjusting (skipping?) for those configs/platforms.

I have seen that and am investigating...

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -115,48 +103,36 @@ public void testSchema() throws Exception {
         super.testSchema();
     }
 
+    @Override
+    public void testSubsystem() throws Exception {
+        standardSubsystemTest(null);
```
bstansberry, line 108: If you put all the expressions in micrometer.xml you can then delete all the other test methods except this one and testSchema. And delte micrometer-expressions.xml All the other tests duplicate what standardSubsystemTest does.

bstansberry (changes requested):
This is ok except the imports comment. My comment about how to simplify SubystemParsingTestCase is optional.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -27,6 +12,13 @@
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 
+import java.io.IOException;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Locale;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.*;
+
```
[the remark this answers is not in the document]
jasondlee: Sorry. Stupid IDE setting. :)

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
jasondlee: @bstansberry Done. If you wouldn't mine verifying that I deleted only the tests you had in mind, I'd appreciate it. :)

bstansberry: @jasondlee Actually you can delete testSchema and testSubsystem as well. It turns out both are @Test methods in the superclass, and here in testSchema you just call the superclass and in the testSubsystem you do directly what the superclass does. So no need to override the superclass.

jasondlee: Ugh. I should have caught that. Changes incoming.

bstansberry:
@jasondlee Tests are failing, I believe because MicrometerSetupTask.startOpenTelemetryCollector() is failing.

Note that unfortunately the way the wf-arq integration works is that ServerSetup failures are silently swallowed, so you'll need logic to catch and react to them.

jasondlee:
> @jasondlee Tests are failing, I believe because MicrometerSetupTask.startOpenTelemetryCollector() is failing.
> 
> Note that unfortunately the way the wf-arq integration works is that ServerSetup failures are silently swallowed, so you'll need logic to catch and react to them.

Right. Thanks. Haven't had a chance to dig in just yet, but I'll try to get to that this afternoon.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/OpenTelemetryCollectorContainer.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String imageName = "otel/opentelemetry-collector";
+    private static final String imageVersion = "0.74.0";
+
+    private String otlpEndpoint;
+    private String prometheusUrl;
+
+    public OpenTelemetryCollectorContainer() {
+        super(DockerImageName.parse(imageName + ":" + imageVersion));
+        setExposedPorts(Lists.newArrayList(HTTP_OTLP_PORT, HEALTH_CHECK_PORT, PROMETHEUS_PORT));
+        setWaitStrategy(Wait.forHttp("/").forPort(HEALTH_CHECK_PORT));
```
bstansberry, line 37: Any reason to not wait for all 3?

Also, KeycloakContainer sets a startup timeout and startup attempts, presumably to try to make the oidc tests more robust.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
```diff
@@ -15,61 +15,112 @@
...
     private final ModelNode micrometerSubsystem = Operations.createAddress("subsystem", "micrometer");
+    private boolean extensionAdded = false;
+    private boolean subsystemAdded = false;
 
     @Override
-    public void setup(final ManagementClient managementClient, final String containerId) throws Exception {
-        if (modificationsNeeded()) {
-            execute(managementClient, Operations.createAddOperation(micrometerExtension), true);
+    public void setup(final ManagementClient managementClient, String containerId) throws Exception {
+        startOpenTelemetryCollector();
```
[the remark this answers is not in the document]
bstansberry: Also, I'm pretty sure a ServerSetupTask runs before any @BeforeClass method is invoked, so you need to check AssumeTestGroupUtil.isDockerAvailable().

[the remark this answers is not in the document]
jasondlee: I have that `isDockerAvailable()` call in a `@BeforeClass` in the test class. Will that work there, or should I test here as well?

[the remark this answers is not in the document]
jasondlee: @bstansberry :

> The FT uses of this setup don't need this collector afaict, so I suggest factoring this out or conditionally enabling it somehow.

I can do that, but:
- If the endpoint is configured with an non-functioning URL, the log will be filled with connected refused errors and stack traces
- Starting the collector in the test will mean another restart of WildFly, as I'll need to get the ports exposed by TC and configure WF accordingly. That's probably not a LOT of time, but it's at least something to consider.

If both of those are OK, I'll make the change posthaste.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/OpenTelemetryCollectorContainer.java
jasondlee: Added.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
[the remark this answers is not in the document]
bstansberry: @jasondlee Re refactoring...

The first point is true, although looking at https://ci.wildfly.org/repository/download/WF_MasterLinuxJdk11/363789:id/testsuite/integration/microprofile/target/wildfly/standalone/log/server.log there's only one connection refused log.

Re the 2nd I wasn't suggesting starting it in the test. I meant something like use a separate ServerSetup class for the collector setup. (A test can have more than one ServerSetup; they start in order of declaration and it's LIFO when they stop.

This aspect is ok as is. If the collector is like some other testcontainers uses and leads to intermittent failures we can revisit disabling it for the FT test.

Re isDockerAvailable()....

First, in the @BeforeClass method you need to use assumeDockerAvailable(), not isDockerAvailable().

Second, yes, you need to call isDockerAvailable() in the ServerSetupTask, as the setup() method is called before any @BeforeClass method on the test. Odd, but true. Arquillian even deploys any @Deployment before it will call the @BeforeClass. I think it's a case where @BeforeClass etc are handled by JUnit and Arq does all its stuff before letting JUnit do it's stuff.

To verify my memory I tossed some logging in your branch and this is what came out:

```
22:53:21,242 SEVERE [org.wildfly.test.integration.observability.micrometer.MicrometerSetupTask] (main) Starting OpenTelemetry Container
22:53:21,511 SEVERE [org.wildfly.test.integration.observability.micrometer.MicrometerSetupTask] (main) OpenTelemetry Container failed to start.
22:53:23,941 SEVERE [org.wildfly.test.integration.observability.micrometer.MicrometerOtelIntegrationTestCase] (main) @BeforeClass docker available? false
```

bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #16521, opened 2023-02-09
[WFLY-17616] Upgrade SmallRye Reactive Messaging to 4.4.0

kabir:
https://issues.redhat.com/browse/WFLY-17616

PR #16281, opened 2022-11-14
[WFLY-15405] AMQP Connector for reactive messaging

kabir:
https://issues.redhat.com/browse/WFLY-15405 - Support for the AMQP connector in the MP Reactive Messaging subsystem

Includes #17343 so this can be tested with the latest SmallRye Reactive Messaging which will likely be merged by the time this is tested

PR #16689, opened 2023-04-04
[WFLY-17839] Update mutiny vertx client to 3.0.0

kabir:
https://issues.redhat.com/browse/WFLY-17839

This should have come in with https://github.com/wildfly/wildfly/pull/16521 but was missed. Although the tests work without the upgrade, this is the version used by SmallRye Reactive Messaging 4.3.0. Also, https://github.com/wildfly/wildfly/pull/16281 fails without upgrade when rebased on main.


bstansberry (approved):

bstansberry:
Thanks @kabir

PR #16693, opened 2023-04-05
WFLY-17835: Consolidate license resources

michpetrov:
Issue: [WFLY-17835](https://issues.redhat.com/browse/WFLY-17835)

This is mostly to move everything that can be moved into a single location, and to update the merger file (currently it's putting an empty line between every tag). Updating the content of the configuration files will be done in another PR, when the license plugin gets updated. I have verified this generates files that are identical to the current setup.

I can squash this into a single commit, I just wanted to separate moving the text files and moving/merging the config files so it's clearer to review.


galleon-pack/galleon-feature-pack/pom.xml
```diff
@@ -72,6 +73,28 @@
...
+                                    <directory>${ee.galleon.shared.resources.directory}/packages/docs.licenses</directory>
+                                </resource>
+                                <resource>
+                                    <targetPath>${basedir}/target/resources/packages/docs.licenses.merge</targetPath>
+                                    <directory>${ee.galleon.shared.resources.directory}/packages/docs.licenses.merge</directory>
+                                </resource>
+                            </resources>
+                            <overwrite>true</overwrite>
+                        </configuration>
+                    </execution>
```
bstansberry, line 97: This is ok for now but we'll definitely refactor how this works at some point.

But this is fine; the whole business of using the maven-resources-plugin to compose feature packs needs to be replaced; the way you need to deal with this use case here is more evidence of that.

bstansberry (approved):

bstansberry:
Thanks @michpetrov -- good progress!

bstansberry:
BTW, I preferred the two commits -- each are complete pieces of work and easier to understand separated.

PR #16698, opened 2023-04-06
[WFLY-17852] Annotations in resource adapter module are not processed

gaol:
Issue: https://issues.redhat.com/browse/WFLY-17852

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
+++ b/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
@@ -35,6 +35,7 @@ import org.jboss.as.connector.metadata.xmldescriptors.IronJacamarXmlDescriptor;
 import org.jboss.as.connector.services.mdr.AS7MetadataRepository;
 import org.jboss.as.connector.services.resourceadapters.deployment.ResourceAdapterDeploymentService;
 import org.jboss.as.connector.services.resourceadapters.deployment.registry.ResourceAdapterDeploymentRegistry;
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/ParsedRaDeploymentProcessor.java
```diff
@@ -225,6 +226,7 @@ public static ServiceBuilder<ResourceAdapterDeployment> process(final ConnectorX
             } else {
                 builder.addDependency(ConnectorServices.CCM_SERVICE, CachedConnectionManager.class, raDeploymentService.getCcmInjector());
             }
+            builder.requires(ConnectorServices.BOOTSTRAP_CONTEXT_SERVICE.append(Constants.DEFAULT_NAME));
```
gaol, line 229: Depending on default bootstrap context service because when [the requiredWorkderContexts is set up](https://github.com/wildfly/wildfly/blob/28.0.0.Beta1/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/annorar/AnnoResourceAdapter.java#L56) in resource adapter, the [BootstrapContextCoordinator.getInstance().getDefaultBootstrapContext()](https://github.com/ironjacamar/ironjacamar/blob/3.0.0.Final/deployers/common/src/main/java/org/jboss/jca/deployers/common/AbstractResourceAdapterDeployer.java#L1208-L1217) needs to be set up as well.

gaol:
@tadamski would you please review ? thanks

tadamski (approved):

bstansberry:
Thanks @gaol and @tadamski

PR #16701, opened 2023-04-06
WFLY-17859 Channel add operations created by legacy parser can fail if run before jgroups subsystem is added

pferraro:
https://issues.redhat.com/browse/WFLY-17859

### DIFF
--- a/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemResourceDefinition.java
+++ b/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemResourceDefinition.java
@@ -40,10 +40,16 @@ import org.jboss.as.clustering.controller.SimpleResourceRegistrar;
 import org.jboss.as.clustering.controller.SubsystemRegistration;
 import org.jboss.as.clustering.controller.SubsystemResourceDefinition;
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationContext;
+import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.OperationStepHandler;
+import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.capability.RuntimeCapability;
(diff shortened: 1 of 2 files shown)


clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemResourceDefinition.java
```diff
@@ -110,11 +116,31 @@ public void register(SubsystemRegistration parentRegistration) {
...
+
+        ExecuteIfAbsentStepHandler(OperationStepHandler handler) {
+            this.handler = handler;
+        }
+
+        @Override
+        public void execute(OperationContext context, ModelNode operation) throws OperationFailedException {
+            PathAddress address = context.getCurrentAddress();
+            Resource parent = context.readResourceFromRoot(address.getParent(), false);
+            if (!parent.hasChild(address.getLastElement())) {
```
bstansberry, line 141: There is a still race between L140 and L142.

pferraro: @bstansberry Do you mean that operations generated by 2 subsystem parsers will execute concurrently?

bstansberry: @pferraro Yes, that is part of what parallel boot does. Operations are grouped by subsystem and the groups are executed concurrently. Different subsystems updating each other is not allowed. I must not have noticed this when reviewing #6736.

TBH this is probably good enough. The InfinispanSubsystemXMLReader logic has been there since 2014 and apparently wasn't a big problem. Your fix makes it even less of one. Over time parsing documents using these is getting less and less common, so the whole thing seems like an edge case. But, your marking the JIRA as Critical makes me wonder if I'm not understanding something.

pferraro: Yeah - the problematic logic in InfinispanSubsystemXMLReader only exists to support XML schemas from AS 7.x.  Hopefully, we can remove this soon?
I sent you a DM explaining the reason for the critical priority.

bstansberry:
Thanks @pferraro

PR #16713, opened 2023-04-07
[WFLY-17856] The undertow-servlet module needs a dependency on the Elytron EE Jakarta Authorization module

darranl:
It now implements an SPI from this module.

https://issues.redhat.com/browse/WFLY-17856


bstansberry:
Thanks @darranl

PR #16715, opened 2023-04-10
WFLY-17875 Channels added by ancient Infinispan schema can fail if no stack was defined

pferraro:
https://issues.redhat.com/browse/WFLY-17875

Verified by @emmartins


bstansberry:
Thanks @pferraro and @emmartins

PR #16723, opened 2023-04-11
WFLY-17881 Fix Jakarta EE 10 TCK jpa/core test failures (related to Byte[]/Character[] mapping changes in ORM 6.2)

scottmarlow:
https://issues.redhat.com/browse/WFLY-17881

Change the default for Byte[]/Character[] mapping to `legacy` to workaround Platform TCK test failures when running against Derby which doesn't support SQL Array type.


scottmarlow:
I'm not sure why yet but my prior change https://github.com/scottmarlow/wildfly/tree/hibernate.type.wrapper_array_handling worked but I'm still seeing local TCK failures with this change so I might need to use the https://github.com/scottmarlow/wildfly/tree/hibernate.type.wrapper_array_handling changes instead.  Will investigate further.

bstansberry:
Thanks @scottmarlow

PR #16730, opened 2023-04-12
[WFLY-17890] forces jacorb's migrate() op to use elytron security

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-17890

### DIFF
--- a/legacy/jacorb/src/main/java/org/jboss/as/jacorb/JacORBSubsystemConstants.java
+++ b/legacy/jacorb/src/main/java/org/jboss/as/jacorb/JacORBSubsystemConstants.java
@@ -77,6 +77,7 @@ public final class JacORBSubsystemConstants {
     public static final String NAMING = "naming";
     public static final String NAMING_EXPORT_CORBALOC = "export-corbaloc";
     public static final String NAMING_ROOT_CONTEXT = "root-context";
+    public static final String ELYTRON = "elytron";
     public static final String IDENTITY = "identity";
     public static final String INTEROP = "interop";
     public static final String INTEROP_SUN = "sun";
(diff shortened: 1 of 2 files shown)


tadamski (approved):
LGTM

bstansberry:
@emmartins @tadamski I think a change right after this line could be good:

https://github.com/wildfly/wildfly/blob/main/legacy/jacorb/src/main/java/org/jboss/as/jacorb/MigrateOperation.java#L156

Check 'jacorbModel' (not the transformed 'openjdkModel')  for 'identity', 'IDENTITY', 'client' or 'CLIENT' and if present add something to 'warnings'.  IIOPSubsystemAdd rejects client and identity at runtime so presumably there's a behavior difference between those and 'elytron' that justifies a warning.

emmartins:
@bstansberry From what I understood "client" and "identity" on jacorb were options using legacy security, which originally we would migrate to "identity" on iiop, but then we introduced Elytron and "elytron" option was added to iiop... And now we have no other option than to use "elytron" since we have no legacy security alternatives. 

Anyway I will leave up to @tadamski to decide with you how the op should behave, maybe it would even make sense for him to takeover on the PR.

bstansberry:
Agreed; this is Tomek's decision. If a config used 'client' or 'identity' and us changing it to 'elytron' isn't going to affect the server behavior, then for sure we should not include a warning message. If there is some difference the user might want to be aware of, the warning message is the way to provide that info without just failing the migration.

tadamski:
As there are differences in behavior between client, identity and elytron interceptors my vote would be to add such warning. I have opened a PR with it in https://github.com/emmartins/wildfly/pull/1. ElytronSASClientInterceptor has good javadoc so I decided to refer to it in the warning message.

bstansberry:
Thanks @emmartins and @tadamski. I'm going to merge this one as-is and will push up a separate PR for the https://github.com/tadamski/wildfly/tree/WFLY-17890 commit.

PR #16732, opened 2023-04-13
[WFLY-17886] IIOP: make sure that configuration of client security ty…

tadamski:
…pe installs requirement on legacy-security capability

https://issues.redhat.com/browse/WFLY-17886

client security configuration relies on legacy-security as well so it should behave the same way as identity security regarding capability registration.


bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #16734, opened 2023-04-14
[WFLY-17833] Make Angus Activation services visible to org.jboss.as.s…

yersan:
…ar module

Jira issue: https://issues.redhat.com/browse/WFLY-17833


yersan:
I'm not sure what should be the correct approach here, I would expect that the startService method of a MBean registration be executed under the application module classloader where the required services are already visible, however it uses the "org.jboss.as.sar" module classloader and I'm not sure if that is correct.

In any case, maybe this fix can be accepted as a workaround the issue for the incoming release.

bstansberry (approved):
Thanks @yersan. Can you file a JIRA to look into the TCCL aspect?

yersan:
Thanks for the review @bstansberry

> Can you file a JIRA to look into the TCCL aspect?
Sure, see [WFLY-17896](https://issues.redhat.com/browse/WFCORE-6196) for the next steps related to the TCCL

bstansberry:
Thanks @yersan

PR #16736, opened 2023-04-14
WFLY-17898 Deploy fails when legacy SSO is configured

pferraro:
https://issues.redhat.com/browse/WFLY-17898

Deploying a WAR associated with a host that configures legacy SSO will now successfully deploy, but will log the following:
```
17:03:12,610 WARN  [org.wildfly.extension.undertow] (MSC service thread 1-3) WFLYUT0112: Deployment root-app1.war associated with a host configured with legacy SSO: [
    ("subsystem" => "undertow"),
    ("server" => "default-server"),
    ("host" => "default-host"),
    ("setting" => "single-sign-on")
]


bstansberry:
@pferraro @fjuma @darranl @fl4via I don't understand the log WARN part.  Reading the discussion on https://issues.redhat.com/browse/JBEAP-22898, it seems that we've removed functionality without deprecating it. Ok, c'est la vie. But if we had deprecated it, what we likely would have done would be to leave the resource for DC use in a mixed-domain and for admin-only configuration migration use, while failing if the resource is configured on a non-admin-only server. Leaving the resource but having it semi-silently not function doesn't get around the point of deprecating things first, so we might as well just fail it if it's used in a normal mode server.

pferraro:
@bstansberry I've updated this to reject the legacy SSO add operation if not in admin-only mode.

bstansberry (approved):

fjuma (approved):
Thanks @pferraro! Just one minor nit, some of the commits are missing a WFLY issue reference. Would be good to update that.

bstansberry:
Thanks @pferraro and @fjuma

PR #16741, opened 2023-04-17
[WFLY-17605] Add in topology update wait time to stabilize before beginning test to avoid race condition.

rachmatowicz:
This PR does the following:
- for the test case TwoConnectorsEJBFailoverTestCase, which tests a single EJB client interacting with cluster members via the remoting protocol (on port 4447) and the remote+http protocol (on port 8080), adds in a delay to the first client interaction with the clustered deployment, to allow the cluster time to stablize when run on slow servers and prevent the client invoing on the cluster before it is ready.
-
For more details, see: https://issues.redhat.com/browse/WFLY-17605


pferraro:
Adding hold label.  I'll trigger a few runs to verify stabilization.

bstansberry:
@rachmatowicz @pferraro Any news on this one?

pferraro:
I think I triggered extra runs but forgot to verify them.  I'll have to go through the run history to verify (or re-run if those are lost).

bstansberry:
@pferraro @rachmatowicz Can we get this one through?

rhusar:
@bstansberry @pferraro @rachmatowicz Here's the solution to this problem of now knowing if this is stabilized - opened https://github.com/wildfly/wildfly/pull/18213 which incorporates this stabilization fix and marks the test as intermittent so that we can continuously test this in our clustering lab before enabling on upstream CI.

rhusar:
The change has been incorporated. This can be closed now.

PR #16746, opened 2023-04-17
[WFLY-17906] Remove invalid dependency io.opentelemetry:opentelemetry-sdk-extension

jbliznak:
https://issues.redhat.com/browse/WFLY-17906

Such artifact doesn't exist thus I am simply removing it. If instead there should have been different artifact, then please close this PR.


jasondlee (approved):
No, that looks good. Not sure where that came from, but my guess is a bad search/replace.

bstansberry:
Thanks @jbliznak and @jasondlee

PR #16747, opened 2023-04-18
[WFLY-17908] Removing JDK 8 specific excludes from maven enforcer plugin configurations

ropalka:
https://issues.redhat.com/browse/WFLY-17908

### DIFF
--- a/ee-feature-pack/common/pom.xml
+++ b/ee-feature-pack/common/pom.xml
@@ -3834,11 +3834,7 @@
                                 <configuration>
                                     <rules>
                                         <banTransitiveDependencies>
-                                            <excludes>
-                                                <!-- Ignore jdk jars because they are system scope -->
-                                                <exclude>com.sun:tools</exclude>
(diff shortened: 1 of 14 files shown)


bstansberry (approved):

bstansberry (changes requested):
@ropalka I'm doubtful about removing the plugin config entirely from the various poms.

See https://github.com/wildfly/wildfly/pull/16747/files#diff-b2fea476e535fb7131e979d33a9af3d8a5084b5ea53a003e80c32bfcba29867cR417-R418

The feature pack maven module is not checking the modules it integrates for transitives; it counts on those modules doing it themselves. This change turns that off, so we're effectively not checking for transitives any more.

Perhaps the design could be rethought, but changing it is out of scope for WFLY-17908, which is just about the JDK 8 specifics.

ropalka:
Fixed @bstansberry - thanks for identifying this issue!

bstansberry (approved):

bstansberry:
Thanks @ropalka

PR #16749, opened 2023-04-18
WFLY-17910 Deprecate org.wildfly.clustering.web.annotation.Immutable for removal

rhusar:
https://issues.redhat.com/browse/WFLY-17910


rhusar:
@pferraro Should we do this? :-)

darranl:
@rhusar / @pferraro Can we get a decision on if we should do this?  We have a couple of days to get the PR merged but would be good to have the decision so we know if this really needs to be on the list.

rhusar:
Hey @darranl, we have uncovered a problem with usage of this annotation where it doesn't work exactly as we have documented. So, let me close this until we fix it.

PR #16754, opened 2023-04-18
Disable JPA2LCTestCase and JpaStatisticsTestCase on OpenJ9

bstansberry:
https://ci.wildfly.org/viewLog.html?buildId=365232& is a custom run of the nightly OpenJ9 job with this in place.


bstansberry:
Closing; disabling these is not the right solution to the problem we are seeing.

PR #16757, opened 2023-04-19
[20.x] Upgrade jboss-ejb-client to allow running on Java 17

eirbjo:
Wildfly 20.x currently ships with a version of `jboss-ejb-client` which breaks the stricter encapsulation rules introduced in Java 17.

`jboss-ejb-client` has fixed this issue in later versions, so it would be good to update Wildfly 20.x to use version which works on Java 17 and later.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -376,7 +376,7 @@
         <version.org.jboss.activemq.artemis.integration>1.0.2</version.org.jboss.activemq.artemis.integration>
         <version.org.jboss.arquillian.core>1.4.0.Final</version.org.jboss.arquillian.core>
         <version.org.jboss.common.jboss-common-beans>2.0.1.Final</version.org.jboss.common.jboss-common-beans>
-        <version.org.jboss.ejb-client>4.0.33.Final</version.org.jboss.ejb-client>
+        <version.org.jboss.ejb-client>4.0.52.Final</version.org.jboss.ejb-client>
         <version.org.jboss.ejb-client-legacy>3.0.3.Final</version.org.jboss.ejb-client-legacy>
         <version.org.jboss.ejb3.ext-api>2.3.0.Final</version.org.jboss.ejb3.ext-api>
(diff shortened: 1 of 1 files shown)


bstansberry:
Hi @eirbjo -- Thanks for the report. If @chengfang and @tadamski are ok with this change to the 20.x branch I'm fine with merging it, but please note that we are not going to produce any further releases in the WildFly 20 series. We don't even have CI for that branch.

tadamski (commented):
I think that from ejb-client point of view this upgrade shouldn't cause any problems. I'm not sure it is going to help much though. Apart of lack of the release, my understanding is that there may be a number of other components requiring modifications to work with jdk17.

eirbjo:
> Apart of lack of the release, my understanding is that there may be a number of other components requiring modifications to work with jdk17.

My work here was mainly to assess the impact of running a Java EE 8 implementation on a JRE where the deprecated `java.security.Identity` class and some related classes are removed. This removal is an ongoing discussion in the OpenJDK project: https://bugs.openjdk.org/browse/JDK-8191136

With the change in this PR, I was able to build Wildfly 20 and run it on my custom Java 21 build where the APIs are removed.

So the modified Wildfly has been tested to run on Java 21. I used the `ejb-security-context-propagation` quickstart project to verify this:

https://github.com/wildfly/quickstart/tree/20.x/ejb-security-context-propagation

So while I did not run a full test of all features, at least this quickstart had no problems running on 21.

chengfang:
diff: https://github.com/wildfly/jboss-ejb-client/compare/4.0.33.Final...4.0.52.Final

There are many changes in the diff and they haven't been tested with WildFly 20.x. But if the new version works for your purpose, I have no objection.

chengfang (approved):

tadamski (approved):

bstansberry:
Thanks @eirbjo, @chengfang and @tadamski

PR #16759, opened 2023-04-20
[WFLY-17920] Move HostExcludesTestCase test case to the next iteratio…

yersan:
…n release

Jira issue: https://issues.redhat.com/browse/WFLY-17920


ropalka (approved):
Verified locally it  fixes the regression.

PR #16780, opened 2023-04-25
[WFLY-15324] Dynamically generating key/trust stores in Kafka SSL test

khermano:
Issue: https://issues.redhat.com/browse/WFLY-15324


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/reactive/messaging/kafka/ssl/RunKafkaWithSslSetupTask.java
bstansberry, line 1: @khermano Since we're creating the store files as part of the test execution now, please delete them in the tearDown method. Thanks!

bstansberry (changes requested):

kabir (approved):

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/reactive/messaging/kafka/ssl/RunKafkaWithSslSetupTask.java
khermano: Sorry about that. I changed that.

PR #16785, opened 2023-04-26
[WFLY-17960] Do not register the org.jboss.resteasy.plugins.server.se…

jamezp:
…rvlet.ResteasyBootstrap listener on the servlet. The servlet already handles creating this. This ends up crating two different RESTEasy deployments.

https://issues.redhat.com/browse/WFLY-17960

Upstream: #16784


bstansberry (approved):

bstansberry:
Thanks @jamezp

PR #16778, opened 2023-04-25
[WFLY-17953] Migrate to using the Shrinkwrap Resolvers for resolving …

jamezp:
…test artifacts rather than using non-public JBoss Modules utilities.

https://issues.redhat.com/browse/WFLY-17953

PR #16790, opened 2023-04-26
[WFLY-17953] Disable Maven Central when remote repositories are defin…

jamezp:
…ed using using the org.jboss.model.test.maven.repository.urls property.

https://issues.redhat.com/browse/WFLY-17953

Follows up on #16778

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/MavenArtifactResolver.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/MavenArtifactResolver.java
@@ -1,0 +1,99 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/shared/MavenArtifactResolver.java
```diff
@@ -0,0 +1,100 @@
...
+     * Resolves the Maven artifact and returns the artifact file associated with the coordinates.
+     *
+     * @param mavenCoords the maven coordinates
+     *
+     * @return the artifact file
+     */
+    public static File resolveSingleFile(final String mavenCoords) {
+        return createResolver("pom.xml")
+                .resolve(mavenCoords)
+                .withTransitivity()
```
bstansberry, line 48: The uses this is replacing all used .withoutTransitivity()

testsuite/shared/src/main/java/org/jboss/as/test/shared/MavenArtifactResolver.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenRemoteRepositories;
+import org.jboss.shrinkwrap.resolver.api.maven.repository.MavenRemoteRepository;
+
+/**
+ * A simple utility used to retrieve libraries for tests from Maven. If the artifact does not exist locally, it will
+ * be downloaded.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+// TODO (jrp) rename this
```
bstansberry, line 35: Can be removed.

bstansberry (changes requested):

PR #16798, opened 2023-04-28
[WFLY-14721] Add tests to XTS and transactions(JTA) subsystem for attributes that allow expression

petrberan:
https://issues.redhat.com/browse/WFLY-14721

### DIFF
--- a/transactions/src/test/resources/org/jboss/as/txn/subsystem/full-expressions.xml
+++ b/transactions/src/test/resources/org/jboss/as/txn/subsystem/full-expressions.xml
@@ -4,7 +4,7 @@
             <socket socket-binding="${test.socket-binding:txn-socket-id}" socket-process-id-max-ports="${test.socket.process.id-max.ports:10}"/>
         </process-id>
     </core-environment>
-    <recovery-environment socket-binding="${test.recovery.socket-binding:txn-recovery-environment" status-socket-binding="${test.recovery.status-socketbinding:txn-status-manager}"
+    <recovery-environment socket-binding="${test.recovery.socket-binding:txn-recovery-environment}" status-socket-binding="${test.recovery.status-socketbinding:txn-status-manager}"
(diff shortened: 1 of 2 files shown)


transactions/src/test/resources/org/jboss/as/txn/subsystem/full-expressions.xml
```diff
@@ -16,5 +16,5 @@
             <xid-location name="${test.table:fooTable}" batch-size="${test.batch:10}" immediate-cleanup="${test.immediate:false}"/>
         </commit-markable-resource>
     </commit-markable-resources>
-    
+    <client stale-transaction-time="${test.stale-transaction-time:600}"/>
```
mmusgrov, line 19: That is the default value but I guess it does no harm to specify it explicitly.

mmusgrov (approved):
The changes look fine, thanks very much.

jmfinelli:
Hi @petrberan. As [WFLY-14721](https://issues.redhat.com/browse/WFLY-14721) looks a bit out-of-date, could you please add a comment in that JIRA to explain why the modifications in this PR only partially address the description of the issue? Thanks

jmfinelli:
Thanks @petrberan. LGTM! I'll leave the approval to @mmusgrov

mmusgrov:
+1 the two changes still look OK

mmusgrov:
> @mmusgrov Parsing of the XML is `TransactionSubsystemxxParser` responsibility (the xx standing for 60/50/40/... depending on the version). The syntax error was noticed by me after a week of figuring out the problem with the `jndi-name` attribute - once the `socket-binding` was parsed into Model Node it was missing the usual "expression" prefix.
> 

Nice work.

> Perhaps a build regex check resulting in a warning if an incomplete expression was found would be a good addition

We need a regex expert, do we have any ;)

PR #16812, opened 2023-05-05
[WFLY-17991] / [WFLY-17976] ee-security should depend on elytron and ensure the JACC policy is enabled by default.

darranl:
https://issues.redhat.com/browse/WFLY-17991
https://issues.redhat.com/browse/WFLY-17976


docs/src/main/asciidoc/_elytron/Elytron_and_Jakarta_Authorization.adoc
```diff
@@ -11,33 +11,20 @@ endif::[]
...
-----
-
-After executing the command above, please reload the server
-configuration as follows:
-
-[source,options="nowrap"]
-----
-[standalone@localhost:9990 /] reload
-----
+specification. This policy provider is active default in the default
```
fjuma, line 21: s/active default/active by default

fjuma (approved):
Thanks @darranl!

bstansberry (approved):

bstansberry:
@darranl Failures seem to be related.

darranl:
Hopefully that fix does it.

One thing I have noticed I am sure we used to fail earlier but a lot of tests now are just ignoring set up failures and ploughing  on so we end up with a cryptic test failure after the set up failed rather than an error reporting what went wrong during set up so we can jump straight to the point that needs a fix.

darranl:
Added ready-for-merge label as this has two reviews and CI is now green.

PR #16813, opened 2023-05-05
WFLY-18002 ExpirationMetaData.isExpired() should return true if expiration time is exactly now.

pferraro:
https://issues.redhat.com/browse/WFLY-18002

### DIFF
--- a/clustering/ee/spi/src/main/java/org/wildfly/clustering/ee/expiration/ExpirationMetaData.java
+++ b/clustering/ee/spi/src/main/java/org/wildfly/clustering/ee/expiration/ExpirationMetaData.java
@@ -37,7 +37,7 @@ public interface ExpirationMetaData extends Expiration {
     default boolean isExpired() {
         if (this.isImmortal()) return false;
         Instant lastAccessedTime = this.getLastAccessTime();
-        return (lastAccessedTime != null) ? lastAccessedTime.plus(this.getTimeout()).isBefore(Instant.now()) : false;
+        return (lastAccessedTime != null) ? !lastAccessedTime.plus(this.getTimeout()).isAfter(Instant.now()) : false;
     }
 
     /**


clustering/ee/spi/src/main/java/org/wildfly/clustering/ee/expiration/ExpirationMetaData.java
```diff
@@ -37,7 +37,7 @@ public interface ExpirationMetaData extends Expiration {
     default boolean isExpired() {
         if (this.isImmortal()) return false;
         Instant lastAccessedTime = this.getLastAccessTime();
-        return (lastAccessedTime != null) ? lastAccessedTime.plus(this.getTimeout()).isBefore(Instant.now()) : false;
+        return (lastAccessedTime != null) ? !lastAccessedTime.plus(this.getTimeout()).isAfter(Instant.now()) : false;
```
kabir, line 40: I might be misunderstanding something but won't this still return false if lastAccessedTime+timeout == now?

pferraro: I don't think so.
Current behavior is:
```java
instant.isBefore(now)
```
i.e. instant < now.  However, we want this to be inclusive, such that instant <= now

```java
instant.isAfter(now)
```
means instant > now, thus the inverse:
```java
!instant.isAfter(now)
```
means instant <= now, which is what we want.

bstansberry: 👍 It's the subtle added '!' in the diff that's key.

kabir (approved):

bstansberry (approved):

bstansberry:
Thanks @pferraro and @kabir

PR #16814, opened 2023-05-05
WFLY-18002 ExpirationMetaData.isExpired() should return true if expiration time is exactly now.

pferraro:
https://issues.redhat.com/browse/WFLY-18002

Upstream PR: https://github.com/wildfly/wildfly/pull/16813


fjuma (approved):
Thanks @pferraro!

PR #16816, opened 2023-05-09
[WFLY-18004]: Upgrading messaging subystem version to 16.0

ehsavoie:
Upgrading to 16.0 to prepare for management model changes

Jira: https://issues.redhat.com/browse/WFLY-18004


bstansberry:
Included in #17202

PR #16818, opened 2023-05-09
Revert "WFLY-17958 Persistence container bytecode enhancement should …

jamezp:
…be enabled by default"

https://issues.redhat.com/browse/WFLY-17958


bstansberry:
@jamezp if this is ok with @scottmarlow (and i've seen discussion that makes me expect it is), then this is fine with me. Go ahead and merge it if he approves and it passes CI.

scottmarlow:
> @jamezp if this is ok with @scottmarlow (and i've seen discussion that makes me expect it is), then this is fine with me. Go ahead and merge it if he approves and it passes CI.

Sorry, I now have a different change in mind that I want to validate locally (I don't want to lose the new test and also would like to keep part of the other changes made like removing checking of old Hibernate settings.  I think that this pr can be closed.

PR #16782, opened 2023-04-25
WFLY-17958 Persistence container bytecode enhancement should be enabled by default

scottmarlow:
https://issues.redhat.com/browse/WFLY-17958

The workaround of setting `<property name="jboss.as.jpa.classtransformer" value="true"/>` is provided in the issue.  We have supported this property for many years and will likely continue to support it in the future as it has served us well with Hibernate ORM.  With other persistence providers, we have always defaulted `jboss.as.jpa.classtransformer` on.

This issue + pull request is about changing the default for `jboss.as.jpa.classtransformer` from false to true, which means applications will now need to opt out of (JPA/Persistence container managed) bytecode enhancement if they wish.

PR #16819, opened 2023-05-09
WFLY-17958 Persistence container bytecode enhancement should be disabled by default

scottmarlow:
https://issues.redhat.com/browse/WFLY-17958

This change reverts part of the https://github.com/wildfly/wildfly/pull/16782 change by reversing the default for ByteCode enhancement to be disabled by default.  

To summarize, with this change, bytecode enhancement is disabled by default now.

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
@@ -238,6 +238,7 @@ public class Configuration {
     private static final String EE_DEFAULT_DATASOURCE = "java:comp/DefaultDataSource";
     // key = provider class name, value = module name
     private static final Map<String, String> providerClassToModuleName = new HashMap<String, String>();
+    public static final String HIBERNATE = "Hibernate";
 
     static {
         // always choose the default hibernate version for the Hibernate provider class mapping
@@ -281,7 +282,14 @@ public class Configuration {
         if (pu.getProperties().containsKey(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER)) {
             return Boolean.parseBoolean(pu.getProperties().getProperty(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER));
         }
+        final String provider = pu.getPersistenceProviderClassName();
+        if (provider != null && provider.contains(HIBERNATE)) {
+            return false;
+        }
         return true;
+    }
+    private static boolean isHibernateProvider(String provider) {
+        return provider == null || provider.contains("Hibernate");
     }
 
     // key = provider class name, value = adapter module name
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
@@ -238,7 +238,7 @@ public class Configuration {
     private static final String EE_DEFAULT_DATASOURCE = "java:comp/DefaultDataSource";
     // key = provider class name, value = module name
     private static final Map<String, String> providerClassToModuleName = new HashMap<String, String>();
-    public static final String HIBERNATE = "Hibernate";
+    private static final String HIBERNATE = "Hibernate";
 
     static {
         // always choose the default hibernate version for the Hibernate provider class mapping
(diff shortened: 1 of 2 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
```diff
@@ -238,6 +238,7 @@ public class Configuration {
     private static final String EE_DEFAULT_DATASOURCE = "java:comp/DefaultDataSource";
     // key = provider class name, value = module name
     private static final Map<String, String> providerClassToModuleName = new HashMap<String, String>();
+    public static final String HIBERNATE = "Hibernate";
```
jamezp, line 241: This should be private.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
```diff
@@ -281,8 +282,15 @@ public static boolean needClassFileTransformer(PersistenceUnitMetadata pu) {
...
             return Boolean.parseBoolean(pu.getProperties().getProperty(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER));
         }
+        final String provider = pu.getPersistenceProviderClassName();
+        if (provider != null && provider.contains(HIBERNATE)) {
+            return false;
+        }
         return true;
     }
+    private static boolean isHibernateProvider(String provider) {
+        return provider == null || provider.contains("Hibernate");
```
jamezp, line 292: Why not use the constant here? Also shouldn't it be:
```java
return provider != null && provider.contains(HIBERNATE);
```

That would match what's above. If not, this can be deleted as I don't see it used anywhere.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/hibernate/classfiletransformertest/ClassFileTransformerTestCase.java
```diff
@@ -91,10 +92,10 @@ public void testhibernate_ejb_use_class_enhancer() throws Exception {
     }
 
     @Test
-    public void testHibernateByteCodeEnhancement() {
+    public void testHibernateByteCodeEnhancementIsDisabledByDefault() {
         // Note: ManagedTypeHelper is an internal Hibernate ORM class, if it is removed or renamed then this test can be updated
         // accordingly.
-        assertTrue("Employee class is not bytecode enhanced", org.hibernate.engine.internal.ManagedTypeHelper.isManagedType(Employee.class));
+        assertFalse("Employee class is not bytecode enhanced", org.hibernate.engine.internal.ManagedTypeHelper.isManagedType(Employee.class));
```
jamezp, line 98: Should this be "Employee class is bytecode enhanced" if it's disable by default and we want to ensure the bytecode enhancing has not been done?

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
```diff
@@ -281,8 +282,15 @@ public static boolean needClassFileTransformer(PersistenceUnitMetadata pu) {
         if (pu.getProperties().containsKey(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER)) {
             return Boolean.parseBoolean(pu.getProperties().getProperty(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER));
         }
+        final String provider = pu.getPersistenceProviderClassName();
+        if (provider != null && provider.contains(HIBERNATE)) {
+            return false;
+        }
         return true;
```
jamezp, line 289: Couldn't this just be:
```java
return isHibernateProvider(provider);
```

jamezp (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
scottmarlow: thanks for catching ^

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
scottmarlow: Fixing

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
scottmarlow: I'll explain why in a comment.

jamezp (approved):

PR #16834, opened 2023-05-11
Module refactors (WFLY-17684, WFLY-17918)

bstansberry:
This is a combination of #16786 and #16833. Meant to test the combination and it resolves the merge conflict between the two.


bstansberry:
#16833 was merged independently so I'm going to close this. #16786 is the for the other part.

PR #16840, opened 2023-05-12
[WFLY-18016] Refactoring - eliminating usage of deprecated CapabilityServiceBuilder.addCapabilityRequirement() methods

ropalka:
https://issues.redhat.com/browse/WFLY-18016

### DIFF
--- a/datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceOperations.java
+++ b/datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceOperations.java
@@ -25,11 +25,14 @@ import static org.jboss.as.controller.security.CredentialReference.handleCredent
 import static org.jboss.as.controller.security.CredentialReference.rollbackCredentialStoreUpdate;
 import static org.wildfly.extension.datasources.agroal.AbstractDataSourceDefinition.CREDENTIAL_REFERENCE;
 
-import jakarta.transaction.TransactionSynchronizationRegistry;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
+import io.agroal.api.AgroalDataSource;
 import io.agroal.api.configuration.supplier.AgroalConnectionFactoryConfigurationSupplier;
 import io.agroal.api.configuration.supplier.AgroalConnectionPoolConfigurationSupplier;
 import io.agroal.api.configuration.supplier.AgroalDataSourceConfigurationSupplier;
+import jakarta.transaction.TransactionSynchronizationRegistry;
 import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.AbstractRemoveStepHandler;
 import org.jboss.as.controller.CapabilityServiceBuilder;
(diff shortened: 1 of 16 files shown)


datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceService.java
```diff
@@ -27,6 +27,7 @@
 import io.agroal.api.transaction.TransactionIntegration;
 import io.agroal.narayana.NarayanaTransactionIntegration;
 import org.ietf.jgss.GSSException;
+import org.jboss.as.server.deployment.DelegatingSupplier;
```
bstansberry, line 30: I'd prefer not to use the server module as a utility lib.  

Extensions can use the deployment package in DUPs, but this isn't a DUP. (At some point we should factor out that package into its own module.)

AIUI using this just eliminates the need for some null checks Optional.isPresent checks.

ropalka: Fixed @bstansberry  . There was introduced a new DelegatingSupplier class in EE subsystem and AGROAL subsystem was rewritten to don't use DelegatingSupplier at all.
We're using DelegatingSupplier not because of Optional.isPresent() checks but because there are some places where supplier injection happens later than at the service installation time.

datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceOperations.java
```diff
@@ -39,6 +40,12 @@
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceName;
+import org.wildfly.common.function.ExceptionSupplier;
+import org.wildfly.security.auth.client.AuthenticationContext;
+import org.wildfly.security.credential.source.CredentialSource;
+
+import java.util.function.Consumer;
+import java.util.function.Supplier;
```
bstansberry, line 48: Don't change this unless there is some other change needed, but the imports here are out of order.

The jakarta.transaction one was already out of order due to my sed change from javax to jakarta, which took it out of the "java. and javax. before others" category.

datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceService.java
```diff
@@ -115,7 +126,7 @@ public void start(StartContext context) throws StartException {
 
         if (jta || xa) {
             TransactionManager transactionManager = ContextTransactionManager.getInstance();
-            TransactionSynchronizationRegistry transactionSynchronizationRegistry = transactionSynchronizationRegistryInjector.getValue();
+            TransactionSynchronizationRegistry transactionSynchronizationRegistry = transactionSynchronizationRegistrySupplier.get();
```
bstansberry, line 129: I think transactionSynchronizationRegistrySupplier can be null if jta is false.

datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceOperations.java
ropalka: Fixed

datasources-agroal/src/main/java/org/wildfly/extension/datasources/agroal/DataSourceService.java
ropalka: Correct - thanks for spotting this out ! It is fixed now.

bstansberry (approved):

bstansberry:
Thanks @ropalka

PR #16845, opened 2023-05-16
[WFLY-16013]: Print more informative failure messages when working with broadcast/discovery group resources

ehsavoie:
The broadcast-group and discovery-group resources has been deprecated and substituted by socket-*-group and jgroups-*-group variants.
These new resources have different validation rules than the original resources, which may cause confusion for users who are not aware of these changes.

Jira: https://issues.redhat.com/browse/WFLY-16013


TomasHofman (approved):
Thanks, I think it's still desirable change :+1:

bstansberry:
Thanks @ehsavoie and @TomasHofman

PR #16846, opened 2023-05-16
[WFLY-17349] cloned org.jboss.as.test.integration.jpa.hibernate.entit…

istudens:
…y entities into seperate entities for webtxem tests

Issue: https://issues.redhat.com/browse/WFLY-17349

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/webtxem/TestServlet.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/webtxem/TestServlet.java
@@ -35,10 +35,10 @@ import jakarta.servlet.http.HttpServletResponse;
 import jakarta.transaction.Status;
 import jakarta.transaction.UserTransaction;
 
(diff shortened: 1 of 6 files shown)


istudens:
@scottmarlow can you review it please? thanks

scottmarlow (approved):

rhusar (approved):
I am also under impression that this will fix the issue - basically what's happening is we are creating/deleting and recreating the same tables in rapid succession in H2 and their related structures, but it's just racey in H2. Using different entities should solve the situation. Although this solution is not exactly elegant, I would prefer to e.g. implement a custom table mapping or use a different schema name.

Also note that while this might not fail frequently in our lab it fails quite frequently on faster machines which makes this very annoying :-)

Rerunning the failed unrelated tests to give this a chance to get in eventually.

bstansberry:
Thanks @istudens, @scottmarlow and @rhusar

rhusar:
I am afraid this did not help afterall :-(

> Caused by: org.hibernate.exception.SQLGrammarException: could not prepare statement [Table "WEBTXEMFLIGHT" not found; SQL statement:

PR #16841, opened 2023-05-12
[WFLY-18017] Remove elytron profile remnants

bstansberry:
https://issues.redhat.com/browse/WFLY-18017

PR #16855, opened 2023-05-18
[WFLY-18017] Do not execute the enable-embedded-messaging.cli script …

jamezp:
…by default.

https://issues.redhat.com/browse/WFLY-18017

Follows up on #16841 which in combo with #16833 seems to have created an issue.


fjuma (approved):
Thanks @jamezp!

PR #16856, opened 2023-05-18
[WFLY-15260] Add support for securing the management console with OIDC

fjuma:
https://issues.redhat.com/browse/WFLY-15260

Depends on https://github.com/wildfly/wildfly-core/pull/5509 and https://github.com/hal/console/pull/853
Includes https://github.com/wildfly/wildfly/pull/16768

Note that tests for accessing the management console when secured with OIDC cannot currently be added to the WildFly testsuite. The redirect from HAL to the Keycloak login page is handled by javascript so we can't just use `HttpClient` to simulate the log in like we do with the existing OIDC tests. We are looking into adding tests to the HAL testsuite instead.


skyllarr (approved):

darranl (approved):

fjuma:
I've rebased this.

fjuma:
Some of the jobs were hitting strange compilation failures due to the Core upgrade not being picked up for some reason.

PR #16859, opened 2023-05-19
WFLY-18036 Marshalling optimizations are not getting applied to @SessionScoped @Stateful EJBs

pferraro:
https://issues.redhat.com/browse/WFLY-18036


pferraro:
Rebasing so that we can test this along with recent fixes.

darranl:
@rhusar looks like one more here asking for your review I may be able to bundle in if you are Ok with it?

rhusar (approved):
I think this looks good (also we have pretested this in a subset of QE tests) - cc @darranl

PR #16863, opened 2023-05-21
[WFLY-17798][WFLY-18033] Deprecate the javax.* alias modules; replace dependencies on them with the jakarta.* ones

bstansberry:
https://issues.redhat.com/browse/WFLY-17798
https://issues.redhat.com/browse/WFLY-18033

Builds on #16786


yersan (approved):
We need to do the same with the modules shipped with  wildfly-core

jbliznak:
```
     (Note that an xml comment is used here instead of the usual 'jboss.api=deprecated'
     property child element only because the 'module-alias' element and associated
     Java classes do not support property child elements.)
```
I think it would be beneficial to allow marking module-alias as deprecated as it seems a perfect fit for such scenario (deprecate something in favor of something else but still provide compatibility via alias, however, give users a warning about future required migration)

bstansberry:
> ```
>      (Note that an xml comment is used here instead of the usual 'jboss.api=deprecated'
>      property child element only because the 'module-alias' element and associated
>      Java classes do not support property child elements.)
> ```
> 
> I think it would be beneficial to allow marking module-alias as deprecated as it seems a perfect fit for such scenario (deprecate something in favor of something else but still provide compatibility via alias, however, give users a warning about future required migration)

@jbliznak I agree it would be good but this would be a significant RFE. AIUI an alias module is maintained in JBoss Modules solely as a string in a Set<String> maintained by the Module object that represents the target module. If you access a Module object using the alias name you get the target module, not something specific to the alias. To do anything useful with anything else parsed from a module-alias element file we would have to change what is stored and somehow expose that. Without breaking anything.

Also, from a JBoss Modules point of view, the jboss.api property we use to mark deprecation is just an opaque property. It's WildFly Core that gives it meaning. To JBM, logically an alias module and its target should have the same properties; otherwise one is no longer an alias.

Via https://issues.redhat.com/browse/WFCORE-6279 @yersan has put something in place to address a good chunk of the use case of warning about using alias modules.

@ropalka re this conversation.

darranl:
@bstansberry FYI this one needs a rebase

PR #16864, opened 2023-05-21
[WFLY-18039] Remove uses of deprecated AttributeDefinition API

bstansberry:
https://github.com/bstansberry/wildfly/pull/new/WFLY-18039


yersan:
Failed test are unrelated to this Jira; added an integration Job to get a bit more confidence 

https://ci.wildfly.org/buildConfiguration/WF_WildFlyCoreIntegrationExperiments/371856

pferraro:
Rather than need to reference each attribute twice per read/write, I would suggest adding a utility methods to the parsers/writers to both reduce verbosity and likelihood of bugs/regressions, e.g.
https://github.com/wildfly/wildfly/blob/main/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemXMLReader.java#L945
https://github.com/wildfly/wildfly/blob/main/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemXMLWriter.java#L232

pferraro (changes requested):
Everything looks technically correct, but see my comment above.

bstansberry:
@pferraro Thanks. I added a commit to clean out the dual use of the same constant in one line.

pferraro (approved):

yersan (approved):

PR #16866, opened 2023-05-22
[WFLY-18040] EJB: make system wide client interceptors a static part …

tadamski:
…of client context service

https://issues.redhat.com/browse/WFLY-18040

I have created this fix as an optimization of EJBClientInterceptors usage. I think IdentityEJBClientInterceptor can be configured globally (and later possibly other interceptors of this kind) so that EJBClientContext can be shared between all deployments if no additional metadata is provided.

I also plan to create another fix to OptionMap configuration propagation (which seems to don't work now), but this optimization should be able to solve customer problems for now.

@fjuma @chengfang could you please review?


chengfang (approved):

fjuma (approved):
Thanks @tadamski!

PR #16869, opened 2023-05-23
WFLY-18052 Drop problematic transitive dependency exclusions for org.wildfly.core:wildfly-controller and org.wildfly.core:wildfly-server

pferraro:
https://issues.redhat.com/browse/WFLY-18052

This makes it impossible to run full-integration tests from wildfly-core in the event that that a new dependency is added to one of these modules.

### DIFF
--- a/bean-validation/pom.xml
+++ b/bean-validation/pom.xml
@@ -139,12 +139,6 @@
         <dependency>
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-server</artifactId>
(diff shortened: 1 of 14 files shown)


jamezp (approved):
I think this makes sense. IMO exclusions should be happening where the inherited dependency is defined anyway, unless there is some special reason as to why this is not possible.

bstansberry:
Thanks @pferraro and @jamezp

PR #16872, opened 2023-05-24
[WFLY-17891] Updating DeploymentModulesListTestCase to not require that WildFly installs a static module with a slot

khermano:
Issue: https://issues.redhat.com/browse/WFLY-17891

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/moduleslisting/DeploymentModulesListTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/moduleslisting/DeploymentModulesListTestCase.java
@@ -29,6 +29,7 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.api.ServerSetup;
 import org.jboss.as.arquillian.container.ManagementClient;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.client.helpers.Operations;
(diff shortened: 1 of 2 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/moduleslisting/DeploymentModulesListTestCase.java
```diff
@@ -25,10 +25,12 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
+import integration.deployment.moduleslisting.UserModuleSetupTask;
```
bstansberry, line 28: @khermano Please use the fully qualified class name, not just the relative part.

After ~ 25 years working with Java, I had no idea you could do this!

bstansberry (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/moduleslisting/DeploymentModulesListTestCase.java
bstansberry: Ah, I suspect you can't do that. This doesn't fail because the UserModuleSetupTask that gets used is in the same package and doesn't require an import. So the incorrect import statement is never used. Please remove it.

khermano: I am sorry. I accidentally removed the beginning of the package path in the UserModuleSetupTask, which is why such an import exists.
I fixed it. Thank you for the review.

bstansberry (approved):

bstansberry:
Thanks @khermano!

khermano:
Thank you for the review :)

PR #16881, opened 2023-05-25
[WFLY-18055] Consume upstream Mojarra artifacts directly

jasondlee:
https://issues.redhat.com/browse/WFLY-18055

Remove com.sun.faces dependency
Add org.glassfish:jakarta-faces dependency
Modify modules to reflect dependency changes


ee-feature-pack/galleon-shared/src/main/resources/license/licenses.xml
```diff
@@ -5067,4 +5067,5050 @@
             </licenses>
         </dependency>
     </dependencies>
+||||||| parent of e966483f7c ([WFLY-18055] Investigate how to consume upstream Mojarra directly)
```
fjuma, line 5070: This should be removed.

jasondlee: The joy of merges. :)

jsf/subsystem/src/test/resources/modules/org/jboss/as/jsf-injection/1.2/module.xml
```diff
@@ -33,9 +33,10 @@
     </resources>
```
fjuma, line 33: We could probably remove this since JSF 1.2 won't work anymore here.

fjuma, line 33: We could probably remove this since JSF 1.2 won't work anymore here.

clustering/faces/mojarra/pom.xml
```diff
@@ -54,6 +54,8 @@
         <dependency>
             <groupId>jakarta.faces</groupId>
```
fjuma, line 55: Just noticed that `org.glassfish` is used as the `groupId` for the `jakarta.faces` `artifactId` in the file above but here `jakarta.faces` is used as the `groupId`.

Is that intentional?

clustering/faces/mojarra/pom.xml
```diff
@@ -54,6 +54,8 @@
...
             <artifactId>jakarta.faces-api</artifactId>
+            <!-- Temporary override until upstream publishes update -->
+            <version>4.0.1</version>
```
fjuma, line 58: Why is this needed?

jasondlee: With the fix above, it is not. I'll get these both cleaned up.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/faces/api/main/module.xml
```diff
@@ -24,23 +24,50 @@
...
+                    <path name="jakarta/faces/flow/builder"/>
+                    <path name="jakarta/faces/lifecycle"/>
+                    <path name="jakarta/faces/mock"/>
+                    <path name="jakarta/faces/model"/>
+                    <path name="jakarta/faces/push"/>
+                    <path name="jakarta/faces/render"/>
+                    <path name="jakarta/faces/validator"/>
+                    <path name="jakarta/faces/view"/>
+                    <path name="jakarta/faces/view/facelets"/>
+                    <path name="jakarta/faces/webapp"/>
```
fjuma, line 64: Have we done any checks to make sure that the number of classes before the change and after the change match?

jasondlee: I have not verified the number of classes, but I pulled all of those package names directly from the source via `grep | cut | sort`, so it should be exhaustive.

jsf/subsystem/src/test/resources/modules/org/jboss/as/jsf-injection/1.2/module.xml
jasondlee: Done

ee-feature-pack/galleon-shared/src/main/resources/license/licenses.xml
jasondlee: Done

fjuma (approved):
Thanks @jasondlee!

PR #16885, opened 2023-05-26
WFLY-18065 Distributed @SessionScoped @Stateful EJBs require excessive cache transactions per invocation

pferraro:
https://issues.redhat.com/browse/WFLY-18065

### DIFF
--- a/weld/ejb/src/main/java/org/jboss/as/weld/ejb/StatefulSessionObjectReferenceImpl.java
+++ b/weld/ejb/src/main/java/org/jboss/as/weld/ejb/StatefulSessionObjectReferenceImpl.java
@@ -52,6 +52,7 @@ public class StatefulSessionObjectReferenceImpl implements SessionObjectReferenc
     private final SessionID id;
     private final StatefulSessionComponent ejbComponent;
     private final Map<Class<?>, ServiceName> viewServices;
+    private volatile boolean removed = false;
 
     private transient Map<String, ManagedReference> businessInterfaceToReference;
 
@@ -122,6 +123,7 @@ public class StatefulSessionObjectReferenceImpl implements SessionObjectReferenc
     @Override
     public void remove() {
         try (StatefulSessionBean<SessionID, StatefulSessionComponentInstance> bean = this.ejbComponent.getCache().findStatefulSessionBean(this.id)) {
+            this.removed = true;
             if (bean != null) {
                 bean.remove();
             }
@@ -130,8 +132,6 @@ public class StatefulSessionObjectReferenceImpl implements SessionObjectReferenc
 
     @Override
     public boolean isRemoved() {
-        try (StatefulSessionBean<SessionID, StatefulSessionComponentInstance> bean = this.ejbComponent.getCache().findStatefulSessionBean(this.id)) {
-            return (bean == null) || bean.isRemoved();
-        }
+        return this.removed;
     }
 }


weld/ejb/src/main/java/org/jboss/as/weld/ejb/StatefulSessionObjectReferenceImpl.java
```diff
@@ -125,13 +126,12 @@ public void remove() {
             if (bean != null) {
                 bean.remove();
             }
+            this.removed = true;
```
bstansberry, line 129: @pferraro Should this be the first line in the method? There's a bit of a race here otherwise, where the bean can be gone but isRemoved() returns false. Moving this line first there is still a race, but it's a problem only if it matters whether isRemoved() returns true before that actual removal work is done.

The only use I see of isRemoved() is in this class itself, but I think my IDE isn't finding uses of SessionObjectReference properly.

BTW, both this code and the old code don't comply with the behavior specified in the SessionObjectReference javadoc.

pferraro: > Should this be the first line in the method?

My original thinking what that, if the remove handler on the bean triggers some other invocation handled by weld, it would otherwise fail with a NoSuchEJBException.
However, I think any invocation triggered within the beans remove handler would be handled by the EJB container, rather than weld.

> There's a bit of a race here otherwise, where the bean can be gone but isRemoved() returns false. 

I think that's a valid point.  While the container is required to serialize SFSB invocations, this object resides outside of the synchronization interceptor, so technically the remove() method could be invoked by concurrent threads.

pferraro: I've updated this, such that isRemoved() will no longer return false if the bean is in the process of being, or has already been, removed.

bstansberry: Thanks for the discussion, Paul.

bstansberry (approved):

manovotn:
@bstansberry @pferraro this PR breaks one of Weld's tests as it makes it unable to determine if the bean reference was removed or not in time.
The test can be seen here - https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/enterprise/weld1234/Weld1234Test.java
With the original backing issue here - https://issues.redhat.com/browse/WELD-1234

The short of it is that if you have a stateful bean whose method throws an exception, such bean should be removed and any subsequent method invocation on that bean should be able to detect that via `org.jboss.weld.ejb.api.SessionObjectReference`.
The above linked Weld test shows that's not the case.
Weld has a [special case handling](https://github.com/weld/core/blob/master/modules/ejb/src/main/java/org/jboss/weld/module/ejb/EnterpriseBeanProxyMethodHandler.java#L126-L128) for invocations of `toString` on a removed EJB in which case it won't throw `NoSuchEJBException` - looking at the [forum reference](https://developer.jboss.org/message/772357) from Weld issue, this is because of logging.

Can we revert this or look for another way of handling it?

pferraro:
@manovotn Thanks for pointing this out.  I will revisit this today and modify the fix accordingly.

pferraro:
Looking into this further, it is my opinion that the original fix for WELD-1234 was not ideal - as it made the assumption that SessionObjectReference.isRemoved() is inexpensive and thus requires that it be called on *every* stateful ejb invocation.  As I see it, the best way forward would be to modify weld in one of the following ways:

1. Leave this PR in place, and instead remove the isRemoved() check per invocation and instead catch a possible NoSuchEjbException during method invocation and return a fabricated value if the method was toString().
   e.g.
```java
    try {
        Object returnValue = Reflections.invokeAndUnwrap(proxiedInstance, method, args);
        BeanLogger.LOG.callProxiedMethod(method, proxiedInstance, args, returnValue);
        return returnValue;
    } catch (NoSuchEjbException e) {
        if (isToStringMethod(method)) {
            return businessInterface.getName() + " [REMOVED]";
        }
        throw e;
    }
```
2. Revert this PR and modify EnterpriseBeanProxyMethodHandler.invoke(...) to first test whether the method is toString() before calling isRemoved(), and thus short-circuit the condition before calling the potentially expensive operation.
    e.g.
```java
        if (isToStringMethod(method) && reference.isRemoved()) {
            return businessInterface.getName() + " [REMOVED]";
        }
```

IMO, #1 is preferable.  Thoughts @manovotn ?

manovotn:
> IMO, the former is the better option. Thoughts @manovotn ?

I agree and I basically arrived at the same conclusion after browsing Weld code some more.
This really seems to be an edge case for logging when the bean was removed but you don't want to blow up.
I'll create a Weld issue and send a PR tomorrow, thanks for your help @pferraro.

pferraro:
Great.  I had already prepared https://github.com/weld/core/pull/2862 just in case...

PR #16886, opened 2023-05-26
WFLY-18066 ByteBufferMarshalledValue generates duplicate buffers during a single marshalling operation

pferraro:
https://issues.redhat.com/browse/WFLY-18066

### DIFF
--- a/clustering/marshalling/spi/src/main/java/org/wildfly/clustering/marshalling/spi/ByteBufferMarshalledValue.java
+++ b/clustering/marshalling/spi/src/main/java/org/wildfly/clustering/marshalling/spi/ByteBufferMarshalledValue.java
@@ -45,12 +45,25 @@ public class ByteBufferMarshalledValue<T> implements MarshalledValue<T, ByteBuff
     private transient volatile T object;
     private transient volatile ByteBuffer buffer;
 
+    /**
+     * Constructs a marshalled value from the specified object and marshaller.
+     * @param object the wrapped object
+     * @param marshaller a marshaller suitable for marshalling the specified object
+     */
     public ByteBufferMarshalledValue(T object, ByteBufferMarshaller marshaller) {
         this.marshaller = marshaller;
         this.object = object;
     }
 
+    /**
+     * Constructs a marshalled value from the specified byte buffer.
+     * This constructor is only public to facilitate marshallers of this object (from other packages).
+     * The byte buffer parameter must not be read outside the context of this object.
+     * @param buffer a byte buffer
+     */
     public ByteBufferMarshalledValue(ByteBuffer buffer) {
+        // Normally, we would create a defensive ByteBuffer.asReadOnlyBuffer()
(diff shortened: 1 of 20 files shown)


clustering/marshalling/spi/src/main/java/org/wildfly/clustering/marshalling/spi/ByteBufferMarshalledValue.java
```diff
@@ -64,11 +65,18 @@ public synchronized boolean isEmpty() {
...
+            buffer = this.marshaller.write(this.object);
+            // N.B. Refrain from logging wrapped object
+            // If wrapped object contains an EJB proxy, toString() will trigger an EJB invocation!
+            LOGGER.debugf("Marshalled size of %s object = %d bytes", this.object.getClass().getCanonicalName(), buffer.limit() - buffer.arrayOffset());
+        }
+        return buffer;
+    }
+
+    public synchronized OptionalInt size() {
+        return (this.buffer != null) ? OptionalInt.of(this.buffer.limit() - this.buffer.arrayOffset()) : this.marshaller.size(this.object);
```
bstansberry, line 79: A comment explaining this algorithm would be good. It's not clear why passing a buffer into the constructor (so this.buffer != null) means the state of that buffer means this.buffer.arrayOffset() is safe to call (i.e. it's a heap buffer) or that this.buffer.limit() - this.buffer.arrayOffset() is the correct value when size() is called.

bstansberry (changes requested):

clustering/marshalling/spi/src/main/java/org/wildfly/clustering/marshalling/spi/ByteBufferMarshalledValue.java
pferraro: Valid point.  In practice, these are always array backed buffers - but we should always have hasArray() checks (and alternate handling) for completeness.

pferraro:
Added hold label while I finish the unit test.

clustering/marshalling/spi/src/main/java/org/wildfly/clustering/marshalling/spi/ByteBufferMarshalledValue.java
pferraro: Fixed.

pferraro: Oh - I see what you mean now.  Yes - I will add comments clarifying the lifecycle of this property, which is invalidated once the object is restored from the buffer.

pferraro: Done.

bstansberry (approved):
Thanks, @pferraro -- the comments are very helpful.

bstansberry:
Thanks @pferraro

PR #16894, opened 2023-05-31
[WFLY-12019] HostRemove should not remove services unless allow-resou…

bstansberry:
…rce-service-restart=true is set.

https://issues.redhat.com/browse/WFLY-12019


gaol:
@bstansberry looks good to me, I opened a PR https://github.com/bstansberry/wildfly/pull/63 to your branch with a test case.

gaol:
Sorry that the test failures are related, I opened another PR https://github.com/bstansberry/wildfly/pull/64 to your branch to fix the test failures. :)

gaol:
@bstansberry I would like to close this one, and use https://github.com/wildfly/wildfly/pull/16902 instead, which fixed the test failure, WDYT ?

gaol:
I am closing it :) 
replaced by #16902

PR #16903, opened 2023-06-05
WFLY-18069 Eliminate unnecessary buffer copies when writing an object with known size via ProtoStream

pferraro:
https://issues.redhat.com/browse/WFLY-18069

Uses non-buffering encoder during marshalling whenever possible, e.g. the backing OutputStream is a ByteArrayOutputStream.
Relocates reference tracking logic from the AnyMarshaller to the ProtoStreamWriter.writeAny(...) and ProtoStreamReader.readAny() method implementations.
Replaces thread local based context propagation during read/write/size operations with per-operation context instances.
Adds per-operation buffer size caching to reduce heap/cpu usage during a write operation.


darranl:
Triggered a CI run as I just merged three PRs - if this is reviewed (and still passes) it can likely go straight in as nothing to aggregate with.

rhusar (approved):
👍🏼 I couldn't find any issues here.

PR #16904, opened 2023-06-05
[WFLY-17563] Restore *module.xml necessary for manual installation of different jsf implementations

jasondlee:
https://issues.redhat.com/browse/WFLY-17563

Change javax.* to jakarta.*


pferraro:
@jasondlee Question - the distributable-web subsystem adds specific JSF-related modules to the deployment classpath based on this logic:
https://github.com/wildfly/wildfly/blob/main/clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/deployment/DistributableWebDeploymentProcessor.java#L93

If a user manually installs a different JSF implementation, is there a way to determine this from the DUP above?

bstansberry (approved):

PR #16906, opened 2023-06-05
WFLY-17853 WFLY-17978 Batch job fails to restart on server resume after server suspend

chengfang:
https://issues.redhat.com/browse/WFLY-17853
Batch job fails to restart on server resume after server suspend

https://issues.redhat.com/browse/WFLY-17978
Add a test using remote ejb client for restarting batch job after server suspend and resume


chengfang:
the 1 test failure (the new test) in Galleon Linux job is caused by the use of remote ejb in batch-jberet layer. I'll separate this test into its own test class and exclude it from layered test.

chengfang:
This PR does not refresh after I pushed more changes to the source branch. So close it before opening a new one.

chengfang:
superseded by https://github.com/wildfly/wildfly/pull/16917

PR #16907, opened 2023-06-06
WFLY-17967 Add MicroProfile Config layer dependency to MicroProfile LRA layers

xstefank:
https://issues.redhat.com/browse/WFLY-17967

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-coordinator/layer-spec.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-coordinator/layer-spec.xml
@@ -26,6 +26,7 @@
         <layer name="cdi"/>
         <layer name="jaxrs"/>
         <layer name="transactions"/>
+        <layer name="microprofile-config" optional="true"/>
     </dependencies>
 
     <feature spec="subsystem.microprofile-lra-coordinator"/>
--- a/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-participant/layer-spec.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-participant/layer-spec.xml
@@ -25,6 +25,7 @@
     <dependencies>
         <layer name="cdi"/>
         <layer name="jaxrs"/>
+        <layer name="microprofile-config" optional="true"/>
     </dependencies>
 
     <feature spec="subsystem.microprofile-lra-participant"/>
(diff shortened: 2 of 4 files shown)


xstefank:
@bstansberry, can you check please if this is everything that is needed please or do I need to add it also somewhere else?

darranl:
@xstefank FYI you are in the triagers group so should be able to assign reviewers yourself.

Also would suggest trying to include some others who have worked in these subsystems so Brian is not the default reviewer for all PRs.

xstefank:
@darranl thanks

galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-participant/layer-spec.xml
```diff
@@ -25,6 +25,7 @@
     <dependencies>
         <layer name="cdi"/>
         <layer name="jaxrs"/>
+        <layer name="microprofile-config"/>
```
bstansberry, line 28: @xstefank Can this layer function properly if the microprofile-config-smallrye subsystem is not part of the standalone.xml configuration and the user doesn't try to configure mp.lra.propagation.active?

If yes adding 'optional="true"' to this is a possibility as it lets the user exclude that layer and thus not have the subsystem in their config. If it won't work properly this definitely needs to be there.

Even if it won't work properly it's valid for you to decide that you always want it added. (You could always choose to make it optional later; otoh making it required later can be a compatibility problem.) I just want to make sure this PR is doing what you intend.

galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-lra-coordinator/layer-spec.xml
```diff
@@ -26,6 +26,7 @@
         <layer name="cdi"/>
         <layer name="jaxrs"/>
         <layer name="transactions"/>
+        <layer name="microprofile-config"/>
```
bstansberry, line 29: Same comment as the one I made for the participant layer.

galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/rts/lra-coordinator/main/module.xml
```diff
@@ -44,6 +44,7 @@
     <module name="jakarta.servlet.api"/>
     <module name="org.eclipse.microprofile.openapi.api"/>
     <module name="io.smallrye.fault-tolerance"/>
+    <module name="io.smallrye.config" optional="true"/>
```
bstansberry, line 47: This being optional implies that you were thinking in terms of the layer dependency being optional as well.

This being added at all implies that something was missing before, i.e. we have a bug. In that case adding some kind of regression test would be good.

bstansberry (commented):
Hi @xstefank See some questions below...

xstefank:
@bstansberry can you take a look again please?

bstansberry (approved):

bstansberry:
Thanks @xstefank

PR #16917, opened 2023-06-07
WFLY-17853 WFLY-17978 Batch job fails to restart on server resume after server suspend

chengfang:
https://issues.redhat.com/browse/WFLY-17853
Batch job fails to restart on server resume after server suspend

https://issues.redhat.com/browse/WFLY-17978
Add a test using remote ejb client for restarting batch job after server suspend and resume

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
@@ -536,8 +536,8 @@ public class JobOperatorService extends AbstractJobOperator implements WildFlyJo
                         }
                         try {
                             final long newId;
-                            // If the user is not null we need to restart the job with the user specified
-                            if (user == null) {
+                            // If the user is not null or $local, we need to restart the job with the user specified
+                            if (user == null || user.equals("$local")) {
                                 newId = restart(id, RESTART_PROPS);
                             } else {
                                 newId = privilegedRunAs(user, () -> restart(id, RESTART_PROPS));
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -2739,6 +2739,9 @@
                                     <excludes>
                                         <!-- Test requires agroal -->
                                         <exclude>org/jboss/as/test/integration/batch/repository/JdbcRepositoryTestCase.java</exclude>
+
+                                        <!-- Test requires remote ejb -->
+                                        <exclude>org/jboss/as/test/integration/batch/ejb/**/*</exclude>
                                     </excludes>
                                 </configuration>
                             </execution>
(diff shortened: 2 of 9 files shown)


batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
```diff
@@ -536,8 +536,8 @@ private void restartStoppedJobs() {
                         }
                         try {
                             final long newId;
-                            // If the user is not null we need to restart the job with the user specified
-                            if (user == null) {
+                            // If the user is not null or $local, we need to restart the job with the user specified
+                            if (user == null || user.equals("$local")) {
```
fjuma, line 540: @darranl This PR introduces a new test case that runs as a remote client invoking a remote EJB that starts a batch job execution. 

The test makes use of the `JBOSS_LOCAL_USER` mechanism (because it does not explicitly disable this mechanism and doesn't specify a username/password to use).

This means that the test makes use of the `$local` user for the remoting connection and starts the batch job using the `$local` user.

After the batch job has started, the test suspends the server and then resumes it. The batch job is then restarted.

When restarting a job, the `batch` subsystem will use `securityDomain.getAnonymousSecurityIdentity().createRunAsIdentity(user, false)` to obtain the security identity to restart the batch job as.

The `batch` subsystem makes use of the `ApplicationDomain` by default. This means the `ApplicationDomain` will be used in this call. Because the `$local` user won't be found in the default realm, restarting the batch job will fail. Note that this behaviour isn't specific to the `$local` user, this would also fail for any user that's not part of the default realm.

To prevent this, this PR adds a special case in `JobOperatorService` to prevent the `$local` user from being used when restarting a batch job.

IMO, I don't think we should add a special case for the `$local` user here. Instead, I think it would make more sense for the new test to make use of another user instead. To me, the test case doesn't really make sense with the `$local` user and it feels like this change is being made to get the test to pass.

Even with the fix for `$local`, the same issue would also occur if a user from a non-default realm was used.

Do you have any thoughts on this?

fjuma: @chengfang We discussed this PR on the Elytron team call today.

As mentioned on [WFLY-17853](https://issues.redhat.com/browse/WFLY-17853), we currently don't have a way to restore the $local user. Ideally, a token based approach would be used instead in the future ([WFLY-14409](https://issues.redhat.com/browse/WFLY-14409) is the main JIRA issue that tracks this work).

The `$local` identity is a hardcoded value that's specified in the configuration and anyone can change it.

We don't think the `$local` user should be hardcoded in the `JobOperatorService`.

Instead, if we want to add a new test case like this, the test in this PR should be updated to make use of a different user.

fjuma (changes requested):
As mentioned in the comments, the new test being added should be updated to make use of a different user.

Thanks!

chengfang:
@fjuma thanks to you the security team for evaluating this. I think it make sense to table this issue till the token-based authentication is in place. The new tests are mainly for the purpose of verifying this issue and will also be deferred.

PR #16926, opened 2023-06-12
WFLY-18101 Migrate clustering subsystem tests to use AbstractSubsystemSchemaTest from wildfly-core

pferraro:
https://issues.redhat.com/browse/WFLY-18101


rhusar (approved):
All of these look good.

(n.b. I think I will eventually rename next time I touch mod_cluster subsystem xml files from the generic `subystem_x_y.xml` to match the subsystem name).

PR #16928, opened 2023-06-12
WFLY-18103 Migrate MicroProfile fault-tolerance subsystem test to use AbstractSubsystemSchemaTest from wildfly-core

pferraro:
https://issues.redhat.com/browse/WFLY-18103


rhusar (approved):
Looks good - thanks.

PR #16948, opened 2023-06-15
[WFLY-18115] Opentelemetry sampler-type cannot be configured correctly

jasondlee:
https://issues.redhat.com/browse/WFLY-18115

Add code to translate between WF-friendly and otel-specified parameter values for sampler-type


bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #16958, opened 2023-06-16
[WFLY-18157] Add Jakarta Faces API dep back to BOM

jasondlee:
https://issues.redhat.com/browse/WFLY-18157

Add dependency declaration back to BOM

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -3256,6 +3256,20 @@
                 </exclusions>
             </dependency>
 
+            <!-- This dependency is not used by the app server itself, but is defined here to allow
+                 user applications to pull in the dependency version as part of the BOM -->
+            <dependency>
+                <groupId>jakarta.faces</groupId>
+                <artifactId>jakarta.faces-api</artifactId>
+                <version>${version.jakarta.faces.jakarta-faces-api}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+
             <dependency>
                 <groupId>org.glassfish.expressly</groupId>
                 <artifactId>expressly</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -429,6 +429,7 @@
         <version.jakarta.ejb.jakarta-ejb-api>4.0.1</version.jakarta.ejb.jakarta-ejb-api>
         <version.jakarta.enterprise.concurrent.jakarta-enterprise.concurrent-api>3.0.2</version.jakarta.enterprise.concurrent.jakarta-enterprise.concurrent-api>
         <version.jakarta.enterprise>4.0.1</version.jakarta.enterprise>
+        <version.jakarta.faces.jakarta-faces-api>4.0.0</version.jakarta.faces.jakarta-faces-api>
```
bstansberry, line 432: @jasondlee I believe this should be 4.0.1. That's what it was before, and the diff between the two tells me the Mojarra code that produced them is different, with 4.0.1 being closer to the Mojarra was are using:

https://github.com/jakartaee/faces/compare/4.0.0-RELEASE...4.0.1-RELEASE#diff-f35103f068e191d5d8933637b2d4217d8f90a78728ccab1bada951a8d1c83590L111-R111

bstansberry (changes requested):

boms/common-ee/pom.xml
```diff
@@ -3256,6 +3256,18 @@
                 </exclusions>
             </dependency>
 
+            <dependency>
```
jbliznak, line 3259: Since this is uncommon case, what about a comment that we include it even though the same classes are in org.glassfish:jakarta.faces to allow propagation downstream (eg. to BOMs)?

jbliznak (approved):
I will leave it for Brian to confirm the dependency is added to the correct place, but I can confirm that this change makes WF BOMs compile again.

bstansberry (approved):

bstansberry:
Thanks @jasondlee and @jbliznak

PR #16965, opened 2023-06-22
[WFLY-18170] Use a default value for the jakarta.faces.FACELETS_BUFFE…

jamezp:
…R_SIZE parameter was not set.

https://issues.redhat.com/browse/WFLY-18170

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
@@ -3,6 +3,7 @@ package org.jboss.as.jsf.deployment;
 import java.util.ArrayList;
 import java.util.List;
 
+import jakarta.faces.application.ViewHandler;
 import org.jboss.as.jsf.logging.JSFLogger;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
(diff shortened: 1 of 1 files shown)


jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
```diff
@@ -61,6 +86,11 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                 JSFLogger.ROOT_LOGGER.lazyBeanValidationEnabled();
             }
         }
+        // Set a default buffer size as 1024 is too small
+        final JBossWebMetaData webMetaData = metaData.getMergedJBossWebMetaData();
+        if (!hasContextParam(webMetaData, "faces.FACELETS_BUFFER_SIZE")) {
```
bstansberry, line 91: The param is "jakarta.faces.FACELETS_BUFFER_SIZE". Per the spec we should also check "facelets.BUFFER_SIZE" for compatibility.

jamezp: Looks like a type on my side, but it also says:

> The runtime must also consider the facelets.BUFFER_SIZE param name as an alias to this param name for backwards compatibility with existing facelets tag libraries.

bstansberry:
Thanks @jamezp

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
```diff
@@ -23,6 +24,30 @@ public class JSFMetadataProcessor implements DeploymentUnitProcessor {
...
     private static final String LAZY_BEAN_VALIDATION_PARAM = "com.sun.faces.enableLazyBeanValidation";
 
+
+    private static final int defaultBufferSize;
+
+    // This is copied from org.wildfly.extension.undertow.ByteBufferPoolDefinition to come up with a decent default for
+    // the jakarta.faces.FACELETS_BUFFER_SIZE property. We calculate this because the default is 1024, which is very
+    // small, https://jakarta.ee/specifications/faces/4.0/jakarta-faces-4.0.html#a6088. Per the spec we could use -1
+    // as the default, but Mojarra does not currently support that. Once https://github.com/eclipse-ee4j/mojarra/issues/5262
+    // is resolved, and we can upgrade, we should be able to default to -1.
```
fjuma, line 34: @jamezp Should we create a WFLY issue to track defaulting to -1 once the fix for https://github.com/eclipse-ee4j/mojarra/issues/5262 is available in a release?

jamezp: @fjuma That's not a bad idea, https://issues.redhat.com/browse/WFLY-18174.

fjuma: Thanks @jamezp!

PR #16966, opened 2023-06-22
[WFLY-18172] Upgrade ASM to 9.5

ropalka:
https://issues.redhat.com/browse/WFLY-18172

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -544,7 +544,7 @@
         <version.org.keycloak.keycloak-saml-wildfly-subsystem>18.0.2</version.org.keycloak.keycloak-saml-wildfly-subsystem>
         <version.org.kohsuke.metainf-services>1.8</version.org.kohsuke.metainf-services>
         <version.org.opensaml.opensaml>4.2.0</version.org.opensaml.opensaml>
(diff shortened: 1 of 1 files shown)


ropalka:
@jimma is the ASM upgrade compatible with webservice modules?
@scottmarlow is the ASM upgrade compatible with hibernate modules?
@kabir is the ASM upgrade compatible with smallrye modules?

kabir (approved):
SR uses 9.5 already, e.g  https://github.com/smallrye/smallrye-config/blob/main/pom.xml#L39

scottmarlow:
> @scottmarlow is the ASM upgrade compatible with hibernate modules?

Yes as [Hibernate ORM](https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml#L26) + Hibernate Search do not depend on ASM (Hibernate ORM 5.3 did but Hibernate ORM 6.0 only depends on bytebuddy which private packages its own copy of ASM).

scottmarlow (approved):

bstansberry:
Thanks @ropalka

PR #16971, opened 2023-06-26
[WFLY-17841] Adding back org.jboss.vfs as implicit deployment dependency

yersan:
Jira issue:https://issues.redhat.com/browse/WFLY-17841


bstansberry:
Thanks @yersan

PR #17005, opened 2023-07-06
[WFLY-18226] Use UTF-8 for reading log files.

jamezp:
https://issues.redhat.com/browse/WFLY-18226


jamezp:
This is passing on Linux for me, even overriding the encoding, but failing on Windows locally. I guess we'll see how CI does. There are other ways around this if needed.

jamezp:
Looks like CI is failing too. I'll re-think this in the morning.

PR #17006, opened 2023-07-07
WFLY-18227 WFLY-18228 WFLY-18229 Upgrade to Hibernate Search 6.2.0.Final, Elasticsearch client 8.8.8, Avro 1.11.2

yrodiere:
https://issues.redhat.com/browse/WFLY-18227
https://issues.redhat.com/browse/WFLY-18228
https://issues.redhat.com/browse/WFLY-18229

Unfortunately I wasn't able to have WildFly force Hibernate Search to use its "Lucene 9 ID schema" for indexes; Hibernate Search's configuration is too dynamic to allow that reliably (see https://hibernate.atlassian.net/browse/HSEARCH-4897). We will have to handle that later.

It's still worth upgrading to 6.2 in WildFly, though. Be it just for the new [features](https://hibernate.org/search/releases/6.2/#whats-new).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -498,7 +498,7 @@
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.0</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
         <version.org.hibernate.commons.annotations>6.0.6.Final</version.org.hibernate.commons.annotations>
         <version.org.hibernate>6.2.6.Final</version.org.hibernate>
-        <version.org.hibernate.search>6.2.0.CR1</version.org.hibernate.search>
+        <version.org.hibernate.search>6.2.0.Final</version.org.hibernate.search>
         <version.org.hibernate.validator>8.0.0.Final</version.org.hibernate.validator>
         <version.org.hornetq>2.4.9.Final</version.org.hornetq>
         <version.org.infinispan>14.0.11.Final</version.org.infinispan>
--- a/pom.xml
+++ b/pom.xml
@@ -453,7 +453,7 @@
         <version.net.shibboleth.utilities.java-support>8.0.0</version.net.shibboleth.utilities.java-support>
         <version.org.apache.activemq.artemis>2.26.0</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>1.0.2</version.org.apache.activemq.artemis.native>
-        <version.org.apache.avro>1.11.0</version.org.apache.avro>
+        <version.org.apache.avro>1.11.2</version.org.apache.avro>
         <version.org.apache.cxf>4.0.0</version.org.apache.cxf>
         <version.org.apache.cxf.xjcplugins>4.0.0</version.org.apache.cxf.xjcplugins>
         <version.org.apache.httpcomponents.httpasyncclient>4.1.5</version.org.apache.httpcomponents.httpasyncclient>
--- a/pom.xml
+++ b/pom.xml
@@ -489,7 +489,7 @@
         <version.org.eclipse.microprofile.telemetry>1.0</version.org.eclipse.microprofile.telemetry>
         <version.org.eclipse.persistence.eclipselink>4.0.0</version.org.eclipse.persistence.eclipselink>
         <version.org.eclipse.yasson>3.0.2</version.org.eclipse.yasson>
-        <version.org.elasticsearch.client.rest-client>7.16.3</version.org.elasticsearch.client.rest-client>
+        <version.org.elasticsearch.client.rest-client>8.8.2</version.org.elasticsearch.client.rest-client>
         <version.org.glassfish.expressly>5.0.0</version.org.glassfish.expressly>
         <version.org.glassfish.jakarta.enterprise.concurrent>3.0.0</version.org.glassfish.jakarta.enterprise.concurrent>
         <version.org.glassfish.jaxb>4.0.1</version.org.glassfish.jaxb>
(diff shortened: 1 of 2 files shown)


scottmarlow (approved):

scottmarlow:
Copying from [WFLY-18227](https://issues.redhat.com/browse/WFLY-18227):
```
Changes since 6.2.0.CR1: https://hibernate.atlassian.net/issues/?jql=project=10061+AND+fixVersion=32168
See also https://hibernate.org/search/releases/6.2/
Migration guide from 6.1: https://docs.jboss.org/hibernate/search/6.2/migration/html_single/
```

@yrodiere I see that one of the changes is upgrading to use `Avro 1.11.2` but WildFly currently contains `avro-1.11.0.jar` (as module name=`org.apache.avro`) but I am not sure that is used by the Hibernate Search module in WildFly?

scottmarlow:
Hibernate Search in WildFly does have a module dependency on lasticsearch-rest-client-7.16.3.jar, should WildFly upgrade to `Elasticsearch 8.8.2` as part of this pull request?

yrodiere:
Ouch. You are right. Fortunately those are private modules. I'll create tickets and add that to the PR.

yrodiere:
Done. Thanks for catching that @scottmarlow!

scottmarlow (approved):

scottmarlow:
> Done. Thanks for catching that @scottmarlow!

Thank you for the quick update!

scottmarlow (changes requested):
@yrodiere which Hibernate Search module needs access to the Avro classes?  

Can you confirm that applications using Hibernate Search do not need direct access to the Avro classes as far as you understand?

yrodiere:
> @yrodiere which Hibernate Search module needs access to the Avro classes?
> 
> Can you confirm that applications using Hibernate Search do not need direct access to the Avro classes as far as you understand?

They do not. Nothing changed in this area.

scottmarlow:
> > @yrodiere which Hibernate Search module needs access to the Avro classes?

wildfly/preview/galleon-local/src/main/resources/modules/system/layers/base/org/hibernate/search/mapper/orm/coordination/outboxpolling/main/module.xml seems to reference Avro classes in the org.hibernate.search.mapper.orm.coordination.outboxpolling module.  

Note that file wildfly/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java also mentions Avro with a comment:
```
// Used by outboxpolling TODO this doesn't seem right
                    "org.apache.avro"
```

The HibernateSearchProcessor does reference the `org.hibernate.search.mapper.orm.coordination.outboxpolling` so it does seem like the Avro classes are indeed referenced.

> > Can you confirm that applications using Hibernate Search do not need direct access to the Avro classes as far as you understand?
> 
> They do not. Nothing changed in this area.

Got it, thanks!

yrodiere:
> ```
> // Used by outboxpolling TODO this doesn't seem right
>                     "org.apache.avro"
> ```
> 
> 
>     
>   
> 
> The HibernateSearchProcessor does reference the `org.hibernate.search.mapper.orm.coordination.outboxpolling` so it does seem like the Avro classes are indeed referenced.

I don't know where this comment comes from, but this dependency is completely expected. This module requires Avro for it's implementation. The API does not require Avro so it's not exposed to users.



> It looks like Hibernate Search has an optional dependency on the `org.hibernate.search.mapper.orm.coordination.outboxpolling`

Correct. It is not exposed to users however (this is what you asked previously).



> which I think only exists in the `WildFly Preview` build but is not in the `WildFly` build.

Correct. This technology is incubating.

> Should the Avro module also only exist in the `WildFly Preview` build since those are only used by the `org.hibernate.search.mapper.orm.coordination.outboxpolling` module?

Fine by me. Last I checked the Avro module was used by something else (I forgot what exactly), but if it's no longer the case... It can move out to preview as far as I'm concerned. It's internal anyway, isn't it?

I just hope this won't make it harder to move the outbox-polling module out of preview (which should happen eventually) :)

scottmarlow:
> I don't know where this comment comes from, but this dependency is completely expected.

https://github.com/wildfly/wildfly/blob/main/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java contains that comment.  

>> Should the Avro module also only exist in the WildFly Preview build since those are only used by the org.hibernate.search.mapper.orm.coordination.outboxpolling module?

> Fine by me. Last I checked the Avro module was used by something else (I forgot what exactly), but if it's no longer the case... It can move out to preview as far as I'm concerned. It's internal anyway, isn't it?

I don't see any other references to the Avro module currently.

> I just hope this won't make it harder to move the outbox-polling module out of preview (which should happen eventually) :)

I defer to @bstansberry on whether Avro should also move to preview only.

scottmarlow:
org.wildfly.test.integration.microprofile.faulttolerance.micrometer.FaultToleranceMicrometerIntegrationTestCase + 
org.jboss.as.test.manualmode.messaging.ha.ReplicatedFailoverTestCase + org.jboss.as.test.clustering.cluster.web.ConcurrentCoarseWebFailoverTestCase test failures don't look related to this upgrade.

scottmarlow:
> @scottmarlow @yrodiere I filed https://issues.redhat.com/browse/WFLY-18245 to track moving the avro module to WildFly Preview. I don't see that as super-high priority though, and I think it should be handled separately from this PR as I'd like to get this merged today or tomorrow.
> 
> FYI, the related https://issues.redhat.com/browse/JBEAP-25175 is more of a priority though.

Thanks Brian

scottmarlow (approved):

bstansberry:
Thanks @yrodiere and @scottmarlow

PR #17007, opened 2023-07-07
WFLY-18179 & WFLY-18230 Avoid redundant no-op runtime steps in undertow and security subsystems

pferraro:
https://issues.redhat.com/browse/WFLY-18179
https://issues.redhat.com/browse/WFLY-18230

... and assorted changes required by https://github.com/wildfly/wildfly-core/pull/5563

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
@@ -21,6 +21,7 @@
 package org.jboss.as.jaxrs;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.OperationStepHandler;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.SimpleResourceDefinition;
 
@@ -66,9 +67,10 @@ public class JaxrsSubsystemDefinition extends SimpleResourceDefinition {
 
     @Override
     public void registerAttributes(ManagementResourceRegistration registration) {
-       for (AttributeDefinition definition : JaxrsAttribute.ATTRIBUTES) {
-          registration.registerReadWriteAttribute(definition, null, new JaxrsParamHandler(definition));
-       }
+        OperationStepHandler handler = new JaxrsParamHandler(JaxrsAttribute.ATTRIBUTES);
+        for (AttributeDefinition definition : JaxrsAttribute.ATTRIBUTES) {
+            registration.registerReadWriteAttribute(definition, null, handler);
+        }
     }
 
     @Override
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
@@ -20,6 +20,7 @@
  */
 package org.jboss.as.jaxrs;
 
+import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.SimpleResourceDefinition;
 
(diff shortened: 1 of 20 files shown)


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
```diff
@@ -63,6 +64,13 @@ public class JaxrsSubsystemDefinition extends SimpleResourceDefinition {
                  .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE));
     }
 
+    @Override
+    public void registerAttributes(ManagementResourceRegistration registration) {
+       for (AttributeDefinition definition : JaxrsAttribute.ATTRIBUTES) {
+          registration.registerReadWriteAttribute(definition, null, new JaxrsParamHandler(definition));
```
bstansberry, line 70: FYI re https://issues.redhat.com/browse/WFLY-18257. I'm not requesting a change; just FYI.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterDefinitions.java
```diff
@@ -45,8 +45,8 @@ public class FilterDefinitions extends PersistentResourceDefinition {
 
     public FilterDefinitions() {
         super(PATH_ELEMENT, UndertowExtension.getResolver(PATH_ELEMENT.getValue()),
-                new AbstractAddStepHandler(),
-                ReloadRequiredRemoveStepHandler.INSTANCE
+                new ModelOnlyAddStepHandler(),
+                ModelOnlyRemoveStepHandler.INSTANCE
```
bstansberry, line 49: This remove handler change seems out of scope and perhaps changes the reported read-operation-description behavior. There were some recent bug-fixing changes in this area and I'm reluctant to change this as part of this work.

undertow/src/main/java/org/wildfly/extension/undertow/handlers/HandlerDefinitions.java
```diff
@@ -43,8 +43,8 @@ public class HandlerDefinitions extends PersistentResourceDefinition {
 
     public HandlerDefinitions() {
         super(new SimpleResourceDefinition.Parameters(PATH_ELEMENT, UndertowExtension.getResolver(PATH_ELEMENT.getValue()))
-                .setAddHandler(new AbstractAddStepHandler())
-                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
+                .setAddHandler(new ModelOnlyAddStepHandler())
+                .setRemoveHandler(ModelOnlyRemoveStepHandler.INSTANCE)
```
bstansberry, line 47: Same comment as for the remove handler in FilterDefinitions.

bstansberry (commented):
@pferraro This all looks fine, except I'm concerned about the 2 undertow remove handler changes.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterDefinitions.java
pferraro: This change is the essence of this PR. This resource has no runtime - thus the existing handlers create a completely unnecessary runtime step.
If this resource contains any child resources (i.e. actual filter resource with runtime handling), removing this resource will recursively remove its children and put the server into reload required.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemDefinition.java
pferraro: That's worthwhile change to make now (though, eventually, these attribute definition parameters will be made obsolete by https://github.com/wildfly/wildfly-core/pull/5563).

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterDefinitions.java
pferraro: > This remove handler change seems out of scope and perhaps changes the reported read-operation-description behavior.

Can you elaborate?  I don't see any recent changes to the ReadOperationDescriptionHandler.  Yes, this would change the reported "restart-required" value for the remove operation - but that's the whole point.

bstansberry: Ok; I see now both of these resources are direct children of the subsystem root and truly have no runtime behavior, i.e. no parent whose behavior is influenced by these.

I believe the thing I was talking about related to FilterRefDefinition, which is a different case.

Thanks, and sorry for the noise.

bstansberry (approved):
@pferraro This all looks fine, except I'm concerned about the 2 undertow remove handler changes.

bstansberry:
Thanks @pferraro

PR #17013, opened 2023-07-12
[WFLY-18222] Optimise Kafka layer rule

kabir:
I'm trying to get away with no WFLY issue since this is WIP/unsupported code which has no effect on WIldFly as-is


bstansberry:
@kabir Please add WFLY-18222 to the PR title so it can be associated with that in the git log via the merge commit.

kabir:
@bstansberry done, thanks

bstansberry:
Thanks @kabir

PR #17019, opened 2023-07-13
WFLY-18249: add new attribute to use with datasource capability reference

michpetrov:
Issue: [WFLY-18249](https://issues.redhat.com/browse/WFLY-18249)

Btw. I noticed the capabilities in this subsystem are not documented in the wildfly-capabilities repo (looks like there's a PR hanging - wildfly/wildfly-capabilities#32)

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
+++ b/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
@@ -58,6 +58,13 @@ import org.wildfly.transaction.client.provider.remoting.RemotingTransactionServi
  */
 public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDefinition {
 
+    /**
+     * Represents the data source capability
+     *
+     * @see <a href="https://github.com/wildfly/wildfly-capabilities/blob/main/org/wildfly/data-source/capability.adoc">Capability documentation</a>
+     */
+    static final String DATA_SOURCE_CAPABILITY_NAME = "org.wildfly.data-source";
+
     static final RuntimeCapability<Void> TRANSACTION_CAPABILITY = RuntimeCapability.Builder.of("org.wildfly.transactions", Void.class)
             .build();
 
@@ -222,13 +229,20 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
             .setDefaultValue(ModelNode.FALSE)
             .addAlternatives(CommonAttributes.USE_JOURNAL_STORE)
             .addAlternatives(CommonAttributes.USE_HORNETQ_STORE)
-            .setRequires(CommonAttributes.JDBC_STORE_DATASOURCE)
             .setFlags(AttributeAccess.Flag.RESTART_JVM)
             .setAllowExpression(false).build();
+    public static final SimpleAttributeDefinition JDBC_STORE_DATASOURCE_NAME = new SimpleAttributeDefinitionBuilder(CommonAttributes.JDBC_STORE_DATASOURCE_NAME, ModelType.STRING, true)
+            .setFlags(AttributeAccess.Flag.RESTART_JVM)
+            .setXmlName(Attribute.DATASOURCE_NAME.getLocalName())
+            .addAlternatives(CommonAttributes.JDBC_STORE_DATASOURCE)
+            .setCapabilityReference(DATA_SOURCE_CAPABILITY_NAME, TRANSACTION_CAPABILITY)
+            .build();
     public static final SimpleAttributeDefinition JDBC_STORE_DATASOURCE = new SimpleAttributeDefinitionBuilder(CommonAttributes.JDBC_STORE_DATASOURCE, ModelType.STRING, true)
             .setFlags(AttributeAccess.Flag.RESTART_JVM)
             .setXmlName(Attribute.DATASOURCE_JNDI_NAME.getLocalName())
-            .setAllowExpression(true).build();
+            .addAlternatives(CommonAttributes.JDBC_STORE_DATASOURCE_NAME)
+            .setAllowExpression(true)
+            .setDeprecated(ModelVersion.create(6)).build();
     public static final SimpleAttributeDefinition JDBC_ACTION_STORE_TABLE_PREFIX =
             new SimpleAttributeDefinitionBuilder(CommonAttributes.JDBC_ACTION_STORE_TABLE_PREFIX, ModelType.STRING, true)
             .setFlags(AttributeAccess.Flag.RESTART_JVM)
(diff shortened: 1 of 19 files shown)


tadamski (approved):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -245,6 +252,7 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
     public static final SimpleAttributeDefinition JDBC_STORE_DATASOURCE = new SimpleAttributeDefinitionBuilder(CommonAttributes.JDBC_STORE_DATASOURCE, ModelType.STRING, true)
             .setFlags(AttributeAccess.Flag.RESTART_JVM)
             .setXmlName(Attribute.DATASOURCE_JNDI_NAME.getLocalName())
+            .setCapabilityReference(DATA_SOURCE_CAPABILITY_NAME, TRANSACTION_CAPABILITY)
```
pferraro, line 255: This does not look correct to me.
According to:
https://github.com/wildfly/wildfly/blob/main/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemAdd.java#L347
... and ...
https://github.com/wildfly/wildfly/blob/main/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemAdd.java#L371
... the JDBC_STORE_DATASOURCE attribute expects a JNDI name, not a reference to a DataSource resource.

To implement this correctly, you will need to add a new attribute (as an alternative to the existing one) that accepts a data-source reference, and deprecate the existing attribute.  The resource add operation handler would then resolve the DataSource dependency via its capability reference, if defined.

pferraro (changes requested):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
michpetrov: @pferraro I see, I would've expected some validation in that case. 

That said the "capability reference" doesn't change the datatype, we're still passing a string, in this case the basic name of the datasource. Unfortunately I'm not seeing a way to obtain the JNDI name from the capability reference so I might have to scratch this.

pferraro: It is not possible to configure the Arjuna ObjectStore with a DataSource instance?

michpetrov: I'm not familiar enough with it to say, but the current implementation uses some string-based config - https://github.com/wildfly/wildfly/blob/9d48cbde13d7da3db315161ccb5aa1cfd2308b50/transactions/src/main/java/org/jboss/as/txn/service/ArjunaObjectStoreEnvironmentService.java#L107-L109

I will take this to Zulip, I can't get this to work even with the proper configuration

darranl:
What happens if this attribute is set to something else?  A change such as this risks preventing boot from succeeding using an existing configuration, if that existing configuration would also fail to boot that may not be an issue but if this now triggers a failed boot this may be unexpected by users.

michpetrov:
@pferraro I have reworked the PR to add a separate attribute.

Btw, `commit-markable-resources` also uses JDNI names but the lookup for these is performed outside the subsystem so we would need a change in Narayana.

darranl (changes requested):
If we are doing a model / schema bump please separate the bump into it's own commit and the changes in their own commit - this makes it clearer in standard tooling what has actually changed.

michpetrov:
@darranl ok, but if I do the version bump separately the first commit will not build. Is that ok? I thought that was a no-no.

pferraro:
> ok, but if I do the version bump separately the first commit will not build. Is that ok? I thought that was a no-no.

@michpetrov The management model bump and/or schema version bump would precede the commit that actually changes the model/schema.

michpetrov:
@pferraro I don't think I follow, is the version bump just this line specifically? https://github.com/wildfly/wildfly/blob/07ee34510e722a4894da96db3d739d2ff1f304fa/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionExtension.java#L62

I would assume that the bump also needs to include this
https://github.com/wildfly/wildfly/blob/07ee34510e722a4894da96db3d739d2ff1f304fa/transactions/src/main/java/org/jboss/as/txn/subsystem/Namespace.java#L48-L54

and this
https://github.com/wildfly/wildfly/blob/07ee34510e722a4894da96db3d739d2ff1f304fa/transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionExtension.java#L139

But that would mean I have to include the new parser as well (and now we're cascading into including most of the commit); I tried checking how it was done previously but the last version bump was also done in a commit with significant code changes.

darranl:
> But that would mean I have to include the new parser as well (and now we're cascading into including most of the commit); I tried checking how it was done previously but the last version bump was also done in a commit with significant code changes.

Yes it is quite likely the bump commit contains the majority of changes, then we can focus on the second commit to see the changes subsequently made to the model / integration etc..

michpetrov:
If you want me to separate the changes to the model, i.e. the subsystem definition, from the version bump then the only way I see how is instead of adding a new parser I'd just copy the old one and change it after in the second commit. Otherwise the only thing I wouldn't have to include in the first commit are the tests, and only because the changes don't make them fail.

michpetrov:
@darranl would this work? https://github.com/michpetrov/wildfly/compare/main...michpetrov:wildfly:wfly-18249-split

mmusgrov (approved):
I took a look and these changes are a good addition - it's nice to see less reliance on JNDI names for the config (we did something similar with the quarkus jta extension).

darranl:
Hi @michpetrov unfortunately no that split still does not work, I can not see in the GitHub UI the changes made to the schema after it was bumped.

michpetrov:
@darranl well, there's no change to the schema _after_. Do you want me to bump the version without changing anything? Because that seems counter-intuitive.

pferraro (changes requested):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -222,13 +229,21 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
...
-            .setRequires(CommonAttributes.JDBC_STORE_DATASOURCE)
+            //.setRequires(CommonAttributes.JDBC_STORE_DATASOURCE_NAME)
             .setFlags(AttributeAccess.Flag.RESTART_JVM)
             .setAllowExpression(false).build();
+    public static final SimpleAttributeDefinition JDBC_STORE_DATASOURCE_NAME = new SimpleAttributeDefinitionBuilder(CommonAttributes.JDBC_STORE_DATASOURCE_NAME, ModelType.STRING, true)
+            .setFlags(AttributeAccess.Flag.RESTART_JVM)
+            .setXmlName(Attribute.DATASOURCE_NAME.getLocalName())
+            .addAlternatives(CommonAttributes.JDBC_STORE_DATASOURCE)
+            .setCapabilityReference(DATA_SOURCE_CAPABILITY_NAME, TRANSACTION_CAPABILITY)
+            .setAllowExpression(true).build();
```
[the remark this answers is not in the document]
michpetrov: I've changed it but what's the problem here? Whether the name is a String or an expression that resolves to a String shouldn't matter, should it?. (I did find several attributes that allow expressions and have a capability reference though those are in other subsystems)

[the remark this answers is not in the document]
pferraro: Capability reference validation happens during model validation.  Expressions are resolved at runtime - and can evaluate to different values per host.

darranl:
@michpetrov yes bump the version first without making any changes so the files are just a copy of the last version with things like the version updated so they are valid but nothing else.

Then in the second commit make the actual changes to the schema so we can see what is different from the new baseline to the feature being added.

michpetrov:
@darranl rebased

darranl:
@tadamski / @michpetrov I am wondering if this can now be adapted to the stability levels and we can get this swept into WildFly 32 following the approach Brian describes here:

https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/message/NKQVG2PDU7C5HU2BARNODHIXI2TOEXNJ/

darranl (changes requested):
I think at this stage try and work with @tadamski regarding if we can update to a lower stability level and come in via the WildFly feature development process.

PR #17025, opened 2023-07-16
[WFLY-18261] Add a SECURITY.md

bstansberry:
https://issues.redhat.com/browse/WFLY-18261

This is a copy of what's on wildfly.org. I thought about doing just simple words and then a link, or simple words, a sentence saying to use the list, and then a link for more info. But this is short enough it seems best to just directly state the information vs losing people who don't want to follow further links after they already came to a logical place for info.


fjuma (approved):
Thanks @bstansberry!

PR #17028, opened 2023-07-17
WFLY-18264 Convert TimerAttributeDefinition to ObjectListAttributeDefinition

pferraro:
https://issues.redhat.com/browse/WFLY-18264


ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/AbstractEJBComponentResourceDefinition.java
```diff
@@ -157,6 +165,15 @@ public abstract class AbstractEJBComponentResourceDefinition extends SimpleResou
...
 
+    public static final ObjectTypeAttributeDefinition TIMER = new ObjectTypeAttributeDefinition.Builder("timer",
+            TIME_REMAINING, NEXT_TIMEOUT, CALENDAR_TIMER, PERSISTENT, INFO, new ObjectTypeAttributeDefinition.Builder(SCHEDULE.getName(), SCHEDULE.getValueTypes()).setSuffix("schedule").build())
+            .setStorageRuntime()
+            .build();
+
+    public static final ObjectListAttributeDefinition TIMERS = new ObjectListAttributeDefinition.Builder("timers", TIMER)
+            .setStorageRuntime()
+            .build();
+
```
chengfang, line 176: the above 2 fields are only used locally, can be package private.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/TimerResourceDefinition.java
```diff
@@ -64,19 +63,19 @@ public class TimerResourceDefinition<T extends EJBComponent> extends SimpleResou
     private static final ResourceDescriptionResolver RESOURCE_DESCRIPTION_RESOLVER = EJB3Extension
             .getResourceDescriptionResolver(EJB3SubsystemModel.TIMER);
     // attributes, copy of TimerAttributeDefinition
```
chengfang, line 65: Avoid reference to the deleted class.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/AbstractEJBComponentRuntimeHandler.java
```diff
@@ -321,4 +348,136 @@ protected String resolveRuntimeName(final OperationContext context, final PathEl
...
+    private static void addScheduleDetailDate(ModelNode schedNode, Date value, String detailName) {
+        final ModelNode node = schedNode.get(detailName);
+        if (value != null) {
+            node.set(value.getTime());
+        }
+    }
+
+    private static void logTimerFailure(final String componentName, final EJBException e) {
+        ROOT_LOGGER.failToReadTimerInformation(componentName);
+    }
```
chengfang, line 482: Maybe we can include the exception message in the log, if the stack trace is too long for a warning log.

pferraro: This method was moved here from the TimerAttributeDefinition class - but I agree with you - I see no reason to omit the exception.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/TimerResourceDefinition.java
pferraro: Good point

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/deployment/AbstractEJBComponentResourceDefinition.java
pferraro: Indeed.

pferraro:
@chengfang Thanks for the review.  I've update the commit with your suggested changes.

chengfang (approved):

PR #17038, opened 2023-07-21
WFLY-18279 Update HostExcludesTestCase configuration to work with WF30

pferraro:
https://issues.redhat.com/browse/WFLY-18279

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
@@ -36,7 +36,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -90,7 +89,7 @@ public class HostExcludesTestCase extends BuildConfigurationTestBase {
     private final boolean isFullDistribution = AssumeTestGroupUtil.isFullDistribution();
     private final boolean isPreviewGalleonPack = AssumeTestGroupUtil.isWildFlyPreview();
 
-    private static final String MAJOR = "29.";
+    private static final String MAJOR = "30.";
 
     /**
      * Maintains the list of expected extensions for each host-exclude name for previous releases.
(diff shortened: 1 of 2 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
```diff
@@ -210,7 +210,8 @@ private enum ExtensionConf {
                 "org.wildfly.extension.microprofile.lra-participant",
                 "org.wildfly.extension.microprofile.telemetry"
         ), true),
-        CURRENT(MAJOR, WILDFLY_28_0, getCurrentAddedExtensions(), getCurrentRemovedExtensions(), true);
+        WILDFLY_29_0("WildFly29.0", WILDFLY_28_0, true),
+        CURRENT(MAJOR, WILDFLY_29_0, getCurrentAddedExtensions(), getCurrentRemovedExtensions(), true);
```
yersan, line 214: Just for the record, it is good to have it adjusted now (even if the product version is still 29) but it's not strictly necessary to do it on each new release if we have not added/removed extensions on the current release.

The test won't fail if we have moved the product version to the next one (e.g. 30) but we have not added/removed extensions on the current release (e.g. 29).

yersan (approved):
LGTM

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
pferraro: Without this change, it seems that all full-integration tests on wildfly-core are failing, e.g. https://ci.wildfly.org/viewLog.html?buildId=381306&buildTypeId=WildFlyCore_PullRequest_WildFlyCoreFullIntegrationLinuxJdk11
Was I mistaken?

pferraro: > The test won't fail if we have moved the product version to the next one (e.g. 30) but we have not added/removed extensions on the current release (e.g. 29).

Ah - I see.  So, were the product version changed (I assume, via an update to org.wildfly.core:wildfly-version + a wildfly-core upgrade?), this PR would not have been strictly necessary.  Correct?

bstansberry:
Thanks, @pferraro. Can I punish you for doing this by asking you to update the WF release process doc to include this as an aspect of the "bump the version" release task? This is something I should do as part of the version bump commit but I never remember!

pferraro:
> Thanks, @pferraro. Can I punish you for doing this by asking you to update the WF release process doc to include this as an aspect of the "bump the version" release task? This is something I should do as part of the version bump commit but I never remember!

Sure, no problem.

pferraro:
@yersan Was I also meant to have moved the extensions referenced within getCurrentRemovedExtensions() to the enum value for WILDFLY_29_0?

testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
yersan: Ups, sorry, I missed this comment and my review was incorrect since there were extensions removed in the current release (WF29).

yersan:
@pferraro Yes, right, I missed those extensions. They were removed, so we need to move them as removed extensions for the new enum we are creating. If we don't move them, the test will fail as soon as we moved the major version in the pom.xml: see https://github.com/wildfly/wildfly/pull/17038/files#diff-d4696c308808a57f331264e1d20639f0ef8a33968635726f1d64a3ecb14e93d5L387-L389

yersan:
See https://github.com/pferraro/wildfly/pull/13 these are the additional changes required for this PR

pferraro:
@yersan PR updated to include modules removed in WF29.

yersan:
Hi @pferraro, a check style issue: "[20:01:33]W:	 [Step 5/6] [ERROR] /opt/buildAgent/work/e34a6f994de9f7c6/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java:39:8: Unused import - java.util.Collections. [UnusedImports]"

bstansberry:
Thanks @pferraro and @yersan

PR #17050, opened 2023-07-30
[WFLY-18296] Remove the org.jboss.jts dependency on jdk.console

bstansberry:
https://issues.redhat.com/browse/WFLY-18296

PR #17051, opened 2023-07-30
[WFLY-18296] Remove the org.jboss.jts dependency on jdk.console

bstansberry:
https://issues.redhat.com/browse/WFLY-18296
Upstream: #17050


bstansberry:
@mmusgrov Is this ok?

mmusgrov (approved):
We do not need to graph the Narayana related transaction stats in WildFly. The plugin was developed when jconsole was the primary management application for transactions, but WildFly brings its own set of management APIs so removing the dependency is the correct thing to do.

jamezp (approved):

PR #17054, opened 2023-08-01
[WFLY-18194] JCA: fix datasource security xsd schema

tadamski:
https://issues.redhat.com/browse/WFLY-18194

### DIFF
--- a/connector/src/main/resources/schema/wildfly-datasources_7_0.xsd
+++ b/connector/src/main/resources/schema/wildfly-datasources_7_0.xsd
@@ -885,19 +885,20 @@
     </xs:complexContent>
   </xs:complexType>
   <xs:complexType name="dsSecurityType">
-      <xs:sequence>
-        <xs:choice>
-          <xs:sequence>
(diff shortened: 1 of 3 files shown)


connector/src/main/resources/schema/wildfly-datasources_7_0.xsd
```diff
@@ -908,14 +908,26 @@
...
+            <xs:element name="user-name" type="xs:token" minOccurs="0">
+              <xs:annotation>
+                <xs:documentation>
+                  <![CDATA[[
+                    Specify the username used when creating a new connection.
+                            Ex: <user-name>sa</user-name>
+                  ]]>
+                </xs:documentation>
+              </xs:annotation>
+            </xs:element>
```
pferraro, line 922: Rather than repeat the user-name element: (user-name, password) | (user-name, credential-reference)
why not use: (user-name, (password | credential-reference))

pferraro (changes requested):

pferraro (approved):

bstansberry:

PR #17057, opened 2023-08-03
[WFLY-18305] Enable dependabot for patch release upgrades.

jamezp:
[WFLY-18305](https://issues.redhat.com/browse/WFLY-18305)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,16 @@
 
 version: 2
 updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+       # Only allow for patch release upgrades
+      - dependency-name: '*'
+        update-types: ['version-update:semver-major', 'version-update:semver-minor']
+    rebase-strategy: "disabled"
+    open-pull-requests-limit: 5
   - package-ecosystem: "github-actions"
     # Workflow files stored in the
     # default location of `.github/workflows`


darranl:
I know you are not touching code but re "No JIRA required." - it seems like a change to how we maintain the project that could be worth a mention in release notes and the Jira gives us that automatic mention.

.github/dependabot.yml
```diff
@@ -2,6 +2,16 @@
...
 version: 2
 updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+       # Only allow for patch release upgrades
+      - dependency-name: '*'
+        update-types: ['version-update:semver-major', 'version-update:semver-minor']
```
rhusar, line 12: What's up with mixing single quotes and double quotes in the same yaml file? 😆

.github/dependabot.yml
```diff
@@ -2,6 +2,16 @@
...
+    rebase-strategy: "disabled"
+    open-pull-requests-limit: 5
```
rhusar, line 14: I like the conservative 5 here, CI is going to be in demand! :)

.github/dependabot.yml
jamezp: That is a copy/paste issue :) We can fix it, but for whatever reason I do it all the time.

jamezp:
> I know you are not touching code but re "No JIRA required." - it seems like a change to how we maintain the project that could be worth a mention in release notes and the Jira gives us that automatic mention.

I'd argue an issue doesn't do much good. All it would really say is "Enable dependabot for dependency upgrades" effectively what the commit message already says. I don't really see a benefit with it, but I can file one.

bstansberry:
> I'd argue an issue doesn't do much good. All it would really say is "Enable dependabot for dependency upgrades" effectively what the commit message already says. I don't really see a benefit with it, but I can file one.

Please do. Thanks. I don't care about the commit message or PR title, but I think this should be recorded.

.github/dependabot.yml
bstansberry: Yes. If we can't keep regular merged update activity going with a queue this size we're doing something wrong.

xstefank (approved):
I would enable it also for minors since they are also expected to be non-breaking.

jamezp:
> I would enable it also for minors since they are also expected to be non-breaking.

I've definitely seen minors that break things. Jackson is a good example of that. I know there are others too so I'm not just picking on them. I'm sure I've done it :)

bstansberry:
Thanks @jamezp, @darranl, @rhusar and @xstefank

PR #17059, opened 2023-08-03
WFLY-18304 Enable manual start of github actions CI workflow [skip ci]

chengfang:
https://issues.redhat.com/browse/WFLY-18304

### DIFF
--- a/.github/workflows/build-manual.yml
+++ b/.github/workflows/build-manual.yml
@@ -1,0 +1,59 @@
+name: Manual Build and Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      os:
+        description: 'OS'
+        required: false
+        default: 'ubuntu-latest'
+        type: choice
+        options:
+          - ubuntu-latest
+          - windows-latest
+      jdk-distribution:
+        description: 'JDK Distribution'
+        required: false
+        default: 'temurin'
+        type: choice
+        options:
+          - temurin
+          - semeru
+          - microsoft
+          - oracle
+          - zulu
+          - corretto
+          - liberica
+      jdk-version:
+        description: 'JDK Version'
+        required: true
+        type: string
+      args:
+        description: 'Arguments like -Dquickly | -DskipTests | -DallTests'
+        required: false
+        default: '-Dquickly'
+        type: string
+      timeout:
+        description: 'Job Timeout Minutes'
+        required: false
+        default: 120
+        type: number
+
+jobs:
+  build:
+    name: Build and Test
+    runs-on: ${{ inputs.os }}
+    timeout-minutes: ${{ fromJSON(inputs.timeout) }}
+    strategy:
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK ${{ inputs.jdk-distribution }} ${{ inputs.jdk-version }}
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ inputs.jdk-distribution }}
+          java-version: ${{ inputs.jdk-version }}
+          cache: 'maven'
+      - name: Build and test with Maven
+        run: mvn -U -B -ntp clean install ${{ inputs.args }}
--- a/.github/workflows/build-only.yml
+++ b/.github/workflows/build-only.yml
@@ -2,24 +2,6 @@ name: WildFly Build
 
 on:
   pull_request:
-  workflow_dispatch:
-    inputs:
-      os:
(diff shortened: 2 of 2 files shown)


.github/workflows/build-manual.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: actions/setup-java@v3
+        with:
+          distribution: ${{ inputs.jdk-distribution }}
+          java-version: ${{ inputs.jdk-version }}
+          cache: 'maven'
+      - name: Clear any possible SNAPSHOT dependencies in the local maven repository
+        run: |
+          LOCAL_REPO="$HOME/.m2/repository"
+          echo "Cleaning SNAPSHOTS from $LOCAL_REPO"
+          if [ -d "$LOCAL_REPO" ]; then find "$LOCAL_REPO" -name "*SNAPSHOT*" | xargs -I {} rm -rfv "{}"; fi
```
jamezp, line 62: Do we need this in this case?

.github/workflows/build-manual.yml
```diff
@@ -0,0 +1,65 @@
...
+        description: 'JDK Version'
+        required: false
+        default: '11'
+        type: choice
+        options:
+          - 11
+          - 17
+          - 18
+      args:
+        description: 'Arguments like -Dquickly | -DskipTests | -DallTests'
```
jamezp, line 37: I realize this is an example, but I'm not sure if this will work with `-DallTests`. If it does, the `timeout-minutes` of 120 is definitely not enough time.

.github/workflows/build-manual.yml
```diff
@@ -0,0 +1,65 @@
...
+          - liberica
+      jdk-version:
+        description: 'JDK Version'
+        required: false
+        default: '11'
+        type: choice
+        options:
+          - 11
+          - 17
+          - 18
```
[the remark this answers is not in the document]
chengfang: Good point. I think we can make it a required, string type, without default (any default value will become obsolete).

.github/workflows/build-manual.yml
chengfang: It should be able to run with -DallTests, with a proper timeout value. But if we make the timeout very long like 4 hours, it starts to lose meaning for other more common use cases like a quick build. How about make timeout configurable with a default value 120?

.github/workflows/build-manual.yml
chengfang: It was copied from `build-only.yml`. Not sure if it's really needed in either case. I guess it's more of a just-in-case safety measure.

.github/workflows/build-manual.yml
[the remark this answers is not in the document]
jamezp: Yes. I've done that with the RESTEasy manual-build: https://github.com/resteasy/resteasy/blob/b0e68f867c7c72ca9947b3338ea2d3523993fa08/.github/workflows/manual-build.yml#L8-L17

.github/workflows/build-manual.yml
jamezp: IIRC there are some tests that don't pass with GitHub Actions, however it's been a long time since I ran it. Making it configurable seems reasonable either way.

.github/workflows/build-manual.yml
jamezp: Yes. The reason that was there was to ensure WildFly builds in a clean environment. There was an issue with a profile not correctly building if the server wasn't built first, so the other workflow just verifies that. I don't think we need it here though.

chengfang:
FYI, there seems to be a bug in github actions when resolving an expression to a number type for `timeout-minutes`.

https://github.com/orgs/community/discussions/55332

Hence the workaround using `${{ fromJSON(inputs.timeout) }}`

jamezp (approved):
This looks good to me.

.github/workflows/build-manual.yml
```diff
@@ -0,0 +1,59 @@
...
+        default: 'temurin'
+        type: choice
+        options:
+          - temurin
+          - semeru
+          - microsoft
+          - oracle
+          - zulu
+          - corretto
+          - liberica
```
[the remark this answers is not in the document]
chengfang: Not too sure about this one. It's quite some typing if users were to enter these weird names. Hopefully vendors will keep these names for long enough time.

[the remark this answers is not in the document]
jamezp: I don't think it's that big of a deal to use a string, but if we run into issues we can change it later.

PR #17060, opened 2023-08-03
WFLY-18306 Default Infinispan remote-timeout should not be less than the default lock-timeout

pferraro:
https://issues.redhat.com/browse/WFLY-18306


darranl:
Do we really need a schema bump?  Unless we update the parsers to be schema aware regarding default values this change actually impacts all of the older schema versions.  Maybe just the model bump so tranformers can at least be correct?

pferraro:
> Do we really need a schema bump? Unless we update the parsers to be schema aware regarding default values this change actually impacts all of the older schema versions. Maybe just the model bump so tranformers can at least be correct?

Technically, the default value as defined in the XSD would be wrong otherwise - but I get your point, that configuration migrated from an older version will have the new default applied - contrary to the default value defined in those schemas.
So, maybe the XSD version bump is unnecessary?

bstansberry (approved):

rhusar (approved):
👍🏼 LGTM!

bstansberry:
Thanks @pferraro @darranl and @rhusar 

If it turns out we don't need the 17.1 api version we should tweak this before we release WF 30.

PR #17061, opened 2023-08-03
[WFLY-18307] Upgrade ByteBuddy to 1.14.5

ropalka:
https://issues.redhat.com/browse/WFLY-18307


bstansberry:
@ropalka -- As you know, @scottmarlow is looking into the suitability of this upgrade, which is only used by ORM. Once he approves, this can get merged. Thanks!

scottmarlow:
https://github.com/wildfly/wildfly/pull/17140 looks to bring Hibernate ORM 6.2.8.Final into WildFly and Hibernate ORM 6.2.8.Final is now using ByteBuddy 1.14.7 so I think once we have upgraded Hibernate ORM, we can also switch to the even newer 1.14.7

scottmarlow:
replaced by https://github.com/wildfly/wildfly/pull/17140

PR #17062, opened 2023-08-04
Mark FaultToleranceMicrometerIntegrationTestCase as intermittently failing effectively @Ignore-ing it.

rhusar:
Unfortunately the cause of this is not clear yet, we need to ignore in the meantime. Tracker for resolution is https://issues.redhat.com/browse/WFLY-18080


pferraro (approved):

bstansberry:
Thanks @rhusar and @pferraro

PR #17065, opened 2023-08-06
WFLY-18298 Set up JRE testing.

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-18298

### DIFF
--- a/.github/workflows/jdk-jre-build.yml
+++ b/.github/workflows/jdk-jre-build.yml
@@ -1,0 +1,61 @@
+name: WildFly JRE build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  wildfly-jre-build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: ['11', '17', '21-ea']
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup JRE ${{ matrix.java }} for building
+        id: setup-java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
+          cache: 'maven'
+          java-package: 'jre'
+
+      - name: Set the JRE Home
+        id: jre-home
+        run: |
+          echo "JRE_HOME=$JAVA_HOME" >> "$GITHUB_OUTPUT"
+
+      - name: Setup JDK ${{ matrix.java }} for building
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '11'
+          cache: 'maven'
+
+      - name: Build and verify WildFly with JRE ${{ matrix.java }}
+        run: |
+          mvn -B clean install -DskipTests -Denforcer.skip=true -Dcheckstyle.skip=true
+          cd testsuite
+          echo "Running: mvn -B verify -Dcontainer.java.home=${{ steps.jre-home.outputs.JRE_HOME }} -fae"
+          mvn -B verify -Dcontainer.java.home=${{ steps.jre-home.outputs.JRE_HOME }} -fae
+
+      - name: Upload Surefire Reports on Failure
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: surefire-reports-${{ matrix.java }}
+          path: '**/surefire-reports/'
+
+      - name: Upload Server Logs on Failure
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: server-logs-${{ matrix.java }}
+          path: '**/server.log'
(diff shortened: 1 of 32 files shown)


RanabirChakraborty:
@bstansberry when using jdk11 at that time JRE testing is failing with - https://github.com/wildfly/wildfly/actions/runs/5777713339/job/15658186148#step:4:60467 
```
Error:  Failed to execute goal on project wildfly-galleon-pack: Could not resolve dependencies for project org.wildfly:wildfly-galleon-pack:pom:30.0.0.Beta1-SNAPSHOT: Could not find artifact org.wildfly:wildfly-ee-galleon-pack:zip:30.0.0.Beta1-SNAPSHOT in jboss-public-repository-group (https://repository.jboss.org/nexus/content/groups/public/) -> [Help 1]
```

and with jdk17 it's failing for `jdk.jconsole` presence. 

```
{"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-15" => "WFLYCTL0158: Operation handler failed: org.jboss.modules.ModuleLoadError: jdk.jconsole"}}
	at org.wildfly.galleon.plugin.config.generator.BaseConfigGenerator.doHandle(BaseConfigGenerator.java:144)
	at org.wildfly.galleon.plugin.config.generator.BaseConfigGenerator.endBatch(BaseConfigGenerator.java:162)
	at org.wildfly.galleon.plugin.config.generator.ForkedConfigGenerator.executeScript(ForkedConfigGenerator.java:63)
	at org.wildfly.galleon.plugin.config.generator.ForkedConfigGenerator.forkedForEmbedded(ForkedConfigGenerator.java:45)
	at org.wildfly.galleon.plugin.server.ForkedProcessRunner.main(ForkedProcessRunner.java:56)
```

.github/workflows/jdk-jre-test.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  jdk-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up for JDK testing
```
bstansberry, line 17: s/testing/building

bstansberry (changes requested):
Thanks @RanabirChakraborty. I'd like a couple minor things changed. If you rebase on main before you push again we'll see how the test goes. I've merged the fix for the jdk.console issue. It's possible something else will fail (we have another known issue related to org.glassfish.jaxb) but if it does, that's good; it means the test is helping improve WF. :)

bstansberry:
@RanabirChakraborty I don't know. If you can't figure it out please ask on the zulip thread.

Also, please have a look at https://stackoverflow.com/questions/21638697/disable-maven-download-progress-indication and tweak the maven commands to get rid of all the download progress logging. I had time to spend a few minutes trying to figure out what might have went wrong, but I spent those minutes scrolling through download progress logging and had to give up. :)

RanabirChakraborty:
@jamezp @bstansberry can you please review this once.

testsuite/integration/manualmode/src/test/resources/arquillian.xml
```diff
@@ -52,7 +54,7 @@
                 <!-- AS7-4070 -->
                 <property name="waitForPorts">${as.debug.port:8787} ${as.managementPort:9990}</property>
                 <property name="waitForPortsTimeoutInSeconds">8</property>
-                <property name="modulePath">${basedir}/target/wildfly/modules</property>
```
jamezp, line 55: I'm not sure this should have been removed. It's an odd looking property to me, but it looks like it's been that way for a while so we should likely keep it.

.github/workflows/jdk-jre-build.yml
```diff
@@ -0,0 +1,47 @@
...
+          distribution: 'temurin'
+          java-version: '11'
+          cache: 'maven'
+
+      - name: Build and verify WildFly with JRE ${{ matrix.java }}
+        run: |
+          mvn -B clean install -DskipTests -Denforcer.skip=true -Dcheckstyle.skip=true -Dts.basic
+          cd testsuite
+          echo "Running: mvn -B verify -Dcontainer.java.home=${{ steps.jre-home.outputs.JRE_HOME }} -Dts.basic -fae"
+          mvn -B verify -Dcontainer.java.home=${{ steps.jre-home.outputs.JRE_HOME }} -Dts.basic -fae
```
jamezp, line 47: We should add something like this so we can gather the logs for failed jobs:
```yaml

      - uses: actions/upload-artifact@v3
        if: failure()
        with:
          name: surefire-reports-${{ matrix.java }}
          path: '**/surefire-reports/'
      - uses: actions/upload-artifact@v3
        if: failure()
        with:
          name: server-logs-${{ matrix.java }}
          path: '**/server.log'
```

jamezp (changes requested):
The failures look to be some race conditions in tests. We might want to have a look at them, but we will likely get some better information if we add the log archive uploads.

jamezp (approved):
This looks good to me now. I placed a hold on this for now. We may need to not use the `-Dts.basic`, but I want to see how CI does.

jamezp:
I just re-ran the JRE tests on here. I think these are going to be intermittent failures. I have not had a chance to look at these tests, but they're could be some race conditions in them given the failures reported.

jamezp (approved):

bstansberry (approved):

bstansberry:
Thanks @RanabirChakraborty and @jamezp

PR #17070, opened 2023-08-07
[WFLY-18311] Eliminate Hibernate Validator dependency on legacy Xerces

jasondlee:
https://issues.redhat.com/browse/WFLY-18311

Remove Xerces module dep


bstansberry:
Thanks @jasondlee

PR #17073, opened 2023-08-08
[WFLY-18326] Verify that jboss.server.name can be resolved as a JVM o…

yersan:
…ption in a managed server

Jira issue: https://issues.redhat.com/browse/WFLY-18326

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/management-tasks/JVM_settings.adoc
@@ -113,6 +113,12 @@ This ability is useful when you need to configure JVM settings without specifyin
 }
 ----
 
+=== Other server properties that can be resolved as JVM options on domain mode
+
+In addition to the aforementioned server properties, the Host Controller can resolve the following `jboss.server-xyz` properties as JVM options:
+
+* jboss.server.name
+
 
 == Standalone Server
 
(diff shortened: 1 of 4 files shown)


bstansberry:
@yersan Why is this in full WF? It seems like something WildFly Core can test.

yersan:
@bstansberry It takes advantage of existing test cases available in the WildFly Test Suite. The test case deploys a Servlet and I think that was the original intention of placing the test cases in WildFly instead of wildfly core, to avoid dealing with Undertow and activate it on WildFly Core.

yersan:
@bstansberry , yes, reading the EAP7-823 test plan, the original intention was to add the test cases in WildFly instead of WildFly Core just because Undertow was already prepared out of the box to process .war files in WildFly and I think I wanted to avoid more complications dealing with this detail in Core.

If that's not an option, looking at it after some time, it seems that we could try to move all of them to WildFly Core (the EAP7-823 and this enhancement), we only need a servlet deployed, and that seems doable.

Maybe, could we get this done in the WildFly test suite now and later move everything to WildFly Core? Does it make sense to you? in any case, I've just noticed this PR missed add the new resolvable property to the WildFly documentation

bstansberry (approved):
Sorry, @yersan -- this is fine. I hadn't had enough coffee when I looked at this and it didn't register that this was a minor update to an existing test.

bstansberry:
@yersan Note that I think JVMServerPropertiesTestCase could be easily enough moved to WildFly Core. Just create a UndertowServiceActivator subclass and override getHttpHandler(). There are a number of tests in core that do this. There's no need to use a servlet.

But that's a nice to have for someone to do some day. Not at all urgent or in scope for this issue.

yersan:
Yes @bstansberry , thanks, you're right, we only would need a way to verify the expected system properties on each managed server. I think when I prepared them the first time I went directly to WildFly full because I was thinking of a war application returning the properties, and at that time I thought wildfly core was unable to do something like that. I'll create a follow-up Jira to move those tests once this part is finally merged.

yersan:
Converted to draft since it still needs a WF Core upgrade, I'm now going to add the pending Javadocs

bstansberry (approved):

yersan:
@bstansberry , this one seems to be ready to go, I've created https://issues.redhat.com/browse/WFCORE-6497 as follow up once this is merged.

bstansberry:
Thanks @yersan

PR #17075, opened 2023-08-09
[WFLY-18047]:Remove legacy Xerces dependency from webservice modules

jimma:
https://issues.redhat.com/browse/WFLY-18047

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/com/sun/xml/messaging/saaj/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/com/sun/xml/messaging/saaj/main/module.xml
@@ -40,7 +40,6 @@
         <module name="java.xml"/>
         <module name="jakarta.xml.soap.api"/>
         <module name="jakarta.activation.api"/>
-        <module name="org.apache.xerces" services="import" />
     </dependencies>
 
 </module>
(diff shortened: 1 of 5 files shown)


asoldano (changes requested):
@jimma you seem to need to remove an unused import of org.jboss.wsf.spi.deployment.Endpoint in the WSDependenciesProcessor.java

jimma:
> @jimma you seem to need to remove an unused import of org.jboss.wsf.spi.deployment.Endpoint in the WSDependenciesProcessor.java

Ah,Yes. Fixed and pushed again.  Thanks @asoldano

asoldano (approved):

jbliznak:
Could this test be allowed again after these fixes? https://github.com/wildfly/wildfly/blob/main/testsuite/integration/ws/src/test/java/org/jboss/as/test/integration/ws/wsrm/ReliableServiceTestCase.java#L49

jimma:
Thanks for bring this up @jbliznak. Yes. That's the test I am going to have a look after all other tests are passed.

bstansberry:
@jimma To fix the LayersTestCase failures you need to update org.jboss.as.test.shared.LayersTestBase.NOT_USED_COMMON.  This PR eliminates the last use of the org.apache.xerces module by any layer, so the test demands that we account for why the module still exists. The NOT_USED_COMMON array is where we do that. Basically the fix would be adding an element to the array preceded by an explanatory comment.

The reason the module still exists is because it is a public module. It was deprecated in WF 29, but it's still public.

We can/should discuss what we want to do with the module on wildfly-dev but I think doing anything about that is out of scope for the JIRA you're fixing here.

jimma:
Done. Thanks @bstansberry

PR #17090, opened 2023-08-11
[WFLY-18245] Move the org.apache.avro module from the wildfly-ee feat…

bstansberry:
…ure pack to the wildfly-preview feature pack

https://issues.redhat.com/browse/WFLY-18245


bstansberry:
@yrodiere FYI.

yrodiere (approved):

PR #17067, opened 2023-08-07
WFLY-18233 Optimize payload of ATTRIBUTE granularity mapping in distributed session manager using Infinispan's Cache.compute(...)

pferraro:
https://issues.redhat.com/browse/WFLY-18233

Requires https://github.com/wildfly/wildfly/pull/17238

PR #17096, opened 2023-08-12
WFLY-18315 Optimize payload of distributed session metadata

pferraro:
https://issues.redhat.com/browse/WFLY-18315

Requires https://github.com/wildfly/wildfly/pull/17067

Also fixes https://issues.redhat.com/browse/WFLY-18389


rhusar (approved):
👍🏼 I don't see any issues.

PR #17097, opened 2023-08-14
[WFLY-18345] added jdk.security.jgss to module

gaol:
Issue: https://issues.redhat.com/browse/WFLY-18345

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/apache/ws/security/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/apache/ws/security/main/module.xml
@@ -34,6 +34,7 @@
     </resources>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


jimma:
Would it be better that we make this dependency as optional ?

bstansberry:
@gaol Thanks! And good news. It looks like this module can reasonably safely be regarded as a 'JRE' module.  @fjuma or @darranl FYI in case you disagree.

For further context, please see https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/thread/U2MYHTRWBDCZ22XT6DIIVETYPFZ2P7DO/ Basically all the 'JRE' distributions I checked include the jdk.security.jgss module.

jimma:
@gaol @bstansberry  This looks good to me as this module exists in all java runtimes.

jimma (approved):

gaol:
Updated to fix format.

PR #17104, opened 2023-08-14
[WFLY-18353] Update TestCustomPrincipalTransformer to no longer rely on the Elytron subsystem

jessicarod7:
https://issues.redhat.com/browse/WFLY-18353

This fix is also needed to fully implement wildfly/wildfly-core#5561.

### DIFF
--- a/testsuite/integration/elytron/pom.xml
+++ b/testsuite/integration/elytron/pom.xml
@@ -269,11 +269,6 @@
             <artifactId>wildfly-cli</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.wildfly.core</groupId>
-            <artifactId>wildfly-elytron-integration</artifactId>
-            <scope>test</scope>
-        </dependency>
         <!-- Required for Java 11 -->
         <dependency>
             <groupId>org.wildfly.arquillian</groupId>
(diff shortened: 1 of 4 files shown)


fjuma (changes requested):
Thanks @cam-rod! Just a very minor comment, would be good to update the commit message and issue title to something like "Update the TestCustomPrincipalTransformer to no longer rely on the elytron subsystem".

Would you be able to check if we can also remove the `org.wildfly.core.wildfly-elytron-integration` test dependency now too?

cam-rod:
> Thanks @cam-rod! Just a very minor comment, would be good to update the commit message and issue title to something like "Update the TestCustomPrincipalTransformer to no longer rely on the elytron subsystem".
> 
> Would you be able to check if we can also remove the `org.wildfly.core.wildfly-elytron-integration` test dependency now too?

Updated the title and removed the subsystem dependency

fjuma:
@cam-rod Looks like there are some compilation failures.

cam-rod:
Thanks for the heads up, looks like the CredentialStore tests were trying to reuse a subsystem error message. I directly copied the string over and everything seems fine now.

fjuma (approved):
Thanks @cam-rod!

bstansberry:
Thanks @cam-rod and @fjuma

PR #17116, opened 2023-08-17
WFLY-18365: ejb timer - add new attribute to use with datasource capa…

### LINKED ISSUE #801: EJB: datasource by name in EJB timer configuration
### Description

Add ability to configure datasource by name in EJB timer configuration.

### Issue Contact

tadamski

michpetrov:
…bility

Issue: [WFLY-18365](https://issues.redhat.com/browse/WFLY-18365)


chengfang:
This looks like a new RFE. Is there an analysis document?

michpetrov:
@chengfang ah, you mean this has to go through wildfly-proposals? I can try.

chengfang:
@michpetrov yes, this is a significant addition/change, and will impact various parties, and so following the RFE process will help to track it.

tadamski (changes requested):
@michpetrov 
Could you please preserve the test for datasource-jndi-name and create another one for datasource-name so that both of the configuration possibilities are checked. 

Also, it seems to me that this PR requires transformer addition.

michpetrov:
@tadamski I assume there's no annotation to run the tests with different configurations? As it is I'm looking at introducing a bunch of abstract classes.
As for the transformer, I don't think there's a way to convert "name" to "jndi-name", not unless I can read the datasource subsystem from the context (I haven't checked).

PR #17120, opened 2023-08-21
Fix issue links pattern for format checks

xstefank:
No issue is required.

@wildfly-bot[bot] skip format

### DIFF
--- a/.github/wildfly-bot.yml
+++ b/.github/wildfly-bot.yml
@@ -274,12 +274,12 @@ wildfly:
   emails:
     - mstefank@redhat.com
     - pberan@redhat.com
-    - rbudinsk@redhat.com 
+    - rbudinsk@redhat.com
     - brian.stansberry@redhat.com
   
   format:
     description:
       regexes:
-        - pattern: "JIRA:\\s+https://issues.redhat.com/browse/WFLY-\\d+|https://issues.redhat.com/browse/WFLY-\\d+"
-          message: "The description must contain a link to the JIRA issue \n(JIRA: https://issues.redhat.com/browse/WFLY-XYZ or https://issues.redhat.com/browse/WFLY-XYZ)"
+        - pattern: "https://issues.redhat.com/browse/WFLY-\\d+"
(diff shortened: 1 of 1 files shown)


xstefank:
@wildfly-bot test

The-Huginn (approved):
LGTM

.github/wildfly-bot.yml
```diff
@@ -280,6 +280,6 @@ wildfly:
   format:
     description:
       regexes:
-        - pattern: "JIRA:\\s+https://issues.redhat.com/browse/WFLY-\\d+|https://issues.redhat.com/browse/WFLY-\\d+"
-          message: "The description must contain a link to the JIRA issue \n(JIRA: https://issues.redhat.com/browse/WFLY-XYZ or https://issues.redhat.com/browse/WFLY-XYZ)"
+        - pattern: "https://issues.redhat.com/browse/WFLY-\\d+|https://issues.jboss.org/browse/WFLY-\\d+"
+          message: "The description must contain a link to the JIRA issue \n(https://issues.redhat.com/browse/WFLY-XYZ or https://issues.jboss.org/browse/WFLY-XYZ)"
```
jamezp, line 284: We should leave off the "org https://issues.jboss.org/browse/WFLY-XYZ)" IMO. We should be using issues.redhat.com as that is what issues.jboss.org redirects to. I think allowing it in the pattern is fine though.

The-Huginn:
I have just notices we have there a few trailing space characters i.e. `\s`. Might be worth it removing them as well. Such as `directories:\s` under `-id: clustering`

.github/wildfly-bot.yml
xstefank: well, ok, but still it seems weird that it is still a valid link and we fail the format check

xstefank:
@bstansberry this could go in

.github/wildfly-bot.yml
jamezp: Sorry, I realize it's merged now. What I meant was IMO it would be okay keep it from failing with the pattern, but we shouldn't encourage the usage in the error message.

That said, the only way I could see the jboss.org URL being used is if it's manually typed.

PR #17128, opened 2023-08-23
[WFLY-18348] PermissionUtils should delegate to the WildFly Core variant where possible

lvydra:
https://issues.redhat.com/browse/WFLY-18348

### DIFF
--- a/boms/standard-test/pom.xml
+++ b/boms/standard-test/pom.xml
@@ -501,18 +501,6 @@
                 <version>${version.org.wildfly.extras.creaper}</version>
                 <scope>test</scope>
             </dependency>
-
-            <dependency>
-                <groupId>xom</groupId>
-                <artifactId>xom</artifactId>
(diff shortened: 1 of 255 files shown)


rhusar (changes requested):
This creates a lot of failures triggered by `Caused by: java.lang.NoClassDefFoundError: nu/xom/Node` since clearly org.jboss.as.test.shared.PermissionUtils is still relying on xom.

lvydra:
Hi @rhusar , thank you for your comment. It seems that the problem is in wildfly-core dependency not yet including changes done in the https://issues.redhat.com/browse/WFCORE-6461.

bstansberry:
When I merged the WFCORE-6461 fix I set the wrong Fix Version. So it's not yet in the core release the WF uses.

bstansberry (approved):

bstansberry:
Thanks @lvydra!

PR #17134, opened 2023-08-30
[WFLY-17451] add filter/filter-ref capabilites to enforce runtime req…

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-17451

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
@@ -6,13 +6,12 @@
 package org.wildfly.extension.undertow.filters;
 
 import org.jboss.as.controller.AbstractAddStepHandler;
+import org.jboss.as.controller.CapabilityServiceBuilder;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.Service;
-import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
-import org.jboss.msc.service.ServiceTarget;
 import org.wildfly.extension.undertow.UndertowService;
 
 import java.util.function.Consumer;
@@ -32,12 +31,11 @@ class FilterAdd extends AbstractAddStepHandler {
     @Override
     protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
         final String name = context.getCurrentAddressValue();
-        final ServiceTarget target = context.getServiceTarget();
-        final ServiceBuilder<?> sb = target.addService(UndertowService.FILTER.append(name));
-        final Consumer<PredicateHandlerWrapper> serviceConsumer = sb.provides(UndertowService.FILTER.append(name));
-        PredicateHandlerWrapper wrapper = this.factory.createHandlerWrapper(context, model);
-        sb.setInstance(Service.newInstance(serviceConsumer, wrapper));
-        sb.setInitialMode(ServiceController.Mode.ON_DEMAND);
-        sb.install();
+        final CapabilityServiceBuilder<?> csb = context.getCapabilityServiceTarget().addCapability(FilterCapabilities.FILTER_CAPABILITY.getDefinition());
+        final Consumer<PredicateHandlerWrapper> serviceConsumer =  csb.provides(FilterCapabilities.FILTER_CAPABILITY.getDefinition(), UndertowService.FILTER.append(name));
+        final PredicateHandlerWrapper wrapper = this.factory.createHandlerWrapper(context, model);
+        csb.setInstance(Service.newInstance(serviceConsumer, wrapper));
+        csb.setInitialMode(ServiceController.Mode.ON_DEMAND);
+        csb.install();
     }
 }
--- a/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
@@ -1,0 +1,64 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.wildfly.extension.undertow.filters;
+
+import static org.wildfly.extension.undertow.Capabilities.CAPABILITY_FILTER;
+import static org.wildfly.extension.undertow.Capabilities.CAPABILITY_HOST_FILTER_REF;
+import static org.wildfly.extension.undertow.Capabilities.CAPABILITY_LOCATION_FILTER_REF;
+
+import java.util.function.Function;
+
+import org.jboss.as.clustering.controller.Capability;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.capability.QuaternaryCapabilityNameResolver;
+import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.capability.TernaryCapabilityNameResolver;
+import org.jboss.as.controller.capability.UnaryCapabilityNameResolver;
+import org.wildfly.extension.undertow.UndertowFilter;
+
+/**
+ * @author Paul Ferraro
+ * @author baranowb
+ */
+enum FilterCapabilities implements Capability {
+
+    FILTER_CAPABILITY(CAPABILITY_FILTER, PredicateHandlerWrapper.class),
+    FILTER_HOST_REF_CAPABILITY(CAPABILITY_HOST_FILTER_REF, UndertowFilter.class, TernaryCapabilityNameResolver.GRANDPARENT_PARENT_CHILD),
+    FILTER_LOCATION_REF_CAPABILITY(CAPABILITY_LOCATION_FILTER_REF, UndertowFilter.class, QuaternaryCapabilityNameResolver.GREATGRANDPARENT_GRANDPARENT_PARENT_CHILD);
+
+    private final RuntimeCapability<Void> definition;
+
+    FilterCapabilities(String name, Class<?> serviceValueType) {
+        this.definition = RuntimeCapability.Builder.of(name, true, serviceValueType)
+                .setDynamicNameMapper(UnaryCapabilityNameResolver.DEFAULT).build();
+    }
+
+    FilterCapabilities(String name, Class<?> serviceValueType, Function<PathAddress, String[]> nameResolver) {
+        this.definition = RuntimeCapability.Builder.of(name, true, serviceValueType)
+                .setDynamicNameMapper(nameResolver).build();
+    }
+
+    @Override
+    public RuntimeCapability<Void> getDefinition() {
+        return this.definition;
+    }
+}
--- a/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
@@ -1,23 +1,6 @@
 /*
- * JBoss, Home of Professional Open Source.
- * Copyright 2023, Red Hat, Inc., and individual contributors
- * as indicated by the @author tags. See the copyright.txt file in the
- * distribution for a full listing of individual contributors.
- *
- * This is free software; you can redistribute it and/or modify it
- * under the terms of the GNU Lesser General Public License as
- * published by the Free Software Foundation; either version 2.1 of
- * the License, or (at your option) any later version.
- *
- * This software is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
- * Lesser General Public License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public
- * License along with this software; if not, write to the Free
- * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
- * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
  */
 package org.wildfly.extension.undertow.filters;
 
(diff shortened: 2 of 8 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
```diff
@@ -54,12 +52,11 @@ class FilterAdd extends AbstractAddStepHandler {
...
-        final ServiceBuilder<?> sb = target.addService(UndertowService.FILTER.append(name));
-        final Consumer<HandlerWrapper> serviceConsumer = sb.provides(UndertowService.FILTER.append(name));
+        CapabilityServiceBuilder<?> csb = context.getCapabilityServiceTarget().addCapability(FilterCapabilities.FILTER_CAPABILITY.getDefinition());
+        Consumer<HandlerWrapper> serviceConsumer =  csb.provides(FilterCapabilities.FILTER_CAPABILITY.getDefinition(), UndertowService.FILTER.append(name));
         HandlerWrapper wrapper = this.factory.createHandlerWrapper(context, model);
-        sb.setInstance(Service.newInstance(serviceConsumer, wrapper));
-        sb.setInitialMode(ServiceController.Mode.ON_DEMAND);
-        sb.install();
+        csb.setInstance(new FilterProviderService(serviceConsumer, wrapper));
+        csb.setInitialMode(ServiceController.Mode.ACTIVE);
```
bstansberry, line 59: I don't think changing the mode should be necessary for this JIRA. ON_DEMAND is the natural fit for the use of this service.

baranowb: Well, idea was to have it more pro-active about error reporting, but I dont have strong inclination to die on this hill

bstansberry: FilterProviderService doesn't really do anything in start and has no dependencies, so it's not going to error.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
```diff
@@ -54,12 +53,20 @@ class FilterAdd extends AbstractAddStepHandler {
...
-        final ServiceTarget target = context.getServiceTarget();
-        final ServiceBuilder<?> sb = target.addService(UndertowService.FILTER.append(name));
-        final Consumer<HandlerWrapper> serviceConsumer = sb.provides(UndertowService.FILTER.append(name));
+        CapabilityServiceBuilder<?> csb = context.getCapabilityServiceTarget().addCapability(FilterCapabilities.FILTER_CAPABILITY.getDefinition());
+        Consumer<HandlerWrapper> serviceConsumer =  csb.provides(FilterCapabilities.FILTER_CAPABILITY.getDefinition(), UndertowService.FILTER.append(name));
         HandlerWrapper wrapper = this.factory.createHandlerWrapper(context, model);
-        sb.setInstance(Service.newInstance(serviceConsumer, wrapper));
-        sb.setInitialMode(ServiceController.Mode.ON_DEMAND);
-        sb.install();
+        csb.setInstance(new FilterProviderService(serviceConsumer, wrapper));
```
pferraro, line 59: Why introduce a redundant FilterProviderService class?  The implementation is identical to the previous logic.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
```diff
@@ -54,12 +53,20 @@ class FilterAdd extends AbstractAddStepHandler {
...
     }
+
+    @Override
+    protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource)
+            throws OperationFailedException {
+        super.recordCapabilitiesAndRequirements(context, operation, resource);
+        //NOTE: its dynamic by design.
+        context.registerCapability(FilterCapabilities.FILTER_CAPABILITY.getDefinition().fromBaseCapability(context.getCurrentAddress()));
+    }
+
```
pferraro, line 71: You do not need to override this method.  Capability registration is already handled via the super implementation.
See: https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java#L216

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
```diff
@@ -0,0 +1,90 @@
...
+                if(include) {
+                    elements.add(e.getValue());
+                } else if(e.getKey().equals(ModelDescriptionConstants.SERVER)){
+                    elements.add(e.getValue());
+                    include = true;
+                }
+            }
+            return elements.toArray(new String[elements.size()]);
+        }
+    });
```
pferraro, line 72: I do not think you can reuse the same RuntimeCapability implementation for both contexts (i.e. per-host vs per-location), since the name resolution logic if different, as is the number of dynamic segments.  This will otherwise break galleon.  I ran into a similar problem when intitially modeling capabilities for default caches in the Infinispan subsystem.  Essentially, a capability with a given name must have a fixed number of name segments.

Rather, you'll want to parameterize the FilterRefDefinition with its RuntimeCapability.  That way the host can register the variant with 3 dynamic name segments, and the location can register the variant with 4 dynamic name segments.

FYI, once https://github.com/wildfly/wildfly-core/pull/5516 is merged, you can use [1] for the host filter-ref and [2] for the location filter-ref.

[1] https://github.com/pferraro/wildfly-core/blob/WFCORE-6311/controller/src/main/java/org/jboss/as/controller/capability/TernaryCapabilityNameResolver.java#L27
[2] https://github.com/pferraro/wildfly-core/blob/WFCORE-6311/controller/src/main/java/org/jboss/as/controller/capability/QuaternaryCapabilityNameResolver.java#L28

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRefCapability.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+import java.util.function.Function;
+
+import org.jboss.as.clustering.controller.Capability;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+
+class FilterRefCapability implements Capability {
```
pferraro, line 14: What is the purpose of this class?  It does not seem to be referenced anywhere.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRefDefinition.java
```diff
@@ -128,15 +131,44 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
...
             sb.install();
         }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource)
+                throws OperationFailedException {
+            super.recordCapabilitiesAndRequirements(context, operation, resource);
+            //NOTE: its dynamic by design.
+            context.registerCapability(FilterCapabilities.FILTER_REF_CAPABILITY.getDefinition().fromBaseCapability(context.getCurrentAddress()));
+        }
```
pferraro, line 152: You do not need to override this method. Capability registration is already handled via the super implementation.
See: https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java#L216

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRemove.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource)
+            throws OperationFailedException {
+        super.recordCapabilitiesAndRequirements(context, operation, resource);
+
+        context.deregisterCapability(FilterCapabilities.FILTER_CAPABILITY.getDefinition().getDynamicName(context.getCurrentAddress()));
+    }
+
+}
```
pferraro, line 53: You do not need to subclass ServiceRemoveStepHandler.
Capability deregistration is already handled via the super implementation.
See: https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractRemoveStepHandler.java#L175

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRefDefinition.java
```diff
@@ -128,15 +131,44 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
...
+            super(addOperation);
+        }
+
+        @Override
+        protected void recordCapabilitiesAndRequirements(OperationContext context, ModelNode operation, Resource resource)
+                throws OperationFailedException {
+            super.recordCapabilitiesAndRequirements(context, operation, resource);
+            context.deregisterCapability(FilterCapabilities.FILTER_REF_CAPABILITY.getDefinition().getDynamicName(context.getCurrentAddress()));
+        }
+
```
pferraro, line 172: You do not need to subclass ServiceRemoveStepHandler.
Capability deregistration is already handled via the super implementation.
See: https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractRemoveStepHandler.java#L175

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
baranowb: AFAIR Brian said it has to be done this way.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRefCapability.java
baranowb: ha. power out, thought it did not save and must have picked another name.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRemove.java
baranowb: Same with this one, explicitly requested by Brian.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
baranowb: Ok, I will rebase it once #5516 is in.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
pferraro: Provided that the filter's ResourceDefinition registers the RuntimeCapability (and it should), the capability will be auto-registered with the OperationContext by the super implementation.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterRemove.java
pferraro: Provided that the filter's ResourceDefinition registered the RuntimeCapability (and it should), the capability will be auto-deregistered with the OperationContext by the super implementation.

baranowb:
So far Im unable to reproduce this failure manually.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterAdd.java
bstansberry: @baranowb @pferraro My Sept 7, 2023 comment about using this method was in reference to FilterRefAdd, not FilterAdd.

The issue is this:

https://github.com/wildfly/wildfly/pull/17134/files#diff-8a27bf46cb96032958c3cb8549c72bb98a745b73e2f3aa2756eaeb698f532a1fR130

```
final Supplier<HandlerWrapper> fSupplier = csb.requiresCapability(Capabilities.CAPABILITY_FILTER, HandlerWrapper.class, name);
```

This is a capability requirement that is not enforced at Stage.MODEL; it's only enforced at Stage.RUNTIME via this service dependency. So an admin-only server can be used to create an invalid config that's only detected when the config is switched out of admin-only. And the bad config doesn't abort boot, which it should.

The issue is this capability reference is not done via an attribute; it's the filter-ref resource's own name that specifies the name of the required filter. So it's at the level of the filter-ref resource that we need to record the capability requirement in Stage.MODEL. AbstractAddStepHandler. recordCapabilitiesAndRequirements is the place I could think of where such a requirement can be registered.

bstansberry:
@baranowb @fl4via @pferraro My impression is this one could have gotten in but just stalled out. Can we get it done?

I'd like to get the PR queue down to around 40 open PRs, about 1/2 the current size.

baranowb:
ATM Im stuck at: "WFLYCTL0158: Operation handler failed: java.lang.IllegalArgumentException: WFLYCTL0394 - either I missed something or it class with subsequent changes.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
```diff
@@ -0,0 +1,64 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
bstansberry, line 4: Please change this to the standard WildFly header.

bstansberry (changes requested):
Thanks @baranowb. Besides the header issue I noted, I sent you https://github.com/baranowb/wildfly/pull/1 to illustrate how the filter-ref requirement for the filter needs to be done. With what's here now the system will not detect that the config is wrong until it tries to do service wiring. So it won't detect it in an --admin-only server, where that service wiring code is not executed.

The way this needs to be handled is not ideal because of how Galleon works, but what I did is next best. Comments on my PR to you explain what I mean.

baranowb:
Thanks. I will compare to what I have locally, since it is still failing, I might be missing something that you have set up properly.

undertow/src/main/java/org/wildfly/extension/undertow/filters/FilterCapabilities.java
```diff
@@ -0,0 +1,64 @@
...
+import org.jboss.as.controller.capability.RuntimeCapability;
+import org.jboss.as.controller.capability.TernaryCapabilityNameResolver;
+import org.jboss.as.controller.capability.UnaryCapabilityNameResolver;
+import org.wildfly.extension.undertow.UndertowFilter;
+
+/**
+ * @author Paul Ferraro
+ * @author baranowb
+ */
+enum FilterCapabilities implements Capability {
```
pferraro, line 25: FYI, org.jboss.as.clustering.controller.Capability is deprecated.

baranowb: https://github.com/wildfly/wildfly/blob/main/clustering/common/src/main/java/org/jboss/as/clustering/controller/Capability.java  ? will be ?

bstansberry: It's not javadoced as deprecated and it's used elsewhere in this subsystem and in the related mod_cluster subsystem. So I'm not going to delay merging this if this is the only issue with this PR.

bstansberry: Haha; I see the other use in the Undertow subsystem is removed below. But still, I don't want to further block on this point.

bstansberry:
Thanks @baranowb -- I'm closing this because #20249 picked it up and is merged. Sorry for forgetting to come back to this.

PR #17135, opened 2023-08-30
[WFLY-18380]: Message-destination-type is ejb-jar.xml is ignored.

ehsavoie:
* Processing Message-destination-type to convert it to an activation config property called 'messageDestinationType'

Issue: https://issues.redhat.com/browse/WFLY-18380

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MessageDrivenComponentDescriptionFactory.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MessageDrivenComponentDescriptionFactory.java
@@ -118,16 +118,7 @@ public class MessageDrivenComponentDescriptionFactory extends EJBComponentDescri
                 deploymentDescriptorEnvironment = new DeploymentDescriptorEnvironment("java:comp/env/", beanMetaData);
 
                 messagingType = beanMetaData.getMessagingType();
-                final ActivationConfigMetaData activationConfigMetaData = beanMetaData.getActivationConfig();
-                if (activationConfigMetaData != null) {
-                    final ActivationConfigPropertiesMetaData propertiesMetaData = activationConfigMetaData
-                            .getActivationConfigProperties();
-                    if (propertiesMetaData != null) {
-                        for (final ActivationConfigPropertyMetaData propertyMetaData : propertiesMetaData) {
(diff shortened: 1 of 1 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/MessageDrivenComponentDescriptionFactory.java
```diff
@@ -181,22 +172,35 @@ private String getMessageListenerInterface(final CompositeIndex compositeIndex,
...
+                    activationConfigProps.put(propName, propValue);
+                }
             }
         }
+        if (mdb.getAcknowledgeMode() != null && !mdb.getAcknowledgeMode().isBlank()) {
+            activationConfigProps.put("acknowledgeMode", mdb.getAcknowledgeMode());
+        }
+        if (mdb.getMessageDestinationType() != null && !mdb.getMessageDestinationType().isBlank()) {
+            activationConfigProps.put("destinationType", mdb.getMessageDestinationType());
+            activationConfigProps.put("messageDestinationType", mdb.getMessageDestinationType());
```
chengfang, line 196: There are test failures caused by the above line:
```
java.lang.AssertionError: Unknown activation config property: messageDestinationType
  at org.junit.Assert.fail(Assert.java:89)
  at org.jboss.as.test.integration.ejb.management.deployments.EjbJarRuntimeResourceTestBase.testComponent(EjbJarRuntimeResourceTestBase.java:289)
  at org.jboss.as.test.integration.ejb.management.deployments.EjbJarRuntimeResourceTestBase.testNoTimerMDB(EjbJarRuntimeResourceTestBase.java:135)
```
Because `messageDestinationType` is not recognized as a valid activation config property name by the test. 

@ehsavoie Do we need this property `messageDestinationType` in addition to the similar standard one `destinationType`?

ehsavoie: well I wasn't sure if some resrouce adapter might be expecting it. I can simply remove it

chengfang (approved):

bstansberry (approved):

bstansberry:
Thanks @ehsavoie and @chengfang

PR #17138, opened 2023-08-31
WFLY-18401 Infinispan parsers should generated operations for resources that would otherwise be auto-added

pferraro:
https://issues.redhat.com/browse/WFLY-18401

As requested by @ehsavoie


ehsavoie (approved):
Nicely done ! :)

PR #17140, opened 2023-09-01
WFLY-18399 Upgrade to Hibernate 6.2.8.Final release

scottmarlow:
https://issues.redhat.com/browse/WFLY-18399 for the Hibernate ORM upgrade to 6.2.8
https://issues.redhat.com/browse/WFLY-18307 for the ByteBuddy upgrade to 1.14.7

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -504,7 +504,7 @@
         <version.org.glassfish.soteria>3.0.0</version.org.glassfish.soteria>
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.0</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
         <version.org.hibernate.commons.annotations>6.0.6.Final</version.org.hibernate.commons.annotations>
-        <version.org.hibernate>6.2.6.Final</version.org.hibernate>
+        <version.org.hibernate>6.2.8.Final</version.org.hibernate>
         <version.org.hibernate.search>6.2.1.Final</version.org.hibernate.search>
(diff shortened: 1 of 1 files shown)


Sanne:
@scottmarlow could we also upgrade ByteBuddy to match the version we build with?

It will improve some details regarding compatibility with Java 21.

scottmarlow:
> @scottmarlow could we also upgrade ByteBuddy to match the version we build with? We're using version `1.14.7` in this version of Hibernate ORM.
> 
> It will improve some details regarding compatibility with Java 21.

@Sanne yes, definitely!  I just requested that on https://github.com/wildfly/wildfly/pull/17061 that was upgrading to `1.14.5`, so if that bumps to `1.14.7`, I think we are good.

Sanne:
Ok that works for me, but considering ORM "owns" the Bytebuddy dependency (Last time I checked we were the only users), wouldn't it be simpler for you to regulalry update both at once in the same commit?

That way, should you need forwards/backwards ports, you'll also remember more easily that they need to go together.

As you wish though - just a suggestion ;)

scottmarlow:
> Ok that works for me, but considering ORM "owns" the Bytebuddy dependency (Last time I checked we were the only users), wouldn't it be simpler for you to regulalry update both at once in the same commit?
> 
> That way, should you need forwards/backwards ports, you'll also remember more easily that they need to go together.
> 
> As you wish though - just a suggestion ;)

Done, thanks!

bstansberry (approved):

PR #17141, opened 2023-09-01
[WFLY-18406] Upgrade ironjacamar to 3.0.6.Final

tadamski:
https://issues.redhat.com/browse/WFLY-18406

Upgrade because of refactor in IronJacamar which removes usage on jboss-threads blocking API


tadamski:
@bstansberry ok, I will make some notes about it in my docs

bstansberry:
@tadamski It looks like it was the GH actions that were affected by the artifact not being available when they ran. It looks like  the TeamCity jobs had lots of ordinary test failures.

PR #17145, opened 2023-09-01
WFCORE-6473 Remove references to obsolete remoting resource

pferraro:
https://issues.redhat.com/browse/WFCORE-6473

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <feature-group-spec name="remoting" xmlns="urn:jboss:galleon:feature-group:1.0">
-    <!-- TODO Temporary fork to override use of security-realms. -->
     <feature spec="subsystem.remoting">
         <param name="worker" value="default"/>
         <feature spec="subsystem.remoting.http-connector">
--- a/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml
@@ -2,9 +2,7 @@
 <feature-group-spec name="remoting" xmlns="urn:jboss:galleon:feature-group:1.0">
     <!-- TODO Temporary fork to override use of security-realms. -->
     <feature spec="subsystem.remoting">
-        <feature spec="subsystem.remoting.configuration.endpoint">
-            <unset param="worker"/>
-        </feature>
+        <param name="worker" value="default"/>
         <feature spec="subsystem.remoting.http-connector">
             <param name="http-connector" value="http-remoting-connector"/>
             <param name="connector-ref" value="default"/>
--- a/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/embedded-activemq.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/embedded-activemq.xml
@@ -84,8 +84,4 @@
             <param name="jms-connection-factory" value="java:jboss/DefaultJMSConnectionFactory"/>
         </feature>
     </feature>
-    <!-- remote naming -->
-    <feature spec="subsystem.naming">
-        <feature spec="subsystem.naming.service.remote-naming"/>
-    </feature>
 </feature-group-spec>
--- a/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/unsecured-basic-profile.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/feature_groups/unsecured-basic-profile.xml
@@ -12,9 +12,7 @@
     </feature>
 
     <feature-group name="remoting"/>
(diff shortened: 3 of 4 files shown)


bstansberry:
@fl4via FYI re this and other stuff like WFCORE-6472

ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remoting.xml
```diff
@@ -2,9 +2,7 @@
 <feature-group-spec name="remoting" xmlns="urn:jboss:galleon:feature-group:1.0">
     <!-- TODO Temporary fork to override use of security-realms. -->
```
bstansberry, line 3: This can be updated, as it's now a fork to add the http-connector.

pferraro: Updated.

ee-feature-pack/galleon-shared/src/main/resources/feature_groups/remote-naming.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <feature-group-spec name="remote-naming" xmlns="urn:jboss:galleon:feature-group:1.0">
+    <feature-group name="remoting"/>
```
bstansberry, line 3: I'm not sure about this aspect. Feature groups are just config chunks; layers are the things that have dependency relationships. This is a kind of a dependency relationship, and by duplicating the config chunk it runs the risk of falling out of sync with what the layers specify.

pferraro: There is no "remote-naming" layer - there is only a "remote-naming" feature group, which is not usable without remoting.  Do you suggest we create one?  Or do you mean to add the "remoting" feature group to every layer that uses the "remote-naming" feature group?

bstansberry: The latter. IOW, instead of the deletion at https://github.com/wildfly/wildfly/pull/17145/files#diff-d27c5509481591a1bc0a83645c0fc7f9e507ad6dfb360b3d7ddd6138bed5eff4L4-L11, replace it with a simplified reference to the remoting feature-group.

The two layers that use remote-naming FG already have layer dependencies on the remoting layer.

The remoting feature-group provides config elements that if used mean other things are required, but it doesn't include those in its own definition. So this is no different.

Probably the embedded-activemq feature-group shouldn't embed remote naming either; it should be included  in the embedded-activemq layer. AFAICT the only uses of that feature-group are in the layer.

I realize this is kind of a gray area and TBH I'm thinking it through as I write. But perhaps it's like this...

The bulk of feature-groups are just config chunks. They provide a specific set of configuration but don't attempt to include other reusable things that their own chunk needs.

Then we have layers, which depend on other layers for things they need, and add feature groups / features for other things they need not provided by their dependencies.

Finally we have a subset of feature groups that are basically used to support construction of the OOTB configs, particularly for domain.xml. These are kind of a special case where we allow composition in a feature group, which, unlike standalone.xml we can't generate from layers. But these are more of a specialized use case.

@jfdenise @ehsavoie FYI re this discussion.

ehsavoie: I agree with the fact that the feature groups should be used in a layer as opposed to used by another feature group as from a provisioning perspective le granularity level should be the layer

jfdenise: +1 @bstansberry , fg including other fg due to the requirement of layers doesn't help readability. Better to explicitly add multiple fg directly in the layers.

pferraro: > The latter. IOW, instead of the deletion at https://github.com/wildfly/wildfly/pull/17145/files#diff-d27c5509481591a1bc0a83645c0fc7f9e507ad6dfb360b3d7ddd6138bed5eff4L4-L11, replace it with a simplified reference to the remoting feature-group.
> 
> The two layers that use remote-naming FG already have layer dependencies on the remoting layer.
> 
> The remoting feature-group provides config elements that if used mean other things are required, but it doesn't include those in its own definition. So this is no different.
> 
> Probably the embedded-activemq feature-group shouldn't embed remote naming either; it should be included in the embedded-activemq layer. AFAICT the only uses of that feature-group are in the layer.

Understood now.  Will fix.

pferraro: Fixed.

bstansberry: Thanks, @pferraro. I'll have a look.

FYI, Paul and @ehsavoie and @jfdenise this discussion sparked a thought so I filed https://issues.redhat.com/browse/WFLY-18440.

bstansberry:
I'm not going to formally approve this as I want @fl4via to approve it and someone may merge it if I've approved, but all this LGTM, @pferraro.

fl4via (approved):

PR #17146, opened 2023-09-04
[WFLY-217] Strict location check of persistence.xml deployments

ChristinaDsl:
https://issues.redhat.com/browse/WFLY-217

https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a12236

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
@@ -44,6 +44,7 @@ import org.jboss.as.server.deployment.module.ResourceRoot;
 import org.jboss.metadata.parser.util.NoopXMLResolver;
 import org.jboss.vfs.VirtualFile;
 import org.jipijapa.plugin.spi.PersistenceUnitMetadata;
+import org.jboss.as.weld.logging.WeldLogger;
 
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLStreamReader;
@@ -175,6 +176,9 @@ public class PersistenceUnitParseProcessor implements DeploymentUnitProcessor {
             VirtualFile persistence_xml = deploymentRoot.getRoot().getChild(META_INF_PERSISTENCE_XML);
             parse(persistence_xml, listPUHolders, deploymentUnit);
             PersistenceUnitMetadataHolder holder = normalize(listPUHolders);
+            if (!holder.equals(null) && !holder.equals("")) {
+                WeldLogger.ROOT_LOGGER.couldNotLoadPersistenceXmlFromInvalidLocation(persistence_xml.getPathName());
+            }
             deploymentRoot.putAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS, holder);
             addApplicationDependenciesOnProvider( deploymentUnit, holder);
             markDU(holder, deploymentUnit);
--- a/weld/common/src/main/java/org/jboss/as/weld/logging/WeldLogger.java
+++ b/weld/common/src/main/java/org/jboss/as/weld/logging/WeldLogger.java
@@ -266,4 +266,10 @@ public interface WeldLogger extends BasicLogger {
 
     @Message(id = 58, value = "Persistence unit '%s' removed.")
     IllegalStateException persistenceUnitRemoved(String scopedPuName);
+
+    @LogMessage(level = Logger.Level.WARN)
+    @Message(id = 59, value = "Persistence units are defined in the '%s' file, " +
+            "which is not a designated location for persistence.xml files in the Jakarta Persistence specification. " +
+            "Support for such non-standard locations may be removed in a future release.")
+    void couldNotLoadPersistenceXmlFromInvalidLocation(Object persistenceXmlPath);
 }


jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
```diff
@@ -76,7 +76,7 @@
  */
 public class PersistenceUnitParseProcessor implements DeploymentUnitProcessor {
 
-    private static final String WEB_PERSISTENCE_XML = "WEB-INF/classes/META-INF/persistence.xml";
+    private static final String WEB_PERSISTENCE_XML = "WEB-INF/classes/persistence.xml";
```
bstansberry, line 79: Per https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#_footnotedef_87 the existing code is correct.

weld/common/src/main/java/org/jboss/as/weld/logging/WeldLogger.java
```diff
@@ -266,4 +266,7 @@ public interface WeldLogger extends BasicLogger {
 
     @Message(id = 58, value = "Persistence unit '%s' removed.")
     IllegalStateException persistenceUnitRemoved(String scopedPuName);
+
+    @Message(id = 59, value = "Could not load peristence.xml due to its invalid location")
```
bstansberry, line 270: Typo

s/peristence/persistence

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -134,6 +134,9 @@ private String getScopedPUName(final DeploymentUnit deploymentUnit, final String
         PersistenceUnitMetadata scopedPu;
         scopedPu = PersistenceUnitSearch.resolvePersistenceUnitSupplier(deploymentUnit, persistenceUnitName);
         if (null == scopedPu) {
+            if (persistenceUnitName.equals("")) {
+                throw WeldLogger.ROOT_LOGGER.couldNotLoadPersistenceXmlFromInvalidLocation();
```
bstansberry, line 138: Do we know this means an invalid location? Can we end up here just because there's no persistence.xml at all?

bstansberry (changes requested):
Thanks @ChristinaDsl.

I've made some specific notes, but overall @scottmarlow will need to approve this.

Scott, AIUI once my comment re WEB_PERSISTENCE_XML is addressed, the change here is to ignore a persistence.xml in an ear's own META-INF dir.

1) Do you want this breaking change still? The discussion on WFLY-217 is quite old and may predate any release where we became as strict about breaking changes as we are now.

2) I think if we do ignore that location we should check it and log a WARN if found. Then if things fail the WARN will be helpful.

scottmarlow:
> Thanks @ChristinaDsl.
> 
> I've made some specific notes, but overall @scottmarlow will need to approve this.
> 
> Scott, AIUI once my comment re WEB_PERSISTENCE_XML is addressed, the change here is to ignore a persistence.xml in an ear's own META-INF dir.
> 
>     1. Do you want this breaking change still? The discussion on WFLY-217 is quite old and may predate any release where we became as strict about breaking changes as we are now.

Regarding making this a breaking change, we mean that the current pull request breaks current applications that just work.

> 
>     2. I think if we do ignore that location we should check it and log a WARN if found. Then if things fail the WARN will be helpful.

I agree that we should log a warning instead of throwing an (deployment) exception.  The difference is that the logged warning only shows up in the app server console (e.g. server.log file as well) but a thrown exception causes the application deployment to fail with the thrown exception which means the application must be corrected before it can deploy.  Since such applications have been working for several years now, I think logging the warning is a better approach now.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
ChristinaDsl: Ok, I will change this part

weld/common/src/main/java/org/jboss/as/weld/logging/WeldLogger.java
ChristinaDsl: You 're right, I 'll make this one too

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
ChristinaDsl: It is for both cases. Wrong location and no persistence.xml at all..
I will make it only for wrong location

ChristinaDsl: It's better to make two different cases: The first one checks if not exists at all and throw warning, the second one checks if exist in wrong location and also throws warning?
OR 
Make one case which cover both of them and warn the developer that "perstistence.xml not exist or its location is wrong"
?

ChristinaDsl: Ok, thank you, I will make it this way

bstansberry:
This PR hasn't had activity for a long time and I don't think the current state is correct. PersistenceUnitMetadataHolder.equals("") will always return false. If it would return true I don't think code in that location is relevant to the WFLY-217 issue, as that block is limited to jars in the ear's lib folder and per the WFLY-217 description such. location is valid. AIUI the problematic location is the META-INF dir of the ear itself.

Please feel free to reopen.

PR #17148, opened 2023-09-05
[WFLY-18332] Update LayersTestBase and LayersTestCase(s)

bstansberry:
https://issues.redhat.com/browse/WFLY-18332

Refactor to allow more understandable maintenance of the expected sets of modules that are unreferenced or not used in layers.

Update the expected sets to reflect current reality.

This update is a prerequisite for completing https://issues.redhat.com/browse/WFCORE-6456 as it corrects all the broken things found by the testing I'm adding for that.

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
@@ -205,6 +205,10 @@ public abstract class LayersTestBase {
             "org.wildfly.extension.microprofile.reactive-streams-operators-smallrye",
             "org.wildfly.reactive.mutiny.reactive-streams-operators.cdi-provider",
             "io.vertx.client",
+            // Extension not included in the default config
+            "org.wildfly.extension.microprofile.metrics-smallrye",
+            // Extension not included in the default config
+            "org.wildfly.extension.microprofile.opentracing-smallrye",
             // Injected by jaxrs subsystem
             "org.jboss.resteasy.microprofile.config",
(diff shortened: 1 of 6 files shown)


bstansberry:
I added the hold label as there's more I want to do on this but it's been a while since I've had time to finish it off.

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -179,16 +205,115 @@ public class LayersTestBase {
...
+     *
+     * @param first the first array to combine. Cannot be {@code null}
+     * @param others other arrays to combine. Can be {@code null}
+     * @return a set containing all of the elements in the arrays
+     */
+    @SuppressWarnings("SameParameterValue")
+    protected static Set<String> concatArrays(String[] first, String[]... others) {
+        // TODO Delegate to LayersTest.concatArrays once WFCORE-6481 is integrated
+        if (others == null || others.length == 0) {
+            return new HashSet<>(Arrays.asList(first));
```
jamezp, line 308: Really not important, but we could use `Set.of(first)`.

jamezp (approved):
LGTM

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
bstansberry: Good point; old habits die hard. I'll do that in the wf core code that will replace this.

bstansberry:
Thanks @jamezp!

PR #17151, opened 2023-09-06
WIP WFLY-18414 Create a standard-test-expansion bom

TomasHofman:
https://issues.redhat.com/browse/WFLY-18414


bstansberry (approved):

bstansberry:
Thanks @TomasHofman

TomasHofman:
You were faster than me, I put this up with "WIP" to first see all the CI jobs pass :).

But I think the only change I would make is removing the jaxen dependency that I commented out. I can do that in a subsequent PR.

Regarding the dependency version properties for XP test deps, are we OK leaving them in the main POM or should I move them straight into the new BOM?

TomasHofman:
I created https://github.com/wildfly/wildfly/pull/17160

PR #17153, opened 2023-09-06
Fix for WFLY-18430 and WFLY-18790

jfdenise:
Issues: 
* https://issues.redhat.com/browse/WFLY-18430
* https://issues.redhat.com/browse/WFLY-18790


bstansberry:
Hey @jfdenise -- it's great to see this. Apologies as I may be a while before I can do a deep dive on this; it's not a 15 minute review!

One general comment from a quick scan is it would be good in the existing classes where we use GlowUtil to disable stuff to have comments explaining to some degree why. Over time that kind of thing will be critical to helping maintainers test correctly with Glow.

jfdenise:
@bstansberry , I have opened https://github.com/wildfly/wildfly/pull/17181 that is a subset of this Draft PR. It contains only fixes in layers metadata and layers metadata unit tests.  So no change in WildFly integration tests.

It would be great if we could have this in for WF 30.

bstansberry:
@jfdenise When playing with my WF Core stuff I realized we need to add 'overwriteProvisionedServer' to all these wildfly-maven-plugin provisioning configs. Otherwise you have to use mvn clean to get an updated provisioning if you build a module twice.  IIRC galleon-maven-plugin always re-provisions, so we should get the configure the same behavior with wildfly-maven-plugin.

I'll update https://github.com/wildfly/wildfly-core/pull/5769 tomorrow to fix that in core.

jfdenise:
@bstansberry , thank-you. Yes, I noticed that. I am currently not using it prior to reach a full green run. It could hide some wrong phase orders (eg: some resources copied to target/wildfly dir that should happen after provisioning). When all is green, I will add the 'overwriteProvisionedServer' option.

bstansberry:
@jfdenise WDYT about leaving testsuite/layers and testsuite/layers-expansion alone with this? Moving *all* provisioning (except the build/dist modules) to wildfly-maven-plugin costs us our intended test coverage of the galleon-maven plugin.

The benefit of moving to the channel is we'd detect if some updated version in a channel caused one of the many LayersTestCase installations to fail to boot. That's not really the original purpose of testing those installations, but there is *some* value there.

IOW, please regard this comment as an invitation to discuss how we'll test galleon-maven plugin now that we are dropping so much of its use. :)

jfdenise:
@bstansberry , ok to revert the changes in layers testsuite. That will keep a good coverage of the Galleon maven-plugin: layers provisioning and default configuration provisioning.
Question: Why the build/dist modules will be not using wildfly-maven-plugin + channels?

jfdenise:
@bstansberry , I reverted the changes for layers and layers-expansion.

bstansberry:
> Question: Why the build/dist modules will be not using wildfly-maven-plugin + channels?

@jfdenise Maybe they will at some point but that is out of scope for the testing work I'm trying to get done now in a short period of time. So i'm just putting that to the side. My immediate goal is being able to test an externally managed channel. Even https://issues.redhat.com/browse/WFCORE-6615 is kind of out-of-scope, but it was simple to do so I included it in https://github.com/wildfly/wildfly-core/pull/5777. But if that commit would lead to delays in the rest I'd drop it for now.

bstansberry:
> @bstansberry , ok to revert the changes in layers testsuite. That will keep a good coverage of the Galleon maven-plugin: layers provisioning and default configuration provisioning.

And feature pack creation. :)

jfdenise:
@bstansberry , for the feature-pack creation I think we are now good, The new CI that tests galleon-plugins with WildFly SNAPSHOT is in place: https://github.com/wildfly/galleon-plugins/pull/271

testsuite/integration/basic/pom.xml
```diff
@@ -1136,17 +1448,7 @@
                                 <jboss-maven-dist/>
                                 <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
                             </plugin-options>
-                            <feature-packs>
-                                <feature-pack>
-                                    <groupId>${testsuite.ee.galleon.pack.groupId}</groupId>
-                                    <artifactId>${testsuite.ee.galleon.pack.artifactId}</artifactId>
-                                    <version>${testsuite.ee.galleon.pack.version}</version>
```
bstansberry, line 1143: I could be wrong but I think the behavior provided by these 3 expressions is being lost.

The goal here of them is to allow control of whether the wildfly-ee or wildfly feature pack is used for provisioning. Standard test runs use the default, which is the wildfly FP. But by setting these props, nightly CI jobs can force the use of wildfly-ee, and we can thus track that that kind of provisioning results in installations that pass the tests.

https://ci.wildfly.org/viewType.html?buildTypeId=WF_MainNightlyJobs_GalleonAltFpLinuxJdk11 does this.

bstansberry: Maybe just including another feature pack here will do the trick:

https://github.com/wildfly/wildfly/pull/17153/commits/9f649d7f806acd24d12766ebd89b05932a3be244#diff-100b3a1c5caf72adb576b1434ecc9d3f53b284e79cd07baea4b2fcba616730ffR1137-R1143

If the properties are specifying wildfly-ee then it's redundant as wildfly-test-feature-pack depends on wildfly-ee. But if the props specify wildfly FP then we are bringing that FP into the provisioning.

bstansberry:
@jfdenise I made some review comments above. Is this ready to move out of draft state? It would be good to be able to merge it as soon as it's approved and passes CI.

jfdenise:
@bstansberry , I commented the usage of GlowUtil, all other comments implemented. Moved to ready for review (resynced with main right now).

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/InfinispanServerUtil.java
```diff
@@ -36,10 +37,13 @@ public class InfinispanServerUtil {
...
-                .property("infinispan.client.rest.auth_username", "testsuite-driver-user")
+        InfinispanServerRuleBuilder builder = InfinispanServerRuleBuilder
+                .config(absoluteConfigurationFile);
+        // When WildFly Glow is instantiating the deployment, we don't want to start any server.
+        if (!GlowUtil.isGlowScan()) {
+            builder.property(TestSystemPropertyNames.INFINISPAN_TEST_SERVER_DIR, INFINISPAN_SERVER_HOME);
+        }
+        INFINISPAN_SERVER_RULE = builder.property("infinispan.client.rest.auth_username", "testsuite-driver-user")
                 .property("infinispan.client.rest.auth_password", "testsuite-driver-password")
                 .numServers(1)
```
pferraro, line 48: Better to use:
```java
builder.numServers(GlowUtil.isGlowScan() ? 0 : 1);`

pferraro (changes requested):

jfdenise:
@bstansberry , coud you tell me what is the expected fix for imports? BTW, I found some duplicated resources in tests (wrongly added by this PR) that I removed in an extra commit.

bstansberry:
@jfdenise Order is:

import static

import java
import javax

import other

bstansberry (approved):

PR #17154, opened 2023-09-06
[WFLY-18403] making client-ssl-context and server-ssl-context indepen…

istudens:
…dent from each other

https://issues.redhat.com/browse/WFLY-18403


tadamski (commented):
I think we also need to take into cosideration the runtime behavior. I have created alternative PR https://github.com/wildfly/wildfly/pull/17211 which tries to do that and split the client and server SSL config - @istudens could you please take a look?

istudens:
Superseded by https://github.com/wildfly/wildfly/pull/17211

PR #17168, opened 2023-09-11
[WFLY-18429] Duplicated classes

marcosgopen:
https://issues.redhat.com/browse/WFLY-18429

Duplicated classes in narayana-jts-integration-6.0.1.Final-redhat-00001.jar and wildfly-transactions-8.0.0.GA-redhat-20230811.jar
Interface already present in dependency named 'narayana-jts-integration'


mmusgrov (approved):
As the PR description implies, the ejb3 and transactions sub-subsystems both depend on the maven artifact `org.jboss.narayana.jts:narayana-jts-integration` which contains the duplicated class, therefore it is safe to remove it from the transactions subsystem.  At some point in the future we reserve the right to add it to an SPI.

bstansberry:
Thanks @marcosgopen and @mmusgrov

PR #17176, opened 2023-09-13
[WFLY-18445] Remove XP transformer tests for Microprofile extensions

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-18445

### DIFF
--- a/microprofile/config-smallrye/src/test/java/org/wildfly/extension/microprofile/config/smallrye/MPConfigSubsystemParsingTestCase.java
+++ b/microprofile/config-smallrye/src/test/java/org/wildfly/extension/microprofile/config/smallrye/MPConfigSubsystemParsingTestCase.java
@@ -22,35 +22,16 @@
 
 package org.wildfly.extension.microprofile.config.smallrye;
 
-import static org.jboss.as.weld.Capabilities.WELD_CAPABILITY_NAME;
-import static org.junit.Assert.assertTrue;
-import static org.wildfly.extension.microprofile.config.smallrye.MicroProfileConfigExtension.CONFIG_SOURCE_PATH;
-import static org.wildfly.extension.microprofile.config.smallrye.MicroProfileConfigExtension.SUBSYSTEM_PATH;
-import static org.wildfly.extension.microprofile.config.smallrye.MicroProfileConfigExtension.VERSION_1_0_0;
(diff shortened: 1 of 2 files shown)


bstansberry:
@yersan This is a valid test overall that should be retained. Using the EAP 7.4 controller as a convenience for "something before WF 26 when we changed the subsystem" doesn't mean the test is incorrect and should be removed.

WildFly supports mixed domains back to WF 23. This includes all extensions we ship that work in domain mode.

yersan:
Hi @bstansberry , ok, got it, I will close this one. It even could deserve some discussion via a different channel (Zulip Email). The original idea with this was to try to get rid of any kind of maintenance of any EAP XP based model controller in WildFly Legacy Tests, after all, MP extensions are not supported on an Expansion Pack. 

Probably the approach for this after reading your feedback could have been to keep WildFly specific Model Controllers instead of migrating to EAP based ones at https://github.com/wildfly/wildfly-core/pull/5640 (with the consequent WildFly Legacy Tests maintenance)

We have already done the WildFly Legacy Tests though but I guess we do not need to maintain them anymore for future XP versions.

yersan:
Just to be more clear:

> Probably the approach for this after reading your feedback could have been to keep WildFly specific Model Controllers instead of migrating to EAP based ones at https://github.com/wildfly/wildfly-core/pull/5640 (with the consequent WildFly Legacy Tests maintenance)

Indeed, this seems to be the correct approach, rollback wildfly/wildfly-core#5640 to keep using WF based Model Controllers and close https://github.com/wildfly/wildfly-legacy-test/pull/31. We should rename the Model Controller enum from EAP_XP_4 to WF26 to clarify it is a WF-based model controller instead of an EAP one.

kabir:
@yersan as discussed in chat. How about:
- deprecating ModelTestControllerVersion.EAP_XP_4 in wildfly core. In case external users are using this for whatever weird reason!
- Moving the test here to use the WildFly 26 ModelTestControllerVersion.
- Dropping this test in the XP branch

yersan:
@kabir I agree with all, but without dropping this test in the XP branch if it can continue using the WF26 Model Controller, just to avoid the additional maintenance of XP branch

bstansberry:
@yersan I'm going to close this and the JIRA since as discussed here we don't intend to remove these. Changing the tests can be done under a different JIRA.

PR #17180, opened 2023-09-15
[WFLY-18451] Increase JVM heap size for tests in testsuite/domain

gaol:
Issue: https://issues.redhat.com/browse/WFLY-18451


yersan (approved):

bstansberry:
Thanks @gaol and @yersan

PR #17181, opened 2023-09-15
Fix for WFLY-18436, Layers metadata requires unit testing

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-18436

Contains Layers metadata fixes and unit tests.

### DIFF
--- a/galleon-pack/layer-metadata-tests/src/test/resources/glow/latest/provisioning-bare-metal.xml
+++ b/galleon-pack/layer-metadata-tests/src/test/resources/glow/latest/provisioning-bare-metal.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" ?>
+
+<installation xmlns="urn:jboss:galleon:provisioning:3.0">
+    <feature-pack location="${full.maven.groupId}:wildfly-galleon-pack:${project.version}"/>
+</installation>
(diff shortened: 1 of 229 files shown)


galleon-pack/layer-metadata-tests/src/test/resources/glow/latest/provisioning-bare-metal.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" ?>
+
+<installation xmlns="urn:jboss:galleon:provisioning:3.0">
+    <feature-pack location="${ee.maven.groupId}:wildfly-galleon-pack:${project.version}"/>
```
bstansberry, line 4: This should be full.maven.groupId.

bstansberry (changes requested):
Thanks @jfdenise -- I just noticed one small thing.

galleon-pack/layer-metadata-tests/src/test/resources/glow/latest/provisioning-bare-metal.xml
jfdenise: @bstansberry , good catch! Thank-you, fixed.

bstansberry (approved):

PR #17190, opened 2023-09-15
[WFLY-18452] Add Guava as a test dependency.

ivassile:
Issue: https://issues.redhat.com/browse/WFLY-18452


bstansberry:
Thanks @ivassile

PR #17201, opened 2023-09-20
[WFLY-18526] Moves BOMs to the Server repository

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18526


boms/user/ee/ee-with-tools/pom.xml
```diff
@@ -0,0 +1,389 @@
...
+    </parent>
+
+    <packaging>pom</packaging>
+
+    <artifactId>wildfly-ee-with-tools-builder</artifactId>
+
+    <name>WildFly BOMs: EE with Tools Builder</name>
+    <description>This artifact builds a bill of materials (BOM) for WildFly EE with Deployment and Testing Tools</description>
+
+    <properties>
```
emmartins, line 42: I don't like this here, but I also don't like this BOM at all due to adding artifacts we do not manage and/or support... IMHO we should consider to remove this BOM now!

boms/user/README.adoc
```diff
@@ -0,0 +1,31 @@
...
+----
+<dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>org.wildfly.bom</groupId>
+            <artifactId>wildfly-ee-with-tools</artifactId>
+            <scope>import</scope>
+            <type>pom</type>
+            <version>${wildfly.version}</version>
+        </dependency>
```
jbliznak, line 28: this is just a presentation variant without affecting functionality but it seems to me this form is more commonly used. After all, it is used like that in the code of this very PR.
```suggestion
        <dependency>
            <groupId>org.wildfly.bom</groupId>
            <artifactId>wildfly-ee-with-tools</artifactId>
            <version>${wildfly.version}</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
```

jbliznak: I can see it is also in original https://github.com/wildfly/boms/blob/main/README.adoc so take it just as a nice to have :)

boms/user/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+        <version>30.0.0.Beta1-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <groupId>org.wildfly.bom</groupId>
+    <artifactId>wildfly</artifactId>
+
+    <packaging>pom</packaging>
+
+    <name>WildFly BOMs: Parent Aggregator</name>
```
jbliznak, line 42: We should really try hard to change this name and description.
More so that in contrast to other modules at `/boms`, this new one doesn't contain BOMs per se, it contains bom-builder projects so its name could be misleading.

jbliznak:
The proposal https://github.com/wildfly/wildfly-proposals/pull/539/ mentions `The BOMs sources should be moved to the Server repository, and placed in a new /boms/user directory. The BOMs sources repository history should be included in such move.`

It doesn't seem this "keep history" requirement is included in this PR (yet?).

boms/user/pom.xml
emmartins: I would prefer to not change this for now, it's the WildFly BOMs project that may be found in this directory (as stated by the README too), being this the parent aggregator module. If you look down the tree then builders are identified as such in name and artifactId too.

boms/user/README.adoc
emmartins: change done, thanks

emmartins:
> The proposal [wildfly/wildfly-proposals#539](https://github.com/wildfly/wildfly-proposals/pull/539) mentions `The BOMs sources should be moved to the Server repository, and placed in a new /boms/user directory. The BOMs sources repository history should be included in such move.`
> 
> It doesn't seem this "keep history" requirement is included in this PR (yet?).

History should be present now, previously somehow I got ridden of it on a rebase, thanks for the warning.

boms/user/pom.xml
jbliznak: fair enough, my point stands, but leaving it up to other reviewers then.

bstansberry: I think it's a fair point but I'd like to leave it for later.

With this our names will be one of "WildFly: xxx" "WildFly Preview: xxx" or "WildFly BOMs: xxx". Those all relate to different end-user consumables, so it's a reasonable categorization.

I wouldn't want to change the name of the other bom modules to "WildFly BOMs: xxx" as that means all sorts of other groupings are ok too, e.g. "WildFly Clustering: xxx". And that will just become a mess (as it was until sometime in the last year or so when we did some housekeeping.)

bstansberry (approved):
There are some tweaks I'd like to see, but I'd like to get this in today, so I'll merge it and send up a couple PRs with my tweaks.

One will remove the ee-with-tools module, since AIUI both @emmartins and @jbliznak indicated wanting to see it go away. So we should do that for 30 Beta. If we change our mind we can always add it back later, but if we don't remove it now it needs to remain for 30 Final.

bstansberry:
I ended up needing to revert this as I hadn't noticed that the new modules weren't included in the build. Including them, via #17224 resulted in CI failures that were not easy to understand.

emmartins:
looking at it

PR #17202, opened 2023-09-21
[WFLY-17309]: Upgrade Apache Artemis to 2.31.0 

ehsavoie:
* Upgrading to Apache Artemis 2.31.0
* Upgrading Artemis native to 2.0.0
* Upgrading wildfly-artemis-integration to 2.0.2.Final
* Fixing json response changes in Artemis 2.28
* Defining legacy behavior on paging through AddressSettings
* Upgrading messaging subystem version to 16.0 to prepare for management model changes
* Adding a new *max-read-page-bytes* attribute to be able to configure Artemis accordingly.

Issue:
https://issues.redhat.com/browse/WFLY-17309
https://issues.redhat.com/browse/WFLY-17700
https://issues.redhat.com/browse/WFLY-18004
https://issues.redhat.com/browse/WFLY-18000


ehsavoie:
@bstansberry the 2 failures aren't related

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingDefinition.java
```diff
@@ -124,6 +124,13 @@ public class AddressSettingDefinition extends PersistentResourceDefinition {
             .setAllowExpression(true)
             .build();
 
+    public static final SimpleAttributeDefinition MAX_READ_PAGE_BYTES = create("max-read-page-bytes", ModelType.INT)
+            .setDefaultValue(new ModelNode(-1))
```
bstansberry, line 128: Is this the right default, versus, say, no default? Artemis defaults to 2 * PAGE_SIZE_BYES.

Using the legacy value -1 saves us a potential server config migration issue, but at the cost of being stuck with -1 for a long time.

BTW the handling of default values for a number of attributes (including PAGE_SIZE_BYES) is incorrect in AddressSettings.createSettings. It checks if the relevant ModelNode is defined and if not it ignores the default.

bstansberry: Note that I don't pretend to understand all this, but does https://github.com/apache/activemq-artemis/pull/4552/files address one of the concerns with not using -1, i.e. that the server might silently stop processing?

ehsavoie: 1. I agree we can do this but that will require users to set a proper value or they might meet with https://issues.redhat.com/browse/WFLY-17700
2. I've created https://issues.redhat.com/browse/WFLY-18547 to track that

ehsavoie:
@mnovak1 WDYT about the default value ? Do you agree with this PR ?

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingDefinition.java
bstansberry: Ok, thanks. We talked and I'm ok with this or with some other default or even no default (which you don't think is a good idea, which is fine.)

Note also that we can tweak this between now and 30 Final in 2 weeks. But we should be get some form of this in by tomorrow or it should likely wait for WF 31.

bstansberry:
Thanks @ehsavoie

PR #17211, opened 2023-09-25
[WFLY-18403] Making client-ssl-context and server-ssl-context independent

tadamski:
https://issues.redhat.com/browse/WFLY-18403

This is an alternative to https://github.com/wildfly/wildfly/pull/17154 

@istudens WDYT?

### DIFF
--- a/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
+++ b/iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
@@ -27,14 +27,17 @@ public class ConfigValidator {
         validateSocketBindings(context, resourceModel);
 
         final boolean supportSSL = IIOPRootDefinition.SUPPORT_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
-        final boolean serverRequiresSsl = IIOPRootDefinition.SERVER_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
-        final boolean clientRequiresSsl = IIOPRootDefinition.CLIENT_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
+        final boolean serverRequiresSSL = IIOPRootDefinition.SERVER_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
+        final boolean clientRequiresSSL = IIOPRootDefinition.CLIENT_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
 
-        final boolean sslConfigured = isSSLConfigured(context, resourceModel);
+        final boolean securityDomainConfigured = IIOPRootDefinition.SECURITY_DOMAIN.resolveModelAttribute(context, resourceModel).isDefined();
 
-        validateSSLConfig(supportSSL, sslConfigured, serverRequiresSsl, clientRequiresSsl);
-        validateSSLSocketBinding(context, resourceModel, sslConfigured, warnings);
-        validateIORTransportConfig(context, resourceModel, supportSSL, serverRequiresSsl, warnings);
+        final boolean serverSSLConfigured = IIOPRootDefinition.SERVER_SSL_CONTEXT.resolveModelAttribute(context, resourceModel).isDefined() || securityDomainConfigured;
+        final boolean clientSSLConfigured = IIOPRootDefinition.CLIENT_SSL_CONTEXT.resolveModelAttribute(context, resourceModel).isDefined() || securityDomainConfigured;
+
+        validateSSLConfig(supportSSL, serverSSLConfigured, clientSSLConfigured, serverRequiresSSL, clientRequiresSSL);
+        validateSSLSocketBinding(context, resourceModel, serverSSLConfigured, clientSSLConfigured, warnings);
+        validateIORTransportConfig(context, resourceModel, supportSSL, serverRequiresSSL, warnings);
         validateORBInitializerConfig(context, resourceModel);
 
         return warnings;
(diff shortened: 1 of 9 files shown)


istudens:
@tadamski This looks good to me.

soul2zimate:
This change blocks https://issues.redhat.com/browse/CMTOOL-351 which is a downstream dependency.

Hi @bstansberry, since this is reviewed a long time ago, would you consider merging it?

darranl:
I am not sure if we can make that change without a model version bump, it looks like this could have transformer implications as well.

darranl:
The question here is not so much about the issue type, rather do the required changes constitute a model change.  The issue type does not negate the need for model updates to be handled correctly.

If previously valid configurations could now be broken with this change and if we can not operate in mixed domain due to different validation requirements additional consideration would be needed.

emmartins:
Please note that this is needed for https://issues.redhat.com/browse/CMTOOL-351

tadamski:
@darranl I started the refactor that you suggested (upgrade schema and implement transformers - new commit in this PR) but I think I cannot simply reject/discard attributes - I think I need only to reject as one of the attributes client-ssl-context/server-ssl-context is set to true and other is set to false as this is the configuration that is going to fail in previous schemas. The only thing in the API that seems a solution for this may be a customer transformer so I tried to implement it but it never gets executed. I have added a mock reject (in the commit) to verify that the registrations are OK and the reject handler is executed. Is nonexecuting CustomTransformer a bug or am I doing something wrongly?

emmartins:
Is there any news or plan about getting this done?

bstansberry:
@tadamski I sent https://github.com/tadamski/wildfly/pull/5 to you to fix the problems with this. It passes CI. If it looks good please squash it into your commit and push this branch.

ResourceTransformationDescriptionBuilder.setCustomResourceTransformer only does resource transformation, and not operation transformation. So it can work for when a legacy host connects to the Domain Controller and asks for all the resources, but it would not properly deal with a case where the hosts are already connected but with no iiop-subsystem in the config and then the user does a /subsystem=iiop-openjdk:add. Perhaps a corner case, but, still, it should work.

But a custom RejectAttributeChecker can work and that's what I did. A RejectAttributeChecker can get access to the resource's current model and use that do checks and reject if not ok.

I believe you didn't see the transformer you passed to setCustomResourceTransformer getting called because of how the transformer test infrastructure works. It creates the 'main' KernelServices for the current version. To start the 'legacy' kernel services it works out all the ops to create the 'main' config, transforms the *operations* and then uses those transformed ops to configure the 'legacy' one. But, as I said above what's passed tp ResourceTransformationDescriptionBuilder.setCustomResourceTransformer is only used for resource transformation, and not operation transformation. So it wouldn't get called. There are other workflows where it would get called, but I assume the test failed before it got to those.

Let's try and get this merged for 39 Beta and get one more thing out of the PR queue. Please ping me in Zulip when it is ready.

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
```diff
@@ -54,6 +55,7 @@ public void initializeParsers(ExtensionParsingContext context) {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME,Namespace.IIOP_OPENJDK_2_0.getUriString(), IIOPSubsystemParser_2_0::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME,Namespace.IIOP_OPENJDK_2_1.getUriString(), IIOPSubsystemParser_2_1::new);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME,Namespace.IIOP_OPENJDK_3_0.getUriString(), IIOPSubsystemParser_3_0::new);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME,Namespace.IIOP_OPENJDK_3_1.getUriString(), IIOPSubsystemParser_3_1::new);
```
rhusar: Why 3.1? We always bump major schema version in WildFly for each release. Not sure why this violates that rule. Is that's not the case any more @bstansberry ?

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
```diff
@@ -34,7 +34,8 @@ public class IIOPExtension implements Extension {
 
     static final ParentResourceDescriptionResolver SUBSYSTEM_RESOLVER = new SubsystemResourceDescriptionResolver(SUBSYSTEM_NAME, IIOPExtension.class);
 
-    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(3);
+    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(3,1);
```
rhusar: nitpick: run formatter (missing space)

rhusar (changes requested):

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
bstansberry: I don't care. I want this done and closed. :)

But yes, it should be 4.0.  The 'connector' modules have been our traditional violators about this.

rhusar: If you say so... 🫤 but this has been opened since Sep 25, 2023 so I do get that :)

My take on that is that violations like this proliferate, as others learn from other subsystems creating more confusion.

rhusar: Nitpick: run formatter on this block, missing space...

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
tadamski: fixed

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
tadamski: I have updated the PR so that the version is 4.0 now

tadamski: fixed

tadamski:
@bstansberry Thanks for transformer fix - I have merged it into this PR

@rhusar I have fixed formatting bugs and changed the schema to 4.0

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
rhusar: Thanks! :)

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/IIOPExtension.java
rhusar: 👍🏼 resolving conversation

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
```diff
@@ -27,14 +27,17 @@ public static List<String> validateConfig(final OperationContext context, final
...
+        final boolean serverRequiresSSL = IIOPRootDefinition.SERVER_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
+        final boolean clientRequiresSSL = IIOPRootDefinition.CLIENT_REQUIRES_SSL.resolveModelAttribute(context, resourceModel).asBoolean();
 
-        final boolean sslConfigured = isSSLConfigured(context, resourceModel);
+        final boolean securityDomainConfigured = IIOPRootDefinition.SECURITY_DOMAIN.resolveModelAttribute(context, resourceModel).isDefined();
 
-        validateSSLConfig(supportSSL, sslConfigured, serverRequiresSsl, clientRequiresSsl);
-        validateSSLSocketBinding(context, resourceModel, sslConfigured, warnings);
-        validateIORTransportConfig(context, resourceModel, supportSSL, serverRequiresSsl, warnings);
+        final boolean severSSLConfigured = IIOPRootDefinition.SERVER_SSL_CONTEXT.resolveModelAttribute(context, resourceModel).isDefined() || securityDomainConfigured;
```
rhusar:
```suggestion
        final boolean serverSSLConfigured = IIOPRootDefinition.SERVER_SSL_CONTEXT.resolveModelAttribute(context, resourceModel).isDefined() || securityDomainConfigured;
```

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
```diff
@@ -27,14 +27,17 @@ public static List<String> validateConfig(final OperationContext context, final
...
+        final boolean clientSSLConfigured = IIOPRootDefinition.CLIENT_SSL_CONTEXT.resolveModelAttribute(context, resourceModel).isDefined() || securityDomainConfigured;
+
+        validateSSLConfig(supportSSL, severSSLConfigured, clientSSLConfigured, serverRequiresSSL, clientRequiresSSL);
```
rhusar:
```suggestion
        validateSSLConfig(supportSSL, serverSSLConfigured, clientSSLConfigured, serverRequiresSSL, clientRequiresSSL);
```

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
```diff
@@ -27,14 +27,17 @@ public static List<String> validateConfig(final OperationContext context, final
...
+        validateSSLSocketBinding(context, resourceModel, severSSLConfigured, clientSSLConfigured, warnings);
```
rhusar:
```suggestion
        validateSSLSocketBinding(context, resourceModel, serverSSLConfigured, clientSSLConfigured, warnings);
```

rhusar (changes requested):
Just a small typo, otherwise I think this looks good.

iiop-openjdk/src/main/java/org/wildfly/iiop/openjdk/ConfigValidator.java
tadamski: fixed

rhusar (commented):
Hm, one last thing – do we really need a new schema? The schemas are identical. It appears to me if we are literally not changing anything in the schema, there is no need to introduce one. Just a model version bump.

```xml
[rhusar@puglia schema]$ diff wildfly-iiop-openjdk_3_0.xsd wildfly-iiop-openjdk_4_0.xsd
8,9c8,9
<            targetNamespace="urn:jboss:domain:iiop-openjdk:3.0"
<            xmlns="urn:jboss:domain:iiop-openjdk:3.0"
---
>            targetNamespace="urn:jboss:domain:iiop-openjdk:4.0"
>            xmlns="urn:jboss:domain:iiop-openjdk:4.0"
12c12
<            version="3.0">
---
>            version="4.0">
```

tadamski:
@rhusar tbh I always thought that schema file is necessary as a formal reference whenever schema version changes. If it is not the case I can just remove it as it indeed doesn't introduce any changes.

rhusar:
> @rhusar tbh I always thought that schema file is necessary as a formal reference whenever schema version changes. If it is not the case I can just remove it as it indeed doesn't introduce any changes.

I meant whether we need a new schema meaning version/parser/xsd. There are no changes in that schema, both of these were already optional.

```xml
        <xs:attribute name="server-ssl-context" type="xs:string" use="optional"/>
        <xs:attribute name="client-ssl-context" type="xs:string" use="optional"/>
```

i.e. the enforcement was never in the XSD, so making these independent and optional in the model was always the case in the XSD. What is the value of adding a new schema if they are exactly the same?

tadamski:
@rhusar I have removed 4.0 schema and left only model bump.

rhusar (approved):
LGTM, I don't see any more issues, what a saga :-)

tadamski:
I don't think those testing failures are related.

rhusar:
> I don't think those testing failures are related.

Correct, these are known to be flaky. No need to rerun.

rhusar:
Adding 39.x label so that it gets proper consideration for inclusion (since Brian suggested 39 above)

bstansberry (approved):
Thanks @tadamski !

PR #17212, opened 2023-09-25
WFLY-18537 Memory leak :reload operation

marcosgopen:
https://issues.redhat.com/browse/WFLY-18537

Memory leak :reload operation

This PR needs the update of WFTC because this issue depends on  (https://github.com/wildfly/wildfly-transaction-client/pull/172) which is released in WFTC  3.0.1.Final

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/service/LocalTransactionContextService.java
+++ b/transactions/src/main/java/org/jboss/as/txn/service/LocalTransactionContextService.java
@@ -50,6 +50,7 @@ public final class LocalTransactionContextService implements Service<LocalTransa
     private final InjectedValue<XAResourceRecoveryRegistry> xaResourceRecoveryRegistryInjector = new InjectedValue<>();
     private final InjectedValue<ServerEnvironment> serverEnvironmentInjector = new InjectedValue<>();
     private final int staleTransactionTime;
+    private JBossLocalTransactionProvider provider;
 
     public LocalTransactionContextService(final int staleTransactionTime) {
         this.staleTransactionTime = staleTransactionTime;
@@ -62,7 +63,8 @@ public final class LocalTransactionContextService implements Service<LocalTransa
         builder.setXAResourceRecoveryRegistry(xaResourceRecoveryRegistryInjector.getValue());
         builder.setXARecoveryLogDirRelativeToPath(serverEnvironmentInjector.getValue().getServerDataDir().toPath());
         builder.setStaleTransactionTime(staleTransactionTime);
-        final LocalTransactionContext transactionContext = this.context = new LocalTransactionContext(builder.build());
+        this.provider = builder.build();
+        final LocalTransactionContext transactionContext = this.context = new LocalTransactionContext(this.provider);
         // TODO: replace this with per-CL settings for embedded use and to support remote UserTransaction
         doPrivileged((PrivilegedAction<Void>) () -> {
             LocalTransactionContext.getContextManager().setGlobalDefault(transactionContext);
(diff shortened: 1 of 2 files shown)


transactions/src/main/java/org/jboss/as/txn/service/LocalTransactionContextService.java
```diff
@@ -50,6 +50,7 @@ public final class LocalTransactionContextService implements Service<LocalTransa
     private final InjectedValue<XAResourceRecoveryRegistry> xaResourceRecoveryRegistryInjector = new InjectedValue<>();
     private final InjectedValue<ServerEnvironment> serverEnvironmentInjector = new InjectedValue<>();
     private final int staleTransactionTime;
+    private JBossLocalTransactionProvider provider;
```
chengfang, line 53: is `provider` thread-safe? As a data point in line 47 `context` is declared `volatile`, and both fields seems to have similar purpose. Or can we not add `provider`, and get the provider from `this.context` in `stop` method?

marcosgopen: Thanks for the review @chengfang .  Good point!
What about setting start() and stop() method as synchronized instead?

marcosgopen: Ok, but there is no getter in LocalTransactionContext for the provider, and that is private (see https://github.com/wildfly/wildfly-transaction-client/blob/master/src/main/java/org/wildfly/transaction/client/LocalTransactionContext.java#L71). I don't know exactly how to "get the provider from this.context in stop method"?

chengfang: sorry, I missed that. I really wished it could be public. Then how about making `provider` volatile too?

marcosgopen: I am not against using `volatile` on the `provider`. If in the stop() method there was the `this.provider = null` assignment I would agree. But here in the stop method the provider reference does not change.
IMO if we need to have the provider more thread-safe we would better execute the start and the stop method not in parallel (i.e. with synchronized methods) .

marcosgopen:
I would like to mention that I tested this PR against the [check-leak tool]([check-leak](https://github.com/check-leak/check-leak)) as described in https://issues.redhat.com/browse/JBEAP-25588 and the tool does not report any memory leaks anymore.

transactions/src/main/java/org/jboss/as/txn/service/LocalTransactionContextService.java
chengfang: Okay, that's good point. I think what we have should be good.

chengfang (approved):

TomasHofman:
Hello @marcosgopen @chengfang , we use transaction client 1.1.15 in EAP 7.4, this issue looks to be present there also. Do you think it would be possible to backport there?

transactions/src/main/java/org/jboss/as/txn/service/LocalTransactionContextService.java
bstansberry: Note that MSC won't invoke start and stop concurrently.

bstansberry:
Thanks @marcosgopen and @chengfang

chengfang:
@TomasHofman sure, I'll work on the backport in wildfly-transaction-client and release.

TomasHofman:
Many thanks @chengfang !

PR #17214, opened 2023-09-26
[WFLY-18534] Change the WildFly license to ASL 2.0

bstansberry:
https://issues.redhat.com/browse/WFLY-18534


jasondlee (approved):
Thanks to wildfly-bot, I'm listed as a reviewer, so I'll respond, but I will say this: I did not personally look at each of the 5,000+ files changed. That said, yay for the license change. :)

chengfang:
do we also need to update https://github.com/wildfly/wildfly/blob/main/COMMITMENT ? 

The content of that file seems premised on gpl licenses.

bstansberry:
@chengfang probably, unless @n1hility disagrees. I interpreted it as kind of harmless but off topic. It may still be meaningful for WildFly as a whole (which provides a lot of LGPL software) but it's applicability to this particular repo is questionable.

Either way I don't think it's urgent for 30 Beta.

chengfang (approved):

PR #17230, opened 2023-09-27
[WFLY-18547] Set attributes to default values if available

The-Huginn:
https://issues.redhat.com/browse/WFLY-18547

### DIFF
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
@@ -51,7 +51,8 @@ class AddressSettingAdd extends AbstractAddStepHandler {
     protected void performRuntime(final OperationContext context, final ModelNode operation, final ModelNode model) throws OperationFailedException {
         final ActiveMQServer server = getActiveMQServer(context, operation);
         if (server != null) {
-            final AddressSettings settings = createSettings(context, model);
+            boolean isRootAddressMatch = server.getConfiguration().getWildcardConfiguration().getAnyWordsString().equals(context.getCurrentAddressValue());
+            final AddressSettings settings = createSettings(context, model, isRootAddressMatch);
             server.getAddressSettingsRepository().addMatch(context.getCurrentAddressValue(), settings);
         }
     }
@@ -60,12 +61,11 @@ class AddressSettingAdd extends AbstractAddStepHandler {
      * Create a setting.
      *
      * @param context the operation context
-     * @param config the detyped config
+     * @param config  the detyped config
      * @return the address settings
-     *
      * @throws OperationFailedException if the model is invalid
      */
-    static AddressSettings createSettings(final OperationContext context, final ModelNode config) throws OperationFailedException {
+    static AddressSettings createSettings(final OperationContext context, final ModelNode config, boolean isRootAddressMatch) throws OperationFailedException {
         final AddressSettings settings = new AddressSettings();
         if (config.hasDefined(AddressSettingDefinition.ADDRESS_FULL_MESSAGE_POLICY.getName())) {
             final AddressFullMessagePolicy addressPolicy = AddressFullMessagePolicy.valueOf(AddressSettingDefinition.ADDRESS_FULL_MESSAGE_POLICY.resolveModelAttribute(context, config).asString());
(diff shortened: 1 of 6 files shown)


messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
```diff
@@ -103,52 +117,81 @@ static AddressSettings createSettings(final OperationContext context, final Mode
...
-        }
-        if (config.hasDefined(AddressSettingDefinition.SLOW_CONSUMER_POLICY.getName())) {
-            final SlowConsumerPolicy slowConsumerPolicy = SlowConsumerPolicy.valueOf(AddressSettingDefinition.SLOW_CONSUMER_POLICY.resolveModelAttribute(context, config).asString());
-            settings.setSlowConsumerPolicy(slowConsumerPolicy);
-        }
-        if (config.hasDefined(AddressSettingDefinition.SLOW_CONSUMER_THRESHOLD.getName())) {
-            settings.setSlowConsumerThreshold(AddressSettingDefinition.SLOW_CONSUMER_THRESHOLD.resolveModelAttribute(context, config).asLong());
-        }
+        settings.setExpiryDelay(
+                config.hasDefined(EXPIRY_DELAY.getName())
```
ehsavoie, line 121: You don't have to use such circumventions:
EXPIRY_DELAY.resolveModelAttribute(context, config).asLong() will use the default value if non is defined

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
```diff
@@ -103,52 +117,81 @@ static AddressSettings createSettings(final OperationContext context, final Mode
...
+                        ? EXPIRY_DELAY.resolveModelAttribute(context, config).asLong()
+                        : EXPIRY_DELAY.getDefaultValue().asLong()
+        );
+        settings.setDefaultLastValueQueue(
+                config.hasDefined(LAST_VALUE_QUEUE.getName())
```
ehsavoie, line 126: same as as ^^

ehsavoie (changes requested):

The-Huginn (commented):
I can see Artemis Settings already have all the values set to the default values. Should I then close this PR?

ehsavoie:
I think you misunderstood my comment.
I wasn't clear so I'll try to improve:
you don't have to check if the attribute is defined as in the following test:
 config.hasDefined(EXPIRY_DELAY.getName())
you 'just' need to set the value to the resolved value as the resolved value will be set to the default value if it is not defined thus:
 settings.setExpiryDelay(config.hasDefined(EXPIRY_DELAY.getName())? EXPIRY_DELAY.resolveModelAttribute(context, config).asLong()
                        : EXPIRY_DELAY.getDefaultValue().asLong());

can be simplified to:

settings.setExpiryDelay(EXPIRY_DELAY.resolveModelAttribute(context, config).asLong());

The-Huginn:
Yes, sorry, I didn't realize, it takes (eventually) default value in `resolveModeAttribute` internally. Thanks for the detailed explanation.

ehsavoie (approved):

bstansberry:
@The-Huginn @ehsavoie CI is showing failures in AddressSettingsTestCase

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
```diff
@@ -22,6 +18,25 @@
...
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.PAGE_SIZE_BYTES;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.REDELIVERY_DELAY;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.REDELIVERY_MULTIPLIER;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.REDISTRIBUTION_DELAY;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.SEND_TO_DLA_ON_NO_ROUTE;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.SLOW_CONSUMER_CHECK_PERIOD;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.SLOW_CONSUMER_POLICY;
+import static org.wildfly.extension.messaging.activemq.AddressSettingDefinition.SLOW_CONSUMER_THRESHOLD;
+import static org.wildfly.extension.messaging.activemq.CommonAttributes.DEAD_LETTER_ADDRESS;
+import static org.wildfly.extension.messaging.activemq.CommonAttributes.EXPIRY_ADDRESS;
```
bstansberry, line 38: Please move these static imports back up to the top. That's the WildFly code style. Thanks!

bstansberry (changes requested):

The-Huginn:
After some searching, I believe the cause is in the `AddressSettings.merge()` as internally the `AddressSettings` are using boxed types (Integer) instead of primitives (int), thus default value is null instead of some, for example numerical value.
By always setting the value none of the settings is set to null, so it does not take the value from the merged.
This is the case for the test `testSpecifiedAddressSetting()`.
Due to the order of the test, after it fails it does not remove the address and the other 2 tests fail as well (Not related)

Please let me know, how you would like to proceed @ehsavoie 

I am not pushing the re-ordered imports just yet

ehsavoie (changes requested):
@The-Huginn you are correct during the merge operation null as a special meaning: thus if we set a value on a sub-setting the super-setting won't override its value so I guess we need to ensure that the if is there for all attribute

The-Huginn:
@ehsavoie now I am wondering, if this issue is still relevant, or is obsolete now. As each getter provides a default value in `AddressSettings`

ehsavoie:
@The-Huginn I've updated the issue

ehsavoie:
@The-Huginn well in the end the attributes until MAX_READ_PAGE_BYTES should use the same if element.
Maybe we could also put some comment here

bstansberry:
It sounds like this issue has moved from one of 'use the defaults' to one of 'fix the management API read-resource-description output to better reflect what actually happens.'

bstansberry:
FYI @ehsavoie and I were chatting about some thoughts around this, but I'll bow out and leave it to him to decide what he wants.

Thanks so much @The-Huginn for the research on this!

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingDefinition.java
```diff
@@ -108,7 +108,7 @@ public class AddressSettingDefinition extends PersistentResourceDefinition {
             .build();
 
     public static final SimpleAttributeDefinition MAX_READ_PAGE_BYTES = create("max-read-page-bytes", ModelType.INT)
-            .setDefaultValue(new ModelNode(-1))
+            .setDefaultValue(new ModelNode(2 * 10 * 1024 * 1024))
```
ehsavoie, line 111: Leave that one to -1

ehsavoie:
The best workaround is to set the default values on the root wildcard address thus they will act as default values when the address settings are merged during resolution.

ehsavoie (approved):

ehsavoie (approved):

bstansberry:
I added the Critical label to this as it fell into a crack and needs to be rescued and merged.

bstansberry:
Superseded by #18026, which resolves the conflicts.

PR #17234, opened 2023-09-27
WFLY-18557 appclient -secmgr should be allowed under Java 21

scottmarlow:
https://issues.redhat.com/browse/WFLY-18557

I tested the appclient.sh change locally but am not sure how to test the appclient.bat change as I don't have a Windows machine or vm setup.  Any suggestions?

I did include a change to the appclient.sh that did but I didn't make a similar change to appclient.bat:
```
SECURITY_MANAGER_SET=`echo $JAVA_OPTS | $GREP "java\.security\.manager"`
if [ "x$SECURITY_MANAGER_SET" != "x" ]; then
    SECURITY_MANAGER_SET_TO_ALLOW=`echo $JAVA_OPTS | $GREP "java\.security\.manager=allow"`
    if [ "x$SECURITY_MANAGER_SET_TO_ALLOW" = "x" ]; then
        echo "ERROR: The use of -Djava.security.manager has been removed. Please use the -secmgr command line argument or SECMGR=true environment variable."
        exit 1
    fi
fi
```

So, we need to test the appclient.bat on Java 21 with the -secmgr option.

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/content/bin/appclient.bat
+++ b/ee-feature-pack/galleon-shared/src/main/resources/content/bin/appclient.bat
@@ -116,8 +116,10 @@ if "x%JBOSS_MODULEPATH%" == "x" (
   set  "JBOSS_MODULEPATH=%JBOSS_HOME%\modules"
 )
 
+setlocal EnableDelayedExpansion
 call "!DIRNAME!common.bat" :setSecurityManagerDefault
 set "JAVA_OPTS=!JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
+setlocal DisableDelayedExpansion
 
 rem Set the module options
 set "MODULE_OPTS="
--- a/ee-feature-pack/galleon-shared/src/main/resources/content/bin/appclient.bat
+++ b/ee-feature-pack/galleon-shared/src/main/resources/content/bin/appclient.bat
@@ -116,6 +116,9 @@ if "x%JBOSS_MODULEPATH%" == "x" (
   set  "JBOSS_MODULEPATH=%JBOSS_HOME%\modules"
 )
 
+call "!DIRNAME!common.bat" :setSecurityManagerDefault
+set "JAVA_OPTS=!JAVA_OPTS! !SECURITY_MANAGER_CONFIG_OPTION!"
+
 rem Set the module options
 set "MODULE_OPTS="
 if "%SECMGR%" == "true" (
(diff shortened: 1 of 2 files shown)


scottmarlow:
The appclient.bat script needs more changes as per Windows failure:
{quote}
java.lang.AssertionError: Expected an exit value 0f 0 got 1
Command Executed:
C:\BuildAgent\work\e34a6f994de9f7c6\testsuite\scripts\target\wildfly\bin\appclient.bat "-v"
Environment:
Output:
Calling "C:\BuildAgent\work\e34a6f994de9f7c6\testsuite\scripts\target\wildfly\bin\appclient.conf.bat"
'"!DIRNAME!common.bat"' is not recognized as an internal or external command,
operable program or batch file.
Error: Could not find or load main class !JAVA_OPTS!
Caused by: java.lang.ClassNotFoundException: !JAVA_OPTS!
java.lang.AssertionError:
Expected an exit value 0f 0 got 1
Command Executed:
C:\BuildAgent\work\e34a6f994de9f7c6\testsuite\scripts\target\wildfly\bin\appclient.bat "-v"
Environment:
Output:
Calling "C:\BuildAgent\work\e34a6f994de9f7c6\testsuite\scripts\target\wildfly\bin\appclient.conf.bat"
'"!DIRNAME!common.bat"' is not recognized as an internal or external command,
operable program or batch file.
Error: Could not find or load main class !JAVA_OPTS!
Caused by: java.lang.ClassNotFoundException: !JAVA_OPTS!
{quote}

bstansberry (changes requested):
@scottmarlow It's great that AppClientScriptTestCase checks these. Kudos to @jamezp for adding those!!!!

But, for this you also need to deal with the Powershell script appclient.ps1.

But the existing test will cover it. :)

Note also that we have nightly CI jobs set up for SE 21 so we can do custom runs of those using this branch instead of wildfly/wildfly main.

bstansberry:
> @jamezp would it make sense to add -secmgr to AppClientScriptTestCase?
> 

Yes. It should be simple enough to check if system property 'security.manager' is true and if so add "-secmgr" to the params passed to the script.start. The effect is the CI jobs with the security manager on will test that.

> @bstansberry or should I find a way to run Windows locally?
> 
Let's see if SE 21 for Windows is available for addition to ci.wf.org.

https://wildfly.zulipchat.com/#narrow/stream/174177-ci-servers/topic/SE.2021

bstansberry:
> > @jamezp would it make sense to add -secmgr to AppClientScriptTestCase?
> 
> Yes. It should be simple enough to check if system property 'security.manager' is true and if so add "-secmgr" to the params passed to the script.start. The effect is the CI jobs with the security manager on will test that.
> 

@scottmarlow @jamezp FYI I filed https://issues.redhat.com/browse/WFCORE-6532 to add a similar thing to our testing of domain.sh and standalone.sh.  I left that one unassigned with a 'hacktoberfest' label.

PR #17237, opened 2023-09-28
WFLY-18568 Disable cache operation client events if near cache is disabled

pferraro:
https://issues.redhat.com/browse/WFLY-18568


rhusar (approved):
LGTM

PR #17240, opened 2023-10-02
WFLY-18575 Remote caches used by HotRod cache store should never trigger client listeners

pferraro:
https://issues.redhat.com/browse/WFLY-18575


rhusar (approved):
LGTM

PR #17248, opened 2023-10-04
[WFLY-18578] Allow channel-based overriding of the org.jboss.as.produ…

bstansberry:
…ct data

Use artifacts to store the JBoss-Product-Release-Name and JBoss-Product-Release-Version MANIFEST.MF entries to make it possible to update that info via a channel

https://issues.redhat.com/browse/WFLY-18578

Incorporates https://issues.redhat.com/browse/WFLY-18791 as it's the foundation for the test.

@jmesnil @wolfc @spyrkob @jfdenise @darranl @bmaxwell whether we decide to make use of this capability in the near term or not, I think this is the correct way to do this. The old approach was a mix of pom magic, galleon magic and JBoss Modules magic. Too much magic.


spyrkob:
Is the expectation that if we wanted to separate the reported version from build version, we'd build the product-conf jar separately and inject it into the channel?

jfdenise:
@bstansberry , I tried locally, I opened PR against your branch for small adjustments: https://github.com/bstansberry/wildfly/pull/65

bstansberry:
> Is the expectation that if we wanted to separate the reported version from build version, we'd build the product-conf jar separately and inject it into the channel?

@spyrkob Yes. Getting all that going would be the harder part; this part is simpler.

bstansberry:
> @bstansberry , I tried locally, I opened PR against your branch for small adjustments: [bstansberry#65](https://github.com/bstansberry/wildfly/pull/65)

Thanks, @jfdenise -- I merged that.

bstansberry:
@spyrkob Note that there are other ways we could approach things; this is just one and it involves the least change to the main server code. We can discuss options elsewhere. I decided to push this forward since looking at how we make this data available currently is overly magical and it's valid to update it.

bstansberry:
I've found the cause of the Windows failures.

bstansberry:
I'll retest one #17368 is merged.

boms/common-expansion/pom.xml
```diff
@@ -486,17 +486,6 @@
...
-                <groupId>io.opentelemetry</groupId>
-                <artifactId>opentelemetry-exporter-otlp-common</artifactId>
-                <version>${version.io.opentelemetry.opentelemetry}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>*</groupId>
-                        <artifactId>*</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
```
bstansberry, line 499: Note that this change was off-topic. It's a duplicate entry in the bom that maven was complaining about, so I removed it.

jfdenise:
@bstansberry , I am wandering if we really need the channel artifact. What the WildFly Maven plugin is using is the channel manifest (maven coordinates or URL). The maven repos in use to build the channel are the ones coming from the maven context.

bstansberry:
@jfdenise I combined the channel.yaml and manifest.yaml stuff into a single maven module per FP. I disabled attaching the channel.yaml as an artifact for now.

jfdenise:
@bstansberry , the Maven plugin that builds the feature-pack deploys the channel (https://github.com/wildfly/galleon-plugins/blob/main/maven-plugin/src/main/java/org/wildfly/galleon/maven/AbstractFeaturePackBuildMojo.java#L398). We should make that optional and disable it in WildFly. Perhaps that it could be enough and the channel module not required?

bstansberry:
@jfdenise I'm not sure whether that needs to be configurable or not. Well, it does, but it doesn't seem super urgent. However, we should *never* promote (e.g. blogs, docs etc) such a manifest independent of a conscious decision by the relevant project that that's what it wants to do. We don't have that for WF or WF Core.

I'd prefer to leave the module in WF. I don't like the artifactId of our feature packs and if we started promoting a WF channel I'd need convincing to use that artifactId.

jfdenise:
@bstansberry , thank-you. I created https://issues.redhat.com/browse/WFGP-262 to track it.

ee-feature-pack/channel/src/main/resources/wildfly-ee-channel.yaml
```diff
@@ -0,0 +1,9 @@
+---
+schemaVersion: "2.0.0"
+repositories:
+  - id: "central"
+    url: "https://repo1.maven.org/maven2/"
+  - id: "jboss-public"
+    url: "https://repository.jboss.org/nexus/content/groups/public/"
+manifest:
+  maven: "org.wildfly:${project.artifactId}:${project.version}"
```
spyrkob, line 9: the maven coordinates should be broken up like:
```
manifest:
  maven:
    groupId: "org.wildfly"
    artifactId: "${project.artifactId}"
    version: "${project.version}"
```

Also does the channel need to define the manifest version? I think it would make more sense to have the major version included into the artifactId of the manifest so the updates can be consumed if needed

pom.xml
```diff
@@ -569,6 +570,22 @@
 
             <!-- Keep sorted -->
 
+            <dependency>
+                <groupId>${full.maven.groupId}</groupId>
+                <artifactId>wildfly-channel</artifactId>
+                <version>${full.maven.version}</version>
+                <type>yaml</type>
+                <classifier>channel</classifier>
```
spyrkob, line 578: since the channels are not attached to the reactor, I think this dependency could be removed? Or does it make sense to keep it in case we need to attache the channels in future?

testsuite/integration/manualmode/pom.xml
```diff
@@ -383,8 +397,62 @@
...
+                                    </manifest>
+                                </channel>
+                            </channels>
+                            <feature-packs>
+                                <feature-pack>
+                                    <groupId>${ee.maven.groupId}</groupId>
+                                    <artifactId>wildfly-ee-galleon-pack</artifactId>
+                                    <version>${ee.maven.version}</version>
+                                </feature-pack>
+                                <feature-pack>
```
spyrkob, line 445: Is it necessary to include prospero feature pack? IIUC the test only verifies the interaction between wildfly channel and the product-conf-override-manifest.yaml. The integration tests in prospero already verify [installing Wildfly + prospero feature pack](https://github.com/wildfly-extras/prospero/blob/main/integration-tests/src/test/java/org/wildfly/prospero/it/featurepacks/WildflyFpTest.java).

bstansberry: Right; @spyrkob -- that can be removed. That's left over from when I also used this execution for the testing I later moved into https://github.com/wildfly/wildfly-core/pull/5724.

pom.xml
bstansberry: Yeah, that can be removed. When I started all this I added some random off-topic bits in case we decided to push and deploy channels as part of WF 31. But now it's right at the deadline and we need more discussion before doing that, so these can be removed.

bstansberry: Fixed.

testsuite/integration/manualmode/pom.xml
bstansberry: Fixed.

bstansberry:
Thanks @spyrkob -- I believe I've addressed the issues you raised.

spyrkob (approved):
LGTM!

bstansberry:
Thanks @spyrkob and @jfdenise!

PR #17260, opened 2023-10-06
WFLY-18568 (Revised) Disable cache operation client events if near cache is disabled.

pferraro:
https://issues.redhat.com/browse/WFLY-18568

Reverts previous fix. Instead applies HotRod client flags per operation, as these are thread local.


clustering/web/hotrod/src/main/java/org/wildfly/clustering/web/hotrod/sso/coarse/CoarseSessionsFactory.java
```diff
@@ -65,6 +68,7 @@ public Map.Entry<String, Map<D, S>> findEntryContaining(S session) {
 
     @Override
     public boolean remove(String id) {
-        return this.cache.withFlags(Flag.FORCE_RETURN_VALUE).remove(new CoarseSessionsKey(id)) != null;
+        this.cache.withFlags(this.ignoreReturnFlags).remove(new CoarseSessionsKey(id));
```
bstansberry, line 71: @pferraro So using Flag.FORCE_RETURN_VALUE was wrong; i.e. this bit isn't just about including Flag.SKIP_LISTENER_NOTIFICATION?

pferraro: We don't actually need the return value here, so, yes, this is an additional optimization.

PR #17261, opened 2023-10-06
[WFLY-18605][WFLY-18606] reevaluated test excludes pro ts.preview and ts.bootable…

istudens:
….preview profiles

Issue:
https://issues.redhat.com/browse/WFLY-18605      (testsuite/integration/elytron)
https://issues.redhat.com/browse/WFLY-18606      (testsuite/integration/web)

### DIFF
--- a/testsuite/integration/web/pom.xml
+++ b/testsuite/integration/web/pom.xml
@@ -39,7 +39,6 @@
         <bootable.copy-external-taglib-module-xml.phase>none</bootable.copy-external-taglib-module-xml.phase>
         <bootable.create-external-taglib-module.phase>none</bootable.create-external-taglib-module.phase>
         <bootable-jar-web-profile-packaging.phase>none</bootable-jar-web-profile-packaging.phase>
-        <bootable-jar-web-legacy-security-packaging.phase>none</bootable-jar-web-legacy-security-packaging.phase>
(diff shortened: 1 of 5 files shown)


fjuma (approved):
Thanks @istudens!

testsuite/integration/web/pom.xml
```diff
@@ -1288,18 +1288,14 @@
                                         <arquillian.xml>arquillian-bootable.xml</arquillian.xml>
                                     </systemPropertyVariables>
                                     <excludes>
-                                        <!--TODO re-evaluate these exclusions from EE 9 work-->
                                         <!-- requires legacy-security -->
                                         <exclude>org/jboss/as/test/integration/web/security/**/*TestCase.java</exclude>
```
bstansberry, line 1292: @istudens @fjuma This seems wrong. Not just here or elsewhere related to ts.preview or ts.bootable, but perhaps anywhere in this pom.  We don't have legacy security anywhere any more and these tests run and pass on some jobs.

Of course it's possible some need to be excluded from some profiles for other reasons, e.g. they require adding a module etc.

istudens: @bstansberry That's correct and I missed this. There is no legacy security even in bootable jar profiles, I've changed the PR accordingly. Thanks.

bstansberry (approved):

bstansberry:
Thanks @istudens!

PR #17263, opened 2023-10-06
[WFLY-18590] SimpleTimerMDBTestCase: change waiting time

tadamski:
https://issues.redhat.com/browse/WFLY-18585

The bug is hard to reproduce and even if I managed to do this I can't see any obvious failures in the log. Maybe we can try giving the bean more waiting time and see if it helps.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/TimedObjectTimerServiceMDB.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/TimedObjectTimerServiceMDB.java
@@ -35,9 +35,9 @@ public class TimedObjectTimerServiceMDB implements TimedObject , MessageListener
         timerService.createTimer(100, null);
     }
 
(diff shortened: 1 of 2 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/TimedObjectTimerServiceMDB.java
```diff
@@ -37,7 +37,7 @@ public void createTimer() {
 
     public static boolean awaitTimerCall() {
         try {
-            latch.await(2, TimeUnit.SECONDS);
+            latch.await(5, TimeUnit.SECONDS;
```
pferraro, line 40: This doesn't compile, BTW... :)
That said, rather than bumping the wait time, I would suggest using org.jboss.as.test.shared.TimeoutUtil.adjust(...) to scale the timeout value to accommodate laggy test hosts.

pferraro (changes requested):

bstansberry:

PR #17266, opened 2023-10-09
[WFLY-14769] Access checking for txn namespace lookup

tadamski:
https://issues.redhat.com/browse/WFLY-14769

This is a simpler solution which injects access check to transaction-client library and does not modify naming subsystem.


bstansberry (approved):

bstansberry:
Thanks @tadamski. You've been very patient!

PR #17285, opened 2023-10-12
WFLY-18635 Drop support for scattered cache

pferraro:
https://issues.redhat.com/browse/WFLY-18635


rhusar (approved):
Looks good to me. Having said that, my preference would be to notify the users that the behavior is defaulting to a dist cache via a log message.

pferraro:
@rhusar That is already case within the resource description.

PR #17290, opened 2023-10-13
[WFLY-18642] reevaluated ts.preview and ts.bootable.preview profiles

istudens:
Issue: https://issues.redhat.com/browse/WFLY-18642

### DIFF
--- a/testsuite/integration/microprofile/pom.xml
+++ b/testsuite/integration/microprofile/pom.xml
@@ -389,6 +389,8 @@
                                 <!--Dependencies for lra specs -->
                                 <layer>microprofile-lra-coordinator</layer>
                                 <layer>microprofile-lra-participant</layer>
+                                <!-- ts.bootable.preview tests require this -->
+                                <layer>microprofile-rest-client</layer>
                             </layers>
                         </configuration>
                     </execution>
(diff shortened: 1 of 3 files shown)


testsuite/integration/microprofile/pom.xml
```diff
@@ -389,6 +389,8 @@
                                 <!--Dependencies for lra specs -->
                                 <layer>microprofile-lra-coordinator</layer>
                                 <layer>microprofile-lra-participant</layer>
+                                <!-- ts.bootable.preview tests require this -->
+                                <layer>microprofile-rest-client</layer>
```
istudens, line 393: @bstansberry I couldn't realize why `ts.bootable.preview` is missing `org.eclipse.microprofile.restclient` module. It seems `wildfly-preview-expansion-bom` inherits it from `wildfly-common-expansion-dependency-management`, but when running with `ts.bootable.preview`, the `test-wildfly.jar` is missing it (while with `ts.bootable` it is there as expected).

istudens:
@bstansberry Can you review this please? Thanks

testsuite/integration/microprofile/pom.xml
bstansberry: Thanks, @istudens for pointing this out!

This is actually a bug in WFP:  https://issues.redhat.com/browse/WFLY-18702.

Other than this change, which shouldn't be needed when that bug is fixed, this PR LGTM. But the CI run is out of date, so what I'm going to do is take your branch and add a commit fixing WFLY-18702, which will include removing these 2 lines. I'll send that up and it will get fresh CI.

bstansberry: https://github.com/wildfly/wildfly/pull/17349

PR #17300, opened 2023-10-17
[WFLY-18673] GitHub Actions Run Reductions

jamezp:
Reduce the number of jobs that run per PR. Ignore specific files for CI runs and ensure previous runs are cancelled if a branch/PR is updated and a push is done.

https://issues.redhat.com/browse/WFLY-18673

### DIFF
--- a/.github/workflows/dep-diff-pull_request.yml
+++ b/.github/workflows/dep-diff-pull_request.yml
@@ -5,7 +5,9 @@ name: Dependency Tree Input Builder
 on:
   pull_request:
     branches:
-    - main
+      - main
+    paths:
+      - "**/pom.xml"
 env:
   # The modules to check for dependencies. If there is more than one they are comma separated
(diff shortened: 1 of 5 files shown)


.github/workflows/dep-diff-pull_request.yml
```diff
@@ -5,7 +5,9 @@ name: Dependency Tree Input Builder
 on:
   pull_request:
     branches:
-    - main
+      - main
+    paths:
+      - "**/pom.xml"
```
bstansberry, line 10: @luck3y is the pull-player kickoff of the ci-wf-org jobs still gated by completion of this job? If so we should rethink that.

luck3y: @bstansberry As far as I am aware there is no gating, they run async of one another.

bstansberry (approved):

bstansberry:
Thanks @jamezp and @luck3y

PR #17304, opened 2023-10-17
[WFLY-18657] Fix i18n of exception messages in ee

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-18657
[WFLY-18657] Fix i18n of exception messages in ee


bstansberry:
Thanks @soul2zimate

PR #17307, opened 2023-10-17
[WFLY-18660] Fix i18n of exception messages in messaging-activemq

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-18660

[WFLY-18660] Fix i18n of exception messages in messaging-activemq

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/messaging-activemq/injection/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/messaging-activemq/injection/main/module.xml
@@ -24,7 +24,6 @@
         <module name="jakarta.jms.api"/>
         <module name="jakarta.resource.api"/>
         <module name="jakarta.transaction.api"/>
-
         <module name="org.jboss.as.controller"/>
         <module name="org.jboss.as.ee"/>
         <module name="org.jboss.as.server"/>
--- a/messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/_private/MessagingLogger.java
+++ b/messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/_private/MessagingLogger.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.messaging.activemq._private;
+
+import jakarta.jms.IllegalStateRuntimeException;
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * Date: 10.06.2011
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@MessageLogger(projectCode = "WFLYMSGAMQINJ", length = 4)
+public interface MessagingLogger extends BasicLogger {
+    /**
+     * The logger with the category of the package.
+     */
+    MessagingLogger ROOT_LOGGER = Logger.getMessageLogger(MessagingLogger.class, "org.wildfly.extension.messaging-activemq");
+
+    /**
+     * Create an exception when calling a method not allowed on injected JMSContext.
+     *
+     * @return an {@link IllegalStateRuntimeException} for the error.
+     */
+    @Message(id = 1, value = "It is not permitted to call this method on injected JMSContext (see Jakarta Messaging 2.0 spec, §12.4.5).")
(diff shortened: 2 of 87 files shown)


ehsavoie (approved):

soul2zimate:
failures are related, checking.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/messaging-activemq/injection/main/module.xml
```diff
@@ -25,6 +25,7 @@
         <module name="jakarta.resource.api"/>
         <module name="jakarta.transaction.api"/>
 
+        <module name="org.jboss.as.cli"/>
```
bstansberry, line 28: @soul2zimate @ehsavoie This makes me uncomfortable. How about the MessagingLogger methods just return String and the callers in the subsystem modules construct the OperationFormatException themselves?

There's a separation of concerns problem in this extension, where the 'subsystem' module is using the 'injection' module to hold its logger class. And of all the messages in the logger, only 1 is related to 'injection'! That's not clean but I'm not suggesting fixing that. Well, I am suggesting fixing that, just not necessarily here. :-)  But I don't think we should worsen the separation of concerns problem by bringing CLI types into 'injection'.

See also https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Logger.20interface.20'duplication'/near/399726322

bstansberry (changes requested):

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/messaging-activemq/injection/main/module.xml
soul2zimate: @bstansberry Thank you for the review and suggestion. I have updated the fix.

messaging-activemq/injection/src/main/java/org/wildfly/extension/messaging/activemq/_private/MessagingLogger.java
```diff
@@ -0,0 +1,34 @@
...
+     * The logger with the category of the package.
+     */
+    MessagingLogger ROOT_LOGGER = Logger.getMessageLogger(MessagingLogger.class, "org.wildfly.extension.messaging-activemq");
+
+    /**
+     * Create an exception when calling a method not allowed on injected JMSContext.
+     *
+     * @return an {@link IllegalStateRuntimeException} for the error.
+     */
+    @Message(id = 1, value = "(WFLYMSGAMQ0068) It is not permitted to call this method on injected JMSContext (see Jakarta Messaging 2.0 spec, §12.4.5).")
```
jamezp, line 31: Adding "(WFLYMSGAMQ0068)" isn't a bad idea, but do we just want to do a clean cut? I've got no strong opinion either way, but if we add it, it will likely be there forever.

bstansberry: I'm the one who suggesting adding this in the zulip thread, but now that I've read what this message actually says I'm fine with a complete cut. I think it's very unlikely that this id is important as a search term, since the message itself (and the associated stack trace) is very precise and points to the spec section where readers can learn more. I don't see what any search result would add to that.

I don't think it does any harm either. The WFLYMSGAMQ0068 id is burned so it's never going to be reused.

My vote is @soul2zimate and @ehsavoie decide what they want and that's fine. :)

jamezp: This makes sense to me. I don't think it hurts anything having it really, just wondered if we wanted to cut it out now or live with it forever is all :)

soul2zimate: @ehsavoie I'd like to remove `(WFLYMSGAMQ0068)` prefix in the message to have a clean cut unless you have an opposite preference

ehsavoie: @soul2zimate I agree.

soul2zimate: Thanks @ehsavoie @bstansberry @jamezp . I have removed `(WFLYMSGAMQ0068)` prefix in the message

jamezp (approved):
This looks good to me. One minor note. The package move to `_private` isn't really necessary. You could update the `module.xml` to do something like:
```xml
<exports>
    <exclude path="org/wildfly/extension/messaging/activemq/logging"/>
</exports>
```

This doesn't need to be changed unless anyone else wants it done. It's just more of an FYI :)

bstansberry:
Thanks @soul2zimate, @ehsavoie and @jamezp

PR #17315, opened 2023-10-17
[WFLY-18668]:  WARN message from Artemis when starting WildFly.

ehsavoie:
* Validation should occur at the ned.

Issue: https://issues.redhat.com/browse/WFLY-18668


bstansberry (approved):

bstansberry:
Thanks @ehsavoie

PR #17329, opened 2023-10-19
[WFLY-16929] ForwardedHandlerTestCase fails with security manager

jasondlee:
https://issues.redhat.com/browse/WFLY-16929

Add permissions.xml to allow DNS lookups

### DIFF
--- a/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/ForwardedHandlerTestCase.java
+++ b/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/ForwardedHandlerTestCase.java
@@ -74,34 +74,33 @@ public class ForwardedHandlerTestCase {
     @ContainerResource
     private ManagementClient managementClient;
 
-    @Deployment(name = FORWARDED_HANDLER_NO_UT_HANDLERS)
-    public static WebArchive deployWithoutUndertowHandlers() {
-        return ShrinkWrap.create(WebArchive.class, FORWARDED_HANDLER_NO_UT_HANDLERS + ".war")
-                .addPackage(ForwardedHandlerTestCase.class.getPackage())
+    private static WebArchive createBaseDeployment(String name) {
+        return ShrinkWrap.create(WebArchive.class, name + ".war")
                 .addClass(TestSuiteEnvironment.class)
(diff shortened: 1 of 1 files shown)


ropalka (approved):

darranl:
Do we have the actual error anywhere of the failed permissions check?

I would like to see what is the justification that these permissions need to be granted to the deployment,

darranl (changes requested):
Before we merge I would like to double check why these permissions are needed.

jasondlee:
> Before we merge I would like to double check why these permissions are needed.

@darranl I don't have a specific error message to stacktrace. I was comparing the deployments for the test that was working with the ones that were not and noticed the discrepancy. I can dig down deeper into the server to find where that might be failing and report back.

bstansberry:
@jasondlee @darranl @fl4via @ropalka This is likely the problem:

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/handlers/ForwardedTestHelperServlet.java#L43

ServletRequest.getLocalName() will attempt to do a reverse DNS lookup and will return the raw IP if that fails. If that is what Undertow does, then its behavior is correct and this is a test issue where the test is expected the DNS lookup to work. If it wants it to work it needs to give itself permission.

From ServletRequest.getLocalName() javadoc:

```
Returns the fully qualified name of the address returned by getLocalAddr(). If the engine cannot or chooses not to resolve the hostname (to improve performance), this method returns the IP address.
```

So this is a case of 'cannot'. That seems allowable per the method javadoc. It could also be allowable under the 'chooses not  to... (to improve performance)' rationale, as setting up a privileged block to make it work would have a perf cost.

bstansberry:
Note that https://issues.redhat.com/secure/attachment/13063081/stacktrace.log shows the logging of the failed permission check. It's only at DEBUG, which indicates the failure is handled. It's in forwarded-servlet-no-ut-handlers.war which only contains the servlet I mentioned. The test failure is due to an unexpected value from that req.getLocalName() call.

bstansberry (approved):

darranl (approved):
Brian has explained why the deployment needs the permission to be granted.

ropalka:
Thank you @bstansberry for detailed justification!

bstansberry:
Thanks @jasondlee, @ropalka and @darranl

PR #17330, opened 2023-10-20
[WFLY-18562] re-evaluated test exclusions for ts.preview and ts.boota…

istudens:
…ble.preview profiles

Issue: https://issues.redhat.com/browse/WFLY-18562


istudens:
I need to test this on CI. Locally it fails on ModelControllerMBeanTestCase due to `java.lang.IllegalArgumentException: Default value not supported for ArrayType and TabularType` exception on "jboss.as:management-root=server" -> getOperation("addSubsystemMicrometer") -> getParameterInfo("exposedSubsystems") and I need to consult this with Kabir as I can't understand why it fails.

bstansberry:
@istudens If that's the last thing blocking this (I saw some other CI failures but didn't investigate, nor have I done a code review), I suggest filing a WFLY against the Micrometer subsystem, and disabling the test with ts.preview with a comment noting the JIRA.

I suspect this actually has nothing to do with Preview, except that we include Micrometer in the default config there, so this test picks up issues there. We support Micrometer in standard WF though, it's just not in the default config.

I suspect this isn't a Micrometer bug per se, it's just it has an unusual thing, where an attribute of ModelType.LIST has a default value. That's a valid thing but it sounds like JMX might not like that.

istudens:
This needs https://issues.redhat.com/browse/WFCORE-6604 to fix the JMX issue.

bstansberry:
@istudens The WFCORE-6604 fix is integrated into full WF now.

bstansberry (approved):

bstansberry:
Thanks @istudens!

PR #17333, opened 2023-10-20
[WFLY-18682] Upgrading WildFly Http Client to 2.0.5.Final

ropalka:
https://issues.redhat.com/browse/WFLY-18682

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/http-client/naming/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/http-client/naming/main/module.xml
@@ -15,6 +15,7 @@
         <module name="java.naming"/>
         <module name="org.wildfly.http-client.common"/>
         <module name="org.wildfly.client.config"/>
+        <module name="org.jboss.ejb-client" />
         <module name="io.undertow.core" />
         <module name="org.jboss.logging"/>
         <module name="org.jboss.marshalling" />
--- a/pom.xml
+++ b/pom.xml
@@ -539,7 +539,7 @@
         <version.org.ow2.asm>9.5</version.org.ow2.asm>
         <version.org.reactivestreams>1.0.4</version.org.reactivestreams>
         <version.org.wildfly.core>22.0.1.Final</version.org.wildfly.core>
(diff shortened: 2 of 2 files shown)


ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/http-client/naming/main/module.xml
```diff
@@ -15,6 +15,7 @@
         <module name="java.naming"/>
         <module name="org.wildfly.http-client.common"/>
         <module name="org.wildfly.client.config"/>
+        <module name="org.jboss.ejb-client" />
```
darranl, line 18: @ropalka does this mean there will also need to be a change to dependencies on clients using WildFly HTTP Client?

ropalka: No when such clients use it either via maven dependency or JBoss Modules dependency mechanism @darranl .

darranl: On the clients though are we sure they will have the required dependency, they could use a strategy like we do ourselves and exclude transitive dependencies?

darranl:
Just noticed a couple of errors in the Windows job looked directly related to the HTTP Client, running that job again.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/http-client/naming/main/module.xml
ropalka: If ejb-client will not be on the classpath they will get ClassNotFoundException and will need to include it @darranl

ropalka:
I just upgraded to 2.0.4.Final version that should sort out the Windows issues @darranl

darranl:
@ropalka FYI I am still seeing the same CI failure on the run from today.

darranl:
Thank you very much @ropalka and @tadamski - CI is green so merging.

darranl (approved):

PR #17337, opened 2023-10-24
[WFLY-18677] Disable testsuite/layers[-expansion] provisioning if -DskipTests is set

xjusko:
https://issues.redhat.com/browse/WFLY-18677


bstansberry:
Thanks @xjusko!

PR #17343, opened 2023-10-25
[WFLY-18693] Upgrade to SmallRye Reactive Messaging 4.11.0

kabir:
https://issues.redhat.com/browse/WFLY-18693

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -404,16 +404,16 @@
         <version.io.reactivex.rxjava3>3.1.6</version.io.reactivex.rxjava3>
         <version.io.smallrye.open-api>3.6.2</version.io.smallrye.open-api>
         <version.io.smallrye.reactive-utils>2.6.0</version.io.smallrye.reactive-utils>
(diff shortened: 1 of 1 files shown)


bstansberry:
@kabir This kind of thing is going to be problematic for the 'experimental features' thing, as it would lead to flagging up an entire MP interface.

https://github.com/smallrye/smallrye-reactive-messaging/compare/4.9.0..4.10.1#diff-338d8d1e1e397b9122f2f07ea57524bbc47dbaed04afeccd496cd60e89d0750dR73-R75

BTW do we still need to use io.smallrye.reactive:smallrye-reactive-messaging-api for the org.eclipse.microprofile.reactive-messaging.api module?

kabir:
Good catch. I will check if we need to use the API still. I think so since other  stuff relies on that, and I can't see a way to pull in just sections of classes.

kabir:
@bstansberry I've realised this is actually fine. We have a few occurrences of this sort of stuff already, which I dealt with in my initial prototype

bstansberry:
Thanks @kabir

PR #17345, opened 2023-10-26
[WFLY-18697] Switch ejb subsystem test dep from groovy-all to groovy

bstansberry:
https://issues.redhat.com/browse/WFLY-18697

This also includes the dependabot upgrade commit from #17272. Reviewing that is what led me to file WFLY-18697.


tadamski (approved):
LGTM

PR #17362, opened 2023-10-31
WFLY-18708 Disable counter-productive "distributable" behavior in Mojarra

pferraro:
https://issues.redhat.com/browse/WFLY-18708

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
@@ -10,6 +10,7 @@ import java.util.List;
 
 import jakarta.faces.application.ViewHandler;
 import org.jboss.as.jsf.logging.JSFLogger;
+import org.jboss.as.jsf.subsystem.JSFResourceDefinition;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
 import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
@@ -18,7 +19,10 @@ import org.jboss.as.web.common.WarMetaData;
 import org.jboss.metadata.javaee.spec.ParamValueMetaData;
 import org.jboss.metadata.web.jboss.JBossServletMetaData;
 import org.jboss.metadata.web.jboss.JBossWebMetaData;
+import org.jboss.metadata.web.spec.ListenerMetaData;
 import org.jboss.metadata.web.spec.MultipartConfigMetaData;
+
+import com.sun.faces.config.ConfigureListener;
 
 /**
  * @author Stuart Douglas
@@ -66,11 +70,12 @@ public class JSFMetadataProcessor implements DeploymentUnitProcessor {
         if(JsfVersionMarker.isJsfDisabled(deploymentUnit)) {
             return;
         }
-        if(metaData == null || metaData.getMergedJBossWebMetaData() == null || metaData.getMergedJBossWebMetaData().getServlets() == null) {
+        JBossWebMetaData webMetaData = (metaData != null) ? metaData.getMergedJBossWebMetaData() : null;
+        if (webMetaData == null || webMetaData.getServlets() == null) {
             return;
         }
         JBossServletMetaData jsf = null;
-        for(JBossServletMetaData servlet : metaData.getMergedJBossWebMetaData().getServlets()) {
+        for(JBossServletMetaData servlet : webMetaData.getServlets()) {
             if(JAVAX_FACES_WEBAPP_FACES_SERVLET.equals(servlet.getServletClass())) {
                 jsf = servlet;
             }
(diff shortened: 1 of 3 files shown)


jasondlee (approved):

pferraro:
@jasondlee FYI, I opened an issue upstream https://github.com/eclipse-ee4j/mojarra/issues/5345 to allow applications to override this property, and provided a fix here: https://github.com/eclipse-ee4j/mojarra/pull/5347
Once that is merged and integrated, this fix can be simplified.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
```diff
@@ -81,22 +85,32 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
...
         // First check the legacy facelets.BUFFER_SIZE property which is required for backwards compatibility
         if (!hasContextParam(webMetaData, "facelets.BUFFER_SIZE")) {
             // The legacy parameter has not been set, set a default buffer if the current parameter name has not been set.
-            setContextParameterIfAbsent(metaData.getMergedJBossWebMetaData(), ViewHandler.FACELETS_BUFFER_SIZE_PARAM_NAME, Integer.toString(defaultBufferSize));
+            setContextParameterIfAbsent(webMetaData, ViewHandler.FACELETS_BUFFER_SIZE_PARAM_NAME, Integer.toString(defaultBufferSize));
+        }
+        // If web application is distributable, disable counter-productive "distributable" logic in Mojarra
+        if (JsfVersionMarker.getVersion(deploymentUnit).equals(JsfVersionMarker.JSF_4_0) && (webMetaData.getDistributable() != null)) {
+            ListenerMetaData mojarraListenerMetaData = new ListenerMetaData();
+            mojarraListenerMetaData.setListenerClass(ConfigureListener.class.getName());
```
bstansberry, line 107: @pferraro @jasondlee Does this work if the JSF impl is MyFaces?

pferraro: AIUI, if the JSF impl is provided by the application (e.g. to use MyFaces or anything else), the JsfVersionMarker.getVersion(deploymentUnit) will return JsfVersionMarker.WAR_BUNDLES_JSF_IMPL.

bstansberry: @pferraro We provide integration of MyFaces outside of the war. The latest mechanism for that is https://github.com/wildfly-extras/wildfly-myfaces-feature-pack.

Gating on JsfVersionMarker.JSF_4_0 may cover that as well; I'm just too lazy to research how it works.

pferraro: Ah... in that case, this fix (as is) will be problematic as it explicitly attaches the bootstrap listener for Mojarra.
@jasondlee Is there any way to distinguish the intended Faces implementation from this DUP?

pferraro: After further inspection, I think this logic is OK, so long as the application selects their preferred JSF implementation via the `org.jboss.jbossfaces.JSF_CONFIG_NAME` servlet context init parameter.
However, if the user instead installed the myfaces feature pack and configured the jsf subsystem to use a specific slot, e.g. myfaces, it is unclear to me how this is meant to work since `JsfVersionMarker.getVersion(deploymentUnit)` will still return "Mojarra-4.0" according to the logic within JSFVersionProcessor.  That does not seem right.  
@jasondlee Can you clarify how this is supposed to work?

jasondlee: @pferraro TBH, I'm not sure ATM. That logic pre-dates me, but I'll take a look and figure it out.

pferraro: OK - I think it is sufficient to verify the slot of the implementation module, since the myfaces feature pack uses same module name but with "myfaces" slot.
Please verify the new logic.

bstansberry:
I'll defer to @jasondlee on the approval for this one. I raised my concern about myfaces above and it looks like @pferraro addressed that so I'll defer to Jason as to whether that was correctly done.

Note that in my memory of history, using 'JSFResourceDefinition.DEFAULT_SLOT' could be problematic as we also supported installing different Mojarra versions, using the same pre-Galleon mechanism we used for MyFaces.

But: 

- I don't know if we'd ever do such a thing again, e.g. let a user choose Mojarra 4 or a future Mojarra 5. (Or perhaps we'd choose based on some higher level thing like EE 11 vs EE 10.)
- If we did, perhaps Mojarra 5 will have Paul's fix, so this workaround is not needed. Or Mojarra 4.x will have Paul's fix and this workaround will be gone.
- In any case adding support for choosing will almost certainly involve touching this code if it still exists and we can deal with it then.

jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFMetadataProcessor.java
pferraro: @jasondlee FYI, https://github.com/eclipse-ee4j/mojarra/pull/5347 was merged and will allow us to simplify this logic, i.e. rather than register servlet context listeners, we can just add the requisite Mojarra-specific servlet context init parameter (which will be safely ignored by other implementations).

petrberan:
Hi @pferraro @bstansberry @jasondlee as we've got PR in EAP 7/8 that depends on this PR, what can we do to get this merged?

bstansberry:
@petrberan Merging this requires an approval from @jasondlee.

jasondlee (approved):

bstansberry:
Thanks @pferraro and @jasondlee. @petrberan -- FYI

PR #17371, opened 2023-11-06
Add externalid

albfan:
https://issues.redhat.com/browse/EAPSUP-1336

Add a column external id to ease search from app creating timer

this would need https://github.com/jakartaee/enterprise-beans/pull/151


tadamski (approved):
LGTM

tadamski (changes requested):
@albfan Sorry, I approved this PR to quickly, but it would require further assessment.
I assume this PR is related to [EAPSUP-1336 (https://issues.redhat.com/browse/EAPSUP-1336). We would need to establish how this issue is going to be moved forward - an RFE process may be necessary. In all cases, we would need tests and upstream wildfly JIRA.

bstansberry:
This has been inactive for a long time and appears to be a draft, so I'm going to close it. Please feel free to reopen if there is further work. But, since this was first filed we have published https://docs.wildfly.org/wildfly-proposals/FEATURE_PROCESS.html to describe better how feature development in WildFly works. If, as appears to be the case, this is about a new feature, please do the steps in the feature process before reopening.

PR #17374, opened 2023-11-06
[WFLY-18533] Do not register the resteasy-validator-provider on deplo…

jamezp:
…yments if bean validation is not registered. Remove the resteasy-validator-provider module dependency from the resteasy-core module. Also remove the unneeded module dependencies on the subsystem.

https://issues.redhat.com/browse/WFLY-18533

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-core/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-core/main/module.xml
@@ -29,10 +29,8 @@
         <module name="jakarta.xml.bind.api"/>
         <module name="jakarta.ws.rs.api"/>
         <module name="com.ibm.async.asyncutil"/>
-        <module name="org.hibernate.validator" optional="true" services="import"/>
         <module name="org.jboss.logging"/>
         <module name="org.jboss.resteasy.resteasy-core-spi"/>
-        <module name="org.jboss.resteasy.resteasy-validator-provider" optional="true" services="export" export="true"/>
         <module name="org.reactivestreams"/>
     </dependencies>
 </module>
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -146,16 +146,6 @@
 
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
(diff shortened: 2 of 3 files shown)


ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/as/jaxrs/main/module.xml
```diff
@@ -41,19 +41,8 @@
...
-        <module name="org.jboss.resteasy.resteasy-client"/>
-        <module name="org.jboss.resteasy.resteasy-client-api"/>
         <module name="org.jboss.resteasy.resteasy-core"/>
         <module name="org.jboss.resteasy.resteasy-core-spi"/>
-        <module name="org.jboss.resteasy.resteasy-jaxb-provider" optional="true"/>
-        <module name="org.jboss.resteasy.resteasy-jackson2-provider" optional="true"/>
-        <module name="org.jboss.resteasy.resteasy-jsapi" optional="true"/>
-        <module name="org.jboss.resteasy.resteasy-multipart-provider" optional="true"/>
-        <module name="org.jboss.resteasy.resteasy-spring" optional="true" />
-        <module name="org.jboss.resteasy.resteasy-validator-provider" optional="true"/>
```
jamezp, line 56: The subsystem doesn't actually need these modules, but it does add them to deployments in a DUP. I wasn't sure if we needed to keep them as optional dependencies or not based on that.

kabir: @jamezp I think ResourceDefinition.registerAdditionalRuntimePackages() is intended for this. I see I am using that for Reactive Messaging

jamezp: Yes, sorry I missed that. I wasn't sure if it's better to keep the modules here or to add them in the `ResourceDefinition`. The `ResourceDefinition` seems more in the spirit of JBoss Modules and Galleon.

jamezp: I've just reverted the `module.xml` back for now. A follow up can be done later.

PR #17388, opened 2023-11-09
Allow 8 open dependabot PRs

bstansberry:
It's taking quite a while on average to get dependabot PRs merged or rejected, so let's give it more room to suggest things.


jamezp (approved):

bstansberry:
Thanks @jamezp

PR #17397, opened 2023-11-10
[WFLY-18736] Remove okhttp dependency

jasondlee:
https://issues.redhat.com/browse/WFLY-18735
https://issues.redhat.com/browse/WFLY-18736

This PR incorporates changes for WFLY-18735 and WFLY-18736.

[WFLY-18735] Upgrade to SmallRye OpenTelemetry 2.6.0


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/exporter/InMemorySpanExporter.java
```diff
@@ -39,8 +67,11 @@ public List<SpanData> getFinishedSpanItems(int spanCount) {
 
     public void assertSpanCount(int spanCount) {
         try {
-            Awaitility.await().pollDelay(3, SECONDS).atMost(30, SECONDS)
+            System.err.println("***** Waiting " + pollDelay + " for spans to be ready");
+            Awaitility.await().pollDelay(pollDelay)
+                    .atMost(pollDelay.multipliedBy(2))
                     .untilAsserted(() -> Assert.assertEquals(spanCount, finishedSpanItems.size()));
+            System.err.println("***** All spans found");
```
bstansberry, line 74: Please either comment out all these System.err calls or use a logger (which will put them in target/test.log).

Generally, commenting out is better as logging adds CI burden and it adds up.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemAdd.java
```diff
@@ -34,25 +38,41 @@ class OpenTelemetrySubsystemAdd extends AbstractBoottimeAddStepHandler {
...
 
+        String exporter = OpenTelemetrySubsystemDefinition.EXPORTER.resolveModelAttribute(context, model).asString();
+
+        if (EXPORTER_JAEGER.equals(exporter)) {
+            if (context.isNormalServer()) {
+                throw new OperationFailedException(OTEL_LOGGER.jaegerIsNoLongerSupported());
+            } else {
+                OTEL_LOGGER.warn(OTEL_LOGGER.jaegerIsNoLongerSupported());
+            }
+        }
```
bstansberry, line 49: Please convert this to a static method so it can be used for write-attribute as well. And then OpenTelemetrySubsystemDefinition would need to override registerAttributes with a new implementation that handles EXPORTER differently with a handler that calls the static function.

BTW, this comment is what led me to start this thread: https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Write.20attribute.20handler.20registration

observability/opentelemetry-api/src/main/java/org/wildfly/extension/opentelemetry/api/WildFlyOpenTelemetryConfig.java
```diff
@@ -19,38 +20,56 @@ public final class WildFlyOpenTelemetryConfig implements OpenTelemetryConfig {
...
+        addValue(config, OTEL_METRICS_EXPORTER, "none");
+
+        if (exporter.equals("otlp")) {
+            addValue(config, OTEL_EXPORTER_OTLP_ENDPOINT, endpoint);
+            addValue(config, OTEL_EXPORTER_OTLP_PROTOCOL, "grpc");
+            addValue(config, OTEL_EXPORTER_OTLP_TIMEOUT, exportTimeout);
+        } else {
+            Logger.getLogger(WildFlyOpenTelemetryConfig.class)
+                    .warn("An unexpected exporter type was found: " +
+                            exporter + ". Using the default.");
```
bstansberry, line 64: This should fail with an IAE. The Stage.MODEL code should correctly handle parameter validation and this class should reject any mistakes it makes.

Or, you could remove the exporter param altogether and hard-code the otlp behavior.

bstansberry (changes requested):
@jasondlee This requires a management API version bump, a new schema (as the default exporter changed, plus the attribute documentation can be updated, and it needs a transformer to deal with the https://docs.wildfly.org/30/Extending_WildFly.html#attribute-has-a-different-default-value case.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/exporter/InMemorySpanExporter.java
jasondlee: That wasn't supposed to make it to GH. :) I'll clean that up.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemAdd.java
bstansberry: To be more clear, I'm referring to extracting https://github.com/wildfly/wildfly/pull/17397/files#diff-72d387b5d42a4c7352e43b0c764284eb154c6e34e15d6e9009d6d4867089a943R43-R49 out into a method.

jasondlee: Done.

observability/opentelemetry-api/src/main/java/org/wildfly/extension/opentelemetry/api/WildFlyOpenTelemetryConfig.java
jasondlee: Done

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
```diff
@@ -21,15 +21,14 @@
                         </xs:documentation>
```
bstansberry, line 21: You need to create a new schema (and parser.)

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemDefinition.java
```diff
@@ -163,6 +170,20 @@ protected OpenTelemetrySubsystemDefinition() {
         );
     }
 
+    public static String getExporter(OperationContext context, ModelNode model) throws OperationFailedException {
```
bstansberry, line 173: You need to deal with the write attribute handler as well. I'll send an example in a few minutes.

bstansberry (changes requested):

bstansberry:
@jasondlee See https://github.com/jasondlee/wildfly/pull/14

There's more to do (new schema version, test file for subsystem tests, transformer tests, etc) but that covers the write-attribute validation I was talking about, fixes the 'IAE' problem (java.lang.IllegalArgumentException), sorts some other things.

observability/opentelemetry/src/main/resources/schema/wildfly-opentelemetry_1_0.xsd
jasondlee: I remembered about that part after I left the house. On it now.

bstansberry (approved):
Looks good, but please squash. I don't care about my name on my commits.

bstansberry:
Thanks, @jasondlee

PR #17399, opened 2023-11-13
WFLY-18740 On cache writes, Infinispan store=hotrod throws ISE: Only byte[] instances are supported currently

pferraro:
https://issues.redhat.com/browse/WFLY-18740


bstansberry:
Thanks @pferraro

PR #17402, opened 2023-11-14
Fix for WFLY-18741, Add a github action shared file to build and test WildFly

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-18741

An example of the Galleon plugins Github Actions: https://github.com/jfdenise/galleon-plugins/blob/main/.github/workflows/ci.yml
An Action run: https://github.com/jfdenise/galleon-plugins/actions/runs/6862670363

Another Action run example from Galleon project this time (just validate that build + -Dts.layers tests are passing): https://github.com/jfdenise/galleon/actions/runs/6864936818

### DIFF
--- a/.github/workflows/shared-wildfly-build-and-test.yml
+++ b/.github/workflows/shared-wildfly-build-and-test.yml
@@ -1,0 +1,148 @@
+# A workflow that can be called from other workflows as a way to build and test a SNAPSHOT of WildFly with overridden dependencies.
+
+# Input parameters:
+# * test-arguments: An optional argument allowing to provide arguments to configure testing. Note that '-DallTests' should not be passed as not all tests will work on GitHub Actions.
+# * build-arguments: Arguments used during build. That is where you can override the version of the dependency.
+# * maven-repo-path: The path to a tar.gz file of a maven repo that contains the built dependency(ies) you are using.
+# * maven-repo-name: The name that identifies the maven repo to be downloaded.
+# * java-versions: A string containing a JSON array of JDK versions. For example: "['11', '17']"
+# * os: A string containing a JSON array of operating systems. For example: "['ubuntu-latest', 'windows-latest']"
+# 
+# Example Usage (WildFly main branch is built and tested with a custom build of a dependency named 'my-dependency'.
+#
+#jobs:  
+#  my-dependency-build:
+#    name: my-dependency
+#    runs-on: ubuntu-latest
+#    outputs:
+#      my-dependency-version: ${{ steps.version.outputs.my-dependency-version }}
+#    strategy:
+#      fail-fast: false
+#    steps:
+#      - uses: actions/checkout@v4
+#      - name: Set up JDK
+#        uses: actions/setup-java@v3
+#        with:
+#          java-version: '11'
+#          distribution: 'temurin'
+#          cache: 'maven'
+#      - name: Build and Test my-dependency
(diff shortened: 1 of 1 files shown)


.github/workflows/shared-wildfly-build-and-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Build and Test WildFly SNAPSHOT ${{ inputs.build-arguments }}  ${{ inputs.test-arguments }}
+        run: mvn -U -B -ntp clean install ${{ inputs.build-arguments }} ${{ inputs.test-arguments }}
+        shell: bash
+        # We need this due to long file names on Windows: https://github.com/actions/upload-artifact/issues/240
+      - name: Find reports
+        if: failure()
+        run: |
+          {
+            echo 'SUREFIRE_REPORTS<<EOF'
+            find . -path '**/surefire-reports/*.txt'
```
jamezp, line 73: Minor, but we should leave the *.txt off IMO. I've been capturing the `.xml` files too as they can be helpful. I've written a tool that parses them and gives information to easily get data from the failed tests.

.github/workflows/shared-wildfly-build-and-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.maven-repo-name }}
+          path: .
+      - name: Extract Maven Repo
+        shell: bash
+        run: |
+         tar -xzf ${{ inputs.maven-repo-path }} -C ~
+      - name: Build and Test WildFly SNAPSHOT ${{ inputs.build-arguments }}  ${{ inputs.test-arguments }}
+        run: mvn -U -B -ntp clean install ${{ inputs.build-arguments }} ${{ inputs.test-arguments }}
```
jamezp, line 65: Good call on the `-ntp` argument. I should start using that more :)

.github/workflows/shared-wildfly-build-and-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      java-versions:
+        description: "JDK versions array"
+        required: true
+        default: ""
+        type: string
+      os:
+        description: "OS array"
+        required: true
+        default: ""
+        type: string
```
jamezp, line 35: We should remove the default here if they are required.

jamezp (changes requested):
I think overall this makes sense. One small request, can you add a comment on the top that shows how to use this for a reference? Something similar to what is here https://github.com/wildfly/wildfly/blob/dbf1071f017d21dab7ee5c86aa20302f7586c050/.github/workflows/shared-wildfly-build.yml#L13-L40.

We also might want to note that things like `-DallTests` should not be passed as not all tests will work on GitHub Actions.

.github/workflows/shared-wildfly-build-and-test.yml
jfdenise: Done.

.github/workflows/shared-wildfly-build-and-test.yml
jfdenise: Done

jfdenise:
@jamezp , thank-you I added usage, notes and followed your suggestions.

jamezp (approved):

bstansberry:
Thanks @jfdenise and @jamezp

PR #17403, opened 2023-11-14
[WFLY-18743] Change NoAuditLogTestCase class to use the system default encoding to read the log file

Dkafetzis:
https://issues.redhat.com/browse/WFLY-18743


jamezp:
Have we seen issues with this somewhere?

Dkafetzis:
@jamezp Haven't found any issues of it for linux since UTF-8 is the default file encoding for it, but the test tends to fail for older versions of windows and also recently it started failing on CI (again for windows test runs) for jboss as per these issues [issue1](https://issues.redhat.com/browse/JBEAP-25744) and [issue2](https://issues.redhat.com/browse/JBEAP-25745)

jamezp (approved):

bstansberry:
Thanks @Dkafetzis and @jamezp

PR #17408, opened 2023-11-15
[WFLY-18742] Provisioning micrometer and opentelemetry layers issue

jasondlee:
https://issues.redhat.com/browse/WFLY-18742

Add module dependency to provide access to missing class in failing test


bstansberry:
Thanks @jasondlee

PR #17411, opened 2023-11-16
Fix for WFLY-18746, Revisit telemetry layers inclusion rules

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-18746

* microprofile-telemetry shouldn't contain rules related to opentelemetry API usage. This layer is provisioned if all its dependencies are discovered.
* opentelemetry and microprofile-telemetry layers are not add-ons.


jasondlee (approved):

bstansberry:
Thanks @jfdenise and @jasondlee

PR #17044, opened 2023-07-25
[WFLY-18163] Add org.wildfly.extension.elytron.jaas-realm module exclusion 

skyllarr:
https://issues.redhat.com/browse/WFLY-18163

This PR is required so that the CI on https://github.com/wildfly/wildfly-core/pull/5587 passes

This is a **feature** https://issues.redhat.com/browse/WFLY-18163

PR #17417, opened 2023-11-17
[WFLY-18163] Add tests for JaasSecurityRealm via a custom-realm resource

skyllarr:
https://issues.redhat.com/browse/WFLY-18163
Related to https://github.com/wildfly/wildfly/pull/17044

Requires:
https://issues.redhat.com/browse/WFCORE-6411

### DIFF
--- a/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
+++ b/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
@@ -1,0 +1,84 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.wildfly.test.integration.elytron.realm;
+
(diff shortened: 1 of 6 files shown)


testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * A custom {@link CallbackHandler} used in the JAAS security realm tests. It implements the
+ * {@code setSecurityInfo} method that has been historically used to populate custom handlers. Also, its {@code handle}
+ * implementation will handle any kind of credential by calling {@code toString} and then {@code toCharArray} on the opaque
+ * object.
+ *
+ * @author <a href="mailto:sguilhen@redhat.com">Stefan Guilhen</a>
+ */
+public class TestCallbackHandler implements CallbackHandler {
```
darranl, line 42: I suspect a lot of the time this is all the CallbackHandler will need to do - is there any reason we don't add a simple CallBackHandler ourselves that can be used for this scenario?

darranl (approved):
My comment is just for something possibly in the future, otherwise looks good.

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public void setSecurityInfo(final Principal principal, final Object evidence) {
+        this.principal = principal;
+        this.evidence = (Evidence) evidence;
+    }
+
+    @Override
+    public void handle(Callback[] callbacks) throws UnsupportedCallbackException {
+        if (callbacks == null)
+            throw new IllegalArgumentException("The callbacks argument cannot be null");
```
darranl, line 64: Minor minor issue, would prefer some parenthesis here.

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
Skyllarr: @darranl We do have a simple default callback handler implementation that gets used when none is provided by the user: https://github.com/wildfly-security/wildfly-elytron/blob/2.x/auth/realm/base/src/main/java/org/wildfly/security/auth/realm/JaasSecurityRealm.java#L309 I added the custom callback handler here just so it gets tested as a parameter to the realm

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/realm/TestCallbackHandler.java
Skyllarr: @darranl Fixed. Thanks for review!

fjuma (approved):

yersan:
Superseded by https://github.com/wildfly/wildfly/pull/17622, already merged

PR #17419, opened 2023-11-17
[WFLY-18764] Update HostExcludesTestCase configuration to work with WF31

skyllarr:
https://issues.redhat.com/browse/WFLY-18764


yersan (approved):
This looks good to me. However, notice this is an optional task when there have been no changes in the list of extensions during the latest release development cycle. The test will fail only if we have added new extensions or if we want to add new ones and we have not moved it on to the next exclusion version.

bstansberry:
Thanks @Skyllarr and @yersan

PR #17420, opened 2023-11-17
[WFLY-18763] Add an example standalone-core-microprofile.xml example configuration.

jamezp:
https://issues.redhat.com/browse/WFLY-18763

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -91,6 +91,9 @@
                                     <groupId>${full.maven.groupId}</groupId>
                                     <artifactId>wildfly-galleon-pack</artifactId>
                                     <version>${full.maven.version}</version>
+                                    <included-packages>
+                                        <name>docs.examples.configs</name>
+                                    </included-packages>
                                 </feature-pack>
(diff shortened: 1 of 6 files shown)


bstansberry (changes requested):
@jamezp This package should be in galleon-pack/galleon-shared, not galleon-local. Otherwise it will not be included in the wildfly-preview feature pack.

bstansberry:
Replying to myself -- no it shouldn't be in galleon-shared because the wildfly-preview FP itself has the same package with its own WFP-specific tasks.xml file. So the fix is to duplicate the tasks.xml 'config' element into the WFP file.

bstansberry (approved):

darranl:
@jamezp The test failure looks related

jamezp:
@darranl It was indeed. This should fix it https://github.com/wildfly/wildfly/pull/17420/files#diff-56647cdad56f631ce6554abcb6ab804f04f61cabed523e74b29d1b6c996dafcdR334-R336

PR #17422, opened 2023-11-20
[WFLY-18765] Add Locale parameter while calling toUpperCase and toLowerCase methods

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-18765

[WFLY-18765] Add Locale parameter while calling toUpperCase and toLowerCase methods

### DIFF
--- a/jdr/src/main/java/org/jboss/as/jdr/util/JdrZipFile.java
+++ b/jdr/src/main/java/org/jboss/as/jdr/util/JdrZipFile.java
@@ -12,6 +12,7 @@ import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.Locale;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
 import java.util.zip.ZipOutputStream;
@@ -151,7 +152,7 @@ public class JdrZipFile {
 
     public String getProductDirName() {
         if(this.productDirName == null)
-            this.productDirName = String.format("%s-%s", this.env.getProductName().replace(" ", "_").toLowerCase(), this.env.getProductVersion().split("\\.")[0]);
+            this.productDirName = String.format("%s-%s", this.env.getProductName().replace(" ", "_").toLowerCase(Locale.getDefault()), this.env.getProductVersion().split("\\.")[0]);
         return this.productDirName;
     }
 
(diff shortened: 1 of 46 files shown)


rhusar (approved):
The mod_cluster bit (and others) look good.

Using e.g. turkish locale would break session stickiness if using param since `JSESSIONID` would become

```
[rhusar@ribera ~]$ java ... -Duser.language=tr -Duser.country=TR -Duser.variant=TR Scratch
jsessıonıd
```

soul2zimate:
Hi @bstansberry , May I ask you review and approval on this ? I'd like to backport this to downstream.

jdr/src/main/java/org/jboss/as/jdr/util/JdrZipFile.java
```diff
@@ -151,7 +152,7 @@ public void addAsString(InputStream stream, String path) throws Exception {
 
     public String getProductDirName() {
         if(this.productDirName == null)
-            this.productDirName = String.format("%s-%s", this.env.getProductName().replace(" ", "_").toLowerCase(), this.env.getProductVersion().split("\\.")[0]);
+            this.productDirName = String.format("%s-%s", this.env.getProductName().replace(" ", "_").toLowerCase(Locale.getDefault()), this.env.getProductVersion().split("\\.")[0]);
```
bstansberry, line 155: For WF and EAP, Locale.ENGLISH is more correct as we're not going to use anything else for this.

But WF is extensible, so it's possible community users are using something else, so +1 to this. It's the same behavior as what we currently have.

jasondlee: Why is this the only place that calls (redundantly, as Brian noted) `Locale.getDefault()`?

bstansberry (approved):
@soul2zimate I've left some comments in case they spark any thoughts from you, but I think this is ok. Thanks!

jdr/src/main/java/org/jboss/as/jdr/util/JdrZipFile.java
bstansberry: I think explicitly saying 'Locale.getDefault()' is preferable to saying nothing, as it let's us distinguish an intentional choice from the common case of just forgetting to specify a locale.

petrberan:
Hi @bstansberry if those minor changes are not preventing merging the PR and we have donwstream issue for this one, would you mind merging it? If you'd like those changes handled, please let us know

soul2zimate:
Hi @bstansberry,  I don't really have any more to input for this PR. Since this is approved, could you merge it ? Thanks.

darranl:
Large PRs like this are really not great to coordinate PR reviews, this has touched many subsystems and GitHub is not great at indicating if an approval is for a whole PR or a subset so you risk getting stuck waiting for the future reviews.

bstansberry (approved):

jasondlee (approved):

bstansberry:
Thanks @soul2zimate, @rhusar and @jasondlee

soul2zimate:
Thank you all for the reviews.

PR #17441, opened 2023-11-23
[WFLY-17180][JBJCA-1402] added validation-query-timeout attribute to …

istudens:
…datasources

Issue: https://issues.redhat.com/browse/WFLY-17180
Related issue: https://issues.redhat.com/browse/JBJCA-1402


istudens:
Marking this as Draft since it needs an IronJacamar upgrade which has not been released yet.

bstansberry:
@istudens This has been inactive for a long time so I'm going to close it. Please feel free to reopen if work begins again. It looks like the IJ part of this was done quite a while ago.

This is a feature though, so before reopening please check the process requirements at https://docs.wildfly.org/wildfly-proposals/FEATURE_PROCESS.html and do the necessary preparatory steps. I suggest discussing with Tomek to pick a target stability level.

PR #17442, opened 2023-11-24
WFLY-18782 wildfly-model-test & commons-text should only be present in test scope

TomasHofman:
Issue: https://issues.redhat.com/browse/WFLY-18782

The org.wildfly.core:wildfly-model-test dependency in the testsuite/integration/ module should have a test scope.

The org.apache.commons:commons-text dependency should be excluded from the org.apache.activemq:artemis-server dependency, so that it's eliminated from the compile scope dependency tree. This module is not present in our jboss-modules so it's not expected to be needed at runtime.

...Waiting for CI results...

### DIFF
--- a/testsuite/integration/pom.xml
+++ b/testsuite/integration/pom.xml
@@ -460,6 +460,7 @@
         <dependency>
            <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-model-test</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.wildfly.core</groupId>


messaging-activemq/subsystem/pom.xml
```diff
@@ -98,6 +98,12 @@
         <dependency>
             <groupId>org.apache.activemq</groupId>
             <artifactId>artemis-server</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-text</artifactId>
+                </exclusion>
+            </exclusions>
```
ehsavoie, line 106: I'm trying to understand why you are excluding it form artemis server as I can't find any direct dependecy on it. Maybe the right artefact to exclude is org.apache.commons:commons-configuration2

TomasHofman: Hm, it's not there in this version of Artemis. I didn't expect there to be difference from EAP 8. I removed this bit.

ehsavoie: :+1:

ehsavoie (approved):

PR #17445, opened 2023-11-27
[WFLY-18499] adds resteasy mp client useful artifacts to WildFly BOMs

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18499
Adds deps needed by apps, such as MP REST Client quickstart

### DIFF
--- a/boms/user/microprofile/pom.xml
+++ b/boms/user/microprofile/pom.xml
@@ -179,6 +179,16 @@
                                     <groupId>io.smallrye.reactive</groupId>
                                     <artifactId>smallrye-reactive-messaging-kafka-api</artifactId>
                                 </dependency>
+                                <!-- include RESTeasy MP REST Client -->
+                                <dependency>
+                                    <groupId>org.jboss.resteasy.microprofile</groupId>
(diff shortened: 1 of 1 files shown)


boms/user/ee/ee-with-tools/pom.xml
```diff
@@ -134,6 +134,10 @@
                                     <groupId>org.jboss.arquillian.protocol</groupId>
                                     <artifactId>arquillian-protocol-servlet-jakarta</artifactId>
                                 </dependency>
+                                <dependency>
+                                    <groupId>org.jboss.resteasy</groupId>
+                                    <artifactId>resteasy-undertow-cdi</artifactId>
+                                </dependency>
```
jamezp, line 140: We should not add this. This is for the Jakarta REST `SeBootstrap` which is for SE only environments and should not be used in a Jakarta EE Container.

jamezp (changes requested):

jamezp (approved):

bstansberry:
Thanks @emmartins and @jamezp

PR #17437, opened 2023-11-22
[WFLY-18779] Upgrade WildFly Core to 23.0.0.Beta2

yersan:
Jira issue:

https://issues.redhat.com/browse/WFLY-18779

---

Tag: https://github.com/wildfly/wildfly-core/releases/tag/23.0.0.Beta2
Diff: https://github.com/wildfly/wildfly-core/compare/23.0.0.Beta1...23.0.0.Beta2

---

<details>
<h2>        Bug
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6578'>WFCORE-6578</a>] -         [CVE-2023-3171] WildFly heap exhaustion via deserialization
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6591'>WFCORE-6591</a>] -         Core modules still have dependencies on org.jboss.as.security
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6592'>WFCORE-6592</a>] -         NPE in ThreadFactoryService
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6597'>WFCORE-6597</a>] -         WFCORE-4296 - Illegal reflective access by org.wildfly.extension.elytron.SSLDefinitions when started by ps1 script
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6601'>WFCORE-6601</a>] -         ModuleSpecification.setSystemDependency does not reset the allDependencies list
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6604'>WFCORE-6604</a>] -         JMX can&#39;t set default value for an attribute of ModelType.LIST
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6607'>WFCORE-6607</a>] -         ModuleSpecification.addSystemDependency treats a duplicate add as an excluded dependency
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6608'>WFCORE-6608</a>] -         Inconsistent user dependency views in ModuleSpecification
</li>
</ul>
    
<h2>        Task
</h2>
<ul>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6332'>WFCORE-6332</a>] -         Better handling of licenses
</li>
<li>[<a href='https://issues.redhat.com/browse/WFCORE-6599'>WFCORE-6599</a>] -         Control the maven repos used by dependabot
</li>
[...]

PR #17448, opened 2023-11-27
WFLY-18788 Replace Consumers/Functions with utility methods from wildfly-common

pferraro:
https://issues.redhat.com/browse/WFLY-18788

Requires https://github.com/wildfly/wildfly/pull/17437


bstansberry:
Thanks @pferraro

PR #17451, opened 2023-11-29
[WFLY-18799] Fix rendering of distributed-realm and failover-realm and alphabetize security realms table

skyllarr:
https://issues.redhat.com/browse/WFLY-18799

### DIFF
--- a/docs/src/main/asciidoc/_elytron/Elytron_Subsystem.adoc
+++ b/docs/src/main/asciidoc/_elytron/Elytron_Subsystem.adoc
@@ -260,11 +260,18 @@ not be exposed to manage the realm. However other services that depend
 on the realm will still be able to perform a type check and cast to gain
 access to the modification API.
 
+|<<distributed-security-realm, distributed-realm>> |A security realm definition for authentication and authorization identities distributed between multiple security realms.
+
+|<<failover-security-realm, failover-realm>> |A realm definition that is an aggregation of two realms, one for default behaviour and
+second for cases when first realm is unavailable.
+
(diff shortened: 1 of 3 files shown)


bstansberry (changes requested):
@Skyllarr This doesn't work because the docs you are linking to are not incorporated into the overall guide; they are just files in the source.

The list of includes in https://raw.githubusercontent.com/wildfly/wildfly/main/docs/src/main/asciidoc/_elytron/Subsystem_Component_Documentation.adoc needs to include all the adoc files.

Also, I think it would be good to alphabetize the realms table (https://docs.wildfly.org/30/WildFly_Elytron_Security.html#realms) and any other tables that are not alphabetized. Doing them all is reasonably out of scope for this JIRA, but I think at least fixing the 'Realms' table as part of this would be good.

Skyllarr:
> @Skyllarr This doesn't work because the docs you are linking to are not incorporated into the overall guide; they are just files in the source.
> 
> The list of includes in https://raw.githubusercontent.com/wildfly/wildfly/main/docs/src/main/asciidoc/_elytron/Subsystem_Component_Documentation.adoc needs to include all the adoc files.
> 
> Also, I think it would be good to alphabetize the realms table (https://docs.wildfly.org/30/WildFly_Elytron_Security.html#realms) and any other tables that are not alphabetized. Doing them all is reasonably out of scope for this JIRA, but I think at least fixing the 'Realms' table as part of this would be good.

@bstansberry Thank you! I fixed it and alphabetized the security realms table

bstansberry (approved):

bstansberry:
Thanks @Skyllarr

PR #17457, opened 2023-11-30
Dependabot: Ignore dependencies exclusive to Hibernate Search

yrodiere:
See https://github.com/wildfly/wildfly/pull/17455#issuecomment-1832711680

@bstansberry Do I need a JIRA even for this?


bstansberry:
Thanks @yrodiere.

No, we don't need a JIRA.

I mostly care about JIRAs for stuff that's interesting for release note readers (so production code changes). Secondarily for explanations of why stuff was done when it's not obvious from the code.

PR #17469, opened 2023-12-04
[WFLY-18820]: Queue creation might fail if auto-create-addresses is set to false.

ehsavoie:
* Creating address before creating queue.

Issue: https://issues.redhat.com/browse/WFLY-18820

### DIFF
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/DestinationConfiguration.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/DestinationConfiguration.java
@@ -84,13 +84,24 @@ public class DestinationConfiguration {
             connection.start();
             QueueRequestor requestor = new QueueRequestor((QueueSession) session, managementQueue);
             Message m = session.createMessage();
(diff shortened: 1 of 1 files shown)


messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/DestinationConfiguration.java
```diff
@@ -84,12 +84,24 @@ public void createQueue(ConnectionFactory cf, Queue managementQueue, String queu
...
+            Message reply = requestor.request(m);
+            ROOT_LOGGER.infof("Creating address %s returned %s", queueName, reply);
+            requestor.close();
+            if (!reply.getBooleanProperty("_AMQ_OperationSucceeded")) {
+                String body = reply.getBody(String.class);
+                if (!destinationAlreadyExist(body)) {
+                    throw ROOT_LOGGER.remoteDestinationCreationFailed(queueName, body);
+                }
+            }
+            ROOT_LOGGER.infof("Address %s has been created", queueName);
```
bstansberry, line 97: @ehsavoie The two infof calls need to be debugf.

bstansberry (changes requested):

bstansberry (approved):

bstansberry:
Thanks @ehsavoie

PR #17470, opened 2023-12-05
[WFLY-18817] Test that provisioning channel information is logged at boot

bstansberry:
https://issues.redhat.com/browse/WFLY-18817

This builds on #17248 by adding an additional test method to the test added there. But I needed to rework the way provisioning is done to get it to work.

Requires https://github.com/wildfly/wildfly-core/pull/5724


spyrkob:
@bstansberry do you foresee more tests requiring prospero in the Wildfly testsuite? I'm wondering if it would be better to create a mock prospero module that could be added to the provisioned server to test the integration up to the API level and test the implementation in prospero codebase instead. This way we could avoid adding dependency on prospero to Wildfly codebase

bstansberry:
@spyrkob Potentially, yes. The management integration @yersan did could be tested in WF.

Some kind of integration TCK for that that could executed in both places would make sense.

spyrkob:
> https://github.com/wildfly/wildfly/pull/17470/files#diff-ab7939563ab2ed2a32ec4f069f5755fe052e9555777de64f3c0dbe8489bbdde3R455-R467 -- otherwise it fails to resolve org.jboss.universe:community-universe and org.jboss.universe.provider:wildfly-providers when installing the Prospero FP.

The problem here is that when creating the prospero-wildfly-galleon-pack feature pack, the reference to `org.wildfly:wildfly-ee-galleon-pack` becomes `<dependency location="wildfly-ee@maven(org.jboss.universe:community-universe):current#30.0.0.Final">`. To resolve this back to maven coordinates, Galleon requires universe artifacts. Maybe we could add an option to the wildfly-galleon-maven-plugin to preserve the maven GA of the dependencies? Alternatively the wildfly channel could define the universe artifacts.

> https://github.com/wildfly/wildfly/pull/17470/files#diff-ab7939563ab2ed2a32ec4f069f5755fe052e9555777de64f3c0dbe8489bbdde3R440 and the same in all the other executions. Otherwise each execution hangs for 15 secs after CliMain completes and then exec-maven-plugin warns about 2 daemon threads that don't respond to interrupt.

This seems to be an issue with jgit implementation. It creates some daemon thread in an executor that is never shutdown. I created https://github.com/wildfly-extras/prospero/issues/517 to track it and I'll try to submit a fix upstream.

bstansberry:
Thanks for the analysis, @spyrkob. I added the test I need to https://github.com/wildfly/wildfly-core/pull/5724, and the other uses of integrated-prospero are also there.

I started a zulip thread where we can talk about how to handle the universe thing:

https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Defining.20Galleon.20universe.20artifacts.20in.20a.20WF.20channel

So, given all that I'll close this. I'll leave the branch for a while though, in case it's useful in some way.

I don't see a need for this test in WF, although something like it in prospero's TS might make sense, as I surfaced a few issues by doing it.

PR #17472, opened 2023-12-05
WFLY-18823 Fix Flaky AbstractValidationUnitTest

alexbadia1:
Fixes: https://issues.redhat.com/browse/WFLY-18823

In AbstractValidationUnitTes list schema files in a deterministic order and preserve order with LinkedHashMaps for JBOSS_SCHEMAS_MAP and CURRENT_JBOSS_SCHEMAS_MAP

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/distribution/validation/AbstractValidationUnitTest.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/distribution/validation/AbstractValidationUnitTest.java
@@ -25,8 +25,10 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


testsuite/shared/src/main/java/org/wildfly/test/distribution/validation/AbstractValidationUnitTest.java
```diff
@@ -152,6 +154,7 @@ public class AbstractValidationUnitTest {
             final File schemaDir = new File(JBOSS_DIST_DIR, SCHEMAS_LOCATION);
 
             final File[] xsds = schemaDir.listFiles(new SchemaFilter(EXCLUDED_SCHEMA_FILES.toArray(new String[0])));
+            Arrays.sort(xsds);
```
bstansberry, line 157: @alexbadia1 How does this address the JIRA description statement that:

"the order of the `SCHEMA_SOURCES` matters since some schemas may be dependent on each other, (i.e. one schema importing another)" ?

It makes the order deterministic, but I'm not sure why lexicographical ordering of path names results in avoiding incorrect orders between dependent schemas.

I think your PR's fine; being deterministic is good, and if it passes we know the lexicographical sort order is an ok won to stick to.

bstansberry:
Thanks @alexbadia1!

PR #17479, opened 2023-12-11
[WFLY-18818]:Print log info to let user know the cxf dependenc…

jimma:
https://issues.redhat.com/browse/WFLY-18818

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSClassVerificationProcessor.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSClassVerificationProcessor.java
@@ -62,7 +62,9 @@ public class WSClassVerificationProcessor implements DeploymentUnitProcessor {
             for (AbstractEndpoint ejbEndpoint : wsDeployment.getEjbEndpoints()) {
                 verifyEndpoint(ejbEndpoint, moduleClassLoader, deploymentReflectionIndex);
             }
-            verifyApacheCXFModuleDependencyRequirement(unit);
+            if (!hasCxfModuleDependency(unit)) {
+                checkCXFModuleDependency(unit);
+            }
         }
     }
 
@@ -99,28 +101,6 @@ public class WSClassVerificationProcessor implements DeploymentUnitProcessor {
             throw WSLogger.ROOT_LOGGER.declaredEndpointInterfaceClassNotFound(endpointInterfaceClassName, endpointClass);
         }
     }
-
-    private void verifyApacheCXFModuleDependencyRequirement(DeploymentUnit unit) {
-        if (!hasCxfModuleDependency(unit)) {
-            //notify user if he clearly forgot the CXF module dependency
-            final CompositeIndex index = unit.getAttachment(Attachments.COMPOSITE_ANNOTATION_INDEX);
-            final DotName[] dotNames = {WEB_SERVICE_ANNOTATION, WEB_SERVICE_PROVIDER_ANNOTATION};
-            for (final DotName dotName : dotNames) {
-                for (AnnotationInstance ai : index.getAnnotations(dotName)) {
-                    AnnotationTarget at = ai.target();
-                    if (at instanceof ClassInfo) {
-                        final ClassInfo clazz = (ClassInfo)ai.target();
-                        for (DotName dn : clazz.annotationsMap().keySet()) {
-                            if (dn.toString().startsWith("org.apache.cxf")) {
-                                WSLogger.ROOT_LOGGER.missingModuleDependency(dn.toString(), clazz.name().toString(), "org.apache.cxf");
-                            }
-                        }
-                    }
-                }
-            }
-        }
-    }
-
     static boolean hasCxfModuleDependency(DeploymentUnit unit) {
         final ModuleSpecification moduleSpec = unit.getAttachment(Attachments.MODULE_SPECIFICATION);
         for (ModuleDependency dep : moduleSpec.getUserDependenciesSet()) {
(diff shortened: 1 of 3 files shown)


bstansberry:
@jimma Any logging that was reasonably likely to be unnecessary would need to be INFO or lower, as many users treat WARN as something they have to deal with and get annoyed about anything incorrect.

jamezp:
If I'm understanding correctly a warning might be right here. What his is doing is checking if there is a module dependency on `org.apache.cxf`, `org.apache.cxf.impl` or `org.jboss.ws.cxf.jbossws-cxf-client`. If the dependency is not there and there is an annotation with a package prefix of `org.apache.cxf`, it warns that a module dependency is required for this to work.

I'm not sure we need to log a warning for every annotation, but it does feel a warning or even error might be appropriate. I assume the deployment will not work as expected.

jimma:
> @jimma Any logging that was reasonably likely to be unnecessary would need to be INFO or lower, as many users treat WARN as something they have to deal with and get annoyed about anything incorrect.

@bstansberry This changed to not check the cxf module dependency in parent deployment and sibling deployment because of the ModuleSpecification isn't thread safe. I think this covers most of the dependency declaration. If this deployment processor doesn't find the cxf module dependency in the current deployment, it prints some warning message to ask user checks manually. This warning message is kind of reminder or request for user to check the module dependency is correctly configured. Even we can check the cxf module dependency in parent and sibling deployment after we fix [WFCORE-6639](https://issues.redhat.com/browse/WFCORE-6639), the warning message can still be incorrect when user configure this cxf dependency as a global module.

jimma:
> If I'm understanding correctly a warning might be right here. What his is doing is checking if there is a module dependency on `org.apache.cxf`, `org.apache.cxf.impl` or `org.jboss.ws.cxf.jbossws-cxf-client`. If the dependency is not there and there is an annotation with a package prefix of `org.apache.cxf`, it warns that a module dependency is required for this to work.
> 
That's correct. I only removed the check the module dependency from parent and sibling deployment because of the  ModuleSpecification thread safe issue.  

> I'm not sure we need to log a warning for every annotation, but it does feel a warning or even error might be appropriate. I assume the deployment will not work as expected.
Good point about log a waring message for every annotation @jamezp. This can be improved to print only one warning message as it is only one module dependency is required for this deployment. 
If there isn't cxf module dependency in the current deployment and it is in parent ear or it is in the sibling deployment, it still works, but it beyonds the current deployment processor duty and we can print a warning message to remind user it needs this cxf dependency and user should make sure there is this module dependency whether from parent, sibling or the global module.

webservices/server-integration/src/main/java/org/jboss/as/webservices/deployers/WSClassVerificationProcessor.java
```diff
@@ -99,68 +101,32 @@ void verifyJwsEndpoint(final Class<?> endpointClass, final WebService webService
...
+        final DotName[] dotNames = {WEB_SERVICE_ANNOTATION, WEB_SERVICE_PROVIDER_ANNOTATION};
+        for (final DotName dotName : dotNames) {
+            for (AnnotationInstance ai : index.getAnnotations(dotName)) {
+                AnnotationTarget at = ai.target();
+                if (at instanceof ClassInfo) {
+                    final ClassInfo clazz = (ClassInfo)ai.target();
+                    for (DotName dn : clazz.annotationsMap().keySet()) {
+                        if (dn.toString().startsWith("org.apache.cxf")) {
+                            WSLogger.ROOT_LOGGER.checkModuleDependency(dn.toString(), clazz.name().toString(), "org.apache.cxf");
+                        }
```
jamezp, line 126: I would just return here after you log.

jimma: Done.

bstansberry:
@jimma Leaving this message at WARN will essentially regress the https://issues.redhat.com/browse/WFLY-13832 / https://issues.redhat.com/browse/JBEAP-19682 bug fix.

Changing it to INFO avoids that; otherwise we can go with https://issues.redhat.com/browse/WFCORE-7100 / https://github.com/wildfly/wildfly-core/pull/6285 and bend the rule about how to use ModuleSpecification.

jimma:
@bstansberry @jamezp  Fixed. Please review again. Thanks.

webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
```diff
@@ -225,19 +225,9 @@ public interface WSLogger extends BasicLogger {
...
-
     @Message(id = 54, value = "Config %s: %s handler chain with id %s doesn't exist")
     OperationFailedException missingHandlerChain(String configName, String handlerChainType, String handlerChainId);
 
-    // @Message(id = 55, value = "Config %s, %s handler chain %s: doesn't contain handler with name %s")
-    // OperationFailedException missingHandler(String configName, String handlerChainType, String handlerChainId, String handlerName);
-
-    //@LogMessage(level = ERROR)
-    //@Message(id = 56, value = "Method invocation failed with exception: %s")
-    //void methodInvocationFailed(@Cause Throwable cause, String message);
```
bstansberry, line 239: Please restore the commented methods. They provide documentation of what the message id was used for and help prevent reuse for other purposes by showing any gaps in the currently used id #s was intentional.

webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
```diff
@@ -264,10 +254,6 @@ public interface WSLogger extends BasicLogger {
     @Message(id = 64, value = "Could not update WS server configuration because of existing WS deployment on the server.")
     DisabledOperationException couldNotUpdateServerConfigBecauseOfExistingWSDeployment();
 
-    @LogMessage(level = WARN)
-    @Message(id = 65, value = "Annotation '@%s' found on class '%s'. Perhaps you forgot to add a '%s' module dependency to your deployment?")
-    void missingModuleDependency(String annotation, String clazz, String module);
```
bstansberry, line 269: Please restore this but comment it out.

webservices/server-integration/src/main/java/org/jboss/as/webservices/logging/WSLogger.java
```diff
@@ -277,12 +263,6 @@ public interface WSLogger extends BasicLogger {
     @Message(id = 67, value = "Could not activate the webservices subsystem.")
     void couldNotActivateSubsystem(@Cause Throwable cause);
 
-//    @Message(id = 68, value = "Service %s not available")
-//    OperationFailedException serviceNotAvailable(String serviceName);
-
-//    @Message(id = 69, value = "String format password is required")
-//    IllegalArgumentException invalidPasswordType();
```
bstansberry, line 284: Please restore.

bstansberry (changes requested):
The main change LGTM; thanks!

I requested some things be restored to the logger file.

bstansberry:
Simple via of TC test status (mostly for my own benefit this today): https://ci.wildfly.org/project.html?projectId=WF_PullRequest&branch_WF_PullRequest=pull%2F17479

PR #17491, opened 2023-12-15
[WFLY-15405][WFLY-18852] AMQP Connector for Reactive Messaging follow-up

kabir:
Follow-up to https://github.com/wildfly/wildfly/pull/16281. Those commits are also included in here, so the last three ones are the relevant ones.

https://issues.redhat.com/browse/WFLY-15405
https://issues.redhat.com/browse/WFLY-18852

Note that for the WFLY-18852 commit, I have mainly focussed on trimming the <exclusions> from galleon-pack/galleon-shared/pom.xml when those already exist in boms/common-expansion/pom.xml. Additionally, I moved some of those exclusions to the bom for the subsystems I am most familiar with.

I originally tried to have no exclusions in  galleon-pack/galleon-shared/pom.xml and have them all managed by the BOM, but that gave compile errors, so I went with the approach outlined above. If it needs further cleaning up, I think that should be another Jira, possibly one for each subsystem.


bstansberry (approved):

bstansberry:
Thanks @kabir

PR #17508, opened 2023-12-28
WFLY-18878 26.x Upgrade Infinispan to 13.0.21.Final

FireBurn:
https://issues.redhat.com/browse/WFLY-18878

Hoping to convince folk to cut a new release of 26.x with this fix


jamezp:
Thank you for your contribution. WildFly 26 is no longer being developed with our focus current focus on WildFly 31 and beyond.

You could manually update a local install, just know the combination has not been tested.

PR #17512, opened 2024-01-03
[WFLY-18882] Add resource adapter subsystem dependencies based on cap…

tadamski:
…ability

https://issues.redhat.com/browse/WFLY-18882

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersRootResourceDefinition.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersRootResourceDefinition.java
@@ -8,6 +8,7 @@ import static org.jboss.as.connector.subsystems.resourceadapters.ResourceAdapter
 
 import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.SimpleResourceDefinition;
+import org.jboss.as.controller.capability.RuntimeCapability;
 import org.jboss.as.controller.operations.common.GenericSubsystemDescribeHandler;
 import org.jboss.as.controller.registry.ManagementResourceRegistration;
 import org.jboss.as.controller.registry.RuntimePackageDependency;
@@ -18,6 +19,8 @@ import org.jboss.as.controller.registry.RuntimePackageDependency;
  */
 public class ResourceAdaptersRootResourceDefinition extends SimpleResourceDefinition {
 
+    static final RuntimeCapability<Void> RESOURCE_ADAPTERS_SUBSYSTEM = RuntimeCapability.Builder.of(Capabilities.RESOURCE_ADAPTERS_SUBSYSTEM_CAPABILITY_NAME).build();
+
     private final boolean runtimeOnlyRegistrationValid;
 
     @Override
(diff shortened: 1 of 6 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersRootResourceDefinition.java
```diff
@@ -18,6 +19,10 @@
  */
 public class ResourceAdaptersRootResourceDefinition extends SimpleResourceDefinition {
 
+    private static final String RESOURCE_ADAPTERS_SUBSYSTEM_CAPABILITY = "org.wildfly.resource-adapters";
```
ehsavoie, line 22: Shouldn't this be named RESOURCE_ADAPTERS_SUBSYSTEM_CAPABILITY_NAME and be shared in a "Capabilities.java " class instead of duplicating the constant on several places.

@see https://github.com/wildfly/wildfly/blob/main/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/Capabilities.java for example

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersRootResourceDefinition.java
```diff
@@ -18,6 +19,10 @@
...
+
+    static final RuntimeCapability<Void> RESOURCE_ADAPTERS_SUBSYSTEM = RuntimeCapability.Builder.of(RESOURCE_ADAPTERS_SUBSYSTEM_CAPABILITY).build();
```
ehsavoie, line 24: This should go in "Capabilities.java " to be easily referenced

ehsavoie (commented):
I suggested small changes to align with how we are managing capabilities in other subsystems

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/ResourceAdaptersRootResourceDefinition.java
tadamski: @ehsavoie this was the intial solution but I have removed it after @pferraro review in https://github.com/wildfly/wildfly/pull/15422 stating:
"Normally, your "org.wildfly.resource-adapters" would expose the value provided by the service installed via the resource-adapters subsystem add operation - however, that service does not sufficiently encapsulate its provided value, so we definitely do not want this exposed via a public contract."

pferraro: @tadamski My comment was not in reference to the public static String in Capabilities.java, but rather to the service installed by the resource providing this capability.  In other words, services exposed to the capability registry should be well encapsulated.  Normally, your "org.wildfly.resource-adapters" would expose a service providing some value (e.g. the report directory location), however ResourceAdaptersSubsystemService currently exposes itself, which is what I meant by "poorly encapsulated".

tadamski: @pferraro Is changing ResourceAdaptersSubsystemService type to Void and acceptable solution?

pferraro: I did not mean to suggest that any changes to ResourceAdaptersSubsystemService were needed within the context of your fix for WFLY-18882.
This was intended as a general comment about future improvements.

> Is changing ResourceAdaptersSubsystemService type to Void and acceptable solution?

ResourceAdaptersSubsystemService cannot implement Service<Void> because other services depend on the value it provides.
However, it is unclear to me what exactly ResourceAdaptersSubsystemService is meant to provide.
Clearly there is a need to expose the report directory location (in which case, the corresponding capability should probably be named accordingly).
However, I do not fully understand the purpose of the ResourceAdaptersSubsystemService.getAdapters() method.  It seems to me that resource adapters defined by the management model do not require a separate data structure to be referenceable.  Additionally, resource adapters defined by a deployment are already referenceable via a DeploymentUnit attachment or similar.
Additionally, ignoring the thread safety issues with the data structure returned by this method (see https://issues.redhat.com/browse/WFLY-18881), if it contents are meant to reference some object T, it should probably do so by storing references to Supplier<T> rather than using ServiceNames.

tadamski: @pferraro I have reverted the PR to the state in which capability is present in Capabilities class. I'm working on a subsystem refactor based on your review, but I would prefer to merge it as a part of another jira hopefully sometime next week.

pferraro (approved):

PR #17524, opened 2024-01-13
[WFLY-18916] [WFLY-18917] Upgrade RESTEasy and RESTEasy MicroProfile

jamezp:
If separate PR's would be preferred, these can work alone. However, I figured it might be better to test them together.

- https://issues.redhat.com/browse/WFLY-18916
- https://issues.redhat.com/browse/WFLY-18917


bstansberry:
Thanks @jamezp

PR #17528, opened 2024-01-16
WFLY-18922 Give the Apache Lucene module access to jdk.management

yrodiere:
https://issues.redhat.com/browse/WFLY-18922

I checked manually (through debug mode and by checking `server.log`) that this solves the problem: `org.apache.lucene.util.RamUsageEstimator` now has access to what it needs and no longer logs a warning.


bstansberry (approved):

bstansberry:
Thanks @yrodiere

PR #17537, opened 2024-01-19
WFLY-18890 docs/schema does not contain orm_3-1.xsd

scottmarlow:
https://issues.redhat.com/browse/WFLY-18890

### DIFF
--- a/jpa/hibernate6/src/main/resources/schema/orm_3_1.xsd
+++ b/jpa/hibernate6/src/main/resources/schema/orm_3_1.xsd
@@ -1,0 +1,2325 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2022 Oracle and/or its affiliates. All rights reserved.
+
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License v. 2.0 which is available at
+    http://www.eclipse.org/legal/epl-2.0,
+    or the Eclipse Distribution License v. 1.0 which is available at
+    http://www.eclipse.org/org/documents/edl-v10.php.
+
+    SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+
+-->
+
+<!-- Jakarta Persistence API object/relational mapping file schema -->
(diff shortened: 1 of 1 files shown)


scottmarlow:
This change isn't enough, perhaps we need to reference the `jipijapa-xsd` artifact somewhere in the WildFly build.

Output of `jar tf ~/.m2/repository/org/wildfly/jipijapa-xsd/31.0.0.Final-SNAPSHOT/jipijapa-xsd-31.0.0.Final-SNAPSHOT.jar`:
> META-INF/MANIFEST.MF
META-INF/
schema/
META-INF/maven/
META-INF/maven/org.wildfly/
META-INF/maven/org.wildfly/jipijapa-xsd/
schema/orm_3_1.xsd
META-INF/maven/org.wildfly/jipijapa-xsd/pom.xml
META-INF/maven/org.wildfly/jipijapa-xsd/pom.properties
META-INF/INDEX.LIST

bstansberry:
If you just create a jpa/hibernate6/src/main/resources/schema dir and put it there, that's all you should need.

You could also just put it in the existing jpa/subsystem/src/main/resources/schema dir. That's a bit less 'elegant' because the schema is for Hibernate, not the subsystem in general. But it doesn't really matter.

The 'group' element in wildfly-feature-pack.xml refers to a maven groupId, which for full WF modules is the 'org.wildfly' that's already there. No need for a new entry.

scottmarlow:
> If you just create a jpa/hibernate6/src/main/resources/schema dir and put it there, that's all you should need.
> 
> You could also just put it in the existing jpa/subsystem/src/main/resources/schema dir. That's a bit less 'elegant' because the schema is for Hibernate, not the subsystem in general. But it doesn't really matter.
> 
> The 'group' element in wildfly-feature-pack.xml refers to a maven groupId, which for full WF modules is the 'org.wildfly' that's already there. No need for a new entry.

Thanks @bstansberry, I now can see orm_3_1.xsd under dist/target/wildfly-31.0.0.Final-SNAPSHOT/docs/schema!

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #17540, opened 2024-01-19
[WFLY-18298] Upgrade H2 from 2.2.220 to 2.2.224

bstansberry:
https://issues.redhat.com/browse/WFLY-18928

This aligns us with ORM 6.4

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -378,7 +378,7 @@
         <version.com.google.guava>32.1.2-jre</version.com.google.guava>
         <version.com.google.guava.failureaccess>1.0.1</version.com.google.guava.failureaccess>
         <version.com.google.protobuf>3.19.6</version.com.google.protobuf>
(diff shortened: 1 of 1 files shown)


bstansberry:
@scottmarlow @tadamski please review

scottmarlow:
Hibernate ORM 6.4.* is also using `2.2.224` so this change will be aligned with that.  

https://github.com/h2database/h2database/issues/3944 was reported with 2.2.224 but not clear it the reported `database corruption` can happen with earlier versions as well so that is something to be aware of but still I think we should move forward to 2.2.224.

scottmarlow (approved):

tadamski (approved):
I don't see any problems with that update too

bstansberry:
Thanks @scottmarlow and @tadamski

PR #17546, opened 2024-01-22
[WFLY-18935] JCA: resource adapters subsystem services refactor

tadamski:
https://issues.redhat.com/browse/WFLY-18935

@pferraro this is a PR with the refactor after the review in https://github.com/wildfly/wildfly/pull/17512, but I have only split services into two, so that they are imho better named and encapsulated, but it seems to me that the dependency mechanism here is correct. For the reference it was introduced in https://issues.redhat.com/browse/WFLY-1701 to make sure that RA deployment DUPs don't run only after necessary services are present. I have also looked at another usage of org.jboss.as.server.deployment#addDeploymentDependency it seems pretty similiar to me to what is done in this case.


bstansberry:
@tadamski There doesn't appear to be anything particularly problematic about this PR; apologies it sat so long. If your PRs get stalled due to inattention please start a thread in [#Pull requests](https://wildfly.zulipchat.com/#narrow/channel/274477-Pull-requests) to ask for input.

connector/src/main/java/org/jboss/as/connector/services/resourceadapters/deployment/AbstractResourceAdapterDeploymentService.java
```diff
@@ -110,7 +110,9 @@ public abstract class AbstractResourceAdapterDeploymentService {
     protected final InjectedValue<TransactionIntegration> txInt = new InjectedValue<TransactionIntegration>();
     protected final InjectedValue<CachedConnectionManager> ccmValue = new InjectedValue<CachedConnectionManager>();
     protected final InjectedValue<ExecutorService> executorServiceInjector = new InjectedValue<ExecutorService>();
-    protected final InjectedValue<ResourceAdaptersSubsystemService> resourceAdaptersSubsystem = new InjectedValue<>();
+    protected final InjectedValue<CopyOnWriteArrayListMultiMap> configuredAdaptersInjector = new InjectedValue<>();
```
bstansberry, line 113: I don't see this used anywhere except in the getter, which is only used to set up the injection. Is that because the injection isn't actually needed, and all that is need is a guarantee that RESOURCEADAPTERS_CONFIGURED_ADAPTERS_SERVICE is started before this service starts?

bstansberry (commented):
Just one question; otherwise this looks ok.

This kind of thing would be a good opportunity to move this area of code away from the deprecated MSC 1 APIs and to the current ones, but given that this has already been open for almost two years and I'm not going to ask for that.

If you respond and don't get a reply within a few working days, please ping me in https://wildfly.zulipchat.com/#narrow/channel/274477-Pull-requests. GitHub @ mentions are not reliable with me as I'm involved with way too many projects and PRs.

connector/src/main/java/org/jboss/as/connector/services/resourceadapters/deployment/AbstractResourceAdapterDeploymentService.java
tadamski: it is used in deployment processors to prevent race condition between connector subsystem services and resource adapter deployments:
https://github.com/wildfly/wildfly/blob/main/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/RarDependencyProcessor.java#L59
https://github.com/wildfly/wildfly/blob/main/connector/src/main/java/org/jboss/as/connector/deployers/ra/processors/RaXmlDependencyProcessor.java#L45

bstansberry: @tadamski Ok, my interpretation of that then is it's the 'all that is needed is a guarantee that RESOURCEADAPTERS_CONFIGURED_ADAPTERS_SERVICE is started before this service starts' case. We don't have to have this field. It's possible to wire up the dependency without this field.

But having the field is no big deal, and certainly not something to block merging this over.

bstansberry (approved):

PR #17556, opened 2024-01-24
[WFLY-18938] Add a profile to disable wildscribe generation

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-18938 

Add a profile to disable wildscribe generation

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -177,5 +177,32 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <!-- Profile to disable time consuming wildscribe generation -->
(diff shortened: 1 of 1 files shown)


rhusar (changes requested):
I don't think this is actually needed. Just run the following to skip the model description generation:

`mvn clean install -f docs/pom.xml -Dwildscribe.skip=true`

bstansberry:
Thanks @soul2zimate, but I'm going to close this as @rhusar is correct. Sorry to have wasted your time! Please feel free to reopen if you disagree.

PR #17557, opened 2024-01-24
WFLY-18949 Update resteasy-spring to 3.1.1.Final

liweinan:
https://issues.redhat.com/browse/WFLY-18949

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -514,7 +514,7 @@
         <version.org.jboss.resteasy>6.2.7.Final</version.org.jboss.resteasy>
         <version.org.jboss.resteasy.extensions>2.0.1.Final</version.org.jboss.resteasy.extensions>
         <version.org.jboss.resteasy.microprofile>2.1.5.Final</version.org.jboss.resteasy.microprofile>
-        <version.org.jboss.resteasy.spring>3.0.4.Final</version.org.jboss.resteasy.spring>
+        <version.org.jboss.resteasy.spring>3.1.1.Final</version.org.jboss.resteasy.spring>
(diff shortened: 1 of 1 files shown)


liweinan:
Since Spring6, The Springframework has moved its baseline to JDK17:

- https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6/

@jamezp I guess this update should be on hold right?

jamezp:
Spring 6 itself already requires Java SE 17 and that is already required to work with Jakarta EE 10, so this should not be a blocker as it's already the case :)

jamezp:
For some reason the UI didn't update, but the Linux - JDK 11 job did pass here https://ci.wildfly.org/buildConfiguration/WF_PullRequest_GalleonLinuxJdk11/414414?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false

liweinan:
@jamezp Thanks for helping on this!

liweinan:
There is a known regression issue in this release:

- [resteasy-spring / SpringBeanProcessor broken after fixing issue #145 (v3.1.1.final) #229](https://github.com/resteasy/resteasy-spring/issues/229)

And it's just fixed in the main branch and waiting for some input from community:

- [Error booting up with spring boot 3.2.2 RESTEASY013015: could not find the type for bean named #306](https://github.com/resteasy/resteasy-spring-boot/issues/306)

I plan to do a new release of `resterasy-spring` and then do the update in WildFly.

Thanks for checking this James and Brian!

PR #17560, opened 2024-01-25
[WFLY-18954] Deprecate javaee.api module in favor of wildflyee.api module

ropalka:
https://issues.redhat.com/browse/WFLY-18954


bstansberry:
@ropalka @bmaxwell This change doesn't seem valid. The wildflyee.api module is 'private' which is a more restricted setting than 'deprecated'. This change documents the alias as having a less restricted use than the target module.

PR #17561, opened 2024-01-25
[WFLY-18954] Deprecate javaee.api module in favor of wildflyee.api module

ropalka:
https://issues.redhat.com/browse/WFLY-18954

Upstream PR: https://github.com/wildfly/wildfly/pull/17560


bstansberry:
I don't think this is correct in general, but even if we want it in WF 32,  it's not something that needs changing in a micro.

PR #17563, opened 2024-01-26
Fix for WFLY-18964, Upgrade to Galleon 6 Beta2, upgrade all provisioning tooling

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-18964


jfdenise:
@bstansberry , that is the Galleon tooling that supports Upward compatibility. In Beta for now, would be Final for WF 32. It would be great to have it in place for the WF 32 dev cycle. I am expecting some more Beta to come related to support for stability.

bstansberry (approved):

PR #17565, opened 2024-01-26
[WFLY-18881] CopyOnWriteArrayListMultiMap refactor

tadamski:
https://issues.redhat.com/browse/WFLY-18881

____

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/util/CopyOnWriteArrayListMultiMap.java
+++ b/connector/src/main/java/org/jboss/as/connector/util/CopyOnWriteArrayListMultiMap.java
@@ -5,55 +5,37 @@
 
 package org.jboss.as.connector.util;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 public class CopyOnWriteArrayListMultiMap<K, V> {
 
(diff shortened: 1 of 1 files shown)


tadamski:
@pferraro could you please review?

connector/src/main/java/org/jboss/as/connector/util/CopyOnWriteArrayListMultiMap.java
```diff
@@ -18,42 +18,31 @@ public List<V> get(K k) {
         return cache.get(k);
     }
 
-    public synchronized List<V> remove(K k) {
+    public List<V> remove(K k) {
         return cache.remove(k);
     }
 
-    public synchronized void putIfAbsent(K k, V v) {
+    public void putIfAbsent(K k, V v) {
```
pferraro, line 25: This will be prone to lost updates, e.g. for concurrent calls to putIfAbsent(...) for the same key, where no map entry yet exists.  Rather, I think you want something like:
```java
cache.computeIfAbsent(k, key -> new CopyOnWriteArrayList<>()).addIfAbsent(v);
```

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/util/CopyOnWriteArrayListMultiMap.java
tadamski: fixed

pferraro (approved):

bstansberry:

PR #17566, opened 2024-01-28
[WFLY-18969] Give the Apache Lucene module access to jdk.unsupported

ropalka:
https://issues.redhat.com/browse/WFLY-18969

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/apache/lucene/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/apache/lucene/main/module.xml
@@ -25,8 +25,9 @@
 
     <dependencies>
         <module name="java.logging"/>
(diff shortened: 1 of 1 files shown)


bstansberry:
@yrodiere Any concerns?

yrodiere (approved):
I don't see why this removes the comment, but otherwise this makes sense. Thanks!

ropalka:
The comment about fixed JIRA and it's number is not necessary. Git users can gain this information via git blame command if commit is properly documented @yrodiere .

yrodiere:
@ropalka Yes I know of `git blame`. It works until it doesn't, for example when you starting moving lines around. Like in this very PR for example :) https://github.com/wildfly/wildfly/blame/ee60b1e75e31bb1446f2b4045f81be304d857bc6/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/apache/lucene/main/module.xml#L29

Anyway, I approved the PR, as far as I'm concerned it's good to merge.

ropalka:
You're right @yrodiere !

PR #17567, opened 2024-01-28
[WFLY-18969] Give the Apache Lucene module access to jdk.unsupported

ropalka:
https://issues.redhat.com/browse/WFLY-18969

Upstream PR: https://github.com/wildfly/wildfly/pull/17566


bstansberry:
Thanks @ropalka

PR #17582, opened 2024-02-02
WFLY-18990 Remove illegal uses of Operation.hasOptionalCapability(String, null, null) in JSF subsystem

pferraro:
https://issues.redhat.com/browse/WFLY-18990

Blocks https://github.com/wildfly/wildfly-core/pull/5818


jasondlee (approved):
LGTM. Thanks, @pferraro.

PR #17585, opened 2024-02-02
WFLY-18993 Remove illegal uses of Operation.hasOptionalCapability(String, null, null) from JCA subsystem

pferraro:
https://issues.redhat.com/browse/WFLY-18993


tadamski (approved):

bstansberry:

PR #17588, opened 2024-02-02
WFLY-18996 Remove illegal uses of Operation.hasOptionalCapability(String, null, null) in batch-jberet subsystem

pferraro:
https://issues.redhat.com/browse/WFLY-18996


jamezp (approved):

liweinan (approved):
Thanks for the update Paul @pferraro !

PR #17569, opened 2024-01-28
[WFLY-18971] Upgrade ByteBuddy to 1.14.11

ropalka:
https://issues.redhat.com/browse/WFLY-18971

PR #17591, opened 2024-02-05
WFLY-19029 Hibernate ORM 6.4+ should export services to consumer classpath

Sanne:
[ Updated, this is now [https://issues.redhat.com/browse/WFLY-19029](https://issues.redhat.com/browse/WFLY-19029) ]

Hi @scottmarlow , leaving this as a draft here:

@mbladel has a fix in Hibernate ORM which you'll also need (so this should be adapted as ORM version 6.4.4 is released)

My changes to the modules here are not strictly required to make those failing tests pass, but I'd still recommend including these so as to ensure the services are loaded even when ORM is being initialized in some other way, e.g. by a 3rd party framework.

See also:
 - https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/Adapting.20WildFly.20module.20for.20Hibernate.20ORM
 - https://hibernate.atlassian.net/browse/HHH-17705

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate6/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate6/main/module.xml
@@ -23,7 +23,7 @@
         <module name="jakarta.validation.api"/>
         <module name="jakarta.xml.bind.api"/>
 
-        <module name="org.hibernate"/>
+        <module name="org.hibernate" services="import"/>
         <module name="org.hibernate.commons-annotations"/>
         <module name="org.infinispan.core" services="import"/>
         <module name="org.infinispan.protostream"/>
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
@@ -44,4 +44,11 @@
         <!-- If the PostgreSQL JDBC driver classes are accessible, Hibernate ORM will use them to tailor PostgresDialect and CockroachDialect behavior -->
         <module name="org.postgresql.jdbc" optional="true"/>
     </dependencies>
+
+    <provides>
+        <service name="org.hibernate.bytecode.spi.BytecodeProvider">
+            <with-class name="org.hibernate.bytecode.internal.bytebuddy.BytecodeProviderImpl"/>
+        </service>
+    </provides>
+
 </module>


ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate6/main/module.xml
```diff
@@ -23,7 +23,7 @@
         <module name="jakarta.validation.api"/>
         <module name="jakarta.xml.bind.api"/>
 
-        <module name="org.hibernate"/>
+        <module name="org.hibernate" services="import"/>
```
scottmarlow, line 26: Thanks Sanne!  The 3rd party framework idea sounds interesting!

bstansberry:
@Sanne @scottmarlow If there's a 6.4.4 in the works, fyi re https://github.com/wildfly/wildfly/pull/17569, which is a request to update Byte Buddy, I believe to resolve some issue with running on the upcoming SE 22. @ropalka can explain. We generally would not change Byte Buddy to something other than what Hibernate tested with the ORM version we've integrated, so I wouldn't merge that without Scott's ok.

scottmarlow:
> @Sanne @scottmarlow If there's a 6.4.4 in the works, fyi re #17569, which is a request to update Byte Buddy, I believe to resolve some issue with running on the upcoming SE 22. @ropalka can explain. We generally would not change Byte Buddy to something other than what Hibernate tested with the ORM version we've integrated, so I wouldn't merge that without Scott's ok.

I cherry-picked the https://github.com/hibernate/hibernate-orm/pull/7785 (main branch) change and the WildFly tests that previously failed with ORM 6.4.3.Final passed with my testing, so I think we will expect to upgrade to ORM 6.4.4.Final when available based on the ORM pr#7785.  

Regarding the changes in this pr for third part frameworks.  I think we are still discussing the intention of that change in the https://hibernate.zulipchat.com/#narrow/stream/132094-hibernate-orm-dev/topic/Adapting.20WildFly.20module.20for.20Hibernate.20ORM room that Sanne mentioned above.

I created an upstream (main ORM branch) upgrade to ByteBuddy 1.14.11 via https://github.com/hibernate/hibernate-orm/pull/7790 so we can get test results of that change.  IMO if that passes, we should look at what changed to see what we are getting with the upgrade as shown by https://github.com/raphw/byte-buddy/compare/byte-buddy-1.14.7...byte-buddy-1.14.11 (one of the changes is for Java SE 22 which likely helps WildFly as Brian mentioned above).

Sanne:
@bstansberry I'm always happy to try a Bytebuddy upgrade on Hibernate - it's usually painless and that would trigger the test runs we need.

@ropalka just let us know when there's a need to upgrade, I try to stay on top of necessary/recommended upgrades but also we're refraining from automated upgrades so as to not cause you all downstream work when there's no reason to.
But also let us know the reasons please? We've already been testing on JDK22 so I'm missing something.

Thanks @scottmarlow for sending the upgrade PR !

Sanne:
> We've already been testing on JDK22 so I'm missing something.

We figured that out: we were setting the `net.bytebuddy.experimental` flag still on these JDKs, which makes the use of an older unsupported version of ByteBuddy more lenient. This is fixed now.
 - https://github.com/hibernate/hibernate-orm/pull/7793

scottmarlow:
Just opened https://issues.redhat.com/browse/WFLY-18989 for upgrading WildFly to 6.4.4

Sanne:
Thanks for spotting that - created https://issues.redhat.com/browse/WFLY-19029 and un-drafted this (after updating it).

Sanne:
@wildfly-bot check

scottmarlow (approved):

bstansberry:
Thanks @Sanne and @scottmarlow

PR #17605, opened 2024-02-07
[WFLY-19014][WFLY-19035] Upgrade jboss-ejb-client to 5.0.6.Final, wildfly-http-client to 2.0.7.Final

tadamski:
https://issues.redhat.com/browse/WFLY-19014


bstansberry:
@tadamski This has relevant CI failures.

tadamski:
@bstansberry This update changes ejb-client handling of context data (removing of context data that is not in jboss.returned.keys). I have updated the test so that it passes but I haven't thought of making parallel update to http client and now the behavior is inconsistent, hence the test cannot be configured to pass on both remote and http client. I have commented the http test and I'm working on wildfly-http-client update which will make this handling consistent and which will uncomment this test.

tadamski:
OK. Now it should work.

bstansberry (approved):

PR #17616, opened 2024-02-09
WFLY-19020 Application client applications should be able to deploy persistence units

scottmarlow:
https://issues.redhat.com/browse/WFLY-19020

Application client applications should be able to deploy persistence units (note that they won't have a second level cache enabled).


scottmarlow:
This pull request is considered draft as we need more confidence that we have recreated the reported issue.  I recreated deployment failures of a local app with the second level cache enabled but I think that is a different problem.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow. As a user-reported issue this seems like a good one to port to 31.x

PR #17602, opened 2024-02-07
[WFLY-19010] Elytron registry is in common module

kabir:
https://issues.redhat.com/browse/WFLY-19010

PR #17619, opened 2024-02-12
31.x - [WFLY-19010] Elytron registry is in common module

kabir:
Without this if only the amqp connector is provisioned, and not kafka there will be an error configuring ssl

https://issues.redhat.com/browse/WFLY-19010


jmesnil:
Upstream PR: https://github.com/wildfly/wildfly/pull/17602

jmesnil (approved):

jmesnil:
@kabir thanks!

PR #17626, opened 2024-02-12
[WFLY-16545] Make the Jakarta REST @ApplicationPath, @Provider and @P…

jamezp:
…ath bean defining annotations.

https://issues.redhat.com/browse/WFLY-16545

We are only doing this at a community stability level at this point as it's a change in behavior. Per the [11.2.3](https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1#cdi) Jakarta REST Specification:

> In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or use application scope.

In resteasy-cdi there is currently a CDI portable extension which adds the scope of the bean, if not already defined, to the resource, provider or application. In future versions of Jakarta REST the specific annotations; `@ApplicationPath`, `@Provider` and `@Path` will be defined as bean defining annotations. Part of this PR is planning ahead for that.

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
@@ -8,6 +8,9 @@ package org.jboss.as.jaxrs;
 
 import static org.jboss.as.weld.Capabilities.WELD_CAPABILITY_NAME;
 
+import jakarta.ws.rs.ApplicationPath;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.ext.Provider;
 import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
@@ -26,6 +29,8 @@ import org.jboss.as.jaxrs.logging.JaxrsLogger;
 import org.jboss.as.server.AbstractDeploymentChainStep;
 import org.jboss.as.server.DeploymentProcessorTarget;
 import org.jboss.as.server.deployment.Phase;
+import org.jboss.as.weld.Capabilities;
+import org.jboss.as.weld.WeldCapability;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceTarget;
 import org.jboss.resteasy.spi.ResteasyDeployment;
@@ -68,6 +73,20 @@ class JaxrsSubsystemAdd extends AbstractBoottimeAddStepHandler {
 
         JaxrsServerConfig serverConfig = createServerConfig(operation, context);
         JaxrsServerConfigService.install(serviceTarget, serverConfig);
+
+        // If CDI is available, we must support the annotations below as bean defining annotations. In the next release
+        // of the Jakarta REST specification, these will be stereotyped annotations and a scope will be assigned.
+        // Currently, the RESTEasy portable extension assigns the scope.
+        if (context.hasOptionalCapability(Capabilities.WELD_CAPABILITY_NAME, null, null)) {
+            final WeldCapability preProcessingWeldCapability = context.getCapabilityRuntimeAPI(Capabilities.WELD_CAPABILITY_NAME, WeldCapability.class);
+            // Per section 11.2.3 of the Jakarta REST 3.1 specification:
+            // In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource
+            // classes, providers and Application subclasses. Providers and Application subclasses MUST be singletons or
+            // use application scope.
+            preProcessingWeldCapability.addBeanDefiningAnnotation(ApplicationPath.class)
+                    .addBeanDefiningAnnotation(Provider.class)
+                    .addBeanDefiningAnnotation(Path.class);
+        }
     }
 
     private static JaxrsServerConfig createServerConfig(ModelNode configuration, OperationContext context) throws OperationFailedException {
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CDIBean.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CDIBean.java
@@ -4,9 +4,12 @@
  */
 package org.jboss.as.test.integration.jaxrs.integration.cdi;
 
+import jakarta.enterprise.context.ApplicationScoped;
+
 /**
  * @author Stuart Douglas
  */
+@ApplicationScoped
 public class CDIBean {
 
     public String message() {
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CdiInjectionTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CdiInjectionTestCase.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.jaxrs.integration.cdi;
+
+import java.lang.annotation.Annotation;
+import java.util.Set;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.RequestScoped;
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.Annotated;
+import jakarta.enterprise.inject.spi.Bean;
+import jakarta.enterprise.inject.spi.InjectionPoint;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.core.Application;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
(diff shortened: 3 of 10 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CDIBean.java
```diff
@@ -4,9 +4,12 @@
  */
 package org.jboss.as.test.integration.jaxrs.integration.cdi;
 
+import jakarta.enterprise.context.ApplicationScoped;
+
 /**
  * @author Stuart Douglas
  */
+@ApplicationScoped
```
jamezp, line 12: This annotation is required to make the `CDIBean` a real CDI bean if the `bean-discovery-mode` is set to `annotated`.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/openapi/service/multimodule/TestEjb.java
```diff
@@ -6,11 +6,13 @@
 package org.wildfly.test.integration.microprofile.openapi.service.multimodule;
 
 import jakarta.ejb.Stateless;
+import jakarta.enterprise.context.Dependent;
 
 /**
  * @author Joachim Grimm
  */
 @Stateless
+@Dependent
```
jamezp, line 15: Stateless EJB's must be dependent scoped.

testsuite/integration/rts/src/test/java/org/wildfly/test/extension/rts/common/WorkRestATResource.java
```diff
@@ -32,7 +32,7 @@
 import org.jboss.jbossts.star.util.TxSupport;
 
 @Path(WorkRestATResource.PATH_SEGMENT)
-public final class WorkRestATResource {
+public class WorkRestATResource {
```
jamezp, line 35: Making this final makes the type non-proxyable.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
```diff
@@ -68,6 +74,23 @@ public void execute(DeploymentProcessorTarget processorTarget) {
 
         JaxrsServerConfig serverConfig = createServerConfig(operation, context);
         JaxrsServerConfigService.install(serviceTarget, serverConfig);
+
+        // If CDI is available and this is community level support, add the Jakarta REST type annotations as bean defining annotations.
+        // We currently only support community level as RESTEasy 6.2 specially handles CDI non-proxyable resources. In
+        // future versions of RESTEasy, this will not be done as the Jakarta REST specification itself will assign these
+        // annotations as @Stereotype annotations making bean defining annotations.
+        if (context.getStability() == Stability.COMMUNITY && context.hasOptionalCapability(Capabilities.WELD_CAPABILITY_NAME, null, null)) {
```
pferraro, line 82: Should this not also be available for servers using lower stability levels?
If so, I think you want:
```java
if (context.getStability().enables(Stability.COMMUNITY) && ...)
```

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
```diff
@@ -68,6 +74,23 @@ public void execute(DeploymentProcessorTarget processorTarget) {
 
         JaxrsServerConfig serverConfig = createServerConfig(operation, context);
         JaxrsServerConfigService.install(serviceTarget, serverConfig);
+
+        // If CDI is available and this is community level support, add the Jakarta REST type annotations as bean defining annotations.
+        // We currently only support community level as RESTEasy 6.2 specially handles CDI non-proxyable resources. In
+        // future versions of RESTEasy, this will not be done as the Jakarta REST specification itself will assign these
+        // annotations as @Stereotype annotations making bean defining annotations.
+        if (context.getStability().enables(Stability.COMMUNITY) && context.hasOptionalCapability(Capabilities.WELD_CAPABILITY_NAME, null, null)) {
```
jamezp, line 82: After discussing a bit with @pferraro on [Zulip](https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers/topic/Jakarta.20REST.20CDI/near/421147943), I think we should remove the stability check. I'm only holding off for now to let CI run.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/openapi/service/multimodule/TestEjb.java
manovotn: Just curious if this is actually needed?
In an EE environment (which I am not sure applies to MP tests?), CDI should discover EJB beans as well and basically assign its own scope if it's not explicitly added there.
See https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#default_bean_discovery_ee

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CdiInjectionTestCase.java
```diff
@@ -0,0 +1,91 @@
...
+    public void checkProvider() throws Exception {
+        checkResponse("provider", CDIProvider.class);
+    }
+
+    @Test
+    public void checkResource() throws Exception {
+        checkResponse("resource", CDIResource.class);
+    }
+
+    private void checkResponse(final String path, final Class<?> expectedType) throws Exception {
```
manovotn, line 75: Wouldn't this pass even previously as the classes might not be CDI beans but would still work as REST components?
What I mean is, shouldn't you test if those classes are resolvable as CDI beans instead (i.e. injectable, or you could even check their `Bean<?>` metadata for scopes etc).

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/openapi/service/multimodule/TestEjb.java
jamezp: For some reason it wasn't being discovered. Let me look into possible reasons why. I had some similar issues in the RESTEasy test suite too where I need to add a bean defining annotation. Maybe there is something more to this.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/integration/cdi/CdiInjectionTestCase.java
jamezp: That's a good point. I just copied this test from an old commit and it could probably just run in the container and lookup beans instead of trying to use a resource for them.

manovotn: You could use `Instance` and look those classes up as CDI beans programmatically.
If you want to check their metadata, use `Instance.select(...).getHandle().getBean()`; from there you can see scopes, stereotypes etc. used on that given bean.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/openapi/service/multimodule/TestEjb.java
manovotn: Well, it's in an EJB jar, TBF I am not sure what rules apply in that case :thinking:

jamezp: This doesn't make it much clearer https://jakarta.ee/specifications/enterprise-beans/4.0/jakarta-enterprise-beans-spec-core-4.0#a195. I can say I still need to add the `@Dependent` annotation if the EJB is in the WAR too.

What doesn't make sense is why it works when the Jakarta REST annotations are not BDA's.

What I've noticed in debug is the `WeldDeploymentMarker.mark()` is not invoked if the annotations are not BDA's. I'm not sure in that case the `@Inject` works. I actually think the test is wrong. The resource is never invoked so you never see an NPE. However, if you invoke the endpoint you'll get an NPE.

That said, I don't know why the `@Stateless` annotations is not working as a BDA.

manovotn: Ah, that's a good link because:

> An Enterprise Beans **packaged into a CDI bean archive** and not annotated with `jakarta.enterprise.inject.Vetoed` annotation, is considered a CDI-enabled bean.

Like you say, the archive here isn't a CDI bean archive since there is nothing that would mark it as one. I think it makes sense, after all, the EJB part is an integration layer, not the core CDI.

> That said, I don't know why the `@Stateless` annotations is not working as a BDA.

It looks like it works as a BDA if it's in an archive that's a bean archive.

So adding the annotation to this test seems to make sense after all, sorry for the fuss.

manovotn (approved):
LGTM

testsuite/integration/rts/src/test/java/org/wildfly/test/extension/rts/common/WorkRestATResource.java
mmusgrov: +1

bstansberry:
@jamezp What's the plan with this one?

Above you describe this as being only for 'community' stability but beyond adding module dependency I don't see anything in the PR related to checking stability.

jamezp:
This is a great question. There is an RFE for this, but it's in the back log so maybe for now I should just close this until it makes it off the back log. What we have now works, bit really the REST spec itself should be changing this to truly make the annotations CDI qualifiers.

PR #17632, opened 2024-02-13
[WFLY-18913] Restore the ignored service-name attribute in the opentelemetry subsystem

jasondlee:
https://issues.redhat.com/browse/WFLY-18913

- If set, assign value of opentelemetry.service-name to otel.service.name
- Set otel.service.name with deployment name only if not already set
- Add test to verify honoring of service-name
- Create shared base class for custom Testcontainers
- Add JaegerContainer to aid in capturing exported traces
- Integrate JaegerContainer with OpenTelemetryCollectorContainer
- Various code clean ups in test code

____


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/BaseContainer.java
```diff
@@ -0,0 +1,72 @@
+package org.wildfly.test.integration.observability.container;
```
darranl, line 1: Missing copyright

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/JaegerContainer.java
```diff
@@ -0,0 +1,90 @@
+package org.wildfly.test.integration.observability.container;
```
darranl, line 1: Missing copyright

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/BaseContainer.java
```diff
@@ -0,0 +1,72 @@
...
+                if (bytes != null) {
+                    debugLog(new String(bytes));
+                }
+            }));
+        }
+    }
+
+    @Override
+    public SELF withExposedPorts(Integer... ports) {
+        getExposedPorts().addAll(exposedPorts);
```
bstansberry, line 51: @jasondlee I don't see what this does.

GenericContainer.getExposedPorts() returns a list that is only referred to in a local variable. So modifying that list and then letting that value go out of scope doesn't affect anything.

Also if the intent is to ignore the 'ports' param and just rely on the 'exposedPorts' passed into the constructor, the javadoc should say that.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/OpenTelemetryCollectorContainer.java
```diff
@@ -0,0 +1,88 @@
...
+        super.start();
+        otlpGrpcEndpoint = "http://localhost:" + getMappedPort(OTLP_GRPC_PORT);
+        otlpHttpEndpoint = "http://localhost:" + getMappedPort(OTLP_HTTP_PORT);
+        prometheusUrl = "http://localhost:" + getMappedPort(PROMETHEUS_PORT) + "/metrics";
+    }
+
+    @Override
+    public synchronized void stop() {
+        if (jaegerContainer != null) {
+            jaegerContainer.stop();
```
bstansberry, line 67: If you're nulling out INSTANCE should you null out jaegerContainer as well?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/BaseContainer.java
jasondlee: That's an override from the super class that doesn't appear to be needed anymore, so I'll remove it.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/container/OpenTelemetryCollectorContainer.java
jasondlee: Probably. Done.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
```diff
@@ -5,118 +5,68 @@
...
 
-        ServerReload.reloadIfRequired(managementClient);
-    }
+            executeOp(managementClient, clearAttribute("undertow", STATISTICS_ENABLED));
+            if (subsystemAdded) {
+                executeOp(managementClient, Operations.createRemoveOperation(micrometerSubsystem));
+            }
+            if (extensionAdded) {
+                executeOp(managementClient, Operations.createRemoveOperation(micrometerExtension));
+            }
```
bstansberry, line 67: L61-67 should be outside the 'if (dockerAvailable) block, since they are outside of it in setup.

bstansberry (changes requested):

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
jasondlee: Good catch. Fixed.

bstansberry (approved):

PR #17637, opened 2024-02-14
[WFLY-18163] Make it possible to use JaasSecurityRealm via a custom-realm resource

fjuma:
https://issues.redhat.com/browse/WFLY-18163

Supersedes https://github.com/wildfly/wildfly/pull/17622

I have made the updates @yersan suggested in https://github.com/wildfly/wildfly/pull/17622 to add `javax.xml.stream.api` back to `LayersTestBase` and opened this new PR since @Skyllarr is now on PTO.

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
@@ -98,7 +98,6 @@ public abstract class LayersTestBase {
      * Packages that are always expected to be included in the return value of {@link #getExpectedUnreferenced()}.
      */
     public static final String[] NOT_REFERENCED_COMMON = {
-            "javax.xml.stream.api",
             // injected by ee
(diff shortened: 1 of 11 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -249,7 +250,7 @@ public abstract class LayersTestBase {
             "ibm.jdk",
             "javax.api",
             "javax.sql.api",
-            "javax.xml.stream.api",
+            "javax.xml.stream.api"
```
Skyllarr, line 253: @fjuma Thank you for submitting this PR fix! I noticed there is a comma missing so the build will fail

fjuma: @Skyllarr Thanks very much for spotting that! I've fixed this.

PR #17643, opened 2024-02-15
WFLY-19040 Configure SNI hostnames per cluster if possible

pferraro:
As of Infinispan 14.0.18, HotRod client enables hostname validation by default - so we need to configure SNI hostnames per cluster if possible.
Auto-disable hostname validation if unsupported by server configuration.

https://issues.redhat.com/browse/WFLY-19040


jmesnil:
@rhusar Could you please review this PR? We'd like to incorporate it in WildFly 31.0.1.Final but I want to merge it in upstream first before backporting it to the `31.x` branch.

rhusar (approved):
LGTM, the fix was also verified by QE.

PR #17645, opened 2024-02-16
Fix for WFLY-19021, WildFly provisioning to support WildFly stability

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-19021

Also bumping Galleon and WildFly Galleon plugins to latest beta.

NOTE: The minimum stability of all built feature-pack has been set to `experimental`.

### DIFF
--- a/ee-feature-pack/galleon-feature-pack/pom.xml
+++ b/ee-feature-pack/galleon-feature-pack/pom.xml
@@ -182,6 +182,7 @@
                             <fork-embedded>${galleon.fork.embedded}</fork-embedded>
                             <generate-channel-manifest>true</generate-channel-manifest>
                             <deploy-channel-manifest>false</deploy-channel-manifest>
+                            <minimum-stability>experimental</minimum-stability>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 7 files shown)


jfdenise:
Th failure for cloud tests is unrelated. Seems an infrastructure error. Should be re-run at some point.

bstansberry:
@jfdenise Let's discuss this in a thread in https://wildfly.zulipchat.com/#narrow/stream/174184-wildfly-developers.

AIUI this PR is just using a new setting that achieves the same thing as existing settings that you remove. So it's no big deal and is mergeable.  But WFLY-19021 overall is a Feature Request, with an associated analysis at https://github.com/wildfly/wildfly-proposals/pull/555, and we're working to use the new WildFly feature dev process for such things. We're kind of bypassing that with this work though, e.g. the plugins are being updated and released and now being merged into WF. I don't want to just ignore any process requirements and keep merging things, so let's talk about what we need to do to adapt this to process.

Step 1 is to pick a stability level for WFLY-19021 itself. That doesn't map to --stability etc, but the levels are general concepts that go beyond Paul's work and the related stuff.

bstansberry:
FYI, this works great in https://github.com/wildfly-extras/mvc-krazo-feature-pack/pull/18

jfdenise:
@bstansberry , I updated the Proposal to align with the new process. 
A new Galleon Plugins release would be needed to align with the updated proposal. I am going to set this PR as draft until we have an approved proposal.

bstansberry:
/retest

@jfdenise Please move this out of draft state. Your plan re testing is enough to keep moving forward.

bstansberry:
Thanks @jfdenise!

PR #17658, opened 2024-02-22
[WFLY-18955] micrometer Galleon layer should contain rules to be discovered

jasondlee:
https://issues.redhat.com/browse/WFLY-18955

- Add rules to layer-spec.xml

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/layers/standalone/micrometer/layer-spec.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/layers/standalone/micrometer/layer-spec.xml
@@ -8,6 +8,8 @@
         <prop name="org.wildfly.rule.add-on-depends-on" value="only:cdi"/>
         <prop name="org.wildfly.rule.add-on" value="observability,micrometer"/>
         <prop name="org.wildfly.rule.add-on-description" value="Support for Micrometer."/>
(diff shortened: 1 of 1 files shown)


galleon-pack/galleon-shared/src/main/resources/layers/standalone/micrometer/layer-spec.xml
```diff
@@ -8,6 +8,8 @@
         <prop name="org.wildfly.rule.add-on-depends-on" value="only:cdi"/>
         <prop name="org.wildfly.rule.add-on" value="observability,micrometer"/>
```
jfdenise, line 9: @jasondlee , so this layer is also an add-on? Is there a use case to explicitly enable this layer, even if the deployment doesn't use the micrometer API?

jasondlee: @jfdenise I think so. An admin may want server metrics exported via Micrometer even if no deployments are explicitly using Micrometer.

jfdenise: @jasondlee , Thank-you!

jfdenise (approved):

bstansberry:
Thanks @jasondlee and @jfdenise

PR #17670, opened 2024-02-27
WFLY-19071/WFLY-19072: Fixing AddressSettings default values and Artemis 2.32 dependencies

ehsavoie:
* Setting the default values on the root of AddressSettings so that
    they are used by all child AddressSettings.
  * Fixing the JSON produced to be retro compatible.
  * Adding module de.dentrassi.crypto.pem-keystore
  * Moving from sunb.jdk to java.base   

Jira: https://issues.redhat.com/browse/WFLY-19072
      https://issues.redhat.com/browse/WFLY-19071

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/de/dentrassi/crypto/pem-keystore/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/de/dentrassi/crypto/pem-keystore/main/module.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="de.dentrassi.crypto.pem-keystore">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${de.dentrassi.crypto:pem-keystore}"/>
+    </resources>
+
+
+    <dependencies>
+        <module name="org.bouncycastle.bcpkix" export="true" services="export"/>
+        <module name="org.bouncycastle.bcprov" export="true" services="export"/>
+    </dependencies>
+
+    <provides>
+        <service name="java.security.Provider">
+            <with-class name="de.dentrassi.crypto.pem.PemKeyStoreProvider"/>
+        </service>
+    </provides>
+</module>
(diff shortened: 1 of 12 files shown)


rhusar (changes requested):

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -252,8 +252,8 @@ public abstract class LayersTestBase {
             "javax.sql.api",
             "javax.xml.stream.api",
             // TODO see if org.apache.activemq.artemis.client can use specific JDK module(s)
```
bstansberry, line 254: Please remove the TODO comment.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/de/dentrassi/crypto/pem-keystore/main/module.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="de.dentrassi.crypto.pem-keystore">
```
bstansberry, line 8: @fjuma FYI about this new module.

bstansberry (changes requested):
@ehsavoie Just one minor change, plus I wanted to make Farah (and @darranl) aware of this new crypto lib.

bstansberry (approved):

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ManagementUtil.java
```diff
@@ -28,6 +33,11 @@
  */
 public class ManagementUtil {
 
+    private static final String SNAKE_REPLACEMENT = "$1-$2";
+    private static final String CAMEL_REGEXP = "([a-z])([A-Z]+)";
+    private static final Map<String, AttributeDefinition> ADDESS_SETTINGS_ATTRIBUTES =
```
rhusar, line 38: This should be `ADDRESS_SETTINGS_ATTRIBUTES`

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ManagementUtil.java
```diff
@@ -28,6 +33,11 @@
...
+            Arrays.stream(AddressSettingDefinition.ATTRIBUTES).collect(Collectors.toMap(AttributeDefinition::getName, att -> att)) ;
```
rhusar, line 39: Replace with `Function.identity()`

rhusar (changes requested):
Also, `ADDESS_SETTINGS_ATTRIBUTES` is never used, what's up with that? :)

rhusar (approved):
LGTM now, feedback has been addressed.

bstansberry:
Thanks @ehsavoie and @rhusar

PR #17671, opened 2024-02-27
[WFLY-19073] fix schema link jboss-app_7_0.xsd

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-19073


rhusar (approved):
LGTM, TLDR the old link doesn't resolve, the new one does.

bstansberry (changes requested):
@baranowb Please just delete this file. With this change this file becomes the same as the one in JBoss Metadata.

I don't know why this one file was treated differently from all the other schemas that come from JBoss Metadata; I suspect it's just a leftover that was meant to be deleted long ago but was missed. In any case, since the file in JBM has the correct content there's no reason to have the wildfly repo continue to manage this schema.

bstansberry:
LOL @baranowb! I got curious about the reason for this and I find:

https://github.com/wildfly/wildfly/commit/21f50b2535bb93a3d7175020b3e3e23bf83262c5

Commit message is:

'AS7-3401 Temporarily include the jboss-app_7_0.xsd within the AS7 source. Once a new version of JBoss Metadata is released, we'll fetch it from that jar'

I suppose 12 years still counts as 'Temporarily'. :-D

rhusar (changes requested):
Just checking the link fixed in the metadata project and it's correct in there - https://github.com/jboss/metadata/blob/main/ear/src/main/resources/schema/jboss-app_7_0.xsd#L56 - so yes, just delete the file altogether.

rhusar (approved):
LGTM now

bstansberry (approved):

bstansberry:
Thanks @baranowb and @rhusar

PR #17694, opened 2024-03-06
[WFLY-19087]: External messaging resources can't be updated.

ehsavoie:
* misusage of context.isRuntimeOnlyRegistrationValid() which made all of the resources read-only

Jira: https://issues.redhat.com/browse/WFLY-19087


bstansberry (commented):
@ehsavoie 

This seems like it doesn't fully fix the misuse, more like it hides it?

All these definition classes end up with an field 'registerRuntimeOnly' that seems to not really mean 'registerRuntimeOnly'. I haven't looked at them all but it seems like it's "register as read only because ???".

I think the variable names should be corrected to reflect their actual use, javadoc added explaining the meaning of the params etc. Otherwise this is sure to get broken again when a maintainer thinks they understand but doesn't really and a code reviewer doesn't either.

Even now as a reviewer I'd need to investigate each of these classes to check that the changed to 'false' is right and I'd have to think too hard to do it.

ehsavoie:
@bstansberry I tried to make the intentions clearer by renaming parameters and adding some Javadoc.

PR #17697, opened 2024-03-06
[WFLY-19100] Fixes Datasource subsystem 5.0 and 6.0 parsing

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19100

Notes:
* The issue seems to be affecting only the ds security parsing, thus the simple fix. 
* Tested with all default configs from all affected WFLY versions.


tadamski (approved):
thanks @emmartins LGTM

bstansberry:
Thanks @emmartins and @tadamski

PR #17018, opened 2023-07-13
[WFLY-16532][Preview] add ability to configure additional scope for authentication request

PrarthonaPaul:
Analysis doc: https://github.com/wildfly/wildfly-proposals/pull/527
WFLY Issue: https://issues.redhat.com/browse/WFLY-16532
Related ELY Issue: https://issues.redhat.com/browse/ELY-2574
Depends on: https://github.com/wildfly-security/wildfly-elytron/pull/1925
Depends on: https://github.com/wildfly/wildfly/pull/17705

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> Wildfly issue links:
> * [WFLY-19176](https://issues.redhat.com/browse/WFLY-19176)
> * [WFLY-19078](https://issues.redhat.com/browse/WFLY-19078)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #17705, opened 2024-03-11
WFLY-19078 [Preview] - Update the elytron-oidc-client subsystem parser to enumerate schema versions

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFLY-19078

____


elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * Enumerated the schema versions for the elytron-oidc-client subsystem.
+ * @author Prarthona Paul
+ */
+
+public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<ElytronOidcSubsystemSchema> {
+    VERSION_1_0(1),
+    VERSION_2_0(2),
+    VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY),
```
pferraro, line 42: Since there is no community-specific namespace for 2.0, you don't need to increment the version.
e.g.
```java
VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY)
```

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcExtension.java
```diff
@@ -51,19 +39,19 @@ static ResourceDescriptionResolver getResourceDescriptionResolver(final String..
...
         final ManagementResourceRegistration registration = subsystem.registerSubsystemModel(new ElytronOidcSubsystemDefinition());
         registration.registerOperationHandler(GenericSubsystemDescribeHandler.DEFINITION, GenericSubsystemDescribeHandler.INSTANCE);
     }
 
     public void initializeParsers(ExtensionParsingContext context) {
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemParser_1_0.NAMESPACE_1_0, ELYTRON_OIDC_SUBSYSTEM_PARSER_1_0);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemParser_2_0.NAMESPACE_2_0, CURRENT_PARSER);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_1_0.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_1_0);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_2_0.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_2_0);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_3_0_COMMUNITY.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_3_0_COMMUNITY);
```
pferraro, line 55:
```java
context.setSubsystemXmlMappings(SUBSYSTEM_NAME, EnumSet.allOf(ElytronOidcSubsystemSchema.class));
```

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Enumeration of elytron-oidc-client subsystem model versions.
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+enum ElytronOidcClientSubsystemModel implements SubsystemModel {
+    VERSION_1_0_0(1, 0, 0),
+    VERSION_2_0_0(2, 0, 0),
+    VERSION_3_0_0_COMMUNITY(3, 0, 0),
```
pferraro, line 19: Only management model changes to the default stability level require incrementing the subsystem model version.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
```diff
@@ -0,0 +1,38 @@
...
+import org.wildfly.subsystem.resource.ManagementResourceRegistrar;
+import org.wildfly.subsystem.resource.ManagementResourceRegistrationContext;
+import org.wildfly.subsystem.resource.ResourceDescriptor;
+import org.wildfly.subsystem.resource.SubsystemResourceDefinitionRegistrar;
+
+/**
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public class ElytronOidcClientSubsystemRegistrar implements SubsystemResourceDefinitionRegistrar {
```
pferraro, line 24: Where is this class used?
I can't seem to find anything that calls this object's register(...) method.
Did you mean to reference this from your Extension?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
PrarthonaPaul: I will change it to just version 1 and 2. 
In that case, should I remove the following function? 
https://github.com/wildfly/wildfly/blob/89884940717ae10cf0e39689a19932607e978d6f/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java#L47-L51

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: I have changed the community version to version 2 and updated the schema and test files. 
Quick question, when attributes in 2.0 community move to default, will they be part of subsystem version 3.0 or 2.0?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcExtension.java
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
pferraro: Since 2.0 (default) already exists, you would increment to 3.0.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
pferraro: If the management model (for any stability level) changes, you still need to increment the model version.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: So, would i go back to ` VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY),`?

pferraro: You would only introduce community:3.0 if you are adding new features to the existing community:2.0.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
PrarthonaPaul: I have removed the register(...) method.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcExtension.java
PrarthonaPaul: I had to change this back to what it back because I was getting the following error: 
```
javax.xml.stream.XMLStreamException: WFLYCTL0505: elytron-oidc-client subsystem namespace urn:wildfly:elytron-oidc-client:community:3.0 is not enabled by the current stability level
```

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
PrarthonaPaul: I have incremented the model version to 3 and included a from3(...) method

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
fjuma: @PrarthonaPaul Since there isn't a community-specific namespace for 2.0, you can use:

`VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY)`.

Later, if we were to add new features to the existing community:2.0, that's when we'd use `VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY)`. Or if we were to make your feature part of the default stability level instead, we'd introduce VERSION_3_0.

PrarthonaPaul: Thanks @fjuma for clarifying. 
So, just to confirm, this means in the model versions we would just have version 2 and 1, no version 3. And the contents of from3(...) function (please see below) would be shifted to from2(...). Is that correct? 
https://github.com/wildfly/wildfly/blob/74260706d505c5df38e840d6ba58d09b309d54a6/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java#L47-L54

pferraro: The model version of a subsystem identifies a set of potential features exposed by a subsystem, independent of whether or not they are enabled (or indeed enablable) by a user's server configuration.  For a given version of WildFly, the ModelVersion of a subsystem is fixed.  This value is merely descriptive, and is largely invisible to users.  In general, a subsystem increments this version when the set of potential features has changed since the previous release.

That said, for this PR, you want to increment your subsystem model version to 3.0.0.

In contrast, a subsystem's XML schema describes what can be configured for a given subsystem.  A given version of WildFly can boot using XML that complies with any schema version of a given subsystem.  However, for a given subsystem, _what_ can be configured is constrained by the stability level of the server process.  That we version these using numbers is merely a convention (albeit a useful one), as is the inclusion of the stability level to the namespace uri - the uniqueness of the namespace URI is all that actually matters.

In general, we need to create a new schema version if the current schema version was already published by a previous WildFly release.

_Technically_, the only requirement for a new schema is that its namespace should be distinct from any existing one.  In other words, you _could_ name these however you wanted, so long as you take care to only register schemas supported by the current server process.  

That said, requiring subsystems to perform conditional schema registration depending on the stability level of the server is something we want to avoid.  To that effect, the ExtensionParsingContext, i.e. the thing with which you register your subsystem parsers, will introspect the Stability associated with a given SubsystemSchema and respond appropriately.

That said, for this PR, option 1 is clearly not an option, since it requires a new/distinct schema version.

If you used "wildfly:elytron-oicd-client:3.0" for the new namespace, the default logic of SubsystemSchema.getStability() would return Stability.DEFAULT.  This would otherwise cause ExtensionParsingContext to treat this as a valid parser for a server started using --stability=default, which would be incorrect.  Technically, you could override the value return by this method, but that would likely introduce confusion, as it would not be clear to which stability level the schema applies.

That leaves 2 obvious options:
* "wildfly:elytron-oicd-client:community:2.0"
* "wildfly:elytron-oicd:client:community:3.0"

Either will technically work with the existing infrastructure as they are distinct from the existing "wildfly:elytron-oicd-client:2.0".

Let's say the next release of WildFly adds a new preview feature to this subsystem.  For a server running in preview, the new schema includes changes from this PR.  Using the first option, the next schema becomes "wildfly:elytron-oicd-client:preview:2.0".  However, using the 2nd option, the next schema becomes "wildfly:elytron-oicd:client:preview:4.0".

Now consider a subsequent release of WildFly adding a new default feature to this subsystem.  We now have 3 schemas to manage, depending on the stability level of the server.
If we follow the conventions from the first option, the new names become: 
"wildfly:elytron-oicd:client:preview:3.0"
"wildfly:elytron-oicd:client:community:3.0"
"wildfly:elytron-oicd:client:3.0"
If we follow the conventions from the second option, the new names become:
"wildfly:elytron-oicd:client:community:3.0"

darranl: Are you missing two namespaces in the last option @pferraro 

FYI I like "wildfly:elytron-oicd-client:community:2.0" as that seems to be what @Skyllarr did in her PR ;-)

fjuma: +1, I agree that going with 2.0 makes sense and matches what Diana did as well.

fjuma:
@pferraro When you get a chance, please take a look at @PrarthonaPaul's updates for this one. Thanks!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: Thanks @fjuma, @pferraro  and @darranl for your input. 
I have added a new commit to convert the schema to preview:2.0 since we have decided to move WFLY-17143 and WFLY-16532 to preview as well. I have also switched back to not incrementing the model version to 3.0.0.

fjuma: @PrarthonaPaul The model version still needs to be incremented to 3.0.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+import java.util.Collections;
+/**
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public class ElytronOidcClientSubsystemRegistrar {
+
+    static final String NAME = "elytron-oidc-client";
```
fjuma, line 26: Is this variable needed? Can ElytronOidcExtension.SUBSYSTEM_NAME be used?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -28,14 +31,25 @@ public String getSubsystemName() {
...
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0 });
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion() });
     }
 
     private static void from2(ChainedTransformationDescriptionBuilder chainedBuilder) {
-        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_2_0_0, VERSION_1_0_0);
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion());
         builder.rejectChildResource(PathElement.pathElement(SECURE_SERVER));
     }
+    private static void from3(ChainedTransformationDescriptionBuilder chainedBuilder) {
```
fjuma, line 46: I don't think transformers are needed for preview changes.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -28,14 +31,25 @@ public String getSubsystemName() {
     public void registerTransformers(SubsystemTransformerRegistration registration) {
 
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
+        // 3.0.0 (WildFly 32) to 1.0.0 (WildFly 29)
+        from3(chainedBuilder);
         // 2.0.0 (WildFly 29) to 1.0.0 (WildFly 28)
         from2(chainedBuilder);
 
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0 });
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion() });
```
fjuma, line 39: I think this should be:

chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0.getVersion() });

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ProviderDefinition.java
```diff
@@ -28,6 +30,7 @@
  */
 class ProviderDefinition extends SimpleResourceDefinition {
 
+    static final ResourceRegistration PATH = ResourceRegistration.of(PathElement.pathElement(ElytronOidcDescriptionConstants.PROVIDER), Stability.DEFAULT);
```
fjuma, line 33: Just to check, where does this get used?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,256 @@
...
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.BEARER_ONLY;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.CLIENT_ID;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.ENABLE_BASIC_AUTH;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.MIN_TIME_BETWEEN_JWKS_REQUESTS;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.PUBLIC_CLIENT;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.PUBLIC_KEY_CACHE_TTL;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.RESOURCE;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.TOKEN_MINIMUM_TIME_TO_LIVE;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.TURN_OFF_CHANGE_SESSION_ID_ON_LOGIN;
+import static org.wildfly.extension.elytron.oidc.SecureDeploymentDefinition.USE_RESOURCE_ROLE_MAPPINGS;
```
bstansberry, line 32: Static imports come first.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -17,6 +13,13 @@
 import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
 import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
 
+
+import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_1_0_0;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_2_0_0;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_3_0_0;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcDescriptionConstants.SECURE_DEPLOYMENT;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcDescriptionConstants.SECURE_SERVER;
```
bstansberry, line 21: Static imports comes first.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,256 @@
...
+import org.jboss.as.controller.Feature;
+import org.jboss.as.controller.PersistentSubsystemSchema;
+import org.jboss.as.controller.PersistentResourceXMLDescription;
+import org.jboss.as.controller.SubsystemSchema;
+import org.jboss.as.controller.xml.VersionedNamespace;
+import org.jboss.as.version.Stability;
+import org.jboss.staxmapper.IntVersion;
+
+import java.util.EnumSet;
+import java.util.Map;
```
bstansberry, line 18: java and javax packages come before other packages, after static imports

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.wildfly.extension.elytron.oidc;
+
+import org.jboss.as.subsystem.test.AbstractSubsystemSchemaTest;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.util.EnumSet;
+import java.util.Properties;
```
bstansberry, line 14: java and javax packages come before others, after static imports

bstansberry (dismissed):

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
bstansberry: @fjuma @PrarthonaPaul  They aren't needed. But....

In the draft process doc I encourage writing them, under the theory that waiting to do so may result in problems not being caught until we try to promote a feature to 'default'. But that was before we'd done any actual non-default resources or attributes, so we haven't really thought through what this means in general.

I'm interested in what @yersan plus you, @darranl and @pferraro and (and everybody really) thinks about this. 

Prerequisite: I'm assuming adding transformation stuff for resources/attributes that don't exist at 'default' stability works at all; i.e. the kernel transformation code doesn't blow up when processing non-existent stuff. If that assumption is wrong, then we should go with Option 1 below.

Some options:

1) Drop the suggestion to write transformers from the process doc. Require PRs that add them to remove them.

2) Keep the suggestion but ask people who add them to comment out the functional bits; just leave the 'framing' in place.

3) Allow them, and require that they are correct if present. (I don't think what's here is correct.)

By correct I mean:

3.a) if the resources/attributes being transformed were at 'default' then the transformation is correct.
3.b) while the resources/attributes are not at default, the transformation does no harm.

I think 3.b) is the tricky one, but in most cases it should be fine. If my memory is correct and a secondary HC can't register unless the DC is running at 'default' and the secondary is too, then the common case of 'reject this new thing' if present will be fine, as 'this new thing' cannot be part of the config.

If we go with option 3, for cases where 3.b) is an issue then we can ask that the problematic part is commented out.

Options 1 and 2 are also valid. In the end we need to examine the transformation rules when the feature moves to default, even if the code already exists and is just starting to take effect. Maybe we'll be more likely to check the transformers if every PR moving resources/attributes to default is required to have a transformer section; i.e. it can never rely on one that was previously written.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ProviderDefinition.java
PrarthonaPaul: Here: https://github.com/wildfly/wildfly/blob/1983491071bff113d53a789292cb2389bb7d692f/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java#L100

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
PrarthonaPaul: removed NAME and using ElytronOidcExtension.SUBSYSTEM_NAME instead

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: moved to the top

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: moved to the top, after static imports

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTestCase.java
PrarthonaPaul: moved

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
bstansberry: > @bstansberry This is something we should enforce via checkstyle rules.

+1. I have no idea why this wasn't done in, say, 2011.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,255 @@
...
+                                .addAttribute(CredentialDefinition.CLIENT_KEY_PASSWORD, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                                .addAttribute(CredentialDefinition.TOKEN_TIMEOUT, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                                .addAttribute(CredentialDefinition.CLIENT_KEY_ALIAS, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                                .addAttribute(CredentialDefinition.ALGORITHM, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                                .build())
+                        .addChild(factory.builder(RedirectRewriteRuleDefinition.PATH)
+                                .addAttribute(RedirectRewriteRuleDefinition.REPLACEMENT, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                                .build())
+                        .build())
+                .addChild(factory.builder(SecureServerDefinition.PATH)
```
fjuma, line 193: I think we need to be a bit careful here. The `secure-server` resource was only introduced in version 2.0. So I think we need something similar to the type of logic Diana had added in her subsystem schema refactoring changes (see https://github.com/wildfly/wildfly-core/commit/89eac6a4a095c78950650a6253c1e6fffdc7aa6a#diff-9e0841048dde4b289fc01a5c76980785b1322d5fa7003454599ec4194ab67e4bR88-R91) so that `builder.addChild` is only called if `this.since(ElytronOidcSubsystemScheam.VERSION_2_0)`

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/RealmDefinition.java
```diff
@@ -28,6 +33,9 @@
  */
 class RealmDefinition extends SimpleResourceDefinition {
 
+    static final ResourceRegistration PATH = ResourceRegistration.of(PathElement.pathElement(ElytronOidcDescriptionConstants.REALM), Stability.DEFAULT);
+    static final SimpleAttributeDefinition TYPE = new SimpleAttributeDefinitionBuilder("type", ModelType.STRING)
```
fjuma, line 37: Just to check, where is TYPE being used?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -28,14 +28,26 @@ public String getSubsystemName() {
     public void registerTransformers(SubsystemTransformerRegistration registration) {
 
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
+        // 3.0.0 (WildFly 32) to 1.0.0 (WildFly 29)
```
fjuma, line 31: s/1.0.0/2.0.0

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Enumeration of elytron-oidc-client subsystem model versions.
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+enum ElytronOidcClientSubsystemModel implements SubsystemModel {
+    VERSION_1_0_0(1, 0, 0),
+    VERSION_2_0_0(2, 0, 0),
+    VERSION_3_0_0(3, 0, 0),
```
pferraro, line 19: Generally, I would encourage adding a comment correlating model versions with WF versions.
e.g.
```java
  VERSION_3_0_0(3, 0, 0), // WildFly 32.0-present
```

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
pferraro: true - whoever enables it would need to be willing to fix the errors!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * Enumerated the schema versions for the elytron-oidc-client subsystem.
+ * @author Prarthona Paul
+ */
+
+public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<ElytronOidcSubsystemSchema> {
+    VERSION_1_0(1, Stability.DEFAULT),
+    VERSION_2_0(2, Stability.DEFAULT),
+    VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW),
```
pferraro, line 41: Generally, I would encourage adding a comment correlating schema versions with WF versions.
e.g.
```java
  VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW), // WildFly 32.0-present
```

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,255 @@
...
+                        .addAttribute(ProviderAttributeDefinitions.REGISTER_NODE_PERIOD, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.SOCKET_TIMEOUT_MILLIS, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.SSL_REQUIRED, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.TOKEN_SIGNATURE_ALGORITHM, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.TOKEN_STORE, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.TRUSTSTORE, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.TRUSTSTORE_PASSWORD, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
+                        .addAttribute(ProviderAttributeDefinitions.VERIFY_TOKEN_AUDIENCE, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER)
```
pferraro, line 97: Why not ... ?
```java
.addAttributes(Stream.of(ProviderAttributeDefinitions.ATTRIBUTES))
```

(same for other resources)

pferraro (changes requested):

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: We need to specify the parser and marshaller specifically, because otherwise when I added it as above, I got a lot of parser errors. Adding the attributes one at a time was the only way I could fix it.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTestCase.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public ElytronOidcClientSubsystemTestCase(ElytronOidcSubsystemSchema schema) {
+        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), schema, ElytronOidcSubsystemSchema.CURRENT.get(schema.getStability()));
+    }
+
+    @Override
+    protected void compareXml(String configId, String original, String marshalled) {
+        //
+    }
```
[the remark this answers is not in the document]
PrarthonaPaul: This was in the original subsystem parser tests. When I removed it for this test, the tests failed (maybe it is because the order of the attributes specified in the XML file is not as expected.)

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -28,14 +28,26 @@ public String getSubsystemName() {
...
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0 });
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0.getVersion() });
     }
 
     private static void from2(ChainedTransformationDescriptionBuilder chainedBuilder) {
-        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_2_0_0, VERSION_1_0_0);
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion());
         builder.rejectChildResource(PathElement.pathElement(SECURE_SERVER));
     }
+    // TODO: Figure out how to handle the transformer with a preview level attribute being added to the subsystem
```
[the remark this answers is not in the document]
PrarthonaPaul: I have added a new issue here: https://issues.redhat.com/browse/WFLY-19177

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: Fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/RealmDefinition.java
PrarthonaPaul: I am not using it anymore. 
I have removed it. 
Thanks for catching it!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.staxmapper.XMLExtendedStreamReader;
+import org.wildfly.subsystem.resource.SubsystemResourceDefinitionRegistrar;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+import java.util.Collections;
+/**
+ * @author <a href="mailto:prpaul@redhat.com">Prarthona Paul</a>
+ */
+
+public class ElytronOidcClientSubsystemRegistrar {
```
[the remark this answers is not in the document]
PrarthonaPaul: I have removed this class. I think initially I was trying to implement SubsystemResourceDefinitionRegistrar, but did not need it. 
Thanks for catching it

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemRegistrar.java
```diff
@@ -0,0 +1,58 @@
...
+
+    static final PathElement PATH = SubsystemResourceDefinitionRegistrar.pathElement(ElytronOidcExtension.SUBSYSTEM_NAME);
+    static final AttributeParser SIMPLE_ATTRIBUTE_PARSER = new AttributeElementParser();
+    static final AttributeMarshaller SIMPLE_ATTRIBUTE_MARSHALLER = new AttributeElementMarshaller();
+
+     static class AttributeElementMarshaller extends AttributeMarshaller.AttributeElementMarshaller {
```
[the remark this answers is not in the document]
PrarthonaPaul: removed this class

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcExtension.java
```diff
@@ -51,19 +44,19 @@ static ResourceDescriptionResolver getResourceDescriptionResolver(final String..
...
         final ManagementResourceRegistration registration = subsystem.registerSubsystemModel(new ElytronOidcSubsystemDefinition());
         registration.registerOperationHandler(GenericSubsystemDescribeHandler.DEFINITION, GenericSubsystemDescribeHandler.INSTANCE);
     }
 
     public void initializeParsers(ExtensionParsingContext context) {
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemParser_1_0.NAMESPACE_1_0, ELYTRON_OIDC_SUBSYSTEM_PARSER_1_0);
-        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemParser_2_0.NAMESPACE_2_0, CURRENT_PARSER);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_1_0.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_1_0);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_2_0.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_2_0);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, ElytronOidcSubsystemSchema.VERSION_2_0_PREVIEW.getNamespace().toString(), ElytronOidcSubsystemSchema.VERSION_2_0_PREVIEW);
```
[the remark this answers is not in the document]
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemDefinition.java
```diff
@@ -12,9 +12,9 @@
 import java.util.Collections;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.PersistentResourceDefinition;
```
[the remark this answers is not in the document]
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/RealmDefinition.java
```diff
@@ -28,6 +33,9 @@
  */
 class RealmDefinition extends SimpleResourceDefinition {
 
+    static final ResourceRegistration PATH = ResourceRegistration.of(PathElement.pathElement(ElytronOidcDescriptionConstants.REALM), Stability.DEFAULT);
```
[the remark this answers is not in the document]
PrarthonaPaul: fixed here and below

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
```diff
@@ -15,34 +15,47 @@
...
+    public ExpressionsTestCase(ElytronOidcSubsystemSchema schema) {
         super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension());
     }
 
     @Test
     public void testExpressions() throws Throwable {
         if (services != null) return;
         String subsystemXml = "oidc-expressions.xml";
-        services = super.createKernelServicesBuilder(new DefaultInitializer()).setSubsystemXmlResource(subsystemXml).build();
+        services = super.createKernelServicesBuilder(new DefaultInitializer(Stability.PREVIEW)).setSubsystemXmlResource(subsystemXml).build();
```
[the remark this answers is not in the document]
PrarthonaPaul: This test has been changes to not be parametrized anymore. So, I left it as this.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/OidcTestCase.java
```diff
@@ -16,29 +16,38 @@
...
+    public OidcTestCase(ElytronOidcSubsystemSchema schema) {
         super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension());
     }
 
     @Before
     public void prepare() throws Throwable {
         if (services != null) return;
         String subsystemXml = "oidc.xml";
-        services = super.createKernelServicesBuilder(new DefaultInitializer()).setSubsystemXmlResource(subsystemXml).build();
+        services = super.createKernelServicesBuilder(new DefaultInitializer(Stability.PREVIEW)).setSubsystemXmlResource(subsystemXml).build();
```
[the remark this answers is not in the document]
PrarthonaPaul: same here

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: fixed!

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
pferraro: This is because these attributes do not override their parser/marshaller, such that AttributeDefinition.getParser() and getAttributeMarshaller() does not return the expected SIMPLE_ATTRIBUTE_PARSER and SIMPLE_ATTRIBUTE_MARSHALLER, respectively.
That said, I think it was a poor design choice that XML behavior was ever hardwired into the AttributeDefinition, so I do think it is preferable to specify custom parsing/marshalling behavior in your PersistentResourceXMLDescription.
To accommodate this, I've opened https://issues.redhat.com/browse/WFCORE-6763 to add a `PersistentResourceXMLDescription.Builder.addAttributes(Stream<AttributeDefinition>, AttributeParser, AttributeMarshaller) method`.
In the meantime, you can use:
```java
Stream.of(ProviderAttributeDefinitions.ATTRIBUTES).forEach(attribute -> builder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
```
or similar.

PrarthonaPaul: Okay, perfect! 
I believe the reason why we are using a custom marshaller and parser is for easy migration from the legacy keycloak adapter. The elytron team wanted to have the structure of the new subsystem be as close to the older one as possible.
I will update this and can create an issue to update it to use `PersistentResourceXMLDescription.Builder.addAttributes(Stream<AttributeDefinition>, AttributeParser, AttributeMarshaller)` in the future once WFCORE-6763 is merged.  
Thank you!

PrarthonaPaul: I have made these updates. I could not add `SecureDeploymentDefinition.ALL_ATTRIBUTES` directly because `CredentialDefinition.CREDENTIAL` and `RedirectRewriteRuleDefinition.REDIRECT_REWRITE_RULE` were added to `SecureDeploymentDefinition.ALL_ATTRIBUTES` and were clashing with when I added them as children. 
So, I had to create an array first including all the attributes minus `credentials` and `redirect-rewrite-rule`

Thanks @pferraro 
Please let me know is this PR looks good (or if there are any changes I should make). Once this is good to go, I can squash the commits and add it to WFLY-16532.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTestCase.java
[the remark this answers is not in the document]
pferraro: The subsystem test framework should already normalize the XML such that the attributes are ordered consistently.  If this comparison is failing, please investigate.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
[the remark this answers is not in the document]
pferraro: Referencing the stability of the test schema will prevent you or someone having to keep this value in sync when your test schema version/stability changes, i.e. DRY.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
yersan: > If my memory is correct and a secondary HC can't register unless the DC is running at 'default' and the secondary is too, then the common case of 'reject this new thing' if present will be fine, as 'this new thing' cannot be part of the config.

That was what I had in my mind, but weeks ago I learnt that's not true, and there is a reason. At this moment we are only allowing domain mode for paired stability levels. E.g DC community -- Secondary HC community, preview -- preview, default -- default ... and so on. If we do not allow it, and we restrict it to default -- X, then users would be forced to run Domain Mode always in default stability level, however, community is the default stability level for WildFly. This situation "forces" us to always create transformers. Notice we are not taking into account whether we are running a mixed domain (different host controller versions).

Another point is we have always discouraged mixed domain usage in WildFly, correct me if I am wrong. Basically, because our mixed domain integration testsuite is limited to only certain versions. This is also a bit more problematic now since we are not forced to bump the schema of any new attribute/operation added at the community stability level and lower levels. Maybe should we? 

Having said that, if we want to continue "allowing" mixed domains in WildFly we are forced to write always transformers and, I understand we are also forced to bump the schema model on each feature that involves management resource changes.

Another alternative would be to restrict the HC registration to only on the default stability level for mixed domains only. That would leverage the requirements of writing transformers only when we are promoting to default, but we could add them if we want for testing purposes and to prepare the changes for the future promotion of the feature to default stability level. @bstansberry I guess that would make your option 3) to fit well.

Am I missing something? Maybe this should be moved to Zulip.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: I have created https://issues.redhat.com/browse/WFLY-19177 to track the changes needed to add attributes using a custom parser and marshaller.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
bstansberry: @yersan Yes, we should discuss this in Zulip. We should see if we can figure out how to have different rules for *mixed domain* as opposed to domain mode overall.

In any case what Prarthona has here is fine for 32 Beta, and really for 32 Final. I don't think we want to support mixed domains involving preview stability level HCs.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -0,0 +1,145 @@
...
+import org.jboss.as.controller.SubsystemSchema;
+import org.jboss.as.controller.parsing.ParseUtils;
+import org.jboss.as.controller.xml.VersionedNamespace;
+import org.jboss.as.version.Stability;
+import org.jboss.dmr.ModelNode;
+import org.jboss.staxmapper.IntVersion;
+import org.jboss.staxmapper.XMLExtendedStreamReader;
+
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
```
bstansberry, line 43: @PrarthonaPaul This shouldn't block merging this, but these should move to immediately after the java.* ones.

fjuma (approved):
Adding my approval on this one since follow up issues have been created to address the remaining items (these have also been linked from [WFLY-19078](https://issues.redhat.com/browse/WFLY-19078)).

fjuma:
@PrarthonaPaul Just noticed, there are failures in `MigrateOperationTestCase` that do look related:

```
java.lang.AssertionError: WFLYCTL0158: Operation handler failed: java.lang.NoClassDefFoundError: Could not initialize class org.wildfly.extension.elytron.oidc.ElytronOidcSubsystemDefinition
  at org.junit.Assert.fail(Assert.java:89)
  at org.junit.Assert.assertTrue(Assert.java:42)
  at org.jboss.as.model.test.ModelTestUtils.checkOutcome(ModelTestUtils.java:119)
  at org.jboss.as.subsystem.test.AbstractSubsystemTest.checkOutcome(AbstractSubsystemTest.java:161)
  at org.keycloak.subsystem.adapter.extension.MigrateOperationTestCase.testMigrateDefaultKeycloakConfig(MigrateOperationTestCase.java:73)
```

fjuma:
> @PrarthonaPaul Just noticed, there are failures in `MigrateOperationTestCase` that do look related:
> 
> ```
> java.lang.AssertionError: WFLYCTL0158: Operation handler failed: java.lang.NoClassDefFoundError: Could not initialize class org.wildfly.extension.elytron.oidc.ElytronOidcSubsystemDefinition
>   at org.junit.Assert.fail(Assert.java:89)
>   at org.junit.Assert.assertTrue(Assert.java:42)
>   at org.jboss.as.model.test.ModelTestUtils.checkOutcome(ModelTestUtils.java:119)
>   at org.jboss.as.subsystem.test.AbstractSubsystemTest.checkOutcome(AbstractSubsystemTest.java:161)
>   at org.keycloak.subsystem.adapter.extension.MigrateOperationTestCase.testMigrateDefaultKeycloakConfig(MigrateOperationTestCase.java:73)
> ```

This test does pass for me locally though with your [scope PR](https://github.com/wildfly/wildfly/pull/17018) that includes the changes here so perhaps there was some other change needed that you already included there.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
PrarthonaPaul: @bstansberry @pferraro 
Would you like me to create an unassigned WFLY issue for the time being to track this?

PR #17721, opened 2024-03-14
[WFLY-19049] Use testcontainers to start Mail server for running mail…

yersan:
… tests

Jira issue: https://issues.redhat.com/browse/WFLY-19049

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTestCase.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.mail.basic;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+import java.io.IOException;
+import java.net.SocketPermission;
+import java.net.URL;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.test.integration.common.HttpRequest;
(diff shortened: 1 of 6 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTestCase.java
```diff
@@ -0,0 +1,128 @@
...
+
+            // POP3 Mail server
+            op = Util.createAddOperation(PathAddress.parseCLIStyleAddress("/subsystem=mail/mail-session=mail-test-basic/server=pop3"));
+            op.get("outbound-socket-binding-ref").set("mail-test-basic-pop3-binding");
+
+            ManagementOperations.executeOperation(managementClient.getControllerClient(), op);
+
+            // Reload to apply the changes
+            ServerReload.reloadIfRequired(managementClient);
+        }
```
rhusar, line 115: Just use `org.jboss.as.test.shared.ManagementServerSetupTask` instead.

rhusar (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTestCase.java
yersan: Ok, got it, thanks @rhusar . The operations don't need to go in a batch. Out of curiosity, what would be the advantage vs the current one?

yersan: Well, I guess, they are applied faster, any other?

yersan: ahh, ok, I see it, if one of them fails, there won't be inconsistencies with partially applied operations, that's good

bstansberry:
@rhusar is this ok now?

rhusar (approved):
LGTM now, cool stuff!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTestCase.java
rhusar: Sorry forgot to reply, but yes apart from getting a consistent server config, it's also simple to use and less verbose = less LOC to manage :)

bstansberry:
Thanks @yersan and @rhusar

PR #17726, opened 2024-03-15
[WFLY-15178] Enable the legacy ParseAndMarshalModelsTestCase's

jamezp:
https://issues.redhat.com/browse/WFLY-15178

This removes all testing exception JBoss EAP 7.4. I also added, in a separate commit in case we don't want it yet, JBoss EAP 8.0 testing as well.

For the 7.4 configuration files, I had to remove the legacy `security-realms` and convert those to use Elytron. I used the `enable-elytron-se17.cli` script to enable this. Hopefully it doesn't modify the XML files too much.


bstansberry (approved):

bstansberry:
Thanks @jamezp

PR #17737, opened 2024-03-19
[WFLY-19147] Let `mp.health.disable-default-procedures=true` disable all the default procedures when set at application level

fabiobrz:
~Tests for [WFLY-19147](https://issues.redhat.com/browse/WFLY-19147)~

~**Requires** https://github.com/eclipse/microprofile-health/pull/328~

FYI @xstefank 

Fix [WFLY-19147](https://issues.redhat.com/browse/WFLY-19147)

____


fabiobrz:
Converting to draft since I might move the test in the microprofile module and use Arquillian as usual without requiring anything from the TCK dependency.

bstansberry:
@fabiobrz Yeah this would have to be moved. The microprofile-tck modules are tcks and nothing more.

testsuite/integration/microprofile-tck/health/pom.xml
```diff
@@ -89,12 +89,42 @@
...
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <dependenciesToScan>
+                                <dependency>${ts.microprofile-tck-tests-dependency}</dependency>
+                            </dependenciesToScan>
+                            <systemPropertyVariables>
+                                <!-- Disable default procedures as the MP Metrics TCK must be run without them -->
```
xstefank, line 101: OK, now that we noticed this funny typo, it should be MP Health not Metrics :D

xstefank (changes requested):
So yes, this needs to move from TCK module to testsuite/integration/microprofile. I'll do the review once this is done since it'll change the PR a lot.

testsuite/integration/microprofile-tck/health/pom.xml
fabiobrz: Thanks @xstefank - will fix.

fabiobrz:
> @fabiobrz Yeah this would have to be moved. The microprofile-tck modules are tcks and nothing more.

Yup, thanks @bstansberry - will refactor accordingly

testsuite/integration/microprofile-tck/health/pom.xml
fabiobrz: Fixed, resolving.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
```diff
@@ -38,8 +39,21 @@ public class MicroProfileHealthReporter {
...
     private boolean userChecksProcessed = false;
+    private final AtomicReference<String> defaultProceduresDeactivatorReference = new AtomicReference<>();
+
+    public void deactivateDefaultProcedures(final String deploymentName) {
+        // If a deployment disables default procedures, that will affect all the default procedures,
+        //  including those relating to the global server context, see https://issues.redhat.com/browse/WFLY-19147
+        if (this.defaultProceduresDeactivatorReference.get() == null) {
+            MicroProfileHealthLogger.LOGGER.deploymentDisablingDefaultProcedures(deploymentName);
+        }
+        this.defaultProceduresDeactivatorReference.compareAndSet(null, deploymentName);
```
xstefank, line 51: I'm not sure why you need to keep the track of the deployment name?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresTestBase.java
```diff
@@ -0,0 +1,65 @@
...
+        // we expect none
+        final List<CHECK_TYPE> livenessChecks = retrieveChecks(managementClient, "live");
+        // we expect none
+        final List<CHECK_TYPE> startedChecks = retrieveChecks(managementClient, "started");
+
+        // let's just get the total checks count
+        Assert.assertEquals(1, readinessChecks.size() + livenessChecks.size(), startedChecks.size());
+        // then we expect the only one we find, implemented by the application level Health Check is returning "UP"
+
+        testReadinessChecks(readinessChecks);
```
xstefank, line 58: verify that this is indeed the application check and not something else

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresTestBase.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    protected void testReadinessChecks(final List<CHECK_TYPE> readinessChecks) throws IOException {
```
xstefank, line 61: and you don't need this as it doesn't test anything that you already didn't test

xstefank (changes requested):

xstefank:
@jmesnil please also take a look. I think this is the correct approach to per deployment overrides with what I had in mind when we declared this in the MP Health specification. In MP, we always aim at single deployment per runtime.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
fabiobrz: I though it could be useful anyway, e.g. improving subsequent log messages that would read such name.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresTestBase.java
fabiobrz: Ok. Removed this implementation by making the method abstract.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresTestBase.java
fabiobrz: Not sure I get it, the check name is validated in overridden implementation by subclasses.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
```diff
@@ -175,8 +175,13 @@ It is possible to disable all these server procedures by using the MicroProfile
...
+1. __When the server starts__, to determine if its _server_ procedures should be disabled or enabled. It can be set using the system property `mp.health.disable-default-procedures` or the environment variable `MP_HEALTH_DISABLE_DEFAULT_PROCEDURES`. Setting this property in a deployment is ignored at that time.
+2. __When an application is deployed__, to determine if WildFly should disable _all_ the default procedures, including
+_server_ ones.
+Setting this property in a deployment has higher priority than the global server configuration, based on the principle that
+microservices applications - and specifically MicroProfile based - usually endorse a single-deployment model.
+With respect to this scenario, it is quite natural to let the deployment override the global configuration.
+When set to `false` (i.e. the default behavior) WildFly will also add a readiness check if the deployment does not provide any. At that time, setting this property in a `microprofile-config.properties` file in the deployment would be taken into account. (with the usual priority rules for MicroProfile Config properties).
```
xstefank, line 184: set to true?

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
```diff
@@ -100,9 +115,11 @@ public SmallRyeHealth getLiveness() {
 
     public SmallRyeHealth getReadiness() {
         final Map<HealthCheck, ClassLoader> serverChecks = new HashMap<>();
-        serverChecks.putAll(serverReadinessChecks);
+        if (defaultProceduresShouldBeAdded()) {
```
xstefank, line 118: store the result and reuse it

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporterService.java
```diff
@@ -61,17 +61,18 @@ private MicroProfileHealthReporterService(Supplier<ServerProbesService> serverPr
...
+        // here the property value is read and stored when the runtime is starting
+        final boolean serverDisablingDefaultProcedures = ConfigProvider.getConfig().getOptionalValue("mp.health.disable-default-procedures", Boolean.class).orElse(false);
         // MicroProfile Health supports the mp.health.default.readiness.empty.response to let users specify default empty readiness responses
         final String defaultReadinessEmptyResponse = ConfigProvider.getConfig().getOptionalValue("mp.health.default.readiness.empty.response", String.class).orElse("DOWN");
         // MicroProfile Health supports the mp.health.default.startup.empty.response to let users specify default empty startup responses
         final String defaultStartupEmptyResponse = ConfigProvider.getConfig().getOptionalValue("mp.health.default.startup.empty.response", String.class).orElse("DOWN");
         healthReporter = new MicroProfileHealthReporter(emptyLivenessChecksStatus, emptyReadinessChecksStatus,
-            emptyStartupChecksStatus, defaultServerProceduresDisabled,
+            emptyStartupChecksStatus, serverDisablingDefaultProcedures,
```
xstefank, line 72: I think the original name was better

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresTestBase.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    @OperateOnDeployment("MicroProfileHealthDisabledDefaultProceduresTestBase")
+    public void testDisabledDefaultProcedures() throws IOException {
+        // we expect one
+        final List<CHECK_TYPE> readinessChecks = retrieveChecks(managementClient, "ready");
+        // we expect none
+        final List<CHECK_TYPE> livenessChecks = retrieveChecks(managementClient, "live");
+        // we expect none
+        final List<CHECK_TYPE> startedChecks = retrieveChecks(managementClient, "started");
```
xstefank, line 53: is there no global health operation that aggregates this?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthUtils.java
```diff
@@ -77,4 +80,33 @@ static void testManagementOperation(ModelNode response, boolean mustBeUP, String
             fail("Probe named " + probeName + " not found in " + result);
         }
     }
+
+    static List<JsonObject> retrieveHttpEndPointChecks(final String healthURL) throws IOException {
```
xstefank, line 84: don't duplicate code, reuse it in the original methods

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc
fabiobrz: Fixed

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
fabiobrz: Fixed, thanks.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthUtils.java
fabiobrz: Done, thanks

fabiobrz:
HI @jmesnil, @xstefank - any news about the changes proposed here?

xstefank (approved):
@jmesnil please review.

@bstansberry this is a breaking change. But I agree with @fabiobrz that this is how we should align with the spec.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/deployment/CDIExtension.java
```diff
@@ -67,24 +67,28 @@ private void afterDeploymentValidation(@Observes final AfterDeploymentValidation
...
             if (!disableDefaultprocedure) {
                 // no startup probes are present in the deployment. register a startup check so that the deployment is considered started
                 defaultStartupCheck = new DefaultStartupHealthCheck(module.getName());
                 reporter.addStartupCheck(defaultStartupCheck, module.getClassLoader());
             }
         }
+        // https://issues.redhat.com/browse/WFLY-19147 - let the per-application `mp.health.disable-default-procedures`
+        // configuration setting affect the Health checks response global server configuration
+        if (disableDefaultprocedure) {
+            reporter.deactivateDefaultProcedures(module.getName());
```
bstansberry, line 90: module.getName() isn't the user-understandable deployment name. We don't have the info here to create an 'ideal' user-understandable deployment name in all cases, but removing any preceding ServiceModuleLoader.MODULE_PREFIX will come pretty close.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/_private/MicroProfileHealthLogger.java
```diff
@@ -52,4 +52,12 @@ public interface MicroProfileHealthLogger extends BasicLogger {
     @Message(id = 6, value = "")
     OperationFailedException seeDownstream();
     */
+
+    @LogMessage(level = WARN)
+    @Message(id = 7, value = "Disabling default procedures according to the %s deployment configuration")
+    void disablingDefaultProcedures(String deploymentName);
```
bstansberry, line 58: If this is spec defined behavior that we document as being acceptable for a microservice-style installation, it seems wrong to WARN about it.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/_private/MicroProfileHealthLogger.java
```diff
@@ -52,4 +52,12 @@ public interface MicroProfileHealthLogger extends BasicLogger {
...
+
+    @LogMessage(level = WARN)
+    @Message(id = 8, value = "Default procedures were disabled already, based on the %s deployment configuration")
+    void defaultProceduresDisabledAlready(String deploymentName);
```
bstansberry, line 62: Same comment as above.

bstansberry (changes requested):
Can we always signal MicroProfileHealthReporter, when we deploy something, whether defaults are disabled or not? And then MPHR detects cases where some deployments disable default procedures and others don't?  That seems like the problematic usage we want to WARN about.

This adds complexity though as that kind of tracking perhaps should only work with top level deployments, while CDIExtension processes subdeployments too. But maybe that's not a problem; i.e. if someone deploys an ear they should configure this at the ear level not at the subdeployment level.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/_private/MicroProfileHealthLogger.java
fabiobrz: +1, thanks. Now there's just one WARN that we output upon the problematic condition you described in the review summary.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/_private/MicroProfileHealthLogger.java
fabiobrz: Fixed (hopefully...), see my previous comment.

fabiobrz:
> Can we always signal MicroProfileHealthReporter, when we deploy something, whether defaults are disabled or not? And then MPHR detects cases where some deployments disable default procedures and others don't? That seems like the problematic usage we want to WARN about.
> 
> This adds complexity though as that kind of tracking perhaps should only work with top level deployments, while CDIExtension processes subdeployments too. But maybe that's not a problem; i.e. if someone deploys an ear they should configure this at the ear level not at the subdeployment level.

Thanks @bstansberry - I tried to address (quickly tbh, so it might need some polish) what you suggested, which I agree on. I added a couple of tests for a new use case, i.e. deploying multiple WAR, which actually should be a discouraged proactice when it comes to Microservices.
Then I tried to do the same for the EAR case that you mentioned, but there we're blocked by https://issues.redhat.com/browse/WFLY-12835

WDYT?

bstansberry:
Hi @fabiobrz -- thanks for trying that out. I don't think it can be done in a robust manner though; I don't think my idea was a good one. As an alternative I sent you https://github.com/fabiobrz/wildfly/pull/18

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/_private/MicroProfileHealthLogger.java
fabiobrz: And now changed in to an `INFO`, resolving.

fabiobrz:
> Hi @fabiobrz -- thanks for trying that out. I don't think it can be done in a robust manner though; I don't think my idea was a good one. As an alternative I sent you [fabiobrz#18](https://github.com/fabiobrz/wildfly/pull/18)

Thanks @bstansberry! I merged your changes into this PR feature branch.

bstansberry (approved):

bstansberry:
Thanks @fabiobrz and @xstefank

PR #17740, opened 2024-03-19
[WFLY-19150] Clean up server.jvm.args handling

bstansberry:
- Fix places that were not passing through the 'server.jvm.args' maven property value to surefire 

- Remove some (but not all) redundant -Dmaven.repo.local=${settings.localRepository} setting, as it's in server.jvm.args via surefire.system.args 

- Remove crufty -Dorg.wildfly.unsupported.skip.jakarta.transformer=true settings 

- Remove overridden setting of the surefire 'server.jvm.args' system property that just does the same as the global config setting in testsuite/integration/pom.xml

https://issues.redhat.com/browse/WFLY-19150

### DIFF
--- a/testsuite/domain/pom.xml
+++ b/testsuite/domain/pom.xml
@@ -341,8 +341,6 @@
                                     <argLine>-Djava.util.logging.manager=org.jboss.logmanager.LogManager ${jvm.args.ip.client} ${jvm.args.timeouts} ${surefire.system.args}</argLine>
                                     
                                     <systemPropertyVariables>
-                                        <server.jvm.args>-Dmaven.repo.local=${settings.localRepository}</server.jvm.args>
-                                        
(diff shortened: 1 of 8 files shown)


testsuite/domain/pom.xml
```diff
@@ -341,8 +341,6 @@
                                     <argLine>-Djava.util.logging.manager=org.jboss.logmanager.LogManager ${jvm.args.ip.client} ${jvm.args.timeouts} ${surefire.system.args}</argLine>
                                     
                                     <systemPropertyVariables>
-                                        <server.jvm.args>-Dmaven.repo.local=${settings.localRepository}</server.jvm.args>
```
bstansberry, line 344: Note that a few lines below what the GH diff UI shows is another line that configures this same property. So this one is cruft.

luck3y (approved):

bstansberry:
Thanks @luck3y

PR #17742, opened 2024-03-20
WFLY-14559: Added six RESTEasy parameters to jaxrs

ronsigal:
https://issues.redhat.com/browse/WFLY-14559


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
```diff
@@ -24,10 +24,16 @@ public class JaxrsConstants {
     public static final String RESTEASY_GZIP_MAX_INPUT = "resteasy-gzip-max-input";
     public static final String RESTEASY_JNDI_RESOURCES = "resteasy-jndi-resources";
     public static final String RESTEASY_LANGUAGE_MAPPINGS = "resteasy-language-mappings";
+    public static final String RESTEASY_MATCH_CACHE_ENABLED = "resteasy-match-cache-enabled";
```
jamezp, line 27: Do we want to use the `resteasy-` prefix on these? I don't have a strong opinion, but leaving it off is less typing for administrators. I left it off for the tracing attributes.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void testRejectingTransformersEAP74() throws Exception {
+        FailedOperationTransformationConfig transformationConfig = new FailedOperationTransformationConfig();
+
+        transformationConfig.addFailedAttribute(PathAddress.pathAddress(JaxrsExtension.SUBSYSTEM_PATH),
+                new FailedOperationTransformationConfig.NewAttributesConfig(JaxrsAttribute.TRACING_TYPE, JaxrsAttribute.TRACING_THRESHOLD));
+
+        testRejectingTransformers(transformationConfig, ModelTestControllerVersion.EAP_7_4_0);
```
jamezp, line 60: We probably need a `EAP_8_0_0` test here too.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
```diff
@@ -0,0 +1,78 @@
...
+    public void testExpressions() throws Exception {
+        standardSubsystemTest("jaxrs-expressions.xml");
+    }
+
+    @Test
+    public void testRejectingTransformersEAP74() throws Exception {
+        FailedOperationTransformationConfig transformationConfig = new FailedOperationTransformationConfig();
+
+        transformationConfig.addFailedAttribute(PathAddress.pathAddress(JaxrsExtension.SUBSYSTEM_PATH),
+                new FailedOperationTransformationConfig.NewAttributesConfig(JaxrsAttribute.TRACING_TYPE, JaxrsAttribute.TRACING_THRESHOLD));
```
jamezp, line 58: We need to add the new attributes here for both EAP7 and EAP8 testing.

Also note those two attributes shouldn't fail with EAP8 or WildFly 27+ testing.

jamezp (changes requested):
We need transformers for the attributes. They can all be discarded if the default value is set, but they need to be rejected if a different value is set. See the `org.jboss.jaxrs.ResteasyExtensionTransformerRegistration`.

As you said too the `ResteasyAttributeTestCase` is failing as well :)

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
ronsigal: I can't remember where, but somewhere there's a transformation from, e.g., "resteasy-language-mappings" to "resteasy.language.mappings". Wouldn't omitting "resteasy" screw that up?

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
ronsigal: I've got an updated transformation test JaxrsSubsystem30TestCase and a new JaxrsSubsystem40TestCase running, but when I add an EAP_8_0_0 test to JaxrsSubsystem40TestCase, 
```
ModelVersion subsystemModelVersion = controllerVersion.getSubsystemModelVersion(JaxrsExtension.SUBSYSTEM_NAME);

```
fails with a NullPointerException. I have no idea ...

ronsigal: As soon as the current round of tests finish, I'll upload what I've got so far.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
jamezp: That's easily fixable though :) This is how I did it for the tracing attributes. I just simply added them as context parameters https://github.com/wildfly/wildfly/blob/1f68b0ad54b7899650a782a8efdde0cec196865c/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java#L138-L143.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
jamezp: We'll just have to leave that off. It doesn't look like it's working, so nothing you can do there.

jaxrs/src/main/java/org/jboss/as/jaxrs/ResteasyExtensionTransformerRegistration.java
```diff
@@ -32,14 +33,49 @@ public String getSubsystemName() {
...
+                        JaxrsAttribute.RESTEASY_PROXY_IMPLEMENT_ALL_INTERFACES)
+                .addRejectCheck(RejectAttributeChecker.DEFINED,
+                        JaxrsAttribute.TRACING_TYPE, JaxrsAttribute.TRACING_THRESHOLD,
+                        JaxrsAttribute.RESTEASY_MATCH_CACHE_ENABLED,
+                        JaxrsAttribute.RESTEASY_MATCH_CACHE_SIZE,
+                        JaxrsAttribute.RESTEASY_PATCH_FILTER_DISABLED,
+                        JaxrsAttribute.RESTEASY_PATCH_FILTER_LEGACY,
+                        JaxrsAttribute.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR,
+                        JaxrsAttribute.RESTEASY_PROXY_IMPLEMENT_ALL_INTERFACES);
+    }
```
jamezp, line 60: This method should have have changed. Only the new method should be added.

jamezp (changes requested):

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
ronsigal: Oh, I see. Well, I have mixed feelings. Once could argue that the tracing parameters are a class unto themselves, but the six new parameters are more like the existing ones, which introduces two different conventions.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
ronsigal: So there's nothing to do here?

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
ronsigal: And nothing to do here?

jaxrs/src/main/java/org/jboss/as/jaxrs/ResteasyExtensionTransformerRegistration.java
ronsigal: Ok

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
jamezp: Yeah. It's more opinion UX than real UX probably :)

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
```diff
@@ -0,0 +1,119 @@
...
+                        JaxrsAttribute.RESTEASY_PATCH_FILTER_LEGACY,
+                        JaxrsAttribute.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR,
+                        JaxrsAttribute.RESTEASY_PROXY_IMPLEMENT_ALL_INTERFACES
+                        ));
+
+        testRejectingTransformers74(transformationConfig, ModelTestControllerVersion.EAP_7_4_0);
+    }
+
+//    @Test
+    public void testRejectingTransformersEAP80() throws Exception {
```
jamezp, line 72: Minor change here, but not critical. We should probably leave the `@Test`, but add `@Ignore` with a reason as to why. We can reference https://issues.redhat.com/browse/WFCORE-6700.

ronsigal: Oh, OK, good.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
ronsigal: What if we add a shortened version of **every** parameter, all at one time, in a new feature request. Then the old names work and each has a shortened version as well.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
ronsigal: Done

ronsigal:
Failure on Galleon Linux - JDK 11: org.jboss.as.test.integration.messaging.jms.context.notclosinginjectedcontext.NotClosingInjectedContextTestCase

Not relevant

ronsigal:
Linux - JDK 11: several Elytron tests failed. Also not relevant

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
jamezp: We can just keep the `resteasy` prefix. There isn't really a way to allow for two different attribute names that I'm aware of.

jamezp (approved):
Overall I think this is okay. Looking at the help descriptions, I don't find them all that helpful. However, they are all like this so we could improve this in a follow up pull request. I've https://issues.redhat.com/browse/WFLY-19190 as a follow up for this.

darranl:
I am removing the 32.x label on this one, this PR looks to be adding at the "default" stability level and does not have the required approvals to go in at that level.

This is a candidate that could make it into WildFly under the WildFly feature development process but if we want to wait to get it in at the default stability level we will need to wait until this is prioritised internally and the appropriate process followed.

ronsigal:
>if we want to wait to get it in at the default stability level we will need to wait until this is prioritised internally and the appropriate process followed

@jamezp, what do you think?

bstansberry (changes requested):
The new attributes need to be at community stability level.

ronsigal:
> The new attributes need to be at community stability level.

Looking at "WildFly Feature Development Process" (https://docs.google.com/document/d/15_yKhW74-X9s2zUhs_ZUuZ3h-RlMfH5xWmHHsfYf1AA/edit?pli=1#heading=h.vaxfllj8snno), I think what's missing (other than changing the Stability Level check mark in the analysis document) seems to be:

* Feature team: 3rd party
* Test Verification: code review, presumably with the 3rd party

Am I missing anything? Any suggestions for a 3rd party?

jamezp:
I don't really have a suggestion for a 3rd party, but I would maybe send a message on Zulip in wildfly-developers and ask there.

ronsigal:
> send a message on Zulip

Will do. Thanks.

jamezp:
I'm closing this as a new one will be coming.

PR #17750, opened 2024-03-22
[WFLY-19160] EJB: create multinode wildfly-http-client test with PROV…

tadamski:
…IDER_URL

https://issues.redhat.com/browse/WFLY-19160

____


tadamski (commented):
Failing tests seem unrelated.

testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/ejb/http/EjbOverHttpProviderUrlTestCase.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Collections;
+import java.util.Hashtable;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.COMPOSITE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STEPS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.test.shared.PermissionUtils.createFilePermission;
+import static org.jboss.as.test.shared.PermissionUtils.createPermissionsXmlAsset;
```
bstansberry, line 46: Please put the static imports first, then java.*

bstansberry (changes requested):
Minor import ordering thing; otherwise LGTM.

Just a suggestion -- perhaps use a base class? This and EjbOverHttpTestCase look to be 95% the same; just a small different setting the the ServerSetupTask and then a small difference in how the InitialContext is established. (I may be wrong; I wasn't super-careful about the comparison.)

testsuite/integration/multinode/src/test/java/org/jboss/as/test/multinode/ejb/http/EjbOverHttpProviderUrlTestCase.java
tadamski: @bstansberry I have refactored all ejb-over-http multinode tests into one (and moved the static imports).

tadamski:
@bstansberry Those failing tests were intermittent and the reason is inconsistent deployment configuration by arquillian. If there are many deployments on one node (such as client in EjbOverHttpTestCase) there is a race condition and only the first provided config is used in all the tests. Because of that, there were interchangeable failures of wrong-credentials/good-credentials tests. I have created ARQ-2233 to deal with this issue and for now I have ignored wrong credentials tests case (and it's deployment).

bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #17751, opened 2024-03-22
WFLY-19161: add documentation for configuration file aliases

michpetrov:
Issue: [WFLY-19161](https://issues.redhat.com/browse/WFLY-19161)

____

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/Management_Clients.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/Management_Clients.adoc
@@ -161,10 +161,33 @@ standard configuration files, e.g. `standalone-full.xml`,
 similar to the default `standalone.xml` file but includes additional
 subsystems not present in the default configuration. If you prefer to
 use one of these files as your server configuration, you can specify it
-with the _[line-through]*c*_ or _-server-config_ command line argument:
+with the *_-c_* or *_--server-config_* command line argument:
 
 * `bin/standalone.sh -c=standalone-full.xml`
 * `bin/standalone.sh --server-config=standalone-ha.xml`
+
+==== Aliases
+
+The configuration files can also be referenced by an alias instead of the file name:
+
+* `bin/standalone.sh -c ha`
+
+.Available aliases
+|===
+| Configuration file               | Aliases
+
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/_admin-guide/Management_Clients.adoc
```diff
@@ -161,11 +161,34 @@ standard configuration files, e.g. `standalone-full.xml`,
...
+
+* `bin/standalone.sh -c ha`
+
+.Available aliases
+|===
+| Configuration file | Aliases
+
+| standalone-full.xml     | full
+
+| *-ha.xml                | ha
```
bstansberry, line 181: @michpetrov What does '*' mean here and in the others?

bstansberry:
@michpetrov Adding this content in https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/Getting_Started_Guide.adoc#starting-wildfly-with-an-alternate-configuration would be good too.

darranl:
Adding 32.x as we need the doc in Final, not adding other labels as not strictly necessary for Beta.

docs/src/main/asciidoc/_admin-guide/Management_Clients.adoc
bstansberry: @michpetrov Why not just spell it out? Any symbol people will try and interpret what it means and may come up with novel interpretations, or will just wonder why they're needing to use mental energy.

bstansberry:
> @bstansberry I assume a link from that section to this one will do?

OK, that works. That section doesn't list all the available files, so copying in this content that does list them all would be confusing.

docs/src/main/asciidoc/_admin-guide/Management_Clients.adoc
michpetrov: @bstansberry well, it's superfluous information and AFAIK it's not good UX to have each line start with the same word, but I'm not a docs person so if you don't mind it I'll keep it there

bstansberry (approved):

bstansberry:
Thanks @michpetrov!

PR #17760, opened 2024-03-26
[WFLY-19167] war containing JDBC drivers deploys an extra driver

gaol:
Issue: https://issues.redhat.com/browse/WFLY-19167

____

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/datasource/JdbcDriversInWarTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/datasource/JdbcDriversInWarTestCase.java
@@ -11,6 +11,7 @@ import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.as.controller.client.helpers.Operations;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.test.integration.jca.JcaMgmtBase;
+import org.jboss.as.test.integration.management.util.MgmtOperationException;
 import org.jboss.dmr.ModelNode;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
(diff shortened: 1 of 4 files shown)


The-Huginn:
Apologies for the spam in the description by the wildfly-bot :sweat_smile: . We accidentally had 2 versions deployed, updating one's another actions. Please feel free to remove all the clutter from the description.

gaol:
no worries @The-Huginn :)

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/datasource/JdbcDriversInWarTestCase.java
```diff
@@ -0,0 +1,138 @@
...
+                .addAsServiceProviderAndClasses(Driver.class, TestDriver.class));
+        war.addAsLibrary(ShrinkWrap.create(JavaArchive.class, "testDriver2.jar")
+                .addClasses(TestDriver2.class, DummyDataSource.class, DummyXADataSource.class)
+                .addAsServiceProviderAndClasses(Driver.class, TestDriver2.class));
+        return war;
+    }
+
+    @Test
+    public void testJdbcDrivers() throws Exception {
+        String driverName = DEPLOYMENT;
```
tadamski, line 54: @gaol maybe it will also be useful to verify the number of deployed drivers?

tadamski (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/datasource/JdbcDriversInWarTestCase.java
gaol: right, good catch ! 
will try to get the jdbc driver using short name and it should fail.
checking number of deployed jdbc drivers maybe confusing as we need to filter out the default `h2` driver.

gaol:
updated according to the feedback by checking if the jdbc driver with the short name is not installed. please review again, @tadamski  thanks :)

tadamski (approved):
@gaol LGTM now. Thanks for fixing it!

bstansberry:
Thanks @gaol and @tadamski

PR #17764, opened 2024-03-26
[WFLY-19170] Make WildFly Preview run at preview stability OOTB

bstansberry:
https://issues.redhat.com/browse/WFLY-19170

____

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #17753, opened 2024-03-22
[WFLY-19164] Move MVCKrazoSmokeTestCase to testsuite/preview/manualmode

bstansberry:
https://issues.redhat.com/browse/WFLY-19163
https://issues.redhat.com/browse/WFLY-19164

PR #17766, opened 2024-03-27
[WFLY-19062] Add preview stability support for Jakarta MVC to standard WildFly

bstansberry:
https://issues.redhat.com/browse/WFLY-19062

This also incorporates fixes for:

https://issues.redhat.com/browse/WFLY-19170 (see also #17764)
https://issues.redhat.com/browse/WFLY-19163 (see also #17753)

____

### DIFF
--- a/testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/mvc/MVCKrazoSmokeTestCase.java
+++ b/testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/mvc/MVCKrazoSmokeTestCase.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.test.manual.mvc;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.COMPOSITE;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.EXTENSION;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STEPS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+
+import org.jboss.arquillian.container.test.api.ContainerController;
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.container.test.api.TargetsContainer;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.junit.InSequence;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ServerSetupTask;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.controller.operations.common.Util;
+import org.jboss.as.test.integration.common.HttpRequest;
+import org.jboss.as.test.shared.ServerReload;
+import org.jboss.as.test.shared.TestSuiteEnvironment;
+import org.jboss.dmr.ModelNode;
+import org.jboss.logging.Logger;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.shrinkwrap.descriptor.api.Descriptors;
+import org.jboss.shrinkwrap.descriptor.api.webapp25.WebAppDescriptor;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+//@ServerSetup(MVCKrazoSmokeTestCase.Setup.class) TODO migrate this to testsuite/integration/basic and let WFARQ call the ServerSetupTask
+public class MVCKrazoSmokeTestCase {
+
+    private static final Logger log = Logger.getLogger(MVCKrazoSmokeTestCase.class.getName());
+
+    // TODO migrate this to testsuite/integration/basic and drop the ContainerController aspect used in manualmode
+    private static final String CONTAINER = "stability-preview";
+    // TODO migrate this to testsuite/integration/basic and let WFARQ call the ServerSetupTask via @ServerSetup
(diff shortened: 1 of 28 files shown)


testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -83,7 +83,13 @@ public abstract class LayersTestBase {
      * Included in the return value of {@link #getExpectedUnusedInAllLayers()}
      * only when testing provisioning from the wildfly feature pack.
      */
-    public static final String[] NO_LAYER_WILDFLY = {};
+    public static final String[] NO_LAYER_WILDFLY = {
+            // Preview stability 'mvc-krazo' layer cannot be provisioned in OOTB standard wildfly
```
jfdenise, line 87: When reading the comment, I thought at default standalone config. <perhaps adding a mention of layer.

testsuite/layers-expansion/pom.xml
```diff
@@ -760,6 +760,7 @@
                                                 <layer>messaging-activemq</layer>
                                                 <layer>micrometer</layer>
                                                 <layer>mod_cluster</layer>
+                                                <layer>mvc-krazo</layer>
```
jfdenise, line 763: Perhaps add a comment that says that being preview, nothing is provisioned although present. I understand that this comment would have to be removed when promoting this subsystem.

testsuite/layers-expansion/pom.xml
```diff
@@ -869,6 +870,7 @@
...
```
jfdenise, line 873: Perhaps add a comment that says that being preview, nothing is provisioned although present. I understand that this comment would have to be removed when promoting this subsystem.

jfdenise (approved):

testsuite/layers-expansion/pom.xml
bstansberry: This one is a mistake. It should not be in this execution until it will actually be provisioned. Adding it here would be part of promoting the feature to community.

But commenting it out with a comment is better than deleting it. :)

testsuite/layers-expansion/pom.xml
jfdenise: I think that is the wildfly full all layers for HA.

bstansberry: I misread where in the code this line is. +1 -- this is the same as your other comment above and I'll do the same re commenting it out.

bstansberry: @jfdenise Done. (I moved the whole thing lower into the section with 'wildfly' FP only layers.)

scottmarlow:
Pasting from https://issues.redhat.com/browse/WFLY-19062:
> Adding it to wildfly-ee instead is also a possibility to consider, but as Jakarta MVC is not yet a part of any Jakarta EE platform or profile, I think it's reasonable to consider it as more subject to incompatible change than most wildfly-ee technologies, and thus a better fit for 'wildfly'.

:+1:  That makes sense!

boms/common-ee/pom.xml
```diff
@@ -4808,6 +4808,37 @@
                 <version>${version.org.reactivestreams}</version>
             </dependency>
 
+            <dependency>
```
scottmarlow, line 4811: Should non-EE artifacts be added to boms/common-ee or boms/common-expansion?

scottmarlow:
Pasting from https://github.com/eclipse-ee4j/krazo/issues/352:
> We could implement a PoC for a global WildFly module, so we can run the TCK against it and submit it as PR to WildFly itself.

@erdlet could it make sense for the WildFly project to run the Eclipse Krazo TCK against the https://github.com/eclipse-ee4j/krazo/tree/master/tck?  Perhaps via the https://ci.eclipse.org/krazo/job/tck-wildfly job?

erdlet:
Hi @scottmarlow 

thanks for integratin Krazo into Wildfly.

TBH I retired from Krazo,so I guess its better to ask @ivargrimstad or @chkal.

Sorry that I can't help you more

boms/common-ee/pom.xml
bstansberry: @scottmarlow Non-EE artifacts can go in common-ee. But a realize now these particular ones should be in common-expansion not common-ee.

The 'ee' refers to the 'wildfly-ee' feature pack, which is much more than EE; it's just the name we chose for the 'base' FP back in the WF 19/20 days we when split out the 'expansion' stuff. Down stream the wildfly-ee feature pack becomes the 'eap' feature pack used for base EAP (which is much more the Jakarta EE), while the 'wildfly' feature pack becomes the 'eap-xp' FP used for EAP XP. Upstream I refer to things specific to the 'wildfly' feature pack (boms, testsuite modules etc) as 'expansion'. But the feature pack itself is just 'wildfly' because in community 99%+ of people only deal with that FP directly, with wildfly-ee as an internal detail.

The boms/xxx-ee modules are meant to relate to wildfly-ee, while boms/xxx-expansion relate to the 'wildfly' feature pack. These deps are only relevant to the 'wildfly' FP so I will move them to boms/common-expansion.

bstansberry:
@scottmarlow We have a TCK runner for this:

https://github.com/wildfly-extras/mvc-krazo-feature-pack/tree/main/testsuite/tck

That can be run against any WildFly release by passing system props to maven. The README in that module describes that. Snapshots can be tested locally that way. WildFly Preview 31 passes that, and I'm confident this PR does too. (An earlier variant did but I've tweaked some irrelevant stuff since I last tried.)

I'm going to add a github action to that repo that will take a GH repo and tag/branch/sha as params, values meant to point to a wildfly repo. It will then check out that repo and build wildfly, making the resulting maven repository from the build available to a call to run the TCK. This will allow testing WF snapshots on GH instead of locally. That action could then be called from some broader infrastructure, e.g. if we wanted some action in wildfly-tck-runners to call everything, or perhaps a jenkins job.

Note that I expect that repo may some day go away and the various pieces get moved, e.g. the bulk of it into this repo and the tck runner as a subtree in wildfly-tck-runners. But for now I like having it external as it gives me experience of the issues that come with maintaining something like this externally to the main WF code base. That something we'll be doing more of over time and it's good to build expertise.

boms/common-ee/pom.xml
bstansberry: @scottmarlow I've moved these to common-expansion. While doing it I realized one of them wasn't needed at all so I removed that.

When doing that I also rebased on current main and took the opportunity to squash down from three commits to two.

darranl:
Just adding the "Critical" label so we can discuss.

bstansberry:
TCK results are here: https://github.com/wildfly-extras/mvc-krazo-feature-pack/actions/runs/8539883565/

testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/mvc/MVCKrazoSmokeTestCase.java
```diff
@@ -0,0 +1,171 @@
...
+    private static Deployer deployer;
+
+    // TODO migrate this to testsuite/integration/basic and inject the http listener URL
+//    @ArquillianResource
+//    private URL url;
+
+
+    // TODO migrate this to testsuite/integration/basic and drop this
+    @Test
+    @InSequence(-1)
```
jamezp, line 93: This should not block merging, but any reason not to do this in a `@Before` method?

scottmarlow:
Nice to see the TCK tests passed! `Tests run: 138, Failures: 0, Errors: 0, Skipped: 0`

scottmarlow (approved):

jamezp (approved):
I"m approving this because I don't think my comments should blocking merging and can be changed in a follow up PR.

PR #17800, opened 2024-04-08
[WFLY-19217] Upgrade to SmallRye Reactive Messaging 4.21.0

kabir:
Routine update to the latest and greatest https://issues.redhat.com/browse/WFLY-19217

____

> Wildfly issue links:
> * [WFLY-19310](https://issues.redhat.com/browse/WFLY-19310)


kabir:
@bstansberry This is because the Mutiny upgrade pulls in org.jctools:jctools-core:4.0.3. Adding a module dependency on org.jctools from the Mutiny module does not work, since the current jctools version in WildFly  2.1.2 which is missing some classes that Mutiny needs.  The current user of jctools in WildFly is artemis, which is currently on version 2.32.0.

I see that the next version of artemis is 2.33.0, and that [upgrades to jctools 4.0.2](https://github.com/apache/activemq-artemis/blob/2.33.0/pom.xml#L115). So I will try to update here to 4.0.3 and see how that goes.

PR #17803, opened 2024-04-08
[WFLY-19177] [preview] Update elytron-oidc-client subsystem transformer to add a transformer from model version 3 to model version 2

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFLY-19177
Related discussions: https://github.com/wildfly/wildfly/pull/17705#discussion_r1536073227

____

### DIFF
--- a/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
+++ b/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
@@ -7,6 +7,9 @@ package org.wildfly.extension.elytron.oidc;
 
 import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_1_0_0;
 import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_2_0_0;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcClientSubsystemModel.VERSION_3_0_0;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcDescriptionConstants.SCOPE;
+import static org.wildfly.extension.elytron.oidc.ElytronOidcDescriptionConstants.SECURE_DEPLOYMENT;
 import static org.wildfly.extension.elytron.oidc.ElytronOidcDescriptionConstants.SECURE_SERVER;
 
 import org.jboss.as.controller.ModelVersion;
@@ -14,6 +17,8 @@ import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.transform.ExtensionTransformerRegistration;
 import org.jboss.as.controller.transform.SubsystemTransformerRegistration;
 import org.jboss.as.controller.transform.description.ChainedTransformationDescriptionBuilder;
+import org.jboss.as.controller.transform.description.DiscardAttributeChecker;
+import org.jboss.as.controller.transform.description.RejectAttributeChecker;
 import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
 import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
 
@@ -29,25 +34,32 @@ public class ElytronOidcSubsystemTransformers implements ExtensionTransformerReg
 
         ChainedTransformationDescriptionBuilder chainedBuilder = TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(registration.getCurrentSubsystemVersion());
         // 3.0.0 (WildFly 32) to 2.0.0 (WildFly 29)
-//        from3(chainedBuilder);
+        from3(chainedBuilder);
         // 2.0.0 (WildFly 29) to 1.0.0 (WildFly 28)
         from2(chainedBuilder);
 
-        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_1_0_0.getVersion() });
+        chainedBuilder.buildAndRegister(registration, new ModelVersion[] { VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion() });
     }
 
     private static void from2(ChainedTransformationDescriptionBuilder chainedBuilder) {
         ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_2_0_0.getVersion(), VERSION_1_0_0.getVersion());
         builder.rejectChildResource(PathElement.pathElement(SECURE_SERVER));
     }
-    // TODO: Figure out how to handle the transformer with a preview level attribute being added to the subsystem
-//    private static void from3(ChainedTransformationDescriptionBuilder chainedBuilder) {
-//        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_3_0_0.getVersion(), VERSION_2_0_0.getVersion());
-//        builder.addChildResource(PathElement.pathElement(SECURE_SERVER))
-//                .getAttributeBuilder();
-//
-//        builder.addChildResource(PathElement.pathElement(SECURE_DEPLOYMENT))
-//                .getAttributeBuilder();
-//    }
+
+    private static void from3(ChainedTransformationDescriptionBuilder chainedBuilder) {
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_3_0_0.getVersion(), VERSION_2_0_0.getVersion());
+
+        builder.addChildResource(PathElement.pathElement(SECURE_SERVER))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, SCOPE)
+                .setDiscard(DiscardAttributeChecker.ALWAYS, SCOPE)
+                .end();
+
+        builder.addChildResource(PathElement.pathElement(SECURE_DEPLOYMENT))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, SCOPE)
+                .setDiscard(DiscardAttributeChecker.ALWAYS, SCOPE)
+                .end();
+    }
 
 }
--- a/elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
+++ b/elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.elytron.oidc;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+
+import org.jboss.as.controller.ModelVersion;
+import org.jboss.as.controller.PathAddress;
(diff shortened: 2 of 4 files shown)


elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -29,25 +34,31 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
...
+        builder.addChildResource(PathElement.pathElement(SECURE_SERVER))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, SCOPE)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, SCOPE);
+
+        builder.addChildResource(PathElement.pathElement(SECURE_DEPLOYMENT))
+                .getAttributeBuilder()
+                .addRejectCheck(RejectAttributeChecker.DEFINED, SCOPE)
+                .setDiscard(DiscardAttributeChecker.UNDEFINED, SCOPE);
+        builder.rejectChildResource(PathElement.pathElement(SCOPE));
```
fjuma, line 61: This line shouldn't be needed since `scope` isn't a child resource (it's an attribute for `secure-server` and `secure-deployment`).

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -0,0 +1,79 @@
+package org.wildfly.extension.elytron.oidc;
```
fjuma, line 1: Missing copyright header.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    protected void compareXml(String configId, String original, String marshalled) {
+        //
+    }
+
+    @Override
+    protected String getSubsystemXml() throws IOException {
+        return readResource("elytron-oidc-client-transform.xml");
```
fjuma, line 77: I also see an `elytron-oidc-client-reject.xml` file but it seems to be empty. Is that getting used anywhere?

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
PrarthonaPaul: Fixed, but I think discard and rejection is not working right still, causing the tests to fail when scope is defined in the XML files and pass when it is commented out.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
PrarthonaPaul: added.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
PrarthonaPaul: It should be there now. I hadn't written a rejection test before the last push. 
Thanks!

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public ElytronOidcClientSubsystemTransformerTestCase(ModelTestControllerVersion controllerVersion, ModelVersion version) {
+        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_2_0_PREVIEW, ElytronOidcSubsystemSchema.CURRENT.get(Stability.PREVIEW));
+        this.controllerVersion = controllerVersion;
+        this.modelVersion = version;
+    }
+
+    @Test
+    public void testTransformations() throws Exception {
+        final ModelVersion version = modelVersion;
```
PrarthonaPaul, line 54: Other transformer tests I saw use `controller.getSubsystemModelVersion(getMainSubsystemName())`, but `ModelTestControllerVersion.EAP_8_0_0` does not have a field for elytron-oidc-client, and the above returns null, so I had to use `modelversion`.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemTransformers.java
```diff
@@ -29,25 +35,32 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
...
-//                .getAttributeBuilder();
-//    }
+
+    private static void from3(ChainedTransformationDescriptionBuilder chainedBuilder) {
+        ResourceTransformationDescriptionBuilder builder = chainedBuilder.createBuilder(VERSION_3_0_0.getVersion(), VERSION_2_0_0.getVersion());
+
+        builder.addChildResource(PathElement.pathElement(SECURE_SERVER))
+                .getAttributeBuilder()
+                .addRejectCheck(new RejectAttributeChecker.SimpleRejectAttributeChecker(ModelNode.FALSE), SCOPE)
+                .setDiscard(DiscardAttributeChecker.ALWAYS, SCOPE)
```
PrarthonaPaul, line 56: For some reason if I use the following, the scope attribute does not get discarded or rejected properly. It gets sent to version 2, which does not recognize this attribute. 
```
.addRejectCheck( RejectAttributeChecker.DEFINED, SCOPE)
.setDiscard(DiscardAttributeChecker.UNDEFINED, SCOPE)
``` 
So, I had to change it to the reject and discard checkers above. Not entirely sure if this is correct, however.

Skyllarr: While the `.setDiscard(DiscardAttributeChecker.ALWAYS, SCOPE)` looks okay to me, I am not sure about the reject check. Since from the javadoc of SimpleRejectAttributeChecker: `Rejects the attribute if the value is equal to the specified value.` 

so it seems that this will reject the attribute SCOPE only if the value of it is false. And we want to reject it with any value

PrarthonaPaul: Thanks @Skyllarr 
I have updated it to 
`.addRejectCheck( RejectAttributeChecker.DEFINED, SCOPE)`

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -0,0 +1,125 @@
...
+
+        checkSubsystemModelTransformation(services, version, null, false);
+        ModelNode transformed = services.readTransformedModel(version);
+        Assert.assertTrue(transformed.isDefined());
+    }
+
+    @Test
+    public void testRejection() throws Exception {
+        testRejectingTransformers(ModelTestControllerVersion.EAP_8_0_0, "elytron-oidc-client-reject.xml", new FailedOperationTransformationConfig()
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronOidcDescriptionConstants.SCOPE)),
```
Skyllarr, line 65: Just wondering why this doesn't have to reference SECURE-DEPLOYMENT in the path.. As it now looks for scope attribute on the elytron-oidc-client subsystem directly or ?

skyllarr (commented):
Hi @PrarthonaPaul , just a reminder to remembersquashing the commits into a single commit

elytron-oidc-client/src/test/resources/org/wildfly/extension/elytron/oidc/elytron-oidc-client-reject.xml
```diff
@@ -0,0 +1,16 @@
+<subsystem xmlns="urn:wildfly:elytron-oidc-client:preview:2.0">
```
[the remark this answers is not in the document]
PrarthonaPaul: Added!

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
PrarthonaPaul: I have changed it to 
`SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronOidcDescriptionConstants.SECURE_DEPLOYMENT, "wildfly-reject-deployment-with-scope.war"), PathElement.pathElement(ElytronOidcDescriptionConstants.SCOPE))`
I think `PathElement.pathElement(ElytronOidcDescriptionConstants.SCOPE)` is a wildcard path, so what you suggested is definitely less ambiguous and it works better. 
Thanks for catching it!

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -0,0 +1,127 @@
...
+        ModelNode transformed = services.readTransformedModel(version);
+        Assert.assertTrue(transformed.isDefined());
+    }
+
+    @Test
+    public void testRejection() throws Exception {
+        testRejectingTransformers(ModelTestControllerVersion.EAP_8_0_0, "elytron-oidc-client-reject.xml", new FailedOperationTransformationConfig()
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronOidcDescriptionConstants.SECURE_DEPLOYMENT, "wildfly-reject-deployment-with-scope.war"), PathElement.pathElement(ElytronOidcDescriptionConstants.SCOPE)),
+                        new FailedOperationTransformationConfig.NewAttributesConfig(ElytronOidcDescriptionConstants.SCOPE))
+                .addFailedAttribute(SUBSYSTEM_ADDRESS.append(PathElement.pathElement(ElytronOidcDescriptionConstants.SECURE_SERVER, "wildfly-reject-deployment-with-scope.war"), PathElement.pathElement(ElytronOidcDescriptionConstants.SCOPE)),
```
Skyllarr, line 67: @PrarthonaPaul This should be `wildfly-with-scope.war` instead of the "wildfly-reject-deployment-with-scope.war" r

PrarthonaPaul: Thanks for catching it! 
Should be good now. 
Thanks @Skyllarr

skyllarr (approved):
@PrarthonaPaul just a minor note that the commit message should say from model version 3 to model version 2

bstansberry (approved):

PR #17810, opened 2024-04-10
[WFLY-19224] Suppress CVE-2023-1973 as WildFly no longer uses Undertow's authentication mechanisms.

darranl:
https://issues.redhat.com/browse/WFLY-19224

@bstansberry this is a little easier to see what is needed to suppress a CVE.

Firstly it is not required that we have an until (and if we are scanning older branches this could be an issue) but I was adding these as a mechanism to ensure we at least triage again in the future.

On the original PR you mentioned also adding comments to record in the file why the suppression was added, here I have added it in the notes section which means it will also be included in the resulting HTML report of suppressed CVEs so we can quickly check why it was suppressed there.

____

### DIFF
--- a/sca-overrides/owasp-suppressions.xml
+++ b/sca-overrides/owasp-suppressions.xml
@@ -216,5 +216,15 @@
       ]]></notes>
       <packageUrl regex="true">^pkg:maven/org\.jboss\.resteasy\.spring/resteasy\-spring@.*$</packageUrl>
       <cve>CVE-2018-1051</cve>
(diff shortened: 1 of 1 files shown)


jamezp (approved):

bstansberry (approved):
Excellent; thanks @darranl -- that's a nice improvement over an xml comment

bstansberry:
@darranl Please take care of resolving the JIRA. JIRA is acting up right now so I can't do it.

PR #17833, opened 2024-04-19
See what happens if we support @ManagedBean without the class being p…

bstansberry:
…resent

Just for fun in case @jamezp wants to play.


bstansberry:
I'm going to close this to help clean up the PR queue.

PR #17834, opened 2024-04-22
[WFLY-19206] MailTesterServlet would need better strategy to read back the messages sent

Dkafetzis:
https://issues.redhat.com/browse/WFLY-19206

____

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
@@ -7,7 +7,6 @@ package org.jboss.as.test.integration.mail.basic;
 
 
 import java.io.IOException;
-import java.util.concurrent.TimeUnit;
 
 import jakarta.annotation.Resource;
 import jakarta.mail.Address;
@@ -24,6 +23,7 @@ import jakarta.servlet.annotation.WebServlet;
 import jakarta.servlet.http.HttpServlet;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
+import org.jboss.as.test.shared.TimeoutUtil;
 
 /**
  * A servlet that sends an email using SMTP and verifies the message is in the recipient inbox using pop3.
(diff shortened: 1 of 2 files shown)


yersan (changes requested):
@Dkafetzis thanks for the changes. Maybe four seconds could be enough but it is a static value that does not depend on the one who is triggering the testsuite.

Instead of a static number of tries, I was expecting here the usage of https://github.com/wildfly/wildfly-core/blob/035d41260346840f0073a90875997f2b8f16df17/testsuite/shared/src/main/java/org/jboss/as/test/shared/TimeoutUtil.java#L35 to create a timeout that can be checked in the loop.

That's the WildFly standard approach for this. Using TimeoutUtil#adjust allows any testsuite caller to configure a factor `ts.timeout.factor` that modifies the final computed value for the time out. So, slow systems can increment the timeout. 

That makes the test more resilent, four seconds looks enough, but better if you make it configurable by using TimeoutUtil#adjust.

yersan (changes requested):
Hi @Dkafetzis , this is how we usually approach this: https://github.com/wildfly/wildfly-core/blob/ad531bdb218d15084267bed3c88269e9a1b78a4e/testsuite/manualmode/src/test/java/org/jboss/as/test/manualmode/auditlog/AuditLogBootingSyslogTestCase.java#L128

The idea is not to have a long pause (4s now at the beginning and 4s variable on each loop), the idea is to calculate a timeout using TimeoutUtil.java, 4s or even more, and add small pauses (500ms) on each loop before query again the server.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
```diff
@@ -47,27 +47,42 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                try {
+                    tryToReadEmails(emailContent);
+                    return;
+                } catch (MessagingException | IOException e){
+                    throw new ServletException(e);
+                } catch (ServletException e){
+                    if (e.getMessage().equals("Message read is not equals to the message sent")){
+                        throw e;
+                    }
+                }
```
yersan, line 62: Looks ok, but as a suggestion to make it more simpler:

1. tryToReadEmails could just return true/false. With that method, we are now interested only into know whether the emails were read as we expected. I mean, if `messages.length != 0 and the message text is the expected one` return true, otherwise false.
2. Any MessagingException or IOException can be ignored by the caller, and only go out of the loop if the method returned true.

it is up to you @Dkafetzis , I am fine as it is if the test case passes, but simplifying it a bit more would be nice.

Dkafetzis: @yersan I agree that we can ignore the MessagingException and the IOException, but I'd like to keep some way to know if the received message is different from what is expected and not just keep retrying till timeout in this case, since this indicates a different problem and I think it would be helpful to know if it fails because of this and not just due to the mail server not being up and running yet.

yersan (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
yersan: Thanks @Dkafetzis , I'm more in favor of simplicity, but that's ok for me for this test case.

Dkafetzis: @yersan Great, I applied the changes as requested then. Should be ready now

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
```diff
@@ -66,8 +85,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
             if (!emailContent.equals(messages[0].getContent().toString().trim())) {
                 throw new ServletException("Message read is not equals to the message sent");
             }
-        } catch (MessagingException | InterruptedException e) {
-            throw new ServletException(e);
-        }
+
+        } catch (MessagingException | IOException ignored) {}
```
yersan, line 89: @Dkafetzis you should not ignore them in the method, you have to ignore them outside, otherwise, as soon as any of these exceptions are ignored here, the test will pass.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
```diff
@@ -47,9 +47,28 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
 
-            // Wait a bit to ensure the server has processed the message
-            TimeUnit.SECONDS.sleep(1);
+            long endTime = System.currentTimeMillis() + TimeoutUtil.adjust(5000);
+            boolean success = false;
+            do {
+                Thread.sleep(500);
+                try {
+                     tryToReadEmails(emailContent);
+                    return;
```
yersan, line 56: a minor indentation issue

yersan (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
Dkafetzis: @yersan Done

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/mail/basic/MailTesterServlet.java
Dkafetzis: @yersan True, I've changed it

yersan (changes requested):
@Dkafetzis I forgot an additional detail, read permission over `ts.timeout.factor` due to the `TimeoutUtil` usage. See the failed Job:

`Permission check failed (permission "("java.util.PropertyPermission" "ts.timeout.factor" "read")`

The deployment explicitly needs to be granted to use the `ts.timeout.factor` property. You can find in the code other deployments adding this required permission, we need to do the same here.

Dkafetzis:
@yersan Thanks for the detail, it is done now

yersan (approved):

bstansberry:
Thanks @Dkafetzis and @yersan

PR #17799, opened 2024-04-07
[WFLY-19214] Use the wildfly-maven-plugin for build, ee-build and pre…

bstansberry:
…view/build

https://issues.redhat.com/browse/WFLY-19214

Builds on #17805

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> Wildfly issue links:
> * [WFLY-19222](https://issues.redhat.com/browse/WFLY-19222)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #17838, opened 2024-04-22
[WFLY-19267] Support running the standard WildFly testsuite against installations provisioned with the 'latest standard channels'

bstansberry:
https://issues.redhat.com/browse/WFLY-19267

This builds on #17799, which deals with a similar use case. The WFLY-19214 approach there is more flexible but more verbose in cases where the latest version of standard channels manifests are wanted.

____

> Wildfly issue links:
> * [WFLY-19214](https://issues.redhat.com/browse/WFLY-19214)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1503,6 +1503,41 @@
         </profile>
 
         <profile>
+            <!-- Use the latest version of the standard channels when provisioning. -->
+            <id>latest.standard.channels.profile</id>
+            <activation><property><name>latest.standard.channels</name></property></activation>
+            <properties>
+                <!-- Allow online provisioning so wildfly-maven-plugin can resolve the latest manifests -->
+                <galleon.offline>false</galleon.offline>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.wildfly.plugins</groupId>
+                        <artifactId>wildfly-maven-plugin</artifactId>
+                        <version>${version.org.wildfly.plugin}</version>
+                        <configuration>
+                            <channels>
+                                <channel>
+                                    <manifest>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1502,6 +1502,70 @@
...
+                                    </manifest>
+                                </channel>
+                                <channel>
+                                    <manifest>
+                                        <groupId>${channels.maven.groupId}</groupId>
+                                        <artifactId>wildfly</artifactId>
+                                    </manifest>
+                                </channel>
+                            </channels>
+                        </configuration>
```
yersan, line 1534: There is no wildfly-preview channel attached here, so I'm not sure how the expected channels would be injected when this plugin configuration is merged with the one on preview/build.

Would it become an issue?

I guess the same is applied to the ee-build. The plugin configuration merged will attach wildfly-ee and wildfly channels. Is that expected for an ee-build?

bstansberry: In both cases, yes the limitations you point out are expected. This profile as is will not work for testing WFP or for testing wildfly-ee by itself.

https://issues.redhat.com/browse/WFLY-19268 will replace this, but it is out of scope for this task, which is really just about having an upstream variant of an **urgent** downstream change that tests a combination of these two FPs. Neither wildfly-ee-only testing nor wildfly-preview are relevant to the downstream change.

https://issues.redhat.com/browse/WFLY-19268 will take advantage of the greater flexibility we have in wildfly-maven-plugin 5, which is not used downstream.

yersan (dismissed):

yersan (approved):

bstansberry:
Thanks @yersan. I'll be following up with another PR related to this that deal with bootable jar testing.

PR #17844, opened 2024-04-23
[WFLY-19274] When SocketsAndInterfacesTestCase searches for a non-def…

bstansberry:
…ault NIC it should ignore those with no addresses

https://issues.redhat.com/browse/WFLY-19274

https://ci.wildfly.org/viewLog.html?buildId=440377& is a custom test run to see if this fixes the recurring failure we're seeing in that nightly job.

____


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/management/api/SocketsAndInterfacesTestCase.java
```diff
@@ -81,6 +81,7 @@ public void before() throws IOException {
             testNic = getNic(testHost);
         } else {
             testNic = getNonDefaultNic();
+            org.junit.Assume.assumeFalse("No usable non-default nic is available", testNic == null);
```
bstansberry, line 84: Not that prior to this change testNic being null was also possible and would have produced an NPE. Either way, if a suitable nic cannot be found this test should be skipped.

yersan (approved):
LGTM, although the Galleon error still needs to be fixed and the Job should be completed for this PR.

bstansberry:
Thanks @yersan

PR #17847, opened 2024-04-25
WFLY-19282 Update HostExcludesTestCase configuration to work with WF33

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-19282

____

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
@@ -72,7 +72,7 @@ public class HostExcludesTestCase extends BuildConfigurationTestBase {
     private final boolean isFullDistribution = AssumeTestGroupUtil.isFullDistribution();
     private final boolean isPreviewGalleonPack = AssumeTestGroupUtil.isWildFlyPreview();
 
-    private static final String MAJOR = "32.";
+    private static final String MAJOR = "33.";
 
     /**
      * Maintains the list of expected extensions for each host-exclude name for previous releases.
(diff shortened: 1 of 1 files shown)


rhusar:
@yersan is this what needs to happen after every release?

yersan (approved):
Hello @rhusar , thanks for this. 

> is this what needs to happen after every release when we have added new extensions

Yes, for added or removed extensions. It is a pain, but it is the way how we do not forget to have the Enums updated with the list of extensions added or removed during the version recently released. Probably needs another thinking, this restriction is a pain and another task to do for the one who is releasing :(

> or if we want to add new ones and we have not moved it on to the next exclusion version?

If you want to add / remove more extensions and you have not moved it I think there is no restrictions there. But notice how it is expected to work: firstly the test discover whether there are extensions added/removed by checking the available extensions SPI implementations. The test will fail if it finds a new extension that is not tracked under the CURRENT Enum. If the test fails you are forced to add/removed the extension to/from the CURRENT enum, which is the enum that represents the current version under development. So, you usually add/remove extensions when you add or removed from the server certain extension. That will part of the extension PR development.

To sum us, the test forces us to track the extensions for the current release under development and forces us to add a new Enum with the added/removed extensions of the latest release if we have moved to a new release. Add to it all the details about the different distribution flavours ee, standard, preview. We only track the enums for ee and standard.

yersan:
Thanks @rhusar

rhusar:
@yersan, updating the HostExcludesTestCase seems to me to be a task worth mentioning in the release procedure document that @jmesnil @bstansberry use. Can you please consider adding it? Basically, almost after every release all nightly jobs that we run and all PRs that we test fail and generate lot of noise.

yersan:
> Can you please consider adding it? 

Thanks for the suggestion @rhusar. This is already tracked in the release procedure.

rhusar:
> > Can you please consider adding it?
> 
> Thanks for the suggestion @rhusar. This is already tracked in the release procedure.

Right - I saw Brian mention this on Zulip of Friday that we just missed it. All good 👍🏼

PR #17848, opened 2024-04-26
[WFLY-19283] Migrate Micrometer extension to new wildfly-subsystem approach

jasondlee:
https://issues.redhat.com/browse/WFLY-19283

Migrate classes
Modify module deps

____


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -32,8 +33,10 @@ public VersionedNamespace<IntVersion, MicrometerSubsystemSchema> getNamespace()
 
     @Override
     public PersistentResourceXMLDescription getXMLDescription() {
-        return builder(org.wildfly.extension.micrometer.MicrometerExtension.SUBSYSTEM_PATH, this.namespace)
-                .addAttributes(MicrometerSubsystemDefinition.ATTRIBUTES)
-                .build();
+        PersistentResourceXMLDescription.PersistentResourceXMLBuilder builder =
+                builder(MicrometerSubsystemRegistrar.PATH, this.namespace);
```
pferraro, line 37: As long as you're touching this now, perhaps update so not to rely on deprecated API?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,189 @@
...
+        context.addStep(new AbstractDeploymentChainStep() {
+            @Override
+            public void execute(DeploymentProcessorTarget processorTarget) {
+                processorTarget.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, DEPENDENCIES,
+                        DEPENDENCIES_MICROMETER, new MicrometerDependencyProcessor());
+                processorTarget.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, POST_MODULE,
+                        POST_MODULE_MICROMETER,
+                        new MicrometerDeploymentProcessor(exposeAnySubsystem, exposedSubsystems, wildFlyRegistry));
+            }
+        }, RUNTIME);
```
pferraro, line 158: Move this to your ResourceDescriptor, e.g.
```java
.withDeploymentChainContributor(new Consumer<>() {
    @Override
    public void accept(DeploymentProcessorTarget target) {
        //...
    }
}
```

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,189 @@
...
+        ServiceDependency<Executor> executor = ServiceDependency.on(MANAGEMENT_EXECUTOR, Executor.class);
+        ServiceDependency<ProcessStateNotifier> processStateNotifier =
+                ServiceDependency.on(PROCESS_STATE_NOTIFIER, ProcessStateNotifier.class);
+
+        Supplier<MicrometerCollector> collectorSupplier = () ->
+                new MicrometerCollector(mccf.get().createClient(executor.get()), processStateNotifier.get(),
+                        wildFlyRegistry);
+
+        context.addStep((operationContext, modelNode) -> {
+            MicrometerCollector micrometerCollector = collectorSupplier.get();
```
pferraro, line 171: This doesn't look right.  This will cause a second collector to be created.
Rather, you want to capture the value created by the service.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -24,7 +24,8 @@ public static Iterable<MicrometerSubsystemSchema> parameters() {
     }
 
     public SubsystemParsingTestCase(MicrometerSubsystemSchema schema) {
-        super(MicrometerExtension.SUBSYSTEM_NAME, new MicrometerExtension(), schema, MicrometerSubsystemSchema.CURRENT);
+        super(MicrometerConfigurationConstants.NAME,
+                new MicrometerExtension(), schema, MicrometerSubsystemSchema.CURRENT);
```
pferraro, line 28: Wow, aggressive line wrapping!  Are you using an 80x24 terminal? :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,189 @@
...
+    public static final ParentResourceDescriptionResolver RESOLVER =
+            new SubsystemResourceDescriptionResolver(MicrometerConfigurationConstants.NAME, MicrometerSubsystemRegistrar.class);
+
+    static final RuntimeCapability<Void> MICROMETER_COLLECTOR_RUNTIME_CAPABILITY =
+            RuntimeCapability.Builder.of(MICROMETER_MODULE + ".wildfly-collector", MicrometerCollector.class)
+                    .addRequirements(CLIENT_FACTORY_CAPABILITY, MANAGEMENT_EXECUTOR, PROCESS_STATE_NOTIFIER)
+                    .build();
+    static final RuntimeCapability<Void> MICROMETER_REGISTRY_RUNTIME_CAPABILITY =
+            RuntimeCapability.Builder.of(MICROMETER_MODULE + ".registry", WildFlyRegistry.class)
+                    .build();
```
pferraro, line 77: Is this used anywhere?

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
jasondlee: Definitely. I think I looked at that code and all it said was "deprecated" with no alternative offered. Point me to it?

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
jasondlee: Ha! No, it is overly aggressive. Just forgot to fix that.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: Good catch. No.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: I had
`
MicrometerCollector micrometerCollector = 
    (MicrometerCollector) operationContext.getServiceRegistry(false)
    .getRequiredService(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY
    .getCapabilityServiceName()).getValue();
`
but I got an NPE. Is it ok to cache in the Supplier and do a null-check there, or is there a Better Way?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: I had it that way, but I need model values. How/where would I cache those for use here?

jasondlee: I think I've found something works. Thoughts welcome. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: How's that?

pferraro: > MicrometerCollector micrometerCollector =  (MicrometerCollector) operationContext.getServiceRegistry(false) .getRequiredService(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY .getCapabilityServiceName()).getValue();

This is incorrect.  The installed service does not implement org.jboss.msc.Value.getValue() and thus will always return null.

pferraro: > Is it ok to cache in the Supplier and do a null-check there, or is there a Better Way?

```java
AtomicReference<MicrometerCollector> captor = new AtomicReference<>();

context.addStep((operationContext, modelNode) -> {
        MicrometerCollector collector = captor.get();
        // Given that this step runs in the VERIFY stage, and our collector service was started eagerly, the collector reference _should_ be non-null.
        if (collector == null) return;
        // ...

return CapabilityServiceInstaller.builder(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY, collectorSupplier)
        .requires(List.of(mccf, executor, processStateNotifier))
        .withCaptor(captor::set) // capture the provided value
        .asActive() // Start actively
        .build();
```

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
pferraro: There should be @deprecated javadoc tags:
https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/PersistentResourceXMLDescription.java#L541

jasondlee: Ah, ok. That wasn't popping up in IDEA, but I'll open that source directly. Thanks.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -44,7 +44,6 @@ public void deploy(DeploymentPhaseContext deploymentPhaseContext) throws Deploym
                 deploymentPhaseContext,
```
pferraro, line 44: This service does not look particularly complicated.  Why not create a ServiceInstaller from a Supplier<MetricRegistration>?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,189 @@
...
+
+        AtomicReference<MicrometerCollector> captor = new AtomicReference<>();
+
+        context.addStep((operationContext, modelNode) -> {
+            MicrometerCollector collector = captor.get();
+            // Given that this step runs in the VERIFY stage, and our collector service was started eagerly, the
+            // collector reference _should_ be non-null.
+            if (collector != null) {
+                ImmutableManagementResourceRegistration rootResourceRegistration = context.getRootResourceRegistration();
+                Resource rootResource = context.readResourceFromRoot(EMPTY_ADDRESS);
```
pferraro, line 175: Why perform a fully recursive read of the whole model if we might only need to read specific subsysems?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,189 @@
...
+        exposedSubsystems = MicrometerSubsystemRegistrar.EXPOSED_SUBSYSTEMS.unwrap(context, model);
+        exposeAnySubsystem = exposedSubsystems.remove("*");
+        endpoint = MicrometerSubsystemRegistrar.ENDPOINT.resolveModelAttribute(context, model).asStringOrNull();
+        step = MicrometerSubsystemRegistrar.STEP.resolveModelAttribute(context, model).asLong();
+        wildFlyRegistry = endpoint != null ?
+                new WildFlyOtlpRegistry(new WildFlyMicrometerConfig(endpoint, step)) :
+                new NoOpRegistry();
+
+        try {
+            new JmxMicrometerCollector(wildFlyRegistry).init();
```
pferraro, line 155: Doesn't this imply a requirement on "org.wildfly.management.jmx"?

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
jasondlee: Something like 
```
ServiceInstaller.builder((Supplier<MicrometerDeploymentService>) () -> {
    return ...; 
});
```?

File: observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java [hunk 9]
```diff
@@ -0,0 +1,189 @@
...
+
+        return registration;
+    }
+
+    @Override
+    public ResourceServiceInstaller configure(OperationContext context, ModelNode model) throws OperationFailedException {
+        exposedSubsystems = MicrometerSubsystemRegistrar.EXPOSED_SUBSYSTEMS.unwrap(context, model);
+        exposeAnySubsystem = exposedSubsystems.remove("*");
+        endpoint = MicrometerSubsystemRegistrar.ENDPOINT.resolveModelAttribute(context, model).asStringOrNull();
+        step = MicrometerSubsystemRegistrar.STEP.resolveModelAttribute(context, model).asLong();
```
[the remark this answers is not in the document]
jasondlee: Fixed.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: I do not know. :) That code has been that way (coming from the metrics subsystem) from long before my time. I'm definitely open to improving, but I can't explain/defend it. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: Same as above. I'm not sure. It's been that way for years and works, but I can't attest to its correctness.

bstansberry:
FYI, I wrote the :read-resource-description(operations,recursive) output for this subsystem to a file after building main and then this branch. Then I ran diff -uw against the files. Got this:

```
-- /Users/bstansberry/tmp/orig_rrd.txt	2024-05-01 18:19:40
+++ refactored_rrd.txt	2024-05-02 07:09:25
@@ -3,6 +3,11 @@
     "result" => {
         "description" => "Micrometer subsystem",
         "stability" => "default",
+        "capabilities" => [{
+            "name" => "org.wildfly.extension.micrometer.wildfly-collector",
+            "dynamic" => false,
+            "stability" => "default"
+        }],
         "attributes" => {
             "endpoint" => {
                 "type" => STRING,
@@ -88,6 +93,7 @@
                 },
                 "reply-properties" => {},
                 "read-only" => false,
+                "restart-required" => "all-services",
                 "runtime-only" => false
             },
             "read-children-types" => {
@@ -386,7 +392,7 @@
                 "reply-properties" => {},
                 "stability" => "default",
                 "read-only" => false,
-                "restart-required" => "all-services",
+                "restart-required" => "resource-services",
                 "runtime-only" => false
             },
             "map-remove" => {
```

The 'restart-required` bits relate to the 'add` and 'remove` ops, respectively.

I haven't dug into the details but my instinct is the first two diffs are fine; the refactor is correcting mistakes. The 'remove' setting change seems wrong.

jasondlee:
> ```
> 
> ```
> 
> 
>     
>   
> 
> The 'restart-required` bits relate to the 'add` and 'remove` ops, respectively.
> 
> I haven't dug into the details but my instinct is the first two diffs are fine; the refactor is correcting mistakes. The 'remove' setting change seems wrong.

Good catch. My guess is default values at play with the new API? At any rate, I've added this:

```
    .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
    .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
```

Does that look right @pferraro ?

pferraro:
> > ```
> > 
> > ```
> > The 'restart-required` bits relate to the 'add` and 'remove` ops, respectively.
> > I haven't dug into the details but my instinct is the first two diffs are fine; the refactor is correcting mistakes. The 'remove' setting change seems wrong.
> 
> Good catch. My guess is default values at play with the new API? At any rate, I've added this:
> 
> ```
>     .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
>     .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
> ```
> 
> Does that look right @pferraro ?

The behavior of the generic add/remove resource operation handlers is driven by these flags - so it is important to set them appropriately (if your intended behavior is meant to differ from the default behavior).  The ResourceDescriptor.Builder will automatically override these based on specific described behavior - however, I just realized that the remove operation restart level when calling withDeploymentChainContributor(...) is not overriding the value set when calling withRuntimeHandler(...).  For now, set these explicitly.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,213 @@
...
+import org.wildfly.subsystem.service.ServiceDependency;
+import org.wildfly.subsystem.service.capability.CapabilityServiceInstaller;
+
+class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistrar, ResourceServiceConfigurator {
+    private static final String MICROMETER_MODULE = "org.wildfly.extension.micrometer";
+    private static final String MICROMETER_API_MODULE = "org.wildfly.micrometer.deployment";
+
+    static final String CLIENT_FACTORY_CAPABILITY = "org.wildfly.management.model-controller-client-factory";
+    static final String MANAGEMENT_EXECUTOR = "org.wildfly.management.executor";
+    static final String PROCESS_STATE_NOTIFIER = "org.wildfly.management.process-state-notifier";
```
pferraro, line 68: N.B. Once https://issues.redhat.com/browse/WFCORE-6818 is complete (this week maybe?) you can drop these.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,213 @@
...
+                .addCapability(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY)
+                .addAttributes(List.of(ENDPOINT, STEP, EXPOSED_SUBSYSTEMS))
+                .withDeploymentChainContributor(target -> {
+                    target.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, DEPENDENCIES, DEPENDENCIES_MICROMETER,
+                            new MicrometerDependencyProcessor());
+                    target.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, POST_MODULE, POST_MODULE_MICROMETER,
+                            new MicrometerDeploymentProcessor(deploymentConfig.get()));
+                })
+                .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
+                .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
```
pferraro, line 136: N.B. This will become redundant following https://github.com/wildfly/wildfly-core/pull/5989

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -21,32 +18,29 @@
...
-                exposeAnySubsystem,
-                exposedSubsystems);
+        ServiceInstaller.builder(() -> new MicrometerDeploymentService(
+                        deploymentUnit.getAttachment(DeploymentModelUtils.DEPLOYMENT_RESOURCE),
+                        deploymentUnit.getAttachment(DeploymentModelUtils.MUTABLE_REGISTRATION_ATTACHMENT),
+                        createDeploymentAddressPrefix(deploymentUnit),
+                        config.getCollectorSupplier(),
+                        config.getSubsystemFilter()
+                )).asActive()
+                .build();
```
pferraro, line 43: I think you are missing `.install(deploymentPhaseContext)`? :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -21,32 +18,29 @@
...
         final DeploymentUnit deploymentUnit = deploymentPhaseContext.getDeploymentUnit();
 
-        MicrometerDeploymentService.install(deploymentPhaseContext.getServiceTarget(),
-                deploymentPhaseContext,
-                deploymentUnit.getAttachment(DeploymentModelUtils.DEPLOYMENT_RESOURCE),
-                deploymentUnit.getAttachment(DeploymentModelUtils.MUTABLE_REGISTRATION_ATTACHMENT),
-                registrySupplier,
-                exposeAnySubsystem,
-                exposedSubsystems);
+        ServiceInstaller.builder(() -> new MicrometerDeploymentService(
```
pferraro, line 36: This doesn't look right.  The various ServiceInstaller.Builder factory methods generate generic Service implementations.  This particular method expects a Supplier of the value to be produced by the Service.
Presumably, this service produces a `MetricRegistration`, right?
So, perhaps you want something like this:

```java
Resource resource = deploymentUnit.getAttachment(DeploymentModelUtils.DEPLOYMENT_RESOURCE);
ManagementResourceRegistration registration = deploymentUnit.getAttachment(DeploymentModelUtils.MUTABLE_REGISTRATION_ATTACHMENT);
PathAddress address = ...;
Supplier<MetricRegistration> factory = () -> this.config.getCollectorSupplier().get().collectResourceMetrics(resource, registration, deploymentAddress::append, this.config.getSubsystemFilter());
ServiceInstaller.builder(factory)
    .onStop(MetricRegistration::unregister)
    .build()
    .install(deploymentPhaseContext);
```

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,213 @@
...
+
+            @Override
+            public Predicate<String> getSubsystemFilter() {
+                return subsystem -> exposeAnySubsystem || exposedSubsystems.contains(subsystem);
+            }
+
+            @Override
+            public Supplier<MicrometerCollector> getCollectorSupplier() {
+                return collectorSupplier;
+            }
```
pferraro, line 182: This will result in a 2nd Micrometer collector instance being created by your DUP.  I doubt that was your intention.
Either change this method to return `captor::get`; or, better yet, change this method signature to `MicrometerCollector getCollector()` and have this method return `captor.get()`

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
pferraro: On 2nd read, it seems to me you can remove this method all together.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
jasondlee: Probably. :P Fixed.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -0,0 +1,207 @@
...
+
+    static final PathElement PATH = SubsystemResourceDefinitionRegistrar.pathElement(MicrometerConfigurationConstants.NAME);
+    public static final ParentResourceDescriptionResolver RESOLVER =
+            new SubsystemResourceDescriptionResolver(MicrometerConfigurationConstants.NAME, MicrometerSubsystemRegistrar.class);
+
+    static final RuntimeCapability<Void> MICROMETER_COLLECTOR_RUNTIME_CAPABILITY =
+            RuntimeCapability.Builder.of(MICROMETER_MODULE + ".wildfly-collector", MicrometerCollector.class)
+                    .addRequirements(CLIENT_FACTORY_CAPABILITY, MANAGEMENT_EXECUTOR, PROCESS_STATE_NOTIFIER)
+                    .build();
+    static final ServiceName MICROMETER_COLLECTOR = MICROMETER_COLLECTOR_RUNTIME_CAPABILITY.getCapabilityServiceName();
```
pferraro, line 78: This does not appear to be referenced anywhere.

pferraro (approved):
Other than one potentially orphaned static reference, this looks good.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: Removed.

bstansberry (approved):
The read-resource-description diff looks good.

PR #17856, opened 2024-04-29
[WFLY-19298] Update fasterxml to 2.17.0

spyrkob:
Issue: https://issues.redhat.com/browse/WFLY-19298

____

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,7 +385,7 @@
         <version.antlr>4.13.0</version.antlr>
         <version.com.carrotsearch.hppc>0.8.1</version.com.carrotsearch.hppc>
         <version.com.fasterxml.classmate>1.5.1</version.com.fasterxml.classmate>
-        <version.com.fasterxml.jackson>2.15.4</version.com.fasterxml.jackson>
+        <version.com.fasterxml.jackson>2.17.0</version.com.fasterxml.jackson>
         <version.com.fasterxml.jackson.databind>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.databind>
         <version.com.fasterxml.jackson.jr.jackson-jr-objects>${version.com.fasterxml.jackson}</version.com.fasterxml.jackson.jr.jackson-jr-objects>
         <version.com.github.ben-manes.caffeine>3.1.8</version.com.github.ben-manes.caffeine>


bstansberry:
This requires approval from the various leads responsible for FasterXML use:

@jamezp (RESTEasy)
@rhusar (com.microsoft.azure.storage module)
@yrodiere (Hibernate Search stuff)
@scottmarlow (Hibernate ORM)
@fjuma (org.wildfly.security.elytron-http-oidc and org.wildfly.security.elytron-jose-*)
@jasondlee (OpenTelemetry)
@pferraro (OpenAPI)
@kabir (VertX and Kafka)

scottmarlow:
Hibernate ORM 6.4.4.Final is currently using 2.14.1 as is the ORM `main` branch.

kabir (approved):

fjuma (approved):
Thanks @spyrkob!

yrodiere:
> Hibernate ORM 6.4.4.Final is currently using 2.14.1 as is the ORM main branch.

WildFly should probably upgrade to ORM 6.5 though, and that's not the `main` branch. Did you have a look at ORM 6.5?

> Let's see what @yrodiere has to say about Search but I think 7.1.0.Final is on 2.15.2 and main looks to be on 2.17.0.

I'll defer to @marko-bekhta for Search.

> I think the underlying question is whether there could be Search/ORM releases on fasterxml `2.17.0` if that could work for some future versions of all components?

We could upgrade in ORM 6.6, sure, though obviously that'll take a few weeks/months.

We need to have a look, though. I suspect ORM is not actually that deeply coupled with Jackson, at least if it's only about JSON columns, but I might be missing something. If coupling is indeed loose, it might be enough to just run the ORM test suite with the dependency upgraded.

@sebersole or @beikov may have an answer ready for you, otherwise I'll have a closer look, but not before next week unfortunately.

beikov:
>I suspect ORM is not actually that deeply coupled with Jackson, at least if it's only about JSON columns, but I might be missing something. If coupling is indeed loose, it might be enough to just run the ORM test suite with the dependency upgraded.

You're right, it's an "optional" dependency. Upgrading Jackson should not be a problem, so +1 from my side on upgrading in Wildfly.

scottmarlow:
> > I suspect ORM is not actually that deeply coupled with Jackson, at least if it's only about JSON columns, but I might be missing something. If coupling is indeed loose, it might be enough to just run the ORM test suite with the dependency upgraded.
> 
> You're right, it's an "optional" dependency. Upgrading Jackson should not be a problem, so +1 from my side on upgrading in Wildfly.

Optional dependencies or not, I think the Hibernate ORM testsuite needs to pass with the fasterxml upgrade before WildFly upgrades to it to ensure that the ORM integration fasterxml works as expected.

rhusar (approved):
Fine for `com.microsoft.azure.storage` module. 👍🏼

marko-bekhta:
For Hibernate Search:
- 7.2 (currently in development) Jackson was upgraded to 2.17 a while ago in https://hibernate.atlassian.net/browse/HSEARCH-5106 
- 7.1 build passes with 2.17.0: https://github.com/hibernate/hibernate-search/pull/4141

marko-bekhta (approved):

yrodiere:
Thanks @beikov @scottmarlow .

I created:

* https://hibernate.atlassian.net/browse/HHH-18050 to upgrade to Jackson 2.17.0 in Hibernate ORM 6.6 (+ corresponding PR)
* https://github.com/hibernate/hibernate-orm/pull/8284 to test Hibernate ORM 6.5 against Jackson 2.17.0 (we won't merge this PR, it's just for testing purposes)

@scottmarlow Will you please make the upgrade to Hibernate ORM 6.5 in WildFly?

beikov:
It's a good first step for Wildfly to try out ORM 6.5, but I would like to ask you to make sure that version doesn't end up in a release, because ORM 6.6 is around the corner and it would be best for Wildfly to use that ORM version in its next release.

yrodiere:
> It's a good first step for Wildfly to try out ORM 6.5, but I would like to ask you to make sure that version doesn't end up in a release, because ORM 6.6 is around the corner and it would be best for Wildfly to use that ORM version in its next release.

Clarifying: *ideally* WildFly's next version would use Hibernate ORM 6.6 if it's ready by the feature freeze date, but failing that shipping Hibernate ORM 6.5 would definitely be better than the current Hibernate ORM 6.4. So yes, please upgrade @scottmarlow :)

We're going on a tangent here though, let's move to Zulip if more discussion is necessary.

Back to discussing Jackson here :)

yrodiere (approved):
Hibernate ORM 6.5 works fine with Jackson 2.17: https://github.com/hibernate/hibernate-orm/pull/8284
I'll leave it to @scottmarlow to upgrade to Hibernate ORM 6.5 in WildFly, as WildFly is  currently using Hibernate ORM 6.4 (but shouldn't).

scottmarlow:
> Hibernate ORM 6.5 works fine with Jackson 2.17: [hibernate/hibernate-orm#8284](https://github.com/hibernate/hibernate-orm/pull/8284) I'll leave it to @scottmarlow to upgrade to Hibernate ORM 6.5 in WildFly, as WildFly is currently using Hibernate ORM 6.4 (but shouldn't).

Created https://issues.redhat.com/browse/WFLY-19306 for trying to upgrade to ORM 6.5 in WildFly (note that WildFly Preview is already including ORM 6.5).

jamezp:
There is an issue with RESTEasy with this upgrade. Previously `java.util.Optional` would serialize to something like:
```json
{"empty":false,"present":true}
```
This happened when the `com.fasterxml.jackson.datatype.jdk8.Jdk8Module` was not registered with the `ObjectMapper`. In 2.17 this now throws an exception instead because the `Jdk8Module` is not installed by default. It could be related to https://github.com/FasterXML/jackson-modules-java8/issues/294.

One could argue that RESTEasy should register some default modules in a default `ObjectMapper`. However, there could be users relying on this old behavior. IMO the old behavior is wrong, but I'll need to make changes in RESTEasy to fix the support for this. If we feel this is needed please let me know so we can file an issue and fix this.

jamezp:
I've filed [RESTEASY-3502](https://issues.redhat.com/browse/RESTEASY-3502) for the RESTEasy issue. I can disable the tests that fail in RESTEasy if a newer version of Jackson is found. However, this could be considered a behavior change and if we proceed with the upgrade we should document that.

darranl:
@jamezp Could you please provide your update in a review, this shows as approved in the GitHub UI but I think with your comment it should either be "Changed Requested" or even have a "Hold" label to indicate we are not ready.

jasondlee (approved):

scottmarlow:
https://issues.redhat.com/browse/WFLY-19306 is now for upgrading WildFly to the Hibernate ORM 6.6.0.Final release

bstansberry:
@jamezp @darranl It would be good if we could have a decision on this by Sep 23 so we can get it into 34 Beta.

scottmarlow (approved):

jamezp:
I know there are 2 tests in the RESTEasy test suite that will fail with the upgrade. However, I don't think that should block this. We might need to document which modules need to be enabled. I can't remember off the top of my head which ones.

kabir (approved):

jamezp (approved):

bstansberry:
Thanks @spyrkob

PR #17857, opened 2024-04-29
[WFLY-18582] Add prometheus endpoint to micrometer extension

### LINKED ISSUE #590: Add prometheus endpoint to micrometer extension

jasondlee:
https://issues.redhat.com/browse/WFLY-18582

Part 1
- Introduce new model/schema version
- Move OTLP-related attributes to a child resource
- Update test to reflect model change
 
Part 2
- Move OTLP registry configuration to a child resource
- Add prometheus support
- Add tests for Prometheus context

Resolves https://github.com/wildfly/wildfly-proposals/issues/590


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/otlp/OtlpRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,102 @@
+package org.wildfly.extension.micrometer.otlp;
```
bstansberry, line 1: header

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyCompositeRegistry.java
```diff
@@ -0,0 +1,95 @@
+package org.wildfly.extension.micrometer.registry;
```
bstansberry, line 1: header

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/MicrometerTransformerTestCase.java
```diff
@@ -0,0 +1,105 @@
+package org.wildfly.extension.micrometer;
```
bstansberry, line 1: header

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -33,9 +35,25 @@ public VersionedNamespace<IntVersion, MicrometerSubsystemSchema> getNamespace()
...
+            builder.addChild(factory.builder(OtlpRegistryDefinitionRegistrar.PATH)
+                    .addAttributes(OtlpRegistryDefinitionRegistrar.ATTRIBUTES.stream())
+                    .setXmlElementName("otlp-registry")
+                    .build());
+        } else {
+            builder.addChild(factory.builder(OtlpRegistryDefinitionRegistrar.PATH)
+                    .addAttributes(MicrometerSubsystemRegistrar.ENDPOINT,
+                            MicrometerSubsystemRegistrar.STEP)
+                    .setXmlElementName("otlp-registry")
+                    .build()
```
bstansberry, line 53: AIUI the if and else clauses do the same thing. Should the 'addChild' statement be removed from the 'else'?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/otlp/OtlpRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,102 @@
...
+
+        ManagementResourceRegistrar.of(descriptor).register(mrr);
+
+        return mrr;
+    }
+
+    @Override
+    public ResourceServiceInstaller configure(OperationContext context, ModelNode model) throws OperationFailedException {
+        String endpoint = OtlpRegistryDefinitionRegistrar.ENDPOINT.resolveModelAttribute(context, model).asStringOrNull();
+        Long step = OtlpRegistryDefinitionRegistrar.STEP.resolveModelAttribute(context, model).asLong();
```
bstansberry, line 85: Nit: This statement won't result in null being assigned and in the end 'step' gets passed to Duration.ofSeconds(long), which would NPE if it was null. So using primitive long here and in WildFlyMicrometerConfig is clearer.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyCompositeRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+//            case GAUGE:
+//                return addGauge(metric, metadata);
+//            case COUNTER:
+//                return addCounter(metric, metadata);
+//            default:
+//                throw MicrometerExtensionLogger.MICROMETER_LOGGER.unsupportedMetricType(metadata.getType().name());
+//        }
+//    }
+
+    private Meter.Id addCounter(WildFlyMetric metric, MetricMetadata metadata) {
```
bstansberry, line 41: This is unused.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyCompositeRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+                .description(metadata.getDescription());
+        Optional<FunctionCounter> first = this.getRegistries().stream().map(builder::register).findFirst();
+        if (first.isPresent()) {
+            return first.get().getId();
+        } else {
+            throw MicrometerExtensionLogger.MICROMETER_LOGGER.errorRegisteringMetric(metadata.getMetricName());
+        }
+    }
+
+    private Meter.Id addGauge(WildFlyMetric metric, MetricMetadata metadata) {
```
bstansberry, line 55: This is unused.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/MicrometerTransformerTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Objects;
+
+import static org.wildfly.extension.micrometer.MicrometerSubsystemModel.VERSION_1_1_0;
```
bstansberry, line 20: This goes first.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
```diff
@@ -29,6 +30,8 @@ public class MicrometerSetupTask extends AbstractSetupTask {
 
     @Override
     public void setup(final ManagementClient managementClient, String containerId) throws Exception {
+        otelCollector = OpenTelemetryCollectorContainer.getInstance();
```
bstansberry, line 33: Why is the 'dockerAvailable' handling being dropped? Is this because WFARQ properly handles failures from OpenTelemetryCollectorContainer.getInstance() now?

If so I believe 'dockerAvailable" can be removed.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
jasondlee: Ah, that's a good catch. The issue here is that the model is changing, but the XML structure is not, and it looks like a little noise was left here in trying to figure out how to handle that correctly. The two branches are slightly different in the the attribute defs come from different places, though the defs themselves are identical, so that is probably a "distinction without a difference". :)

jasondlee: Ah, well. Not QUITE identical. The definition on `MicrometerSubsystemRegistrar` includes `.addFlag(AttributeAccess.Flag.ALIAS)` which I added, iirc, due to some runtime errors at one point related to attribute translation. That might also be noise. @pferraro ?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/otlp/OtlpRegistryDefinitionRegistrar.java
jasondlee: Updated.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyCompositeRegistry.java
jasondlee: Most of that class is no longer used. The class itself may be superfluous. I'll verify then do the needful.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
pferraro: The set of attributes of the root resource should be different between version 1.0 and 2.0, correct?

jasondlee: Probably. I'll dig into that.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public ResourceServiceInstaller configure(OperationContext context, ModelNode model) throws OperationFailedException {
+        String serviceContext = CONTEXT.resolveModelAttribute(context, model).asStringOrNull();
+        boolean securityEnabled = SECURITY_ENABLED.resolveModelAttribute(context, model).asBoolean();
+
+        return ServiceInstaller.builder(ServiceDependency.on(HTTP_EXTENSIBILITY_CAPABILITY, ExtensibleHttpManagement.class))
+                .onStart(ehm -> {
+                    WildFlyPrometheusRegistry prometheusRegistry = new WildFlyPrometheusRegistry();
+                    wildFlyRegistry.add(prometheusRegistry);
+                    ehm.addManagementHandler(serviceContext, securityEnabled,
```
pferraro, line 85: Should this service also remove the management handler on stop?  Perhaps it doesn't matter since the service is active?

jasondlee: Any adds/removes require a restart/reload. Does that answer your questions here?
```
                .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
                .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
```

jasondlee:
FYI: It looks like the Cloud Tests failed because they're using a WildFly image that does not contain the model changes in this build.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Micrometer.adoc
```diff
@@ -22,32 +22,71 @@ This `org.wildfly.extension.micrometer` extension is available to all the standa
...
 
-* `endpoint` - the URL of the metrics collector endpoint (default: http://localhost:4318/v1/metrics)
 * `exposed-subsystems` - a list of strings corresponding to the names of subsystems that exposes their metrics in the
 HTTP metrics endpoints. By default, it is not defined (there will be no metrics exposed by subsystem). The special wildcard "{asterisk}" can be used to expose metrics from _all_ subsystems. The standalone configuration sets this attribute to "{asterisk}".
+
+=== OTLP Registry Model
+
+The `/subsystem=micrometer/registry=otlp` resource defines two attributes:
+
+* `endpoint` - (*required*) the URL of the metrics collector endpoint (default: http://localhost:4318/v1/metrics)
```
kabir, line 80: I'm not sure it is correct to say it is required if there is a default value

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Micrometer.adoc
```diff
@@ -22,32 +22,71 @@ This `org.wildfly.extension.micrometer` extension is available to all the standa
...
 * `step` - the step size, or reporting frequency, to use (in seconds).
 
+=== Prometheus Registry Model
+
+The `/subsystem=micrometer/registry=prometheus` resource defines two attributes:
+
+* `context` - (*required*) The context on which to make metrics available for scraping. This will be exposed on the management interface (e.g, `http://localhost:9090`)
```
kabir, line 87: required with a default

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtensionLogger.java
```diff
@@ -65,4 +65,6 @@ public interface MicrometerExtensionLogger extends BasicLogger {
     @LogMessage(level = WARN)
     @Message(id = 11, value = "Micrometer has been enabled, but no endpoint has been configured. A No-op metrics registry has been configured.")
     void noOpRegistryChosen();
-}
+
+    @Message(id = 12, value = "There was an error registerting the metric '%s'")
```
kabir, line 69: s/registerting/registering

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/otlp/OtlpRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,117 @@
...
+                                wildFlyRegistry.add(new WildFlyOtlpRegistry(new WildFlyMicrometerConfig(endpoint, step)));
+                            }
+                        },
+                        () -> {
+                        }
+                )
+                .asActive()
+                .build();
+
+//        AtomicReference<WildFlyMicrometerConfig> captor = new AtomicReference<>();
```
kabir, line 102: Probably want to delete this

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/MicrometerTransformerTestCase.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import java.util.Objects;
+
+/*
+ * Currently ignore as I keep getting this error:
+ * [ERROR] testTransformation[0](org.wildfly.extension.micrometer.MicrometerTransformerTestCase)  Time elapsed: 0.945 s  <<< ERROR!
+ * java.lang.IllegalStateException: Url file:/home/jdlee/.m2/repository/org/wildfly/core/wildfly-subsystem-test-framework/23.0.0.Beta5/wildfly-subsystem-test-framework-23.0.0.Beta5.jar which is the code source for the following classes has already been set up via other means: [org.jboss.as.subsystem.test.AdditionalInitialization]
+ *
+ * I don't know why, but, for now, I'd like to get a full CI run to verify the other changes. Will come back to this
+ * "soon".
```
kabir, line 33: No high priority Jira linked, no merge :-)

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerPrometheusTestCase.java
```diff
@@ -0,0 +1,149 @@
...
+
+        setPrometheusSecurity(false);
+        makeRequests();
+        metrics = fetchPrometheusMetrics(false);
+        Assert.assertTrue("'demo_counter_total " + REQUEST_COUNT + "' is expected",
+                metrics.contains("demo_counter_total " + REQUEST_COUNT));
+    }
+
+    private void addPrometheusRegistry() throws Exception {
+        ModelNode addOperation = Operations.createAddOperation(prometheusRegistryAddress);
```
kabir, line 109: Rather than this method, and one of the tests calling it (since AFAIK test method order isn't guaranteed), I think I would instead override StabilityServerSetupSnapshotRestoreTasks.Preview and use that class as the setup task. Then I'd add override the `doSetup()` method and add  this code there

kabir (changes requested):
A few small issues and nitpicks

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Micrometer.adoc
jasondlee: There is no default for this one.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerPrometheusTestCase.java
jasondlee: Done

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/MicrometerTransformerTestCase.java
jasondlee: > No high priority Jira linked, no merge :-)

https://issues.redhat.com/browse/WFLY-19679

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public ResourceServiceInstaller configure(OperationContext context, ModelNode model) throws OperationFailedException {
+        String serviceContext = CONTEXT.resolveModelAttribute(context, model).asStringOrNull();
+        boolean securityEnabled = SECURITY_ENABLED.resolveModelAttribute(context, model).asBoolean();
+
+        if (context.getProcessType().isHostController()) {
+            throw MicrometerExtensionLogger.MICROMETER_LOGGER.prometheusNotSupportedOnHostControllers();
+        }
```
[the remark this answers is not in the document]
bstansberry: My guess is ResourceDescriptor.Builder.withOperationTransformation("add", transformation) will do the trick.

The 'transformation' param being a UnaryOperator that wraps the passed in OSH with code that first does the check ^^^.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -24,11 +23,13 @@ public static Iterable<MicrometerSubsystemSchema> parameters() {
...
+        super(MicrometerConfigurationConstants.NAME, new MicrometerExtension(), schema,
+            MicrometerSubsystemSchema.VERSION_1_1_COMMUNITY);
     }
 
     @Override
-    protected String getSubsystemXmlPathPattern() {
-        return "%s_%d_%d.xml";
+    public void testSubsystem() throws Exception {
+        System.out.printf("Testing " + getSubsystemSchema());
+        super.testSubsystem();
```
jasondlee, line 33: I'll clean this up after I see how CI does with these changes over all. I've not had good luck on CI with force pushes in rapid succession. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,61 @@
...
 import org.jboss.staxmapper.IntVersion;
+import org.wildfly.extension.micrometer.otlp.OtlpRegistryDefinitionRegistrar;
+import org.wildfly.extension.micrometer.prometheus.PrometheusRegistryDefinitionRegistrar;
 
 public enum MicrometerSubsystemSchema implements PersistentSubsystemSchema<MicrometerSubsystemSchema> {
-    VERSION_1_0(1, 0), // WildFly 28
-    VERSION_1_1(1, 1), // WildFly 29.0.0.Alpha1
+    VERSION_1_0(1, 0, Stability.DEFAULT), // WildFly 28
+    VERSION_1_1(1, 1, Stability.DEFAULT), // WildFly 29.0.0.Alpha1
+    VERSION_1_1_COMMUNITY(1, 1, Stability.COMMUNITY) // WildFly 35
```
jasondlee, line 27: @bstansberry @pferraro I'm not sure why we would want to bump the version to this instead of 2.0, but if you're ok with it, I'll trust you. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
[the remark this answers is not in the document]
jasondlee: @bstansberry How is this? https://github.com/wildfly/wildfly/pull/17857/files#diff-66585cf48a18907809788f26a06306d21bde06ed899c61933c3ea5862b506de7R65-R90

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,110 @@
...
+    public PrometheusRegistryDefinitionRegistrar(WildFlyCompositeRegistry wildFlyRegistry) {
+        this.wildFlyRegistry = wildFlyRegistry;
+    }
+
+    @Override
+    public ManagementResourceRegistration register(ManagementResourceRegistration parent, ManagementResourceRegistrationContext context) {
+        ResourceDescriptor descriptor = ResourceDescriptor.builder(MicrometerSubsystemRegistrar.RESOLVER.createChildResolver(PATH))
+                .addAttributes(ATTRIBUTES)
+                .withRuntimeHandler(ResourceOperationRuntimeHandler.configureService(this))
+            .withOperationTransformation("add", new UnaryOperator<OperationStepHandler>() {
```
bstansberry, line 69: whitespace

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,110 @@
...
+                @Override
+                public OperationStepHandler apply(OperationStepHandler handler) {
+                    return (context, operation) -> {
+                        ModelNode address = operation.get("address");
+                        String json = operation.toJSONString(false);
```
bstansberry, line 74: These two lines are cruft

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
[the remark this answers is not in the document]
bstansberry: @jasondlee I left a couple comments about side issues but that basically looks right to me.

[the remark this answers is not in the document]
jasondlee: Updates pushed.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,107 @@
...
+                        }
+                        handler.execute(context, operation);
+                    };
+                }
+            })
+                .build();
+        ManagementResourceRegistration mrr = parent.registerSubModel(
+                ResourceDefinition.builder(RESOURCE_REGISTRATION,
+                        descriptor.getResourceDescriptionResolver()).build());
+        ManagementResourceRegistrar.of(descriptor).register(mrr);
```
bstansberry, line 84: The io.prometheus module needs to be recorded as an additionalRuntimePackage, as it should only be an optional dep of any existing modules and Galleon needs to know to provision it if this resource is configured.

See https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/ResourceDefinition.2EregisterAdditionalRuntimePackages.20.2E.2E.2E/near/485906664 for how to do this.

observability/micrometer/src/main/resources/org/wildfly/extension/micrometer/LocalDescriptions.properties
```diff
@@ -10,3 +10,6 @@ micrometer.remove=Operation Removes subsystem
 micrometer.exposed-subsystems=The names of the WildFly subsystems that exposes their metrics (or '*' to expose any subsystem metrics).
 micrometer.endpoint=The URL to which metrics will be published.
 micrometer.step=The step size (reporting frequency) to use in seconds.
+micrometer.registry=Micrometer registries
+micrometer.registry.prometheus.context=The context root from which to serve Prometheus metrics (e.g., "/metrics" or "/prometheus"). Note: Prometheus is not supported on host controllers.
+micrometer.registry.prometheus.security-enabled=Whether or not the context should be secured
```
bstansberry, line 15: This file seems really incomplete. I don't see how it works; I suspect there's a bug being relied upon!

observability/micrometer/src/test/resources/org/wildfly/extension/micrometer/micrometer-community-1.1.xml
```diff
@@ -0,0 +1,8 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<subsystem xmlns="urn:wildfly:micrometer:community:1.1">
+    <otlp-registry endpoint="http://localhost:4318/v1/metrics"/>
```
bstansberry, line 7: Please add the prometheus element to this.

Please also include all legal attributes and use an expression for each.

Basically these files should cover the entire legal xml in order to exercise the subsystem.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerPrometheusTestCase.java
```diff
@@ -0,0 +1,126 @@
+package org.wildfly.test.integration.observability.micrometer;
```
bstansberry, line 1: header

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/setuptask/PrometheusSetupTask.java
```diff
@@ -0,0 +1,28 @@
+package org.wildfly.test.integration.observability.setuptask;
```
bstansberry, line 1: header

bstansberry (changes requested):

observability/micrometer/src/main/resources/org/wildfly/extension/micrometer/LocalDescriptions.properties
bstansberry: @jasondlee It relies on a bug where if text is missing the description uses higher level text.

There should be these items:

micrometer.registry.otlp
micrometer.registry.otlp.add
micrometer.registry.otlp.remove
micrometer.registry.prometheus
micrometer.registry.prometheus.add
micrometer.registry.prometheus.remove

Plus, in theory, micrometer.registry.xxx.yyyy for each attribute of the respective 'xxx'. You have those for  'prometheus'. It seems to work fine without those though, perhaps due to the alias thing. So don't worry about that bit.

jasondlee: Added anyway. :P

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtension.java
```diff
@@ -4,15 +4,24 @@
  */
 package org.wildfly.extension.micrometer;
 
+import org.jboss.as.version.Stability;
 import org.wildfly.subsystem.SubsystemConfiguration;
 import org.wildfly.subsystem.SubsystemExtension;
 import org.wildfly.subsystem.SubsystemPersistence;
 
 public class MicrometerExtension extends SubsystemExtension<MicrometerSubsystemSchema> {
+    static final Stability FEATURE_STABILITY = Stability.DEFAULT;
```
jamezp, line 13: This is only used, as far as I can tell, in this class. It could be private. 

Also, should his be default or community? I'm not certain on this, but I just see other references to community.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtension.java
```diff
@@ -4,15 +4,24 @@
...
         super(SubsystemConfiguration.of(MicrometerConfigurationConstants.NAME,
                         MicrometerSubsystemModel.CURRENT,
                         MicrometerSubsystemRegistrar::new),
                 SubsystemPersistence.of(MicrometerSubsystemSchema.CURRENT));
     }
+
+    @Override
+    public Stability getStability() {
+        return FEATURE_STABILITY;
+    }
```
jamezp, line 25: Is there a need for this since it returns the same value as the default?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
jamezp: I'm not sure I follow this. Is there a reason we didn't upgrade to 2.0?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
jamezp, line 1: Missing copyright.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/WildFlyPrometheusRegistry.java
jamezp, line 1: Missing copyright.

testsuite/integration/microprofile/pom.xml
```diff
@@ -385,7 +385,6 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers</artifactId>
-            <scope>test</scope>
```
jamezp, line 388: Why was the scope removed?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/JaxRsActivator.java
jamezp, line 1: Missing copyright :)

testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/AbstractSetupTask.java
```diff
@@ -16,29 +16,33 @@
...
         if (!Operations.isSuccessfulOutcome(result)) {
             throw new RuntimeException("Failed to execute operation: " + Operations.getFailureDescription(result)
                     .asString());
         }
     }
 
-    protected ModelNode executeRead(final ManagementClient managementClient, ModelNode address) throws IOException {
+    public static ModelNode executeRead(final ManagementClient managementClient, ModelNode address) throws IOException {
         return managementClient.getControllerClient().execute(Operations.createReadResourceOperation(address));
     }
```
jamezp, line 47: Is there a reason these were made public and static?

jamezp (changes requested):
A few missing copyrights. The other comments likely shouldn't block merging though.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtension.java
jasondlee: Actually, both this and the constant above are new here (and the constant above) are new and are both left over from the effort to go into wildfly-ee, so they can be removed. That said, since this never overrode `getStability()` (since it predates that utility), defaulted (har har) to DEFAULT.

testsuite/integration/microprofile/pom.xml
jasondlee: Not sure. That might be a wart from an experiment.

testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/AbstractSetupTask.java
jasondlee: I'm not sure, tbh. I removed static from those and fixed up the one compile problem it caused.

jamezp (approved):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
bstansberry: I *think* what we settled on was adding new lower stability stuff to an existing model results in keeping the same version, different stability.

We have a different problem here though. This PR does 2 things:

1) Rearranges the model for OTLP. That has nothing to do with community stability and requires a new 2.0, Stability.DEFAULT.
2) Adds a new 'community' child resource, and thus a 2.0 Stability.COMMUNITY.

Perhaps we could skip the 2.0 DEFAULT bit if we could absolutely prove (via regression tests) that starting a 'default' stability server with a 1.1 config and modifying an existing 1.1 model attribute results in a persisted config that can be successfully rebooted. IOW, any translation that would introduce the new model structure with the registry=otlp resource would not result in that structure being persisted into xml.

jasondlee: FTR: I have adjusted the version numbers here. The first commit in this PR bumped things to 2.0, then, after some discussion, the second commit bumped it back down to 1.1. I reverted that change and added the 2.0 Community as described here.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Micrometer.adoc
```diff
@@ -22,32 +22,71 @@ This `org.wildfly.extension.micrometer` extension is available to all the standa
...
-[standalone@localhost:9990 /] reload
+[standalone@localhost:9990 /] /subsystem=micrometer/registry=prometheus:add(context="/metrics")
 ----
 
+This will create a context on the management interface from which metrics can be scraped. For example, the above configuration would result in a URL that looks like `http://localhost:9090/metrics`. If you have not disabled the WildFly Metrics subsystem, you will not be able to use `/metrics` as your context, as that is already configured. To disable the WildFly Metrics Subsystem, see the `remove` commands above.
+
 == Management Model
 
-The `/subsystem=micrometer` resource defines three attributes:
+The `/subsystem=micrometer` resource defines one attributes, that is shared by the two supported registries:
```
kabir, line 71: nitpick:
s/one attributes/one attribute

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,62 @@
...
+import org.wildfly.extension.micrometer.otlp.OtlpRegistryDefinitionRegistrar;
+import org.wildfly.extension.micrometer.prometheus.PrometheusRegistryDefinitionRegistrar;
 
 public enum MicrometerSubsystemSchema implements PersistentSubsystemSchema<MicrometerSubsystemSchema> {
-    VERSION_1_0(1, 0), // WildFly 28
-    VERSION_1_1(1, 1), // WildFly 29.0.0.Alpha1
+    VERSION_1_0(1, 0, Stability.DEFAULT), // WildFly 28
+    VERSION_1_1(1, 1, Stability.DEFAULT), // WildFly 29.0.0.Alpha1
+    VERSION_2_0(2, 0, Stability.DEFAULT), // WildFly 35
+    VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY) // WildFly 35
```
kabir, line 28: Probably not important, but I'm wondering why we are doing a major version bump of the schema to 2.0 when for the model version(s) we stayed on 1.1?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemModel.java
```diff
@@ -9,9 +9,10 @@
 
 public enum MicrometerSubsystemModel implements SubsystemModel {
     VERSION_1_0_0(1, 0, 0),
-    VERSION_1_1_0(1, 1, 0);
+    VERSION_1_1_0(1, 1, 0),
+    VERSION_1_1_0_COMMUNITY(1, 1, 0);
```
jamezp, line 13: This should be `VERSION_2_0_0_COMMUNITY(2, 0, 0)` and we need a `VERSION_2_0_0(2, 0, 0)` as well. Basically what you did for `MicrometerSubsystemSchema`.

jamezp (changes requested):

kabir (approved):
A few nitpicks which can be cleaned up later.

I did have a question of the versioning of the xml vs the model, but think I saw/heard something about that somewhere the other day. So that part is probably fine. 

I am mentioning it just in case, so someone more involved is aware (if it is actually an issue)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
bstansberry: Because the model version is wrong :)

https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/WFLY-18582.20Prometheus.20Endpoint.20to.20Micrometer

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,62 @@
...
+        PersistentResourceXMLDescription.Factory factory = factory(this);
+        PersistentResourceXMLDescription.Builder builder =
+            factory.builder(MicrometerSubsystemRegistrar.PATH)
+                .addChild(factory.builder(OtlpRegistryDefinitionRegistrar.PATH)
+                    .addAttributes(OtlpRegistryDefinitionRegistrar.ATTRIBUTES.stream())
+                    .setXmlElementName(OTLP_REGISTRY)
+                    .build());
+
+        builder.addAttributes(MicrometerSubsystemRegistrar.ATTRIBUTES.stream());
+        if (this.since(VERSION_2_0_COMMUNITY)) {
```
pferraro, line 57: N.B. This condition is redundant.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemModel.java
jamezp: @jasondlee See https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/WFLY-18582.20Prometheus.20Endpoint.20to.20Micrometer for some details. Maybe the community version is not needed because the attributes themselves are set to a community stability level.

jasondlee: Yup. Just missed this. Changes incoming...

jasondlee: Fixed

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
jasondlee: Fixed.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtensionTransformerRegistration.java
```diff
@@ -0,0 +1,44 @@
...
+        return MicrometerConfigurationConstants.NAME;
+    }
+
+    @Override
+    public void registerTransformers(SubsystemTransformerRegistration registration) {
+        final ModelVersion currentModel = registration.getCurrentSubsystemVersion();
+        ChainedTransformationDescriptionBuilder chainedBuilder =
+                TransformationDescriptionBuilder.Factory.createChainedSubystemInstance(currentModel);
+
+        // 1.1.0_Community (WildFly 35) to 1.1.0 (WildFly 33)
```
jamezp, line 32: Minor, but this comment is now not quite right :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -70,19 +83,21 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
...
             .setRequired(false)
+            .addFlag(AttributeAccess.Flag.ALIAS)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
 
+    @Deprecated
     public static final SimpleAttributeDefinition STEP = SimpleAttributeDefinitionBuilder
             .create(MicrometerConfigurationConstants.STEP, ModelType.LONG, true)
-            .setAttributeGroup(MicrometerConfigurationConstants.OTLP_REGISTRY)
```
jamezp, line 83: This is more for a model expert as I don't recall, but is removing an attribute from an attribute group a breaking change? @pferraro @yersan

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -70,19 +83,21 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
             "io.micrometer"
     };
 
+    @Deprecated
     public static final SimpleAttributeDefinition ENDPOINT = SimpleAttributeDefinitionBuilder
             .create(MicrometerConfigurationConstants.ENDPOINT, ModelType.STRING)
-            .setAttributeGroup(MicrometerConfigurationConstants.OTLP_REGISTRY)
```
jamezp, line 75: This is more for a model expert as I don't recall, but is removing an attribute from an attribute group a breaking change? @pferraro @yersan

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -70,19 +83,21 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
             "io.micrometer"
     };
 
+    @Deprecated
     public static final SimpleAttributeDefinition ENDPOINT = SimpleAttributeDefinitionBuilder
```
jamezp, line 87: Given the field is deprecated and there is a replacement, should this add `.setDeprecated(ModelVersion.create(2, 0, 0))`?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -70,19 +83,21 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
...
             .create(MicrometerConfigurationConstants.ENDPOINT, ModelType.STRING)
-            .setAttributeGroup(MicrometerConfigurationConstants.OTLP_REGISTRY)
             .setRequired(false)
+            .addFlag(AttributeAccess.Flag.ALIAS)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
 
+    @Deprecated
     public static final SimpleAttributeDefinition STEP = SimpleAttributeDefinitionBuilder
```
jamezp, line 96: Given the field is deprecated and there is a replacement, should this add `.setDeprecated(ModelVersion.create(2, 0, 0))`?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,60 @@
...
+import org.wildfly.extension.micrometer.otlp.OtlpRegistryDefinitionRegistrar;
+import org.wildfly.extension.micrometer.prometheus.PrometheusRegistryDefinitionRegistrar;
 
 public enum MicrometerSubsystemSchema implements PersistentSubsystemSchema<MicrometerSubsystemSchema> {
-    VERSION_1_0(1, 0), // WildFly 28
-    VERSION_1_1(1, 1), // WildFly 29.0.0.Alpha1
+    VERSION_1_0(1, 0, Stability.DEFAULT), // WildFly 28
+    VERSION_1_1(1, 1, Stability.DEFAULT), // WildFly 29.0.0.Alpha1
+    VERSION_2_0(2, 0, Stability.DEFAULT), // WildFly 35
+    VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY) // WildFly 35
```
jamezp, line 28: Nit-pick, but should the comments here be `// WIldFly 36`?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
jasondlee: I think the changes made to these AttributeDefinitions were ill-informed, so I've reverted them.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,60 @@
...
+import org.wildfly.extension.micrometer.otlp.OtlpRegistryDefinitionRegistrar;
+import org.wildfly.extension.micrometer.prometheus.PrometheusRegistryDefinitionRegistrar;
 
 public enum MicrometerSubsystemSchema implements PersistentSubsystemSchema<MicrometerSubsystemSchema> {
-    VERSION_1_0(1, 0), // WildFly 28
-    VERSION_1_1(1, 1), // WildFly 29.0.0.Alpha1
+    VERSION_1_0(1, 0, Stability.DEFAULT), // WildFly 28
+    VERSION_1_1(1, 1, Stability.DEFAULT), // WildFly 29.0.0.Alpha1
+    VERSION_2_0(2, 0, Stability.DEFAULT), // WildFly 36
+    VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY) // WildFly 36
```
jamezp, line 28: This could be right, but I don't really understand it. I just don't see this pattern elsewhere. This will not block my approval though :) 

@pferraro Could you verify this is the correct pattern?

jamezp (dismissed):
I'm approving with the caveat of understanding the model version of 2.0.0 for both default and community.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerService.java
```diff
@@ -68,4 +63,52 @@ private void registerJmxMetrics() {
...
+    void setMicrometerRegistry(WildFlyCompositeRegistry micrometerRegistry) {
+        this.micrometerRegistry = micrometerRegistry;
+    }
+
+
+    public static class Builder {
+        private final MicrometerService service = new MicrometerService();
+
+        public Builder micrometerConfig(WildFlyMicrometerConfig micrometerConfig) {
+            service.setMicrometerConfig(micrometerConfig);
```
kabir, line 89: Why is the service instance cached and setters used to configure that? 
Builders normally temporarily store the fields being set in the builder, and then call the constructor of whatever is being built in the build() method (in which case all the MicrometerService fields can be final again)
I don't see anything so heavy in the construction of MicrometerService warranting this strange approach :-)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/otlp/WildFlyOtlpRegistry.java
```diff
@@ -0,0 +1,111 @@
...
+            } else {
+                return OtlpConfig.super.headers();
+            }
+        }
+
+        public static class Builder {
+            private final WildFlyMicrometerOtlpConfig config = new WildFlyMicrometerOtlpConfig();
+
+            public WildFlyMicrometerOtlpConfig.Builder endpoint(String endpoint) {
+                config.endpoint = endpoint;
```
kabir, line 97: Same observation as before about builders and using final fields in what the builder creates

observability/micrometer/src/test/resources/org/wildfly/extension/micrometer/micrometer-transform-1_1_0.xml
```diff
@@ -0,0 +1,5 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<subsystem xmlns="urn:wildfly:micrometer:1.1"/>
```
kabir, line 5: Where is this used? I don't see any transformer tests :-)

observability/micrometer/src/test/resources/org/wildfly/extension/micrometer/micrometer-reject.xml
```diff
@@ -0,0 +1,7 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<subsystem xmlns="urn:wildfly:micrometer:community:2.0">
```
kabir, line 5: I presume this is for transformer testing, which I cannot see

observability/micrometer/src/test/resources/org/wildfly/extension/micrometer/micrometer-2.0.xml
```diff
@@ -0,0 +1,10 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<subsystem xmlns="urn:wildfly:micrometer:2.0"
+    exposed-subsystems="*">
+    <otlp-registry endpoint="${test.endpoint:http\://localhost\:4318/v1/metrics}"
```
kabir, line 8: When you add transformer tests it would be good to make sure that this xml can be transformed successfully

kabir (changes requested):
@jasondlee I think we need transformers testing?

Also a few other comments

galleon-pack/galleon-shared/pom.xml
```diff
@@ -24,8 +24,6 @@
 
     <name>WildFly: Feature Pack Shared Galleon Content</name>
     <description>Content for the wildfly feature pack that can also be used in the wildfly-preview feature pack</description>
-    <packaging>pom</packaging>
```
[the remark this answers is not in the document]
jasondlee: That looks like a fat fingered error. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerService.java
jasondlee: Yeah, I dunno. That is weird. I've modified that to be a proper builder.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
```diff
@@ -24,11 +23,7 @@ public static Iterable<MicrometerSubsystemSchema> parameters() {
...
 
     public SubsystemParsingTestCase(MicrometerSubsystemSchema schema) {
-        super(MicrometerConfigurationConstants.NAME, new MicrometerExtension(), schema, MicrometerSubsystemSchema.CURRENT);
-    }
-
-    @Override
-    protected String getSubsystemXmlPathPattern() {
-        return "%s_%d_%d.xml";
+        super(MicrometerConfigurationConstants.NAME, new MicrometerExtension(), schema,
+            MicrometerSubsystemSchema.VERSION_2_0_COMMUNITY);
```
pferraro, line 27: This is not correct. The "current" schema will vary depending on the stability assumed by the test.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
```diff
@@ -5,37 +5,59 @@
...
 public enum MicrometerSubsystemSchema implements PersistentSubsystemSchema<MicrometerSubsystemSchema> {
-    VERSION_1_0(1, 0), // WildFly 28
-    VERSION_1_1(1, 1), // WildFly 29.0.0.Alpha1
+    VERSION_1_0(1, 0, Stability.DEFAULT), // WildFly 28
+    VERSION_1_1(1, 1, Stability.DEFAULT), // WildFly 29.0.0.Alpha1
+    VERSION_2_0_COMMUNITY(2, 0, Stability.COMMUNITY) // WildFly 36
     ;
 
-    public static final MicrometerSubsystemSchema CURRENT = VERSION_1_1;
+    public static final Set<MicrometerSubsystemSchema> CURRENT = EnumSet.of(VERSION_1_1, VERSION_2_0_COMMUNITY);
```
pferraro, line 30: From what I can tell, this subsystem will fail to start if the server is run with experimental or preview stability.

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemSchema.java
jasondlee: @pferraro I just started it locally at both preview and experimental, and the server booted without complaint.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/SubsystemParsingTestCase.java
jasondlee: @pferraro How would one modify these types of tests to be Stability-aware? Parameterized over the the different levels?

observability/micrometer/src/test/resources/org/wildfly/extension/micrometer/micrometer-transform-1_1_0.xml
jasondlee: @kabir Transformer tests added

jasondlee:
@bstansberry @pferraro @jamezp Are there any other changes you'd like to see before this week's freeze?

bstansberry (changes requested):
I've got some small things and then the big thing about the context name.

Note I'm not digging through everything here; I'm counting on the feature team for the primary review of this.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -0,0 +1,115 @@
...
+            resourceRegistration.registerAdditionalRuntimePackages(RuntimePackageDependency.required("io.prometheus"));
+        }
+        ManagementResourceRegistrar.of(description).register(resourceRegistration);
+
+        return resourceRegistration;
+    }
+
+    @Override
+    public ResourceServiceInstaller configure(OperationContext context, ModelNode model) throws OperationFailedException {
+        String serviceContext = CONTEXT.resolveModelAttribute(context, model).asStringOrNull();
```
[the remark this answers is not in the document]
jasondlee: @bstansberry Is this something you want done for WF 36 (beta?), or can we file an issue and address it either in .Final or WF 37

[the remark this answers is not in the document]
bstansberry: @jasondlee If this gets into Beta, please file a blocker bug for 36 Final. I don't object to resolving it later.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerExtensionTransformerRegistration.java
```diff
@@ -0,0 +1,123 @@
...
+            public void transformResource(ResourceTransformationContext context, PathAddress address, Resource resource) {
+                if (resource.hasChild(OtlpRegistryDefinitionRegistrar.PATH)) {
+                    Resource otlp = resource.removeChild(OtlpRegistryDefinitionRegistrar.PATH);
+                    resource.getModel().get(STEP.getName()).set(otlp.getModel().get(STEP.getName()));
+                    resource.getModel().get(ENDPOINT.getName()).set(otlp.getModel().get(ENDPOINT.getName()));
+                }
+                context.addTransformedResourceFromRoot(address, resource);
+
+                // TODO I am not sure what happens here, if there were other children which should be included
+                // Normally there seems to be a call to context.processChildren() or for each child individually
```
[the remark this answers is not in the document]
jasondlee: You're right. That's just a wart, I think. I've removed it.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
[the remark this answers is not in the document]
jasondlee: @bstansberry https://issues.redhat.com/browse/WFLY-20482

jamezp (approved):

PR #17862, opened 2024-05-01
[WFLY-19223] LayersTestCase should pass with new org.wildfly.security.elytron-dynamic-ssl module

skyllarr:
https://issues.redhat.com/browse/WFLY-19223

Requires: https://github.com/wildfly/wildfly-core/pull/5947
____

____

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
@@ -54,6 +54,7 @@ public abstract class LayersTestBase {
             "org.jboss.as.xts",
             // TODO should an undertow layer specify this?
             "org.wildfly.event.logger",
+            "org.wildfly.security.elytron-dynamic-ssl",
     };
 
     /**


Skyllarr:
@yersan this PR solves the LayersTestCase failures happening on https://github.com/wildfly/wildfly-core/pull/5947

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -54,6 +54,7 @@ public abstract class LayersTestBase {
             "org.jboss.as.xts",
             // TODO should an undertow layer specify this?
             "org.wildfly.event.logger",
+            "org.wildfly.security.elytron-dynamic-ssl",
```
yersan, line 57: In my opinion, this is not the expected fix.

`test-all-layers` installation contains the `elytron` galleon layer, so the `org.wildfly.security.elytron-dynamic-ssl` should have been provisioned by the `test-all-layers`, which makes this change unnecessary.

The fix should be in WildFly Core PR, I'll comment more about it there. The problem is that `elytron` galleon layer is not provisioning `org.wildfly.security.elytron-dynamic-ssl` package.

yersan (changes requested):
@Skyllarr I think this PR can be closed, in my opinion, this is not the expected fix.

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
yersan: @bstansberry I guess we also are going to need a solution to tweak all of these galleon layers tests based on provisioning stability levels. Not sure if we have already discussed the approach for them.

Skyllarr:
@yersan Thank you for looking into this and explaining, I'm closing this PR

PR #17866, opened 2024-05-02
WFLY-19304: update model definition and bump schema

michpetrov:
Issue: [WFLY-19304](https://issues.redhat.com/browse/WFLY-19304)

Updated the schema to reflect what's actually produced by the XmlWriter. The bigger part is keeping the elements in the correct order, though I did change the order in the writer as well to keep it the same between `datasource` and `xa-datasource`. Since the parser doesn't care about the order this should not affect backwards compatibility.

Second part is changing some elements from "boolean presence" - `<element />` to simple boolean - `<element>true</element>`. The elements declared as "boolean presence" were neither written nor parsed as such, `<elem />` and `<elem>true</elem>` should be equivalent but `<elem />` is actually considered "false", this is potentially breaking backwards compatibility but if so it was broken already.

Thirdly two attributes had allowed values added since that's what's expected.

____

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
@@ -24,8 +24,10 @@ import org.jboss.as.controller.SimpleOperationDefinition;
 import org.jboss.as.controller.SimpleOperationDefinitionBuilder;
 import org.jboss.as.controller.access.constraint.SensitivityClassification;
 import org.jboss.as.controller.access.management.SensitiveTargetAccessConstraintDefinition;
+import org.jboss.as.controller.client.helpers.MeasurementUnit;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
 import org.jboss.as.controller.descriptions.NonResolvingResourceDescriptionResolver;
+import org.jboss.as.controller.operations.validation.EnumValidator;
 import org.jboss.as.controller.operations.validation.StringLengthValidator;
 import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.security.CredentialReference;
@@ -39,6 +41,7 @@ import org.jboss.jca.common.api.metadata.ds.Driver;
 import org.jboss.jca.common.api.metadata.ds.DsPool;
 import org.jboss.jca.common.api.metadata.ds.Statement;
 import org.jboss.jca.common.api.metadata.ds.TimeOut;
+import org.jboss.jca.common.api.metadata.ds.TransactionIsolation;
 import org.jboss.jca.common.api.metadata.ds.Validation;
 import org.jboss.jca.common.api.metadata.ds.XaDataSource;
 
@@ -431,7 +434,8 @@ public class Constants {
     static SimpleAttributeDefinition TRACK_STATEMENTS = new SimpleAttributeDefinitionBuilder(TRACKSTATEMENTS_NAME, ModelType.STRING, true)
             .setAllowExpression(true)
             .setXmlName(Statement.Tag.TRACK_STATEMENTS.getLocalName())
-            .setDefaultValue(new ModelNode(Defaults.TRACK_STATEMENTS.name()))
+            .setDefaultValue(new ModelNode(Defaults.TRACK_STATEMENTS.toString()))
+            .setValidator(EnumValidator.create(Statement.TrackStatementsEnum.class))
             .setRestartAllServices()
             .build();
 
@@ -444,6 +448,7 @@ public class Constants {
 
     static SimpleAttributeDefinition ALLOCATION_RETRY_WAIT_MILLIS = new SimpleAttributeDefinitionBuilder(ALLOCATION_RETRY_WAIT_MILLIS_NAME, ModelType.LONG, true)
             .setXmlName(TimeOut.Tag.ALLOCATION_RETRY_WAIT_MILLIS.getLocalName())
+            .setMeasurementUnit(MeasurementUnit.MILLISECONDS)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
@@ -472,6 +477,7 @@ public class Constants {
 
     static SimpleAttributeDefinition QUERY_TIMEOUT = new SimpleAttributeDefinitionBuilder(QUERYTIMEOUT_NAME, ModelType.LONG, true)
             .setXmlName(TimeOut.Tag.QUERY_TIMEOUT.getLocalName())
+            .setMeasurementUnit(MeasurementUnit.SECONDS)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
@@ -492,6 +498,9 @@ public class Constants {
     static SimpleAttributeDefinition TRANSACTION_ISOLATION = new SimpleAttributeDefinitionBuilder(TRANSACTION_ISOLATION_NAME, ModelType.STRING, true)
             .setXmlName(DataSource.Tag.TRANSACTION_ISOLATION.getLocalName())
             .setAllowExpression(true)
+            .setAllowedValues(TransactionIsolation.TRANSACTION_NONE.name(), TransactionIsolation.TRANSACTION_READ_COMMITTED.name(),
+                    TransactionIsolation.TRANSACTION_READ_UNCOMMITTED.name(), TransactionIsolation.TRANSACTION_REPEATABLE_READ.name(),
+                    TransactionIsolation.TRANSACTION_SERIALIZABLE.name())
             .setRestartAllServices()
             .build();
 
@@ -623,6 +632,7 @@ public class Constants {
 
     static SimpleAttributeDefinition XA_RESOURCE_TIMEOUT = new SimpleAttributeDefinitionBuilder(XA_RESOURCE_TIMEOUT_NAME, ModelType.INT, true)
             .setXmlName(TimeOut.Tag.XA_RESOURCE_TIMEOUT.getLocalName())
+            .setMeasurementUnit(MeasurementUnit.SECONDS)
             .setAllowExpression(true)
             .setRestartAllServices()
             .build();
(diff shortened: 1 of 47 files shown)


michpetrov:
@tadamski could I get a review? I'm planning to update the rest of the connector subsystems too.

tadamski (commented):
@michpetrov sorry for late review. The PR looks good to me regarding configuration processing. I'm just wondering: if the changes from 7.1 to 7.2 are a bugfix (ammending wrong configuration instead of adding anything new), maybe this should be treated as 7.1 fix rather than schema update. OTOH, if the new schema is needed - don't we need any 7.2 -> 7.1 transformers?

michpetrov:
@tadamski I assumed making changes to schema requires a version bump no matter what but if not then this can be a fix for 7.1, I don't think there's anything that requires a transformer, the order of the elements changes but the parser doesn't care about that.

tadamski:
ok, so iiuc this fix makes sure that configuration that was parsed and persisted is consistent with the schema, but it should have no impact on any correct configuration neither in 7.1 nor 7.2 schema. tbh I'm still not certain if in such case a version bump is needed, and another question is: I assume all previous configs would have this problem as well so should we treat it like a bug and ammend all previous xsd that are still supported, or indeed just create 7.2 and say that this is a version since which the config bug was fixed. @bstansberry wdyt?

bstansberry (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourcesExtension.java
```diff
@@ -146,7 +148,7 @@ public class DataSourcesExtension implements Extension {
     public static final String SUBSYSTEM_NAME = Constants.DATASOURCES;
     private static final String RESOURCE_NAME = DataSourcesExtension.class.getPackage().getName() + ".LocalDescriptions";
 
-    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(6, 0, 0);
+    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(7, 2, 0);
```
bstansberry, line 151: Note: this being at 6, 0, 0 instead of 7,0,0 was a serious bug.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
```diff
@@ -492,6 +497,9 @@ public class Constants {
     static SimpleAttributeDefinition TRANSACTION_ISOLATION = new SimpleAttributeDefinitionBuilder(TRANSACTION_ISOLATION_NAME, ModelType.STRING, true)
             .setXmlName(DataSource.Tag.TRANSACTION_ISOLATION.getLocalName())
             .setAllowExpression(true)
+            .setAllowedValues(TransactionIsolation.TRANSACTION_NONE.name(), TransactionIsolation.TRANSACTION_READ_COMMITTED.name(),
+                    TransactionIsolation.TRANSACTION_READ_UNCOMMITTED.name(), TransactionIsolation.TRANSACTION_REPEATABLE_READ.name(),
+                    TransactionIsolation.TRANSACTION_SERIALIZABLE.name())
```
bstansberry, line 502: This validation is incorrect per the existing code. See the uses of TransactionIsolation.customLevel(...) which allows arbitrary values.

Changing that would be a breaking change and the API would still have to allow the values on an HC for mixed-domain use, while rejecting them on a currently running server. Such a change should be handled under a different JIRA, assuming it's wanted.

bstansberry:
@tadamski re your policy questions, AIUI there are three basic changes here:

1) Changes to the management API itself. There are some trivial things (measurement units, case of a default value) and then a serious thing, the addition of validation to 2 attributes.

I believe that these changes have no functional impact vs the existing behavior (or won't once my comments about the validators are addressed) so in a crisis we *might* skip the management API version bump. But given the magnitude of the related changes and the presence of the important-to-the-user validation information, we should bump the management API.

I don't see a need for new transformer code, because there is no change in what a DC processing this 7.2 version of the api would send to an HC that uses an earlier version.

2) The 7.1 and earlier xsds misinform the user re valid configuration.  This is a good reason to issue a new, correct xsd. In general we do not want to update existing xsds. Having multiple variants of the same xsd floating around is confusing and not what users generally expect.

The existing handling of transaction-isolation appears to be inconsistent with the xsd, since arbitrary values are allowed (see TranactionIsolation.customLevel) while the xsd says they are not allowed. But that doesn't require us to update existing xsds -- parsers can be more lenient than the xsd.  Really the new xsd shouldn't have the restriction around these; having it would only be because TranactionIsolation.customLevel is a kind of hidden feature, e.g. kept for backward compatibility but not something we want to advertise via the xsd.

2) A bunch of marshaling changes.

WF requires subsystems to marshal in a format consistent with the xsd, and that includes the order of elements and attributes. The ordering requirement is for good UX -- if a user provides a document that is ordered per the current xsd, we should not reorder when we marshal. We can't marshal in the same order they used if it's inconsistent with the xsd, as we don't preserve the ordering info in memory, but at least we can avoid irrelevant document changes if they order per the xsd.

So, if we are marshaling in an order inconsistent with the current xsd, we have a bug that must be fixed. The fix is either to fix the marshaling, or if we think the marshaling is the order we want, then we issue a new xsd. It sounds like in this case the fix is doing both.

Note that documents used in the subsystem test should be written in xsd order (at least those that use the current xsd), so the subsystem test can validate that the marshaling is also in xsd order, by comparing the input and output xml. I assume that wasn't case previously.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/DataSourcesExtension.java
bstansberry: Well, a potentially serious bug. It seems 7.0.0 and 6.0.0 were equivalent in terms of domain mode, so the bug did no harm.

tadamski:
@bstansberry thanks for the thorough explanation

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
```diff
@@ -431,7 +433,8 @@ public class Constants {
     static SimpleAttributeDefinition TRACK_STATEMENTS = new SimpleAttributeDefinitionBuilder(TRACKSTATEMENTS_NAME, ModelType.STRING, true)
             .setAllowExpression(true)
             .setXmlName(Statement.Tag.TRACK_STATEMENTS.getLocalName())
-            .setDefaultValue(new ModelNode(Defaults.TRACK_STATEMENTS.name()))
+            .setDefaultValue(new ModelNode(Defaults.TRACK_STATEMENTS.toString()))
+            .setAllowedValues(Arrays.stream(Statement.TrackStatementsEnum.values()).map(Statement.TrackStatementsEnum::toString).toArray(String[]::new))
```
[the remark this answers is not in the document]
michpetrov: @bstansberry as far as I can tell `setAlowedValues` doesn't actually enforce any validation so I'm not forcing the user to do anything (but that also means I didn't solve the problem)
I see the better way would be to write the values in lowercase but I'm guessing I need a custom marshaller for that?

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
michpetrov: Ah, I didn't see the customLevel, we're losing the ability to show suggestions in CLI/HAL but the custom level makes that unfeasible anyway.
But the custom level would fail validation against the xsd, I'm assuming we're fine with that?

michpetrov:
@bstansberry the issue with the test files is that only one is getting validated against xsd (hence why this didn't pop up before) I've filed [WFCORE-6812](https://issues.redhat.com/browse/WFCORE-6812) for that. Otherwise I think most of the files are validated in terms of the content being compared with the marshalled output.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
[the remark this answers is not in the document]
bstansberry: @michpetrov Sorry; it seems I misremembered! I think at one point someone investigated having this setting trigger automatic registration of a validator. But maybe it didn't happen for some reason or it just got recorded as an idea.

Anyway since this is the case, using setValidator(EnumValidator) seems better than setAllowedValues. DatasourceModelNodeUtil already calls Statement.TrackStatementsEnum.valueOf with this value and that will fail if the value is wrong. So the code is 'validating', just at the wrong point in operation execution.

EnumValidator also attempts to store the Enum.toString() value in the model, so using it results in accepting both the Enum.toString() and Enum.name() values, but storing toString().

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
bstansberry: We don't validate these user documents against the xsd, so in general, yes.

For the existing xsd versions, for sure yes.

For the new xsd I'll defer to @tadamski as to whether he wants to remove the restriction from the xsd.

If he wants to keep it in the xsd and this custom level thing is some sort of hidden thing, I suppose your setAllowedValues can be kept as well, since you reminded me it's just informational and doesn't trigger validation. If you do, please use the CLI to exercise the custom level thing; I don't remember clearly any more what if anything the CLI may do to try and help enforce allowed values.

michpetrov:
added the EumValidator and rebased

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/Constants.java
tadamski: I gave a quick look at the history of why we needed a custom level since those allowed values are part of SQL standard, and it turns out that the culprit is MySQL's SNAPSHOT level.

MySQL seems like important use case, but since other values are standard based, I'm leaning towards keeping custom level hidden and keep setAllowedValues. Maybe it would also make sense to add a TransactionIsolation.SNAPSHOT (with comment that is SQLServer thing), and keep custom value hidden for some exotic cases.

darranl (changes requested):
Can we please move the schema bump into it's own commit - it is not possible from the GitHub UI to see what has changed in the schema,

michpetrov:
@darranl I've split the fix and refactor into separate commits

bstansberry (approved):
This LGTM.

@tadamski If you'd like to add TransactionIsolation.SNAPSHOT, I suggest you handle that via a separate JIRA so we can get this one done.

I've manually confirmed that the setting transaction-isolation=SNAPSHOT works, so the existing 'hidden' feature works.

bstansberry:
@tadamski Re TransactionIsolation.SNAPSHOT I filed https://issues.redhat.com/browse/WFLY-19677 in case it's wanted. If not, just reject that.

bstansberry:
Thanks @michpetrov and @tadamski

PR #17876, opened 2024-05-07
[WFLY-19320] WildFly Transaction client module is unsupported module

ropalka:
https://issues.redhat.com/browse/WFLY-19320

____

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml
@@ -7,6 +7,10 @@
 
 <module xmlns="urn:jboss:module:1.9" name="org.wildfly.transaction.client">
 
+    <properties>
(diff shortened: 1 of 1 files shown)


ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/transaction/client/main/module.xml
```diff
@@ -7,6 +7,10 @@
 
 <module xmlns="urn:jboss:module:1.9" name="org.wildfly.transaction.client">
 
+    <properties>
+        <property name="jboss.api" value="unsupported"/>
```
bstansberry, line 11: Unsupported is a downstream-only setting. It means "'public' upstream but 'private' downstream" so it should never be used upstream.

bstansberry:
I added the 'hold' label pending further discussion of a deprecation plan, documentation of a migration path, update of the JIRA with more details, etc.

asoldano:
@tadamski @rachmatowicz FYI

bstansberry:
I'm going to close this and the 3 other similar ones. There needs to be discussion, not via the PR queue.

PR #17878, opened 2024-05-08
[WFLY-19324] Migrate OpenTelemetry extension to new wildfly-subsystem approach

jasondlee:

- Add wildfly-subsystem module dep to extension
- Add wildfly-subsystem maven dep
- Migrate Extension to SubsystemExtension<OpenTelemetrySubsystemSchema> 
- Rename OpenTelemetrySubsystemDefinition to OpenTelemetrySubsystemRegistrar
- Modify OpenTelemetrySubsystemRegistrar inheritance and update implementation


observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -60,13 +70,14 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
     };
 
     static final RuntimeCapability<Void> OPENTELEMETRY_CAPABILITY =
-            RuntimeCapability.Builder.of(OPENTELEMETRY_CAPABILITY_NAME)
+            RuntimeCapability.Builder.of(OPENTELEMETRY_CAPABILITY_NAME, Void.class)
```
pferraro, line 73: This implies that the "org.wildfly.extension.opentelemetry" capability provides a "Void" via a service.  However, no such service installed by your ResourceOperationRuntimeHandler.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +169,71 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
+                OpenTelemetrySubsystemRegistrar.RATIO.resolveModelAttribute(context, model).asDoubleOrNull(),
+                context.getCapabilityServiceSupport().hasCapability("org.wildfly.extension.microprofile.telemetry")
         );
+
+        Supplier<WildFlyOpenTelemetryConfig> configSupplier = () -> config;
+        opentelemetryConfig.set(config);
+
+        return CapabilityServiceInstaller.builder(OPENTELEMETRY_CONFIG_CAPABILITY, configSupplier)
+                .asActive()
+                .build();
```
pferraro, line 236: As this is a fixed value (i.e. resolvable the need for dependencies only available after starting), you can simplify this as:

```java
return CapabilityServiceInstaller.builder(OPENTELEMETRY_CONFIG_CAPABILITY, config).build();
```

Though, does this even need to be a service?  Where is it comsumed?

pferraro (changes requested):

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: I added that... for some reason. Did that second param used to be required? Removing it, though, doesn't seem to break anything. That said, I've removed the capability altogether, and am testing that.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: The AtomicReference is consumed here: https://github.com/wildfly/wildfly/blob/c6370a7737e14a1d7cb4344a2b897cfe6d6c476a/observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java#L203

It seems weird to build a service (?) that offers that, but it's not consumed, while the value it offers is accessed via the AtomReference instead. I don't know how to do that more correctly, though, so advice is welcome here.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: Turns out that capability is used by the MP Telemetry subsystem to make sure the OpenTelemetry system is installed.

pferraro: OK - in that case, it should not declare a service type.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
pferraro: In that case, you do not need to install a service.

pferraro (changes requested):

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +169,71 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
+                .addCapability(OPENTELEMETRY_CAPABILITY)
+                .withRuntimeHandler(ResourceOperationRuntimeHandler.configureService(this))
+                .addAttributes(ATTRIBUTES)
+                .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
+                .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
+                .withDeploymentChainContributor(target -> {
+                    // We need to disable vertx's DNS resolver as it causes failures under k8s
+                    if (System.getProperty(VERTX_DISABLE_DNS_RESOLVER) == null) {
+                        System.setProperty(VERTX_DISABLE_DNS_RESOLVER, "true");
+                    }
```
pferraro, line 195: This is a little strange to execute here.  Perhaps move this to a static code block?

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +169,71 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
+                    target.addDeploymentProcessor(OpenTelemetryConfigurationConstants.SUBSYSTEM_NAME,
+                            Phase.DEPENDENCIES,
+                            0x1910,
```
pferraro, line 198: This value should reside in Phase.java within wildfly-core.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +169,71 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
+                            new OpenTelemetryDependencyProcessor());
+                    target.addDeploymentProcessor(OpenTelemetryConfigurationConstants.SUBSYSTEM_NAME,
+                            Phase.POST_MODULE,
+                            0x3810,
```
pferraro, line 202: This value should reside in Phase.java within wildfly-core.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: IIRC, I added that method so that I could build the WildFlyOpenTelemetryConfig instance, for which I need the context and model. To do that in configure(), I have to return a ResourceServiceInstaller. Is there a better way/place to build that WildFlyOpenTelemetryConfig? It doesn't look like I can do that in the DUP, which is the only place that consumes it, ultimately.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: Done.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: Fixed.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryDeploymentProcessor.java
```diff
@@ -24,13 +21,15 @@
...
 import org.wildfly.extension.opentelemetry.api.WildFlyOpenTelemetryConfig;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
 class OpenTelemetryDeploymentProcessor implements DeploymentUnitProcessor {
-    private final boolean useServerConfig;
-    private final WildFlyOpenTelemetryConfig serverConfig;
+    private final AtomicReference<WildFlyOpenTelemetryConfig> configSupplier;
```
pferraro, line 29: Nit: As this DUP only reads this reference, better to use a Supplier<WildFlyOpenTelemetryConfig>.
e.g.
```java
AtomicReference<WildFlyOpenTelemetryConfig> configReference = new AtomicReference<>();
// ...
target.addDeploymentProcessor(..., new OpenTelemetryDeploymentProcessor(configReference::get));

File: observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java [hunk 7]
```diff
@@ -184,31 +248,26 @@ static void validateExporter(OperationContext context, String exporter) throws O
...
+//            if(!attr.getFlags().contains(AttributeAccess.Flag.RESTART_ALL_SERVICES)) {
+//                throw ControllerLogger.ROOT_LOGGER.attributeWasNotMarkedAsReloadRequired(attr.getName(), resourceRegistration.getPathAddress());
+//            }
+//            resourceRegistration.registerReadWriteAttribute(attr, null, handler);
+//        }
+//    }
 
-    static class WildFlyOpenTelemetryConfigSupplier implements Supplier<OpenTelemetryConfig>, Consumer<OpenTelemetryConfig> {
-        private OpenTelemetryConfig config;
+    static class WildFlyOpenTelemetryConfigSupplier implements Supplier<WildFlyOpenTelemetryConfig>,
```
pferraro, line 261: Where is this used?

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
pferraro: > To do that in configure(), I have to return a ResourceServiceInstaller.

If your runtime does not install services you do not need ResourceServiceConfigurator.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: I don't think it is. If not, it will go away before this PR is final

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: Where else would I build `WildFlyOpenTelemetryConfig`?

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +170,76 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
             .build();
 
-    public static final AttributeDefinition[] ATTRIBUTES = {
+    public static final List<AttributeDefinition> ATTRIBUTES = List.of(
             SERVICE_NAME, EXPORTER, ENDPOINT, SPAN_PROCESSOR_TYPE, BATCH_DELAY, MAX_QUEUE_SIZE, MAX_EXPORT_BATCH_SIZE,
             EXPORT_TIMEOUT, SAMPLER, RATIO
-    };
+    );
+
+    private Supplier<WildFlyOpenTelemetryConfig> opentelemetryConfig;
```
pferraro, line 178: If you want to feed your DUPs directly...
```java
private final AtomicReference<OpenTelemetryConfig> openTelemetryConfig = new AtomicReference<>();
```
Then, in your deployment chain contributor, create your OpenTelemetryDeploymentProcessor via:
```java
new OpenTelemetryDeploymentProcessor(this.openTelemetryConfig::get)
```

and then set the value of the atomic reference in your runtime handler.

jasondlee: I don't have a strong preference. I just want to do it The Right Way. :)

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -158,18 +170,76 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
+                OpenTelemetrySubsystemRegistrar.RATIO.resolveModelAttribute(context, model).asDoubleOrNull(),
+                context.getCapabilityServiceSupport().hasCapability("org.wildfly.extension.microprofile.telemetry")
         );
+
+        opentelemetryConfig = () -> config;
+
+        return CapabilityServiceInstaller.builder(OPENTELEMETRY_CONFIG_CAPABILITY, opentelemetryConfig)
+                .provides(ServiceNameFactory.resolveServiceName(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR))
+                .asActive()
+                .build();
```
pferraro, line 242:
```java
return CapabilityServiceInstaller.builder(OPENTELEMETRY_CONFIG_CAPABILITY, config)
        .withCaptor(this.openTelemetryConfig::set)
        .build();
```

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -60,13 +70,14 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
...
     static final RuntimeCapability<Void> OPENTELEMETRY_CAPABILITY =
-            RuntimeCapability.Builder.of(OPENTELEMETRY_CAPABILITY_NAME)
+            RuntimeCapability.Builder.of(OPENTELEMETRY_CAPABILITY_NAME, Void.class)
                     .addRequirements(WELD_CAPABILITY_NAME)
                     .build();
 
-    public static final WildFlyOpenTelemetryConfigSupplier CONFIG_SUPPLIER = new WildFlyOpenTelemetryConfigSupplier();
-    static final RuntimeCapability<WildFlyOpenTelemetryConfigSupplier> OPENTELEMETRY_CONFIG_CAPABILITY =
-            RuntimeCapability.Builder.of(OPENTELEMETRY_MODULE + ".config", false, CONFIG_SUPPLIER).build();
+    public static final RuntimeCapability<Void> OPENTELEMETRY_CONFIG_CAPABILITY =
```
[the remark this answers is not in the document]
pferraro: Nm - I understand now.

observability/opentelemetry-api/src/main/java/org/wildfly/extension/opentelemetry/api/WildFlyOpenTelemetryConfig.java
```diff
@@ -10,8 +10,13 @@
 import java.util.Map;
 
 import io.smallrye.opentelemetry.api.OpenTelemetryConfig;
+import org.wildfly.service.descriptor.NullaryServiceDescriptor;
 
 public final class WildFlyOpenTelemetryConfig implements OpenTelemetryConfig {
+    public static NullaryServiceDescriptor<WildFlyOpenTelemetryConfig> SERVICE_DESCRIPTOR =
+            NullaryServiceDescriptor.of("org.wildfly.extension.opentelemetry.config",
+                WildFlyOpenTelemetryConfig.class);
```
pferraro, line 18: Do consumers need this specific implementation?  If not, just expose the interface.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -64,9 +74,11 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
                     .addRequirements(WELD_CAPABILITY_NAME)
                     .build();
 
-    public static final WildFlyOpenTelemetryConfigSupplier CONFIG_SUPPLIER = new WildFlyOpenTelemetryConfigSupplier();
-    static final RuntimeCapability<WildFlyOpenTelemetryConfigSupplier> OPENTELEMETRY_CONFIG_CAPABILITY =
-            RuntimeCapability.Builder.of(OPENTELEMETRY_MODULE + ".config", false, CONFIG_SUPPLIER).build();
+    public static final String OPENTELEMETRY_CONFIG_SERVICE_NAME = OPENTELEMETRY_MODULE + ".config";
+    public static final RuntimeCapability<Void> OPENTELEMETRY_CONFIG_CAPABILITY =
+            RuntimeCapability.Builder.of(OPENTELEMETRY_CONFIG_SERVICE_NAME, WildFlyOpenTelemetryConfig.class)
+                    .build();
```
pferraro, line 80: You can create this from your ServiceDescriptor.
e.g.
```java
RuntimeCapability.Builder.of(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR).build();
```

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
```diff
@@ -64,9 +74,11 @@ class OpenTelemetrySubsystemDefinition extends PersistentResourceDefinition {
                     .addRequirements(WELD_CAPABILITY_NAME)
                     .build();
 
-    public static final WildFlyOpenTelemetryConfigSupplier CONFIG_SUPPLIER = new WildFlyOpenTelemetryConfigSupplier();
-    static final RuntimeCapability<WildFlyOpenTelemetryConfigSupplier> OPENTELEMETRY_CONFIG_CAPABILITY =
-            RuntimeCapability.Builder.of(OPENTELEMETRY_MODULE + ".config", false, CONFIG_SUPPLIER).build();
+    public static final String OPENTELEMETRY_CONFIG_SERVICE_NAME = OPENTELEMETRY_MODULE + ".config";
```
pferraro, line 77: This name is made redundant by your service descriptor.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: https://github.com/wildfly/wildfly-core/pull/6075

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemRegistrar.java
jasondlee: https://github.com/wildfly/wildfly-core/pull/6075

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/BlarghLogger.java
```diff
@@ -0,0 +1,7 @@
+package org.wildfly.test.integration.observability;
+
+public class BlarghLogger {
+    public static void blargh(String msg) {
```
jasondlee, line 4: Goofy test logging class. I'll remove this before merging.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerOtelIntegrationTestCase.java
```diff
@@ -99,9 +92,13 @@ public void makeRequests() throws URISyntaxException {
 
     @Test
     @InSequence(3)
-    public void checkCounter() {
-        Counter counter = meterRegistry.get("demo_counter").counter();
-        Assert.assertEquals(counter.count(), REQUEST_COUNT, 0.0);
+    public void checkCounter() throws InterruptedException {
+        Thread.sleep(1000); // Allow time for metrics to be pushed
```
bstansberry, line 96: Use TimeoutUtil for this.

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
```diff
@@ -42,7 +55,7 @@ public void execute(DeploymentProcessorTarget processorTarget) {
                         MicroProfileTelemetryExtension.SUBSYSTEM_NAME,
                         Phase.POST_MODULE,
                         Phase.POST_MODULE_MICROPROFILE_TELEMETRY,
-                        new MicroProfileTelemetryDeploymentProcessor());
+                        new MicroProfileTelemetryDeploymentProcessor(openTelemetryConfig::get));
```
bstansberry, line 58: This isn't safe. AFAICT, there's no happens-before / happens-after relationship between the captor passed to L40 being invoked and this supplier being invoked. For it to be safe something would need to establish a dependency relationship between the DeploymentUnitPhaseService for Phase.POST_MODULE and the service installed at L42.

To add such a happens-before/happens-after you'd need to give your service a name and use a DUP running in a phase before POST_MODULE to add a dep on it to the next phase's DUPS. It's a bit kludgy but MicroProfileTelemetryDependencyProcessor could do it.

In practice this is quite unlikely to fail because of all the work that happens between L42 and when the DeploymentUnitPhaseService for Phase.POST_MODULE starts and calls MicroProfileTelemetryDeploymentProcessor

bstansberry, line 58: This isn't safe. AFAICT, there's no happens-before / happens-after relationship between the captor passed to L40 being invoked and this supplier being invoked. For it to be safe something would need to establish a dependency relationship between the DeploymentUnitPhaseService for Phase.POST_MODULE and the service installed at L42.

To add such a happens-before/happens-after you'd need to give your service a name and use a DUP running in a phase before POST_MODULE to add a dep on it to the next phase's DUPS. It's a bit kludgy but MicroProfileTelemetryDependencyProcessor could do it.

In practice this is quite unlikely to fail because of all the work that happens between L42 and when the DeploymentUnitPhaseService for Phase.POST_MODULE starts and calls MicroProfileTelemetryDeploymentProcessor

bstansberry (changes requested):

microprofile/telemetry-smallrye/extension/pom.xml
```diff
@@ -146,6 +146,14 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly</groupId>
```
jamezp, line 150: Minor nit. Elsewhere the `${project.groupId}` is used so it should likely be used here too.

observability/opentelemetry-api/src/main/java/org/wildfly/extension/opentelemetry/api/WildFlyOpenTelemetryConfig.java
```diff
@@ -30,14 +35,16 @@ public final class WildFlyOpenTelemetryConfig implements OpenTelemetryConfig {
     public static final String OTEL_TRACES_SAMPLER_ARG = "otel.traces.sampler.arg";
 
     private final Map<String, String> properties;
+    private boolean mpTelemetryInstalled = false;
```
jamezp, line 38: Any reason for this not to be final?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/BasicMicrometerTestCase.java
jamezp, line 1: This is missing the copyright header.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/BasicMicrometerTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+    @Inject
+    private MeterRegistry meterRegistry;
+
+    // The @ServerSetup(MicrometerSetupTask.class) requires Docker to be available.
+    // Otherwise the org.wildfly.extension.micrometer.registry.NoOpRegistry is installed which will result in 0 counters,
+    // and cause the test fail seemingly intermittently on machines with broken Docker setup.
+    @BeforeClass
+    public static void checkForDocker() {
+        AssumeTestGroupUtil.assumeDockerAvailable();
+    }
```
jamezp, line 35: Given the `@DockerRequired(AssumptionViolatedException.class)`, this should not be needed.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerSetupTask.java
```diff
@@ -6,66 +6,59 @@
...
-    private final ModelNode micrometerSubsystem = Operations.createAddress("subsystem", "micrometer");
-    private boolean extensionAdded = false;
-    private boolean subsystemAdded = false;
+    @Testcontainer
     private OpenTelemetryCollectorContainer otelCollector;
 
     @Override
     public void setup(final ManagementClient managementClient, String containerId) throws Exception {
+        AssumeTestGroupUtil.assumeDockerAvailable();
+
```
jamezp, line 35: This should not be required since `@DockerRequired(AssumptionViolatedException.class)` is on the task.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BasicOpenTelemetryTestCase.java
```diff
@@ -33,6 +37,11 @@ public class BasicOpenTelemetryTestCase extends BaseOpenTelemetryTest {
     @Inject
     private Baggage baggage;
 
+    @BeforeClass
+    public static void checkForDocker() {
+        AssumeTestGroupUtil.assumeDockerAvailable();
+    }
```
jamezp, line 43: Same as the others, this should not be required.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/ContextPropagationTestCase.java
```diff
@@ -38,17 +40,25 @@
...
 
     @BeforeClass
     public static void disableForWFLY19509() {
-        IntermittentFailure.thisTestIsFailingIntermittently("WFLY-19509");
+//        IntermittentFailure.thisTestIsFailingIntermittently("WFLY-19509");
+    }
+
+    @BeforeClass
+    public static void checkForDocker() {
+        AssumeTestGroupUtil.assumeDockerAvailable();
```
jamezp, line 55: Same as above about not being required :)

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationTestCase.java
```diff
@@ -33,19 +36,27 @@
...
 
     @BeforeClass
     public static void disableForWFLY19509() {
-        IntermittentFailure.thisTestIsFailingIntermittently("WFLY-19510");
+//        IntermittentFailure.thisTestIsFailingIntermittently("WFLY-19510");
+    }
+
+    @BeforeClass
+    public static void checkForDocker() {
+        AssumeTestGroupUtil.assumeDockerAvailable();
```
jamezp, line 51: Same as above :)

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BaseOpenTelemetryTest.java
```diff
@@ -35,40 +33,39 @@ public abstract class BaseOpenTelemetryTest {
...
+                        "permissions.xml");
     }
 
-    @BeforeClass
-    public static void checkForDocker() {
-        AssumeTestGroupUtil.assumeDockerAvailable();
+    @ContainerResource
+    ManagementClient managementClient;
+    @ContainerResource
+    ServerManager serverManager;
```
jamezp, line 66: Not a huge deal, but I don't see these being used anywhere.

jamezp (changes requested):
Most of these are minor comments. I do think we need to add the copyright header though.

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
jasondlee: I thought that was how @pferraro said to handle that, though I can't find that discussion. Do you remember, Paul?

microprofile/telemetry-smallrye/extension/pom.xml
bstansberry: Use ${full.maven.groupId}.  Always use that or ${ee.maven.groupId} for artifacts coming out of this project's build. Which depends on whether the artifact is provisioned via wildfly or wildfly-ee feature pack.

bstansberry: Please apply my comment ^^^ throughout the PR.

testsuite/integration/microprofile/pom.xml
```diff
@@ -391,6 +391,13 @@
             <artifactId>wildfly-naming-client</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian</groupId>
+            <artifactId>arquillian-testcontainers</artifactId>
+            <version>1.0.0.Alpha1</version>
```
jamezp, line 398: FWIW I think for now this is fine while we test it, but this should move to the test-bom at some point.

jamezp (approved):
From a testing point of view this looks good to me. Once @bstansberry approves the changes he's asked for.

testsuite/integration/microprofile/pom.xml
jasondlee: Agreed, but standard-test/pom.xml just has the Arquillian BOM, so maybe this should be added there, at which point the version HERE could be removed.

jamezp (approved):

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
bstansberry: @jasondlee I see you added a ServiceName. Adding a couple lines to MicroProfileTelemetryDependencyProcessor.deploy will ensure the service is started before MicroProfileTelemetryDeploymentProcessor tries to read the value.

```
    @Override
    public void deploy(DeploymentPhaseContext phaseContext) {
        addDependencies(phaseContext.getDeploymentUnit());

        // Ensure the OpenTelemetryConfig is available before the next phase DeploymentUnitPhaseService starts
        phaseContext.requires(ServiceDependency.on(ServiceName.parse(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR.getName())));
    }
```

jasondlee: Added, building, and testing...

jasondlee: All clear locally. Pushed.

bstansberry (approved):
My concerns are all addressed.

Note that yesterday I did a read-resource-description for this subsystem with a build a main and a build of a branch with this merged and I saw no differences.

PR #17882, opened 2024-05-10
[32.x][WFLY-12329] managed executor service's execute(), submit() and schedule*() should use RequestController#forceBeginRequest() only when starting

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-12329
Main branch PR: https://github.com/wildfly/wildfly/pull/17881

____


bstansberry:
I don't think this is a good choice for a micro. Even if there were no concerns about the solution (I have some; see the other PR), I think a behavior change like this should bake a while in main.

emmartins:
closing this one since 32.0.1 is released

PR #17885, opened 2024-05-14
[WFLY-19331] Correct the product.docs.server.version

bstansberry:
Roughly related to https://issues.redhat.com/browse/WFLY-19331, which was actually fixed via https://github.com/wildfly/wildfly.github.io/pull/76.

We don't regenerate docs for micros, so the change here shouldn't matter, but it's better to have this correct.

Also includes the fix for https://issues.redhat.com/browse/WFLY-19341, which is required for CI to pass with this change.

Upstream: #17886

____

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
         <ee.maven.groupId>${project.groupId}</ee.maven.groupId>
         <ee.maven.version>${project.version}</ee.maven.version>
         <channels.maven.groupId>${project.groupId}.channels</channels.maven.groupId>
-        <product.docs.server.version>31</product.docs.server.version>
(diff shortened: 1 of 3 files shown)


asoldano (approved):

asoldano:
I've seen TckParticipantTests.testNonJaxRsCompletionStageResponseAndParticipantStatus fail other times on both main and 32.x, it looks unrelated to the changes here.
https://ci.wildfly.org/test/4458590039331980043?currentProjectId=WF_PullRequest&expandTestHistoryChartSection=true&page=1
https://ci.wildfly.org/test/4458590039331980043?currentProjectId=WF_PullRequest32x&expandTestHistoryChartSection=true

PR #17892, opened 2024-05-16
[WFLY-19341] Make the testsuite/galleon tests appropriately scoped

bstansberry:
https://issues.redhat.com/browse/WFLY-19341

____

### DIFF
--- a/testsuite/galleon/update/pom.xml
+++ b/testsuite/galleon/update/pom.xml
@@ -23,6 +23,15 @@
     <packaging>pom</packaging>
     <name>WildFly Test Suite: Galleon update tests</name>
     <description>Install a base version and check that wildfly producers can get updated to the SNAPSHOT version.</description>
+
+    <properties>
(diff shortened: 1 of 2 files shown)


testsuite/galleon/pom.xml
```diff
@@ -22,7 +22,7 @@
     <artifactId>wildfly-ts-galleon-parent</artifactId>
     <packaging>pom</packaging>
     <name>WildFly Test Suite: Galleon tests (parent)</name>
-    <modules>
-        <module>update</module>
-    </modules>
+            <modules>
```
jfdenise, line 25: Not sure that this formatting is required.

jfdenise (approved):

testsuite/galleon/pom.xml
bstansberry: Thanks @jfdenise -- that's the result of my taking an approach to something and then abandoning it incompletely. I'll correct that and merge this.

bstansberry:
Thanks @jfdenise

PR #17927, opened 2024-06-03
[WFLY-19376] Upgrade WildScribe 3.0.0.Final

jmesnil:
* Generate WildFly management model from its experimental stability level
* Display Ⓟ, Ⓒ, Ⓔ  characters next to resp. preview, community, and  experimental in the generated HTML pages

This fixes https://issues.redhat.com/browse/WFLY-19376

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -124,6 +124,7 @@
                     <display-name>${full.dist.product.release.name}</display-name>
                     <display-version>${product.docs.server.version}</display-version>
                     <site-dir>${project.build.directory}/generated-docs/wildscribe</site-dir>
+                    <stability>experimental</stability>
                     <required-extensions>
                         <required-extension>org.wildfly.extension.rts</required-extension>
                         <required-extension>org.jboss.as.xts</required-extension>
--- a/pom.xml
+++ b/pom.xml
@@ -318,7 +318,7 @@
         <version.org.jacoco>0.8.12</version.org.jacoco>
         <version.org.jboss.galleon>6.0.0.Final</version.org.jboss.galleon>
         <version.org.wildfly.glow>1.0.0.Final</version.org.wildfly.glow>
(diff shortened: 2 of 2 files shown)


jamezp (approved):

bstansberry:
This requires a big jump in the minimum maven version:

```
[ERROR] Failed to execute goal org.jboss.wildscribe:wildscribe-maven-plugin:3.0.0.Final:generate-site (generate-site) on project docs: The plugin org.jboss.wildscribe:wildscribe-maven-plugin:3.0.0.Final requires Maven version 3.9.4
```

Currently it's

```
        <!--
            Build configuration.  Override JBoss Parent settings as necessary.
            For example: <version.surefire.plugin>2.11</version.surefire.plugin>
          -->
        <maven.min.version>3.6.0</maven.min.version>
```

We may want to do that (I know it's come up a few times) but we should discuss.

Whatever our min is we should probably find a way to enforce that it works in our various plugins so we don't end up with a plugin requiring an update that we're not willing to accept, or perhaps not detecting that our min has increased. (I got lucky detecting this by building the docs on a machine with an old maven version.)

bstansberry:
Actually, it wasn't a very old maven version; it's 3.9.1.

jamezp:
Good catch. It shouldn't actually require that new of a version, there might just be a dependency marked incorrectly or some property set wrong. I've seen this before. I'll have a look.

jamezp:
This should fix it https://github.com/wildscribe/wildscribe/pull/68. The question becomes, do we want to do a 3.0.1, or go ahead and include some extra changes and do a 3.1.0?

bstansberry:
@jamezp @jmesnil Looking at the open issues and PRs in https://github.com/wildscribe/wildscribe I think getting 3.0.1 out and integrated makes sense. Other than the parent pom bump the other items don't sound like things that are certain to get done quickly.

bstansberry:
Superseded by #17991

PR #17934, opened 2024-06-05
WIP: Add Hibernate Search jakarta-batch integration module to Wildfly preview

marko-bekhta:
Hi @scottmarlow 👋🏻 😃 

We are looking to add a new Hibernate Search preview module to Wildfly. It is for [Jakarta Batch integration](https://docs.jboss.org/hibernate/stable/search/reference/en-US/html_single/#jsr352-integration). 

Hibernate Search offers `hibernate-search-mapper-orm-jakarta-batch-jberet` + `hibernate-search-mapper-orm-jakarta-batch-core` artifacts, that provide this integration. `batch-core` contains the job XML and job implementation details, while `batch-jberet` has JBeret-specific things (e.g. job resolver, cdi stuff etc.).

I've created a preview module and a simple test to check how things work on the Wildfly side and whether we need any other updates to be applied on the Search side.

There are a few problems I've encountered while working on this:
- Wildfly discovers the `org.jberet.spi.JobXmlResolver` from the Search jar with no problem, and the mass indexing job is visible to `jobOperator.start(MassIndexingJob.NAME, jobProps)` but then it fails to find the actual XML packaged in the Search jar. I've tried adding this new `org.hibernate.search.jakarta.batch` and various combinations of `services="import" export="true" etc` in search/jberet/batch Wildfly modules but none helped to make the job XML to be found. 😕 
- `batch-jberet` jar provides an application-scoped bean that should be used by the mass indexing job, but similarly to the problem with the XML, this bean is not getting injected 😕 

I was hoping that maybe you'd have some ideas on how these ^ can be resolved or any pointers to where I should look. Thanks! 😃 

This work is based on a snapshot build of Hibernate Search based on this branch https://github.com/marko-bekhta/hibernate-search/tree/fix/batch, as some changes to Search are required.

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/WildFlyJobXmlResolver.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/WildFlyJobXmlResolver.java
@@ -126,7 +126,12 @@ public class WildFlyJobXmlResolver implements JobXmlResolver {
             return null;
         }
         for (JobXmlResolver resolver : jobXmlResolvers) {
-            final InputStream in = resolver.resolveJobXml(jobXml, classLoader);
+            InputStream in = resolver.resolveJobXml(jobXml, classLoader);
+            if (in == null) {
+                // Fallback: the resolver's resources may be in a module jar
+                // not visible through the deployment classloader
+                in = resolver.resolveJobXml(jobXml, resolver.getClass().getClassLoader());
+            }
             if (in != null) {
                 return in;
             }
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
@@ -24,5 +24,6 @@
         <module name="org.hibernate.search.mapper.pojo" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm.outboxpolling" optional="true" services="import"/>
+        <module name="org.hibernate.search.jakarta.batch" optional="true" services="import" export="true"/>
     </dependencies>
 </module>
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
@@ -21,6 +21,7 @@
         <!-- Optional dependencies to import services (used for resolution of bean references in particular). -->
         <module name="org.hibernate.search.backend.lucene" optional="true" services="import"/>
         <module name="org.hibernate.search.backend.elasticsearch" optional="true" services="import"/>
+        <module name="org.hibernate.search.jakarta.batch" optional="true" services="import" export="true"/>
         <module name="org.hibernate.search.mapper.pojo" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm.outboxpolling" optional="true" services="import"/>
--- a/pom.xml
+++ b/pom.xml
@@ -515,7 +515,7 @@
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.1</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
         <version.org.hibernate.commons.annotations>6.0.6.Final</version.org.hibernate.commons.annotations>
         <version.org.hibernate>6.4.4.Final</version.org.hibernate>
-        <version.org.hibernate.search>7.1.1.Final</version.org.hibernate.search>
+        <version.org.hibernate.search>7.1.2-SNAPSHOT</version.org.hibernate.search>
         <version.org.hibernate.validator>8.0.1.Final</version.org.hibernate.validator>
         <version.org.hornetq>2.4.9.Final</version.org.hornetq>
         <version.org.infinispan>14.0.28.Final</version.org.infinispan>
--- a/testsuite/preview/basic/src/test/java/org/wildfly/test/preview/hibernate/search/batch/HibernateSearchBatchTestCase.java
+++ b/testsuite/preview/basic/src/test/java/org/wildfly/test/preview/hibernate/search/batch/HibernateSearchBatchTestCase.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.preview.hibernate.search.batch;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+import org.hibernate.search.jakarta.batch.core.massindexing.MassIndexingJob;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.test.shared.TimeoutUtil;
+import org.jboss.as.test.shared.util.AssumeTestGroupUtil;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.Asset;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.shrinkwrap.descriptor.api.Descriptors;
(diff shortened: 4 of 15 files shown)


bstansberry:
@liweinan FYI

scottmarlow:
Hi @marko-bekhta, sorry that I missed your ping on this issue earlier this week.  :(

I started building https://github.com/marko-bekhta/hibernate-search/tree/fix/batch locally and will try to look at building the https://github.com/marko-bekhta/wildfly/tree/feat/hibernate-search-jakarta-batch change soon.

scottmarlow:
`mvn clean install -Dtest=org.wildfly.test.preview.hibernate.search.batch.HibernateSearchBatchTestCase` passed for me locally which is a good sign.

marko-bekhta:
hey 😃  no worries 😃  Thanks for taking a look at it!

> `mvn clean install -Dtest=org.wildfly.test.preview.hibernate.search.batch.HibernateSearchBatchTestCase` passed for me locally which is a good sign.

hmm that's strange 😕 does it work for you even if the 
```java
.addAsResource(jobXML(), "META-INF/batch-jobs/hibernate-search-mass-indexing.xml")
```
line is removed (since the user shouldn't really supply that XML as it is shipped with the Search jar)?

pom.xml
```diff
@@ -515,7 +515,7 @@
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.1</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
         <version.org.hibernate.commons.annotations>6.0.6.Final</version.org.hibernate.commons.annotations>
         <version.org.hibernate>6.4.4.Final</version.org.hibernate>
-        <version.org.hibernate.search>7.1.1.Final</version.org.hibernate.search>
+        <version.org.hibernate.search>7.1.2-SNAPSHOT</version.org.hibernate.search>
```
scottmarlow, line 518: Will there be a alpha or beta 7.1.2 that could be used instead of snapshot?

scottmarlow:
> hey 😃 no worries 😃 Thanks for taking a look at it!
> 
> > `mvn clean install -Dtest=org.wildfly.test.preview.hibernate.search.batch.HibernateSearchBatchTestCase` passed for me locally which is a good sign.
> 
> hmm that's strange 😕 does it work for you even if the
> 
> ```java
> .addAsResource(jobXML(), "META-INF/batch-jobs/hibernate-search-mass-indexing.xml")
> ```
> 
> line is removed (since the user shouldn't really supply that XML as it is shipped with the Search jar)?

I'll try that soon.

scottmarlow:
I tried running the test again but this time from the correct folder.  The test fails right away.  Steps that I followed:
1. build wildfly
2. cd testsuite/preview/basic
3. mvn clean install -Dtest=org.wildfly.test.preview.hibernate.search.batch.HibernateSearchBatchTestCase

^ fails with:
> Caused by: java.lang.RuntimeException: The process 95636 is no longer active.
        at org.wildfly.plugin.tools.server.AbstractServerManager.waitFor(AbstractServerManager.java:70)
        at org.wildfly.plugin.tools.server.StandaloneManager.waitFor(StandaloneManager.java:23)
        at org.jboss.as.arquillian.container.CommonManagedDeployableContainer.startInternal(CommonManagedDeployableContainer.java:90

The testsuite/preview/basic/target/wildfly/standalone/log/server.log contains
2024-06-11 15:13:20,248 INFO  [org.jboss.modules] (main) JBoss Modules version 2.1.5.Final
2024-06-11 15:13:20,406 FATAL [org.jboss.as.server] (main) WFLYSRV0239: Aborting with exit code 1

liweinan:
Thanks for checking this Scott!

marko-bekhta:
Hey Scott, thanks for spending time on this one!

> ^ fails with:
> >Caused by: java.lang.RuntimeException: The process 95636 is no longer active.

I've noticed this too... even if I try running any other tests in that preview module even from the main branch (without 
 any changes from this patch) e.g.:

```bash
mvn clean install -Players_.module.profile -pl :wildfly-ts-preview-basic -Dtest=org.wildfly.test.preview.mvc.MVCKrazoSmokeTestCase
```
I get a similar error (just a different process ID)... so I've been building Wildfly, skipping the tests, and then running a particular test from an IDE, which surprisingly works 🤷🏻‍♂️😖.

with your second approach the logs you've shared are similar to what I get from the IDE, and that's where I got stuck:

>There are a few problems I've encountered while working on this:
>
>Wildfly discovers the org.jberet.spi.JobXmlResolver from the Search jar with no problem, and the mass indexing job is visible to jobOperator.start(MassIndexingJob.NAME, jobProps) but then it fails to find the actual XML packaged in the Search jar. I've tried adding this new org.hibernate.search.jakarta.batch and various combinations of services="import" export="true" etc in search/jberet/batch Wildfly modules but none helped to make the job XML to be found. 😕
>
> batch-jberet jar provides an application-scoped bean that should be used by the mass indexing job, but similarly to the problem with the XML, this bean is not getting injected 😕 

there's a line in the logs:
```
HSEARCH500008: No entity manager factory has been created yet. Make sure that the entity manager factory has already been created and that the entity manager factory hasn't been closed.
```
which means that the fallback `ActiveSessionFactoryRegistry`  registry is used instead of the expected `JBeretEntityManagerFactoryRegistry`... 
`JBeretEntityManagerFactoryRegistry` is included in the `hibernate-search-mapper-orm-jakarta-batch-jberet` jar and is annotated as an application-scoped bean. The expectation is that this bean would be available through CDI and will get injected into a job step... but as that doesn't happen, I suspect that the module definition XML is missing something or maybe there's some other way to make the beans from a module accessible?

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/hibernate/search/batch/HibernateSearchBatchTestCase.java
```diff
@@ -0,0 +1,239 @@
...
+                .addClasses(SearchBean.class, IndexedEntity.class, TimeoutUtil.class)
+                .addAsResource(manifest(), "META-INF/MANIFEST.MF").addAsResource(persistenceXml(), "META-INF/persistence.xml")
+                // we shouldn't really need this xml added to the war, since it is packaged with Search,
+                // but: WFLYBATCH000006: Could not find the job XML file in the deployment: hibernate-search-mass-indexing
+                .addAsResource(jobXML(), "META-INF/batch-jobs/hibernate-search-mass-indexing.xml")
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    private static Asset jobXML() {
+        return new StringAsset("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" + "<!--\n"
```
rhusar, line 72: This isn't readable. Can you make this a regular file instead?

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/search/engine/main/module.xml
```diff
@@ -24,5 +24,6 @@
         <module name="org.hibernate.search.mapper.pojo" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm" optional="true" services="import"/>
         <module name="org.hibernate.search.mapper.orm.outboxpolling" optional="true" services="import"/>
+        <module name="org.hibernate.search.jakarta.batch" optional="true" services="import" export="true"/>
```
rhusar, line 27: Let's keep these sorted alphabetically...

rhusar (changes requested):

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/WildFlyJobXmlResolver.java
```diff
@@ -126,7 +126,12 @@ public InputStream resolveJobXml(final String jobXml, final ClassLoader classLoa
             return null;
         }
         for (JobXmlResolver resolver : jobXmlResolvers) {
-            final InputStream in = resolver.resolveJobXml(jobXml, classLoader);
+            InputStream in = resolver.resolveJobXml(jobXml, classLoader);
+            if (in == null) {
+                // Fallback: the resolver's resources may be in a module jar
+                // not visible through the deployment classloader
+                in = resolver.resolveJobXml(jobXml, resolver.getClass().getClassLoader());
+            }
```
marko-bekhta, line 134: So this was the missing bit to make Jakarta Batch / Hibernate Search integration work...

Since Hibernate Search ships its own job XML that lives in its resources, trying to load it through a deployment CL doesn't really work; hence the fallback to checking the module in which the job is defined.

marko-bekhta: hey @scottmarlow 👋🏻 🙂 
just pinging you to get this one back on your radar 🙂

scottmarlow: Looks like test results are still pending...

marko-bekhta: hm, build seems to be 🟢: https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk17/579203 maybe it's just a GH status reporting issue ?

PR #17939, opened 2024-06-07
[WFLY-19136] Add test for testing the example configurations

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-19136


bstansberry:
@pferraro or @rhusar Please review.

rhusar (approved):
LGTM! Thanks @yersan 

I was going to suggest using a distributable deployment as well so that the channel is actually started – but that would fail when starting the channel since we can't reasonably provide a usable credential set here. So this is the best we could do without having real credentials or using a mock cloud service.

pferraro (approved):

bstansberry:
Thanks @yersan, @rhusar and @pferraro

PR #17961, opened 2024-06-14
WFLY-19425: update model definition and bump schema

michpetrov:
Issue: [WFLY-19425](https://issues.redhat.com/browse/WFLY-19425)

The major change here is changing the `option` element in xsd to allow text content (since that's how we write down the key-value pair), and some smaller changes dealing with default values and wrongly declared minimal occurrences.

The second commit is just refactoring the schema since there are some needless repetitions, ideally the `distributed-workmanager` would also just extend `workmanager` but that's not possible without imposing an order on the child elements and that would also require a change in the writer (although I guess the writer always has some implicit order to it, so maybe we should never be using `xs:all`). I will squash it into one commit but wanted to keep it separate for easier comparison.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
@@ -97,6 +97,7 @@ public class JcaExtension implements Extension {
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_4_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_5_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
         context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_6_0.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
+        context.setSubsystemXmlMapping(SUBSYSTEM_NAME, Namespace.JCA_6_1.getUriString(), () -> ConnectorSubsystemParser.INSTANCE);
     }
 
     static final class ConnectorSubsystemParser implements XMLStreamConstants, XMLElementReader<List<ModelNode>>,
@@ -309,6 +310,7 @@ public class JcaExtension implements Extension {
             while (reader.hasNext() && reader.nextTag() != END_ELEMENT) {
                 Namespace readerNs = Namespace.forUri(reader.getNamespaceURI());
                 switch (readerNs) {
+                    case JCA_6_1:
                     case JCA_6_0:
                     case JCA_5_0:
                     case JCA_4_0:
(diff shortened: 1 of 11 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
```diff
@@ -63,7 +63,7 @@ public class JcaExtension implements Extension {
 
     public static final String SUBSYSTEM_NAME = "jca";
 
-    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(6, 0, 0);
+    private static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(6, 1, 0);
```
pferraro, line 66: This jira appears to be specific to XML parsing.  For what reason are you incrementing the model version?

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaExtension.java
michpetrov: @tadamski are the schema and model version not the same thing? I'm increasing the schema version, same as #17866

pferraro: They are not the same thing.
One can change the subsystem model without affecting the XML schema (like you are doing here), just as one can change the XML schema without affecting the subsystem model.

bstansberry:
@michpetrov Please DO NOT squash the two commits. It's much easier to understand the history when we have a commit that does the mechanics of introducing a new version and then a separate one that shows the functional change. For example your second commit makes it easy to see why there is a new xsd from within git; no need to diff the xsd to the previous one.

Please do add WFLY-19425 to the 2nd commit's message.

michpetrov:
alright, undid the model version change and changed the commit message

bstansberry:
@michpetrov Can you move all the functional changes to the second commit please? I've been trying to understand what the change is here and the only way to tell without checking out the code and doing manual diffs is to read your PR description.

Best practice for anything that increments a management API version or introduces a new schema version is to do a commit that does all the boilerplate stuff to add the new version / xsd file, but where the new stuff is no different from the old. So in this case the wildfly-jca_6_1.xsd would be no different from wildfly-jca_6_0.xsd. Then a subsequent commit or commits bring in the actual changes, which are then obvious from reading the commit diff. This makes the overall PR easy to review and makes the history much more understandable for future maintainers.

What you have here is mostly like that, except the xsd changes are partly in the first commit and partly in the second.

michpetrov:
@bstansberry done

connector/src/main/resources/schema/wildfly-jca_6_1.xsd
```diff
@@ -239,9 +160,35 @@
...
+    </xs:element>
+    <xs:element name="elytron-enabled" type="xs:boolean" default="true">
+        <xs:annotation>
+            <xs:documentation>
+                <![CDATA[[
+                Indicates that Elytron is responsible for security for this workmanager. Default is true
+                ]]>
+            </xs:documentation>
+        </xs:annotation>
+    </xs:element>
```
bstansberry, line 187: @michpetrov This means a document containing one or more of these as root elements is valid. It also means one of these would be a legal child of a 'profile' element in standalone/domain/host.xml. Which of course would fail to parse.

michpetrov: @bstansberry I've reverted it, unfortunately this is the only way xsd lets you reuse elements

PR #17962, opened 2024-06-17
WFLY-19419 Distributed timer service should consolidate timeouts that would execute in the past

pferraro:
https://issues.redhat.com/browse/WFLY-19419

Also solves: WFLY-19271 Distributed timer service drops timeout events if server is suspended
https://issues.redhat.com/browse/WFLY-19271

____

> Wildfly issue links:
> * [WFLY-19514](https://issues.redhat.com/browse/WFLY-19514)

### DIFF
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
@@ -138,7 +138,7 @@ public class TimerScheduler<I, V, C> extends AbstractCacheEntryScheduler<I, Immu
 
                         // Safeguard : ensure timeout was not already triggered elsewhere
                         if (currentTimeoutReference.isEmpty()) {
-                            InfinispanEjbLogger.ROOT_LOGGER.debugf("Unexpected timeout event triggered.", id);
+                            InfinispanEjbLogger.ROOT_LOGGER.debugf("Unexpected timeout event triggered for %s", id);
                             return false;
                         }
                         Instant currentTimeout = currentTimeoutReference.get();
@@ -149,9 +149,9 @@ public class TimerScheduler<I, V, C> extends AbstractCacheEntryScheduler<I, Immu
 
                         Timer<I> timer = factory.createTimer(id, metaData, manager, scheduler);
 
-                        InfinispanEjbLogger.ROOT_LOGGER.debugf("Triggering timeout for timer %s [%s]", id, timer.getMetaData().getContext());
+                        InfinispanEjbLogger.ROOT_LOGGER.debugf("Triggering timeout for timer %s @ %s", id, currentTimeout);
 
-                        // In case we need to reset the last timeout
+                        // Capture original last timeout in case we need to reset it
                         Optional<Instant> lastTimeout = metaData.getLastTimeout();
                         // Record last timeout - expected to be set prior to triggering timeout
                         metaData.setLastTimeout(currentTimeout);
@@ -166,6 +166,7 @@ public class TimerScheduler<I, V, C> extends AbstractCacheEntryScheduler<I, Immu
                             InfinispanEjbLogger.ROOT_LOGGER.debugf("EJB component is suspended - could not invoke timeout for timer %s", id);
                             // Reset last timeout
                             metaData.setLastTimeout(lastTimeout.orElse(null));
+                            batch.discard();
                             return false;
                         }
 
--- a/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
+++ b/clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
@@ -141,43 +141,50 @@ public class TimerScheduler<I, V, C> extends AbstractCacheEntryScheduler<I, Immu
                             InfinispanEjbLogger.ROOT_LOGGER.debugf("Unexpected timeout event triggered for %s", id);
                             return false;
                         }
+
+                        Instant now = Instant.now();
                         Instant currentTimeout = currentTimeoutReference.get();
-                        if (currentTimeout.isAfter(Instant.now())) {
-                            InfinispanEjbLogger.ROOT_LOGGER.debugf("Timeout for timer %s initiated prematurely.", id);
+                        if (currentTimeout.isAfter(now)) {
+                            InfinispanEjbLogger.ROOT_LOGGER.debugf("Timeout for timer %s initiated prematurely @ %s", id, currentTimeout);
                             return false;
                         }
 
-                        Timer<I> timer = factory.createTimer(id, metaData, manager, scheduler);
-
-                        InfinispanEjbLogger.ROOT_LOGGER.debugf("Triggering timeout for timer %s @ %s", id, currentTimeout);
-
-                        // Capture original last timeout in case we need to reset it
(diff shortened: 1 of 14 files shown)


bstansberry:
@tadamski @rachmatowicz Please review.

pferraro:
Adding the hold label for now, as there are few things requiring further discussion:
* Yesterday, I discovered that, based on the implementations [here](https://github.com/wildfly/wildfly/blob/32.0.1.Final/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerTask.java#L202) vs [here](https://github.com/wildfly/wildfly/blob/32.0.1.Final/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/CalendarTimerTask.java#L48), there seems to be a discrepancy in behavior of missed timeouts between interval vs ScheduleExpression-based timers.  While the old timer service implementation coalesces missed timeouts into a single event for interval timers, it does not appear to do the same for ScheduleExpression-based timers.  If this is indeed the case, while the proposed changes would bring the behavior of the distributed timer service in line with the old timer service wrt interval timers, it would cause the same behavior wrt ScheduleExpression-timers to diverge.
* @rachmatowicz suggested that we expose a management attribute to allow users to enable/disable the coalescing of missed timeouts, which I thought was a good idea.  If the previous point is valid, then we might actually want to consider an attribute per timer type - which would offer the greatest amount of flexibility wrt compatibility.

pferraro:
Removing hold label.

I've included a reproducer and fix for [WFLY-19514](https://issues.redhat.com/browse/WFLY-19514) to address the issue from my previous comment, where interval vs calendar timers behave differently wrt missed timeouts.  With this change, both timer services will behave consistently for both interval and calendar-based timers.
I've logged a feature request (https://issues.redhat.com/browse/WFLY-19515) to expose ejb3 subsystem configuration for enabling/disabling the coalescing of missed timeouts, as well as deployment descriptor support to enable/disable this behavior per timer service.

rachmatowicz:
@bstansberry @pferraro FYI, actively working on the review. Should be done Friday 12 July.

bstansberry:
I'm going to defer this to 34.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
```diff
@@ -138,45 +138,53 @@ public Boolean call() throws Exception {
...
-
-                        // If timeout callback canceled this timer
-                        if (timer.isCanceled()) {
-                            InfinispanEjbLogger.ROOT_LOGGER.debugf("Timeout callback canceled timer %s", id);
-                            return true;
-                        }
-
                         // Determine next timeout
                         Optional<Instant> nextTimeout = metaData.getNextTimeout();
+                        // WFLY-19361: If next timeout is in the past do not trigger event
```
rachmatowicz, line 159: I have to confess, I don't see how this code works to fire one event to represent a possible sequence of events that may have fired while the timer service is stopped. We have a timer with a getLastTimeout() setting and a getNextTimeout() setting and we have the time now. In line 147, if the next timeout is after now (in the future), don't fire prematurely - agreed. In line 160, if the next timeout is in the past, don't fire anything - not sure how this represents a set of possible firings while the timer service was stopped. In line 163, fire an event if the next firing is exactly equal to the time now - unlikely to be the case. There must be something else going on here, for example in the way getNextTimeout() is calculated? I'm thinking of the case where a single interval timer was created which fires every second, the timer fires once (#1), the timer service is offline for 3 seconds (#2, #3, #4) and then comes back on line, and is now ready to fire again (#5). I'm expecting to see three timeout method invocations - one for #1, one for #2+#3+#4 (to represent those firings which happened when the timer service was offline) and one for #5.   This is why line 160 puzzles me.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
rachmatowicz, line 1: This is complex code. It would help to provide some explanatory comments as to what is going on here, both on the class and the methods (where appropriate), particularly with the InvokeTask.

rachmatowicz (commented):
I can see that the tests pass...so my understanding is just not deep enough at this stage to fully digest everything that is going on, but left some comments for consideration.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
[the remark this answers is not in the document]
pferraro: This is because neither timer service properly handles server suspension.  Both implementations schedule timers to execute, but then skip the application callback when suspended resulting in the timer being rescheduled immediately.  This will be addressed separately by a later fix.

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
pferraro: This works because this logic is only used to determine the _next_ timeout.  Remember that the "last timeout" identifies the last _successful_ timeout.  So, in your example, when the server comes back online, the value for the last timeout will identify execution \#1.  The timer service will schedule the timer to fire per \#2.  When \#2 is successful, then next scheduled event will correspond to \#5.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/ejb/timer/CoalesceMissedTimeoutsTestCase.java
```diff
@@ -0,0 +1,151 @@
...
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
```
bstansberry, line 21: Wrong header

pferraro: Whoops.  Fixing...

pferraro: Fixed

clustering/ejb/infinispan/src/main/java/org/wildfly/clustering/ejb/infinispan/timer/TimerScheduler.java
pferraro: This method is already heavily annotated with comments.

rachmatowicz (approved):
Met with Paul and discussed how LocalScheduler.run() and TimerScheduler.InvokeTask() work together to schedule and execute timeouts in such a way that any timeout callbacks that occur when the TimerService is not started are coalesced into one representative timer firing when the TimerService is restarted, so  this looks good to me.

pferraro:
Thanks @rachmatowicz !

PR #17991, opened 2024-06-28
[WFLY-19376] Upgrade Wildscribe to 3.1.0.Final.

jamezp:
* Generate WildFly management model from its experimental stability level
* Display Ⓟ, Ⓒ, Ⓔ  characters next to resp. preview, community, and  experimental in the generated HTML pages

This fixes https://issues.redhat.com/browse/WFLY-19376

This is a replacement for #17927 only if we want it in sooner.

Note this is an upgrade to 3.1.0.Final instead of 3.0.0.Final. If we'd rather have a 3.0.1.Final, please let me know.


bstansberry (approved):

bstansberry:
Thanks @jmesnil and @jamezp

PR #17992, opened 2024-06-28
WFLY-19473: fix and bump schema

michpetrov:
Issue: [WFLY-19473](https://issues.redhat.com/browse/WFLY-19473)

This is the last one, the issue here is that certain elements in the schema were configured as mutually exclusive - specifically `elytron-enabled` is not mutually exclusive with anything according to the model but was declared as such in both `security` and `recovery`. Some elements had to be reordered as a result.

`elytron-enabled` was also changed from booleanPresence to boolean, but we should just pick one (or they could be attributes) since it looks randomly assigned, e.g. in `xa-pool` we've got
* is-same-rm-override (boolean)
* interleaving (boolean-presenceType)
* no-tx-separate-pools (boolean-presenceType)
* pad-xid (boolean)
* wrap-xa-resource (boolean)
though I imagine this pops in other subsytems too.

In the parser refactor the `pool-fair` attribute was never being written in a non-xa pool, so that is also fixed. I noticed most of the pool attributes are shared with datasources so maybe there should be another refactor of the whole "wildfly/connector" so that both the RA schema and parser import from datasources.

____

> Wildfly issue links:
> * [WFLY-19573](https://issues.redhat.com/browse/WFLY-19573)

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Namespace.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Namespace.java
@@ -34,12 +34,14 @@ public enum Namespace {
 
     RESOURCEADAPTERS_7_0("urn:jboss:domain:resource-adapters:7.0"),
 
-    RESOURCEADAPTERS_7_1("urn:jboss:domain:resource-adapters:7.1");
+    RESOURCEADAPTERS_7_1("urn:jboss:domain:resource-adapters:7.1"),
+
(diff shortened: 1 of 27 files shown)


jamezp:
I think the second commit here has the wrong JIRA id attached. [WFLY-19573](https://issues.redhat.com/browse/WFLY-19573)  was just created today and is about CDI.

michpetrov:
@jamezp yeah, thanks, that was a typo

PR #18016, opened 2024-07-05
[WFLY-19504] Remove MyFaces-related test from JSFDeploymentProcessorTestCase

jasondlee:
https://issues.redhat.com/browse/WFLY-19504

Remove obsolete tests
Clean up remaining code

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/version/JSFDeploymentProcessorTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jsf/version/JSFDeploymentProcessorTestCase.java
@@ -4,47 +4,33 @@
  */
 package org.jboss.as.test.integration.jsf.version;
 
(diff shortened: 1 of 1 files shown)


bstansberry:
@jasondlee Do we have the necessary coverage in the myfaces FP code?

jasondlee:
> @jasondlee Do we have the necessary coverage in the myfaces FP code?

@bstansberry There is *some* testing: https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/testsuite/subsystem/src/test/java/org/wildfly/extras/jsf/myfaces/test/subsystem/SubsystemSanityTestCase.java

That seems a bit light, but I'm not sure off hand how much more we need. I imagine running many of the same JSF tests from the WF testsuites could be useful there, but I have not given that much thought of late, tbh. I've filed an issue on that component to track the question: https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/issues/11

bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #18017, opened 2024-07-06
[WFLY-19287] Fail boot if EE 11 APIs are used and the security manage…

bstansberry:
…r is enabled

https://issues.redhat.com/browse/WFLY-19287


darranl (approved):

bstansberry:
Thanks @darranl

PR #18020, opened 2024-07-08
[WFLY-19339] If Naming subsystem is activated then add both jdk.naming.dns & jdk.naming.rmi dependencies to deployments

ropalka:
https://issues.redhat.com/browse/WFLY-19339

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
+++ b/naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.naming.deployment;
+
+import org.jboss.as.naming.logging.NamingLogger;
+import org.jboss.as.server.deployment.DeploymentPhaseContext;
+import org.jboss.as.server.deployment.DeploymentUnit;
+import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
+import org.jboss.as.server.deployment.DeploymentUnitProcessor;
+import org.jboss.as.server.deployment.Attachments;
+import org.jboss.as.server.deployment.module.ModuleDependency;
+import org.jboss.as.server.deployment.module.ModuleSpecification;
+import org.jboss.modules.Module;
+import org.jboss.modules.ModuleLoadException;
+import org.jboss.modules.ModuleLoader;
+
+/**
+ * A processor which adds jdk naming modules to deployments.
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
+ */
+public final class JdkDependenciesProcessor implements DeploymentUnitProcessor {
+
+    private static final String[] JDK_NAMING_MODULES = {
+            "jdk.naming.dns",
+            "jdk.naming.rmi"
+    };
+
+    @Override
+    public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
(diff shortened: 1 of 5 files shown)


naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+        final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
+        final ModuleSpecification moduleSpecification = deploymentUnit.getAttachment(Attachments.MODULE_SPECIFICATION);
+        final ModuleLoader moduleLoader = Module.getBootModuleLoader();
+        for (String moduleName : JDK_NAMING_MODULES) {
+            try {
+                moduleLoader.loadModule(moduleName);
+                moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, moduleName, false, false, false, false));
```
bstansberry, line 38: FYI:

https://issues.redhat.com/browse/WFCORE-6889

naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+            } catch (ModuleLoadException ex) {
+                NamingLogger.ROOT_LOGGER.debugf("Module not found: %s", moduleName);
```
bstansberry, line 40: I was thinking maybe log this at at INFO (with an i18n message) but I think a user would have to have gone far out of their way to have one of these modules be unavailable. So debug seems ok.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/naming/dns/DnsContextLookupBean.java
```diff
@@ -0,0 +1,28 @@
...
+@Stateless
+public class DnsContextLookupBean {
+
+    public void testDnsContextLookup(String serverAddress) throws Exception {
+        final Hashtable<String, String> env = new Hashtable<>();
+        env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.dns.DnsContextFactory");
+        env.put(Context.PROVIDER_URL, "dns://" + serverAddress);
+
+        Context ctx = new InitialContext(env);
+        ctx.close();
```
bstansberry, line 26: The fails with the SecurityManager enabled; needs some permission.

bstansberry (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/naming/dns/DnsContextLookupBean.java
ropalka: Missing permission was added.

naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
ropalka: OK

naming/src/main/java/org/jboss/as/naming/deployment/JdkDependenciesProcessor.java
ropalka: OK

bstansberry (approved):

bstansberry:
Thanks @ropalka

PR #18039, opened 2024-07-11
[WFLY-19529] Upgrade the OWASP dependency-check plugin to 10.0.2

darranl:
https://issues.redhat.com/browse/WFLY-19529


darranl:
Proposing as a blocker as the prior version is enable to access the NVD database.

Additionally by moving to a property it will make it easier to override later.

bstansberry (approved):

bstansberry:
Thanks @darranl

PR #18045, opened 2024-07-13
[WFLY-19537] Bind misc external testsuite servers to the server address to avoid test env networking issues

bstansberry:
https://issues.redhat.com/browse/WFLY-19537

Various tests on the s390 CI job fail because the server can't connect to test processes listening on InetAddress.getLocalHost(). Avoid this by binding the test processes to the same address as the WF server.

It's not clear why they can't connect but it's safer to KISS and use the same address as the server.


luck3y:
LGTM, assuming CI passes ;)

luck3y (approved):

bstansberry:
Thanks @luck3y

PR #18050, opened 2024-07-15
[WFLY-19478] Remove usage of deprecated SimpleResourceDefinition methods

lvydra:
https://issues.redhat.com/browse/WFLY-19478

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemRootResourceDefinition.java
@@ -17,7 +17,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
-import org.jboss.as.controller.AbstractAddStepHandler;
 import org.jboss.as.controller.AbstractWriteAttributeHandler;
 import org.jboss.as.controller.AttributeDefinition;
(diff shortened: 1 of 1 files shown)


bstansberry (commented):
@lvydra This fix is correct, but reviewing it made me realize the entire JIRA is unnecessary. We can simply remove the deprecated method and when this code compiles against the updated class:

https://github.com/wildfly/wildfly-core/pull/6084

I'm going to put the 'hold' label on this for now, with the assumption we can close it and the JIRA once the core fix is integrated.

bstansberry:
Oh, and thanks! This is helpful.

bstansberry:
Ah, but this one does need to be changed to override the OperationStepHandler variant instead:

```
    /**
     * Overrides the default impl to use a special definition of the add op that includes additional parameter
     * {@link #DEFAULT_CLUSTERED_SFSB_CACHE}
     * {@inheritDoc}
     */
    @Override
    protected void registerAddOperation(ManagementResourceRegistration registration, AbstractAddStepHandler handler, OperationEntry.Flag... flags) {
        OperationDefinition od = new SimpleOperationDefinitionBuilder(ADD, getResourceDescriptionResolver())
                .setParameters(ATTRIBUTES)
                .addParameter(DEFAULT_CLUSTERED_SFSB_CACHE)
                .withFlags(flags)
                .build();
        registration.registerOperationHandler(od, handler);
    }
```

Perhaps revert what you have here and fix that instead?

lvydra:
Hi @bstansberry, thank you for your review, I have updated PR accordingly.

bstansberry (approved):

bstansberry:
Thanks @lvydra

PR #18061, opened 2024-07-17
WFLY-19496 Promote OIDC scope attribute from Preview to community stability level

PrarthonaPaul:
Issue: https://issues.redhat.com/browse/WFLY-19496
TODO: 
- we need to determine if this requires the model version to be incremented.


pedro-hos:
I've included this changes on the https://github.com/wildfly/wildfly/pull/18471

bstansberry:
I put the 'hold' label on this because #18471 incorporates it.

PR #18068, opened 2024-07-22
[WFLY-19569] Revert "[WFLY-18174] Change the default jakarta.faces.FACELETS_BUFFER…

soul2zimate:
…_SIZE size to -1"

This reverts commit fa8b0fe0d207236f207d64281464f23518d39be5.

Issue: https://issues.redhat.com/browse/WFLY-19569

____

> Wildfly issue links:
> * [WFLY-18174](https://issues.redhat.com/browse/WFLY-18174)


bstansberry:
@soul2zimate This needs a JIRA to track and explain the change. WFLY-18714 is included in a release; it can't just be reverted with no paperwork.

@jasondlee FYI.

jamezp:
I'm not too sure how I feel about this. The defaults we used are a complete guess and don't really think we should be making assumptions here. The user can override this by setting the `ViewHandler.FACELETS_BUFFER_SIZE_PARAM_NAME` property.

That said, maybe we want to just not set it by default. This will take some research to ensure we don't mess anything up. However, it does feel like we shouldn't just be assuming defaults. The size [defaults to 1024](https://jakarta.ee/specifications/faces/4.0/apidocs/jakarta/faces/application/viewhandler#FACELETS_BUFFER_SIZE_PARAM_NAME), maybe we shouldn't be setting it to -1 by default.

soul2zimate:
Hi @bstansberry, the issue is https://issues.redhat.com/browse/WFLY-19569, and it comes from downstream. It seems wildfly-bot give an extra issue from the commit message. Yes, I agree more paperwork is helpful in upstream beside the information in the JBEAP issue.

Hi @jamezp, If we don't set a default value at all, doesn't the small size 1024 fail the TCK again like the original [WFLY-18170](https://issues.redhat.com/browse/WFLY-18170)?

bstansberry:
Thanks for the pointer to the JIRA, @soul2zimate. I apologize for missing it before!

bstansberry:
@jasondlee @jamezp What's the story here? Let's get this closed out one way or the other.

Do we pass the TCK with this change? If not, we can't merge this. In that case we either pick a different default or resolve this issue as Won't Fix. If people want to use multi-jsf and use an older 4.0.x Mojarra they need to set a value.

jamezp:
I'd have to test it, but IIRC with this change we would not pass the TCK. However, we should be able to change the setting for the TCK I _think_. There would need to be some testing done on this though.

FWIW my initial concern on this was the original PR reverted back to a hard-coded value for the buffer size. I more agree with not having a default buffer size like this PR is setup to do now.

bstansberry:
@soul2zimate @jamezp I just resolved a conflict in this. If this passes CI I plan to merge it tomorrow and immediately send up a commit to revert that. We'll let the TCK run tomorrow night with this in it.

If it passes, great, but if not I think we need to just resolve this JIRA by reverting the https://github.com/wildfly/wildfly/pull/16965 change and go back to the calculated default, which worked. Those were guesses, but reasonable ones.

I doubt we can work around this in the TCK and in any case we need to get this resolved and stop messing with it. We've now got 3 JBEAP's about this.

jamezp:
Sounds reasonable to me @bstansberry

bstansberry:
@jamezp @soul2zimate There was no magic; the TCK failed. So I reverted this.

So, we can either go back to the calculated defaults, or Won't Fix the JIRAs. Either is fine with me.

I'll try and find time this morning to send a PR going back to the calculated defaults, unless I find one of you has already created one. We can decide what to do by approving that, or not. If I can't merge it by tomorrow, then we Won't Fix.

jamezp:
I don't have a strong opinion. However, the [spec](https://jakarta.ee/specifications/faces/4.0/jakarta-faces-4.0#a6088) states:

> `jakarta.faces.FACELETS_BUFFER_SIZE` —The buffer size to set on the response when the `ResponseWriter` is generated. By default the value is 1024. A value of -1 will not assign a buffer size on the response. This should be increased if you are using development mode in order to guarantee that the response isn’t partially rendered when an error is generated. The runtime must also consider the `facelets.BUFFER_SIZE` param name as an alias to this param name for backwards compatibility with existing facelets tag libraries.

The problem is the TCK really expects a value large than 1024. The fix might leave this PR in and see if we can fix the TCK with a `web.xml` or some other way of setting the `jakarta.faces.FACELETS_BUFFER_SIZE` with a deployment descriptor.

The one thing I do think is `-1` is the wrong default. That was originally, years ago now, my suggestion, but it's wrong IMO.

PR #18071, opened 2024-07-22
[WFLY-19570] Update HostExcludesTestCase configuration to work with WF33

bstansberry:
Also move the product docs version to 34

https://issues.redhat.com/browse/WFLY-19570

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
@@ -72,7 +72,7 @@ public class HostExcludesTestCase extends BuildConfigurationTestBase {
     private final boolean isFullDistribution = AssumeTestGroupUtil.isFullDistribution();
     private final boolean isPreviewGalleonPack = AssumeTestGroupUtil.isWildFlyPreview();
 
-    private static final String MAJOR = "33.";
+    private static final String MAJOR = "34.";
 
     /**
      * Maintains the list of expected extensions for each host-exclude name for previous releases.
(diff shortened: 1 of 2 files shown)


testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java
```diff
@@ -202,7 +202,8 @@ private enum ExtensionConf {
         WILDFLY_32_0("WildFly32.0", WILDFLY_31_0, List.of(
                 "org.wildfly.extension.mvc-krazo"
         ), List.of(), true),
-        CURRENT(MAJOR, WILDFLY_32_0, getCurrentAddedExtensions(), getCurrentRemovedExtensions(), true);
+        WILDFLY_33_0("WildFly32.0", WILDFLY_32_0, List.of(), List.of(), true),
```
jamezp, line 205: Should the first parameter be "wildfly33.0"?

jamezp (approved):

yersan (approved):

yersan:
Just FYI, we are not forced to move the test to the next version if we have not added / removed extensions in the current cycle:

https://github.com/wildfly/wildfly/blob/e0ead744ec33d6ba0a2bf0d51766a87538afb684/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/HostExcludesTestCase.java#L380-L382

Doing it and adding a new configuration entry is harmless and will allow us to evolve the test together with the WildFly versions, but keep in mind it is not necessary to move it on each WidlFly new version.

bstansberry:
Thanks for the FYI @yersan -- I was wondering why not having this done wasn't breaking anything.

We have the requirement to do this as part of the WildFly release process document. Perhaps we could add more discussion there about when it's required, but OTOH perhaps it's simpler to follow the KISS principle and just always do it. I know when it comes to release time I like the KISS principle!

bstansberry:
Thanks @jamezp and @yersan

PR #18076, opened 2024-07-24
[WFLY-19577] Remove usage of AbstractAddStepHandler.attributes from u…

yersan:
…ndertow ServerAdd

Jira issue: https://issues.redhat.com/browse/WFLY-19577

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/ServerAdd.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/ServerAdd.java
@@ -7,6 +7,7 @@ package org.wildfly.extension.undertow;
 
 import static org.wildfly.extension.undertow.ServerDefinition.SERVER_CAPABILITY;
 
+import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -18,6 +19,7 @@ import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
+import org.jboss.as.controller.registry.AttributeAccess;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.as.web.host.CommonWebServer;
 import org.jboss.dmr.ModelNode;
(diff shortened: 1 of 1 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/ServerAdd.java
```diff
@@ -92,7 +94,11 @@ protected void recordCapabilitiesAndRequirements(OperationContext context, Model
         // We currently don't have any attributes that reference capabilities, but we have this code in case that changes
         // since we are not calling the superclass code.
         ModelNode model = resource.getModel();
-        for (AttributeDefinition ad : getAttributes()) {
+
+        Map<String, AttributeAccess> attributeAccessMap = context.getResourceRegistration().getAttributes(PathAddress.EMPTY_ADDRESS);
+        for (Map.Entry<String, AttributeAccess> entry : attributeAccessMap.entrySet()) {
+            AttributeAccess attributeAccess = entry.getValue();
+            AttributeDefinition ad = attributeAccess.getAttributeDefinition();
```
pferraro, line 101: Technically, we should skip requirement recording for runtime attributes and aliases.
See: https://github.com/wildfly/wildfly-core/blob/main/controller/src/main/java/org/jboss/as/controller/AbstractAddStepHandler.java#L259
I don't see any reason not to replace the first half of this method with a call to the super implementation, do you?

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/ServerAdd.java
yersan: @pferraro It looks like this method Java Doc responds to your question:

https://github.com/wildfly/wildfly/pull/18076/files#diff-9274d3139c44afa3e06390c846378a7a27d2b136ed92ac1050b193b86d0063e4R81-R83

pferraro: ok.

pferraro (approved):

bstansberry:
Thanks @yersan and @pferraro

bstansberry:
@yersan Can you send up a 33.x PR for this? It looks like this fixes a regression in WF 33, as that's when the call to the deprecated constructor was removed.

yersan:
> Can you send up a 33.x PR for this?

Done, follow the related link above

PR #18077, opened 2024-07-24
Fix missing variable in logging statement in PersistentTestXAResource.java

logresearch:
Fix missing variable in logging statement


jamezp:
This fix is not quite correct and this is for a test only.  The user is private and cannot be verified. Therefore this PR is being rejected.

PR #18078, opened 2024-07-24
WFLY-19573 Expand WeldCapability to include build compatible extensions

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-19573

The PR is a draft until I can test this somehow, ideally see if @ehsavoie's code can work with this.

~~Note that this initial version collects information on all BCEs inside `WeldPortableExtensionProcessor` so anyone registering them would have to do that in an earlier processor. I am not sure how much of an issue that presents, if any.
We could move it to a later phase by either passing it around into Weld bootstrap (req. changes directly in Weld) or perhaps we could try to modify where in WFLY we register `LiteExtensionTranslator` and do it shortly before starting bootstrap at which point we could retrieve and attach all found BCEs.~~

EDIT: The above is no longer true, see the comment below

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/extensions/buildcompatible/subsystem/DummyBean.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/extensions/buildcompatible/subsystem/DummyBean.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.weld.extensions.buildcompatible.subsystem;
+
+import jakarta.enterprise.context.Dependent;
+
+@Dependent
+public class DummyBean {
+}
--- a/weld/common/src/main/java/org/jboss/as/weld/deployment/WeldPortableExtensions.java
+++ b/weld/common/src/main/java/org/jboss/as/weld/deployment/WeldPortableExtensions.java
@@ -5,11 +5,15 @@
 
 package org.jboss.as.weld.deployment;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
+import jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension;
 import jakarta.enterprise.inject.spi.Extension;
 
 import org.jboss.as.server.deployment.AttachmentKey;
@@ -18,14 +22,16 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessingException;
 import org.jboss.as.weld.logging.WeldLogger;
 import org.jboss.weld.bootstrap.spi.Metadata;
 import org.jboss.weld.bootstrap.spi.helpers.MetadataImpl;
+import org.jboss.weld.exceptions.IllegalArgumentException;
 
 /**
- * Container class that is attached to the top level deployment that holds all portable extension metadata.
+ * Container class that is attached to the top level deployment that holds all portable and build compatible extensions
+ * metadata.
  * <p/>
- * A portable extension may be available to multiple deployment class loaders, however for each PE we
+ * A CDI extension may be available to multiple deployment class loaders, however for each PE we
  * only want to register a single instance.
  * <p/>
- * This container provides a mechanism for making sure that only a single PE of a given type is registered.
+ * This container provides a mechanism for making sure that only a single PE/BCE of a given type is registered.
  *
  * @author Stuart Douglas
  *
@@ -52,6 +58,7 @@ public class WeldPortableExtensions {
     }
 
     private final Map<Class<?>, Metadata<Extension>> extensions = new HashMap<>();
+    private final Collection<Class<? extends BuildCompatibleExtension>> buildCompatibleExtensions = new ArrayList<>();
 
     public synchronized void tryRegisterExtension(final Class<?> extensionClass, final DeploymentUnit deploymentUnit) throws DeploymentUnitProcessingException {
         if (!Extension.class.isAssignableFrom(extensionClass)) {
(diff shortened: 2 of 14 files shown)


manovotn:
After some manual testing I found out this is just too limiting and I am working on a version that would delay registration of `LiteExtensionTranslator` so that modules/extensions have more space to register their BCEs. I'll force push it here once I verify it makes sense.

While I do like my code tested, I have no idea how to write an automated test for this. Ideas are welcome :)

manovotn:
Pushed an updated version. It should now be possible to register build compatible extension from the same places you'd register a classic portable extensions. Weld now delays creation of `LiteExtensionTranslator` up until creating whole `WeldDeployment` in `WeldDeploymentProcessor`.

manovotn:
Ok, both me and @ehsavoie tested this independently and the API as such does its job.
There are some changes I need/want to do on Weld side (https://github.com/weld/core/pull/3017) but that doesn't affect this PR.

It's IMO ready for review, so marking as such.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldPortableExtensionProcessor.java
```diff
@@ -74,15 +73,13 @@ private void loadAttachments(Module module, DeploymentUnit deploymentUnit, WeldP
...
-                Extension extension = new LiteExtensionTranslator(loadedBuildCompatExtensions, module.getClassLoader());
-                // NOTE: I chose to register it under the same dep. unit as other extensions, not sure if this is correct
-                extensions.registerExtensionInstance(extension, deploymentUnit);
+            // load class and register for discovered build compatible extensions
+            for (Class<? extends BuildCompatibleExtension> extensionClass : loadExtensions(module, buildCompatibleExtensionServices, BuildCompatibleExtension.class)) {
+                extensions.registerBuildCompatibleExtension(extensionClass);
             }
+            // prepare metadata for later registration of LiteExtensionTranslator
+            // NOTE: I chose to register it under the same dep. unit as other extensions, not sure if this is correct
+            extensions.prepareLiteExtensionTranslator(module.getClassLoader(), deploymentUnit);
```
bstansberry, line 82: I might be understanding this incorrectly, but it seems that WeldPortableExtensions is scoped to the top level deployment unit, while this method executes for both top level deployments and subdeployments. So I think subdeployments will end up overwriting each other.

bstansberry (changes requested):
I think testing this would require a fake extension in testsuite/integration/manualmode that adds a BCE. That's certainly doable. We can chat about how to do this.

I don't know what @ehsavoie is planning on doing with this, but if it's not going to appear in the main WF code base right away and when there naturally result in the key cases (e.g. ears) being exercised, then I think we need to add a test.

manovotn:
Thanks for your review @bstansberry.
I am currently on PTO, I'll take a look as soon as I am back!

weld/common/src/main/java/org/jboss/as/weld/deployment/WeldPortableExtensions.java
```diff
@@ -72,8 +81,47 @@ public synchronized void registerExtensionInstance(final Extension extension, fi
...
+     *
+     * @param extensionCreator a function that can create an instance of the {{@code LiteExtensionTranslator}} from given class loader and a collection of found BCEs
+     * @throws DeploymentUnitProcessingException if {{@link #prepareLiteExtensionTranslator(ClassLoader, DeploymentUnit)}} wasn't invoked prior to this method
+     *
+     * @see #prepareLiteExtensionTranslator(ClassLoader, DeploymentUnit)
+     */
+    public synchronized void registerLiteExtensionTranslatorIfNeeded(BiFunction<ClassLoader, List<Class<? extends BuildCompatibleExtension>>,
+            Extension> extensionCreator) throws DeploymentUnitProcessingException {
+        if (liteExtensionTranslatorDepUnit == null || liteExtensionTranslatorCl == null) {
+            throw new DeploymentUnitProcessingException("Incorrect setup for Weld's LiteExtensionTranslator initialization");
```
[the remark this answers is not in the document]
manovotn: Good point, will do.

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldPortableExtensionProcessor.java
manovotn: Hm, that would mean the previous code was also incorrect.
Maybe I could use the CL and DU taken from the `WeldDeploymentProcessor`. Does the top level deployment always have visibility to all subdeployments?

manovotn:
> I don't know what @ehsavoie is planning on doing with this, but if it's not going to appear in the main WF code base right away and when there naturally result in the key cases (e.g. ears) being exercised, then I think we need to add a test.

This is related to his work on wfly extension for AI services where he aimed to reuse a BCE from SmallRye project which wasn't possible with current API.
However, for that particular case, the new API proposed here cannot be used immediately as there were changes needed on Weld side as well (https://github.com/weld/core/pull/3017).

I agree on the test, let's chat about that :+1:

weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldPortableExtensionProcessor.java
manovotn: I've done the above and re-tested the scenario we talked through with Emmanuel and that works fine as do other tests with BCE within CDI TCKs. I'll look into writing the test now and will probably ping you either today or tomorrow.

manovotn:
> I think testing this would require a fake extension in testsuite/integration/manualmode that adds a BCE. That's certainly doable. We can chat about how to do this.

I took a look into this testsuite package and I am not sure I understand how  you meant to test it.
The API addition is specifically for programmatic registration of BCE so a test would have to contain a custom deployment processor which would use the `WeldCapability` to perform the registration and so far I saw no other such tests.

When it comes to standard BCE registration (application having it in its sources and including the `META-INF/services` bits), then that's basically unchanged - we attempt to discover that just like we did until now.

bstansberry:
> > I think testing this would require a fake extension in testsuite/integration/manualmode that adds a BCE. That's certainly doable. We can chat about how to do this.
> 
> I took a look into this testsuite package and I am not sure I understand how you meant to test it. The API addition is specifically for programmatic registration of BCE so a test would have to contain a custom deployment processor which would use the `WeldCapability` to perform the registration and so far I saw no other such tests.
> 

You understood. :) That's what would be needed. Likely this would use the org.jboss.as.test.module.util.TestModule class to provision a (wildfly) extension module. I mentioned testsuite/integration/manualmode because often we stick tests that require provisioning custom modules there, but thinking more about it that shouldn't be necessary.

In case it helps I took an existing class I wrote for something else and tweaked it to provide a trivial subsystem that adds a DUP:

https://github.com/bstansberry/wildfly/pull/new/bce_dup_extension

I'm back from PTO now so we can chat about this a bit less async. ;)

manovotn:
Rebased and added an automated test.

manovotn:
I've added a commit from Emmanuel who helped me figure out why the bootable JAR variant of test failed.
Hopefully, the CI will fully pass now.

manovotn:
I don't think the JDK 17 failures are related to this PR and the bootable part now passes.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/weld/extensions/buildcompatible/subsystem/DummyBean.java
```diff
@@ -0,0 +1,7 @@
+package org.jboss.as.test.integration.weld.extensions.buildcompatible.subsystem;
```
bstansberry, line 1: Please add the copyright/license header to all of these.

bstansberry (changes requested):
@manovotn A couple more things. Ping me when you update and I'll keep an eye on CI and merge when done.

bstansberry (approved):

bstansberry:
Thanks @manovotn.

@ehsavoie, FYI.

PR #18082, opened 2024-07-26
WFLY-19584 Fix typo in testsuite pom.xml: goal "provisioning" to "provision"

TomasHofman:
https://issues.redhat.com/browse/WFLY-19584

The wildfly-maven-plugin goal "provisioning" should be "provision"?

I haven't noticed a functional impact, the configuration segments are to disable the goal and for some reason it seems to be disabled even with the typo.

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -3297,14 +3297,14 @@
                             <execution>
                                 <id>cloud-profile-provisioning</id>
                                 <goals>
-                                    <goal>provisioning</goal>
+                                    <goal>provision</goal>
                                 </goals>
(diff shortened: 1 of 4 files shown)


bstansberry (approved):
In at least ts/int/basic the execution wouldn't have been enabled in the first place, as you'd need to activate -Dts.layers to turn it on.

For others I guess the typo doesn't matter because setting the phase to none disables the execution and presumably maven processes that and moves on before it gets to code that tries to do anything with the value of the 'goal' element.

But the above is just for curiousity's sake; thanks for cleaning this up @TomasHofman!

bstansberry:
Thanks @TomasHofman

PR #18096, opened 2024-08-02
[WFLY-19324] Anonymous 'captor' service does not reliably provide Ope…

bstansberry:
…nTelemetryConfig; use the DeploymentUnit AttachmentKey mechanism

Follow up on #17878 to correct a wiring flaw I misunderstood when I reviewed it.

### DIFF
--- a/microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetryDependencyProcessor.java
+++ b/microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetryDependencyProcessor.java
@@ -7,6 +7,7 @@ package org.wildfly.extension.microprofile.telemetry;
 
 import static org.wildfly.extension.microprofile.telemetry.MicroProfileTelemetrySubsystemDefinition.EXPORTED_MODULES;
 
+import org.jboss.as.controller.ServiceNameFactory;
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
 import org.jboss.as.server.deployment.DeploymentUnit;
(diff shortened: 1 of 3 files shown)


microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
```diff
@@ -36,12 +30,6 @@ protected void performBoottime(OperationContext context, ModelNode operation, Mo
 
         super.performBoottime(context, operation, model);
 
-        ServiceInstaller.builder(ServiceDependency.on(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR))
-                .withCaptor(openTelemetryConfig::set)
-                .build()
-                .install(context);
```
pferraro, line 42: This installs a "passive" service that captures the WildFlyOpenTelemetryConfig as soon as the service the provides it is started.  While this correctly makes an existing WildFlyOpenTelemetryConfig available to the MicroProfileTelemetryDeploymentProcessor, it is do not _require_ that one is available.  If the WildFlyOpenTelemetryConfig is meant to be a requirement, then this mechanism is insufficient for this use case.  I assume that is the case here?

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetryDependencyProcessor.java
```diff
@@ -17,15 +17,17 @@
...
     public void deploy(DeploymentPhaseContext phaseContext) {
         addDependencies(phaseContext.getDeploymentUnit());
 
-        // Ensure the OpenTelemetryConfig is available before the next phase DeploymentUnitPhaseService starts
-        phaseContext.requires(ServiceDependency.on(ServiceName.parse(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR.getName())));
+        // Ensure the OpenTelemetryConfig is available to the Phase.POST_MODULE MicroProfileTelemetryDeploymentProcessor
+        // TODO WFCORE-6491 the kernel should support an API such that an OSH can record this requirement without
+        // needing to involve a DUP like this one that is separate from the one that consumes the dependency.
+        phaseContext.addDeploymentDependency(ServiceName.parse(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR.getName()),
+                MicroProfileTelemetryDeploymentProcessor.CONFIG_ATTACHMENT_KEY);
```
pferraro, line 30: Use the cached ServiceName.
```java
phaseContext.addDeploymentDependency(ServiceNameFactory.resolve(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR), MicroProfileTelemetryDeploymentProcessor.CONFIG_ATTACHMENT_KEY);

pferraro (approved):
Given the existing mechanisms, this is the correct approach (of the two).

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
bstansberry: Yes, the WildFlyOpenTelemetryConfig is required. There's also no guarantee that the passive service here is started before the DeploymentUnitPhaseService that runs MicroProfileTelemetryDeploymentProcessor is started.

microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetryDependencyProcessor.java
bstansberry: Thanks for the reminder! Fixed.

bstansberry:
Thanks @pferraro

PR #18097, opened 2024-08-02
[WFLY-19324] Anonymous 'captor' service does not reliably provide Ope…

bstansberry:
…nTelemetryConfig; use the DeploymentPhaseContext ServiceDependency requirement  mechanism

Follow up on https://github.com/wildfly/wildfly/pull/17878 to correct a wiring flaw I misunderstood when I reviewed it.

This is a draft because I also sent up #18096 with a different approach and reviewers can pick one.

### DIFF
--- a/microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
+++ b/microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
@@ -7,8 +7,6 @@ package org.wildfly.extension.microprofile.telemetry;
 
 import static org.wildfly.extension.microprofile.telemetry.MicroProfileTelemetryExtensionLogger.MPTEL_LOGGER;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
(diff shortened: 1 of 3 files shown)


microprofile/telemetry-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/telemetry/MicroProfileTelemetrySubsystemAdd.java
```diff
@@ -36,11 +33,8 @@ protected void performBoottime(OperationContext context, ModelNode operation, Mo
 
         super.performBoottime(context, operation, model);
 
-        ServiceInstaller.builder(ServiceDependency.on(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR))
-                .withCaptor(openTelemetryConfig::set)
-                .build()
-                .install(context);
-
+        final ServiceDependency<WildFlyOpenTelemetryConfig> configServiceDependency =
+                ServiceDependency.on(ServiceName.parse(WildFlyOpenTelemetryConfig.SERVICE_DESCRIPTOR.getName()));
```
pferraro, line 37: A single ServiceDependency instance is not safe for use by multiple services.
Consider the following:
1. Deploy application1
   1. Starting the DEPENDENCIES phase service will inject the WildFlyOpenTelemetryConfig into the ServiceDependency to be available to subsequent phases, as expected.
   2. The WildFlyOpenTelemetryConfig can be read from the ServiceDependency in the POST_MODULE phase, as expected.
1. Initiate deploy of application2 followed by undeploy of application1
   1. Starting the DEPENDENCIES phase service (for application2) will inject the WildFlyOpenTelemetryConfig into the already populated ServiceDependency.
   2. Stopping the DEPENDENCIES phase service (for application1) will uninject the WildFlyOpenTelemetryConfig from the ServiceDependency.
   3. If (1) happens before (2), during the POST_MODULE phase of application2, MicroProfileTelemetryDeploymentProcessor can read a null from the shared ServiceDependency instance.

pferraro (changes requested):

bstansberry:
Thanks @pferraro -- I'll close this then.

PR #18113, opened 2024-08-07
WFLY-19613 Immutability performance optimizations

pferraro:
https://issues.redhat.com/browse/WFLY-19613

Backport of https://github.com/wildfly/wildfly/pull/18112

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionBeanImmutability.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionBeanImmutability.java
@@ -5,12 +5,15 @@
 
 package org.jboss.as.ejb3.component.stateful;
 
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationHandler;
 
 import org.jboss.as.ee.component.ProxyInvocationHandler;
 import org.jboss.invocation.proxy.ProxyFactory;
 import org.kohsuke.MetaInfServices;
 import org.wildfly.clustering.ee.Immutability;
+import org.wildfly.security.ParametricPrivilegedAction;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 /**
  * Immutability test for EJB proxies, whose serializable placeholders are immutable.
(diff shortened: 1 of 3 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionBeanImmutability.java
```diff
@@ -5,27 +5,43 @@
...
  */
 @MetaInfServices(Immutability.class)
 public class StatefulSessionBeanImmutability implements Immutability {
+    private static final ParametricPrivilegedAction<InvocationHandler, Object> GET_INVOCATION_HANDLER = new ParametricPrivilegedAction<>() {
+        @Override
+        public InvocationHandler run(Object object) {
+            // Since there is a low probability that this object is actually a SFSB proxy, avoid Class.getDeclaredField(...) which will typically throw/catch a NoSuchFieldException
+            for (Field field : object.getClass().getDeclaredFields()) {
+                if (field.getName().equals(ProxyFactory.INVOCATION_HANDLER_FIELD)) {
+                    field.setAccessible(true);
```
jamezp, line 30: We should probably use `field.trySetAccessible()` here since `setAccessible(boolean)` is deprecated.

pferraro: Hmm... That's news to me - when was AccessableObject.setAccessible(...) deprecated?

jamezp: Well, it's not. I'm not sure where I got that idea from. Sorry about that.

jamezp (approved):

ejb3/src/main/java/org/jboss/as/ejb3/component/stateful/StatefulSessionBeanImmutability.java
jamezp: Oh, maybe. I guess I thought it was because of modules and the preference towards `VarHandle` now. I do know `isAccessible()` is [deprecated](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/reflect/AccessibleObject.html#isAccessible()).

jamezp:
Thank you @pferraro

PR #18116, opened 2024-08-08
WFLY-19605 Add a CI job to check for non-i18n INFO/WARN/ERROR logging

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-19605

### DIFF
--- a/.github/workflows/check_logging.yml
+++ b/.github/workflows/check_logging.yml
@@ -1,0 +1,22 @@
+name: Check non-i18n logging
+
+on:
+  pull_request:
+    branches:
+      - main
+      - '*.x'
+
+jobs:
+  check-logging:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run logging check
+        env:
+          GITHUB_BASE_REF: '${{ github.base_ref }}'
+        shell: bash
+        run: bash check_logging.sh
--- a/check_logging.sh
+++ b/check_logging.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+BASE_BRANCH="${GITHUB_BASE_REF:-main}"
+DIFF=$(git diff origin/"$BASE_BRANCH"...HEAD || true)
+
+if [ -z "$DIFF" ]; then
+  echo "No diff found to analyze."
+  exit 0
+fi
+
+PATTERNS=(
+  ".info("
+  ".infof("
+  ".warn("
+  ".warnf("
+  ".error("
+  ".errorf("
+  ".fatal("
+  ".fatalf("
+  "System.out.print"
+  "System.err.print"
+  ".printStackTrace"
+)
+
+ERRORS=""
+CURRENT_FILE=""
+
+while IFS= read -r line; do
+  # Update the current file if a new diff starts
+  if [[ "$line" =~ ^diff\ --git\ a\/.*\ b\/.* ]]; then
+    CURRENT_FILE=$(echo "$line" | awk '{print $3}' | sed 's/^a\///')
+  fi
+
+  if [[ "$line" =~ ^\+[^+] ]]; then
+    # Ignore lines in the test directories
+    if [[ "$CURRENT_FILE" != *"src/test/"* && "$CURRENT_FILE" != *"testsuite/"* && "$CURRENT_FILE" != *"check_logging.sh"* ]]; then
+      # Check for any of the patterns, ensuring "//" doesn't precede them
+      for pattern in "${PATTERNS[@]}"; do
+        if [[ "$line" == *"$pattern"* ]]; then
+          # Ensure the pattern is not commented out with "//"
+          if [[ "$line" != *"//"*"$pattern"* ]]; then
+            # Capture the error line and its context
+            ERRORS+="$line\nFile: $CURRENT_FILE\n\n"
+          fi
+        fi
+      done
+    fi
+  fi
+done <<< "$DIFF"
+
+if [ -n "$ERRORS" ]; then
+  echo -e "Logging statements found that should be internationalized or converted to a lower log level:\n"
+  echo -e "$ERRORS"
+  exit 1
+else
+  echo "No problematic logging statements found."
+  exit 0
+fi


check_logging.sh
```diff
@@ -0,0 +1,57 @@
...
+
+while IFS= read -r line; do
+  # Update the current file if a new diff starts
+  if [[ "$line" =~ ^diff\ --git\ a\/.*\ b\/.* ]]; then
+    CURRENT_FILE=$(echo "$line" | awk '{print $3}' | sed 's/^a\///')
+  fi
+
+  if [[ "$line" =~ ^\+[^+] ]]; then
+    # Ignore lines in the test directories
+    if [[ "$CURRENT_FILE" != *"src/test/"* && "$CURRENT_FILE" != *"testsuite/"* ]]; then
```
rhusar, line 35: This also needs to exclude this file itself, otherwise it will always fail :) 

[.github/workflows/check_logging.yml](https://github.com/wildfly/wildfly/pull/18116/files#diff-f65ba3d4fa0016b455b4f0a026ba7cab93315c7db73aa6565a677052292e9b37)

rhusar (changes requested):

check_logging.sh
RanabirChakraborty: Yeah, I haven't added it on purpose so that we can check it's working. I'm excluding this file now.

rhusar: > Yeah, I haven't added it on purpose so that we can check it's working. I'm excluding this file now.

If I were to test this, I would test this in a private fork rather than upstream and would test in actual code rather than this file, but that's just a preference 👍🏼

.github/workflows/check_logging.yml
```diff
@@ -0,0 +1,21 @@
+name: Check PR Logging
```
rhusar, line 1: This needs a better description, this is very ambiguos.

.github/workflows/check_logging.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
```
rhusar, line 6: What's the reason for not including .x branches?

.github/workflows/check_logging.yml
```diff
@@ -0,0 +1,21 @@
...
+  check-logging:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set execute permission
+        run: chmod +x check_logging.sh
```
rhusar, line 18: Why not just replace this step by calling sh directly?

.github/workflows/check_logging.yml
```diff
@@ -0,0 +1,21 @@
...
+
+      - name: Run logging check
+        run: ./check_logging.sh
```
rhusar, line 21: Why not just run `sh check_logging.sh` here.

rhusar (changes requested):

.github/workflows/check_logging.yml
rhusar: How about "Check for non-i18n logging"

RanabirChakraborty:
@rhusar Thank you for the comments, I have fixed them accordingly. Can you please review again

darranl:
If we are doing this in WildFly we should probably do the same from the outset in WildFly Core.

Also it looks like this will not work with block comments but as this only scans the current PR maybe false positives are at merger discretion.

RanabirChakraborty:
Thanks, @darranl I have created a [WFCORE-6955](https://issues.redhat.com/browse/WFCORE-6955) issue for the same.

.github/workflows/check_logging.yml
```diff
@@ -0,0 +1,20 @@
...
+    branches:
+      - main
+      - '*.x'
+
+jobs:
+  check-logging:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
bstansberry, line 14: Please use checkout@v4.

bstansberry (changes requested):

.github/workflows/check_logging.yml
RanabirChakraborty: Thanks! fixed it.

check_logging.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+DIFF=$(git diff origin/main...HEAD || true)
```
yersan, line 3: This diff should take into account when the pull request is against other branches that are not `main`, otherwise it will be making comparisons not only the pull request changes but also any diff between the current branch and `main`

yersan (changes requested):

check_logging.sh
RanabirChakraborty: Most of the PRs for WildFly are against `main` branch only, in that case, if we stop the diff check then it'll pass automatically. And also, I'm not sure how we can fetch the details of a pull request to check which base branch it's targeting. Can you please explain a bit?

yersan: > Most of the PRs for WildFly are against main branch only, in that case, if we stop the diff check then it'll pass automatically.

We produce micro releases, so this should work also with these branches without any manual intervention.


>  I'm not sure how we can fetch the details of a pull request to check which base branch it's targeting.

I have no so much idea either, searching for information about that it looks like that `github.base_ref` property can provide you this information:

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

github.base_ref: The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

This was a quick search only, I'm not sure even if that is exactly what we need, but looks promising. So, it is a matter of investigating it.

RanabirChakraborty: @yersan Thanks! I have fixed it now according to our conversation. Additionally, I have also added the below code in GitHub Action so that we can see base branch is correctly fetched or not. I'll remove the part once you think it's correctly done.

```
- name: Print GitHub context
    run: echo "${{ toJson(github) }}"
```
output:        

    "head_ref": "WFLY-19605",
    "base_ref": "main",

yersan: Thanks @RanabirChakraborty , looking at the code, this looks fixed. 

Is there any proof of how this change works in action? I mean any test done by you where invalid logs are found and so one? That for sure will help to the reviewers.

Notice other revision comments have not been resolved yet.

rhusar: +1 as suggested [above](https://github.com/wildfly/wildfly/pull/18116#discussion_r1722055354), would be nice to see this in action in a private fork example PR.

RanabirChakraborty: @yersan @rhusar, according to your suggestion to test this GitHub Action, I created a dummy repo(https://github.com/RanabirChakraborty/spring-demo) with this GitHub Action present and sent a PR with the non-i18n INFO/WARN/ERROR logging and you can see the action is failing for the PR - https://github.com/RanabirChakraborty/spring-demo/pull/1, You can also check it by yourself.

yersan: Thanks, @RanabirChakraborty, look good from my side, I'm done with my review

yersan (approved):

rhusar (approved):
LGTM now

bstansberry:
Thanks @RanabirChakraborty @rhusar and @yersan

PR #18130, opened 2024-08-15
[WFLY-19327] Validate it's possible to secure two apps within the same EAR with FORM and OIDC credential support

rsearls:
https://issues.redhat.com/browse/WFLY-19327

### DIFF
--- a/testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
+++ b/testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
@@ -18,16 +18,19 @@ import static org.wildfly.security.http.oidc.Oidc.AuthenticationRequestFormat.RE
 import static org.wildfly.security.http.oidc.Oidc.AuthenticationRequestFormat.OAUTH2;
 import static org.wildfly.test.integration.elytron.oidc.client.KeycloakConfiguration.ALLOWED_ORIGIN;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 
 import org.apache.http.Header;
 import org.apache.http.HttpResponse;
+import static org.apache.http.HttpStatus.SC_OK;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.CookieStore;
 import org.apache.http.client.HttpClient;
@@ -43,6 +46,7 @@ import org.apache.http.impl.client.LaxRedirectStrategy;
 import org.apache.http.message.BasicNameValuePair;
 import org.apache.http.protocol.BasicHttpContext;
 import org.apache.http.protocol.HttpContext;
+import org.apache.http.util.EntityUtils;
 import org.jboss.arquillian.container.test.api.OperateOnDeployment;
 import org.jboss.as.arquillian.api.ServerSetupTask;
 import org.jboss.as.arquillian.container.ManagementClient;
@@ -52,6 +56,7 @@ import org.jboss.as.test.http.util.TestHttpClientUtils;
 import org.jboss.as.test.integration.management.ManagementOperations;
 import org.jboss.as.test.integration.security.common.servlets.SimpleSecuredServlet;
 import org.jboss.as.test.integration.security.common.servlets.SimpleServlet;
+import org.jboss.as.test.shared.ManagementServerSetupTask;
 import org.jboss.as.test.shared.TestSuiteEnvironment;
 import org.jboss.as.test.shared.util.AssumeTestGroupUtil;
 import org.jboss.as.version.Stability;
@@ -64,6 +69,7 @@ import org.junit.Test;
 import org.keycloak.representations.idm.RealmRepresentation;
 import org.wildfly.common.iteration.CodePointIterator;
 import org.wildfly.security.jose.util.JsonSerialization;
+import org.wildfly.test.integration.elytron.oidc.client.deployment.OidcWithDeploymentConfigTest;
 import org.wildfly.test.integration.elytron.oidc.client.subsystem.SimpleServletWithScope;
 
 import io.restassured.RestAssured;
@@ -87,6 +93,8 @@ public abstract class OidcBaseTest {
     public static final String AUTH_SERVER_URL_APP = "AuthServerUrlOidcApp";
     public static final String WRONG_PROVIDER_URL_APP = "WrongProviderUrlOidcApp";
     public static final String WRONG_SECRET_APP = "WrongSecretOidcApp";
+    public static final String FORM_WITH_OIDC_EAR_APP = "FormWithOidcApp";
+    public static final String FORM_WITH_OIDC_OIDC_APP = "oidc";
     public static final String DIRECT_ACCCESS_GRANT_ENABLED_CLIENT = "DirectAccessGrantEnabledClient";
     public static final String BEARER_ONLY_AUTH_SERVER_URL_APP = "AuthServerUrlBearerOnlyApp";
     public static final String BEARER_ONLY_PROVIDER_URL_APP = "ProviderUrlBearerOnlyApp";
@@ -117,6 +125,20 @@ public abstract class OidcBaseTest {
     public static final String INVALID_SIGNATURE_ALGORITHM_APP = "InvalidSignatureAlgorithmApp";
     public static final String PS_SIGNED_REQUEST_URI_APP = "PsSignedRequestUriApp";
     public static final String MISSING_SECRET_APP = "MissingSecretApp";
+    public static final String FORM_USER="user1";
+    public static final String FORM_PASSWORD="password1";
+    protected static final String ERROR_PAGE_CONTENT = "Error!";
+
+    // Avoid problem on windows with path
+    public static final String USERS_PATH = new File(
+            OidcWithDeploymentConfigTest.class.getResource("users.properties").getFile()).getAbsolutePath()
+            .replace("\\", "/");
+    public static final String ROLES_PATH = new File(
+            OidcWithDeploymentConfigTest.class.getResource("roles.properties").getFile()).getAbsolutePath()
+            .replace("\\", "/");
+    public static final String ORIGINAL_USERS_PATH = "application-users.properties";
+    public static final String ORIGINAL_ROLES_PATH = "application-roles.properties";
+    public static final String RELATIVE_TO = "jboss.server.config.dir";
 
     private final Stability desiredStability;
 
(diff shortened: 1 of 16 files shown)


testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
```diff
@@ -5,27 +5,7 @@
 
 package org.wildfly.test.integration.elytron.oidc.client;
 
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SYSTEM_PROPERTY;
```
fjuma, line 8: Just a small note, it's good to avoid reordering imports (since it's easier to see the changes related specifically to this PR).

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
```diff
@@ -303,15 +303,17 @@ private void processDeployment(final WarMetaData warMetaData, final DeploymentUn
...
                 applySecurityFunction = builder.requires(capabilitySupport.getCapabilityServiceName(Capabilities.CAPABILITY_APPLICATION_SECURITY_DOMAIN, securityDomainName));
             } else {
                 throw ROOT_LOGGER.deploymentConfiguredForLegacySecurity();
             }
         }
+        else if (isVirtualDomainRequired(deploymentUnit) || isVirtualMechanismFactoryRequired(deploymentUnit)) {
+            securityDomain = builder.requires(securityMetaData.getSecurityDomain());
+        }
+
         if (isVirtualMechanismFactoryRequired(deploymentUnit)) {
```
fjuma, line 317: I think we can move this if statement into the else if:

```
} else if (isVirtualDomainRequired(deploymentUnit) || isVirtualMechanismFactoryRequired(deploymentUnit)) {
        securityDomain = builder.requires(securityMetaData.getSecurityDomain());

        if (isVirtualMechanismFactoryRequired(deploymentUnit)) {
            ...
        }
}

File: testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java [hunk 3]
```diff
@@ -350,6 +368,39 @@ public static WebArchive createOpenIDWithMissingSecretHmacSigningAlgorithm() {
                 .addAsWebInfResource(OidcWithDeploymentConfigTest.class.getPackage(), MISSING_SECRET_WITH_HMAC_ALGORITHM_FILE, "oidc.json");
     }
 
+    @Deployment(name = FORM_WITH_OIDC_EAR_APP, managed = false, testable = false)
+    public static Archive<?> createFormWithOidcDeployment() {
```
fjuma, line 372: Do you think it would also be possible to add a test case to `OidcWithSubsystemConfigTest`? The main difference with the subsystem config tests is that instead of specifying the oidc.json file in the deployment and specifying the `OIDC` auth-method in the deployment, a `secure-deployment` is configured in the `elytron-oidc-client` subsystem. For most cases, we try to add tests to both `OidcWithDeploymentConfigTest` and `OidcWithSubsystemConfigTest` to make sure both methods of configuring OIDC work for the specific use case.

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
```diff
@@ -567,6 +618,17 @@ public void testCorsRequestWithEnableCorsWithInvalidOrigin() throws Exception {
         }
     }
 
+    @Test
+    @InSequence(27)
+    public void testFormWithOidc() throws Exception {
```
fjuma, line 623: Just as a sanity check, were you able to verify that this test fails without the changes in `UndertowDeploymentProcessor` and passes after these changes?

fjuma:
Just a minor comment, please update the title of the PR to include `[WFLY-19327]`.

You can also squash the commits.

fjuma (changes requested):
Thanks very much for the PR, @rsearls! I've added some small comments.

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
```diff
@@ -567,6 +618,17 @@ public void testCorsRequestWithEnableCorsWithInvalidOrigin() throws Exception {
...
+        try {
+            deployer.deploy(FORM_WITH_OIDC_EAR_APP);
+            super.testFormWithOidc();
+        } finally {
+            deployer.undeploy(FORM_WITH_OIDC_EAR_APP);
+        }
+    }
+
```
PrarthonaPaul, line 631: Maybe it would be good to also add some checks where the authentication fails? (incorrect username/passwords/roles, etc)

PrarthonaPaul (commented):
Hi @rsearls 
I have added some minor comments

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
rsearls: ok

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
rsearls: tested.  It failed as it should.

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
```diff
@@ -493,6 +497,38 @@ public void testOpenIDWithMissingSecretHmacSigningAlgorithm() throws Exception {
                 "/" + MISSING_SECRET_APP + SimpleSecuredServlet.SERVLET_PATH).toURI(), true);
     }
 
+    @Test
+    @OperateOnDeployment(FORM_WITH_OIDC_EAR_APP)
+    public void testFormWithOidc() throws Exception {
+        // oidc login
+        // EAR declares context-root to be oidc
+        loginToApp("oidc",
```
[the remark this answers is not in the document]
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
```diff
@@ -45,7 +49,9 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
-@ServerSetup({ OidcWithDeploymentConfigTest.PreviewStabilitySetupTask.class, OidcWithDeploymentConfigTest.KeycloakAndSystemPropertySetup.class })
+@ServerSetup({ OidcWithDeploymentConfigTest.PreviewStabilitySetupTask.class,
+        OidcWithDeploymentConfigTest.KeycloakAndSystemPropertySetup.class,
+        OidcWithDeploymentConfigTest.WflyServerSetupTask.class})
```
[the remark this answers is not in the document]
rsearls: done

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
```diff
@@ -303,15 +303,17 @@ private void processDeployment(final WarMetaData warMetaData, final DeploymentUn
         }
 
         final SecurityMetaData securityMetaData = deploymentUnit.getAttachment(ATTACHMENT_KEY);
-        if (isVirtualDomainRequired(deploymentUnit) || isVirtualMechanismFactoryRequired(deploymentUnit)) {
-            securityDomain = builder.requires(securityMetaData.getSecurityDomain());
-        } else if(securityDomainName != null) {
+        if(securityDomainName != null) {
```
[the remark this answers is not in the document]
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/resources/org/wildfly/test/integration/elytron/oidc/client/deployment/FormWithOidcApp_application.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
rsearls: for OIDC login I am using the common method loginToApp which is
      doing all the checking and reporting for the other tests.
    for testFormCredentials() that type of detail is not return in the
      response. Only the HTTP failure number is returned.

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
bstansberry: Also, the existing import order is the correct one for WildFly, so please restore it.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowDeploymentProcessor.java
fjuma: I don't see this change here.

rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java
rsearls: done

testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/deployment/OidcWithDeploymentConfigTest.java
fjuma: I see there's an error.jsp page configured for the form app. Would it be possible to check that we get to the error page when an invalid username/password is specified (similar to the test here: https://github.com/wildfly/wildfly/blob/main/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/http/FormMechTestBase.java#L142-L151)?

Same for OIDC, would it be possible to check that the wrong password is handled correctly as shown here: https://github.com/wildfly/wildfly/blob/main/testsuite/integration/elytron-oidc-client/src/test/java/org/wildfly/test/integration/elytron/oidc/client/OidcBaseTest.java#L157-L159

rsearls: done

fjuma (approved):
Thanks very much @rsearls!

fjuma:
@rsearls Just a minor thing I just noticed. If you could update the commit message to describe the fix that would be great (e.g., "[WFLY-19327] Ensure it's possible to secure two apps within the same EAR with FORM and OIDC" or something like that)

rsearls:
done

fjuma (approved):
Thanks @rsearls!

bstansberry:
The 'WildFly Preview - Linux - JDK 17' job had failures with related tests so I started it again. My sense is OIDC tests in general have intermittent failures so perhaps that was the case here, but in any case that job will need to pass before we should merge this.

bstansberry (dismissed):
@rsearls I'm 'requesting changes' here, but I don't have anything specific. But there are repeated test failures on the WildFly Preview CI job so those need to be corrected.

rsearls:
The 2 tests that fail are unrelated to elytron security.
JWTIdentityPropagationTest is a known issue.
FineHotRodSessionExpirationTestCase is a clustering test.  It does not declare any security.

fjuma:
@rsearls Just to check, it looks like an additional change was made for the fix:

https://github.com/wildfly/wildfly/pull/18130/files#diff-26e63e6957fd8bc1c91552063b1a65f04dfff54a8ed76f96214f6f7089b7fe0aR43

Can you add a comment here about the reason for introducing that line?

Have we had enough successful test runs since that change was added to see if it solves the intermittent failure you were hitting?

rsearls:
@fjuma 
The failure is a thread timing issue.  For an EAR each subdeployment is
processed by a different thread.  A OIDC subdeployment sets a flag on
the ROOT deploymentUnit that causes a virtural securityDomain name 
to be generated and saved in the subdeployment.  Once the flag is set on the 
ROOT deploymentUnit all subdeployment units processed after the flag
is set, passing through the "if" stmt get a virtural securityDomain name.
This leads to the wrong *AuthenticationMechanism class being selected
and executed.  

Setting the flag on the OIDC subdeployment only does not work.  It breaks
other processing code.   This is the cleanest and simplest fix.

fjuma:
@rsearls Thanks for the explanation, will take a closer look in a bit. Do you think we've had enough test runs to confirm the fix is working as intended?

rsearls:
Yes I do

fjuma:
Thanks very much for the explanation, @rsearls! I took a closer look and ensuring that the auth method really is OIDC in the OIDC `VirtualHttpServerMechanismFactoryNameProcessor` makes sense. Thanks!

PR #18131, opened 2024-08-16
WFLY-19610 Attempt to prevent double post construct invocation for injection targets that are beans

manovotn:
Related to https://issues.redhat.com/browse/WFLY-19610

This PR is to see what CI has to say about this change and reveal any use cases for which it might have been coded that way,
If it passes, we should add a test similar to the reproducer in WFLY-19610 before merging.

cc @jasondlee


manovotn:
Superseeded by https://github.com/wildfly/wildfly/pull/18132

PR #18132, opened 2024-08-16
[WFLY-19610]  @PostConstruct on Servlet may be called twice

jasondlee:
https://issues.redhat.com/browse/WFLY-19610

- Attempt to prevent double post construct invocation for injection targets that are beans
- Add reproducer/test

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/web/postcontstruct/beanjar/SimpleBean.java
+++ b/testsuite/integration/basic/src/test/java/org/wildfly/test/integration/web/postcontstruct/beanjar/SimpleBean.java
@@ -1,0 +1,9 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.test.integration.web.postcontstruct.beanjar;
+
+public class SimpleBean {
+}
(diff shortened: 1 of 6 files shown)


testsuite/integration/basic/src/test/java/org/wildfly/test/integration/web/postcontstruct/beanjar/SimpleBean.java
```diff
@@ -0,0 +1,4 @@
+package org.wildfly.test.integration.web.postcontstruct.beanjar;
+
+public class SimpleBean {
```
manovotn, line 3: Nitpickers note: I'd add a bean defining annotation to this class (`@Dependent`/`@ApplicationScoped`/...).

manovotn (approved):

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/web/postcontstruct/beanjar/SimpleBean.java
jasondlee: I probably would, too, but this is what the reporter provided in the reproducer, so I've tried to change that as little as possible.

bstansberry (changes requested):
Please add the headers to all files.

jasondlee:
> Please add the headers to all files.

Done.

bstansberry (approved):

testsuite/integration/basic/src/test/java/org/wildfly/test/integration/web/postcontstruct/beanjar/SimpleBean.java
manovotn: It makes no difference for the purpose of this test - it's just that you now rely on having `beans.xml` with explicitly stated `all` discovery mode whereas with added annotation, the test would work even with empty `beans.xml` or completely without it :)

manovotn (approved):

bstansberry:
Thanks @jasondlee and @manovotn

PR #18108, opened 2024-08-07
[WFLY-19606] Workaround infinispan ISPN-16346 JDK23 issue

ropalka:
https://issues.redhat.com/browse/WFLY-19606

PR #17910, opened 2024-05-23
WFLY-18857 Replace wildfly clustering modules with versions from wildfly-clustering

pferraro:
~~Requires https://github.com/wildfly/wildfly/pull/17860~~
Includes WFLY-19235 Simplify singleton service installation

https://issues.redhat.com/browse/WFLY-18857
https://issues.redhat.com/browse/WFLY-19235

I tried to make this as granular as possible, but migrating individual modules would have required a lot of intermediate code changes that would only have become obsolete.

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> Wildfly issue links:
> * [WFLY-19139](https://issues.redhat.com/browse/WFLY-19139)
> * [WFLY-16973](https://issues.redhat.com/browse/WFLY-16973)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #18148, opened 2024-08-26
[WFLY-19668] Upgrade Infinispan to 14.0.31.Final

ropalka:
https://issues.redhat.com/browse/WFLY-19668

This PR supersedes https://github.com/wildfly/wildfly/pull/18108


pferraro:
In general, Infinispan/JGroups upgrades should always be validated first against wildfly-clustering (as of https://github.com/wildfly/wildfly/pull/17910).
Also, this needs an additional rxjava upgrade.  I'll send a new PR when ready.

PR #18149, opened 2024-08-26
[WFLY-19671] Remove 'galleon content only' dependencies from the list…

bstansberry:
… of feature pack deps

https://issues.redhat.com/browse/WFLY-19671

### DIFF
--- a/ee-feature-pack/galleon-feature-pack/pom.xml
+++ b/ee-feature-pack/galleon-feature-pack/pom.xml
@@ -180,7 +180,7 @@
                             </licensesConfigFiles>
                             <licensesOutputFile>${license.directory}/wildfly-ee-feature-pack-licenses.xml</licensesOutputFile>
                             <excludedGroups>org.wildfly.galleon-plugins</excludedGroups>
-                            <excludedArtifacts>wildfly-standard-ee-bom|wildfly-ee-feature-pack-galleon-shared|wildfly-ee-feature-pack-galleon-local|wildfly-common-ee-dependency-management|wildfly-jar-boot|wildfly-core-feature-pack-common|wildfly-core-feature-pack-galleon-common|wildfly-core-feature-pack-galleon-pruned|wildfly-elytron\z</excludedArtifacts>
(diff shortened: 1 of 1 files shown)


ee-feature-pack/galleon-feature-pack/pom.xml
```diff
@@ -237,12 +237,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- WFLY-19671. DO NOT include 'galleon content only' dependencies
```
jfdenise, line 240: @bstansberry , I can see some remaining usages of wildfly-core-feature-pack-galleon-common (ban-transitive-deps and update-licenses-xml). BTW, perhaps cleaning wildfly-core-feature-pack-galleon-pruned too?

jfdenise: @bstansberry , another point, just to be sure... Looking at https://github.com/wildfly/wildfly-core/tree/main/core-feature-pack/galleon-common/src/main/resources/packages/org.wildfly.openssl/pm/wildfly and https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/galleon-common/src/main/resources/packages/core.misc.common/pm/wildfly/tasks.xml we can see artifacts that require a version.

These artifacts are resolved thanks to the dependency in https://github.com/wildfly/wildfly-core/blob/main/core-feature-pack/common/pom.xml so we are good. 

Perhaps in wildfly-core we should add a note to not add dependencies to galleon-common, add them to common?

bstansberry: @jfdenise re your second comment, I filed the much broader scoped https://issues.redhat.com/browse/WFCORE-6755. Thanks for pointing this out. I've long had the urge to get core-feature-pack reorganized but never had a reason to push it.

bstansberry: Re your first comment -- good catch. I just pushed an update cleaning those out.

jfdenise (approved):

bstansberry:
BTW in my earlier response to your review I posted a link to some random JIRA, not the one I filed. The right link is https://issues.redhat.com/browse/WFCORE-6961.

PR #18152, opened 2024-08-27
WFLY-19664 Log warning if hibernate.cache.region.factory_class setting is ignored

scottmarlow:
Wildfly issue links:
> * [WFLY-19664](https://issues.redhat.com/browse/WFLY-19664)

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/JpaLogger.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/JpaLogger.java
@@ -6,6 +6,7 @@
 package org.jboss.as.jpa.hibernate;
 
 import static org.jboss.logging.Logger.Level.INFO;
+import static org.jboss.logging.Logger.Level.WARN;
 
 import org.hibernate.boot.archive.spi.ArchiveException;
 import org.jboss.logging.BasicLogger;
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

bstansberry:
Thanks @scottmarlow

scottmarlow:
fyi @Sanne @yrodiere with this change, WildFly 34 will log a warning if the hibernate.cache.region.factory_class setting is ignored due to the JPA/Persistence container causing it to be ignored.

Sanne:
thanks @scottmarlow !

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/JpaLogger.java
```diff
@@ -46,6 +47,10 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 20261, value = "Hibernate ORM did not register LifeCycleListener")
     IllegalStateException HibernateORMDidNotRegisterLifeCycleListener();
 
+    @LogMessage(level = WARN)
+    @Message(id = 20262, value = "Application custom cache region setting is ignored %s=%s")
+    void ignoredCacheRegionSetting(String propertyName, String setting );
```
Sanne, line 52: maybe the message could suggest which property is expected to be used instead in Wildfly?

PR #18153, opened 2024-08-28
[WFLY-19657] Update AppClient parsing to use the "Schemas" approach with stability level support.

darranl:
https://issues.redhat.com/browse/WFLY-19657

This PR will depend upon the following from WildFly Core:
    https://github.com/wildfly/wildfly-core/pull/6134


bstansberry (approved):

darranl:
I assume those JRE builds don't use a merge commit to pull in the latest WildFly Core?

darranl:
Just temporarily adding "hold" - I realise I may need an additional parameter to the ManagementXmlReaderWriter interface so easier without this being merged.

darranl:
@bstansberry FYI I have added one more commit since your last approval (Looks like wildfly-bot has retracted that) - anyway I just need to make a small API change in core to the new interface - the new commit makes WildFly compatible with what is in WildFly Core today and what will be in a future version.

bstansberry (approved):
@darranl I've re-approved.

I'd have merged it but it still has the 'hold' label.

darranl:
Sorry yes that hold label was supposed to have come off with the last commit.

bstansberry:
Thanks @darranl

PR #18154, opened 2024-08-28
[WFLY-19681] minimize lock hold while iterating across existing entries in refresh task

aogburn:
https://issues.redhat.com/browse/WFLY-19681

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
@@ -1099,11 +1099,14 @@ public class DatabaseTimerPersistence implements TimerPersistence, Service {
                                 }
                             }
 
+                            Set<String> timers;
                             synchronized (DatabaseTimerPersistence.this) {
-                                Set<String> timers = knownTimerIds.get(timedObjectId);
-                                for (String timer : existing) {
-                                    TimerImpl timer1 = timerService.getTimer(timer);
-                                    if (timer1 != null && timer1.getState() != TimerState.CREATED) {
+                                 timers = knownTimerIds.get(timedObjectId);
+                            }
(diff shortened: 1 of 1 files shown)


bstansberry:
@aogburn Can you correct the commit message? It's the wrong JIRA.

aogburn:
Yes, changed! Sorry, I moved the original jira instead of my cloned one and forgot to fix up the commit message.

ejb3/src/main/java/org/jboss/as/ejb3/timerservice/persistence/database/DatabaseTimerPersistence.java
```diff
@@ -1099,11 +1099,11 @@ public void run() {
                                 }
                             }
 
-                            synchronized (DatabaseTimerPersistence.this) {
-                                Set<String> timers = knownTimerIds.get(timedObjectId);
-                                for (String timer : existing) {
-                                    TimerImpl timer1 = timerService.getTimer(timer);
-                                    if (timer1 != null && timer1.getState() != TimerState.CREATED) {
+                            Set<String> timers = knownTimerIds.get(timedObjectId);
```
bstansberry, line 1102: All other access to knownTimerIds in this class is guarded by 'synchronized (DatabaseTimerPersistence.this)' (or synchronized (this) in DatabaseTimerPersistence itself.

Except the first lie of loadActiveTimers, which is probably a bug.

Changing the concurrency handling for just one call seems risky.

The sets in the map also aren't thread safe.

aogburn: D'oh, thanks for catching that Brian.  Added the knownTImerIds back within a sync there.

aogburn: Any further thoughts on this @bstansberry @tadamski ?

bstansberry:
Thanks @aogburn

PR #18159, opened 2024-08-30
[WFLY-19692]: Documentation for WFCORE-6960 read-config-as-xml-file

ehsavoie:
Updating CLI recipes for the new operation "read-config-as-xml-file".

Jira: https://issues.redhat.com/browse/WFLY-19692
      https://issues.redhat.com/browse/WFCORE-6960

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
@@ -190,15 +190,30 @@ Assume you have a host that is called "primary"
 
 [source,options="nowrap"]
 ----
-[domain@localhost:9990 /] /host=primary:read-config-as-xml
+[domain@localhost:9990 /] attachment display --operation=/host=primary:read-config-as-xml-file
+----
+
+If you want to get the configuration of the server "main-one" hosted on "primary":
+
+[source,options="nowrap"]
+----
+[domain@localhost:9990 /] attachment display --operation=/host=primary/server=main-one:read-config-as-xml-file
 ----
 
 Just for the domain or standalone
 
 [source,options="nowrap"]
 ----
-[domain@localhost:9990 /] :read-config-as-xml
+[domain@localhost:9990 /] attachment display --operation=:read-config-as-xml-file
 ----
+
+If you want to save the file you can use:
+----
+[standalone@localhost:9990 /] attachment save --operation=:read-config-as-xml-file --file=/home/wildfly/tmp/standalone-runtime.xml
+----
+
+Please note that the operation "read-config-as-xml" that used to return the XML as a String has been deprecated.
+
 
 [[take-a-snapshot-of-what-the-current-domain-is]]
 === Take a snapshot of what the current domain is
@@ -214,6 +229,7 @@ Just for the domain or standalone
     }
 }
 ----
+
 
 [[take-the-latest-snapshot-of-the-host.xml-for-a-particular-host]]
 === Take the latest snapshot of the host.xml for a particular host


docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
```diff
@@ -190,14 +190,19 @@ Assume you have a host that is called "primary"
 
 [source,options="nowrap"]
 ----
-[domain@localhost:9990 /] /host=primary:read-config-as-xml
```
fabiobrz, line 193: The new text should be added and the old one should NOT be removed now, since the `read-config-as-xml` still exists.

docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
```diff
@@ -190,14 +190,19 @@ Assume you have a host that is called "primary"
...
+[domain@localhost:9990 /] attachment display --operation=/host=primary:read-config-as-xml
 ----
 
 Just for the domain or standalone
 
 [source,options="nowrap"]
 ----
-[domain@localhost:9990 /] :read-config-as-xml
```
fabiobrz, line 200: Same as above, let's add rather than replace.

docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
```diff
@@ -190,14 +190,19 @@ Assume you have a host that is called "primary"
...
+[domain@localhost:9990 /] attachment display --operation=:read-config-as-xml
```
fabiobrz, line 200: According to the analysis document we'd actually need to use `read-config-as-xml-as-file`, right?

docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
```diff
@@ -190,14 +190,19 @@ Assume you have a host that is called "primary"
...
+----
+
+If you want to save the file you can use:
+----
+[standalone@localhost:9990 /] attachment save --operation=:read-config-as-xml --file=/home/wildfly/tmp/standalone-runtime.xml
```
fabiobrz, line 205: According to the analysis document we'd actually need to use `read-config-as-xml-as-file`, right?

fabiobrz (changes requested):
Thanks for this PR @ehsavoie - I've dropped a couple of comments and suggestions.
Also please take the following one into account: https://docs.wildfly.org/33/Admin_Guide.html#operations here too the new operation should be mentioned.

docs/src/main/asciidoc/_admin-guide/CLI_Recipes.adoc
```diff
@@ -215,6 +230,8 @@ Just for the domain or standalone
 }
 ----
 
+Please note that the operation"
```
fabiobrz, line 233: Leftover? It seems the sentence which is started by these words has been added above already.

fabiobrz (changes requested):
Thanks @ehsavoie for addressing my remarks.
I am ok about not advertising `read-config-as-xml` and the deprecation note you added.
I added just one comment for what seems to be a leftover typo.

Also you might want to update the PR title in order to have it pass the CI checks, see https://github.com/wildfly/wildfly/pull/18159#issuecomment-2320730648

fabiobrz (approved):
Thanks @ehsavoie, for fixing the typos. Changes LGTM, approving.

bstansberry:
Thanks @ehsavoie and @fabiobrz

PR #18163, opened 2024-08-30
[WFLY-19695] Upgrade to Hibernate Search 7.1.2

marko-bekhta:
https://issues.redhat.com/browse/WFLY-19695

Just in case we won't get the Hibernate Search 7.2 into the next version, here's an update to the currently used 7.1. There are a couple of bug fixes in it (https://in.relation.to/2024/08/30/hibernate-search-maintenance-releases#other-since-last-7-1) that may be useful to the users.


bstansberry:
Thanks @marko-bekhta

PR #18167, opened 2024-09-04
Fix for WFLY-19702, Replace usage of wildfly-jar-maven-plugin with wildfly-maven-plugin

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-19702

Replaced usage in the tests of WildFly JAR Maven plugin with WildFly Maven plugin support for Bootable JAR.

### DIFF
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -1432,13 +1432,31 @@
             </plugin>
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
-                <artifactId>wildfly-jar-maven-plugin</artifactId>
+                <artifactId>wildfly-maven-plugin</artifactId>
+                <version>${version.org.wildfly.plugin}</version>
+                <configuration>
+                    <overwrite-provisioned-server>true</overwrite-provisioned-server>
+                    <record-provisioning-state>false</record-provisioning-state>
+                    <log-provisioning-time>${galleon.log.time}</log-provisioning-time>
+                    <offline-provisioning>${galleon.offline}</offline-provisioning>
+                    <galleon-options>
+                        <jboss-maven-dist/>
+                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                        <optional-packages>passive+</optional-packages>
+                    </galleon-options>
+                    <bootableJar>true</bootableJar>
+                    <skipDeployment>true</skipDeployment>
+                </configuration>
                 <executions>
                     <!-- Package a cloud-profile server with the default 'bootable-jar-packaging' execution -->
                     <execution>
                         <id>bootable-jar-packaging</id>
+                        <goals>
+                            <goal>package</goal>
+                        </goals>
+                        <phase>${ts.bootable-jar-packaging.phase}</phase>
                         <configuration>
-                            <output-file-name>test-wildfly-cloud-profile.jar</output-file-name>
+                            <bootableJarName>test-wildfly-cloud-profile.jar</bootableJarName>
                             <extra-server-content-dirs>
                                 <extraServerContent>${project.build.directory}/extraContent/</extraServerContent>
                             </extra-server-content-dirs>
@@ -1454,7 +1472,7 @@
                         </goals>
                         <phase>${ts.bootable-jar-packaging.phase}</phase>
                         <configuration>
-                            <output-file-name>test-wildfly-jpa.jar</output-file-name>
+                            <bootableJarName>test-wildfly-jpa.jar</bootableJarName>
                             <extra-server-content-dirs>
                                 <extraServerContent>${project.build.directory}/extraContent/</extraServerContent>
                             </extra-server-content-dirs>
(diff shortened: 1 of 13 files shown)


testsuite/integration/basic/pom.xml
```diff
@@ -1432,13 +1432,31 @@
             </plugin>
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
-                <artifactId>wildfly-jar-maven-plugin</artifactId>
+                <artifactId>wildfly-maven-plugin</artifactId>
+                <version>${version.org.wildfly.plugin}</version>
```
jamezp, line 1436: The version shouldn't be needed here. If it is, we should define it somewhere higher up the parent chain in the `<pluginManagement>` section.

testsuite/integration/basic/pom.xml
```diff
@@ -1432,13 +1432,31 @@
...
+                    <record-provisioning-state>false</record-provisioning-state>
+                    <log-provisioning-time>${galleon.log.time}</log-provisioning-time>
+                    <offline-provisioning>${galleon.offline}</offline-provisioning>
+                    <galleon-options>
+                        <jboss-maven-dist/>
+                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                        <optional-packages>passive+</optional-packages>
+                    </galleon-options>
+                    <bootableJar>true</bootableJar>
+                    <skipDeployment>true</skipDeployment>
```
jamezp, line 1448: This is somewhat related to the `wildfly-maven-plugin` more than this change. I assume this is to make this a hallow JAR. If that is the case, I think we need to update the description on the plugins documentation.

testsuite/integration/clustering/pom.xml
```diff
@@ -535,14 +553,28 @@
...
+                            </galleon-options>
+                            <bootableJar>true</bootableJar>
+                            <skipDeployment>true</skipDeployment>
+                            <bootableJarName>test-wildfly-jpa-distributed.jar</bootableJarName>
+                            <packaging-scripts>
                                 <cli-session>
-                                    <script-files>
+                                    <javaOpts>
+                                        <opt>-Dmaven.repo.local=${maven.repo.local}</opt>
+                                    </javaOpts>
```
jamezp, line 572: Again, this is more about the plugin. What makes this required? If it is, is this something we should look at adding to the plugin by default?

testsuite/integration/elytron/pom.xml
```diff
@@ -941,22 +917,45 @@
                     </plugin>
                     <plugin>
                         <groupId>org.wildfly.plugins</groupId>
-                        <artifactId>wildfly-jar-maven-plugin</artifactId>
+                        <artifactId>wildfly-maven-plugin</artifactId>
+                        <version>${version.org.wildfly.plugin}</version>
```
jamezp, line 921: Same note about the version here.

testsuite/integration/microprofile-tck/config/pom.xml
```diff
@@ -111,27 +111,38 @@
                             <goal>execute-commands</goal>
                         </goals>
                         <configuration>
+                            <offline>true</offline>
+                            <jboss-home>${basedir}/target/wildfly</jboss-home>
+                            <stdout>${project.build.directory}/wildfly/standalone/log/weld-configuration.log</stdout>
+                            <java-opts>${modular.jdk.args}</java-opts>
+                            <system-properties>
+                                <maven.repo.local>${maven.repo.local}</maven.repo.local>
```
jamezp, line 119: Is there a reason this is defined here and not in the `<java-opts>`? It seems we should be consistent if they achieve the same thing.

jamezp (changes requested):

testsuite/integration/basic/pom.xml
jfdenise: @jamezp , I followed existing pattern (the version is defined everywhere), introduced minimal changes. I could refactor it in this PR. Perhaps in an extra commit.

testsuite/integration/microprofile-tck/config/pom.xml
jfdenise: @jamezp This is the `execute-commands` goal. System-properties maven-repo-local is already set that way (I just moved the `<configuration>` element.
 The `package` goal has a different `<packaging-scripts>` element that defines the `<java-opts>`.

testsuite/integration/clustering/pom.xml
jfdenise: That is not strictly needed there (I didn't run into issue without it). But that is something to generalize. We are starting an embedded server from a slim installation that resolves JBoss Modules modules maven artifact from the maven local cache.

testsuite/integration/basic/pom.xml
jfdenise: Good point, I logged https://issues.redhat.com/browse/WFMP-270
BTW, you can create an Hollow JAR without this property, but you will get a Warning (Deployment not found). It is cleaner to generalize that and document it (as you suggested).

jamezp (approved):
I think this looks good. I think later we might be able to clean up some of the verbosity, but that feels like a follow up IMO.

bstansberry:
Thanks @jfdenise and @jamezp

PR #18188, opened 2024-09-13
WFLY-19729 *TxClientEJBForwardingTestCase fail while trying to configure legacy security-realm and with security manager

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-19729

### DIFF
--- a/testsuite/shared/src/main/resources/application-users.properties
+++ b/testsuite/shared/src/main/resources/application-users.properties
@@ -6,7 +6,7 @@
 # guest=guest
 # user1=password1
 # user2=password2
-#remoteejbuser=rem@teejbpasswd1
+# remoteejbuser=rem@teejbpasswd1
 guest=b5d048a237bfd2874b6928e1f37ee15e
 user1=23624d2f74dfcb9688651a066d90b97e
 user2=ab3f9e12039435236d89de9023a304b7
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
@@ -187,6 +187,10 @@ public abstract class AbstractClusteringTestCase {
     }
 
     // Node and deployment lifecycle management convenience methods
+
+    /**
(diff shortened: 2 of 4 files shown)


rhusar:
@tadamski @rachmatowicz Can you guys review the updated security configuration in the `ServerSetupTask` or should I ask someone from the security team? Thanks

rhusar:
Hi @fjuma could you or someone from your team please review the elytron configuration bit[^1] of the PR? Much appreciated!

[^1]: specifically lines https://github.com/wildfly/wildfly/pull/18188/files#diff-63ef7e85a9f1b9907ab663d0c705cecc84a3452cfc8f8519f01497f9a230583cL243-R268

rhusar:
Happy 1 month 🍰  anniversary to this PR!

testsuite/shared/src/main/resources/application-users.properties
```diff
@@ -6,7 +6,7 @@
 # guest=guest
 # user1=password1
 # user2=password2
-#remoteejbuser=rem@teejbpasswd1
+# remoteejbuser=rem@teejbpasswd1
```
darranl, line 9: Is there a reason for this additional space?  I don't see where this line gets uncommented.

darranl (changes requested):
The security changes do look good but I am not sure is this test ignored on CI so we don't have real results here.

rhusar:
> The security changes do look good but I am not sure is this test ignored on CI so we don't have real results here.

That's correct. This test is marked as intermittently failing, that's why it's not running in CI.

However, it's not failing intermittently any more - but consistently hence this PR. 

This PR will make the test  runnable again when intermittent tests are enabled. Then the investigation of intermittent failures can continue.

testsuite/shared/src/main/resources/application-users.properties
rhusar: The reason legibility and to follow the same convention as the example `# login=password` above.

It doesn't get uncommented since the password is already encoded on line 13. This is un-encoded version of that password.

darranl: Thank you, GitHub cut off at line 12 - I see now.

darranl (approved):
Thank you for the information, I will merge as is so investigation can continue.

rhusar:
Thanks @darranl !

PR #18194, opened 2024-09-16
[WFLY-19445] an element addition, thus, a version update to the schem…

rsearls:
https://issues.redhat.com/browse/WFLY-19445

This XSD file addition is in preparation for code additions in module 
elytron-oidc-client to use the new element.

### DIFF
--- a/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
+++ b/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemModel.java
@@ -18,8 +18,9 @@ enum ElytronOidcClientSubsystemModel implements SubsystemModel {
     VERSION_2_0_0(2, 0, 0),
     VERSION_3_0_0(3, 0, 0), // WildFly 32.0-present
     VERSION_4_0_0(4, 0, 0), // WildFly 33.0-present
+    VERSION_5_0_0(5, 0, 0), // WildFly 34.0-present
     ;
-    static final ElytronOidcClientSubsystemModel CURRENT = VERSION_4_0_0;
+    static final ElytronOidcClientSubsystemModel CURRENT = VERSION_5_0_0;
 
     private final ModelVersion version;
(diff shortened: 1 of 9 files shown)


fjuma:
@rsearls Thanks for creating this PR for the subsystem bump. In general, it's good to include all the changes needed for the bump in addition to the xsd file. As an example, take a look at some previous subsystem bump PRs (e.g., https://github.com/wildfly/wildfly-core/pull/5999/commits/ceccfb627cde46809dc06f7dd8c334239c01cb6c).

This PR should also be a draft PR since we don't want to merge this until we know that we have an RFE that's ready to be merged (i.e., we've gone through the feature development process) that requires this bump.

bstansberry:
@rsearls A further note to what @fjuma said -- when introducing a new schema or a new management API version, do a commit that does all the boilerplate stuff (a la https://github.com/wildfly/wildfly-core/pull/5999/commits/ceccfb627cde46809dc06f7dd8c334239c01cb6c) but where the new XSD is no different from the old (except xmlns etc). And then a second commit that changes the new schema (and/or the management API classes if those are changing) to bring in whatever is different.

This allows reviewers and future maintainers to easily understand what happened. The boilerplate commit is easy to review as it's just boilerplate familiar to reviewers. And the substantive change is easy to see because there's a git diff that's only about that.

rsearls:
@bstansberry So are you saying in this case submit a file, wildfly-elytron-oidc-client_preview_3_0.xsd with only the element changes
and in the future submit file wildfly-elytron-oidc-client_preview_4_0.xsd  with the version changes?

fjuma:
@rsearls You'll want to introduce a wildfly-elytron-oidc-client_preview_4_0.xsd file that is a copy of the current wildfly-elytron-oidc-client_preview_3_0.xsd file without your new element added. Then in your changes for your RFE, you'll introduce your new element. That helps us to easily see the changes introduced by your RFE and it also makes it possible for multiple people to make use of your subsystem bump commit if they are also working on an RFE with configuration changes.

rsearls:
2nd attempt at new schema components

PR #18203, opened 2024-09-17
The preview boms should not depend on the standard boms

bstansberry:
Just an experiment to see what CI says.


bstansberry:
I ran the WFP-focused set of jobs against this by filing https://github.com/wildfly/wildfly/pull/18204. They all passed.

PR #18227, opened 2024-09-23
[WFLY-19397] Jakarta Data support in WildFly Preview

bstansberry:
https://issues.redhat.com/browse/WFLY-19397

@scottmarlow Doing this is 97% boilerplate, so since I'm sadly responsible for most of the boilerplate details I took a shot at this while watching sports yesterday.

It's a new jakarta.data.api module, a minimal subsystem and a DUP to add the api to the deployment. Plus a galleon layer to allow adding the extension/subsystem. The new module, subsystem and layer are only integrated in WildFly Preview.

We need a subsystem so we can evolve any configuration or add any /deployment=foo.war/subsystem=jakarta-data stuff in the future without breaking people who were accustomed to not needing one. Plus it's just the normal way of adding technologies.

I'll spend a bit of time today trying to port https://github.com/jakartaredhat/hibernate-data-tck/tree/main/runner-web to wildfly-tck-runners. The TCK would be the primary way this would be tested initially.

### DIFF
--- a/boms/preview-ee/pom.xml
+++ b/boms/preview-ee/pom.xml
@@ -110,6 +110,17 @@
                 </exclusions>
             </dependency>
             <dependency>
+                <groupId>jakarta.data</groupId>
+                <artifactId>jakarta.data-api</artifactId>
+                <version>${version.jakarta.data.jakarta-data-api}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
                 <groupId>jakarta.el</groupId>
                 <artifactId>jakarta.el-api</artifactId>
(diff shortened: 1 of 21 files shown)


darranl (changes requested):
FYI the test failure does relate to the new extension.

bstansberry:
@darranl That should be fixed now.

bstansberry:
@darranl @scottmarlow @jamezp A build of this branch passes the Jakarta Data TCK when I use the proposed runner at https://github.com/wildfly/wildfly-tck-runners/pull/184.  So I've moved this out of draft state.

For 34 Final this needs a smoke test, but I don't have time today so I focused on getting the TCK to run and pass, as that's a better proof of the basic validity of the integration for 34 Beta.

@gavinking @starksm64 FYI

jamezp (approved):

gavinking:
Great to see

darranl:
@bstansberry I have a fix for the failure, sending a replacement PR

PR #18235, opened 2024-09-24
[WFLY-19397] Jakarta Data support in WildFly Preview

darranl:
Supersedes https://github.com/wildfly/wildfly/pull/18227

https://issues.redhat.com/browse/WFLY-19397

@scottmarlow Doing this is 97% boilerplate, so since I'm sadly responsible for most of the boilerplate details I took a shot at this while watching sports yesterday.

It's a new jakarta.data.api module, a minimal subsystem and a DUP to add the api to the deployment. Plus a galleon layer to allow adding the extension/subsystem. The new module, subsystem and layer are only integrated in WildFly Preview.

We need a subsystem so we can evolve any configuration or add any /deployment=foo.war/subsystem=jakarta-data stuff in the future without breaking people who were accustomed to not needing one. Plus it's just the normal way of adding technologies.

I'll spend a bit of time today trying to port https://github.com/jakartaredhat/hibernate-data-tck/tree/main/runner-web to wildfly-tck-runners. The TCK would be the primary way this would be tested initially.

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
@@ -23,6 +23,8 @@
         <module name="com.fasterxml.jackson.core.jackson-core" optional="true"/>
         <module name="com.fasterxml.jackson.core.jackson-databind" optional="true"/>
         <module name="jakarta.annotation.api"/>
+        <!-- If Jakarta Data is provisioned, depend on it -->
+        <module name="jakarta.data.api" optional="true"/>
         <module name="jakarta.enterprise.api"/>
         <module name="jakarta.json.bind.api"/>
         <module name="jakarta.persistence.api"/>
--- a/preview/galleon-local/src/main/resources/layers/standalone/jakarta-data/layer-spec.xml
+++ b/preview/galleon-local/src/main/resources/layers/standalone/jakarta-data/layer-spec.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" ?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:2.0" name="jakarta-data">
+    <props>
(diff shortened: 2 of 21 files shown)


darranl:
@bstansberry The last commit adds some changes to get host excludes testing working again.

I am not 100% sure how we are supposed to be handling these preview extensions, can you please check if this is correct?

bstansberry (approved):

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
```diff
@@ -23,6 +23,8 @@
         <module name="com.fasterxml.jackson.core.jackson-core" optional="true"/>
         <module name="com.fasterxml.jackson.core.jackson-databind" optional="true"/>
         <module name="jakarta.annotation.api"/>
+        <!-- If Jakarta Data is provisioned, depend on it -->
+        <module name="jakarta.data.api" optional="true"/>
```
scottmarlow, line 27: This makes sense to include!

preview/galleon-local/src/main/resources/layers/standalone/jakarta-data/layer-spec.xml
```diff
@@ -0,0 +1,18 @@
...
+
+<layer-spec xmlns="urn:jboss:galleon:layer-spec:2.0" name="jakarta-data">
+    <props>
+        <prop name="org.wildfly.rule.annotations" value="jakarta.data"/>
+        <prop name="org.wildfly.rule.class" value="jakarta.data"/>
+    </props>
+    <dependencies>
+        <!-- TODO one or the other of jpa or jpa-distributed is required,
+             but I don't think that can be declared -->
+        <layer name="jpa" optional="true"/>
```
scottmarlow, line 15: :+1:

scottmarlow (approved):

scottmarlow:
Thanks @bstansberry + @darranl !

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/main/module.xml
bstansberry: I believe it's required for Jakarta Data to work.

bstansberry:
Thanks @darranl and @scottmarlow

PR #18250, opened 2024-09-27
[WFLY-19759] Test MicroProfile Reactive messaging multiple deployments

kabir:
https://issues.redhat.com/browse/WFLY-19759

If deployed within an .ear, the channel names need to be unique within that .ear.

If deployed as separate top-level archives, the channel names don't need to be unique.


bstansberry:
Thanks @kabir

PR #18259, opened 2024-10-01
[WFLY-19796] skip programmatic test module creation in bootable jar tests

simkam:
https://issues.redhat.com/browse/WFLY-19796

Tests create tests module in {{module.path}} which is not used by bootable jar at all and it also leads outside of testsuite ([WFLY-19795](https://issues.redhat.com/browse/WFLY-19795)).

\+ simplified `DeploymentArchiveTestCase` so it can be used with bootable jar without a full WildFly build


jbliznak (approved):

bstansberry:
@simkam @jbliznak I'm reluctant re the 2nd test as we don't have a lot of testing of CLI module add/remove, and perhaps none in the context of a .cli file where the module's artifact and config file are package in the .cli archive.

I think the WFLY-19795 concern with that one can be addressed by removing the custom getModulePath and using TestModule.getModulesDirectory(). That method has logic that's meant to guard against writing in build/dist. If that causes the test to fail it means we're not properly configuring surefire.

That leaves the "it's pointless in bootable jar" issue. That is valid but I don't know if it outweighs the loss of coverage.

simkam:
@bstansberry Thanks for pointing out `TestModule` tooling class. I included [WFLY-19795](https://issues.redhat.com/browse/WFLY-19795) fix in this MR. 

I made operations with modules in `DeploymentArchiveTestCase` conditional, skipped with bootablejar. I added comment
```
           // "module add/remove" operations doesn't make sense in bootablejar
            // CLI isn't included in bootablejar and "module" is local command. It operates with module on its own
            // MODULEPATH. It could be used with bootable jar with workaround:
            //   $ JBOSS_HOME=<bootablejar_install_dir> jboss.cli.sh -c
            // install-dir is by default folder with random suffix in /tmp
```
I didn't find a way, how to change CLI's `JBOSS_HOME` or `JBOSS_MODULEPATH` with `CLIWrapper`.

jbliznak (approved):

bstansberry (approved):

bstansberry:
Thanks @simkam and @jbliznak

PR #18273, opened 2024-10-04
[WFLY-19815] Make ProvisioningConsistencyBaseTest more resilent to ch…

yersan:
…anges on project build dir names

Jira issue: https://issues.redhat.com/browse/WFLY-19815

### DIFF
--- a/testsuite/integration/manualmode-expansion/pom.xml
+++ b/testsuite/integration/manualmode-expansion/pom.xml
@@ -370,7 +370,8 @@
                         <node2>${node2}</node2>
                         <node3>${node3}</node3>
                         <mcast>${mcast}</mcast>
-                        <standard.dist.version>${project.version}</standard.dist.version>
+                        <dist.output.dir>dist/target/${server.output.dir.prefix}-${server.output.dir.version}</dist.output.dir>
+                        <build.output.dir>${wildfly.build.output.dir}</build.output.dir>
                     </systemPropertyVariables>
                 </configuration>
                 <executions>
--- a/testsuite/integration/manualmode/pom.xml
+++ b/testsuite/integration/manualmode/pom.xml
@@ -565,7 +565,8 @@
                         <mcast>${mcast}</mcast>
                         <multichannel.jboss.home>${multichannel.jboss.home}</multichannel.jboss.home>
                         <oidc.jboss.home>${oidc.jboss.home}</oidc.jboss.home>
-                        <standard.dist.version>${project.version}</standard.dist.version>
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -232,6 +232,7 @@
              Custom test jobs can override this using -D, e.g. -Dtestsuite.default.build.project.prefix=ee- -->
         <testsuite.default.build.project.prefix></testsuite.default.build.project.prefix>
         <wildfly.build.output.dir>${testsuite.default.build.project.prefix}build/target/${server.output.dir.prefix}-${server.output.dir.version}</wildfly.build.output.dir>
+        <wildfly.dist.output.dir>${testsuite.default.build.project.prefix}dist/target/${server.output.dir.prefix}-${server.output.dir.version}</wildfly.dist.output.dir>
```
bstansberry, line 235: I don't see this used anywhere.

testsuite/integration/manualmode-expansion/pom.xml
```diff
@@ -370,7 +370,8 @@
                         <node2>${node2}</node2>
                         <node3>${node3}</node3>
                         <mcast>${mcast}</mcast>
-                        <standard.dist.version>${project.version}</standard.dist.version>
+                        <dist.output.dir>dist/target/${server.output.dir.prefix}-${server.output.dir.version}</dist.output.dir>
```
bstansberry, line 373: Was it supposed to be used here?

testsuite/integration/manualmode/pom.xml
```diff
@@ -565,7 +565,8 @@
                         <mcast>${mcast}</mcast>
                         <multichannel.jboss.home>${multichannel.jboss.home}</multichannel.jboss.home>
                         <oidc.jboss.home>${oidc.jboss.home}</oidc.jboss.home>
-                        <standard.dist.version>${project.version}</standard.dist.version>
+                        <ee.dist.output.dir>ee-dist/target/${server.output.dir.prefix}-${server.output.dir.version}</ee.dist.output.dir>
```
bstansberry, line 568: and here?

bstansberry (changes requested):
See my previous comments. I think either something is unimplemented, or there's a cruft property left over from an earlier approach that you moved away from.

pom.xml
yersan: Deleted, yes, it was unused.

I was left as cruft from development I forgot to remove; I thought I was going to need to use it but then I realized it was not necessary since `manualmode/pom.xml` and` manualmode-expansion/pom.xml` would run independently of this dynamic value (based on `testsuite.default.build.project.prefix`)

bstansberry:
Thanks @yersan

PR #18283, opened 2024-10-08
[WFLY-19609]:Remove wss4j dependency and doesn't export saaj and jaxb…

jimma:
https://issues.redhat.com/browse/WFLY-19609


bstansberry:
Thanks @jimma

PR #18287, opened 2024-10-08
[WFLY-19831] unstable-api-annotation-index package should be provisio…

yersan:
…ned only when preview stability level is enabled

Jira issue: https://issues.redhat.com/browse/WFLY-19831

The tricky part is that dist/build verifications depend on the galleon-pack stability level configuration, so as soon as we change the galleon-pack provisioning all to "community", these verification files will fail since this annotation-index package won't be provisioned. I think it is ok since what we are verifying is what we are provisioning, although making them more dynamic would be better, I guess it could be done with multiple properties (per galleon pack) in the main pom.xml

### DIFF
--- a/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
+++ b/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
@@ -26,6 +26,6 @@
         <package name="cleanup.domain.log.dir" optional="true"/>
         <package name="cleanup.domain.data.dir" optional="true"/>
         <!-- Unstable API Annotation Index -->
-        <package name="unstable-api-annotation-index.wildfly-ee-feature-pack"/>
+        <package name="unstable-api-annotation-index.wildfly-ee-feature-pack" optional="true" valid-for-stability="preview"/>
     </packages>
 </config>
--- a/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
+++ b/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
@@ -27,5 +27,7 @@
         <package name="cleanup.domain.data.dir" optional="true"/>
         <!-- Unstable API Annotation Index -->
         <package name="unstable-api-annotation-index.wildfly-ee-feature-pack" optional="true" valid-for-stability="preview"/>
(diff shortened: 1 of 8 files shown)


bstansberry:
@yersan I don't understand your 'tricky part' comment. Please post some code links or something to illustrate.

The change itself seems ok, but I don't understand the broader context you're discussing.

yersan:
@bstansberry sure, let me try to explain it:

My comment comes from the fact that we are running the `maven-verifier-plugin` for the dist/build maven modules, and these verifications depend on the Galleon Feature Pack stability level configuration. So, as soon as we change the WildFly Galleon Pack stability levels, we will need to change the verifications we are doing, for example, when we move the server from `community` to `default` stability levels. Or even with lower stability levels.

For example, here we are verifying that the annotation index package is provisioned:

https://github.com/wildfly/wildfly/blob/main/dist/src/verifier/verifications.xml#L36-L39

As soon as we change the WildFly Feature Pack `package-stability-level` configuration from `experimental` to `community`, we will need to adjust the `verifications.xml` file to assert the new provisioning. Ideally, a change in the WildFly Feature Pack stability levels should automatically change the `verifications.xml` assertions. That would make it easier to move the server configuration from one stability level to another. 

And, I just wanted to mention that "tricky" possibility I did not investigate further, because I think it is ok since what we are verifying is what we are provisioning, although making them more dynamic would be better.

bstansberry:
@yersan Ok, thanks. I got confused because I looked at the 'build' module verifications.xml, and not 'dist'. And 'build' doesn't include the verifications you're talking about. So then I wasn't confident I understood what you meant.

We can also revert https://github.com/wildfly/wildfly/commit/477d928d37af1b6b2abe90810ac1f901095a6ca6 and update the restored test to do a check of  /core-service=server-environment 'permissible-stability-levels' and throw AssumptionViolatedException. I suggested using verfications.xml but perhaps that was a bad idea.

bstansberry (dismissed):

yersan:
Windows - JDK 17 is unrelated to this. @bstansberry just FYI, this one is blocking the WildFly Core counterpart, I am done with this

kabir (approved):

bstansberry:
Thanks @yersan

PR #18290, opened 2024-10-09
[WFLY-19780]:webservice endpoint packaged in the ROOT.war should be d…

jimma:
…eployed under '/' root path

https://issues.redhat.com/browse/WFLY-19780


webservices/server-integration/src/main/java/org/jboss/as/webservices/tomcat/WebMetaDataModifier.java
```diff
@@ -91,6 +92,9 @@ private void configureEndpoints(final Deployment dep, final JBossWebMetaData jbo
      * @param jbossWebMD web meta data
      */
     private void modifyContextRoot(final Deployment dep, final JBossWebMetaData jbossWebMD) {
+        if (DefaultDeploymentMappingProvider.instance().getMapping(dep.getSimpleName()) != null) {
```
bstansberry, line 95: @fl4via @pferraro Please review.

I understand DefaultDeploymentMappingProvider  well enough to know if this use of it is valid. For example it just checks for an entry and ignores the value.

PR #18292, opened 2024-10-11
WFLY-19592 [MP Platform 7] MicroProfile Fault Tolerance 4.1 in WildFly Preview

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-19592

### DIFF
--- a/boms/preview-expansion/pom.xml
+++ b/boms/preview-expansion/pom.xml
@@ -34,8 +34,9 @@
         <version.io.opentelemetry.instrumentation>2.5.0</version.io.opentelemetry.instrumentation>
         <version.io.opentelemetry.opentelemetry>1.39.0</version.io.opentelemetry.opentelemetry>
         <version.io.opentelemetry.opentelemetry-semconv>1.25.0-alpha</version.io.opentelemetry.opentelemetry-semconv>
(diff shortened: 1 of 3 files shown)


bstansberry (approved):

rhusar:
@bstansberry Do you think it's worthwhile having this in main in preview or should we close this and reopen against our MP 7 branch instead (having this in proper)?

bstansberry:
@rhusar If you don't mind and it passes CI (which I requested just as a refresh) then I'd rather merge it and let it bake in main, where it will get a lot more testing.

I think that only because you already did the work. If it's a hassle for you to migrate from this to what you'll do in mp-7-wip instead of starting from current main, that's fine; just close it.

If it passes I'll merge tomorrow unless you say not to.

rhusar:
> @rhusar If you don't mind and it passes CI (which I requested just as a refresh) then I'd rather merge it and let it bake in main, where it will get a lot more testing.
> 
> I think that only because you already did the work. If it's a hassle for you to migrate from this to what you'll do in mp-7-wip instead of starting from current main, that's fine; just close it.
> 
> If it passes I'll merge tomorrow unless you say not to.

No problem here - you are right, the more bake and visibility, the better!

bstansberry:
Thanks @rhusar

PR #18295, opened 2024-10-12
WFLY-19806 Restore singleton MDBs to working state (Take II)

pferraro:
https://issues.redhat.com/browse/WFLY-19806
Resolves underlying issue:
https://issues.redhat.com/browse/WFLY-19849

Verified via quickstart (due to lack of integration tests).

### DIFF
--- a/clustering/singleton/server/src/main/java/org/wildfly/clustering/singleton/server/DistributedSingletonServiceBuilder.java
+++ b/clustering/singleton/server/src/main/java/org/wildfly/clustering/singleton/server/DistributedSingletonServiceBuilder.java
@@ -8,6 +8,7 @@ package org.wildfly.clustering.singleton.server;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -29,6 +30,7 @@ import org.wildfly.service.AsyncServiceBuilder;
  */
 public class DistributedSingletonServiceBuilder<T> extends AbstractSingletonServiceBuilder<T> {
 
+    private final ServiceName id = ServiceName.parse(UUID.randomUUID().toString());
     private final Function<ServiceTarget, ServiceBuilder<?>> builderFactory;
     private final List<Map.Entry<ServiceName[], Consumer<Consumer<?>>>> injectors = new LinkedList<>();
     private final SingletonServiceBuilderContext context;
@@ -39,7 +41,8 @@ public class DistributedSingletonServiceBuilder<T> extends AbstractSingletonServ
         super(new AsyncServiceBuilder<>(builder, builder.requires(ServiceName.parse("org.wildfly.management.executor"))), context);
         this.builderFactory = builderFactory;
         this.context = context;
-        this.singleton = singletonFactory.apply(builder).andThen(this.reference);
+        Consumer<Singleton> injector = builder.provides(this.id);
+        this.singleton = singletonFactory.apply(builder).andThen(this.reference).andThen(injector);
     }
 
     @SuppressWarnings("unchecked")
@@ -59,17 +62,14 @@ public class DistributedSingletonServiceBuilder<T> extends AbstractSingletonServ
 
     @Override
     public SingletonServiceBuilder<T> setInstance(Service service) {
-        this.context.setServiceName(ServiceName.parse(service.getClass().getName()));
         this.context.setService(service);
         return this;
     }
 
     @Override
     public SingletonServiceController<T> install() {
-        ServiceName name = this.context.getServiceName();
-        if (name == null) {
-            throw new IllegalStateException();
-        }
+        // Identify singleton service via its id if caller did not provide one
+        this.context.setServiceName(this.id);
         ServiceController<T> controller = this.getDelegate().setInstance(new DistributedSingletonService(this.context, this.builderFactory, this.injectors, this.singleton)).install();
         return new DistributedSingletonServiceController<>(controller, this.reference);
     }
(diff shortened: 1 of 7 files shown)


clustering/singleton/server/src/main/java/org/wildfly/clustering/singleton/server/DistributedSingletonServiceBuilder.java
```diff
@@ -59,17 +62,14 @@ public void accept(V value) {
...
     }
 
     @Override
     public SingletonServiceController<T> install() {
-        ServiceName name = this.context.getServiceName();
-        if (name == null) {
-            throw new IllegalStateException();
-        }
+        // Identify singleton service via its id if caller did not provide one
+        this.context.setServiceName(this.id);
```
bstansberry, line 72: Note to reviewers/maintainers: The 'if caller did not provide one' bit is in the setServiceName implementation.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -594,13 +597,24 @@ private static void addClusteringServices(final OperationContext context, final
...
+                    // We don't want this to start until a deployment requires it
+                    ServiceController<?> controller = org.wildfly.service.ServiceInstaller.builder(Functions.constantSupplier(Boolean.TRUE))
+                            .provides(CLUSTERED_SINGLETON_CAPABILITY.getCapabilityServiceName())
+                            .build()
+                            .install(targetFactory.get().createSingletonServiceTarget(target));
+
+                    // Install well-known on-demand service that will force singleton service to start.
+                    ServiceInstaller.builder(Functions.constantSupplier(Boolean.TRUE))
+                            .provides(ServiceNameFactory.resolveServiceName(CLUSTERED_SINGLETON_BARRIER))
+                            .requires(controller.provides().stream().map(ServiceDependency::on).collect(Collectors.toList()))
```
bstansberry, line 610: This could use a comment. Just reading the nearby code 'controller' provides one ServiceName but here there's complex stream processing happening.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -594,13 +597,24 @@ private static void addClusteringServices(final OperationContext context, final
             ServiceInstaller installer = new ServiceInstaller() {
                 @Override
                 public ServiceController<?> install(RequirementServiceTarget target) {
-                    ServiceBuilder<?> builder = targetFactory.get().createSingletonServiceTarget(target).addService();
-                    return builder.setInstance(org.jboss.msc.Service.newInstance(builder.provides(CLUSTERED_SINGLETON_CAPABILITY.getCapabilityServiceName()), Boolean.TRUE))
-                            .setInitialMode(ServiceController.Mode.ON_DEMAND)
-                            .install();
+                    // Install on-demand singleton service
+                    // We don't want this to start until a deployment requires it
```
bstansberry, line 601: Note to reviewers/maintainers: Unlike traditional MSC APIs, the wildfly-service stuff creates ON_DEMAND services by default. That's why there's no code making these on-demand to satisfy what the comment says.

bstansberry:
@pferraro Based on the JIRAs I added the Blocker and 34.x labels.

Please make sure @tadamski reviews this Monday AM. I don't want delays.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
pferraro: Added.

tadamski (commented):
@bstansberry @pferraro Sorry, it took me some time to get a grasp of this new functional Service API that I wasn't working out very much with, but even after that I'm not sure I have full understanding of how this PR works. My understanding is that what this PR does is assigning another ServiceName to clustered singleton service which is unique ID and, after this service is started on demand, installs a barrier service with dependencies on this unique ID. But TBH I'm not sure if I understand how this barrier service is supposed  to interact with other nodes in the cluster and make sure that other instance of this service are not started on them.

pferraro:
@tadamski Your understanding is correct.  Where previously (WF33), singleton services were named (i.e. ServiceController.getServiceName() != null), they are now anonymous (i.e. ServiceController.getServiceName() == null).  When I made this change I did not account for Singleton MDBs, which is the only use case that installs an on-demand singleton service.  In order for the BARRIER service to be able to force the singleton service to start, it must be able to require some ServiceName provided by the singleton service.  The WLY-19849 fix ensures that its ServiceController.provides() returns a non-empty set, so that the BARRIER service can require it (and thus force it to start).

tadamski:
@pferraro thanks for an explanation, but honestly looking further into that I still don't understand all stuff that's happening. Specifically I don't understand the presence and the behavior of ServiceTargetFactory.DEFAULT_SERVICE_DESCRIPTOR capability. My first question would be why the line mentioned in the comment didn't worked before the fix?
{code}
serviceTarget.addDependency(capabilityServiceSupport.getCapabilityServiceName(ServiceTargetFactory.DEFAULT_SERVICE_DESCRIPTOR));
{code} 
I can't see what it has to do with Singleton service name being null?

pferraro:
> @pferraro thanks for an explanation, but honestly looking further into that I still don't understand all stuff that's happening. Specifically I don't understand the presence and the behavior of ServiceTargetFactory.DEFAULT_SERVICE_DESCRIPTOR capability. 

ServiceTargetFactory provides a mechanism for decorating a ServiceTarget, in this case, with an implementation that instrument singleton services.

> My first question would be why the line mentioned in the comment didn't worked before the fix?
> 
> ```
> serviceTarget.addDependency(capabilityServiceSupport.getCapabilityServiceName(ServiceTargetFactory.DEFAULT_SERVICE_DESCRIPTOR));
> ```

Because this only forces the ServiceTargetFactory to start, not the singleton service itself.

> I can't see what it has to do with Singleton service name being null?

Previously (in WF33), the EJB subsystem used the older singleton service installation API (based on org.wildfly:wildfly-clustering-service) that created named services and thus an on-demand singleton service could be forced to start by depending on its ServiceController.getServiceName().  The new singleton service installation API (based on org.wildfly.core:wildfly-service) creates anonymous services, where ServiceController.getServiceName() is always null.

tadamski (approved):
@pferraro thanks for the thorough explanation; it all makes sense to me know and looking again at the PR I don't have any changes suggestion

bstansberry (approved):

bstansberry:

PR #18310, opened 2024-10-18
WFLY-19867 Upgrade to Hibernate Commons Annotations 7.0.3.Final

yrodiere:
https://issues.redhat.com/browse/WFLY-19867

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -528,7 +528,7 @@
         <version.org.glassfish.jaxb.jaxb-xjc>${version.org.glassfish.jaxb}</version.org.glassfish.jaxb.jaxb-xjc>
         <version.org.glassfish.soteria>3.0.3</version.org.glassfish.soteria>
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.1-jbossorg-1</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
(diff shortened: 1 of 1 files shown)


scottmarlow (approved):

bstansberry:
@yrodiere Looking at the CI failures, it seems this binary was compiled to produce SE 17 binaries, not SE 11.

We plan to move WildFly main to an SE 17 base shortly, so we'll be able to consume this for WF 35 then, but other uses of 7.0.x will not be able to move to this version.

As you noted in the WFLY-19867 description, 7.0.2 was not meant to introduce runtime behavior changes, so not being able to move to it won't harm anyone. So it's more an issue if you ever wanted to fix bugs in a future 7.0.x.

yrodiere:
@bstansberry This binary was definitely intended to target JDK 11; the release job even [uses JDK 11](https://github.com/hibernate/hibernate-commons-annotations/blob/6a076ce58b6182a343a499218e375b6c0f54ad4a/ci/release/Jenkinsfile#L60)...

I suspect something is wrong with the CI setup and/or Gradle (which might be starting its daemon using a different JDK).
Anyway, I'll pin the release target in the Gradle config and do another release...

yrodiere:
I changed this PR to upgrade to 7.0.3.Final, which ships binaries with correct bytecode (Java 11).

bstansberry:
Thanks @yrodiere

PR #18313, opened 2024-10-21
[WFLY-19393] Persistence container bytecode enhancement should be enabled by default

scottmarlow:
https://issues.redhat.com/browse/WFLY-19393

https://hibernate.atlassian.net/browse/HHH-16572 was addressed in Hibernate ORM 6.6.2.Final and this (WFLY-19393) change enables bytecode enhancement for Persistence application deployments by default.

Note that we now expect Jakarta EE 10 TCK test jpa/core/query/flushmode (https://github.com/jakartaee/platform-tck/blob/master/src/com/sun/ts/tests/jpa/core/query/flushmode/Client.java#L355 #flushModeTest5 to pass (with ByteCode enhancement enabled by default) since [HHH-16572](https://hibernate.atlassian.net/browse/HHH-16572) was addressed.

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
@@ -7,7 +7,6 @@ package org.jboss.as.jpa.config;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 import jakarta.persistence.EntityManagerFactory;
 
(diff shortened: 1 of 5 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADelegatingClassFileTransformer.java
```diff
@@ -42,7 +43,10 @@ public ByteBuffer transform(ClassLoader classLoader, String className, Protectio
                     // run as security privileged action
                     @Override
                     public ByteBuffer run() {
-
+                        if(Configuration.ignoreClassTransform(className)) {
+                            // indicate that class change was not made as workaround for https://hibernate.atlassian.net/browse/HHH-16572
```
scottmarlow, line 47: Will create a separate issue tracker for removing the call to check Configuration.ignoreClassTransform(className) if/when HHH-16572 is fixed (when possible).

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #18316, opened 2024-10-22
[WFLY-19871] Adds Channel configuration to user BOMs

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19871

### DIFF
--- a/boms/user/client/ejb-client/pom.xml
+++ b/boms/user/client/ejb-client/pom.xml
@@ -47,6 +47,12 @@
             <version>${ee.maven.version}</version>
             <type>pom</type>
         </dependency>
+        <dependency>
+            <groupId>org.wildfly.channels</groupId>
+            <artifactId>wildfly-ee</artifactId>
+            <version>${ee.maven.version}</version>
+            <type>pom</type>
+        </dependency>
     </dependencies>
 
     <build>
@@ -74,6 +80,14 @@
                             <bomWithDependencies>true</bomWithDependencies>
                             <licenses>true</licenses>
                             <inheritExclusions>UNMANAGED</inheritExclusions>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly-ee</artifactId>
+                                    </manifest>
+                                </channel>
+                            </channels>
                             <includeDependencies>
                                 <!-- ejb client, the main artifact -->
                                 <dependency>
--- a/boms/user/server/ee/preview/pom.xml
+++ b/boms/user/server/ee/preview/pom.xml
@@ -43,6 +43,13 @@
         <dependency>
             <groupId>org.wildfly</groupId>
             <artifactId>wildfly-preview-feature-pack</artifactId>
+            <version>${ee.maven.version}</version>
+            <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.channels</groupId>
+            <artifactId>wildfly-preview</artifactId>
+            <version>${ee.maven.version}</version>
             <type>pom</type>
         </dependency>
     </dependencies>
@@ -70,6 +77,14 @@
                             <bomVersion>${project.version}</bomVersion>
                             <bomName>${full.dist.product.release.name} Preview BOMs: EE</bomName>
                             <bomDescription>${full.dist.product.release.name} Preview: EE Dependency Management</bomDescription>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly-preview</artifactId>
+                                    </manifest>
+                                </channel>
+                            </channels>
                             <includeDependencies combine.children="append">
                                 <!-- include here any <dependency /> exclusive to this BOM -->
                                 <dependency>
--- a/boms/user/server/expansion/preview/pom.xml
+++ b/boms/user/server/expansion/preview/pom.xml
@@ -38,11 +38,18 @@
         </dependencies>
     </dependencyManagement>
 
-    <!-- enforces all BOM artifacts are available -->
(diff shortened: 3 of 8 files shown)


boms/user/server/ee/preview/pom.xml
```diff
@@ -70,6 +83,20 @@
                             <bomVersion>${project.version}</bomVersion>
                             <bomName>${full.dist.product.release.name} Preview BOMs: EE</bomName>
                             <bomDescription>${full.dist.product.release.name} Preview: EE Dependency Management</bomDescription>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly-ee</artifactId>
+                                    </manifest>
+                                </channel>
```
bstansberry, line 92: This shouldn't be used. Standard WF and WFP are separate.

boms/user/server/expansion/preview/pom.xml
```diff
@@ -38,11 +38,24 @@
...
             <artifactId>wildfly-preview-feature-pack</artifactId>
+            <version>${full.maven.version}</version>
+            <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>org.wildfly.channels</groupId>
+            <artifactId>wildfly</artifactId>
+            <version>${full.maven.version}</version>
+            <type>pom</type>
+        </dependency>
```
bstansberry, line 54: This shouldn't be used.

boms/user/server/expansion/preview/pom.xml
```diff
@@ -70,6 +83,20 @@
                             <bomVersion>${project.version}</bomVersion>
                             <bomName>${full.dist.product.release.name} Preview BOMs: Expansion</bomName>
                             <bomDescription>${full.dist.product.release.name} Preview: Expansion Dependency Management</bomDescription>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly</artifactId>
+                                    </manifest>
+                                </channel>
```
bstansberry, line 92: This shouldn't be used.

boms/user/tools/pom.xml
```diff
@@ -68,6 +77,14 @@
...
                             <bomDescription>Dependency Management with Tools useful for ${full.dist.product.release.name}</bomDescription>
                             <inheritExclusions>NONE</inheritExclusions>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly-ee</artifactId>
+                                    </manifest>
+                                </channel>
+                            </channels>
```
bstansberry, line 87: This seems off. I don't think our channels should be controlling the tools stuff. Is it for things like the CLI, which is a tool that also gets provisioned?

bstansberry (changes requested):

boms/user/tools/pom.xml
emmartins: Yes, and there are other artifacts like httpclient too, all come from EE (we moved all from ee-with-tools) thus the usage of ee channel.

boms/user/server/expansion/preview/pom.xml
emmartins: fixed

boms/user/server/expansion/preview/pom.xml
emmartins: fixed

boms/user/server/ee/preview/pom.xml
emmartins: fixed

boms/user/tools/pom.xml
bstansberry: Ok re the channel.

I realize now that standard WildFly controls these versions, even in the wildfly-ee|expansion-preview-with-tools boms.

Not ideal, but ok.

bstansberry (approved):

boms/user/client/ejb-client/pom.xml
```diff
@@ -74,6 +80,14 @@
                             <bomWithDependencies>true</bomWithDependencies>
                             <licenses>true</licenses>
                             <inheritExclusions>UNMANAGED</inheritExclusions>
+                            <channels>
+                                <channel>
+                                    <manifest>
+                                        <groupId>org.wildfly.channels</groupId>
+                                        <artifactId>wildfly-ee</artifactId>
+                                    </manifest>
```
jbliznak, line 88: Guys, this is a problem, specifically the versionless manifest coordinates.

You can see the problem even in the CI run for this PR https://github.com/wildfly/wildfly/actions/runs/11478610181/job/31943175993#step:5:20008 :
```
...
[INFO] Building WildFly BOM Builders: Client EJB Builder 35.0.0.Beta1-SNAPSHOT [129/194]
...
[INFO] --- wildfly-bom-builder-plugin:2.0.7.Final:build-bom (build-bom) @ wildfly-ejb-client-bom-builder ---
...
Oct 23, 2024 11:13:38 AM org.wildfly.channel.maven.VersionResolverFactory$MavenResolverImpl resolveChannelMetadata
INFO: Resolving channel metadata from Maven artifact org.wildfly.channels:wildfly-ee:34.0.0.Final
```

I would expect `org.wildfly.channels:wildfly-ee:35.0.0.Beta1-SNAPSHOT` is resolved here, otherwise the generated BOMs are completely off wrt current build.

jbliznak: reported at https://issues.redhat.com/browse/WFLY-19908

jbliznak: BTW even if the resolution is somehow fixed to look at local repository too, we might still need to explicitly set version here otherwise the build won't be guaranteed to be reproducible anymore - eg. when building 34.0.x for whatever reason, the 35.0.0.Beta1 might be already public and as higher version it will be resolved here which is clearly wrong.

PR #18325, opened 2024-10-28
[WFLY-19895]: Upgrade Apache Artemis to 2.38.0.

ehsavoie:
* Upgrading Apache Artemis to 2.38.0
* Upgrading Jackson to 2.18.0
* Upgrading commons-io to 2.17.0

Jira: https://issues.redhat.com/browse/WFLY-19895
Release Note: https://activemq.apache.org/components/artemis/download/release-notes-2.38.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -485,7 +485,7 @@
         <version.joda-time>2.12.7</version.joda-time>
         <version.net.jcip>1.0</version.net.jcip>
         <version.net.shibboleth.utilities.java-support>8.0.0</version.net.shibboleth.utilities.java-support>
-        <version.org.apache.activemq.artemis>2.37.0</version.org.apache.activemq.artemis>
+        <version.org.apache.activemq.artemis>2.38.0</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>2.0.0</version.org.apache.activemq.artemis.native>
         <version.org.apache.avro>1.12.0</version.org.apache.avro>
         <version.org.apache.cxf>4.0.5</version.org.apache.cxf>


bstansberry:
@jamezp @pferraro @kabir @yrodiere @jasondlee @fjuma Please review re the Jackson upgrade here. On #18348 I asked for input by Nov 22, but for this sooner is better as it's part of a larger piece of work.

@jamezp Any concerns with the commons-io change?

@ehsavoie Why are you proposing the Jackson and commons-io updates? I don't see any Artemis module dependencies on the databind and commons.io modules. (Perhaps something is missing, which would be a different issue.)

pom.xml
```diff
@@ -401,7 +401,7 @@
         <version.antlr>4.13.0</version.antlr>
         <version.com.carrotsearch.hppc>0.8.1</version.com.carrotsearch.hppc>
         <version.com.fasterxml.classmate>1.5.1</version.com.fasterxml.classmate>
-        <version.com.fasterxml.jackson>2.17.0</version.com.fasterxml.jackson>
+        <version.com.fasterxml.jackson>2.18.0</version.com.fasterxml.jackson>
```
jamezp, line 404: If we go with 2.18, we should probably use 2.18.2 that was just released https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.18.2.

jamezp:
These should be fine for RESTEasy. I've added a comment about the version of Jackson.

yrodiere (commented):
The Jackson upgrade is most likely fine as far as Hibernate projects are concerned.

I created https://issues.redhat.com/browse/WFLY-19943 to check it now, and for every future upgrade. I will try to submit a PR soon, and run tests on top of this PR, and report back.

kabir (approved):

fjuma (approved):

bstansberry:
@spyrkob or @TomasHofman any concerns re Jackson 2.18? IIRC either prospero or wildfly-channels uses Jackson.

bstansberry (changes requested):
@ehsavoie I'm 'requesting changes' re my previous comment above about how in the WF runtime there doesn't seem to be any relationship between Artemis and Jackson or commons-io. So something needs clarifying.

pom.xml
bstansberry: Definitely. Otherwise within a day or two we'll just end up with a dependabot PR for everyone to approve.

ehsavoie:
@bstansberry I updated the versions to match those used in Apache Artemis but I don't see a direct dependency either so I can revert that part.

TomasHofman:
> @spyrkob or @TomasHofman any concerns re Jackson 2.18? IIRC either prospero or wildfly-channels uses Jackson.

I tried to rebuild both wildfly-channel and prospero with 2.18.1 and didn't see any issues, tests are passing.

bstansberry (approved):

bstansberry:
Since the Jackson update was withdrawn here (a good thing for formal correctness) but most of the desired oks of it already came in, I went ahead and filed #18391.

@jamezp If the common-io change was ok, please file a JIRA and send up a PR. REST is the only use of that lib.

bstansberry:
Thanks @ehsavoie

PR #18300, opened 2024-10-15
[WFLY-16670]

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-16670

Description: this is different solution than what was presented last week, in short I had to reverse the logical dependency EE depends on Concurrency SPI, now it's Concurrency SPI that depends on EE, which facilitates a lot how to integrate with the EE component framework. 

This fundamental change should make it easier if we ever introduce concurrency extension/subsystems, but most important it also means that we no longer need to break the EeLogger, or LocalDescriptions+ResourceDescriptionResolvers, so the issues we discussed last week are no longer valid. Please note that we can (and probably should) now start to log in a new ConcurrencyLogger (with new IDs) but there is no need to change any older message or touch the EeLogger class any more.

PR #18337, opened 2024-10-29
[WFLY-16670] Split EE Concurrency into its own maven modules

bstansberry:
https://issues.redhat.com/browse/WFLY-16670

Follow-up in #18300 to clarify the org.wildfly.concurrency module dependency set.

Note I think this can be further improved by moving the concurrency-spi artifact into org.jboss.as.ee, which concurrency-impl already depends on. Or, create a separate module for it that both org.jboss.as.ee and org.wildfly.concurrency depend on. That will reduce the org.wildfly.concurrency module.xml deps (which will get duplicated in the EE 11 variant) down to those listed in https://github.com/wildfly/wildfly/compare/main...bstansberry:wildfly:pullRequest18300?expand=1#diff-b30e5483fc05b782b102f27db7772b611f60b64baadf463473cd5cd67a5791e1, eliminating all the management API development deps.


bstansberry:
Merging oer https://github.com/wildfly/wildfly/pull/18300#discussion_r1822128030

PR #18338, opened 2024-10-29
[WFLY-15590] Add READMEs for the 'galleon-shared' and 'galleon-local'…

bstansberry:
… modules

This doesn't do everything that could be done for this issue, but the other modules in these directories are more self-explanatory.

https://issues.redhat.com/browse/WFLY-15590


jasondlee (approved):
Looks great to me! :)

bstansberry:
Thanks @jasondlee

PR #18351, opened 2024-11-06
[WFLY-19694] Add prototype for the equivalent of having a PersistenceProvider.getClassTransformer() to allow entity enhancement and bootstrap of the persistence unit to occur in two phases

scottmarlow:
https://issues.redhat.com/browse/WFLY-19694

The idea is to setup the entity enhancement prior to creating the EntityManagerFactory instead of using org.jipijapa.plugin.spi.TwoPhaseBootstrapCapable + Hibernate ORM org.hibernate.jpa.boot.spi.EntityManagerFactoryBuilder two phase bootstrap.


gavinking:
For reference: https://github.com/jakartaee/persistence/issues/650

scottmarlow:
Replaced by https://github.com/wildfly/wildfly/pull/18863

PR #18353, opened 2024-11-07
[WFLY-19846] [cleanup] Promote MicroProfile Telemetry 2.0 to WildFly Standard

jasondlee:
https://issues.redhat.com/browse/WFLY-19846

- Revert changes to shared OpenTelemetryServerSetupTask
- Create private/local ServerSetupTask for BasicOpenTelemetryTestCase
- Add helper method for metric handling
- Clean up debug logging


rhusar (changes requested):
This actually needs changes:

1. Makes a BasicOpenTelemetryTestCase fail with java.lang.IllegalStateException: Could not connect to Ryuk at localhost:34003
2. needs a git squash

jasondlee:
> This actually needs changes:
> 
>     1. Makes a BasicOpenTelemetryTestCase fail with java.lang.IllegalStateException: Could not connect to Ryuk at localhost:34003
> 
>     2. needs a git squash

1. It passed for me locally, but I have `TESTCONTAINERS_RYUK_DISABLED=true` in my env. I may just add that to the surefire config. Ryuk is almost always the culprit in these test failures.
2. Yes, but git was behaving oddly, trying to squash ~10 commits when I told it 3. I'll try again Monday.

jasondlee:
> This actually needs changes:
> 
>     1. Makes a BasicOpenTelemetryTestCase fail with java.lang.IllegalStateException: Could not connect to Ryuk at localhost:34003
> 
>     2. needs a git squash

Squash done. We'll see how the CI likes #1...

testsuite/integration/microprofile/pom.xml
```diff
@@ -500,6 +500,7 @@
                     <environmentVariables>
                         <!-- dummy env property for MicroProfile Config testing -->
                         <MPCONFIG_TEST_ENV_VAR>integ-basicmpconfig-test-env-var-value</MPCONFIG_TEST_ENV_VAR>
+                        <TESTCONTAINERS_RYUK_DISABLED>true</TESTCONTAINERS_RYUK_DISABLED>
```
rhusar, line 503: I don't think we want this. IIRC Ryuk is used to reap containers on platforms like Linux and disabling it especially on CI will cause problems.

Case in point, the tests fail with something that sounds like it's a direct result of this:

```
Caused by: org.testcontainers.containers.ContainerLaunchException: Timed out waiting for container port to open (localhost ports: [33376, 33377, 33374, 33375] should be listening)
```

rhusar (changes requested):

rhusar (changes requested):
Needs a rebase (leaving this comment so this gets attention if it needs to).

testsuite/integration/microprofile/pom.xml
```diff
@@ -503,6 +503,7 @@
                         <TC_LOGGING>true</TC_LOGGING>
                     </environmentVariables>
                     <systemPropertyVariables>
+                        <testcontainers.reuse.enable>false</testcontainers.reuse.enable>
```
rhusar, line 506: What's the purpose of this?

Reading https://java.testcontainers.org/features/reuse/ - 'Those containers won't stop after all tests are finished.' so this is definitely disabled by default, so I am not sure why this is desired here.

rhusar: 👍🏼

bstansberry:
@jasondlee Please change the target of this PR to 'main'. The mp-7-wip branch has been merged.

jasondlee:
> @jasondlee Please change the target of this PR to 'main'. The mp-7-wip branch has been merged.

@bstansberry Retargeted.

bstansberry:
@jasondlee I put the 39.x label on this not because it has to be done for 39, but just to draw a bit of attention to it. But that was a misuse of that label, so I'm going to remove it.

 It had the Feature label even though the feature was long since done, which probably helped this fall into a crack. Anyone this looks like one of those WIP things that we should try and bring to conclusion.

testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/OpenTelemetrySetupTask.java
```diff
@@ -39,6 +45,8 @@ public void setup(final ManagementClient managementClient, final String containe
 
     @Override
     public void tearDown(final ManagementClient managementClient, final String containerId) throws Exception {
+        otelCollectorContainer.stop();
```
rhusar, line 48: IIRC This needs to be last, not first, otherwise you are going to get unnecessary connection errors.

I already fixed this in org.jboss.as.test.shared.observability.setuptasks.OpenTelemetryWithCollectorSetupTask

rhusar (changes requested):

rhusar:
Also, needs a BIG rebase :)

PR #18370, opened 2024-11-13
WFLY-19937 Update to Narayana LRA 0.0.9.Final

marcosgopen:
https://issues.redhat.com/browse/WFLY-19937

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/main/module.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.jboss.narayana.lra">
+
+    <dependencies>
+        <module name="org.jboss.resteasy.resteasy-client"/>
+        <module name="org.jboss.resteasy.resteasy-core-spi"/>
+        <module name="org.jboss.resteasy.resteasy-core"/>
+        <module name="org.jboss.resteasy.resteasy-jaxb-provider"/>
+        <module name="org.jboss.resteasy.resteasy-json-p-provider"/>
+        <module name="jakarta.transaction.api"/>
+        <module name="jakarta.resource.api"/>
+        <module name="jakarta.ws.rs.api" services="export"/>
+        <module name="jakarta.servlet.api"/>
+        <module name="jakarta.xml.bind.api"/>
+        <module name="jakarta.ejb.api"/>
+        <module name="org.jboss.jts"/>
+        <module name="org.jboss.logging"/>
+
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -550,6 +550,7 @@
         <version.org.jboss.metadata>16.0.0.Final</version.org.jboss.metadata>
         <version.org.jboss.mod_cluster>2.0.4.Final</version.org.jboss.mod_cluster>
         <version.org.jboss.narayana>7.1.0.Final</version.org.jboss.narayana>
```
marcosgopen, line 552: Please notice that narayana 7.1.0.Final is still present because Narayana LRA is now independently released

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/main/module.xml
```diff
@@ -0,0 +1,31 @@
...
+        <module name="org.jboss.jts"/>
+        <module name="org.jboss.logging"/>
+
+        <!-- TODO WFLY-5966 validate the need for these and remove if not needed.
+             Prior to WFLY-5922 they were exported by javax.ejb.api. -->
+        <module name="javax.rmi.api"/>
+        <module name="org.omg.api"/>
+        <module name="java.xml"/>
+    </dependencies>
+</module>
```
xstefank, line 31: Why is this needed?

bstansberry: It's unused and it doesn't do anything, as it has no resources of its own and doesn't export any of its dependencies. So it should be removed, along with https://github.com/wildfly/wildfly/pull/18370/files#diff-0a882b46c5e23c165088809789a9549c4737c55d10d01e072fe0fd9666ee439aR309

bstansberry (changes requested):

pom.xml
```diff
@@ -550,6 +550,7 @@
...
+        <version.org.jboss.narayana.lra>0.0.9.Final</version.org.jboss.narayana.lra>
```
bstansberry, line 553: Does the 0.0.9 version imply anything about stability? If it does, will it be at a stable version in time to be upgrade in WF by December 18?

xstefank: @bstansberry no it doesn't. Marco told me he just chose this version because he wants to do the 1.0.0 as a normal upgrade once integrated to go through all verifications. The code is forked from narayana main and it's the same.

xstefank (changes requested):
Changing my review since changes are required.

xstefank (changes requested):
Test

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/main/module.xml
marcosgopen: Thank you for your feedback! I created it so that it aligns to the new groupID: org.jboss.narayana.lra
I understand your point and since I will be on PTO from tomorrow I will close this PR so that a new one can be raise from somebody else of the Narayana team while I am away.

marcosgopen:
I can see some changes are needed, since I will be on PTO from tomorrow I will close this PR so that a new one can be raised from somebody else of the Narayana team while I am away.

PR #18376, opened 2024-11-13
[WFLY-19924] - OpenTelemetry quickstart application logs warning with NoClassDefFoundError exceptionafter URL request

jasondlee:
https://issues.redhat.com/browse/WFLY-19924

Add netty module dep
Set netty default logger factory

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry/main/module.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/opentelemetry/main/module.xml
@@ -17,6 +17,7 @@
 
     <dependencies>
         <module name="org.wildfly.extension.opentelemetry-api"/>
+        <module name="io.netty.netty-common"/>
         <module name="io.opentelemetry.api" />
         <module name="io.opentelemetry.sdk" />
         <module name="io.opentelemetry.context" />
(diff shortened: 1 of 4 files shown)


emmartins (approved):
LGTM, opentelemetry tracing qs provisioned with this branch no longer throws the exception

jamezp (commented):
LGTM. Should we also do `org.wildfly.extension.microprofile.telemetry.MicroProfileTelemetryExtension` as well? It also uses `io.smallrye.opentelemetry` which has a dependency on `io.vertx.core` which in turn as a dependency on `io.netty.netty-common`.

jasondlee:
> LGTM. Should we also do `org.wildfly.extension.microprofile.telemetry.MicroProfileTelemetryExtension` as well? It also uses `io.smallrye.opentelemetry` which has a dependency on `io.vertx.core` which in turn as a dependency on `io.netty.netty-common`.

All the MP Telemetry module does is allow a deployment to override otel server config via MP Config. All the real work is still done by the `opentelemetry` subsystem, so I don't think it's necessary to update both.

jamezp (approved):

jamezp (approved):

jamezp:
FYI I've queued a new CI job for the failed Galleon Linux run as it doesn't look related.

bstansberry:
Thanks @jasondlee and @jamezp

PR #18378, opened 2024-11-14
[WFLY-19942] io.smallrye.reactive.mutiny.vertx-core module needs to depend on io.netty.netty-transport module

gaol:
Issue: https://issues.redhat.com/browse/WFLY-19942


kabir:
Thanks @gaol

PR #18382, opened 2024-11-14
[mp-7-wip] WFLY-19592 WFLY-19948 Upgrade SmallRye Fault Tolerance from 6.6.1 to 6.6.3 and enable simultaneous metrics collection with MP Telemetry and Micrometer

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-19592
https://issues.redhat.com/browse/WFLY-19948


rhusar:
Let me stick a hold on this for a while as we might be getting the new SR FT release with the metrics changes we wanted.

rhusar:
OK, new 6.6.x with the remaining fixes is on it's way. Keeping the hold.

rhusar:
Updated with latest release of SR FT 6.6.3 which brings in fixes to registering with multiple containers and exposes properties (documented) to finely tune which metrics provider to use.

PR #18394, opened 2024-11-18
WFLY-19943 Add integration tests for Hibernate features relying on Jackson/Jaxb

yrodiere:
https://issues.redhat.com/browse/WFLY-19943


yrodiere:
Amended so that tests pass with the security manager. Other failures seem unrelated -- I'm just changing/adding tests anyway, so if other tests are failing... seems unlikely to be related.

scottmarlow (approved):

bstansberry:
Thanks @yrodiere and @scottmarlow

PR #18396, opened 2024-11-18
WFLY-19891 Fix deadlock when application tries to invoke a timed-out timer referenced from TimerService.getTimers() within a @Timeout method.

pferraro:
https://issues.redhat.com/browse/WFLY-19891


pferraro:
Verified by @jbaesner .

PR #18399, opened 2024-11-18
[WFLY-19967] Hibernate ORM integration should also check for hibernate.cache.region.factory_class being set to jcache

scottmarlow:
https://issues.redhat.com/browse/WFLY-19967

Note that Hibernate ORM actually does the integration with `jcache`.  This change is to just allow Hibernate ORM integration to be reached.

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/service/ServiceContributorImpl.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/service/ServiceContributorImpl.java
@@ -20,6 +20,7 @@ public class ServiceContributorImpl implements ServiceContributor {
     private static final String CONTROL2LCINTEGRATION = "wildfly.jpa.regionfactory";
     private static final String TRANSACTION_PLATFORM = "hibernate.transaction.jta.platform";
     private static final String EHCACHE = "ehcache";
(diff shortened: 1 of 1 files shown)


scottmarlow:
Echoing discussion from https://wildfly.zulipchat.com/#narrow/channel/424923-releases/topic/WildFly.2034.2E0.2E1.20release.20coordination/near/483112212

> Paul Ferraro: Is there a reason for gatekeeping the region factory implementation?
> 
> Scott Marlow: Good question, originally, it helped avoid deployment failures. But we could also remove the gatekeeping of the region factory implementation and let applications fail if they contain incorrect cache settings.
> 
> Paul Ferraro: We could also just log a message in the event that the deployment contains a non-default region factory implementation.

Sanne:
Why would you log something only for JCache and Ehcache?  There are more implementations in the community, and the possibility for people to write their custom ones.

I'd log a warning (and not just a trace) if *anything* else than our recommended solution (Infinispan) is being set.

cc/ @pferraro

bstansberry:
@scottmarlow Please progress this out of draft state, or close it and reopen if you're able to get back to it.

I'm doing some PR queue housekeeping. :)

bstansberry:
Closing per my comment last November.

PR #18402, opened 2024-11-19
[WFLY-19758] Remove ApplicationClientStructureProcessor

aldaris:
Picked up this good first issue here:
https://issues.redhat.com/browse/WFLY-19758

The test run with -Dts.basic was successful, so based on the ticket this might be good to go.


bstansberry:
Thank you, @aldaris!

PR #18405, opened 2024-11-19
WFLY-19974 Update Arquillian Testcontainers to 1.0.0.Alpha3

jasondlee:
https://issues.redhat.com/browse/WFLY-19974

* Bump version
* Move version info to managed property
* Remove explicit Exception use in annotation so tests will fail if the container engine is not available
* Add profile to skip tests where docker is not required

### DIFF
--- a/boms/standard-test-expansion/pom.xml
+++ b/boms/standard-test-expansion/pom.xml
@@ -232,6 +232,13 @@
             </dependency>
 
             <dependency>
+                <groupId>org.jboss.arquillian</groupId>
+                <artifactId>arquillian-testcontainers</artifactId>
+                <version>${version.org.jboss.arquillian.testcontainers}</version>
+                <scope>test</scope>
+            </dependency>
+
+            <dependency>
                 <groupId>org.jboss.weld.se</groupId>
                 <artifactId>weld-se-core</artifactId>
                 <version>${version.org.jboss.weld.weld}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -372,6 +372,7 @@
         <version.org.javassist>3.29.2-GA</version.org.javassist>
         <version.org.jboss.arquillian.core>1.9.1.Final</version.org.jboss.arquillian.core>
         <version.org.jboss.arquillian.jakarta>10.0.0.Final</version.org.jboss.arquillian.jakarta>
+        <version.org.jboss.arquillian.testcontainers>1.0.0.Alpha3</version.org.jboss.arquillian.testcontainers>
         <version.org.jboss.byteman>4.0.23</version.org.jboss.byteman>
         <version.org.jboss.logmanager.commons-logging-jboss-logmanager>1.0.3.Final</version.org.jboss.logmanager.commons-logging-jboss-logmanager>
         <!-- only needed here until wildfly-arquillian has this version properly synced with arquillian itself  -->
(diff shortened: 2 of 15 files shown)


jamezp (commented):
I feel like it might be best to add a profile for setting the system property rather than setting it on CI. Something like `skip-testcontainer-tests`.

jasondlee:
> I feel like it might be best to add a profile for setting the system property rather than setting it on CI. Something like `skip-testcontainer-tests`.

That's a good idea. I can add a profile to this PR.

rhusar (changes requested):
This is missing a profile activation that changes this exception so that tests get skipped on ~~Windows hosts~~ hosts that don't have a container runtime.

jamezp (dismissed):

rhusar (changes requested):
@jasondlee These files need to be updated too:

```
testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/micrometer/FaultToleranceMicrometerIntegrationDisabledTestCase.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/BaseContainer.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/MicrometerSetupTask.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/OpenTelemetryWithCollectorSetupTask.java:@DockerRequired(AssumptionViolatedException.class)
```

rhusar (changes requested):
@jasondlee These classes  in this branch still haven't been updated, I think you may have mixed up the branches...

```
testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/ContextPropagationTestCase.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationTestCase.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/MetricsTestCase.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/BaseContainer.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/MicrometerSetupTask.java:@DockerRequired(AssumptionViolatedException.class)
testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/setuptasks/OpenTelemetrySetupTask.java:@DockerRequired(AssumptionViolatedException.class)
```

jasondlee:
@rhusar Strange. I thought I had searched the whole code base, but I'll double check.

rhusar:
The problem with leaving those (the default exception) in there is that they will continue to be skipped silently and appear all green here on CI status checks.

jasondlee:
Right. I think I just didn't search widely enough. Fixed and testing locally.

rhusar (approved):
Code looks good now, I ll be asleep when CI runs finish, so leaving 👍🏼  here now.

bstansberry:
Thanks @jasondlee, @rhusar and @jamezp

PR #18409, opened 2024-11-21
[WFLY-19981]:Add optional resteasy-patchfilter-disabled element to jaxrs sub…

jimma:
…system

https://issues.redhat.com/browse/WFLY-19981

____

> Wildfly issue links:
> * [WFLY-19881](https://issues.redhat.com/browse/WFLY-19881)


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
```diff
@@ -29,7 +29,7 @@ public class JaxrsExtension implements Extension {
 
     public static final String SUBSYSTEM_NAME = "jaxrs";
 
-    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(4, 0, 0);
+    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(3, 1, 0);
```
jamezp, line 32: Why are we downgrading the model version? The model version and the schema version are managed differently and do not need to be the same value.

jaxrs/src/main/resources/schema/jboss-as-jaxrs_3_1.xsd
jamezp, line 1: I don't see the new attribute or element defined in here.

jamezp (changes requested):
We should wait on changes until a decision #18315 is made.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
jimma: I don't understand why we need to use a different version. Why do we have to use 4.0.0 here ? It seems using the same version with the schema  makes more sense.  Th another reason I changed version here is I saw the transformer register here seems not correct :  
https://github.com/wildfly/wildfly/blob/main/jaxrs/src/main/java/org/jboss/as/jaxrs/ResteasyExtensionTransformerRegistration.java#L35-L39

jimma: It turns out this model version is already used in the previous release and we can't do things to align with schema version. I already changed back and upgraded the model version to 4.1.0 for schema version 3.1.0.

jamezp:
Until I submit a new PR I will leave this open, but this conflicted with #18315 so I'll create a new PR to add this new attribute.

jaxrs/src/main/resources/schema/jboss-as-jaxrs_3_1.xsd
```diff
@@ -0,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="urn:jboss:domain:jaxrs:3.1"
```
bstansberry, line 9: Unless @jamezp disagrees, please do a major version bump for this and the management API version.

That's our standard practice; we don't try and signal the significance of a new version by using major or minor versions.

Leave minor versions to downstream in case they need to insert a new version during their long period between feature releases.

jamezp:
Closing this as replaced by #18446

PR #18411, opened 2024-11-21
[WFLY-19973] @PersistenceContext properties attribute should be processed

scottmarlow:
https://issues.redhat.com/browse/WFLY-19973
https://wildfly.zulipchat.com/#narrow/channel/196266-wildfly-user/topic/.40PersistenceContext.20properties.20attribute.20problem

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/TransactionScopedEntityManager.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/TransactionScopedEntityManager.java
@@ -198,6 +198,11 @@ public class TransactionScopedEntityManager extends AbstractEntityManager implem
         return skipQueryDetach.booleanValue();
     }
 
+    @Override
+    public void setProperty(String propertyName, Object value) {
+        properties.put(propertyName, value);
+        super.setProperty(propertyName, value);
+    }
 
 
     /**
--- a/weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
+++ b/weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
@@ -11,6 +11,7 @@ import java.security.AccessControlContext;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 
@@ -18,6 +19,7 @@ import jakarta.enterprise.inject.spi.InjectionPoint;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.EntityManagerFactory;
 import jakarta.persistence.PersistenceContext;
+import jakarta.persistence.PersistenceProperty;
 import jakarta.persistence.PersistenceUnit;
 import jakarta.transaction.TransactionManager;
 import jakarta.transaction.TransactionSynchronizationRegistry;
@@ -71,7 +73,7 @@ public class WeldJpaInjectionServices implements JpaInjectionServices {
                 public EntityManager call() throws Exception {
                     return new TransactionScopedEntityManager(
                             scopedPuName,
-                            new HashMap<>(),
+                            getProperties(context),
                             persistenceUnitService.getEntityManagerFactory(),
                             context.synchronization(),
                             deploymentUnit.getAttachment(JpaAttachments.TRANSACTION_SYNCHRONIZATION_REGISTRY),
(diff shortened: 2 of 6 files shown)


bstansberry (commented):
@scottmarlow The change looks good but I think this should have a regression test.

scottmarlow:
Hmm, I smell more changes coming...

scottmarlow:
> Hmm, I smell more changes coming...

I reviewed related/other property handling which looks correctly handled.  

We define most properties at the Persistence Unit definition level.  We allow additional container/persistence provider properties to be specified at the Persistence Context level which are [described the JavaDoc here](https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/persistencecontext#properties()):
"
[PersistenceProperty](https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/persistenceproperty)[] properties
(Optional) Properties for the container or persistence provider. Vendor specific properties may be included in this set of properties. Properties that are not recognized by a vendor are ignored.

Default:
    {}
"

scottmarlow:
> @scottmarlow The change looks good but I think this should have a regression test.

@bstansberry I updated the test to cover `PersistenceContext properties`

jpa/subsystem/src/main/java/org/jboss/as/jpa/container/TransactionScopedEntityManager.java
```diff
@@ -198,6 +198,11 @@ protected boolean skipQueryDetach() {
         return skipQueryDetach.booleanValue();
     }
 
+    @Override
+    public void setProperty(String propertyName, Object value) {
+        properties.put(propertyName, value);
```
bstansberry, line 203: I found at least one call path (via PersistenceUnitServiceHandler) that leads to 'properties' being assigned to an immutable map, which would cause this to fail. There may be others.

Thread safety is probably an issue as well since AIUI this is part of the EM API and thus can be called from application.

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
```diff
@@ -122,6 +124,15 @@ private String getScopedPUName(final DeploymentUnit deploymentUnit, final String
         return scopedPu.getScopedPersistenceUnitName();
     }
 
+    private static Map getProperties(PersistenceContext context) {
+        HashMap map = new HashMap();
```
bstansberry, line 128: If TransactionScopedEntityManager cached this map in its 'properties' field that wouldn't be thread safe.

jpa/subsystem/src/main/java/org/jboss/as/jpa/container/TransactionScopedEntityManager.java
scottmarlow: The EntityManager is only expected to be accessed by one application thread at a time but we should ensure that the property added to the HashMap is not left in thread memory:

For example:

1. Servlet request makes call to session bean that adds calls EntityManager.setProperty("myAppProperty", new MyAppClassPropertyValue).  The session bean call happens to run on thread id = 2001.
2. Servlet request than calls the session bean again and the session bean happens to run on different thread id = 2025.  

The HashMap itself should be available on all threads once it is created but I think that values put in the TransactionScopedEntityManager as properties may be cached in CPU registers.

scottmarlow: More about thread safety:

As per https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#obtaining-an-entitymanager

> An entity manager must not be shared among multiple concurrently executing threads, as the entity manager and persistence context are not required to be threadsafe. Entity managers must only be accessed in a single-threaded manner.

weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java
scottmarlow: More about thread safety:

As per https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#obtaining-an-entitymanager

> An entity manager must not be shared among multiple concurrently executing threads, as the entity manager and persistence context are not required to be threadsafe. Entity managers must only be accessed in a single-threaded manner.

bstansberry (approved):
LGTM @scottmarlow -- you fixed the Collections.emptyMap() case and the thread safety aspect is covered by the spec statement about EMs not being thread safe.

bstansberry:
Thanks @scottmarlow

PR #18418, opened 2024-11-21
[WFLY-19985] Upgrade WildFly Arquillian to 5.1.0.Beta7.

jamezp:
https://issues.redhat.com/browse/WFLY-19985

Tag: https://github.com/wildfly/wildfly-arquillian/releases/tag/v5.1.0.Beta7
Diff: https://github.com/wildfly/wildfly-arquillian/compare/5.1.0.Beta6...v5.1.0.Beta7

This also includes a couple fixes due to https://github.com/wildfly/wildfly-arquillian/pull/508. The removal of the `org.wildfly.arquillian:wildfly-arquillian-testenricher-msc` which has a dependency on the `org.wildfly.core:wildfly-server` was causing issues with the shared testsuite module.

### DIFF
--- a/testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
+++ b/testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
@@ -11,7 +11,6 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REA
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_OPERATION_NAMES_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RELOAD_ENHANCED;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.STABILITY;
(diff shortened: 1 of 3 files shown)


testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
```diff
@@ -42,6 +41,7 @@
  * the {@code teardown()} method
  */
 public abstract class StabilityServerSetupSnapshotRestoreTasks implements ServerSetupTask {
+    private static final String SERVER_ENVIRONMENT ="server-environment";
```
rhusar, line 44: Format - missing space.

rhusar (changes requested):

testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
jamezp: I don't have a laptop with me to fix that so I won't be able to do that on Tuesday. 🙂

bstansberry:
Thanks @jamezp and @rhusar

PR #18420, opened 2024-11-22
Fix for WFLY-19989, h2-driver layer rule for jakarta DataSourceDefinition annotation

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-19989

* With those rules, the h2-driver will be discovered by WildFly Glow (starting 1.2.0.Beta1) based on DataSourceDefinition annotation values.
* Upgrade to WildFly Glow 1.2.0.Beta2 that includes the rule
* Added metadata test to cover the rule
* Updated integration tests with a change of discovered content

This fix is expected by quickstart: https://github.com/wildfly/quickstart/pull/973


jfdenise:
Failure is caused by WildFly Glow, preparing a fix and a 1.2.0.Beta2.

bstansberry:
@jfdenise I'm going to merge this but I think adding a test of the url attribute would be good.

PR #18423, opened 2024-11-22
[MP7] WFLY-19974 - Update Arquillian Testcontainers to 1.0.0.Alpha3

jasondlee:
https://issues.redhat.com/browse/WFLY-19974

Port of changes from main.

* Bump version
* Move version info to managed property
* Remove explicit Exception use in annotation so tests will fail if the container engine is not available
* Add profile to skip tests where docker is not required


kabir:
@jasondlee @rhusar @bstansberry please adjust the Jira as needed

PR #18430, opened 2024-11-25
[WFLY-19502][WFLY-16644] Create mixed domains tests against EAP8 and unignore EAP 7.4 legacy tests

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-19502
Jira issue: https://issues.redhat.com/browse/WFLY-16644

We will follow up with another Jira to introduce the latest WidlFly version before the first version where we introduced a community only feature so we will be also testing DC -- HC under community -- community stability level pairs.

This PR requires a core upgrade with:
https://github.com/wildfly/wildfly-core/pull/6259
https://github.com/wildfly/wildfly-core/pull/6204


testsuite/mixed-domain/src/test/java/org/jboss/as/test/integration/domain/mixed/MixedDomainTestSupport.java
```diff
@@ -174,16 +183,15 @@ private void configureSecondaryJavaHome() {
...
-            //Start the primary in admin only  and reconfigure the domain with what
+            //Start the primary in admin only and reconfigure the domain with what
             //we want to test in the mixed domain and have the DomainAdjuster
             //strip down the domain model to something more workable. The domain
             //adjusters will also make adjustments for the legacy version being
             //tested.
             DomainLifecycleUtil primaryUtil = getDomainPrimaryLifecycleUtil();
+            assert primaryUtil.getConfiguration().getStability() == version.getStability();
             primaryUtil.getConfiguration().setAdminOnly(true);
-            //primaryUtil.getConfiguration().addHostCommandLineProperty("-agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=y");
```
rhusar, line 186: BTW I think this was intentionally left out here for connecting a remote debugger since failures here are pretty hard to investigate otherwise.

yersan: Hi @rhusar , yes, I replaced it by using environment variables, you can now the DC with `mvn clean .... -Dwildfly.primary.debug` and the legacy host with "-Dwildfly.secondary.debug"

rhusar: @yersan Excellent, even better! Thanks.

bstansberry:
This all LGTM.

yersan:
Thanks both, just rebased to include it with the incoming Wildfly Core release

yersan:
@luck3y We also need an EAP 8.0 server in the mixed domain directory. Could you add it?

luck3y:
> @luck3y We also need an EAP 8.0 server in the mixed domain directory. Could you add it?

Added, along with WildFly 31.0 Final.

PR #18435, opened 2024-11-26
[WFLY-20007] Remove all non-breaking uses of ModuleIdentifier in Messaging Subsystem

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-20007

[WFLY-20007] Remove all non-breaking uses of `ModuleIdentifier` in Messaging Subsystem

____

> Wildfly issue links:
> * [WFLY-17539](https://issues.redhat.com/browse/WFLY-17539)
> * [WFLY-19988](https://issues.redhat.com/browse/WFLY-19988)
> * [WFLY-19985](https://issues.redhat.com/browse/WFLY-19985)
> * [WFLY-19994](https://issues.redhat.com/browse/WFLY-19994)

### DIFF
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ServerAdd.java
@@ -4,6 +4,8 @@
  */
 package org.wildfly.extension.messaging.activemq;
 
+import static org.jboss.as.controller.ModuleIdentifierUtil.canonicalModuleIdentifier;
+
 import static org.wildfly.extension.messaging.activemq._private.MessagingLogger.ROOT_LOGGER;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -150,7 +152,6 @@ import org.jboss.as.network.SocketBinding;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.Property;
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.msc.service.ServiceBuilder;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceController.Mode;
@@ -717,8 +718,7 @@ class ServerAdd extends AbstractAddStepHandler {
             String className = classModel.get(NAME).asString();
             String moduleName = classModel.get(MODULE).asString();
             try {
-                ModuleIdentifier moduleID = ModuleIdentifier.fromString(moduleName);
-                Module module = Module.getCallerModuleLoader().loadModule(moduleID);
+                Module module = Module.getCallerModuleLoader().loadModule(canonicalModuleIdentifier(moduleName));
                 Class<?> clazz = module.getClassLoader().loadClass(className);
                 return clazz;
             } catch (Exception e) {
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeFactory.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/bridge/JMSBridgeFactory.java
@@ -5,6 +5,8 @@
 
 package org.wildfly.extension.messaging.activemq.jms.bridge;
 
+import static org.jboss.as.controller.ModuleIdentifierUtil.canonicalModuleIdentifier;
+
 import java.util.Properties;
 import org.apache.activemq.artemis.jms.bridge.ConnectionFactoryFactory;
 import org.apache.activemq.artemis.jms.bridge.DestinationFactory;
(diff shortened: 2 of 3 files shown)


ropalka (approved):

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/AbstractCreateJMSBridgeSetupTask.java
```diff
@@ -38,7 +38,7 @@ public void setup(ManagementClient managementClient, String containerId) throws
         jmsBridgeAttributes.get("source-destination").set(QUEUE1_JNDI_NAME);
         jmsBridgeAttributes.get("target-connection-factory").set(CF_JNDI_NAME);
         jmsBridgeAttributes.get("target-destination").set(QUEUE2_JNDI_NAME);
-        jmsBridgeAttributes.get("module").set("org.apache.activemq.artemis:main");
+        jmsBridgeAttributes.get("module").set("org.apache.activemq.artemis");
```
bstansberry, line 41: @soul2zimate Was this change required to get the test to pass, or did you do this more as a bit of tidying?

I'm concerned this indicates we've introduced a breaking change. AIUI in the past appending ":main" was not necessary but wasn't wrong. If the test breaks without this change that indicates it is now "wrong", which is an incompatible change.

@ehsavoie @ropalka @dmlloyd FYI.

jamezp: +1 the module name should allow for a slot in that same pattern.

soul2zimate: Yes, some JMS tests in smoke break without this change.

jamezp: This is due to https://github.com/wildfly/wildfly-core/blob/a62fb1c4ab771bac5790b2eb2d369af39eec3e45/server/src/main/java/org/jboss/as/server/deployment/module/ModuleDependency.java#L151-L153. The simple fix is to use `ModuleIdentifier.fromString()`. However, I don't know if we simply just want a PR for that or not.

jamezp: I'll file a fix for this later today. FWIW we should add the `:main` back even if it's failing and add the core-upgrade-required label as we'll need a core upgrade after the fix is in.

bstansberry: @jamezp I don't think ModuleDependency is relevant here. This PR isn't touching deployment processing code; it's all subsystem-level messaging container resources. The attribute in this test is used to create a subsystem resource, not to set up a ModuleDependency.

I think we'll likely need our own ModuleIdentifier from WF Core. ModuleLoader.loadModuleLocal is using the provided 'name' as is and doesn't try to recognize that "foo:main" and "foo" are equivalent. It was ModuleIdentifier that dealt with the fact that "foo" is the canonical representation of "foo:main". By running strings through it we got canonical strings.

ehsavoie: Yes I remember hitting those kind of issues while working on something else, I didn't dig them, sorry

bstansberry:

bstansberry (changes requested):

bstansberry:
@soul2zimate The https://github.com/wildfly/wildfly-core/pull/6270 API is now available.

testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/jms/AbstractCreateJMSBridgeSetupTask.java
bstansberry: Note that in the https://github.com/wildfly/wildfly-core/pull/6270 I added this bit:

https://github.com/wildfly/wildfly-core/pull/6270/files#diff-867089bef7659adb3ead7e82d208a38aa95d3fda6df7476187f2aa272bf064d4R94-R110

Any AttributeDefinition that represents a module name can add that ParameterCorrector and that will ensure that what gets stored in the model is the correct form. If this 'module' attribute did that and this test left the "org.apache.activemq.artemis:main" value, that would demonstrate it works.

soul2zimate:
Thanks @bstansberry I was away last week. PR is updated.

ropalka (approved):

bstansberry (approved):

bstansberry:
Thanks @soul2zimate and @ropalka

PR #18436, opened 2024-11-26
[WFLY-20008] Remove all non-breaking uses of ModuleIdentifier in Pojo Subsystem

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-20008

[WFLY-20008] Remove all non-breaking uses of `ModuleIdentifier` in Pojo Subsystem

### DIFF
--- a/pojo/src/main/java/org/jboss/as/pojo/KernelDeploymentModuleProcessor.java
+++ b/pojo/src/main/java/org/jboss/as/pojo/KernelDeploymentModuleProcessor.java
@@ -15,7 +15,6 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.server.deployment.module.ModuleDependency;
 import org.jboss.as.server.deployment.module.ModuleSpecification;
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoader;
 import org.jboss.modules.filter.PathFilter;
 import org.jboss.modules.filter.PathFilters;
@@ -29,7 +28,7 @@ import java.util.List;
  */
 public class KernelDeploymentModuleProcessor implements DeploymentUnitProcessor {
 
-    private ModuleIdentifier POJO_MODULE = ModuleIdentifier.create("org.jboss.as.pojo");
+    private final String POJO_MODULE = "org.jboss.as.pojo";
 
     /**
      * Add POJO module if we have any bean factories.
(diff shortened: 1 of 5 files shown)


pojo/src/main/java/org/jboss/as/pojo/KernelDeploymentModuleProcessor.java
```diff
@@ -29,7 +28,7 @@
  */
 public class KernelDeploymentModuleProcessor implements DeploymentUnitProcessor {
 
-    private ModuleIdentifier POJO_MODULE = ModuleIdentifier.create("org.jboss.as.pojo");
+    private String POJO_MODULE = "org.jboss.as.pojo";
```
ropalka, line 31: POJO_MODULE field should be final

ropalka (approved):
Except one minor issue it LGTM

bstansberry:
I put the hold label on this pending resolution of https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/ModuleIdentifier.20deprecation/near/484590414

pojo/src/main/java/org/jboss/as/pojo/KernelDeploymentModuleProcessor.java
soul2zimate: made it `final`

bstansberry:

bstansberry (changes requested):

bstansberry:
@soul2zimate The https://github.com/wildfly/wildfly-core/pull/6270 API is now available.

soul2zimate:
/retest

Hi @bstansberry , The change here doesn't need the wildfly-core API.

ropalka (approved):

bstansberry:
@soul2zimate @ropalka 

See KernelDeploymentXmlDescriptorParser L354

```
    private ModuleConfig parseModuleConfig(final XMLExtendedStreamReader reader) throws XMLStreamException {
        final ModuleConfig moduleConfig = new ModuleConfig();
        final int count = reader.getAttributeCount();
        final Set<Attribute> required = EnumSet.of(Attribute.NAME);
        for (int i = 0; i < count; i++) {
            final Attribute attribute = Attribute.of(reader.getAttributeLocalName(i));
            required.remove(attribute);
            final String attributeValue = reader.getAttributeValue(i);

            switch (attribute) {
                case NAME:
                    moduleConfig.setModuleName(attributeValue);
```

That 'attributeValue' is what gets passed through to ConfigVisitor.loadModule via ModuleConfig.visit.

I think canonicalizing it in the KernelDeploymentXmlDescriptorParser code is fine. I have a vague general preference for doing it as close to the user input as possible so it doesn't end up getting done repeatedly or for strings that come from our own presumably 'safe' static module names.

bstansberry (approved):

ropalka (approved):
Good catch @bstansberry !

PR #18438, opened 2024-11-26
[WFLY-20011] Remove all non-breaking uses of ModuleIdentifier in Naming Subsystem

ropalka:
https://issues.redhat.com/browse/WFLY-20011

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/ExternalContextObjectFactory.java
+++ b/naming/src/main/java/org/jboss/as/naming/ExternalContextObjectFactory.java
@@ -5,6 +5,8 @@
 
 package org.jboss.as.naming;
 
+import static org.jboss.as.controller.ModuleIdentifierUtil.canonicalModuleIdentifier;
+
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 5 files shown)


bstansberry:
I put the hold label on this pending resolution of https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/ModuleIdentifier.20deprecation/near/484590414

bstansberry (changes requested):

bstansberry:
@ropalka The https://github.com/wildfly/wildfly-core/pull/6270 API is now available.

ropalka:
Thanks for info @bstansberry - it is fixed now.

bstansberry (approved):
This breaks the API of ModularReference, which in theory could be used by extension authors outside the main WF code base. (The org.jboss.as.naming module is used in many extensions, but not the ModularReference class.)

I think it's ok; if this breaks someone they have a path to adapt, by using the create methods instead of constructors, and by dealing with a canonical string form of a module identifier themselves.

bstansberry:
Thanks @ropalka

PR #18439, opened 2024-11-26
[WFLY-20005] Persistence/JPA subsystem should remove non-breaking use of ModuleIdentifier

scottmarlow:
https://issues.redhat.com/browse/WFLY-20005

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
@@ -298,6 +298,18 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 25, value = "Couldn't load %s from Jakarta Persistence modules classloader")
     RuntimeException cannotLoadFromJpa(@Cause Throwable cause, String injectionTypeName);
 
+//    /**
+//     * Creates an exception indicating the module, represented by the {@code moduleId} parameter, could not be loaded
+//     * for the adapter, represented by the {@code name} parameter.
+//     *
+//     * @param cause    the cause of the error.
+//     * @param moduleId the module id that was attempting to be loaded.
+//     * @param name     the name of the adapter.
+//     * @return a {@link RuntimeException} for the error.
+//     */
+    //@Message(id = 26, value = "Could not load module %s to add %s adapter to deployment")
+    //RuntimeException cannotLoadModule(@Cause Throwable cause, ModuleIdentifier moduleId, String name);
+
     /**
      * Creates an exception indicating the persistence provider module, represented by the
      * {@code persistenceProviderModule} parameter, had an error loading.
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
@@ -298,18 +298,6 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 25, value = "Couldn't load %s from Jakarta Persistence modules classloader")
     RuntimeException cannotLoadFromJpa(@Cause Throwable cause, String injectionTypeName);
 
-//    /**
-//     * Creates an exception indicating the module, represented by the {@code moduleId} parameter, could not be loaded
-//     * for the adapter, represented by the {@code name} parameter.
-//     *
-//     * @param cause    the cause of the error.
-//     * @param moduleId the module id that was attempting to be loaded.
-//     * @param name     the name of the adapter.
-//     * @return a {@link RuntimeException} for the error.
-//     */
-    //@Message(id = 26, value = "Could not load module %s to add %s adapter to deployment")
-    //RuntimeException cannotLoadModule(@Cause Throwable cause, ModuleIdentifier moduleId, String name);
-
     /**
      * Creates an exception indicating the persistence provider module, represented by the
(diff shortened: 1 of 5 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
```diff
@@ -229,10 +217,12 @@ private int loadPersistenceUnits(final ModuleSpecification moduleSpecification,
...
                     if (providerModuleName != null) {
-                        addOptionalDependency(moduleSpecification, moduleLoader, deploymentUnit, ModuleIdentifier.fromString(providerModuleName));
-                        ROOT_LOGGER.debugf("%s is configured to use persistence provider '%s', adding an optional dependency on module '%s'",
-                                pu.getPersistenceUnitName(), pu.getPersistenceProviderClassName(), providerModuleName);
+                        String moduleIdentifier = providerModuleName;
+                        moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, moduleIdentifier, true, false, false, false));
+                        ROOT_LOGGER.debugf("added %s dependency to %s", moduleIdentifier, deploymentUnit.getName());
                     }
+                    ROOT_LOGGER.debugf("%s is configured to use persistence provider '%s', adding an optional dependency on module '%s'",
+                            pu.getPersistenceUnitName(), pu.getPersistenceProviderClassName(), providerModuleName);
```
yrodiere, line 225: This probably shouldn't be displayed if `providerModuleName` is `null`?

yrodiere (approved):
See also https://hibernate.zulipchat.com/#narrow/channel/132092-hibernate-search-dev/topic/WFLY-20005

As far as I can see this is semantically equivalent to what we had before, apart maybe from a few debug logs. So, +1 from me.

Thanks for taking care of it!

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPADependencyProcessor.java
scottmarlow: Addressed ^

bstansberry (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -298,18 +298,6 @@ public interface JpaLogger extends BasicLogger {
...
-//     * for the adapter, represented by the {@code name} parameter.
-//     *
-//     * @param cause    the cause of the error.
-//     * @param moduleId the module id that was attempting to be loaded.
-//     * @param name     the name of the adapter.
-//     * @return a {@link RuntimeException} for the error.
-//     */
-    //@Message(id = 26, value = "Could not load module %s to add %s adapter to deployment")
-    //RuntimeException cannotLoadModule(@Cause Throwable cause, ModuleIdentifier moduleId, String name);
-
```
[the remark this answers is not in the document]
scottmarlow: Restored.  I removed due to api change but probably okay to leave this unchanged.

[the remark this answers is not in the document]
bstansberry: Thanks. There's no expectation that these methods would ever be uncommented. They are just a form of documentation of the how the message ids were used.

bstansberry:
@scottmarlow The https://github.com/wildfly/wildfly-core/pull/6270 API is now available.

scottmarlow:
> @scottmarlow The [wildfly/wildfly-core#6270](https://github.com/wildfly/wildfly-core/pull/6270) API is now available.

Thanks @bstansberry  for the new API, very helpful!  

Please check https://github.com/wildfly/wildfly/pull/18439/commits/00e6180836ff5b169f242d9114a9728fe9fd9ab6 + https://github.com/wildfly/wildfly/pull/18439/commits/2bcf75bb2663c24cd3063d219eaae2b7029d0ffa  to see updates to use the new https://github.com/wildfly/wildfly-core/pull/6270 API.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/HibernateSearchProcessor.java
```diff
@@ -93,18 +81,20 @@ else if (!providerModule.equals(searchMapperModuleName)) {
 
         // use Search module name specified in persistence unit definition
         if (searchMapperModuleName != null && !IGNORE.equals(searchMapperModuleName)) {
-            ModuleIdentifier moduleIdentifier = ModuleIdentifier.fromString(searchMapperModuleName);
-            moduleSpecification.addSystemDependency(new ModuleDependency(moduleLoader, moduleIdentifier, false, true, true, false));
-            ROOT_LOGGER.debugf("added %s dependency to %s", moduleIdentifier, deploymentUnit.getName());
+            moduleSpecification.addSystemDependency(ModuleDependency.Builder.of(moduleLoader, searchMapperModuleName).
```
bstansberry, line 84: I believe the searchMapperModuleName var has not been canonicalized. It holds a value that was read from PersistenceUnitMetadata.getProperties().getProperty(Configuration.HIBERNATE_SEARCH_MODULE) and I don't expect that value to have been canonicalized.

I was going to suggest a fix, but the only use of this is to pass it into ModuleDependency.Builder, and that handles the canonicalization.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow and @yrodiere

PR #18446, opened 2024-11-28
[WFLY-19981] Add the resteasy-patchfilter-disabled attribute to the jaxrs subsystem.

### LINKED ISSUE #664: Add json merge patch support
### Description

This feature is to support JSON Merge Patch spec as described in: https://tools.ietf.org/html/rfc7386
With Json merge patch support, the client can send a json http request with Content-Type "application/merge-patch+json",
and the json merge request will be directly apply to the target resource with the merge function as spec described.

### Issue Contact

jimma

jamezp:
* https://issues.redhat.com/browse/WFLY-19981
* https://issues.redhat.com/browse/WFLY-20029

This replaces #18409

Proposal: https://github.com/wildfly/wildfly-proposals/pull/367

resolves https://github.com/wildfly/wildfly-proposals/issues/664

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
@@ -25,8 +25,6 @@ public class JaxrsExtension extends SubsystemExtension<JaxrsSubsystemSchema> {
 
     public static final String SUBSYSTEM_NAME = "jaxrs";
 
-    static final ModelVersion CURRENT_MODEL_VERSION = ModelVersion.create(4, 0, 0);
-
     private static final String RESOURCE_NAME = JaxrsExtension.class.getPackage().getName() + ".LocalDescriptions";
     static PathElement SUBSYSTEM_PATH = PathElement.pathElement(ModelDescriptionConstants.SUBSYSTEM, SUBSYSTEM_NAME);
 
@@ -44,14 +42,20 @@ public class JaxrsExtension extends SubsystemExtension<JaxrsSubsystemSchema> {
     }
 
     enum JaxrsSubsystemModel implements SubsystemModel {
-        VERSION_4_0_0,
+        VERSION_4_0_0(ModelVersion.create(4, 0, 0)),
+        VERSION_5_0_0(ModelVersion.create(5, 0, 0)),
         ;
 
-        static final JaxrsSubsystemModel CURRENT = VERSION_4_0_0;
+        static final JaxrsSubsystemModel CURRENT = VERSION_5_0_0;
+        private final ModelVersion version;
+
+        JaxrsSubsystemModel(final ModelVersion version) {
+            this.version = version;
+        }
 
         @Override
         public ModelVersion getVersion() {
-            return CURRENT_MODEL_VERSION;
+            return version;
         }
 
         @Override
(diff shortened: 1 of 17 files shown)


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
```diff
@@ -44,14 +42,20 @@ public JaxrsExtension() {
     }
 
     enum JaxrsSubsystemModel implements SubsystemModel {
-        VERSION_4_0_0,
+        VERSION_4_0_0(ModelVersion.create(4, 0, 0)),
+        VERSION_5_0_0(ModelVersion.create(5, 0, 0)),
         ;
 
-        static final JaxrsSubsystemModel CURRENT = VERSION_4_0_0;
+        static final JaxrsSubsystemModel CURRENT = VERSION_5_0_0;
```
jimma, line 49: I know we should bump this version to the next major one version from Brain's comment, but is there a reason we don't align this version with schema version ?

jamezp: They are just different things really. I don't recall how we ended up off. It had something to do with needing a model change because something was added. Either way, it's common for them not to match.

jaxrs/src/test/java/org/jboss/as/jaxrs/JaxrsSubsystem40TestCase.java
```diff
@@ -0,0 +1,104 @@
...
+        kernelServicesInitializer.addMavenResourceURL("org.wildfly:wildfly-jaxrs:26.0.0.Final");
+        KernelServices kernelServices = builder.build();
+        assertTrue(kernelServices.isSuccessfulBoot());
+        assertTrue(kernelServices.getLegacyServices(subsystemModelVersion).isSuccessfulBoot());
+
+        List<ModelNode> operations = builder.parseXmlResource("jaxrs.xml");
+        ModelTestUtils.checkFailedTransformedBootOperations(kernelServices, subsystemModelVersion, operations, transformationConfig);
+    }
+
+    private void testRejectingTransformers27(FailedOperationTransformationConfig transformationConfig, ModelTestControllerVersion controllerVersion) throws Exception {
```
jimma, line 89: Can we have a better name for this method ?

jamezp: Probably. The idea was for WildFly 27, but it can definitely be something else.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsExtension.java
bstansberry: Yes, IMHO we shouldn't attempt to align these. There are valid reasons for the numbers to get out of sync and that means it's not possible to keep them in alignment without doing things like skipping versions or introducing new models or schemas with no change. Both of which are confusing too.

bstansberry:
Note that github is showing me as a reviewer here, but I'm not planning on approving it. So don't hold up merging waiting for me. I'm leaving approval of this to the feature team handling https://github.com/wildfly/wildfly-proposals/issues/664.

(Don't take this comment as any sort of objection; I'm just not needed and don't want to be in the way.)

darranl:
Retested under https://github.com/wildfly/wildfly/pull/18500

PR #18452, opened 2024-11-28
Add a workaround for WFLY-20030 ReactiveMessaging* tests fail with podman-machine (macOS/Windows)

rhusar:
Workaround for https://issues.redhat.com/browse/WFLY-20030 until we find a proper solution.

### DIFF
--- a/testsuite/integration/microprofile/pom.xml
+++ b/testsuite/integration/microprofile/pom.xml
@@ -503,6 +503,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <!-- TODO Remove workaround for https://issues.redhat.com/browse/WFLY-20030 *ReactiveMessaging* tests fail with podman -->
+                    <reuseForks>false</reuseForks>
                     <environmentVariables>
                         <!-- dummy env property for MicroProfile Config testing -->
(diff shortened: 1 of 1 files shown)


kabir (approved):
Just to emphasise, this issue is not happening on CI. It is during local development using podman.

rhusar:
There's a bunch of [issues](https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxSmJdk17/470525?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&hideProblemsFromDependencies=false&expandBuildChangesSection=true) in the JDK 17 SM job, kick off a few reruns...

rhusar:
Bad news @kabir - these fail consistently and do appear to be a legitimate security manager issue with reactive messaging. Opened https://issues.redhat.com/browse/WFLY-20035 - so sadly, this workaround approach is currently not viable so our tests remain problematic with podman-machine :-(

bstansberry:
@rhusar @kabir I don't get why this would cause SM failures. IOW I'm concerned we have problems elsewhere and somehow this exposed them. The io.vertx.core.logging.LoggerFactory reading the TCCL in a static initializer goes back a long way though, so perhaps this change is somehow causing that static initializer to be run from deployment code whereas before it's always first run from subsystem code.

bstansberry:
Never mind my question; I just read https://issues.redhat.com/browse/WFLY-20035

bstansberry:
Maybe https://github.com/wildfly/wildfly/pull/18466 will fix it.

PR #18461, opened 2024-12-02
[WFLY-20054] Use the galleon-pack/pom.xml as the parent for the 'build' and 'dist'…

bstansberry:
… modules.

Make it easier to use a different groupId and version scheme for these downstream.

Ideally these modules would move under "galleon-pack", and ee-build and ee-dist would be under ee-galleon-pack. Both being similar to the 'preview' structure. But that can be disruptive for so late in WF 35 development as external code may expect the current decade+ old filesystem structure.

https://issues.redhat.com/browse/WFLY-20054

### DIFF
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -12,12 +12,13 @@
 
     <parent>
         <groupId>org.wildfly</groupId>
-        <artifactId>wildfly-parent</artifactId>
+        <artifactId>wildfly-feature-pack-parent</artifactId>
         <!--
         Maintain separation between the artifact id and the version to help prevent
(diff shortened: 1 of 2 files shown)


bstansberry:
@yersan If this is helpful for you please approve and I'll merge it.

yersan (approved):

yersan:
@bstansberry I've approved, it makes sense to tighten these two modules to the galleon pack which is used by them so +1 but only for that reason. As I understand it, switching their group ID to produce an XP galleon pack under a different Group ID is optional. This is a step forward to align with a better structure.

bstansberry:
Thanks @yersan

PR #18462, opened 2024-12-02
[WFLY-20039] Only use testsuite.full.galleon.pack.groupId|artifactId|…

bstansberry:
…version in testsuite/integration/microprofile-tck

https://issues.redhat.com/browse/WFLY-20039


jamezp (approved):

bstansberry:
I added the 'hold' label as it looks like this broke something in the MP REST Client TCK wit -Dts.layers.

bstansberry:
The CI results on this are wacky; there are continually new jobs being spun up for the same commit. When I dig into https://ci.wildfly.org/project.html?projectId=WF_PullRequest&branch_WF_PullRequest=pull%2F18462 I see that the current commit has a passing run on all the jobs. So I'll merge this and cancel the ones that are still running.

bstansberry:
Thanks @jamezp

PR #18467, opened 2024-12-03
[WFLY-20084] Add profiles to wildfly sources to separate base and expansion builds/tests

bstansberry:
…pansion FP

Allow limiting the build to just the base FP or just the expansion FP
    
Control TS execution via aggregator modules decoupled from behavioral configuration of the individual TS modules.

https://issues.redhat.com/browse/WFLY-20084


pom.xml
```diff
@@ -1447,6 +1375,95 @@
...
     <profiles>
+
+        <profile>
+            <id>base.build</id>
+            <activation>
+                <property>
+                    <name>!no.base.build</name>
+                </property>
+            </activation>
+            <modules>
```
yersan, line 1386: Should not these profiles (`!no.base.build` and `!no.expansion.build`) also get control over what profiles are activated in the testsuite? or it is expected to use these profiles in combination with `-P-base-server-tests`, `-P-expansion-server-tests` and / or `-P-Dpreview-server-tests`

I don't see such an interaction between those profiles yet

bstansberry: I expect users to use these in combination.

My goal here is maximum flexibility, not maximum ease of use, as this is all for advanced cases. (Even users building from source is an advanced use case.)

An unfortunate thing is maven does not support profile activation via multiple properties.

testsuite/README.md
```diff
@@ -83,68 +165,70 @@ Run a selected set of modules and tests
 ---------------------------------
```
yersan, line 165:
```
5. Building server configurations
---------------------------------
```

It doesn't render well in the final document, I think you have to use `5.0 Building server configurations` instead

yersan (approved):
Added a minor detail, but lgtm.

It also partially resolves https://issues.redhat.com/browse/WFLY-19749, It looks the missing point is just to describe how to run a single test, in case you want to add here now, we can resolve both

pom.xml
bstansberry: @yersan You already approved this, but BTW, the most relevant flexibility is:

mvn install -P-base-server-tests

That builds everything but only runs the expansion tests.

If instead the only option was

mvn install -Dno.base.build

then the user has to figure out how ensure the wanted base components are available. Some options:

1) Build the snapshots first and cache them in a repo and configure the build to use that repo. This is a valid approach that would make sense in CI workflows that want to build shared bits and then launch parallel tasks to run the tests.

2) Perhaps use released base server components and also do -Dee.maven.version=xxxx so the TS doesn't try to compile against or provision snapshots. This could also be a valid approach in some scenarios, where the user's goal is to test existing released bits. Say in combination with -Dexternal.wildfly.channels where the main goal is to test using later components specified in the channel.

Both of those are valid but at this point I don't want to *require* one of them.

testsuite/README.md
bstansberry: thank you. I fixed a bunch of cases like that; i don't know how I missed this one!

pom.xml
yersan: > mvn install -P-base-server-tests

ok, that's a good one, I was dealing with the possible usage of the `-Dno.base.build` and `-Dno.expansion.build` without realizing about that one.

bstansberry:
Thanks @yersan.

Good point about https://issues.redhat.com/browse/WFLY-19749, but I'll leave that open for now as I don't have time for that single test bit, plus I don't mind if others take a look.  I'll comment about how this change definitely addressed a lot of things though.

PR #18470, opened 2024-12-03
WFLY-20047 Update Narayana LRA 0.0.10.Final (plus WFLY-20050 changes to change the endpoint of the lra-coordinator)

tomjenkinson:
https://issues.redhat.com/browse/WFLY-20047
https://issues.redhat.com/browse/WFLY-20050

### DIFF
--- a/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/CommonAttributes.java
+++ b/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/CommonAttributes.java
@@ -6,7 +6,7 @@ package org.wildfly.extension.microprofile.lra.participant;
 
 interface CommonAttributes {
     String LRA_COORDINATOR_URL = "lra-coordinator-url";
-    String DEFAULT_COORDINATOR_URL = "http://localhost:8080/lra-coordinator/lra-coordinator";
+    String DEFAULT_COORDINATOR_URL = "http://localhost:8080/lra-coordinator";
 
     String PROXY_SERVER = "proxy-server";
     String PROXY_HOST = "proxy-host";
--- a/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/MicroProfileLRAParticipantSubsystemModel.java
+++ b/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/MicroProfileLRAParticipantSubsystemModel.java
@@ -13,7 +13,7 @@ import org.jboss.as.controller.SubsystemModel;
  * @author Paul Ferraro
  */
 enum MicroProfileLRAParticipantSubsystemModel implements SubsystemModel {
-    VERSION_1_0_0(1),
+    VERSION_2_0_0(2),
     ;
 
     private final ModelVersion version;
--- a/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/Namespace.java
+++ b/microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/Namespace.java
@@ -12,13 +12,13 @@ enum Namespace {
     // must be first
     UNKNOWN(null),
 
-    LRA_PARTICIPANT_1_0("urn:wildfly:microprofile-lra-participant:1.0"),
+    LRA_PARTICIPANT_2_0("urn:wildfly:microprofile-lra-participant:2.0"),
     ;
 
     /**
      * The current namespace version.
      */
-    public static final Namespace CURRENT = LRA_PARTICIPANT_1_0;
+    public static final Namespace CURRENT = LRA_PARTICIPANT_2_0;
 
     private final String name;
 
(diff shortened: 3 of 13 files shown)


microprofile/lra/participant/src/main/resources/schema/wildfly-microprofile-lra-participant_1_0.xsd
```diff
@@ -23,7 +23,7 @@
           ]]>
         </xs:documentation>
       </xs:annotation>
-      <xs:attribute name="lra-coordinator-url" type="xs:string" default="http://localhost:8080/lra-coordinator/lra-coordinator"/>
+      <xs:attribute name="lra-coordinator-url" type="xs:string" default="http://localhost:8080/lra-coordinator"/>
```
bstansberry, line 26: To be truly correct there needs to be a new 2.0 version of this schema, with 1.0 unchanged.  We'd need to tweak MicroProfileLRAParticipantSubsystemSchema.getXMLDescription() as well so when it processes the 1.0 namespace and sees no setting for this attribute it explicitly writes the old value.

The problem is AIUI we don't actually want to do that 'correct' behavior because it won't work -- the local coordinator won't understand that old URL.

Probably there should be a 2.0 schema, and the 1.0 documentation text of the 1.0 schema can be updated to explain the behavior on newer servers that support 2.0? Or just leave it unexplained? The thing is the 1.0 xsd exists and is published and still has meaning in contexts other than a WF 35+ server.

bstansberry (changes requested):

microprofile/lra/participant/src/main/resources/schema/wildfly-microprofile-lra-participant_1_0.xsd
tomjenkinson: I could add `Please note that the default value will no longer work to connect to WildFly servers incorporating WFLY-20050` to https://github.com/wildfly/wildfly/blob/main/microprofile/lra/participant/src/main/resources/schema/wildfly-microprofile-lra-participant_1_0.xsd#L20 (I would probably change `The "url"` to `The "lra-coordinator-url"` too) as a separate new WFLY (maybe a Bug or a Task "Update text in schema wildfly-microprofile-lra-participant_1_0.xsd") if you think that would be helpful (and also importantly, if that is the kind of wording change you would agree is correct :) )? Possibly we can do that as a follow up after this pull request?

microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/CommonAttributes.java
```diff
@@ -6,7 +6,7 @@
 
 interface CommonAttributes {
     String LRA_COORDINATOR_URL = "lra-coordinator-url";
-    String DEFAULT_COORDINATOR_URL = "http://localhost:8080/lra-coordinator/lra-coordinator";
+    String DEFAULT_COORDINATOR_URL = "http://localhost:8080/lra-coordinator";
```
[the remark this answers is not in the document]
tomjenkinson: Thank you, I _think_(/hope) I have bumped the management API  version - please can I ask you to have a look at https://github.com/wildfly/wildfly/pull/18470/commits/94b6a1a3d517649af942f3dce9291925870068a7 and see if you it looks good enough?

marcosgopen (approved):
Thanks @tomjenkinson for your PR!, LGTM
If @bstansberry is ok with your latest changes I am approving it too.

microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/MicroProfileLRAParticipantSubsystemModel.java
```diff
@@ -13,7 +13,7 @@
  * @author Paul Ferraro
  */
 enum MicroProfileLRAParticipantSubsystemModel implements SubsystemModel {
-    VERSION_1_0_0(1),
+    VERSION_2_0_0(2),
```
bstansberry, line 16: The old enum value should be retained and a new one added.

Needing the old one is relevant when it comes to domain mode transformers, which aren't needed here, but regardless this should follow the general pattern and have an enum value for each API version that exists.

microprofile/lra/participant/src/main/java/org/wildfly/extension/microprofile/lra/participant/Namespace.java
bstansberry, line 1: This entire file can be removed.

microprofile/lra/participant/src/main/resources/schema/wildfly-microprofile-lra-participant_1_0.xsd
bstansberry: We can definitely do that as a follow up.

The wording is tricky as what will happens depends on the context when a document using this 1.0 schema is parsed. What you wrote is correct if the context is a server before WF 35. If the context is WF 35 or later reading an old config that has a 1.0 schema element, then it won't actually use the default stated in the XSD; it will use the new default.

But anyway, NBD and can definitely be dealt with later.

bstansberry (approved):
@tomjenkinson @marcosgopen This is ok.  

I make 3 comments below. I'll send up a PR in a bit for two of them, as they are trivial, but that doesn't need to be merged for WF 35 Beta and doesn't block this.  The 3rd one is part of our discussion about the 1.0 xsd and that can come later as well.

marcosgopen:
> I am hearing that this change to register on "/" is causing a problem so we need to hold off on merging this /cc @marcosgopen @mnovak1 @xstefank

Yes, since this change is causing a problem we are evaluating to revert the change.

PR #18471, opened 2024-12-03
[WFLY-20037 - WFLY-19496] - OIDC Authentication Request and Scope com…

pedro-hos:
…munity promotion

https://issues.redhat.com/browse/WFLY-20037
https://issues.redhat.com/browse/WFLY-19496

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
@@ -147,8 +147,8 @@ will continue normally.
 
 === Scope Configuration
 
-NOTE: Since this attribute is a preview level feature, the WildFly server's stability level must be set to `preview`
-or lower to enable it. The stability level must not be changed to `community` or `default` after configuring the
+NOTE: Since this attribute is a community level feature, the WildFly server's stability level must be set to `community`
+or lower to enable it. The stability level must not be changed to `default` after configuring the
 `scope` attribute. For more information about the stability levels of the WildFly server, please refer to the
 link:Admin_Guide{outfilesuffix}[Admin Guide], particularly the
 link:Admin_Guide{outfilesuffix}#Feature_stability_levels[Feature stability levels] section.
@@ -175,9 +175,9 @@ NOTE: If both the deployment and subsystem configuration is present, then the su
 
 == Sending a Request Object as a JWT
 
-NOTE: Sending the request object as a JWT is a preview level feature, meaning the WildFly server's stability level must
-be set to `preview` or lower to enable the ability to add `request` or `request_uri` to authentication requests. The
-stability level must not be changed to community or higher after configuring the attributes related to this feature.
+NOTE: Sending the request object as a JWT is a community level feature, meaning the WildFly server's stability level must
+be set to `community` or lower to enable the ability to add `request` or `request_uri` to authentication requests. The
+stability level must not be changed to `default` after configuring the attributes related to this feature.
 For more information about the stability levels of the WildFly server, please refer to the
 link:Admin_Guide{outfilesuffix}[Admin Guide], particularly the
 link:Admin_Guide{outfilesuffix}#Feature_stability_levels[Feature stability levels] section.
--- a/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
+++ b/elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
@@ -94,9 +94,10 @@ public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<Elyt
     VERSION_2_0(2, Stability.DEFAULT),
     VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW), // WildFly 32.0-present
     VERSION_3_0_PREVIEW(3, 0, Stability.PREVIEW), // WildFly 33.0-present
+    VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY)
     ;
 
-    static final Map<Stability, ElytronOidcSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_3_0_PREVIEW, VERSION_2_0));
+    static final Map<Stability, ElytronOidcSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_3_0_PREVIEW, VERSION_3_0_COMMUNITY, VERSION_2_0));
     private static final AttributeParser SIMPLE_ATTRIBUTE_PARSER = new AttributeElementParser();
     private static final AttributeMarshaller SIMPLE_ATTRIBUTE_MARSHALLER = new AttributeElementMarshaller();
 
@@ -149,14 +150,14 @@ public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<Elyt
         Stream.of(providerDefaultAttributes).forEach(attribute -> secureDeploymentDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
         Stream.of(providerDefaultAttributes).forEach(attribute -> secureServerDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
 
-        if (this.since(VERSION_3_0_PREVIEW) && this.enables(AUTHENTICATION_REQUEST_FORMAT)) {
+        if ((this.since(VERSION_3_0_PREVIEW) || this.since(VERSION_3_0_COMMUNITY)) && this.enables(AUTHENTICATION_REQUEST_FORMAT)) {
             Stream.of(requestObjectAttributes).forEach(attribute -> secureDeploymentDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
             Stream.of(requestObjectAttributes).forEach(attribute -> secureServerDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
             Stream.of(requestObjectAttributes).forEach(attribute -> realmDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
             Stream.of(requestObjectAttributes).forEach(attribute -> providerDefinitionBuilder.addAttribute(attribute, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER));
         }
 
-        if (this.since(VERSION_2_0_PREVIEW) && this.enables(SCOPE)) {
+        if ((this.since(VERSION_2_0_PREVIEW) || this.since(VERSION_3_0_COMMUNITY)) && this.enables(SCOPE)) {
             secureDeploymentDefinitionBuilder.addAttribute(SCOPE, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER);
             secureServerDefinitionBuilder.addAttribute(SCOPE, SIMPLE_ATTRIBUTE_PARSER, SIMPLE_ATTRIBUTE_MARSHALLER);
         }
--- a/elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_community_3_0.xsd
+++ b/elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_community_3_0.xsd
@@ -1,0 +1,1584 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="urn:wildfly:elytron-oidc-client:community:3.0"
+           xmlns="urn:wildfly:elytron-oidc-client:community:3.0"
+           elementFormDefault="qualified"
+           attributeFormDefault="unqualified"
+           version="1.0">
+
+    <!-- The subsystem root element -->
+    <xs:element name="subsystem" type="subsystem-type"/>
+
+    <xs:complexType name="subsystem-type">
+        <xs:choice minOccurs="0" maxOccurs="unbounded">
+            <xs:element name="realm" maxOccurs="unbounded" minOccurs="0" type="realm-type">
+                <xs:annotation>
+                    <xs:documentation>
+                        <![CDATA[
+                        Configuration for a Keycloak realm. If you have multiple deployments secured by the same
+                        realm you can share the realm configuration here.
+                        ]]>
+                    </xs:documentation>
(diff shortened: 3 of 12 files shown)


elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -94,9 +94,10 @@ public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<Elyt
     VERSION_2_0(2, Stability.DEFAULT),
     VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW), // WildFly 32.0-present
     VERSION_3_0_PREVIEW(3, 0, Stability.PREVIEW), // WildFly 33.0-present
+    VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY)
```
gabrielpadilh4, line 97: Should it have a comment with the Wildfly version?

gabrielpadilh4 (approved):
LGTM.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
pedro-hos: I think that this is still related to 33.

fjuma:
Thanks for the PR, @pedro-hos! Looks like there are some test failures related to these changes, e.g., see:

https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk17/471813

pedro-hos:
@fjuma thank you for your feedback. Adding the missing files, let's see if the CI shows anything else.

pedro-hos:
@fjuma I think that the error isn't related to the changes. Please, let me know if I need to do something else. Thank you

pedro-hos:
Hi @fjuma do you think that we can add this to next WF version? Thanks

bstansberry:
@pedro-hos @darranl I'd really like to see these two features promoted in WildFly 37. :-) Pedro, if you are able to find someone from the security team to act as SME on these, I'd be willing to be the outside perspective person, unless someone else volunteers.  

This would require an RFE that follows https://docs.wildfly.org/wildfly-proposals/FEATURE_PROCESS.html. Perhaps two, if just one can't be made to cover both. I suspect one can cover; the analysis would just have two RFEs worth of content in Overview / User Stories / Release Note content.

The Jakarta MVC RFE that you're on is an example of a promotion RFE issue and analysis. https://github.com/wildfly/wildfly-proposals/issues/607

pedro-hos:
@bstansberry Thank you for the update. I'll the proposal issue tomorrow morning w/ you as outside perspective person and ping the elytron channel on zullip and the google groups asking for SME volunteer.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
bstansberry, line 97: Note to self / other reviewers:

3.0 COMMUNITY is ok but this promotion illustrates a case not fully described in the documentation of how to do updates in https://github.com/wildfly/wildfly.github.io/blob/main/wildfly-designs/ManagementAPIVersioning.adoc.

A 3.0 COMMUNITY is lower in the stack than the existing 3.0 PREVIEW and introduces new 'community' functionality. So per the way that is written it can't fit in the stack and needs to go on top as 4.0 COMMUNITY.

I think it's ok though because this doesn't actually "add" new functionality that now magically becomes part of a redefined 3.0 PREVIEW. It just makes functionality already available in 3.0 PREVIEW in a preview stability server also available in a community stability server. Nothing that worked in a preview stability server with 3.0 PREVIEW stops working and nothing that didn't work in a preview stability server with 3.0 PREVIEW starts working.

bstansberry: We shouldn't block on this but a nice to have would be

{code}
    VERSION_2_0(2, Stability.DEFAULT), // WildFly 39.0 - present 
    VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW), // WildFly 32.0-33.0
    VERSION_3_0_PREVIEW(3, 0, Stability.PREVIEW), // WildFly 33.0 - present
    VERSION_3_0_COMMUNITY(3, 0, Stability.COMMUNITY) // WildFly 39.0 - present
{code}

Note that VERSION_3_0_PREVIEW is still technically valid, hence the " - present" . It just doesn't have anything not also in VERSION_3_0_COMMUNITY.

elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_community_3_0.xsd
bstansberry, line 1: Per https://github.com/wildfly/wildfly.github.io/blob/main/wildfly-designs/ManagementAPIVersioning.adoc what we'd typically require would be 1 boilerplate commit that adds the management API version and xml schema with no difference in its content vs what it "replaces". And then a separate commit that adds whatever is new that led to having the new thing. This allows reviewers to separately check the boilerplate commit and then check whatever is new.

I'm not going to ask for a refactor of this PR to do this though because:

1) expediency -- I'd like this in WF 39
2) What this "replaces" isn't so obvious. Is it VERSION_2_0, the thing that will no longer take effect in a OOTB  std WildFly? Or is it VERSION_3_0_PREVIEW, which this functionally replaces?  Either way I'm going to need to do manual file diffs to make sure things are right, so if I have to do that I don't see a ton of value in doing a refactor.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
bstansberry, line 1: Tangent: in the future we should add the expressions to all the elytron-oidc-client[-stability]-#.0.xml files, get rid of oidc-expressions.xml, and get rid of this test class. ElytronOidcClientSubsystemTestCase exercises all the elytron-oidc-client[-stability]-#.0.xml files and it should exercise the expressions as well.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
```diff
@@ -44,7 +44,7 @@ public static Collection<Object[]> parameters() {
 
 
     public ElytronOidcClientSubsystemTransformerTestCase(ModelTestControllerVersion controllerVersion, ModelVersion version) {
-        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_3_0_PREVIEW, ElytronOidcSubsystemSchema.CURRENT.get(Stability.PREVIEW));
+        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_3_0_COMMUNITY, ElytronOidcSubsystemSchema.CURRENT.get(Stability.COMMUNITY));
```
bstansberry, line 47: This and a lot of the other similar constructor changes aren't formally correct as VERSION_3_0_COMMUNITY still exists. Better would be to use @Parameterized and cover both.

But I'm ok with just using VERSION_3_0_COMMUNITY.

elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_community_3_0.xsd
bstansberry: I've compared this file to preview_3_0.xsd and 2_0.xsd and the difference is as expected.

elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_community_3_0.xsd
```diff
@@ -0,0 +1,1584 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
```
bstansberry, line 2: If you need to update this PR for some other reason, please add the copyright/license header here. See wildfly-elytron-oidc-client_2_0.xsd.

Bonus points for adding it to other xsds in the same dir that don't have it.

bstansberry (approved):
LGTM. I've made various comments but I'd be ok with this as is.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
darranl: We had a big discussion about this in Zulip a while back, I thought we can to the conclusion that trying to apply meaning to the digits was doomed to fail.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Elytron_OIDC_Client.adoc
```diff
@@ -147,8 +147,8 @@ will continue normally.
 
 === Scope Configuration
 
-NOTE: Since this attribute is a preview level feature, the WildFly server's stability level must be set to `preview`
-or lower to enable it. The stability level must not be changed to `community` or `default` after configuring the
+NOTE: Since this attribute is a community level feature, the WildFly server's stability level must be set to `community`
+or lower to enable it. The stability level must not be changed to `default` after configuring the
```
rhusar, line 151: The 'community' is the default stability level for WildFly. I am not sure whether this isn't more confusing to the users than necessary.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
rhusar: +1 I would prefer we add the documentation which version this was used in. It saves a LOT of time doing version/SCM arecheology.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
```diff
@@ -29,7 +29,7 @@ public class ExpressionsTestCase extends AbstractSubsystemSchemaTest<ElytronOidc
     private KernelServices services = null;
 
     public ExpressionsTestCase() {
-        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_2_0_PREVIEW, ElytronOidcSubsystemSchema.CURRENT.get(Stability.PREVIEW));
+        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_3_0_COMMUNITY, ElytronOidcSubsystemSchema.CURRENT.get(Stability.COMMUNITY));
```
rhusar, line 32: Leverage AbstractSubsystemSchemaTest(String, Extension, S, Set\<S\>) constructor introduced in WFCORE-7416

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/OidcTestCase.java
```diff
@@ -32,7 +32,7 @@ public class OidcTestCase extends AbstractSubsystemSchemaTest<ElytronOidcSubsyst
     private KernelServices services = null;
 
     public OidcTestCase() {
-        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_2_0_PREVIEW, ElytronOidcSubsystemSchema.CURRENT.get(Stability.PREVIEW));
+        super(ElytronOidcExtension.SUBSYSTEM_NAME, new ElytronOidcExtension(), ElytronOidcSubsystemSchema.VERSION_3_0_COMMUNITY, ElytronOidcSubsystemSchema.CURRENT.get(Stability.COMMUNITY));
```
rhusar, line 35: Leverage AbstractSubsystemSchemaTest(String, Extension, S, Set\<S\>) constructor introduced in WFCORE-7416

rhusar (changes requested):
Also, the https://issues.redhat.com/browse/WFLY-21298 should be fixed first before making any further changes to the subsystems.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
darranl: @bstansberry I am looking at how we could do this but an issue I see is we could lose all the non-expression testing.

darranl: Actually checking the config all of our child resources appear to be named, this means I can add variants with expressions without losing the non express testing.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
darranl: I am considering just dropping this test down to cover default stability only.

darranl:
FYI I am going to be submitting a new PR for this one built on top of a management model and schema bump I have prepared.

elytron-oidc-client/src/test/resources/org/wildfly/extension/elytron/oidc/oidc.xml
```diff
@@ -3,7 +3,7 @@
   ~ SPDX-License-Identifier: Apache-2.0
   -->
 
-<subsystem xmlns="urn:wildfly:elytron-oidc-client:preview:3.0">
+<subsystem xmlns="urn:wildfly:elytron-oidc-client:community:3.0">
```
darranl, line 6: I have raised https://redhat.atlassian.net/browse/WFLY-21941 to revisit this test case, we should not be swapping the namespace here as the community variant adds nothing to the test but it does demonstrate that this test is now only aware of a single stability at a time.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/OidcTestCase.java
darranl, line 35: I have raised https://redhat.atlassian.net/browse/WFLY-21941 to revisit this test case, we should not be swapping the namespace here as the community variant adds nothing to the test but it does demonstrate that this test is now only aware of a single stability at a time.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ExpressionsTestCase.java
darranl: I have raised https://redhat.atlassian.net/browse/WFLY-21942 to revisit this test.

elytron-oidc-client/src/test/java/org/wildfly/extension/elytron/oidc/ElytronOidcClientSubsystemTransformerTestCase.java
darranl: I will rework the transofmer testing under https://redhat.atlassian.net/browse/WFLY-21940

Working with transformers on the lower stability levels does not seem a good approach now so I plan to simplify and target default only, additionally this is a parameterised test with a single parameter - I am not sure parameterisation is helping us but I do need to get the test updated so we can test tranformation to different versions.

PR #18482, opened 2024-12-05
[WFLY-20058]: "netty-common" should depend on "org.apache.logging.log4j.api".

ehsavoie:
* Updating the netty-common module.
* removing the infinispan workaround.
* removing the messaging workaround.

Jira: https://issues.redhat.com/browse/WFLY-20058

### DIFF
--- a/observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemExtension.java
+++ b/observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemExtension.java
@@ -5,8 +5,6 @@
 
 package org.wildfly.extension.opentelemetry;
 
-import io.netty.util.internal.logging.InternalLoggerFactory;
-import io.netty.util.internal.logging.JdkLoggerFactory;
 import io.vertx.core.logging.LoggerFactory;
 import org.wildfly.subsystem.SubsystemConfiguration;
 import org.wildfly.subsystem.SubsystemExtension;
@@ -21,9 +19,6 @@ public class OpenTelemetrySubsystemExtension extends SubsystemExtension<OpenTele
                 OpenTelemetrySubsystemModel.CURRENT,
                 OpenTelemetrySubsystemRegistrar::new),
                 SubsystemPersistence.of(OpenTelemetrySubsystemSchema.CURRENT));
-
-        // Initialize the Netty logger factory
-        InternalLoggerFactory.setDefaultFactory(JdkLoggerFactory.INSTANCE);
         // Initialize the Vert.x logger factory
         //noinspection deprecation
         LoggerFactory.initialise();
--- a/clustering/infinispan/extension/pom.xml
+++ b/clustering/infinispan/extension/pom.xml
@@ -112,11 +112,6 @@
         </dependency>
 
         <!-- External dependencies -->
-        <!-- This is only required for the InfinispanExtension to initialize Netty's InternalLoggerFactory -->
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-common</artifactId>
-        </dependency>
         <dependency>
             <groupId>io.reactivex.rxjava3</groupId>
             <artifactId>rxjava</artifactId>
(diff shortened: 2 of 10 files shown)


jamezp (changes requested):
@ehsavoie We need to make the change in all of these extensions:

```
clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanExtension.java
messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingExtension.java
microprofile/reactive-messaging-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/reactive/messaging/MicroProfileReactiveMessagingExtension.java
observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemExtension.java
```

We should also add a comment as to why we're adding the module. Something like:
```xml
<!-- There is no requirement for this in Netty. This avoids a warning from
     JBoss Modules which gets logged when checking the InternalLoggerFactory
     for which logging framework to bind to. Netty isn't aggressive enough
     when attempting to see if log4j 2 is available and JBoss Modules
     is a bit too aggressive with logging the warning. For this reason
     it is best to just process through log4j 2.
 -->
```

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemExtension.java
```diff
@@ -5,9 +5,6 @@
 
 package org.wildfly.extension.opentelemetry;
 
-import io.netty.util.internal.logging.InternalLoggerFactory;
-import io.netty.util.internal.logging.JdkLoggerFactory;
```
jasondlee, line 9: IIRC, there's a netty dep in this module that was added just for this change, so that should probably be removed as it was in the clustering module above.

jasondlee (changes requested):

rhusar (changes requested):
Changes are needed as this removes the fix for the SM that again results in

```
Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getClassLoader")" in code source "(vfs:/content/rx-messaging-ported.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.rx-messaging-ported.war" from Service Module Loader")
```

ehsavoie:
@rhusar I was a bit too enthusiastic in my removal :)

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetrySubsystemExtension.java
jamezp: +1 I do recall you adding this.

jamezp:
> Changes are needed as this removes the fix for the SM that again results in
> 
> ```
> Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getClassLoader")" in code source "(vfs:/content/rx-messaging-ported.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.rx-messaging-ported.war" from Service Module Loader")
> ```

I don't see this failure. Am I missing something?

rhusar:
> > Changes are needed as this removes the fix for the SM that again results in
> > ```
> > Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getClassLoader")" in code source "(vfs:/content/rx-messaging-ported.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.rx-messaging-ported.war" from Service Module Loader")
> > ```
> 
> @rhusar I don't see this failure. Am I missing something?

Here - https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxSmJdk17/472439?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildDeploymentsSection=false&expandBuildChangesSection=true - in reactive messaging.

jamezp:
Ah, I see. Thank you @rhusar

jamezp (approved):

bstansberry:
@jamezp I removed the hold label you'd added, now that you've approved this.

jamezp:
Thank you. That was my bad, I forgot I added it :facepalm:

rhusar (changes requested):
LayersTestCase needs to be updated.

rhusar (approved):
LGTM now.

bstansberry:
Thanks @ehsavoie @rhusar @jamezp @jasondlee

PR #18485, opened 2024-12-06
WFLY-20064 PersistentResourceXMLDescription of elytron-oidc-client subsystem uses marshallers/parsers inconsistent with its schema.

pferraro:
https://issues.redhat.com/browse/WFLY-20064

The attributes of a credential or redirect-rewrite-rule element parse/marshall as XML attributes, not as XML element content.

Fixes errors exposed by wildfly-core integration CI runs using new parser implementation: https://github.com/wildfly/wildfly-core/pull/6261


bstansberry:
Tangent: This is mostly just a style preference, but to me the heavy use of xml elements for simple values in this subsystem's xml schema isn't great. If there's a significant update in the future perhaps reconsider those.

fjuma:
> Tangent: This is mostly just a style preference, but to me the heavy use of xml elements for simple values in this subsystem's xml schema isn't great. If there's a significant update in the future perhaps reconsider those.

Right, I don't like this either. It was originally designed this way to be consistent with the keycloak subsystem that this was replacing to make migration easy. So we kept the same structure as before.

pferraro:
@fjuma Can you review this?  It is blocking wildfly/wildfly-core#6261

fjuma (approved):

bstansberry:
Thanks @pferraro and @fjuma

PR #18488, opened 2024-12-06
[WFLY-20063] add prospero module to the distribution

spyrkob:
Issue: https://issues.redhat.com/browse/WFLY-20063

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -1150,6 +1150,12 @@
             </dependency>
 
             <dependency>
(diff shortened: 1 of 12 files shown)


wolfc:
This PR creates a lot of code duplication with https://github.com/wildfly-extras/prospero. This makes little sense to me.

bstansberry:
Adding all those maven deps to our typical dists seems like too much relative to the benefit, so sticking with a separate feature pack makes more sense.

PR #18489, opened 2024-12-06
WFLY-19689 - Micrometer extension keeps pushing metrics after removal and reload

jasondlee:
https://issues.redhat.com/browse/WFLY-19689

- Add .onRemove() to Micrometer ResourceServiceInstaller to close registry when subsystem is remove 
- Remove default implementation in WildFlyRegister to prevent hiding method in superclass 
- Add test to verify that registry is shut down when micrometer subsystem is removed

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
@@ -44,6 +44,7 @@ import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.wildfly.common.function.Functions;
 import org.wildfly.extension.micrometer.jmx.JmxMicrometerCollector;
 import org.wildfly.extension.micrometer.metrics.MicrometerCollector;
 import org.wildfly.extension.micrometer.registry.NoOpRegistry;
@@ -170,7 +171,6 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
             }
         });
 
-
         AtomicReference<MicrometerCollector> captor = new AtomicReference<>();
 
         context.addStep((operationContext, modelNode) -> {
@@ -187,10 +187,11 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
         }, OperationContext.Stage.VERIFY);
 
         return CapabilityServiceInstaller.builder(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY, collectorSupplier)
-                .requires(List.of(mccf, executor, processStateNotifier))
-                .withCaptor(captor::set) // capture the provided value
-                .asActive() // Start actively
-                .build();
+            .requires(List.of(mccf, executor, processStateNotifier))
+            .withCaptor(captor::set) // capture the provided value
+            .onStop(Functions.closingConsumer())
+            .asActive() // Start actively
+            .build();
     }
 
     public interface MicrometerDeploymentConfiguration {
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyRegistry.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyRegistry.java
@@ -20,7 +20,7 @@ import org.wildfly.extension.micrometer.MicrometerExtensionLogger;
 import org.wildfly.extension.micrometer.metrics.MetricMetadata;
 import org.wildfly.extension.micrometer.metrics.WildFlyMetric;
 
-public interface WildFlyRegistry {
+public interface WildFlyRegistry extends AutoCloseable {
     Meter remove(Meter.Id mappedId);
 
     default Meter.Id addMeter(WildFlyMetric metric, MetricMetadata metadata) {
@@ -34,9 +34,7 @@ public interface WildFlyRegistry {
         }
     }
 
-    default void close() {
-
-    }
+    void close();
 
     private Meter.Id addCounter(WildFlyMetric metric, MetricMetadata metadata) {
         return FunctionCounter.builder(metadata.getMetricName(), metric,
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.integration.observability.micrometer;
+
+import java.io.File;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.io.input.Tailer;
+import org.apache.commons.io.input.TailerListener;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.test.shared.TimeoutUtil;
+import org.jboss.dmr.ModelNode;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.wildfly.plugin.tools.OperationExecutionException;
+import org.wildfly.plugin.tools.server.ServerManager;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+public class MicrometerRemovalTestCase {
+    private static final ModelNode micrometerExtension = Operations.createAddress("extension", "org.wildfly.extension.micrometer");
+    private static final ModelNode micrometerSubsystem = Operations.createAddress("subsystem", "micrometer");
+    private static final ModelNode ADDRESS_SERVER_LOG_DIR = Operations.createAddress("path", "jboss.server.log.dir");
+    private static final String ERROR_MESSAGE = "Failed to publish metrics to OTLP receiver";
+
+    @ContainerResource
+    private ServerManager serverManager;
+    private String logFilePath;
+
+    @Before
+    public void getLogLocation() throws IOException {
+        ModelNode op = Operations.createReadAttributeOperation(ADDRESS_SERVER_LOG_DIR, "path");
+        logFilePath = serverManager.executeOperation(op).asString();
+    }
+
+    @Test
+    public void testRemoval() throws Exception {
+        ServerLogTailerListener listener = new ServerLogTailerListener();
+        try (Tailer ignored = Tailer.builder()
+            .setFile(new File(logFilePath, "server.log"))
+            .setTailerListener(listener)
+            .setDelayDuration(Duration.ofMillis(500))
+            .get()) {
+            enableMicrometer();
+            Thread.sleep(TimeoutUtil.adjust(1000));
+            disableMicrometer();
+            // Micrometer will push one last time while the registry is shutting down. Sleep long enough to allow that to
+            // happen, then clear the log, wait, then check again. The server is configured to push every millisecond, so
+            // 500 should be sufficient to give that time without slowing down the test suite more than necessary.
+            Thread.sleep(TimeoutUtil.adjust(1000));
+            listener.logs.clear();
+            Thread.sleep(TimeoutUtil.adjust(1000));
+            Assert.assertTrue("Micrometer has been removed, but errors are still being logged.",
+                listener.logs.stream().noneMatch(l -> l.contains(ERROR_MESSAGE)));
+        } finally {
+            disableMicrometer();
+        }
+    }
+
+    protected void enableMicrometer() throws IOException {
+        try {
+            if (!resourceExists(micrometerExtension)) {
+                serverManager.executeOperation(Operations.createAddOperation(micrometerExtension));
+            }
+            if (!resourceExists(micrometerSubsystem)) {
+                ModelNode addOp = Operations.createAddOperation(micrometerSubsystem);
+                addOp.get("endpoint").set("http://localhost:4318/v1/metrics/v1/metrics");
+                addOp.get("step").set("1");
+                serverManager.executeOperation(addOp);
+            }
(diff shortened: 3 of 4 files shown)


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -187,10 +187,11 @@ public Predicate<String> getSubsystemFilter() {
         }, OperationContext.Stage.VERIFY);
 
         return CapabilityServiceInstaller.builder(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY, collectorSupplier)
-                .requires(List.of(mccf, executor, processStateNotifier))
-                .withCaptor(captor::set) // capture the provided value
-                .asActive() // Start actively
-                .build();
+            .requires(List.of(mccf, executor, processStateNotifier))
+            .withCaptor(captor::set) // capture the provided value
+            .onRemove(wildFlyRegistry::close)
```
jasondlee, line 192: @pferraro Would appreciate you chiming in on this to make sure it's the intended use of the API.

bstansberry:
@darranl I put the Blocker label on this because it is a 35 Final blocker. It's not a Beta blocker, although good to get in if it's ready.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyRegistry.java
```diff
@@ -34,9 +34,7 @@ default Meter.Id addMeter(WildFlyMetric metric, MetricMetadata metadata) {
         }
     }
 
-    default void close() {
-
-    }
+    void close();
```
jamezp, line 37: Not critical, but IMO this should probably just extend `AutoCloseable`. Keeping the method here to remove the exception seems fine. Again, this is not critical.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
```diff
@@ -0,0 +1,135 @@
...
+        // /path=jboss.server.log.dir:read-attribute(name="path")
+        ModelNode address = Operations.createAddress("path", "jboss.server.log.dir");
+        ModelNode op = Operations.createReadAttributeOperation(address, "path");
+        final ModelNode result = managementClient.getControllerClient().execute(op);
+        logFilePath = result.get("result").asString();
+    }
+
+    @Test
+    public void testRemoval() throws Exception {
+        MicrometerSetupTask setupTask = new MicrometerSetupTask();
```
jamezp, line 53: Is there a reason this is done here and not in a `@ServerSetup()`? The `tearDown()` is not guaranteed to be executed here and could leave the test server in an inconsistent state.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
```diff
@@ -0,0 +1,135 @@
...
+        ServerLogTailerListener listener = new ServerLogTailerListener();
+        try (Tailer ignored = Tailer.builder()
+            .setFile(new File(logFilePath, "server.log"))
+            .setTailerListener(listener)
+            .setDelayDuration(Duration.ofMillis(500))
+            .get()) {
+            setupTask.setup(managementClient, "");
+            Thread.sleep(1000);
```
jamezp, line 61: These sleeps should probably use `TimeoutUtil.adjust(1000)`.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
```diff
@@ -0,0 +1,135 @@
...
+            setupTask.tearDown(managementClient, "");
+            // Micrometer will push one last time while the registry is shutting down. Sleep long enough to allow that to
+            // happen, then clear the log, wait, then check again. The server is configured to push every millisecond, so
+            // 500 should be sufficient to give that time without slowing down the test suite more than necessary.
+            Thread.sleep(1000);
+            listener.logs.clear();
+            Thread.sleep(1000);
+            System.out.println(listener.logs);
```
jamezp, line 69: Is this cruft?

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
```diff
@@ -0,0 +1,135 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Arquillian.class)
+@DockerRequired
+@RunAsClient
+public class MicrometerRemovalTestCase {
+    private static final String ERROR_MESSAGE = "Failed to publish metrics to OTLP receiver";
```
jamezp, line 35: Where does this message come from? If the message is internationalized, this could cause the test to fail.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/WildFlyRegistry.java
jasondlee: Good call. Done. Updated the switch statement above to newer language standard.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
jasondlee: That's a good point. I need to manually control the tear down in the test, but I need to make sure that it IS run in the event of some odd failure.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
jasondlee: Ah, yes. I always forget about that one.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
jasondlee: The `println` is, yes.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
jasondlee: That come from Micrometer. The full error message is "Failed to publish metrics to OTLP receiver (context: url=http://localhost:4318/v1/metrics/v1/metrics, resource-attributes={service.name=wildfly}): java.net.ConnectException: Connection refused". There's no code to search for, so I grabbed something that seemed small enough to be safe-ish without being likely to hit false positives.

rhusar: This is coming from directly from `io.micrometer.registry.otlp.OtlpMeterRegistry` and there is no i18n.

rhusar: @jamezp Speaking of - do we ever run our test suite with non-English locale? I am not aware that we do.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @ContainerResource
+    private ServerManager serverManager;
+    private String logFilePath;
+
+    @Before
+    public void getLogLocation() throws IOException {
+        ModelNode op = Operations.createReadAttributeOperation(ADDRESS_SERVER_LOG_DIR, "path");
+        logFilePath = Operations.readResult(serverManager.executeOperation(op)).asString();
```
jamezp, line 43: We discussed and I know a change is coming, but for others. We know this will fail :)

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/observability/micrometer/MicrometerRemovalTestCase.java
jamezp: @rhusar I don't think anything on ci.wildfly.org is setup that way. It really just depends on where it's ran TBH. I know in the past we've had some CI setup with`cs-CZ.UTF-8`.

jamezp (approved):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
rhusar, line 192: @jasondlee I am afraid is unfortunately not correct. The service supplies a MicrometerCollector but it's closing a WildFlyRegistry. So, it's the MicrometerCollector that should also implement AutoCloseable and do proper cleanup, among other things closing the WildFlyRegistry.

rhusar: Moreover, this should be onStop(..) instead.

rhusar (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
pferraro: onRemove(...) should conclude actions initiated by service installation.
onStop(...) should conclude actions initated by service start.

rhusar: Just to elaborate on my above comment, I am under suspicion, that the collector should NOT start pushing metrics on service installation, but rather than on actual service start, and thus analogously the stop should be onStop. Unless there are reasons that I am not seeing now?

For example: in JGroups channel service case, the provider provides a JChannel, onStart the channel is connected and onStop the channel is disconnected.

rhusar (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -187,10 +186,10 @@ public Predicate<String> getSubsystemFilter() {
         }, OperationContext.Stage.VERIFY);
 
         return CapabilityServiceInstaller.builder(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY, collectorSupplier)
-                .requires(List.of(mccf, executor, processStateNotifier))
-                .withCaptor(captor::set) // capture the provided value
-                .asActive() // Start actively
-                .build();
+            .requires(List.of(mccf, executor, processStateNotifier))
+            .withCaptor(captor::set) // capture the provided value
+            .asActive() // Start actively
```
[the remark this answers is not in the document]
jasondlee: I thought it might, but the test passes, so it looks like it's automatically closed (if it's not, then the interface isn't required, right)?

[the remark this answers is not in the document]
rhusar: 🤷🏼

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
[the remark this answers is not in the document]
jasondlee: TBH, I don't know. :) This API is still pretty new to me, so if you think that we should do that here, I'm happy to add it, but the test to verify that the metrics publication does indeed stop when the subsystem is removed is passing, so it seems like it's working, even if it's not "correct".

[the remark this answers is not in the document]
pferraro: Where would this be automatically closed?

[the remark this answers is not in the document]
jasondlee: I really don't know, but based on the feedback that I should make it Autocloseable, I inferred that there was framework support for that. I'm guessing based on questions now that that is not the case, so I'll certainly add the .onStop(), though I'm not sure why the test is passing with that out. I'll dig some and try to understand that better...

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
bstansberry: I agree with @rhusar here, but I think we may need to defer that the WF 36 as this is a 35 Blocker and needs to be ready for merge in 2 days.  This is convoluted but seems to work.

It's not WildFlyCollector that is pushing metrics on service installation; it's WildFlyOtlpRegistry that is pushing metrics as soon as this configure() method instantiates it.

I think WildFlyRegistry, JmxMicrometerCollector and MicrometerCollector are services, with WildFlyRegistry injected into the others via a Supplier. It's the builder for the WildFlyRegistry service that gets .onStop(Functions::closingConsumer). 

Or simpler they all go in one service that instantiates the WildFlyRegistry and the JmxMicrometerCollector.

In the old code WildFlyOtlpRegistry was constructed (and thus started publishing) in a service start.

jasondlee: Agreed. As I was working through this, there were several things I saw that I wanted to clean up, but I was trying to minimize the changes, given the timeline for this. I have filed a JIRA for this: https://issues.redhat.com/browse/WFLY-20148

rhusar (approved):
The main concerns have been addressed - looks good to me now.

This should be refactored  as discussed on here https://github.com/wildfly/wildfly/pull/18489#discussion_r1881121059 for which https://issues.redhat.com/browse/WFLY-20148 has been filed.

bstansberry:
Thanks @jasondlee @jamezp @rhusar @pferraro

PR #18495, opened 2024-12-09
[WFLY-20072] Include non shaded wildfly-cli artifact in tools user BOM

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-20072


bstansberry:
Thanks @emmartins

PR #18499, opened 2024-12-10
WFLY-20026 run an initial recovery scan when the microprofile-lra sub…

mmusgrov:
…system starts so that pending LRAs are visible

https://issues.redhat.com/browse/WFLY-20026

The PR runs an initial recovery scan when the microprofile-lra subsystem starts so that any in progress LRAs are immediately visible.

### DIFF
--- a/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
+++ b/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
@@ -16,10 +16,43 @@ import org.jboss.msc.service.StopContext;
 import org.wildfly.extension.microprofile.lra.coordinator._private.MicroProfileLRACoordinatorLogger;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.function.Supplier;
+
+import static org.wildfly.extension.microprofile.lra.coordinator.service.LRACoordinatorService.CONTEXT_PATH;
+
 public class LRARecoveryService implements Service {
     private static final Logger log = Logger.getLogger(LRARecoveryService.class);
 
+    private final Supplier<ExecutorService> executorSupplier;
+
     private volatile LRARecoveryModule lraRecoveryModule;
+
+    public LRARecoveryService(Supplier<ExecutorService> executorSupplier) {
+        this.executorSupplier = executorSupplier;
+    }
+
+    private void startRecoveryScan(final StartContext context) {
+        assert executorSupplier != null;
+        final ExecutorService service = executorSupplier.get();
+        final Runnable task = () -> {
+            try {
+                // run an initial recovery scan - skip periodicWorkFirstPass since it is a no-op
+                lraRecoveryModule.periodicWorkSecondPass();
+                context.complete();
+            } catch (Exception e) {
+                MicroProfileLRACoordinatorLogger.LOGGER.failedToRunRecoveryScan(CONTEXT_PATH, e);
+            }
+        };
+        try {
+            service.execute(task);
+        } catch (RejectedExecutionException e) {
+            task.run();
+        } finally {
+            context.asynchronous();
+        }
+    }
 
     @Override
     public synchronized void start(final StartContext context) throws StartException {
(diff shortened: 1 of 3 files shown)


microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
```diff
@@ -36,6 +36,8 @@ public synchronized void start(final StartContext context) throws StartException
                     .noneMatch(rm -> rm instanceof LRARecoveryModule)) {
                     RecoveryManager.manager().addModule(lraRecoveryModule);
                 }
+                // run an initial recovery scan - skip periodicWorkFirstPass since it is a no-op
+                lraRecoveryModule.periodicWorkSecondPass();
```
bstansberry, line 40: @mmusgrov I expect this blocks and thus unless @ropalka says otherwise it should use the MSC asynchronous API (StartContext.asynchronous() and from a runnable executed by an executor, StartContext.complete().

See the class and start method javadoc at https://github.com/jboss-msc/jboss-msc/blob/main/src/main/java/org/jboss/msc/Service.java

There are lots of examples of this in the WF code. The general pattern is to inject an executor that the server kernel provides into your service instance:

https://github.com/wildfly/wildfly/blob/main/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/BatchEnvironmentProcessor.java#L144-L146

and then use it in start

https://github.com/wildfly/wildfly/blob/main/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/job/repository/JdbcJobRepositoryService.java#L45-L61

bstansberry (changes requested):

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
mmusgrov: It can potentially block for up to [two seconds](https://github.com/jbosstm/lra/blob/0.0.10.Final/service-base/src/main/java/io/narayana/lra/LRAConstants.java#L66) for each participant if recovery needs to reissue JAX RS requests to participants. I'll update the PR along the lines you suggest ...

mmusgrov: I did all that but the StartContext.complete() call results in an IllegalStateException: Illegal controller state:

```
    private void startRecoveryScan(final StartContext context) {
        assert executorSupplier != null;
        final ExecutorService service = executorSupplier.get();
        final Runnable task = () -> {
            try {
                // run an initial recovery scan - skip periodicWorkFirstPass since it is a no-op
                lraRecoveryModule.periodicWorkSecondPass();
                context.complete(); // results in java.lang.IllegalStateException: Illegal controller state
            } catch (Exception e) {
                MicroProfileLRACoordinatorLogger.LOGGER.failedToRunRecoveryScan(CONTEXT_PATH, e);
            }
        };
        try {
            service.execute(task);
        } catch (RejectedExecutionException e) {
            task.run();
        } finally {
            context.asynchronous();
        }
    }
```
If I comment out context.complete(); it works - am I calling it in the correct place, I modelled the call based on your [last link](https://github.com/wildfly/wildfly/blob/main/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/job/repository/JdbcJobRepositoryService.java#L45-L61
).

mmusgrov: Ah wait, I was calling `context.complete();` elsewhere too - let me retest that ...

mmusgrov:
@xstefank I tested this PR using the reproducer in the WFLY-20026 issue tracker and I added a narayana test that verifies that calling lraRecoveryModule.periodicWorkSecondPass() is thread safe (https://github.com/jbosstm/lra/pull/23)

bstansberry (approved):

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
```diff
@@ -16,11 +16,44 @@
...
+    private final Supplier<ExecutorService> executorSupplier;
+
     private volatile LRARecoveryModule lraRecoveryModule;
 
+    public LRARecoveryService(Supplier<ExecutorService> executorSupplier) {
+        this.executorSupplier = executorSupplier;
+    }
+
+    private void startRecoveryScan(final StartContext context) {
+        assert executorSupplier != null;
```
xstefank, line 37: maybe this should be done in constructor

xstefank (approved):

bstansberry:
> Thanks Brian, I'll need someone to merge it for me.

Yep; I was waiting for CI.

bstansberry:
Thanks @mmusgrov and @xstefank

PR #18506, opened 2024-12-11
WFLY-20076 Intermittent failures in OpenTelemetry integration test cases

rhusar:
Fixes all intermittent otel test cases.

Introduces org.jboss.as.test.shared.observability.containers.OpenTelemetryCollectorContainer#assertMetrics(java.util.function.Consumer<java.util.List<org.jboss.as.test.shared.observability.signals.PrometheusMetric>>) that continually evaluates assertions provided in a consumer until the state obtained from the Prometheus endpoint matches the expected state or until a timeout elapses.  By default, polls the collector every second for 30 seconds. Returns snapshot of the prometheus registry that passed the assertions; typically ignored.

Resolves https://issues.redhat.com/browse/WFLY-20076

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/OpenTelemetryCollectorContainer.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/OpenTelemetryCollectorContainer.java
@@ -4,21 +4,34 @@
  */
 package org.jboss.as.test.shared.observability.containers;
 
+import static org.junit.Assert.assertTrue;
+
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import jakarta.ws.rs.client.Client;
 import jakarta.ws.rs.client.ClientBuilder;
 import jakarta.ws.rs.client.WebTarget;
+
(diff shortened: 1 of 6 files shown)


bstansberry (approved):

testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/OpenTelemetryCollectorContainer.java
```diff
@@ -76,26 +89,76 @@ public List<JaegerTrace> getTraces(String serviceName) throws InterruptedExcepti
...
-                body = target.request().get().readEntity(String.class);
-                found = body.contains(nameToMonitor);
-                attemptCount++;
+                debugLog("assertMetrics(..) validation passed.");
+
+                return prometheusMetrics;
+            } catch (AssertionError assertionError) {
+                debugLog("assertMetrics(..) validation failed - retrying.");
+                lastAssertionError = assertionError;
+                Thread.sleep(1000);
```
jasondlee, line 132: ThreadUtils.sleep(Duration.ofMillis(1000));

jasondlee (changes requested):

testsuite/shared/src/main/java/org/jboss/as/test/shared/observability/containers/OpenTelemetryCollectorContainer.java
jasondlee: Hrm. I think I was thinking of TimeoutUtils, which you're already using. Take or leave this one. :)

jasondlee (approved):

bstansberry:
Thanks @rhusar and @jasondlee

PR #18518, opened 2024-12-13
[WFLY-20102] only encode and log route for first getRoute call

aogburn:
https://issues.redhat.com/browse/WFLY-20102

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/Server.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/Server.java
@@ -47,6 +47,7 @@ public class Server implements Service<Server> {
     private final Supplier<ServletContainerService> servletContainer;
     private final Supplier<UndertowService> undertowService;
     private final String defaultHost;
+    private String finalRoute = null;
     private final String name;
     private final NameVirtualHostHandler virtualHostHandler = new NameVirtualHostHandler();
     private final List<ListenerService> listeners = new CopyOnWriteArrayList<>();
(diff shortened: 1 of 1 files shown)


bstansberry:
@fl4via Please review.

soul2zimate:
Hi @fl4via , Could you review this one ? We need same fix in downstream. Thanks.

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -162,6 +163,9 @@ public List<UndertowListener> getListeners() {
     }
 
     public String getRoute() {
+        if (this.finalRoute != null) {
```
pferraro, line 166: Since this is a dynamically generated value, it would be best to establish this within Service.start(...).

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
```diff
@@ -47,6 +47,7 @@ public class Server implements Service<Server> {
     private final Supplier<ServletContainerService> servletContainer;
     private final Supplier<UndertowService> undertowService;
     private final String defaultHost;
+    private String finalRoute = null;
```
pferraro, line 50: This needs to be volatile.

pferraro (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/Server.java
bstansberry: Note that doing that may require changes to some unit test stuff that uses a Server but doesn't call start/stop on it.

org.wildfly.mod_cluster.undertow.UndertowEngineTestCase, perhaps others.

PR #18525, opened 2024-12-13
WFLY-20077: do not set default value of derived attributes, remove du…

michpetrov:
…plicate method

Issue: [WFLY-20077](https://redhat.atlassian.net/browse/WFLY-20077)


messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
```diff
@@ -192,13 +161,11 @@ static AddressSettings createDefaultAddressSettings() {
         settings.setAutoCreateAddresses(AddressSettingDefinition.AUTO_CREATE_ADDRESSES.getDefaultValue().asBoolean());
         settings.setAutoDeleteAddresses(AddressSettingDefinition.AUTO_DELETE_ADDRESSES.getDefaultValue().asBoolean());
         settings.setAutoDeleteCreatedQueues(AddressSettingDefinition.AUTO_DELETE_CREATED_QUEUES.getDefaultValue().asBoolean());
-        settings.setMaxReadPageBytes(AddressSettingDefinition.MAX_READ_PAGE_BYTES.getDefaultValue().asInt());
```
ehsavoie, line 195: Why is that removed ??

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
```diff
@@ -192,13 +161,11 @@ static AddressSettings createDefaultAddressSettings() {
...
         settings.setDeadLetterAddress(asSimpleString(DEAD_LETTER_ADDRESS.getDefaultValue(), null));
         settings.setExpiryAddress(asSimpleString(EXPIRY_ADDRESS.getDefaultValue(), null));
         settings.setExpiryDelay(AddressSettingDefinition.EXPIRY_DELAY.getDefaultValue().asLong());
         settings.setDefaultLastValueQueue(AddressSettingDefinition.LAST_VALUE_QUEUE.getDefaultValue().asBoolean());
         settings.setMaxDeliveryAttempts(AddressSettingDefinition.MAX_DELIVERY_ATTEMPTS.getDefaultValue().asInt());
-        settings.setMaxRedeliveryDelay(AddressSettingDefinition.MAX_REDELIVERY_DELAY.getDefaultValue().asLong());
```
ehsavoie, line 201: You should also remove the default value on the attribute definition

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
michpetrov: @ehsavoie For the same reason as the other one, in this case the default for `maxReadPageBytes` is `2 * pageSizeBytes`

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
michpetrov: I wanted to say that would require a schema bump but it looks like the default values are not in the schema to begin with. Still, the attribute does technically have a default value, it just isn't explicitly set, shouldn't we mention that in the description?

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
ehsavoie: It has been set to this value for backward compatible reasons due to changes in paging

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
ehsavoie: Yes a schema bump is required since we need transformation of the default values

ehsavoie: You need also to add transformers

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
michpetrov: It's the same issue. We can end up in a situation where we're ignoring user setting because our default for a MAX_VALUE is less than a user-defined BASE_VALUE.

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
michpetrov: I don't follow. A transformer is for changing old attributes into new attributes, right? Since we don't use the default value anywhere we don't care if the old attributes have it.

ehsavoie: Transformers are there to transform new configuration into old configuration and fail before operation is sent to old systems.

michpetrov: I still don't understand, the default value being present is either a bug (if the default configuration relies on it) or it doesn't matter so adding it back for the old configuration would basically be supporting the buggy behavior.

bstansberry:
@michpetrov @ehsavoie I'd like to see this one get to a mergable state, or, if there's no consensus on the fix, closed.

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
ehsavoie: ok

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
ehsavoie: The default values won't be used because if the value is not defined then it should be computed based on the other attribute value and not a default value

ehsavoie: Thus we should convert the default value to a proper one in the transformer otherwise the buggy value is passed

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
ehsavoie: I agree

ehsavoie (changes requested):

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/AddressSettingAdd.java
michpetrov: Ok, I think I get it. The transformer (AttributeConverter) will compute the correct "default" value and we will pass that because otherwise it would get replaced by the bad default value.

michpetrov: @ehsavoie I've updated the PR, I assume we want a test for this but is there a point in running all the other tests that are copied over from the 16.0 testcase?

ehsavoie: I don't think so

ehsavoie: The version should be 17 not 16.1

michpetrov:
@ehsavoie updated the PR - since the transformation tests are being skipped I don't know what I can do there, so I only kept the one test for EAP 7.4. The only other relevant test I found is in the testsuite so I changed it to check if values are what we expect. Is there anything else?

bstansberry:
@ehsavoie Is this ok?

darranl:
Please keep schema bumps in their own commit so we can use the GitHub UI to see what actually changed in the schema.

However if we are doing schema bumps this feels like it is falling into feature development and may need to follow the feature development process.

michpetrov:
@darranl I'll split it I just wanted to see if the build passes, there's no overlap between the version bump and my changes though

this is primarily a bug fix, does it count as a feature when I'm not adding/removing attributes?

michpetrov:
@ehsavoie can you review this?

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingTransformerRegistration.java
```diff
@@ -78,7 +91,22 @@ private static void registerTransformers_WF_36(ResourceTransformationDescription
...
                 .addChildResource(ADDRESS_SETTING_PATH);
-        rejectDefinedAttributeWithDefaultValue(addressSettings, AddressSettingDefinition.MAX_READ_PAGE_BYTES);
+
+        addressSettings.getAttributeBuilder().setDiscard(new DiscardAttributeChecker.DefaultDiscardAttributeChecker() {
+
+                @Override
+                public boolean isOperationParameterDiscardable(PathAddress address, String attributeName, ModelNode attributeValue, ModelNode operation, TransformationContext context) {
+                    long baseValue = getBaseOrDefaultValue(operation, AddressSettingDefinition.PAGE_SIZE_BYTES);
+
+                    return attributeValue.asLong() == baseValue * 2L || attributeValue.equals(AddressSettingDefinition.MAX_READ_PAGE_BYTES.getDefaultValue());
```
ehsavoie: What is AddressSettingDefinition.MAX_READ_PAGE_BYTES.getDefaultValue() as this was removed ?

ehsavoie (changes requested):

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingTransformerRegistration.java
```diff
@@ -158,6 +187,36 @@ private static void rejectDefinedAttributeWithDefaultValue(ResourceTransformatio
...
                 .setDiscard(DiscardAttributeChecker.UNDEFINED, newAttribute)
-                .addRename(newAttribute, attribute.getName());
+                .addRename(newAttribute, attribute);
+    }
+
+    private static void computeDerivedValue(ResourceTransformationDescriptionBuilder resourceRegistry, AttributeDefinition baseAttribute, AttributeDefinition maxAttribute, long multiplier) {
+        resourceRegistry.getAttributeBuilder()
+                .setValueConverter(new AttributeConverter() {
+
+                    private long computedValue = 0L;
```
ehsavoie: I'm not super sure about creating a field here. Shouldn't computing the value be stateless ?

ehsavoie (changes requested):

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingTransformerRegistration.java
michpetrov: Right, that should be the actual value.

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingTransformerRegistration.java
michpetrov: This was wrong, we need access to the current baseAttribute value and I thought the operation converter could pre-compute the value for the resource converter but that's not how the conversion works. I need to retrieve the value from the transformation context.

michpetrov:
@ehsavoie ok, I think I fixed it but I'm not sure about the `convertOperationParameter` method, it's going to execute only if someone manually sets the value to "undefined", but not for the "undefine-attribute" operation. I'm assuming the `convertResourceAttribute` will get called before the value is retrieved but if that's the case I don't see what we would need the `convertOperationParameter` for.

ehsavoie (approved):

rhusar:
@ehsavoie @michpetrov Can you please resolve the discussions that were completed as a way to indicate that those conversations are resolved so a merger doesn't have to read through them to asses that? Thanks

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/MessagingTransformerRegistration.java
```diff
@@ -78,7 +93,22 @@ private static void registerTransformers_WF_36(ResourceTransformationDescription
...
                 .addChildResource(ADDRESS_SETTING_PATH);
-        rejectDefinedAttributeWithDefaultValue(addressSettings, AddressSettingDefinition.MAX_READ_PAGE_BYTES);
+
+        addressSettings.getAttributeBuilder().setDiscard(new DiscardAttributeChecker.DefaultDiscardAttributeChecker() {
+
+                @Override
+                public boolean isOperationParameterDiscardable(PathAddress address, String attributeName, ModelNode attributeValue, ModelNode operation, TransformationContext context) {
+                    long baseValue = getBaseOrDefaultValue(operation, AddressSettingDefinition.PAGE_SIZE_BYTES);
+
+                    return attributeValue.asLong() == baseValue * 2L || attributeValue.asLong() == 20971520L; // 2*10*1024*1024;
```
rhusar: I would have added a comment for the next guy here


```suggestion
                    // Discard if value matches the derived default (2 * page-size-bytes) or the previous hardcoded default (2 * 10 * 1024 * 1024)
                    return attributeValue.asLong() == baseValue * 2L || attributeValue.asLong() == 20971520L; // 2*10*1024*1024;
```

rhusar (approved):
LGTM

But on tangent, man this subsystem is out of date with current subsystem development mechanism and the MessagingSubsystemParser_18_0.java is a basically a verbatim copy of _17_0 with only functional difference is the namespace string.

rhusar:
Thanks @michpetrov and @ehsavoie

PR #18529, opened 2024-12-13
[WFLY-20120] Limit passing -Djava.security.manager=allow to SE 18 - 23

bstansberry:
https://issues.redhat.com/browse/WFLY-20120


bstansberry:
https://ci.wildfly.org/viewQueued.html?itemId=474551 is a custom run of the SE 24 job with this.

PR #18534, opened 2024-12-16
[WFLY-20015] Remove all non-breaking uses of ModuleIdentifier in Web …

soul2zimate:
Issue: https://issues.redhat.com/browse/WFLY-20015

[WFLY-20015] Remove all non-breaking uses of ModuleIdentifier in Web Services Subsystem

### DIFF
--- a/webservices/server-integration/src/main/java/org/jboss/as/webservices/util/ModuleClassLoaderProvider.java
+++ b/webservices/server-integration/src/main/java/org/jboss/as/webservices/util/ModuleClassLoaderProvider.java
@@ -7,7 +7,6 @@ package org.jboss.as.webservices.util;
 import java.lang.ref.WeakReference;
 
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.modules.ModuleLoadException;
 import org.jboss.wsf.spi.classloading.ClassLoaderProvider;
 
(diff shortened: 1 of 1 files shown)


webservices/server-integration/src/main/java/org/jboss/as/webservices/util/ModuleClassLoaderProvider.java
```diff
@@ -31,7 +31,7 @@ public ClassLoader getWebServiceSubsystemClassLoader() {
     public ClassLoader getServerIntegrationClassLoader() {
         if (integrationClassLoader == null || integrationClassLoader.get() == null) {
             try {
-                Module module = Module.getBootModuleLoader().loadModule(ASIL);
+                Module module = Module.getBootModuleLoader().loadModule(ModuleIdentifierUtil.canonicalModuleIdentifier(ASIL));
```
ropalka, line 34: It is not necessary to use ModuleIdentifierUtil helper class here. The string being used is standard module name.

ropalka (changes requested):

webservices/server-integration/src/main/java/org/jboss/as/webservices/util/ModuleClassLoaderProvider.java
bstansberry: +1.

Generally we should reserve use of ModuleIdentifierUtil.canonicalModuleIdentifier to making sure external inputs are valid, where users might be passing "org.foo:main" etc.

soul2zimate: Fixed.

bstansberry (approved):

ropalka (approved):

PR #18546, opened 2024-12-16
WFLY-20023 Remove all non-breaking uses of ModuleIdentifier in Concurrency

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-20023

### DIFF
--- a/concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
+++ b/concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
@@ -14,7 +14,6 @@ import org.jboss.as.ee.concurrent.handle.SetupContextHandle;
 import org.jboss.as.ee.logging.EeLogger;
 import org.jboss.as.server.CurrentServiceContainer;
 import org.jboss.modules.Module;
-import org.jboss.modules.ModuleIdentifier;
 import org.jboss.msc.service.ServiceContainer;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.msc.service.ServiceName;
@@ -210,7 +209,7 @@ public class ConcurrentContext {
             // switch to EE module classloader, otherwise, due to serialization, deployments would need to have dependencies to other internal modules
             final Module module;
             try {
-                module = Module.getBootModuleLoader().loadModule(ModuleIdentifier.create("org.jboss.as.ee"));
+                module = Module.getBootModuleLoader().loadModule("org.jboss.as.ee");
             } catch (Throwable e) {
                 throw new IOException(e);
             }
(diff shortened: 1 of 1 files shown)


bstansberry (changes requested):
Thanks @RanabirChakraborty. Needs a tweak.

Please be aware of https://wildfly.zulipchat.com/#narrow/channel/174177-ci-servers/topic/Multiple.20jobs when deciding whether to push this branch.

concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
```diff
@@ -210,7 +210,8 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
             // switch to EE module classloader, otherwise, due to serialization, deployments would need to have dependencies to other internal modules
             final Module module;
             try {
-                module = Module.getBootModuleLoader().loadModule(ModuleIdentifier.create("org.jboss.as.ee"));
+                String canonicalModuleName = ModuleIdentifierUtil.canonicalModuleIdentifier("org.jboss.as.ee");
```
ropalka, line 213: It is not necessary to use ModuleIdentifierUtil.canonicalModuleIdentifier @RanabirChakraborty  here because "org.jboss.as.ee" is standard module name.

ropalka (changes requested):

concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
RanabirChakraborty: Okay, but I'm unsure what to use here to replace `ModuleIdentifier`. If we want to replace it with `ModuleIdentifierUtil` then I guess it's the only way. If you have anything else in your mind, please suggest.

bstansberry: @RanabirChakraborty it's simply

module = Module.getBootModuleLoader().loadModule("org.jboss.as.ee");

The key thing I wrote ModuleIdentifierUtil.canonicalModuleIdentifier for was to ensure arbitrary module name strings that external users provide are in the canonical form. This isn't that; it's a static string the WF devs wrote, in the canonical form. It doesn't need to be canonicalized.

An incorrect form would be "org.jboss.as.ee:main", which an external user might think is correct, but isn't.

ropalka (approved):

bstansberry (approved):

bstansberry:
Simple view of TC test status (mostly for my own benefit this today): https://ci.wildfly.org/project.html?projectId=WF_PullRequest&branch_WF_PullRequest=pull%2F18546

bstansberry:
Thanks @RanabirChakraborty and @ropalka

PR #18560, opened 2024-12-17
[WFLY-20157] Remove all non-breaking uses of ModuleIdentifier in Connector subsystem

ropalka:
https://issues.redhat.com/browse/WFLY-20157

PR #18561, opened 2024-12-17
WFLY-20025 Remove all non-breaking uses of ModuleIdentifier in JCA

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-20025

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/services/driver/InstalledDriver.java
+++ b/connector/src/main/java/org/jboss/as/connector/services/driver/InstalledDriver.java
@@ -4,8 +4,6 @@
  */
 package org.jboss.as.connector.services.driver;
 
-import org.jboss.modules.ModuleIdentifier;
-
 /**
  * Metadata describing a JDBC driver that has been installed as a service in the
  * runtime.
@@ -14,7 +12,7 @@ import org.jboss.modules.ModuleIdentifier;
 public final class InstalledDriver {
 
     private final String driverName;
-    private final ModuleIdentifier moduleName;
+    private final String moduleName;
     private final String deploymentUnitName;
     private final String driverClassName;
     private final String dataSourceClassName;
@@ -38,7 +36,7 @@ public final class InstalledDriver {
      * @param minorVersion the driver minor version
      * @param jdbcCompliant whether the driver is JDBC compliant
      */
-    public InstalledDriver(final String driverName, final ModuleIdentifier moduleName, final String driverClassName,
+    public InstalledDriver(final String driverName, final String moduleName, final String driverClassName,
                            final String dataSourceClassName, final String xaDataSourceClassName,
                            final int majorVersion, final int minorVersion, final boolean jdbcCompliant) {
         this.deploymentUnitName = null;
@@ -85,10 +83,10 @@ public final class InstalledDriver {
     /**
      * Gets the name of the module from which the driver was loaded, if it was
      * loaded from the module path.
-     * @return the module name, or {@code null} if {@link #isFromDeployment()}
+     * @return the canonical module name, or {@code null} if {@link #isFromDeployment()}
      *         returns {@code true}
      */
-    public ModuleIdentifier getModuleName() {
+    public String getModuleName() {
         return moduleName;
     }
 
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/AbstractDataSourceService.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/AbstractDataSourceService.java
@@ -355,12 +355,10 @@ public abstract class AbstractDataSourceService implements Service<DataSource> {
                             throw ConnectorLogger.ROOT_LOGGER.cannotDeploy(ConnectorLogger.ROOT_LOGGER.failedToLoadDataSourceClass(dsClsName, e));
                         }
                     }
-
                     String driverName = dataSourceConfig.getDriver();
                     InstalledDriver installedDriver = driverRegistry.getValue().getInstalledDriver(driverName);
                     if (installedDriver != null) {
-                        String moduleName = installedDriver.getModuleName() != null ? installedDriver.getModuleName().getName()
-                                : null;
+                        String moduleName = installedDriver.getModuleName();
                         org.jboss.jca.common.api.metadata.ds.Driver driver = new DriverImpl(installedDriver.getDriverName(),
                                 installedDriver.getMajorVersion(), installedDriver.getMinorVersion(),
                                 moduleName, installedDriver.getDriverClassName(),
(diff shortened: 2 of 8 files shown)


ropalka:
This PR is duplicate of https://github.com/wildfly/wildfly/pull/18560

ropalka (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
```diff
@@ -72,12 +73,24 @@ public void execute(final OperationContext context, final ModelNode operation) t
...
+                        String slot = "";
+                        if (canonicalModId != null && canonicalModId.contains(":")) {
+                            // Extract the slot by splitting the canonicalModId
+                            String[] parts = canonicalModId.split(":");
+                            if (parts.length > 1) {
+                                slot = parts[1];
+                            }
+                        }
+                        driverNode.get(DRIVER_MODULE_NAME.getName()).set(moduleName);
+                        driverNode.get(MODULE_SLOT.getName()).set(slot);
```
bstansberry, line 89: I made this same comment on #18560...

I was thinking of consolidating this kind of thing in ModuleIdentifierUtil:

https://github.com/wildfly/wildfly-core/pull/6282/files#diff-867089bef7659adb3ead7e82d208a38aa95d3fda6df7476187f2aa272bf064d4R26-R41

So this could just be another function to pass in.

RanabirChakraborty: @bstansberry should we wait for the PR to get merged?

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/AbstractDataSourceService.java
```diff
@@ -355,12 +355,11 @@ public CommonDeployment deploy(ServiceContainer serviceContainer) throws DeployE
...
                         }
                     }
-
                     String driverName = dataSourceConfig.getDriver();
                     InstalledDriver installedDriver = driverRegistry.getValue().getInstalledDriver(driverName);
                     if (installedDriver != null) {
-                        String moduleName = installedDriver.getModuleName() != null ? installedDriver.getModuleName().getName()
-                                : null;
+                        // Directly use installedDriver.getModuleName() because it's already a String
+                        String moduleName = installedDriver.getModuleName(); // No need for .getName() here
```
bstansberry, line 362: The "// No need for ..." comment is going to be confusing once this is merged, as it explains a change you're making, not the actual code that people will be reading. There is no String.getName() method.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
bstansberry: Yes, I think so.

The 'name' part can include ':' if it is escaped. "foo\:bar:baz" is valid.

bstansberry: https://github.com/wildfly/wildfly-core/pull/6293

RanabirChakraborty:
@bstansberry can you please review it now?

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
bstansberry: @RanabirChakraborty fyi that API is now available

RanabirChakraborty: Yes, I have used ModuleIdentifierUtil.parseModuleIdentifier method to ensure parsedSlot defaults to "main" if it is null. Let me know If I'm missing something here.

soul2zimate:
@RanabirChakraborty the test failure `JdbcDriverInfoTestCase` seems related.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
bstansberry: @RanabirChakraborty I think it's something more like this:

```
                        String moduleName = driver.getModuleName();
                        assert moduleName != null; // InstalledDriver.getModuleName() is documented as only returning null if isFromDeployment() is true
                        boolean canonicalize = false; // InstalledDriver.getModuleName() should only return canonical names or else the WFLY-20025 fix missed something
                        ModuleIdentifierUtil.parseModuleIdentifier(moduleName, (parsedName, parsedSlot) -> {
                            driverNode.get(DRIVER_MODULE_NAME.getName()).set(parsedName);
                            driverNode.get(MODULE_SLOT.getName()).set(parsedSlot);
                            return null;
                            }, canonicalize, "");
```

That's verbose because I was explaining via code. More concise is:

```
                        ModuleIdentifierUtil.parseModuleIdentifier(driver.getModuleName(), (parsedName, parsedSlot) -> {
                            driverNode.get(DRIVER_MODULE_NAME.getName()).set(parsedName);
                            driverNode.get(MODULE_SLOT.getName()).set(parsedSlot);
                            return null;
                        }, false, "");
```

connector/src/main/java/org/jboss/as/connector/services/driver/InstalledDriver.java
```diff
@@ -88,7 +86,7 @@ public InstalledDriver(final String deploymentUnitName, final String driverClass
      * @return the module name, or {@code null} if {@link #isFromDeployment()}
```
bstansberry, line 86: This should now be:

```
@return the canonical module name, or {@code null} if {@link #isFromDeployment()} returns {@code true}
```

And then this PR needs to be sure only canonical strings are passed to the constructor. (Which probably it does; I haven't double-checked as I write this.)

The constructor javadoc for the moduleName parameter should declare that the moduleName must be canonical.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/GetInstalledDriverOperationHandler.java
bstansberry: Note the 'driverNode.get(MODULE_SLOT.getName()).set(driver.getModuleName() != null ? driver.getModuleName().getSlot() : "");' null check in the existing code is unnecessary per the contract of InstalledDriver. And we know it's not needed in practice as the line before would NPE otherwise.

connector/src/main/java/org/jboss/as/connector/services/driver/InstalledDriver.java
RanabirChakraborty: Thanks! @bstansberry review it once.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/InstalledDriversListOperationHandler.java
```diff
@@ -71,14 +72,16 @@ public void execute(OperationContext context, ModelNode operation) throws Operat
...
                         } else {
                             driverNode.get(DEPLOYMENT_NAME.getName());
-                            driverNode.get(DRIVER_MODULE_NAME.getName()).set(driver.getModuleName().getName());
-                            driverNode.get(MODULE_SLOT.getName()).set(
-                                    driver.getModuleName() != null ? driver.getModuleName().getSlot() : "");
+                            ModuleIdentifierUtil.parseModuleIdentifier(driver.getModuleName(), (parsedName, parsedSlot) -> {
+                                driverNode.get(DRIVER_MODULE_NAME.getName()).set(parsedName);
+                                driverNode.get(MODULE_SLOT.getName()).set(parsedSlot);
+                                return null;
+                            }, false, "");
```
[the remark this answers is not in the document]
RanabirChakraborty: @bstansberry thanks! and looks like the tests are passing now. please have a look.

bstansberry:
Thanks @RanabirChakraborty and @ropalka

PR #18564, opened 2024-12-18
[WFLY-19954] A follow up to add tests on the vertx instance usage in opentelemetry subsystem

gaol:
This is a follow up to add tests on the vertx instance usage in opentelemetry subsystem

Related issue is: https://issues.redhat.com/browse/WFLY-19954

If a new WFLY issue is a better, I can create one to update.

### DIFF
--- a/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
+++ b/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.integration.observability.opentelemetry;
+
+import org.jboss.as.test.shared.ManagementServerSetupTask;
+
+/**
+ * A ServerSetupTask to set up logger categories to file handlers for log messages check in the tests.
+ *
+ * @author <a href="mailto:aoingl@gmail.com">Lin Gao</a>
+ */
+public class LoggingServerSetupTask extends ManagementServerSetupTask {
+
+    public LoggingServerSetupTask() {
+        this(SMALLRYE_OPENTELEMETRY_LOG_FILE, VERTX_FEATURE_PACK_LOG_FILE);
+    }
+
+    static final String SMALLRYE_OPENTELEMETRY_LOG_FILE = "smallrye-opentelemetry.log";
+    static final String VERTX_FEATURE_PACK_LOG_FILE = "vertx-feature-pack.log";
+
+    public LoggingServerSetupTask(String smallryeOpentelemetryLogFile, String vertxSubsystemLogFile) {
+        super(createContainerConfigurationBuilder()
+                .setupScript(createScriptBuilder()
+                        .startBatch()
+                        .add("/subsystem=logging/file-handler=otel-exporter:add(file={relative-to=jboss.server.log.dir, path=%s}, append=false, level=DEBUG)", smallryeOpentelemetryLogFile)
+                        .add("/subsystem=logging/logger=io.smallrye.opentelemetry.implementation.exporters:add(level=DEBUG, handlers=[otel-exporter])")
+                        .add("/subsystem=logging/file-handler=vertx-extension-logger-handler:add(file={relative-to=jboss.server.log.dir, path=%s}, append=false, level=DEBUG)", vertxSubsystemLogFile)
+                        .add("/subsystem=logging/logger=org.wildfly.extension.vertx:add(level=DEBUG, handlers=[vertx-extension-logger-handler])")
+                        .endBatch()
+                        .build())
+                .tearDownScript(createScriptBuilder()
+                        .startBatch()
+                        .add("/subsystem=logging/logger=org.wildfly.extension.vertx:remove()")
+                        .add("/subsystem=logging/file-handler=vertx-extension-logger-handler:remove()")
+                        .add("/subsystem=logging/logger=io.smallrye.opentelemetry.implementation.exporters:remove()")
+                        .add("/subsystem=logging/file-handler=otel-exporter:remove()")
+                        .endBatch()
+                        .build())
+                .build());
+    }
+
+}
--- a/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationWithVertxTestCase.java
+++ b/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationWithVertxTestCase.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.integration.observability.opentelemetry;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.testcontainers.api.DockerRequired;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.test.shared.observability.setuptasks.OpenTelemetryWithCollectorSetupTask;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+/**
+ * A test on opentelemetry to use the Vertx instance defined in the vertx subsystem.
+ * There will be a server log to indicate the usage.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({OpenTelemetryWithCollectorSetupTask.class, OpenTelemetryIntegrationWithVertxTestCase.LoggingWithVertxServerSetupTask.class, VertxSubsystemSetupTask.class})
+@RunAsClient
+@DockerRequired
+public class OpenTelemetryIntegrationWithVertxTestCase extends AbstractOpenTelemetryIntegrationTest {
+
+    private static final String WITH_VERTX_SMALLRYE_OPENTELEMETRY_LOG_FILE = "smallrye-opentelemetry-with-vertx.log";
+    private static final String WITH_VERTX_VERTX_FEATURE_PACK_LOG_FILE = "vertx-feature-pack-with-vertx.log";
+    public static class LoggingWithVertxServerSetupTask extends LoggingServerSetupTask {
+        public LoggingWithVertxServerSetupTask() {
+            super(WITH_VERTX_SMALLRYE_OPENTELEMETRY_LOG_FILE, WITH_VERTX_VERTX_FEATURE_PACK_LOG_FILE);
+        }
+    }
+
+    @Deployment(testable = false)
+    public static WebArchive getDeployment() {
+        return buildBaseArchive("otelinteg-with-vertx");
+    }
+
+    @Test
+    public void testVertxUsageInLog() throws Exception {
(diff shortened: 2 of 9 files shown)


testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+    @ArquillianResource
+    private URL url;
+
+    private static final String MP_CONFIG = "otel.sdk.disabled=false\n" +
+            "otel.metric.export.interval=100";
+
+    static WebArchive createDeployment(String deploymentName) {
+        return ShrinkWrap
+                .create(WebArchive.class, deploymentName + ".war")
+                .addPackage(AbstractOpenTelemetryIntegrationTest.class.getPackage())
```
rhusar, line 50: All the tests extending the AbstractOpenTelemetryIntegrationTest are run as client, so do not pollute the deployment with the test class itself.

rhusar (changes requested):

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
gaol: Thank you @rhusar , updated to only package the REST resources in the deployment.

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.preview.microprofile.telemetry;
```
bstansberry, line 5: Perhaps don't use package org.wildfly.test.preview.

Using a package with 'preview' in it was probably a mistake in testsuite/preview/basic.

The general goal of stuff in WFP is that at some point it gets promoted to standard WildFly. With the vertx subsystem that could be very quickly. Promoting means moving the tests out of testsuite/preview so I think it's better to use a package structure that fits wherever the tests will end up.

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.net.URL;
+import java.util.stream.Collectors;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OUTCOME;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUCCESS;
+import static org.jboss.as.test.shared.observability.setuptasks.ServiceNameSetupTask.SERVICE_NAME;
+
+public abstract class AbstractOpenTelemetryIntegrationTest {
```
bstansberry, line 33: @jasondlee Please review.

Review in general of course, but I put the comment here because soonish these classes will be promoted to standard WildFly and end up in testsuite/integration/microprofile. So now is the time to think whether this is a good name for a base class that will end up in that module, are the packages what you'll want etc. IOW let's plan for promotion.

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/LoggingServerSetupTask.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.OP_ADDR;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOVE;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+
+/**
+ * A ServerSetupTask to set up logger categories to file handlers for log messages check in the tests.
+ *
+ * @author <a href="mailto:aoingl@gmail.com">Lin Gao</a>
+ */
+public class LoggingServerSetupTask implements ServerSetupTask {
```
rhusar, line 27: Sorry but I don't see why this isn't just implementing `org.jboss.as.test.shared.ManagementServerSetupTask` instead which would significantly simplify this class and make it actually readable.

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
rhusar, line 33: I think it's the abstract class that should define `@DockerRequired` since it's the one defining a `@Testcontainer`. I believe these annotations inherited.

rhusar (changes requested):

gaol:
Thank you everyone for the reviews, I will update this pr this week. :)

testsuite/preview/expansion/src/test/java/org/wildfly/test/preview/microprofile/telemetry/AbstractOpenTelemetryIntegrationTest.java
jasondlee: @gaol There's already a base otel test class in ts/integ/mp. To make promoting these tests out of preview, I'd suggest, say, forking and extending that class. It looks like the only new functionality you're adding is the log retrieval, so if all else is the same, there will be much less to merge/migrate when that time comes.

gaol:
Updated this PR with:
* Added a separate commit to remove the `ts.copy-wildfly-standalone-embedded-broker` execution.
* Renamed the package to `org.wildfly.test.integration.observability.opentelemetry` to match the package name of tests in `testsuite/integration/microprofile`.
* Copied the abstract classes and REST resource classes from `testsuite/integration/microprofile` to make the tests easier to promote when the vertx feature pack gets promoted. The copied classes can just be removed at that time.
* Updated both `VertxSubsystemSetupTask` and `LoggingServerSetupTask` to extend `ManagementServerSetupTask` for easier set up and more readable.

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
```diff
@@ -0,0 +1,44 @@
...
+                        .add("/subsystem=logging/logger=org.wildfly.extension.vertx:remove()")
+                        .add("/subsystem=logging/file-handler=vertx-extension-logger-handler:remove()")
+                        .add("/subsystem=logging/logger=io.smallrye.opentelemetry.implementation.exporters:remove()")
+                        .add("/subsystem=logging/file-handler=otel-exporter:remove()")
+                        .endBatch()
+                        .build())
+                .build());
+    }
+
+}
```
jasondlee, line 44: This looks MUCH better. +1 :)

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationWithVertxTestCase.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+/**
+ * A test on opentelemetry to use the Vertx instance defined in the vertx subsystem.
+ * There will be a server log to indicate the usage.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({OpenTelemetryWithCollectorSetupTask.class, ServiceNameSetupTask.class, OpenTelemetryIntegrationWithVertxTestCase.LoggingWithVertxServerSetupTask.class, VertxSubsystemSetupTask.class})
```
jasondlee, line 25: If you're not explicitly checking for a particular service name in a trace, the `ServiceNameSetupTask` is not needed here.

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationWithoutVertxTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * A test on opentelemetry that is not using the Vertx instance defined in the vertx subsystem.
+ * There will be a server log to indicate a Vertx instance is created by smallerye-opentelemetry.
+ */
+@RunWith(Arquillian.class)
+@ServerSetup({OpenTelemetryWithCollectorSetupTask.class, ServiceNameSetupTask.class, LoggingServerSetupTask.class})
```
jasondlee, line 24: Same here.

jasondlee (approved):

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    static final String SMALLRYE_OPENTELEMETRY_LOG_FILE = "smallrye-opentelemetry.log";
+    static final String VERTX_FEATURE_PACK_LOG_FILE = "vertx-feature-pack.log";
+
+    public LoggingServerSetupTask(String smallryeOpentelemetryLogFile, String vertxSubsystemLogFile) {
+        super(createContainerConfigurationBuilder()
+                .setupScript(createScriptBuilder()
+                        .startBatch()
+                        .add("/subsystem=logging/file-handler=otel-exporter:add(file={relative-to=jboss.server.log.dir, path=" + smallryeOpentelemetryLogFile + "}, append=false, level=DEBUG)")
```
rhusar, line 27: FWIW the preferred way is to use the method variant with a pattern and %s expression with params, i.e. this one `org.jboss.as.test.shared.ManagementServerSetupTask.CommandSet#add(java.lang.String, java.lang.Object...)` but this works too...

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
rhusar: Indeed - though there is still room for improvement :)

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/LoggingServerSetupTask.java
gaol: I am glad to know that, thank you :)

testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/OpenTelemetryIntegrationWithVertxTestCase.java
gaol: OK, removed the usage of ServiceNameSetupTask, and use the name of the archive for the tests.

Actually checking for traces is not the main goal for this test case, it tries to send rest request to trigger the tracer so that the necessary messages to be checked will be in the log files.

gaol:
Minor changes were done to address the feedback above, thank you guys :+1:

bstansberry:
Thanks @gaol and @jasondlee

PR #18584, opened 2024-12-19
WFLY-20141 Multiple @Deployment-s in microprofile test suite are attaching unnecessary classes to deployments, e.g. test class itself

rhusar:
Eliminates "WELD-000119: Not generating any bean definitions from ... because of underlying class loading error: Type org.jboss.as.test.shared.observability.containers.OpenTelemetryCollectorContainer" logging

Resolve
https://issues.redhat.com/browse/WFLY-20141

### DIFF
--- a/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase.java
+++ b/testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase.java
@@ -19,6 +19,7 @@ import org.jboss.as.arquillian.api.ServerSetup;
 import org.jboss.as.arquillian.container.ManagementClient;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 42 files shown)


bstansberry:
Hi @rhusar Please reopen once nearly all the 35.x PRs are processed.

rhusar:
OK will do.

bstansberry:
@rhusar Things have calmed down so I reopened this.

jasondlee (approved):

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase.java
```diff
@@ -31,11 +32,11 @@ public class MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase {
...
 
-    @Deployment
+    @Deployment(testable = false)
     public static Archive<?> deployment() {
-        final Archive<?> deployment = ShrinkWrap.create(JavaArchive.class, "MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase.jar")
-                .addClasses(MicroProfileHealthSecuredHTTPEndpointSetupTask.class, EmptyMgmtUsersSetupTask.class);
-        return deployment;
+        // Dummy deployment to trigger @ServerSetup
+        return ShrinkWrap.create(JavaArchive.class, "MicroProfileHealthSecuredHTTPEndpointEmptyMgmtUsersTestCase.jar")
+                .addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
```
bstansberry, line 39: FYI re AssumeTestGroupUtil.emptyJar/War/Ear.

Those methods are in that class because they got added in the context of needing to deal with the fact that Arq would deploy things even though we would later disable the test when a @BeforeClass was called. So checking conditions and returning an empty artifact was a workaround.

That's not a problem any more because now WFARQ let's us stop everything via an AssumptionViolatedException from a ServerSetup.

bstansberry (approved):

bstansberry:
Thanks @rhusar and @jasondlee

PR #18588, opened 2024-12-20
[WFLY-20208] Upgrade Byte Buddy to 1.15.11

bstansberry:
https://issues.redhat.com/browse/WFLY-20208

A custom SE 24 job with this is at https://ci.wildfly.org/viewLog.html?buildId=477102

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -437,7 +437,7 @@
         <version.org.apache.wss4j>3.0.3</version.org.apache.wss4j>
         <version.org.apache.ws.xmlschema>2.3.0</version.org.apache.ws.xmlschema>
         <version.org.bitbucket.jose4j>0.9.6</version.org.bitbucket.jose4j>
-        <version.org.bytebuddy>1.14.18</version.org.bytebuddy>
+        <version.org.bytebuddy>1.15.11</version.org.bytebuddy>
         <version.org.codehaus.woodstox.stax2-api>4.2.2</version.org.codehaus.woodstox.stax2-api>
         <version.org.codehaus.woodstox.woodstox-core>7.0.0</version.org.codehaus.woodstox.woodstox-core>
         <version.org.cryptacular>1.2.5</version.org.cryptacular>


bstansberry:
@yrodiere If you're still around in 2024, WDYT about this?

Not super urgent.

At some point fairly soon we'll need to use this in combination with ORM 6.6.x as we'll need to support SE 24+ in our EE 10 implementation.

Unfortunately some test infrastructure (Mockito 5) now uses Byte Buddy as well so being behind a bit messes with nightly CI on SE 24 quite a bit. But we can live with that for a bit.

yrodiere:
Discussing:

* with the Hibernate team: https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/HHH-18972.20Upgrade.20bytebuddy.20in.206.2E6
* with the Quarkus team: https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/Bytebuddy.201.2E15

bstansberry:
@yrodiere Thanks for digging into this. Paul Ferraro was able to solve our issue with Mockito, which means we can now get useful CI results using SE 24 without this. So that removes the temporary urgency around this and how to deal with it in the ORM context is something we can take more time on.

@scottmarlow Do we not have CI tests that use byte buddy? I'd expect them to fail on https://ci.wildfly.org/viewType.html?buildTypeId=WF_MainNightlyJobs_StandardLinuxJdk24 but I don't see any such failures.

yrodiere:
> @yrodiere Thanks for digging into this. Paul Ferraro was able to solve our issue with Mockito, which means we can now get useful CI results using SE 24 without this. So that removes the temporary urgency around this and how to deal with it in the ORM context is something we can take more time on.

Thanks for the update. FWIW the upgrade doesn't seem to break any tests on the Quarkus side, but it does break a single test in the Hibernate ORM test suite. I'm investigating, but I suspect the test is more at fault than the upgrade. Will see.

> @scottmarlow Do we not have CI tests that use byte buddy? I'd expect them to fail on https://ci.wildfly.org/viewType.html?buildTypeId=WF_MainNightlyJobs_StandardLinuxJdk24 but I don't see any such failures.

FWIW as far as Hibernate ORM is concerned, setting the system property `net.bytebuddy.experimental=true` generally allows tests to pass even if the version of bytebuddy you're using wasn't designed for the bytecode version you're using.
Can't say if the same holds for Mockito, though.

scottmarlow:
> @scottmarlow Do we not have CI tests that use byte buddy? I'd expect them to fail on https://ci.wildfly.org/viewType.html?buildTypeId=WF_MainNightlyJobs_StandardLinuxJdk24 but I don't see any such failures.

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/datasourcedefinition/DataSourceDefinitionJPATestCase.java should use byte buddy.

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/resourceref/DataSourceResourceReferenceTestCase.java should also use byte buddy as well.  

Other CI tests (using Persistence) will also use byte buddy as well once we solve https://hibernate.atlassian.net/browse/HHH-18901 and enable Persistence bytecode enhancement by default.

yrodiere:
FYI [the upgrade to ByteBuddy 1.15.11 was merged into Hibernate ORM's 6.6 branch](https://github.com/hibernate/hibernate-orm/pull/9478). Tests passed. Unless we find last-minute problems with it, it will be part of Hibernate ORM 6.6.5, due to be released this Sunday.

ropalka (approved):

PR #18590, opened 2024-12-20
[WFLY-20209] AbstractParseAndMarshalModelsTestCase does not correctly…

yersan:
… detect if the test is running using the wildfly-ee distribution

Jira issue: https://issues.redhat.com/browse/WFLY-20209

Note: This is not very useful upstream, but it would help on the product testsuite.


bstansberry (approved):

bstansberry:
Thanks @yersan

PR #18593, opened 2024-12-20
[WFLY-19212] Do not include the security manager extension / subsystem in the default configuration for WildFly Preview.

darranl:
https://issues.redhat.com/browse/WFLY-19212

### DIFF
--- a/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
+++ b/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
@@ -256,6 +256,7 @@ public abstract class LayersTestBase {
             "org.wildfly.extension.metrics",
             // Extension not included in the default config
             "org.wildfly.extension.mvc-krazo",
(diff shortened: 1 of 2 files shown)


bstansberry:
LayersTestCase fails on WildFly Preview.

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -256,6 +256,7 @@ public abstract class LayersTestBase {
             "org.wildfly.extension.metrics",
             // Extension not included in the default config
             "org.wildfly.extension.mvc-krazo",
+            "org.wildfly.extension.security.manager",
```
bstansberry, line 259: I think this also needs to be in NO_LAYER_WILDFLY_PREVIEW to fix the LayersTestCase failure.

darranl:
Superseded by https://github.com/wildfly/wildfly/pull/18710

PR #18599, opened 2024-12-27
WFLY-20220 Purge global state if previous JVM was killed or crashed.

pferraro:
https://issues.redhat.com/browse/WFLY-20220


rhusar (approved):
LGTM (n.b. the default UncleanShutdownAction is to fail)

PR #18606, opened 2025-01-02
[WFLY-20228] Upgrade the WildFly Arquillian to 5.1.0.Beta8.

jamezp:
* https://issues.redhat.com/browse/WFLY-20228
* https://issues.redhat.com/browse/WFLY-20219

This fixes [WFLY-20219](https://issues.redhat.com/browse/WFLY-20219) by upgrading the `org.wildfly.plugins:wildfly-plugin-tools` in Arquillian to [1.2.1.Final](https://github.com/wildfly/wildfly-plugin-tools/releases/tag/v1.2.1.Final).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -321,7 +321,7 @@
              on the dependency on the MicroProfile REST Client TCK POM.
          -->
         <version.org.testng.mp.rest.client>7.10.2</version.org.testng.mp.rest.client>
-        <version.org.wildfly.arquillian>5.1.0.Beta7</version.org.wildfly.arquillian>
+        <version.org.wildfly.arquillian>5.1.0.Beta8</version.org.wildfly.arquillian>
         <version.org.wildfly.extras.creaper>2.0.3</version.org.wildfly.extras.creaper>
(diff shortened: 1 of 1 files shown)


bstansberry (approved):

bstansberry:
Thanks @jamezp 

I was concerned about the https://ci.wildfly.org/viewLog.html?buildId=479710&buildTypeId=WF_PullRequest_WindowsJdk17&fromSakuraUI=true failure too, but it passed on your re-run and on a re-run I did. It was suspicious though as it looked like the test told ARQ to stop a container and it then asserted state. It kind of looked like the state assertion might have happened before the container completely stopped, and only a couple seconds after the container began stopping, which doesn't seem like something that should happen and might indicate a problem with ARQ. But I don't see anything in the 5.1.0.Beta8 diff that could cause that.

@pferraro FYI.

jamezp:
Yes. I was concerned about that as well. I ran it a few times locally as well and it seemed to work. I'm not too sure what would have triggered that, but it's the first time the test failed.

bstansberry:
we continued this discussion at https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/WildFly.20ARQ.20upgrade

PR #18608, opened 2025-01-03
[WFLY-20225] & [WFLY-20231] Upgrade the jboss-parent POM from 39 to 47

jamezp:
This also requires a rename of the packages in the `messaging-activemq/injection` Maven module.

* https://issues.redhat.com/browse/WFLY-20225
* https://issues.redhat.com/browse/WFLY-20231

@ehsavoie let me know if you'd prefer to use a different package name.


jamezp:
The galleon failure looks related to this change. I'll do some testing with it tomorrow.

ehsavoie:
@jamezp I didn't like the new package name with the double injection so I did https://github.com/wildfly/wildfly/pull/18609 hoping to fix the galleon issue

PR #18611, opened 2025-01-03
[WFLY-20234] Suppress CVE-2024-52046 as mina-core is a test dependency only.

darranl:
https://issues.redhat.com/browse/WFLY-20234


darranl:
@bstansberry Not an issue if we merge later but would be nice to merge before we branch for 35.x so this known test dependency can be omitted from the scan results.

bstansberry:
Thanks @darranl

PR #18619, opened 2025-01-06
[WFLY-20239] Suppress CVEs against ActiveMQ Classic as we use Artemis ActiveMQ.

darranl:
https://issues.redhat.com/browse/WFLY-20239


darranl:
@bstansberry this would be nice to include before we branch for 35.x but it is not an issue for me to send in a second PR later.

ehsavoie (approved):

PR #18625, opened 2025-01-07
[WFLY-20242] Don't disable running execute-commands for all skipTests…

bstansberry:
… values; only for 'true' or no value

https://issues.redhat.com/browse/WFLY-20242


fjuma (approved):

bstansberry:
Thanks @fjuma

PR #18645, opened 2025-01-11
[WFLY-20194] Use ModuleDependency.Builder instead of deprecated ModuleDependency ctor in MICROMETER

ropalka:
https://issues.redhat.com/browse/WFLY-20194


bstansberry:
Thanks @ropalka

PR #18651, opened 2025-01-13
WFLY-20260 MP Telemetry TCK is never run with the security manager

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-20260

### DIFF
--- a/testsuite/integration/microprofile-tck/telemetry/src/test/java/io/smallrye/opentelemetry/ExceptionMapper.java
+++ b/testsuite/integration/microprofile-tck/telemetry/src/test/java/io/smallrye/opentelemetry/ExceptionMapper.java
@@ -9,7 +9,7 @@ import jakarta.ws.rs.core.Response.Status;
 import jakarta.ws.rs.ext.Provider;
 
 /**
- * Temporal fix to catch exceptions thrown in Jakarta RESTful Web Services endpoints, see https://issues.jboss.org/browse/RESTEASY-1758
+ * Temporary fix to catch exceptions thrown in Jakarta RESTful Web Services endpoints, see https://issues.redhat.com/browse/RESTEASY-1758
  *
  * @author Pavol Loffay
  */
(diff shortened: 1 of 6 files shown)


jasondlee (approved):

rhusar:
@jasondlee Made this as draft, because I found an improvement can be made that permissions could be granted a little more granularly, i.e. per test.

rhusar:
> @jasondlee Made this as draft, because I found an improvement can be made that permissions could be granted a little more granularly, i.e. per test.

Done - I reworked this to as granular as possible with minimum permissions per test case. Looks pretty good. Please, re-review.

jasondlee:
> Done - I reworked this to as granular as possible with minimum permissions per test case. Looks pretty good. Please, re-review.

It looks good, I guess, but I wonder about the long list of Strings that have to be maintained. Is the code complexity added by this very finely-grained access control worth the added cost in code size and maintenance?

rhusar:
> > Done - I reworked this to as granular as possible with minimum permissions per test case. Looks pretty good. Please, re-review.
> 
> It looks good, I guess, but I wonder about the long list of Strings that have to be maintained. Is the code complexity added by this very finely-grained access control worth the added cost in code size and maintenance?

Unfortunately yes, because granting permissions en masse defeats the purpose altogether. Here's a real example where this discovered a real issue: a test was doing a property lookup in a TCK but also the implementation is doing that lookup - but it was missing a privileged block. So, granting that permission to all deployments would hide that issue and never discover it. While this is annoying, the silver lining is that this will be short-lived until security manager support is dropped completely :-)

jasondlee:
> Unfortunately yes, because granting permissions en masse defeats the purpose altogether. Here's a real example where this discovered a real issue: a test was doing a property lookup in a TCK but also the implementation is doing that lookup - but the implementation was missing a privileged block. So, granting that permission to all deployments would hide that issue and never discover it. While this is annoying, the silver lining is that this will be short-lived until security manager support is dropped completely :-)

Fair enough! Approved.

jasondlee (approved):

bstansberry:
Thanks @rhusar and @jasondlee

PR #18655, opened 2025-01-13
WFLY-20257 MP JWT TCK is never run with the security manager

rhusar:
Plus a bit of cleanup.

Resolves
https://issues.redhat.com/browse/WFLY-20257


rhusar:
@darranl FYI, as discussed on Friday's call, did some investigations today. I was able to set this up very granularly, and there are only a few permissions that required by the TCK test classes only, e.g. [some tests just randomly get a class loader in a System.out logging](https://github.com/eclipse/microprofile-jwt-auth/blob/main/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/ProviderInjectionEndpoint.java#L295-L302)...

darranl (approved):

PR #18662, opened 2025-01-15
[WFLY-20282] Do not attempt to install Policy into JDK via byteman configuration

ropalka:
https://issues.redhat.com/browse/WFLY-20282

### DIFF
--- a/testsuite/integration/manualmode/pom.xml
+++ b/testsuite/integration/manualmode/pom.xml
@@ -33,6 +33,7 @@
         <!-- Byteman -->
         <byteman.host>127.0.0.1</byteman.host>
         <byteman.port>17091</byteman.port>
+        <byteman.policy>true</byteman.policy>
 
         <oidc.jboss.home>${project.build.directory}/stability-preview</oidc.jboss.home>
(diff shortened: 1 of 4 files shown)


bstansberry:
@ropalka  Thanks! 

If this setting can be false, why are we setting it to true? Is this because, on SE 24+, byteman deals behind the scenes with whatever 'true' would have allowed? Or was the 'true' just not needed?

If you don't know, please ping the test authors.

The change itself seems fine (unless removing the setting or making it static 'false' is better). If this dynamic fix is the needed one though, we should update the JIRA description with a bit more info so in the future we can understand, or add an explanation comment. It's obscure stuff. :)

ropalka:
You were right @bstansberry - this byteman's policy config option is not necessary at all for both tests to function properly.

bstansberry:
@ropalka it seems the CoreBridgeTimeout one fails when the SM is enabled with this latest iteration.

bstansberry (approved):
LGTM @ropalka. I added a comment to the JIRA showing what fails on SE < 24 when policy=true isn't used.

bstansberry:
Thanks @ropalka

PR #18665, opened 2025-01-15
WFLY-20148 - Clean up WildFlyRegistry handing in MicrometerSubsystemRegistrar

jasondlee:
https://issues.redhat.com/browse/WFLY-20148

- Introduce MicrometerService to encapsulate Micrometer concerns 
- Simplify dependencies from other classes

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
@@ -8,13 +8,14 @@ package org.wildfly.extension.micrometer;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBDEPLOYMENT;
 import static org.wildfly.extension.micrometer.MicrometerExtensionLogger.MICROMETER_LOGGER;
-import static org.wildfly.extension.micrometer.MicrometerSubsystemRegistrar.MICROMETER_COLLECTOR_RUNTIME_CAPABILITY;
+import static org.wildfly.extension.micrometer.MicrometerSubsystemRegistrar.MICROMETER_SERVICE_SERVICE_NAME;
 
 import java.util.function.Supplier;
 
 import io.micrometer.core.instrument.MeterRegistry;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.capability.CapabilityServiceSupport;
+import org.jboss.as.server.deployment.AttachmentKey;
 import org.jboss.as.server.deployment.Attachments;
 import org.jboss.as.server.deployment.DeploymentCompleteServiceProcessor;
 import org.jboss.as.server.deployment.DeploymentModelUtils;
@@ -25,38 +26,34 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.weld.WeldCapability;
 import org.wildfly.extension.micrometer.api.MicrometerCdiExtension;
 import org.wildfly.extension.micrometer.metrics.MetricRegistration;
-import org.wildfly.extension.micrometer.metrics.MicrometerCollector;
 import org.wildfly.subsystem.service.ServiceDependency;
 import org.wildfly.subsystem.service.ServiceInstaller;
 
 class MicrometerDeploymentProcessor implements DeploymentUnitProcessor {
     static final String WELD_CAPABILITY_NAME = "org.wildfly.weld";
+    static final AttachmentKey<MicrometerService> CONFIG_ATTACHMENT_KEY = AttachmentKey.create(MicrometerService.class);
 
-    private final MicrometerSubsystemRegistrar.MicrometerDeploymentConfiguration config;
-
-    MicrometerDeploymentProcessor(MicrometerSubsystemRegistrar.MicrometerDeploymentConfiguration config) {
-        this.config = config;
+    MicrometerDeploymentProcessor() {
     }
 
     @Override
-    public void deploy(DeploymentPhaseContext deploymentPhaseContext) throws DeploymentUnitProcessingException {
-        final DeploymentUnit deploymentUnit = deploymentPhaseContext.getDeploymentUnit();
+    public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitProcessingException {
+        final DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
 
-        ServiceDependency<MicrometerCollector> collector = ServiceDependency.on(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY.getCapabilityServiceName());
-        Supplier<MetricRegistration> factory = () -> collector.get().collectResourceMetrics(
+        ServiceDependency<MicrometerService> serviceDependency = ServiceDependency.on(MICROMETER_SERVICE_SERVICE_NAME);
+        Supplier<MetricRegistration> factory = () -> serviceDependency.get().collectResourceMetrics(
                 deploymentUnit.getAttachment(DeploymentModelUtils.DEPLOYMENT_RESOURCE),
                 deploymentUnit.getAttachment(DeploymentModelUtils.MUTABLE_REGISTRATION_ATTACHMENT),
-                createDeploymentAddressPrefix(deploymentUnit)::append,
-                this.config.getSubsystemFilter());
+                createDeploymentAddressPrefix(deploymentUnit)::append);
         ServiceInstaller.builder(factory)
                 .requires(ServiceDependency.on(DeploymentCompleteServiceProcessor.serviceName(deploymentUnit.getServiceName())))
-                .requires(collector)
+                .requires(serviceDependency)
                 .asActive()
                 .onStop(MetricRegistration::unregister)
                 .build()
-                .install(deploymentPhaseContext);
+                .install(phaseContext);
 
-        registerCdiExtension(deploymentPhaseContext);
+        registerCdiExtension(phaseContext, deploymentUnit.getAttachment(CONFIG_ATTACHMENT_KEY));
     }
 
     @Override
@@ -71,7 +68,7 @@ class MicrometerDeploymentProcessor implements DeploymentUnitProcessor {
         }
     }
 
-    private void registerCdiExtension(DeploymentPhaseContext deploymentPhaseContext) throws DeploymentUnitProcessingException {
+    private void registerCdiExtension(DeploymentPhaseContext deploymentPhaseContext, MicrometerService micrometerService) throws DeploymentUnitProcessingException {
         DeploymentUnit deploymentUnit = deploymentPhaseContext.getDeploymentUnit();
         try {
             CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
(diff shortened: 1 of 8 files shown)


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerService.java
```diff
@@ -0,0 +1,67 @@
+package org.wildfly.extension.micrometer;
```
bstansberry, line 1: header

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -137,65 +136,36 @@ public ManagementResourceRegistration register(SubsystemRegistration parent,
...
+                service.collectResourceMetrics(context.readResourceFromRoot(EMPTY_ADDRESS),
+                    context.getRootResourceRegistration(), Function.identity());
             }
         }, OperationContext.Stage.VERIFY);
 
-        return CapabilityServiceInstaller.builder(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY, collectorSupplier)
+        return CapabilityServiceInstaller.builder(MICROMETER_SERVICE_RUNTIME_CAPABILITY, serviceSupplier)
             .requires(List.of(mccf, executor, processStateNotifier))
             .withCaptor(captor::set) // capture the provided value
+            .onStart(() -> captor.get().start())
```
bstansberry, line 166: @pferraro The Installer.Builder javadoc needs to clarify the javadoc of withCaptor to specify the sequence of operations with respect to tasks provided to onStart/onStop.

What @jasondlee is doing here is quite magical and relies on undocumented behavior.

TBH if it works it feels broken, as I'd expect a service value to not be passed out to any consumer until the onStart work is completed. This relies on it happening before.

Jason, if we conclude that my sense is correct and that this being able to work is broken, you can stop relying on that broken behavior by passing captor into the MicrometerService c'tor and calling captor.set in start.

jasondlee: I thought I could implement Service and it would start automatically, but that doesn't appear to be the case, so I added the explicit onStart() so that the registry and collector are available for my VERIFY step. If that's not correct (or I'm working around a bug), I'm happy to make it right.

pferraro: This should be:
```java
.onStart(MicrometerService::start)
```

pferraro: @bstansberry withCaptor(...) and onStart(...) do not necessarily operate on the same object. 
 Generally, the service installer factory method employed here uses:
1. a factory
2. a mapper

onStart/onStop(...) operates on the object returned from factory, whereas the withCaptor(...) operates on the mapped value, i.e. the actual value provided by the service.  In this case, the mapper is the identity function (i.e. Function.identity()) thus onStart(...) and withCaptor(...) operate on the same object.

The general idea is that the factory produces some low-level implementation whose lifecycle is controlled by JBoss MSC, whereas the mapper wraps/downcasts this object using some high-level interface suitable for consumption by other services.

jasondlee: @pferraro @bstansberry To clarify, is there anything I need to change here, or am I using the API as intended/designed?

bstansberry: @jasondlee This looks ok now. It's no longer relying on 'accept(...)' on the Consumer passed to withCaptor being invoked *before* 'accept(...)' is called on the Consumer passed to onStart.

@pferraro Thanks for the explanation. So am I correct that theCaptor.accept(V value) won't be called until whatever was passed to onStart has been invoked? (It's possible the CI results on the early code I commented on will tell me, but TC is not responding at the moment so I can't look at the history.)

pferraro: @bstansberry
> So am I correct that theCaptor.accept(V value) won't be called until whatever was passed to onStart has been invoked?

Correct.  See: https://github.com/wildfly/wildfly-core/blob/main/service/src/main/java/org/wildfly/service/Installer.java#L442

pferraro: > @pferraro @bstansberry To clarify, is there anything I need to change here, or am I using the API as intended/designed?

The wildfly-service API yes, however, I would question your "MicrometerService" capability.  A RuntimeCapability is meant to expose some high-level API (either statically, or provided by a service) suitable for use by other subsystems, whereas "MicrometerService" is an implementation class that requires runtime dependencies specific to its implementation.

jasondlee: > The wildfly-service API yes, however, I would question your "MicrometerService" capability. A RuntimeCapability is meant to expose some high-level API (either statically, or provided by a service) suitable for use by other subsystems, whereas "MicrometerService" is an implementation class that requires runtime dependencies specific to its implementation.

@pferraro Do you have a recommendation on how that should be implemented instead?

bstansberry: @jasondlee I think something like remove MICROMETER_SERVICE_RUNTIME_CAPABILITY and ....

```
static final ServiceName MICROMETER_SERVICE_SERVICE_NAME = ....;
```

and then in configure()...

```
ServiceInstaller.builder(serviceSupplier)
   .provides(MICROMETER_SERVICE_SERVICE_NAME)
   .... the rest of what you have
```

I haven't tried that it; just seems like it would work from reading code.

The basic point is this service is an internal detail and doesn't need the capability aspect. If nothing else, capabilities, even private ones, should all be registered in https://github.com/wildfly/wildfly-capabilities and that's unneeded effort.

There's a small bit of black magic involved.

```
static final RuntimeCapability<Void> MICROMETER_SERVICE_RUNTIME_CAPABILITY =
        RuntimeCapability.Builder.of(MICROMETER_MODULE + ".service", MicrometerService.class)
            .addRequirements(ModelControllerClientFactory.SERVICE_DESCRIPTOR,
                Capabilities.MANAGEMENT_EXECUTOR,
                ProcessStateNotifier.SERVICE_DESCRIPTOR)
            .build();
```

Those 'addRequirements' tell Galleon that the thing that provides those must be installed. Which is a good thing in theory. But in practice those are provided by the very core of the server kernel, which will always be installed, so Galleon doesn't learn anything from that info, and it's safe to not provide it and just use ServiceName instead of RuntimeCapability.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -25,38 +25,33 @@
...
                 .requires(ServiceDependency.on(DeploymentCompleteServiceProcessor.serviceName(deploymentUnit.getServiceName())))
-                .requires(collector)
+                .requires(serviceDependency)
                 .asActive()
                 .onStop(MetricRegistration::unregister)
                 .build()
                 .install(deploymentPhaseContext);
 
-        registerCdiExtension(deploymentPhaseContext);
+        registerCdiExtension(deploymentPhaseContext, serviceDependency.get());
```
bstansberry, line 54: There's no happens-before relationship here. ServiceDependency.on by itself doesn't create one, so this method can run before MICROMETER_SERVICE_SERVICE_NAME) is started and get() can return null. Fairly unlikely, but it's possible, and CI is less likely to catch it because for most tests we have a booted server (so the service is started) and then we deploy. A typical real world use deploys the app as part of boot.

I suggest adding a hack like we did in MicroProfileTelemetryDependencyProcessor.deploy.

Note the TODO WFCORE-6491 in MicroProfileTelemetryDependencyProcessor points to the wrong JIRA; it should be WFCORE-6941

bstansberry (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -72,6 +64,7 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
             RuntimeCapability.Builder.of(MICROMETER_MODULE + ".wildfly-collector", MicrometerCollector.class)
                     .addRequirements(ModelControllerClientFactory.SERVICE_DESCRIPTOR, Capabilities.MANAGEMENT_EXECUTOR, ProcessStateNotifier.SERVICE_DESCRIPTOR)
                     .build();
+    static final ServiceName MICROMETER_SERVICE_SERVICE_NAME = ServiceName.of("micrometer");
```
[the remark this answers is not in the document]
jasondlee: Done.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerSubsystemRegistrar.java
```diff
@@ -109,26 +102,28 @@ class MicrometerSubsystemRegistrar implements SubsystemResourceDefinitionRegistr
...
-                    target.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, DEPENDENCIES, DEPENDENCIES_MICROMETER,
-                            new MicrometerDependencyProcessor());
-                    target.addDeploymentProcessor(MicrometerConfigurationConstants.NAME, POST_MODULE, POST_MODULE_MICROMETER,
-                            new MicrometerDeploymentProcessor(deploymentConfig.get()));
-                })
-                .withAddOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
-                .withRemoveOperationRestartFlag(OperationEntry.Flag.RESTART_ALL_SERVICES)
-                .build();
+            .withRuntimeHandler(ResourceOperationRuntimeHandler.configureService(this))
+            .addCapability(MICROMETER_COLLECTOR_RUNTIME_CAPABILITY)
```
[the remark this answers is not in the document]
jasondlee: Done (and the now-unreferenced Capability completed removed)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
jasondlee: @bstansberry I've made the changes, I think. Feel free to ask for more if I missed anything.

bstansberry (commented):
One small comment, but in general LGTM.

jasondlee:
> One small comment, but in general LGTM.

@bstansberry Did you forget to save the comment, or am I blind? :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -25,38 +26,34 @@
...
-                .requires(collector)
+                .requires(serviceDependency)
                 .asActive()
                 .onStop(MetricRegistration::unregister)
                 .build()
-                .install(deploymentPhaseContext);
+                .install(phaseContext);
 
-        registerCdiExtension(deploymentPhaseContext);
+        registerCdiExtension(phaseContext, serviceDependency.get());
```
bstansberry, line 56: The missing small comment:

s/serviceDependency.get()/deploymentUnit.getAttachment(CONFIG_ATTACHMENT_KEY)/g

The serviceDependency.get() will work because MicrometerDependencyProcessor has established a situation where this DUP won't even run until that service is started. But it's clearer for maintainers / less fragile if the code here explicitly gets the value via the directly-established mechanism, the attachment.

bstansberry:
Sorry @jasondlee -- I have no idea what happened to it. I probably didn't save it. I redid it.

jasondlee:
> Sorry @jasondlee -- I have no idea what happened to it. I probably didn't save it. I redid it.

No worries. I've done that before. I'll get that update as soon as I get a chance.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
jasondlee: Changed

bstansberry (approved):

bstansberry:
Thanks @jasondlee and @pferraro

PR #18666, opened 2025-01-15
[WFLY-20289] Remove use of ModuleIdentifier from ModuleIdentifierVali…

bstansberry:
…datorBuilder

https://issues.redhat.com/browse/WFLY-20289

This is also the last subtask for https://issues.redhat.com/browse/WFLY-19983, so when this is merged that parent task can be resolved as well.


ropalka (approved):

bstansberry:
Thanks @ropalka

PR #18657, opened 2025-01-13
[WFLY-20273] Adapt microprofile-certification.sh to play nicely with the MicroProfile Telemetry TCK processing

bstansberry:
https://issues.redhat.com/browse/WFLY-20273

This builds on #18656 because it touches the same file, although it doesn't really need to build on it as there'd be no conflicts.

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> Wildfly issue links:
> * [WFLY-20272](https://issues.redhat.com/browse/WFLY-20272)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #18668, opened 2025-01-16
[WFLY-20272][WFLY-20273] Adapt microprofile-certification.sh to reflect current status

bstansberry:
https://issues.redhat.com/browse/WFLY-20272
https://issues.redhat.com/browse/WFLY-20273

Upstream: https://github.com/wildfly/wildfly/pull/18657

@darranl This updates the script used to generate the MP 7.0 content in wildfly/certifications. It correct problems we hit when preparing those for 35.0.0.Final. So 35.0.1 will need this.


rhusar (approved):
LGTM

PR #18669, opened 2025-01-16
[WFLY-20277] Ignore persistence units in app-client container archive when deploying on server

scottmarlow:
https://issues.redhat.com/browse/WFLY-20277

This helps but doesn't solve the https://hibernate.atlassian.net/browse/HHH-18901 problem (`Seeing "java.lang.annotation.AnnotationFormatError: Duplicate annotation for class: interface org.hibernate.bytecode.enhance.spi.EnhancementInfo`).

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
@@ -61,11 +61,14 @@ public class PersistenceUnitParseProcessor implements DeploymentUnitProcessor {
 
     private static final String WEB_PERSISTENCE_XML = "WEB-INF/classes/META-INF/persistence.xml";
     private static final String META_INF_PERSISTENCE_XML = "META-INF/persistence.xml";
+    private static final String META_INF_APPLICATION_CLIENT_XML = "META-INF/application-client.xml";
     private static final String JAR_FILE_EXTENSION = ".jar";
     private static final String LIB_FOLDER = "lib";
 
-    public PersistenceUnitParseProcessor() {
+    private final boolean appclient;
 
+    public PersistenceUnitParseProcessor(boolean appclient) {
+        this.appclient = appclient;
     }
 
     @Override
@@ -89,6 +92,12 @@ public class PersistenceUnitParseProcessor implements DeploymentUnitProcessor {
             List<PersistenceUnitMetadataHolder> listPUHolders = new ArrayList<PersistenceUnitMetadataHolder>(1);
             // handle META-INF/persistence.xml
             final ResourceRoot deploymentRoot = deploymentUnit.getAttachment(Attachments.DEPLOYMENT_ROOT);
+            VirtualFile appclientClientXml = deploymentRoot.getRoot().getChild(META_INF_APPLICATION_CLIENT_XML);
+            if (!appclient && appclientClientXml.exists()) {
+                // if not in appclient container, do not deploy persistence units contained in appclient container archive
+                return;
+            }
+
             VirtualFile persistence_xml = deploymentRoot.getRoot().getChild(META_INF_PERSISTENCE_XML);
             parse(persistence_xml, listPUHolders, deploymentUnit);
             PersistenceUnitMetadataHolder holder = normalize(listPUHolders);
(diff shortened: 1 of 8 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/appclient/PersistenceUnitInAppClientArchiveInServerTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+        return ear;
+    }
+
+    /**
+     * Any deployment failures for this test means the fix for WFLY-20277 has been broken.
+     * This test is expected to simply pass meaning that the test deployment succeeded with the ignored invalid persistence provider.
+     */
+    @Test
+    public void testNothing() throws NamingException {
+        assertTrue("if we don't get a deployer failure than this test has passed", true);
```
bstansberry, line 62: I think better is to use @Deployment(managed = false) and add this to the test:

@ArquillianResource
private Deployer deployer;

Then use the deployer to deploy (and undeploy) inside the test method.

The way it is now if it fails it won't fail the test, it will fail at the class level. That messes with things like CI stats etc and might confuse maintainers.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
```diff
@@ -89,6 +92,12 @@ private void handleJarDeployment(DeploymentPhaseContext phaseContext) throws Dep
             List<PersistenceUnitMetadataHolder> listPUHolders = new ArrayList<PersistenceUnitMetadataHolder>(1);
             // handle META-INF/persistence.xml
             final ResourceRoot deploymentRoot = deploymentUnit.getAttachment(Attachments.DEPLOYMENT_ROOT);
+            VirtualFile appclientClientXml = deploymentRoot.getRoot().getChild(META_INF_APPLICATION_CLIENT_XML);
+            if (!appclient && appclientClientXml.exists()) {
```
bstansberry, line 96: I think this should check the DeploymentTypeMarker from Attachments.DEPLOYMENT_TYPE, as this is an incomplete check vs what ApplicationClientDeploymentProcessor does. And making it complete would be duplicative

If Attachments.DEPLOYMENT_TYPE isn't giving you what's needed we should investigate why not. AFAICT it should.

bstansberry: In zulip chat you mentioned it it wasn't set the way you expected, so I thought there might be an issue. Maybe I misunderstood. We'll see. :)

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/appclient/PersistenceUnitInAppClientArchiveInServerTestCase.java
scottmarlow: Thanks Brian!  Change is added for ^

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitParseProcessor.java
scottmarlow: Thanks Brian!  Change is added for ^

> In zulip chat you mentioned it it wasn't set the way you expected, so I thought there might be an issue. Maybe I misunderstood. We'll see. :)

I was read the words you said but I tried using a marker that the appclient container subsystem sets but that I likely used the wrong marker (probably one that isn't set in the server).

scottmarlow: Thanks Brian! Change is added for ^

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/appclient/PersistenceUnitInAppClientArchiveInServerTestCase.java
```diff
@@ -0,0 +1,80 @@
...
+    public void testNothing() throws NamingException {
+        try {
+            deployer.deploy(ARCHIVE_NAME);
+        } catch (RuntimeException deploymentProblem) {
+            deploymentProblem.printStackTrace();
+            fail("The change for WFLY-20277 has been broken likely because the appclientcontainerarchive.jar contains an invalid persistence.xml (someone needs to verify that).  " +
+                    "Deployment failure message: " + deploymentProblem.getMessage());
+        }
+        finally {
+            deployer.undeploy(ARCHIVE_NAME);
```
bstansberry, line 76: I think this should be in the try block right after the deploy.

If the deploy call fails, this undeploy will fail in the finally block (because it's not deployed) and that failure will hide the real test failure. It's only if the deploy succeeds that undeploy is needed.

bstansberry: Sorry; probably better is to leave it after the try/catch but not in a finally. If it's in the try block an undeploy failure can't be distinguished from a deploy failure.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #18690, opened 2025-01-21
[WFLY-20297] Use the correct maven property when configuring the fat …

bstansberry:
…server to use

https://issues.redhat.com/browse/WFLY-20297


jbliznak (approved):

bstansberry:
Thanks @jbliznak

PR #18692, opened 2025-01-21
WFLY-20315 - Rename testsuite/integration/microprofile

jasondlee:
https://issues.redhat.com/browse/WFLY-20315

- Rename module
- Modify profiles to match rename


rhusar (commented):
> The name of the module is somewhat misleading as it's not intended for just MicroProfile tests. It should be renamed, then, to expansion to make its intended purpose clearer.

Well, when we set this up, it was indeed only meant to include microprofile integration tests above the scope testing of the TCK - until IIRC you added non-MP tests to this testsuite ;-)

Nevertheless, +1 it appears practical to rename and redefine the scope.

jasondlee:
> Well, when we set this up, it was indeed only meant to include microprofile integration tests above the scope testing of the TCK - until IIRC you added non-MP tests to this testsuite ;-)

That's fair, but they were put there as instructed, and this PR came about after repeated mentionings of needing to rename the module to better reflect its purpose. :)

bstansberry:
I just posted on wildfly-dev about this to give people a bi of warning.

jasondlee:
> I just posted on wildfly-dev about this to give people a bi of warning.

That's a good call. Thanks!

bstansberry:
Thanks @jasondlee and @rhusar

PR #18700, opened 2025-01-23

jamezp:

Note for the test changes. Some of the deprecated methods from Hibernate 6.0 have been removed. These changes only affect WildFly Preview and Hibernate 7.0+, see https://docs.jboss.org/hibernate/orm/7.0/migration-guide/migration-guide.html#cleanup.

The changes to setting the ID's was done because the `persist` method does not allow generated id's to, rightfully IMO, be set.


jamezp:
While this isn't against the EE11 branch as it seems a little out of date, I've kicked on an EE11 PR Job https://ci.wildfly.org/buildConfiguration/WF_PullRequestEE11_Aggregator/484362?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true.

bstansberry:
I removed the EE11 label because I'm using that for PRs against the EE11 branch.

Although I suppose we could repurpose it if activity is going to pick up.

PR #18703, opened 2025-01-24
WFLY-20324 and WFLY-20302

marcosgopen:
https://issues.redhat.com/browse/WFLY-20324 Upgrade Narayana to 7.2.0.Final

and 

https://issues.redhat.com/browse/WFLY-20302 ObjectStoreBrowser doesn't allow to register external OSBTypeHandler

### DIFF
--- a/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
+++ b/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
@@ -6,8 +6,15 @@
 package org.wildfly.extension.microprofile.lra.coordinator.service;
 
 import com.arjuna.ats.arjuna.recovery.RecoveryManager;
+import com.arjuna.ats.arjuna.tools.osb.mbean.OSBTypeHandler;
+import com.arjuna.ats.arjuna.tools.osb.mbean.ObjStoreBrowser;
+
+import io.narayana.lra.coordinator.domain.model.FailedLongRunningAction;
+import io.narayana.lra.coordinator.domain.model.LongRunningAction;
 import io.narayana.lra.coordinator.internal.Implementations;
 import io.narayana.lra.coordinator.internal.LRARecoveryModule;
+import io.narayana.lra.coordinator.tools.osb.mbean.LRAActionBean;
+
 import org.jboss.logging.Logger;
 import org.jboss.msc.Service;
 import org.jboss.msc.service.StartContext;
@@ -16,6 +23,7 @@ import org.jboss.msc.service.StopContext;
 import org.wildfly.extension.microprofile.lra.coordinator._private.MicroProfileLRACoordinatorLogger;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
+import java.io.File;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


tomjenkinson (approved):
Looks good to me - thank you!!

bstansberry:
@marcosgopen The CI failures are related.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
```diff
@@ -59,6 +77,15 @@ public synchronized void start(final StartContext context) throws StartException
         // installing LRA recovery types
         Implementations.install();
 
+        // Add ObjStore type handlers
+        objStoreBrowser = ObjStoreBrowser.getInstance();
+
+        for (String[] typeAndBean : LRA_OSB_TYPES) {
+            String typeName = typeAndBean[0].replaceAll("/", File.separator);
```
bstansberry, line 84: This fails on Windows. Probably because File.separator is `\'. See the String.replaceAll javadoc for more details.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
```diff
@@ -29,6 +37,16 @@ public class LRARecoveryService implements Service {
 
     private volatile LRARecoveryModule lraRecoveryModule;
 
+    private ObjStoreBrowser objStoreBrowser;
+
+    private static String[][] LRA_OSB_TYPES = {
```
bstansberry, line 42: This should be final.

pom.xml
```diff
@@ -483,7 +483,7 @@
         <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
         <version.org.jboss.metadata>16.0.0.Final</version.org.jboss.metadata>
         <version.org.jboss.mod_cluster>2.1.0.Final</version.org.jboss.mod_cluster>
-        <version.org.jboss.narayana>7.1.0.Final</version.org.jboss.narayana>
+        <version.org.jboss.narayana>7.2.0.Final</version.org.jboss.narayana>
```
bstansberry, line 486: com.arjuna.common.internal.util.ClassloadingUtility.loadAndInstantiateClass added a call to Class.getDeclaredConstructor which fails, at least in some situations, when the SecurityManager is enabled, leading to the test failures.

The correct fix is tricky as that looks to be a general utility class, so adding a privileged block inside it is likely wrong. More likely such blocks belong in whatever callers need it and are safe to add it.

bstansberry (changes requested):

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
marcosgopen: Thanks @bstansberry , I see. I replaced replaceAll with replace so the regex  is not used.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/service/LRARecoveryService.java
marcosgopen: Thanks @bstansberry. I've updated it.

pom.xml
marcosgopen: I see, thanks @bstansberry.  I wonder if that is a problem related to 7.2.0.Final release though. I will have a look at that and fix it for 7.0.3.Final release.

marcosgopen:
I am closing this PR because of the security manager failing test.

PR #18705, opened 2025-01-24
[WFLY-20325] - Fix the MicroProfile Health CDIExtension in order to reset disabled default procedures configuration on undeploy

fabiobrz:
Resolves https://issues.redhat.com/browse/WFLY-20325

### DIFF
--- a/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
+++ b/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
@@ -8,6 +8,7 @@ package org.wildfly.extension.microprofile.health;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import jakarta.json.Json;
 import jakarta.json.JsonArray;
@@ -27,7 +28,6 @@ public class MicroProfileHealthReporter {
     public static final String DOWN = "DOWN";
     public static final String UP = "UP";
     private final boolean globalDefaultProceduresDisabled;
-    private boolean deploymentDefaultProceduresDisabled = false;
     private final String defaultReadinessEmptyResponse;
     private final String defaultStartupEmptyResponse;
     private final Map<HealthCheck, ClassLoader> healthChecks = new HashMap<>();
@@ -39,18 +39,59 @@ public class MicroProfileHealthReporter {
     private final HealthCheck emptyDeploymentLivenessCheck;
     private final HealthCheck emptyDeploymentReadinessCheck;
     private final HealthCheck emptyDeploymentStartupCheck;
-    private boolean userChecksProcessed = false;
+    private final Map<String, DeploymentConfiguration> deploymentsConfiguration = new ConcurrentHashMap<>();
 
-    public void registerDeploymentDefaultProceduresConfiguration(final String deploymentName, final boolean deploymentDefaultProceduresDisabled) {
-        if (deploymentDefaultProceduresDisabled && !globalDefaultProceduresDisabled) {
-            // Deployment-level setting is overriding the global setting
-            MicroProfileHealthLogger.LOGGER.defaultProceduresDisabledByDeployment(deploymentName);
-        }
-        this.deploymentDefaultProceduresDisabled |= deploymentDefaultProceduresDisabled;
+    /**
+     * Evaluate whether any deployment is disabling default procedures server-wide, i.e. by setting
+     * {@code mp.health.disable-default-procedures=true}
+     * @return {@code true} if any deployment is disabling the default procedures, {@code false} otherwise
+     */
+    private boolean isAnyDeploymentDisablingDefaultProcedures() {
+        return deploymentsConfiguration.entrySet().stream().anyMatch(dc -> dc.getValue().getDisableDefaultProcedures());
     }
 
+    /**
+     * Evaluate whether any user defined checks have been processed, for any deployments.
+     * @return {@code true} if any deployment has being tracked, {@code false} otherwise.
+     */
+    private boolean userChecksProcessed() {
+        return !deploymentsConfiguration.isEmpty();
+    }
+
+    /**
+     * Whether default procedures should be added or not to the returned health checks.
+     * Default procedures will be added when they're <i>not</i> disabled globally,
+     * <b>and</b> they're <i>not</i> disabled by any deployments.
+     * @return {@code true} if default procedures should be added, {@code false} otherwise.
+     */
     private boolean defaultProceduresShouldBeAdded() {
-        return !globalDefaultProceduresDisabled && !deploymentDefaultProceduresDisabled;
+        return !globalDefaultProceduresDisabled && !isAnyDeploymentDisablingDefaultProcedures();
+    }
+
+    /**
+     * Track a deployment configuration.
+     * @param deployment The deployment name
+     * @param disableDefaultProcedures Flag that tells whether the given deployment is disabling default procedures,
+     *                                 see the {@code mp.health.disable-default-procedures} MicroProfile Config property.
+     */
+    public void addDeploymentConfiguration(final String deployment, final Boolean disableDefaultProcedures) {
+        if (disableDefaultProcedures) {
+            // Log: deployment-level setting is overriding the global setting
+            MicroProfileHealthLogger.LOGGER.addDefaultProceduresDisabledByDeployment(deployment);
+        }
+        deploymentsConfiguration.put(deployment, new DeploymentConfiguration(disableDefaultProcedures));
+    }
+
+    /**
+     * Remove a deployment configuration.
+     * @param deployment The deployment name
+     */
+    public void removeDeploymentConfiguration(final String deployment) {
+        if (deploymentsConfiguration.get(deployment).getDisableDefaultProcedures()) {
+            // Log: deployment-level setting is being removed
+            MicroProfileHealthLogger.LOGGER.removeDefaultProceduresDisabledByDeployment(deployment);
+        }
+        deploymentsConfiguration.remove(deployment);
     }
 
     private static class EmptyDeploymentCheckStatus implements HealthCheck {
@@ -119,7 +160,7 @@ public class MicroProfileHealthReporter {
         if (readinessChecks.size() == 0) {
             if (!addDefaultProcedures) {
                 return getHealth(serverChecks, readinessChecks,
-                    userChecksProcessed ? HealthCheckResponse.Status.UP :
+                    userChecksProcessed() ? HealthCheckResponse.Status.UP :
                         HealthCheckResponse.Status.valueOf(defaultReadinessEmptyResponse));
             } else {
                 serverChecks.put(emptyDeploymentReadinessCheck, Thread.currentThread().getContextClassLoader());
@@ -134,7 +175,7 @@ public class MicroProfileHealthReporter {
         if (startupChecks.size() == 0) {
             if (!defaultProceduresShouldBeAdded()) {
                 return getHealth(serverChecks, startupChecks,
-                    userChecksProcessed ? HealthCheckResponse.Status.UP :
+                    userChecksProcessed() ? HealthCheckResponse.Status.UP :
                         HealthCheckResponse.Status.valueOf(defaultStartupEmptyResponse));
             } else {
                 serverChecks = Collections.singletonMap(emptyDeploymentStartupCheck, Thread.currentThread().getContextClassLoader());
@@ -285,7 +326,20 @@ public class MicroProfileHealthReporter {
         startupChecks.remove(check);
     }
 
-    public void setUserChecksProcessed(boolean userChecksProcessed) {
-        this.userChecksProcessed = userChecksProcessed;
+    /**
+     * Holds a deployment MicroProfile Health related configuration.
+     * Currently, the only field is the flag that tells whether the deployment is disabling default
+     * procedures, but it can be expanded in the future.
+     */
+    private static final class DeploymentConfiguration {
+        final Boolean disableDefaultProcedures;
+
+        private DeploymentConfiguration(Boolean disableDefaultProcedures) {
+            this.disableDefaultProcedures = disableDefaultProcedures;
+        }
+
+        public Boolean getDisableDefaultProcedures() {
+            return disableDefaultProcedures;
+        }
     }
 }
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.test.integration.microprofile.health;
+
+import java.io.IOException;
+
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.arquillian.setup.ReloadServerSetupTask;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Validate behavior of MicroProfile Health integration with respect to default procedures, and the behavior when
+ * those are disabled by multiple deployments.
+ *
+ * @author <a href="mailto:fburzigo@redhat.com">Fabio Burzigotti</a>
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+@ServerSetup(ReloadServerSetupTask.class)
+public class MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest {
+
+    private static final String DEPLOYMENT_NAME = "DisabledProcedures1.war";
+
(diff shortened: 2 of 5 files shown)


testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @OperateOnDeployment(DEPLOYMENT_NAME)
+    public void testDisabledDefaultProceduresUndeployment() throws IOException {
+        // The deployment disables the server procedures and since the value of `empty-readiness-checks` is "UP" by default,
+        // then the overall status must be "UP", and contain 5 checks
+        final String httpEndpoint = "/health/ready";
+        final String healthURL = "http://" + managementClient.getMgmtAddress() + ":" + managementClient.getMgmtPort() + httpEndpoint;
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "empty-readiness-checks", 5);
+        deployer.deploy(DEPLOYMENT_NAME);
+        // The deployment disables the server procedures and since the value of `empty-readiness-checks` is "UP" by default,
+        // then the overall status must be "UP", and contain 0 checks
```
xstefank, line 59: This is a little confusing since you expect 1 check in the assertion, fix the comment please

xstefank (approved):
@fabiobrz LGTM but fix that one comment please

fabiobrz:
> @fabiobrz LGTM but fix that one comment please

Hi @xstefank, and thanks for the review!
I actually spotted (thanks to the MicroProfile TS) another issue with the proposed changes, which is due to the logic in MicroProfileHealthReporter, which must set the value only when it has not been set, and to the one in CDIExtension, which must call such method only if the property has been set for the current deployment.
The latest changes are passing both the WildFly testsuite locally, and the MicroProfile test suite, so I'd ask for another look, if you please...
Meantime, let's see ho CI checks go.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
fabiobrz: Fixed, thanks @xstefank !

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
```diff
@@ -40,17 +40,33 @@ public class MicroProfileHealthReporter {
...
+        return defaultProceduresDeactivatorReference.get() != null;
+    }
+
+    public void registerDeploymentDefaultProceduresConfiguration(final String deploymentName) {
+        // Only if not set already, so that the multiple deployments case is handled
+        if (!globalDefaultProceduresDisabled && !deploymentDefaultProceduresDisabled()) {
+            // Log: deployment-level setting is overriding the global setting
             MicroProfileHealthLogger.LOGGER.defaultProceduresDisabledByDeployment(deploymentName);
+            // set
+            this.defaultProceduresDeactivatorReference.compareAndSet(null, deploymentName);
```
xstefank, line 55: but what if multiple deployments disable default procedures and you undeploy the first one? Not sure if I'm understanding it correctly, but I think that would ignore disabling by the second deployment with this.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @OperateOnDeployment(DEPLOYMENT_NAME)
+    public void testDisabledDefaultProceduresUndeployment() throws IOException {
+        // The deployment disables the server procedures and since the value of `empty-readiness-checks` is "UP" by default,
+        // then the overall status must be "UP", and contain 5 checks
+        final String httpEndpoint = "/health/ready";
+        final String healthURL = "http://" + managementClient.getMgmtAddress() + ":" + managementClient.getMgmtPort() + httpEndpoint;
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "empty-readiness-checks", 5);
+        deployer.deploy(DEPLOYMENT_NAME);
+        // The deployment disables the server procedures and since the value of `empty-readiness-checks` is "UP" by default,
+        // then the overall status must be "UP", and contain 1 checks
```
xstefank, line 59: 1 check (only please fix it if you will be pushing more)

fabiobrz: Oh well.. sorry about that. Let's see whether I'll have another chance (:innocent: ) to fix it.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
xstefank: So you assume only one application will be deployed, which is recommended but not required right? I think this might come back to bite later.

fabiobrz: Hi @xstefank - I would say that the implementation proposed here is not assuming that only one application will be deployed.
In fact the code is taking the multi-deployment scenario into account, but with the rules described above, i.e. only one application is responsible for disabling the server-wide procedures, and it is only by undeploying such application that such configuration is reset.

That being said I must admit that such behavior puts the user to a decision when deploying multiple applications, i.e. they will have to enforce a deployment/undeployment order to have the deterministic behavior described above.

Do you still see this as something that could bite us later, given the explanation above?

Thinking of an alternative, though, and this is what is coming to my mid:
1. we could maintain (on deployment/undeployment) a map to track all the applications that do define the property to disable the procedures, and verify whether the map is empty instead than checking an atomic reference for being not null.
2. live with the boolean switch. Each application being _deployed_ will switch it to `true` if it is not already and if such application defines the property. Then, each time an application is being _undeployed_, it will switch is back to `false` if it is not already, and if such application defines the property. 

IMO 1. is the only that covers the multi-deployment scenario, while 2. is similar to this PR proposal, with the difference that the reset would happen when undeploying _any_ of the applications that define the property.

WDYT?

xstefank: TBH, I think you should go with 1. This is what I would expect. I.e., if any deployment sets default procedures to false, then they are indeed disabled.

bstansberry: +1 to tracking the deployments.

Anything that relies on treating a particular deployment as special compared to others that also disabled default procedures is hard for users to understand and a likely source of user mistake or bugs in our code.

fabiobrz: Thanks @xstefank and @bstansberry for your feedback.
I pushed the changes again, taking the multi-deployment scenario into account, both on deploy and updeploy, also for what concerns the user defined checks registration, which affects how the (readiness|startup) checks are rendered, and added some Javadoc.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
fabiobrz: Fixed, thanks.

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,123 @@
...
+                .addAsWebInfResource(new StringAsset("<beans bean-discovery-mode=\"all\"></beans>"), "beans.xml");
+    }
+
+    /**
+     * Verify the server behavior with respect to the MicroProfile Health default procedures, before and after
+     * multiple applications are deployed.
+     *
+     * The test first verifies that all the checks are reported when nothing has been deployed.
+     *
+     * Then, the first applicaiton is deployed, which sets {@code mp.health.disable-default-procedures=true}. The test
```
xstefank, line 81:
```suggestion
     * Then, the first application is deployed, which sets {@code mp.health.disable-default-procedures=true}. The test
```

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,123 @@
...
+     * verifies that default procedures have been disabled, and that they no longer apper in the Health check
```
xstefank, line 82:
```suggestion
     * verifies that default procedures have been disabled, and that they no longer appear in the Health check
```

testsuite/integration/microprofile/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,123 @@
...
+        // server procedures, and since the value of `empty-readiness-checks` is "UP" by default,
+        // then the overall status must be "UP", and contain 1 user defined check
+        deployer.deploy(DEPLOYMENT_NAME);
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "myReadyProbe", 1);
+        // Let's deploy the second deployment as well, and we should see no changes, since both are disabling
+        // the default procedures via MP Config
+        deployer.deploy(ANOTHER_DEPLOYMENT_NAME);
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "myReadyProbe", 1);
+        // Now, let's undeploy the first deployment. Here default procedures are still expected to be disabled, because
+        // the remaining deployment defines mp.health.disable-default-procedures=true. Also, since it does not contains
```
xstefank, line 113:
```suggestion
        // the remaining deployment defines mp.health.disable-default-procedures=true. Also, since it does not contain
```

xstefank (approved):
@fabiobrz I assume you just changed the typos :)

fabiobrz:
Hi @bstansberry are the changes to track the multi-deployment scenario okay from your perspective?

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/health/MicroProfileHealthDisabledDefaultProceduresDeployUndeployTest.java
```diff
@@ -0,0 +1,123 @@
...
+        deployer.deploy(DEPLOYMENT_NAME);
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "myReadyProbe", 1);
+        // Let's deploy the second deployment as well, and we should see no changes, since both are disabling
+        // the default procedures via MP Config
+        deployer.deploy(ANOTHER_DEPLOYMENT_NAME);
+        MicroProfileHealthUtils.testHttpEndPoint(healthURL, true, "myReadyProbe", 1);
+        // Now, let's undeploy the first deployment. Here default procedures are still expected to be disabled, because
+        // the remaining deployment defines mp.health.disable-default-procedures=true. Also, since it does not contain
+        // any user defined check, we expect for 0 checks and an "UP" overall status
+        deployer.undeploy(DEPLOYMENT_NAME);
```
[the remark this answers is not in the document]
fabiobrz: Hi @jmesnil, and thanks for your feedback. Yes the probe would be in the response payload, unless `mp.health.disable-default-procedures=true` is defined - since _that very one_ is a default procedure itself.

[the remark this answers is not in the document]
fabiobrz: Anyway @jmesnil, I'll check the flow again in detail, and will unlock the PR once done.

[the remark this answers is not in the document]
fabiobrz: Hi @jmesnil, and thanks again for your feedback. I am back after reviewing the use cases, the docs, and the code again.

So, according to `MicroProfileHealthReporter`, the test logic is correct, because default procedures are still expected to be _disabled_ by the 2nd deployment configuration, hence no default checks are expected to be seen in the response, see https://github.com/wildfly/wildfly/blob/main/microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java#L126

You're right about the documentation, though, since the sentence is wrong, and it is as well in main, although it's located in a previous paragraph. 

I am going to push one last change where I have it removed.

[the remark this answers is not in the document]
fabiobrz: This is done now, FYI @jmesnil.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
```diff
@@ -39,18 +41,79 @@ public class MicroProfileHealthReporter {
     private final HealthCheck emptyDeploymentLivenessCheck;
     private final HealthCheck emptyDeploymentReadinessCheck;
     private final HealthCheck emptyDeploymentStartupCheck;
-    private boolean userChecksProcessed = false;
+    private final List<String> deploymentUserChecksProcessed = new ArrayList<>();
```
bstansberry, line 44: This needs to use a concurrent collection or all reads and writes need to be in synchronized blocks.

But, I don't see why this field and the two methods that write to it are needed. The methods that call those two methods also call addDeploymentDefaultProceduresConfiguration / removeDeploymentDefaultProceduresConfiguration with the same deploymentName. So the (thread safe) defaultProceduresConfiguration field can be used by userChecksProcessed().

bstansberry:
@jmesnil Since you're looking at this, I'd like to leave final approval and merge to you. You can confirm any fix of the thread safety issue I pointed out in my last comment without needing me.

Tangent:  the subsystem docs about the 'empty' checks (read to understand the discussion about the test) are really confusing. But that's a pre-existing condition.

microprofile/health-smallrye/src/main/java/org/wildfly/extension/microprofile/health/MicroProfileHealthReporter.java
fabiobrz: Thanks for your feedback @bstansberry. 
I used additional add/remove methods backed by a different list, mainly for clarity and based on a Separation of Concern principle.

`userChecksProcessed` was originally meant to be set as soon as a deployment checks were processed.
Such setting can then be used by Health response generation methods, which take it into account to determine whether the server is in the pre-deployment phase or not, as described in the docs, e.g.:

```
...
This allows applications that does not provide readiness checks to still be able to inform cloud containers when they are ready to serve requests. Setting empty-readiness-checks-status to DOWN ensures that the server will not be ready until the application is deployed.
...
```

`addDeploymentDefaultProceduresConfiguration` / `removeDeploymentDefaultProceduresConfiguration` have been added instead to track how deployments are configured with respect to default procedures, and I didn't want to mix the two implementations and the requirements they address. 

Anyway I can definitely follow your advice, and fix the code to keep separate methods, both relying on `defaultProceduresConfiguration`.

fabiobrz:
> @jmesnil Since you're looking at this, I'd like to leave final approval and merge to you. You can confirm any fix of the thread safety issue I pointed out in my last comment without needing me.

Thanks @bstansberry. @jmesnil I've revisited the code following Brian's suggestions, i.e. using the `ConcurrentHashMap` instance that tracks the individual deployments configuration for `userChecksProcessed()`, too.
I caught the chance and made another tiny change, introducing a `record` to hold the configuration as the type for the Map _values_, so that it has more generic semantics. 

> Tangent: the subsystem docs about the 'empty' checks (read to understand the discussion about the test) are really confusing. But that's a pre-existing condition.

+1, I tried to improve it, and we can log a separate WFLY issue to work on the documentation side asynchronously if needed. WDYT?

bstansberry:
@fabiobrz Re the docs improvement, I don't think the docs added for this change require improving the existing docs in order to make sense, so a separate issue is appropriate.

Note that use of a record precludes any porting to EAP XP 5, which baselines on SE 11.

fabiobrz:
> @fabiobrz Re the docs improvement, I don't think the docs added for this change require improving the existing docs in order to make sense, so a separate issue is appropriate.

+1, thanks, I'll log another issue.

> Note that use of a record precludes any porting to EAP XP 5, which baselines on SE 11.

Thanks @bstansberry - I'll fix it. 
BTW we seem to have another one [around](https://github.com/wildfly/wildfly/blob/main/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/messaging/otel/ReactiveMessagingOtelAssertUtils.java#L138) but in the testsuite so maybe not such a big deal? And - BTW again - that is what inspired me :slightly_smiling_face:

bstansberry:
@fabiobrz Re using records I don't object to it; I was just pointing that out in case you were thinking about porting this. My apologies for being unclear.

fabiobrz:
Thanks @bstansberry - if you don't have any objections, I'd be fine with this, now.

Better for me us save electrons and a full round of CI checks just to replace a 1-field POJO with a 1-field record.

Otherwise, if record it must be, I'll push again, and use the POJO for porting to XP5.

Feel free to let me know what you prefer.

bstansberry:
Thanks @fabiobrz, @xstefank and @jmesnil

fabiobrz:
> @fabiobrz Re the docs improvement, I don't think the docs added for this change require improving the existing docs in order to make sense, so a separate issue is appropriate.

I've filed a Jira to track the community documentation updates, see https://issues.redhat.com/browse/WFLY-20463
FYI @bstansberry

PR #18706, opened 2025-01-24

jamezp:

[WFLY-20326] Upgrade Hibernate Search to work with Hibernate 7.0. This also required an upgrade of Apache Lucene.

https://issues.redhat.com/browse/WFLY-20326

This replaces #17934 with the exception is it does not add the new Jakarta Batch integration module.

Note for the test changes. Some of the deprecated methods from Hibernate 6.0 have been removed. These changes only affect WildFly Preview and Hibernate 7.0+, see https://docs.jboss.org/hibernate/orm/7.0/migration-guide/migration-guide.html#cleanup.

The changes to setting the ID's was done because the `persist` method does not allow generated id's to, rightfully IMO, be set.

There was also a bug found, [HHH-19072](https://hibernate.atlassian.net/browse/HHH-19072). which requires a temporary workaround of using the `hibernate.session_factory_jndi_name` property in a tests `persistence.xml`.

cc @marko-bekhta and @yrodiere

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate/main/module.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.hibernate.jipijapa-hibernate">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.wildfly:jipijapa-hibernate6}"/>
+    </resources>
+
+    <dependencies>
+        <module name="jakarta.annotation.api"/>
+        <module name="jakarta.enterprise.api"/>
+        <module name="jakarta.persistence.api"/>
+        <module name="jakarta.transaction.api"/>
+        <module name="jakarta.validation.api"/>
+        <module name="jakarta.xml.bind.api"/>
+
+        <module name="org.hibernate" services="import"/>
+        <module name="org.hibernate.commons-annotations"/>
+        <module name="org.infinispan.core" services="import"/>
+        <module name="org.infinispan.protostream"/>
+        <module name="org.infinispan.hibernate-cache" services="import"/>
+        <module name="org.jboss.as.jpa.spi"/>
+        <module name="io.smallrye.jandex"/>
+        <module name="org.jboss.logging"/>
+        <module name="org.jboss.vfs"/>
+        <module name="org.wildfly.clustering.marshalling.protostream"/>
+        <module name="org.wildfly.common"/>
+    </dependencies>
+</module>
--- a/jpa/hibernate7/pom.xml
+++ b/jpa/hibernate7/pom.xml
@@ -1,0 +1,161 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-jpa-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>36.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jipijapa-hibernate7</artifactId>
+
+    <name>WildFly: Jipijapa Hibernate 7 (JPA 3.2) integration</name>
+
+    <properties>
+        <!-- Use the standard WildFly dependency set -->
+        <dependency.management.pom.artifactId>wildfly-preview-ee-bom</dependency.management.pom.artifactId>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
+                <artifactId>wildfly-preview-test-bom</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <!-- Internal -->
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>jipijapa-spi</artifactId>
+        </dependency>
+
+        <!-- External -->
+
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-annotations</artifactId>
+            <!-- This is a compile-time dependency of this project, but is not needed at compile or runtime by other
+                  projects that depend on this project.-->
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-processor</artifactId>
+            <!-- This is a compile-time dependency of this project, but is not needed at compile or runtime by other
+                  projects that depend on this project.-->
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.hibernate.common</groupId>
+            <artifactId>hibernate-commons-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-scan-jandex</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-hibernate-cache-spi</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-commons</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-component-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.infinispan.protostream</groupId>
+            <artifactId>protostream</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.kohsuke.metainf-services</groupId>
+            <artifactId>metainf-services</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>jakarta.persistence</groupId>
+            <artifactId>jakarta.persistence-api</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss</groupId>
+            <artifactId>jboss-vfs</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>jakarta.transaction</groupId>
+            <artifactId>jakarta.transaction-api</artifactId>
+        </dependency>
+    
+        <dependency>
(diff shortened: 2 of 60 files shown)


jamezp:
This is the EE 11 PR run https://ci.wildfly.org/buildConfiguration/WF_PullRequestEE11_Aggregator/484523?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/jipijapa-hibernate/main/module.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.hibernate.jipijapa-hibernate">
```
jamezp, line 8: Note this was just a change to the module name from `org.hibernate.jipijapa-hibernate6`. The idea being in the future we just provision the module with which ever integration is needed and other modules don't need to change the dependency.

We could add an alias for the old module, but this is private and didn't seem like it would be needed.

jpa/hibernate7/pom.xml
jamezp, line 1: This whole module was copied from `jpa/hibernate6`. There was one change needed for `org.wildfly.persistence.jipijapa.hibernate7.HibernateArchiveScanner` where the `org.hibernate.archive.scan.spi.AbstractScannerImpl` package moved to a new artifact dependency.

Also note the tests were not copied over. They only test Hibernate specific things not found in this module and one of the tests, `ScannerTests`, doesn't even run because it uses JUnit 4. These could likely be removed from the `hibernate6` module, but that is out of scope for this PR.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/hibernate/models/hibernate-models/main/module.xml
```diff
@@ -0,0 +1,26 @@
...
+
+<module xmlns="urn:jboss:module:1.9" name="org.hibernate.models.hibernate-models">
+
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <resources>
+        <artifact name="${org.hibernate.models:hibernate-models}"/>
+        <artifact name="${org.hibernate.models:hibernate-models-jandex}"/>
```
jamezp, line 16: @yrodiere @marko-bekhta I wasn't 100% sure if we should be using the `hibernate-models-jandex` or not. WildFly does use Jandex, so it seemed logical. Please let me know if this is wrong.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+/**
+ * JipiJapa message range is 20200-20299
+ * note: keep duplicate messages in sync between different sub-projects that use the same messages
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ * @author Scott Marlow
+ */
+@MessageLogger(projectCode = "JIPIORMV6")
```
scottmarlow, line 27: Should be JIPIORMV7 as we already have JIPIORMV6

jpa/hibernate7/pom.xml
scottmarlow: I think the ScannerTests protect against future VFS changes breaking compatibility with Hibernate ORM which seems unlikely to happen but you never know.  Some years ago it took a lot of effort to get Hibernate ORM to work with VFS based URLs.  Perhaps those tests should just move to the WildFly testsuite.

jpa/hibernate7/src/main/resources/org.hibernate.cache.internal.proto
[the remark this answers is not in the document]
scottmarlow: +1 for keeping it as I think that is needed for the second level cache handling (@pferraro would know for sure).

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
jamezp: That's a great question. I wonder if we should remove the version. The idea behind this is that id's don't really change. However, I realize this is not a typical case.

scottmarlow:
I only see one test failure in the [Preview](https://ci.wildfly.org/buildConfiguration/WF_PullRequest_WildFlyPreviewLinuxJdk21/484515?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true) results:

```
org.jboss.arquillian.container.spi.client.container.DeploymentException: Cannot deploy HibernateSearchOutboxPollingTestCase.war: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.persistenceunit.\"HibernateSearchOutboxPollingTestCase.war#primary\"" => "java.lang.NoClassDefFoundError: org/hibernate/models/spi/AnnotationDescriptor
    Caused by: java.lang.NoClassDefFoundError: org/hibernate/models/spi/AnnotationDescriptor
    Caused by: java.lang.ClassNotFoundException: org.hibernate.models.spi.AnnotationDescriptor from [Module \"org.hibernate.search.mapper.orm.outboxpolling\" version 8.0.0.Alpha1 from local module loader @6b58b9e9 (finder: local module finder @f14a7d4 (roots: /opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/wildfly/modules,/opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/wildfly/modules/system/layers/base,/opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/modules))]"}}}}
org.jboss.arquillian.container.spi.client.container.DeploymentException:
Cannot deploy HibernateSearchOutboxPollingTestCase.war: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.persistenceunit.\"HibernateSearchOutboxPollingTestCase.war#primary\"" => "java.lang.NoClassDefFoundError: org/hibernate/models/spi/AnnotationDescriptor
    Caused by: java.lang.NoClassDefFoundError: org/hibernate/models/spi/AnnotationDescriptor
    Caused by: java.lang.ClassNotFoundException: org.hibernate.models.spi.AnnotationDescriptor from [Module \"org.hibernate.search.mapper.orm.outboxpolling\" version 8.0.0.Alpha1 from local module loader @6b58b9e9 (finder: local module finder @f14a7d4 (roots: /opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/wildfly/modules,/opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/wildfly/modules/system/layers/base,/opt/buildAgent/work/e34a6f994de9f7c6/testsuite/preview/basic/target/modules))]"}}}}
  at org.jboss.as.arquillian.container.ArchiveDeployer.deployInternal(ArchiveDeployer.java:174)
```

scottmarlow (approved):

jamezp:
Yeah. The failures seem related. I'm going to look at it now to hopefully have a clean run before the weekend :)

jamezp:
Triggered a new EE11 PR run https://ci.wildfly.org/buildConfiguration/WF_PullRequestEE11_Aggregator/484650?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/hibernate/models/hibernate-models/main/module.xml
yrodiere: I think this will result in this being executed:

https://github.com/hibernate/hibernate-models/blob/f66a5f10e7d42ebafe9853f384ba6201b1a0812e/hibernate-models-jandex/src/main/java/org/hibernate/models/jandex/spi/JandexBuildingContextProvider.java#L32-L36

Which, worst case, does nothing (if WildFly doesn't pass the Jandex index to Hibernate ORM), and best case uses a new Jandex-based model retrieval (better performance) or just breaks everything (we'll get feedback!).

So, +1 on doing this in WildFly Preview.

yrodiere (commented):
With the copy-pasting it's hard to know if anything changed in Jipijapa from just the diff -- can I assume nothing changed?

If so, LGTM. @marko-bekhta may want to have a look at the Search part though, in particular where hibernate-models is involved.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
yrodiere: Isn't the point of project code and message IDs that they should never change -- to allow translation in particular, but also (as a side effect) easier web search?

marko-bekhta (commented):
Looks good 😃

jpa/hibernate7/pom.xml
jamezp: The `ScannerTests` don't even run. I can get them to run, but it didn't look like it was doing much. If we want it copied over and setup to run, I can do that.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
jamezp: You shouldn't have matching id's really, however this is kind of a special case. IMO it would be best to drop the 6, but it might be too late for that.

> Isn't the point of project code and message IDs that they should never change -- to allow translation in particular, but also (as a side effect) easier web search?

Yes. The idea behind these is that they're easy to search so if they were translated into a language one can't read, it would be easy to find out what the error message is.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/hibernate/search/mapper/orm/outboxpolling/main/module.xml
```diff
@@ -19,6 +19,7 @@
         <module name="jakarta.xml.bind.api"/>
         <module name="org.jboss.logging" />
         <module name="org.hibernate" />
+        <module name="org.hibernate.models.hibernate-models" services="import"/>
```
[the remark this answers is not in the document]
jamezp: I'm not sure here. I'll have to leave that to your expertise. I added this to get some tests working, but it doesn't mean it's correct :)

jamezp:
> With the copy-pasting it's hard to know if anything changed in Jipijapa from just the diff -- can I assume nothing changed?
> 
> If so, LGTM. @marko-bekhta may want to have a look at the Search part though, in particular where hibernate-models is involved.

@yrodiere The jipijapa stuff was copied over as-is with one exception. I did change the package name and one type. `HibernateArchiveScanner` had to change the import for `org.hibernate.archive.scan.spi.AbstractScannerImpl` because it moved as part of https://hibernate.atlassian.net/browse/HHH-18633.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/hibernate/search/mapper/orm/outboxpolling/main/module.xml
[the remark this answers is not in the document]
marko-bekhta: well ... me neither 🙈 😄 what I can say is ... 

- `hibernate-search-mapper-orm`
- `hibernate-search-mapper-orm-outbox-polling`
- `hibernate-search-mapper-pojo-base`

^ these 3 are directly using some classes from hibernate models.  Then `hibernate-search-mapper-orm` depends on `hibernate-search-mapper-pojo-base` and `hibernate-search-mapper-orm-outbox-polling` depends on `hibernate-search-mapper-orm`. 
So in terms of WildFly modules if we have `hibernate-search-mapper-pojo-base` depending on hibernate models module and `hibernate-search-mapper-orm`/`hibernate-search-mapper-orm-outbox-polling` depend  on the `hibernate-search-mapper-pojo-base` WF module do they also need to directly depend on the hibernate models module? 😃 🙈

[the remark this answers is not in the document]
jamezp: What this would mean,  I think, is that one if not all of the `org.hibernate.search.mapper.*` modules would need to export the `org.hibernate.models.hibernate-models` dependency. I see that seems to be fairly common among these modules, however I don't think we should do that here. If we export the `org.hibernate.models.hibernate-models` then that dependency ends up on deployments too. I don't think we want that as deployments should not need the module.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
jamezp: @bstansberry Do you have a strong opinion on this? In some ways I think leave it and we just have to deal with the version. The other option is to simply break it now and remove the version.

bstansberry: I think use JIPIORMV7 here. And then either duplicate messages (which is what you have now) or add a different logger class elsewhere (spi module?) and put the common messages there under JIPIORM.

Which to do is up to @scottmarlow.

jamezp:
I've created an EE11 PR too now https://github.com/wildfly/wildfly/pull/18721. It will make it easier to retest instead of manually having to kick off runs.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/JpaLogger.java
scottmarlow: My preference is JIPIORMV7

scottmarlow (changes requested):
Pending changing logger to JIPIORMV7, I think the pr will be ready to merge.

scottmarlow (approved):
ah, we already switched to JIPIORMV7

bstansberry:
Thanks @jamezp, @scottmarlow, @yrodiere and @marko-bekhta

PR #18591, opened 2024-12-20
[WFLY-17478] The old security and picketlink subsystems should move under "legacy"

darranl:
https://issues.redhat.com/browse/WFLY-17478

PR #18709, opened 2025-01-24
[WFLY-17478] Move the legacy security modules under legacy.

darranl:
https://issues.redhat.com/browse/WFLY-17478

Supersedes https://github.com/wildfly/wildfly/pull/18591


bstansberry:
Thanks @darranl

PR #18721, opened 2025-01-27

jamezp:

[WFLY-20326] Upgrade Hibernate Search to work with Hibernate 7.0. This also required an upgrade of Apache Lucene.

https://issues.redhat.com/browse/WFLY-20326

Given we've need a couple changes and there have been CI failures, this is the EE11 version of #18706


jamezp:
I cannot reproduce this failure.

jamezp:
Okay, I can reproduce this, but I can reproduce it on main too. This looks like a build issue and is not an issue with this PR.

bstansberry:
#18706 is merged.

PR #18730, opened 2025-01-28
WFLY-20311 Fix memory leak when SFSB is created, but never invoked, and thus never scheduled for expiration.

pferraro:
https://issues.redhat.com/browse/WFLY-20311

SFSBs will now be scheduled for expiration following creation, instead of upon initial invocation.

N.B. This change requires changes to some test assumptions, where a SFSB with a @StatefulTimeout of 0 is assumed to be invokable once.
With this change, such a bean is considered expired immediately following creation, i.e. *before* the 1st invocation, unless created within an active transaction.


tadamski (approved):

bstansberry:

PR #18731, opened 2025-01-30
WFLY-20350 Provided jboss.dist ignored in testsuite/integration/smoke

OndrejKotek:
https://issues.redhat.com/browse/WFLY-20350

### DIFF
--- a/testsuite/integration/smoke/pom.xml
+++ b/testsuite/integration/smoke/pom.xml
@@ -475,83 +475,28 @@
                     </executions>
                 </plugin>
                 <plugin>
-                    <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-resources-plugin</artifactId>
-                    <version>${version.resources.plugin}</version>
-                    <executions combine.self="override">
+                    <groupId>org.wildfly.plugins</groupId>
+                    <artifactId>wildfly-maven-plugin</artifactId>
+                    <version>${version.org.wildfly.plugin}</version>
+                    <executions>
                         <execution>
-                            <id>ts.copy-wildfly</id>
-                            <phase>none</phase>
+                            <id>bootable-jar-provisioning</id>
+                            <goals>
+                                <goal>package</goal>
+                            </goals>
+                            <phase>${ts.bootable-jar-packaging.phase}</phase>
+                            <configuration>
+                                <skipDeployment>true</skipDeployment>
+                                <provisioning-file>target/glow-scan/microprofile.surefire/provisioning.xml</provisioning-file>
+                                <bootableJar>true</bootableJar>
+                                <bootableJarName>test-wildfly-cloud-profile.jar</bootableJarName>
+                            </configuration>
                         </execution>
                     </executions>
                 </plugin>
             </plugins>
         </pluginManagement>
         <plugins>
-            <plugin>
-                <groupId>org.wildfly.plugins</groupId>
-                <artifactId>wildfly-maven-plugin</artifactId>
-                <version>${version.org.wildfly.plugin}</version>
-                <configuration>
-                    <overwrite-provisioned-server>true</overwrite-provisioned-server>
-                    <record-provisioning-state>false</record-provisioning-state>
-                    <log-provisioning-time>${galleon.log.time}</log-provisioning-time>
-                    <offline-provisioning>${galleon.offline}</offline-provisioning>
-                    <galleon-options>
-                        <jboss-maven-dist/>
-                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
-                        <optional-packages>passive+</optional-packages>
-                    </galleon-options>
-                </configuration>
-                <executions>
-                    <execution>
-                        <id>server-provisioning</id>
-                        <goals>
-                            <goal>provision</goal>
-                        </goals>
-                        <phase>generate-resources</phase>
-                        <configuration>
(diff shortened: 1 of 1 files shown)


testsuite/integration/smoke/pom.xml
```diff
@@ -640,6 +640,45 @@
             </build>
         </profile>
 
+        <profile>
```
bstansberry, line 643: Can you instead add a copy global properties and use those in the top level build/plugins section. Then just change the value of the props in this profile instead of redeclaring all the plugin stuff.

OndrejKotek: I'm not sure I understand you well, but I redistributed and minimized the definitions.

darranl:
@OndrejKotek is this ready for another review?  Did your push cover the comments from @bstansberry ?

OndrejKotek:
@darranl Yes, thanks, this is ready for another review, I hope I covered the comments from @bstansberry.

testsuite/integration/smoke/pom.xml
```diff
@@ -640,48 +585,84 @@
...
+            <id>standalone.microprofile.profile</id>
+            <activation>
+                <property>
+                    <name>ts.standalone.microprofile</name>
+                </property>
+            </activation>
+            <properties>
+                <!-- We don't want to use the default jboss.dist for this module, which is provisioned locally -->
+                <jboss.dist>${jbossas.project.dir}/${wildfly.build.output.dir}</jboss.dist>
```
bstansberry, line 661: I think this now should be the default setting, i.e. what's in L33 of the pom. Then server-provisioning-profile switches it to what's currently in L33. This standalone.microprofile.profile doesn't need to deal with it.

testsuite/integration/smoke/pom.xml
```diff
@@ -640,48 +585,84 @@
...
+                <jboss.home>${jboss.dist}</jboss.home>
```
bstansberry, line 662: I don't think this was necessary in the existing code, or now, and can be removed. All it does is repeat what was declared in testsuite/pom.xml and never overridden anywhere.

testsuite/integration/smoke/pom.xml
```diff
@@ -640,48 +585,84 @@
...
                             <execution>
                                 <id>server-provisioning</id>
                                 <goals>
                                     <goal>provision</goal>
                                 </goals>
-                                <phase>none</phase>
-                            </execution>
-                        </executions>
-                    </plugin>
-                    <!-- Turn back on the standard copy based provisioning -->
```
bstansberry, line 673: Now this profile needs to turn off the copy based provisioning. Currently this branch results in 

```
 --- wildfly:5.1.0.Final:provision (server-provisioning) @ wildfly-ts-integ-smoke ---
[INFO] Provisioning server in /Users/bstansberry/dev/wildfly/wildfly/testsuite/integration/smoke/target/wildfly
[INFO] Resolving feature-packs
[INFO] Installing packages
[INFO] Resolving artifacts
[INFO] Generating configurations
[INFO] Delayed generation, waiting...
[INFO] WildFly configuration generation took 1.857 seconds
[INFO] Overall WildFly Galleon Plugin took 2.906 seconds
[INFO] Overall Galleon provisioning took 3.38 seconds
[INFO] 
[INFO] --- resources:3.3.1:resources (default-resources) @ wildfly-ts-integ-smoke ---
[INFO] skip non existing resourceDirectory /Users/bstansberry/dev/wildfly/wildfly/testsuite/integration/smoke/src/main/resources
[INFO] 
[INFO] --- enforcer:3.5.0:enforce (require-java17-build) @ wildfly-ts-integ-smoke ---
[INFO] Skipping Rule Enforcement.
[INFO] 
[INFO] --- checkstyle:3.6.0:checkstyle (check-style) @ wildfly-ts-integ-smoke ---
[INFO] Rendering content with org.apache.maven.skins:maven-fluido-skin:jar:2.0.0-M9 skin
[INFO] Starting audit...
Audit done.
[INFO] 
[INFO] --- build-helper:3.6.0:add-test-source (add-ee-test-source) @ wildfly-ts-integ-smoke ---
[INFO] Test Source directory: /Users/bstansberry/dev/wildfly/wildfly/testsuite/integration/smoke/src/test/ee10/java added.
[INFO] 
[INFO] --- resources:3.3.1:copy-resources (ts.copy-wildfly) @ wildfly-ts-integ-smoke ---
[INFO] Copying 37 resources from target/wildfly to target/wildfly
```

So the last bit overwrites the first bit.

bstansberry (changes requested):
This basic approach is ok, although there are some things that need changes as noted below.

Later I think I'll file an issue to make this always use ts.copy-wildfly unless a profile like ts.layers or ts.bootable that specifically needs Galleon provisioning is activated. IOW don't use jboss.dist to control whether or not Galleon provisioning is used instead of an existing dist; only use it to control the location of that dist. It looks like the testsuite/integration/elytron module would also need to be updated.

testsuite/integration/smoke/pom.xml
OndrejKotek: `jboss.home` property removed from the profile.

testsuite/integration/smoke/pom.xml
OndrejKotek: I tried to check this but not good enough :( Fixed now, thanks!

bstansberry:
Thanks @OndrejKotek

PR #18741, opened 2025-02-04
[WFLY-20091] Provide distributed deployment repository functionality

rachmatowicz:
This PR is does the following:
- adds a new resource to the distributable-ejb subsystem, a module availability registrar provider
- allows the ejb3 subsystem to install amn instance of the registrar to store module deployment information across the cluster
- integrates the registrar with the deployment repository to (1) populate its values and (2) move suspend resume awareness from the deployment repository to the module availability registrar
-  use the values of the module avaipability registrar to (1) generate module updates for  EJB client applications and (2) populate discovery requests for EJB/HTTP discovery

For more information, see: https://issues.redhat.com/browse/WFLY-20091

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/remote/ModuleAvailabilityRegistrarService.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/remote/ModuleAvailabilityRegistrarService.java
@@ -19,6 +19,7 @@ import org.wildfly.clustering.server.manager.Service;
 import org.wildfly.clustering.server.provider.ServiceProviderListener;
 import org.wildfly.clustering.server.provider.ServiceProviderRegistrar;
 import org.wildfly.clustering.server.provider.ServiceProviderRegistration;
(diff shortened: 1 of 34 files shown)


ejb3/src/main/java/org/jboss/as/ejb3/remote/ModuleAvailabilityRegistrarService.java
```diff
@@ -0,0 +1,357 @@
...
+
+    // service interface
+    @Override
+    public boolean isStarted() {
+        return started;
+    }
+
+    @Override
+    public void start() {
+        EjbLogger.ROOT_LOGGER.info("Starting ModuleAvailabilityRegistrarService");
```
rhusar, line 91: If this log message is not temporary, this will have to be i18n-ed if it's meant to be on the info level. Or move it to lower level.

rhusar (changes requested):

PR #18748, opened 2025-02-10
[WFLY-20381] Upgrade to Hibernate ORM 6.6.7.Final

scottmarlow:
https://issues.redhat.com/browse/WFLY-20381

Note that this also addresses https://issues.redhat.com/browse/WFLY-20310 (`Fix Persistence TCK failure in jpa/core/metamodelapi/attribute/Client#getJavaMember when bytecode enhancement is enabled by default`) as proven by a Jakarta EE 10 Platform TCK test run.  

Updated: this pull request includes the fix for [WFLY-19393](https://issues.redhat.com/browse/WFLY-19393) (`Persistence container bytecode enhancement must be enabled by default to ensure better performance`).

Update:  this pull request includes the fix for upgrading Byte Buddy [WFLY-20208](https://issues.redhat.com/browse/WFLY-20208) to version 1.15.11 (the same as Hibernate ORM 6.6.7 includes).

____

> Wildfly issue links:
> * [WFLY-20299](https://issues.redhat.com/browse/WFLY-20299)

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
@@ -7,7 +7,6 @@ package org.jboss.as.jpa.config;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 import jakarta.persistence.EntityManagerFactory;
 
@@ -227,8 +226,6 @@ public class Configuration {
     private static final String EE_DEFAULT_DATASOURCE = "java:comp/DefaultDataSource";
     // key = provider class name, value = module name
     private static final Map<String, String> providerClassToModuleName = new HashMap<String, String>();
-    private static final String HIBERNATE = "Hibernate";
-    private static final String HIBERNATE_ENHANCER = "hibernate.enhancer";
 
     static {
         // always choose the default hibernate version for the Hibernate provider class mapping
@@ -272,34 +269,7 @@ public class Configuration {
         if (pu.getProperties().containsKey(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER)) {
             return Boolean.parseBoolean(pu.getProperties().getProperty(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER));
         }
-        if (isHibernateProvider(pu.getPersistenceProviderClassName())) {
-
-            return isAnyHibernateEnhancerPropertySpecified(pu);
-        }
         return true;
-    }
-
-    /**
-     * Consider jboss.as.jpa.classtransformer to be set to true if any hibernate.enhancer properties are set to true.
-     *
-     * @param pu
-     * @return true if any hibernate.enhancer property is detected otherwise return false
-     */
-    private static boolean isAnyHibernateEnhancerPropertySpecified(PersistenceUnitMetadata pu) {
-
-        Set<String> set = pu.getProperties().stringPropertyNames();
-        for (String key : set) {
-            if (key.startsWith(HIBERNATE_ENHANCER) &&
-                    (true == Boolean.parseBoolean(pu.getProperties().getProperty(key)))) {
-                return true;
-            }
-        }
-        return false;
-    }
-
-    private static boolean isHibernateProvider(String provider) {
-        // If the persistence provider is not specified (null case), Hibernate ORM will be used as the persistence provider.
-        return provider == null || provider.contains(HIBERNATE);
     }
 
     // key = provider class name, value = adapter module name
@@ -440,5 +410,4 @@ public class Configuration {
         }
         return result;
     }
-
 }
--- a/pom.xml
+++ b/pom.xml
@@ -465,7 +465,7 @@
         <version.org.glassfish.soteria>3.0.3</version.org.glassfish.soteria>
         <version.org.glassfish.web.jakarta.servlet.jsp.jstl>3.0.1-jbossorg-1</version.org.glassfish.web.jakarta.servlet.jsp.jstl>
         <version.org.hibernate.commons.annotations>7.0.3.Final</version.org.hibernate.commons.annotations>
-        <version.org.hibernate>6.6.3.Final</version.org.hibernate>
+        <version.org.hibernate>6.6.7.Final</version.org.hibernate>
         <version.org.hibernate.search>7.2.2.Final</version.org.hibernate.search>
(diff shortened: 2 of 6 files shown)


bstansberry:
@scottmarlow Can you update the 'ORM controlled' components as well? Bytebuddy, H2 if it has changed, others I may not remember. https://github.com/wildfly/wildfly/pull/18588 was for Byte Buddy, but I'd prefer it comes in via this PR.

scottmarlow:
> @scottmarlow Can you update the 'ORM controlled' components as well? Bytebuddy, H2 if it has changed, others I may not remember. #18588 was for Byte Buddy, but I'd prefer it comes in via this PR.

Byte Buddy did upgrade to 1.15.11, so will include that change against WFLY-20208.

H2, com.fasterxml.classmate, ANTLR, Hibernate Commons are still the same.

ropalka (approved):

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernatePersistenceProviderAdaptor.java
```diff
@@ -38,6 +42,10 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
     private static final String NONE = SharedCacheMode.NONE.name();
     private static final String UNSPECIFIED = SharedCacheMode.UNSPECIFIED.name();
 
+    //Cache this: more efficient to reuse the same instance across all PUs rather than having Hibernate have to create
+    //a new instance during each bootstrap.
+    private static final BytecodeProvider BYTECODE_PROVIDER = BytecodeProviderInitiator.buildDefaultBytecodeProvider();
```
bstansberry, line 47: @Sanne @scottmarlow This concerns me a bit in the context of a WildFly reload, where this BytecodeProvider instance will now outlive all the WildFly jpa subsystem resources and MSC services that use it. The BytecodeProvider.resetCaches() method implies there may be something to worry about there, unless we know that, in the stop of most WildFly services that happens as part of a reload, the resetCaches method will be invoked on this instance.

Sanne: Good point, and it doesn't have to be a static field - I merely suggested that the instance should be reused as much as possible. If there's a better way to control its lifecycle that would be nice: ideally get rid of it when all JPA providers have been booted, but it would also be fine to just tie the lifecycle of one of these for each PU being booted - slightly less efficient but perhaps easier to manage.

It's safe to invoke the resetCaches as often as you want, even concurrently with the enhancer being used, so it can be invoked by various other hooks.

scottmarlow: +1 for updating it to be a per PU instance.

scottmarlow: well, per PU instance would be to just move it to a local parameter.

scottmarlow: Actually, BytecodeProviderInitiator.buildDefaultBytecodeProvider() looks to be at the Hibernate ORM classloader instead of the application deployment classloader to cache the instance:

```
	public static BytecodeProvider buildDefaultBytecodeProvider() {
		// Use BytecodeProvider's ClassLoader to ensure we can find the service
		return getBytecodeProvider( ServiceLoader.load(
				BytecodeProvider.class,
				BytecodeProvider.class.getClassLoader()
		) );
	}

```

scottmarlow: BytecodeProvider.resetCaches()  does get called pretty frequently (concurrently) at the end of each application class being transformed which does sound wrong.

Look at https://github.com/hibernate/hibernate-orm/blob/b87830af58e30941040e18530be6f8b333f18139/hibernate-core/src/main/java/org/hibernate/jpa/internal/enhance/EnhancingClassTransformerImpl.java#L66-L65 which has:

```
	public byte[] transform(
			ClassLoader loader,
			String className,
			Class<?> classBeingRedefined,
			ProtectionDomain protectionDomain,
			byte[] classfileBuffer)  throws TransformerException {

		//N.B. the "className" argument doesn't use the dot-format but the slashes for package separators.
		final String classNameDotFormat = className.replace( '/', '.' );
		if ( CorePrefixFilter.DEFAULT_INSTANCE.isCoreClassName( classNameDotFormat ) ) {
			//Take care to not transform certain types; this is both an optimisation (we can skip this unnecessary work)
			//and a safety precaution as we otherwise risk attempting to redefine classes which have already been loaded:
			//see https://hibernate.atlassian.net/browse/HHH-18108
			return null;
		}

		try {
			return getEnhancer( loader ).enhance( className, classfileBuffer );
		}
		catch (final Exception e) {
			throw new TransformerException( "Error performing enhancement of " + className, e );
		}
		finally {
			bytecodeProvider.resetCaches();
		}
	}
```

scottmarlow: From previous conversations though, I think we identified that https://github.com/hibernate/hibernate-orm/blob/b752ad96f3b047dc9adea0587f146a129b9ac622/hibernate-core/src/main/java/org/hibernate/bytecode/internal/bytebuddy/BytecodeProviderImpl.java#L74 doesn't have (per application deployment class) state. 

Lets confirm that.

scottmarlow: https://github.com/hibernate/hibernate-orm/blob/f68f6aae50b0f2fb371cceb65cd8812af76b734f/hibernate-core/src/main/java/org/hibernate/metamodel/internal/EntityRepresentationStrategyPojoStandard.java#L305 looks to be one path where application classes are added to the bytecodeProvider cache that is often reset so not really sure if there is reuse between threads of that cache.  From debugging the HHH-18901 reproducer the ByteBuddyState.proxyCache + ByteBuddyState.basicProxyCache are always empty.

bstansberry (approved):

bstansberry:
Thanks @scottmarlow @Sanne and @ropalka

PR #18754, opened 2025-02-12
[WFLY-20296] Add support for more OpenTelemetry Config options

jasondlee:
https://issues.redhat.com/browse/WFLY-20296

Introduce new schema version
Add new attributes/elements


observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryConfigurationConstants.java
```diff
@@ -46,16 +60,41 @@ private OpenTelemetryConfigurationConstants() {}
...
+    public static final String TEMPORALITY_LOWMEMORY = "lowmemory";
+    public static final String TEMPORALITY_CUMULATIVE = "cumulative";
+
+    public static final String HISTOGRAM_AGGREGATION = "histogram-aggregation";
+    public static final String AGGREGATION_EXPONENTIAL = "base2_exponential_bucket_histogram";
+    public static final String AGGREGATION_EXPLICIT = "explicit_bucket_histogram";
+
+    public static final String[] ALLOWED_HISTOGRAM_AGGREGATION = {AGGREGATION_EXPONENTIAL, AGGREGATION_EXPLICIT};
+    public static final String[] ALLOWED_EXEMPLAR_FILTERS = {EXEMPLAR_FILTER_ALWAYS_ON, EXEMPLAR_FILTER_ALWAYS_OFF, EXEMPLAR_FILTER_TRACE_BASED};
     public static final String[] ALLOWED_EXPORTERS = {EXPORTER_JAEGER, EXPORTER_OTLP};
```
TomasHofman, line 87: This is somewhat unrelated to this PR, but maybe an opportunity to look at it.

In the OpenTelemetrySubsystemRegistrar#validateExporter() method it says that "The 'jaeger' exporter is no longer supported. Please update the system to use otlp".

Is it OK to still list "jaeger" as a valid option? It is offered as option in the Web Console, and it is even possible to set it via Web Console and CLI. But then Wildfly fails to reload / reboot. Perhaps it should throw an error during the write-attribute operation already, to not allow the subsystem to enter into invalid state..?

I can open a Jira for this (and work on it too), if it feels like something that should be fixed, but not in this PR.

observability/opentelemetry-api/src/main/java/org/wildfly/extension/opentelemetry/api/WildFlyOpenTelemetryConfig.java
```diff
@@ -66,12 +88,11 @@ public Builder setServiceName(String serviceName) {
...
         public Builder setExporter(String exporter) {
-            if (!exporter.equals("otlp")) {
+            if (!exporter.equals(EXPORTER_OTLP)) {
                 throw new IllegalArgumentException("An unexpected exporter type was found: " + exporter);
             }
-            addValue(OTEL_TRACES_EXPORTER, exporter);
-            addValue(OTEL_LOGS_EXPORTER, exporter);
-            addValue(OTEL_METRICS_EXPORTER, exporter);
+
+            // Remove this?
```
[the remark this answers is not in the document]
jasondlee: Maybe? :P Currently, we don't support any additional exporters, so this is at best a leftover (at worst, it's future-engineering). I'll figure that out before we merge, though.

galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/io/vertx/core/main/module.xml
```diff
@@ -31,6 +31,7 @@
         <module name="io.netty.netty-transport-native-epoll"/>
         <module name="io.netty.netty-transport-native-kqueue"/>
         <module name="java.logging"/>
+        <module name="java.naming"/>
```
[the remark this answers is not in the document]
jasondlee: I don't recall the details, but there was an odd and unexpected failure due to CNFE without that module. I'll remove and test again, though, to be sure, as it certainly does seem unexpected. The failure surprised me, but perhaps other things have been changed that no longer require the module, so I'll verify.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryConfigurationConstants.java
jasondlee: I believe it's there for backwards compatibility, allowing older server configs to start and get a more meaningful failure, rather than just fall over due to, say, a parsing error. Does that sound accurate, @bstansberry ?

bstansberry: This is an interesting discussion but definitely should be handled elsewhere.

It's quite complicated and the answer is not to have the validator reject 'jaeger'.

Use cases:

1) Mixed domain. The 'jaeger' value may be valid for a host controller running previous version servers, so it cannot be rejected by the domain controller. Note this scenario also means the management api description (read-resource-description) means the 'jaeger' value is listed as valid. Because it is valid. (The r-r-d output doesn't try and discriminate between use cases.)

2) Migration. Failing boot of an admin-only server that is using a previously valid config leaves the user no option to correct their config except XML editing. (I mention admin-only here because running in admin-only is a kind of hint that config manipulation is what the user is doing. If the server isn't admin-only we have to assume the purpose of running it is to due normal request handling etc.)

Failing in Stage.MODEL when these conditions are not relevant is valid. The OperationContext provides any OSH with the data it needs to detect these cases. The #1 case being any "executing on a Host Controller" case; i.e. don't be trying to figure out that it's mixed domain. It's not possible for a subsystem to do that.

jasondlee:
> Before moving out of draft state, please separate all the boilerplate around creating new API/xsd versions into a separate commit, followed by a commit with that introduces whatever is actually different about the new version change. All management API changes must be done that way so the two pieces can be properly reviewed.

They were that way, then an overeager squash undid it all, but I'll fix that before we try to merge this.

observability/opentelemetry/src/main/java/org/wildfly/extension/opentelemetry/OpenTelemetryConfigurationConstants.java
bstansberry: Looking more it sounds like OpenTelemetrySubsystemRegistrar#validateExporter() is doing the right thing.

jasondlee: If memory serves, you wrote it. :P

PR #18757, opened 2025-02-13
[WFLY-20393] Ensure that only one bytecode transformer is added for each Persistence Unit (also includes [WFLY-20394])

scottmarlow:
https://issues.redhat.com/browse/WFLY-20393
https://issues.redhat.com/browse/WFLY-20394 also was added which covers `Ensure that bytecode enhancement only happens if all persistence units allow bytecode enhancement`
____

> Wildfly issue links:
> * [WFLY-20394](https://issues.redhat.com/browse/WFLY-20394)

### DIFF
--- a/jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
+++ b/jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
@@ -90,4 +90,6 @@ public interface PersistenceUnitMetadata extends PersistenceUnitInfo {
 
     List<ClassTransformer> getTransformers();
 
+    boolean needsJPADelegatingClassFileTransformer();
+
 }
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.persistence.SharedCacheMode;
 import jakarta.persistence.ValidationMode;
@@ -99,6 +100,8 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
     private volatile ClassLoader cachedTempClassLoader;
 
     private volatile Map<URL, Index> annotationIndex;
+
+    private final AtomicBoolean needsJPADelegatingClassFileTransformer = new AtomicBoolean(false);
 
     @Override
     public void setPersistenceUnitName(String name) {
@@ -377,6 +380,12 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
     @Override
     public List<ClassTransformer> getTransformers() {
         return transformers;
+    }
+
+    @Override
+    public boolean needsJPADelegatingClassFileTransformer() {
+        // WFLY-20393 Ensure that only one internal JPADelegatingClassFileTransformer bytecode transformer is configured for each Persistence Unit
+        return needsJPADelegatingClassFileTransformer.compareAndSet(false, true);
     }
 
     @Override
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
@@ -757,4 +757,15 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 75, value="Illegal to call this method from injected, managed EntityManager")
     IllegalStateException illegalCallOnCloseMethod();
 
+    /**
+     * Logs warning that a mix of bytecode enhanced + non-enhanced in this deployment was detected.
+     * Also see discussion on https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/ORM.20bytecode.20enhancement.20.2B.20WildFly if possible.
+     * Also see https://hibernate.atlassian.net/browse/HHH-19168 + https://hibernate.atlassian.net/browse/HHH-19169
+     *
+     * @param scopedPuNames comma separated list of fully scoped persistence unit names that are non-enhanced
+     */
+    @LogMessage(level = WARN)
+    @Message(id = 76, value = "Some persistence units are configured to be bytecode enhanced but persistence unit(s) { %s } are configured to disable bytecode enhancement.  Bytecode enhancement will be performed.  Consider making all of the persistence units use bytecode enhancement.")
+    void mixedEnhancedAndNotEnhanced(String scopedPuNames);
+
 }
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
@@ -5,6 +5,12 @@
 
 package org.jboss.as.jpa.processor;
 
+import static org.jboss.as.jpa.messages.JpaLogger.ROOT_LOGGER;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import org.jboss.as.jpa.config.Configuration;
 import org.jboss.as.jpa.config.PersistenceUnitMetadataHolder;
 import org.jboss.as.server.deployment.DeploymentPhaseContext;
@@ -40,22 +46,30 @@ public class JPAClassFileTransformerProcessor implements DeploymentUnitProcessor
         // during the call to CreateContainerEntityManagerFactory.
 
         DelegatingClassTransformer transformer = deploymentUnit.getAttachment(DelegatingClassTransformer.ATTACHMENT_KEY);
-        boolean appContainsPersistenceProviderJars = false;  // remove when we revert WFLY-10520
         if ( transformer != null) {
-
+            final List<PersistenceUnitMetadata> notEnhancedPersistenceUnits = new ArrayList<>();
+            int enhancedPersistenceUnitsCount = 0;
             for (ResourceRoot resourceRoot : DeploymentUtils.allResourceRoots(deploymentUnit)) {
                 PersistenceUnitMetadataHolder holder = resourceRoot.getAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS);
                 if (holder != null) {
                     for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
-                        if (Configuration.needClassFileTransformer(pu)) {
+                        if (!Configuration.needClassFileTransformer(pu)) {
+                            notEnhancedPersistenceUnits.add(pu);
+                        } else if ( pu.needsJPADelegatingClassFileTransformer()) {
+                            enhancedPersistenceUnitsCount++;
                             transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
                         }
                     }
                 }
-                if ( pu.needsJPADelegatingClassFileTransformer()) {
-                    transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
-                }
             }
-        }
+
+            if (notEnhancedPersistenceUnits.size() > 0 && enhancedPersistenceUnitsCount > 0 ) {
+                // WFLY-20394 Log warning if applications have a mix of some persistence units with "jboss.as.jpa.classtransformer" set to false and some set to true
+                String persistenceUnitsConfiguredNotEnhanced = notEnhancedPersistenceUnits.
+                        stream().map(PersistenceUnitMetadata::getScopedPersistenceUnitName).
+                        collect(Collectors.joining(", "));
+                ROOT_LOGGER.mixedEnhancedAndNotEnhanced(persistenceUnitsConfiguredNotEnhanced);
+            }
+       }
     }
 }
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
@@ -52,6 +52,9 @@ public class JPAClassFileTransformerProcessor implements DeploymentUnitProcessor
                         }
                     }
                 }
+                if ( pu.needsJPADelegatingClassFileTransformer()) {
+                    transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
+                }
             }
         }
     }


jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
```diff
@@ -90,4 +90,12 @@ public interface PersistenceUnitMetadata extends PersistenceUnitInfo {
 
     List<ClassTransformer> getTransformers();
 
+    Object getLockObject();
```
jamezp, line 93: We shouldn't use this. The `wasDelegatingTransformerAdded` and `setDelegatingTransformerAdded` should handle the locking and not require a consumer to know that.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
```diff
@@ -90,4 +90,12 @@ public interface PersistenceUnitMetadata extends PersistenceUnitInfo {
...
+
+    /**
+      * @return true if a DelegatingClassTransformer was already setup to transform application classes for this persistence unit.
+     */
+    boolean wasDelegatingTransformerAdded();
+
+    void setDelegatingTransformerAdded();
```
jamezp, line 100: IMO this should simply be a CAS. Something like:
```java
boolean compareAndSetDelegatingTransformerAdded(boolean expected, boolean value);
```

jamezp (changes requested):

jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
scottmarlow: Addressed in change just pushed.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceUnitMetadata.java
scottmarlow: Switched to a compare and set, thanks!  And just pushed the change.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -100,6 +101,7 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
 
     private volatile Map<URL, Index> annotationIndex;
 
+    private volatile AtomicBoolean coordinateDelegatingTransformerAdd = new AtomicBoolean(false);
```
jamezp, line 104: This should be final.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -381,11 +383,14 @@ public List<ClassTransformer> getTransformers() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
-        transformers.add(classTransformer);
-        if (ROOT_LOGGER.isTraceEnabled()) {
-            ROOT_LOGGER.tracef("added entity class transformer '%s' for '%s'",
-                    classTransformer.getClass().getName(),
-                    getScopedPersistenceUnitName());
+        // WFLY-20393 Ensure that only one bytecode transformer is added for each Persistence Unit
+        if (true == coordinateDelegatingTransformerAdd.compareAndSet(false, true)) {
```
jamezp, line 387: Not a big deal, but the `true ==` is redundant.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -381,11 +383,14 @@ public List<ClassTransformer> getTransformers() {
...
+            transformers.add(classTransformer);
+            if (ROOT_LOGGER.isTraceEnabled()) {
+                ROOT_LOGGER.tracef("added entity class transformer '%s' for '%s'",
+                        classTransformer.getClass().getName(),
+                        getScopedPersistenceUnitName());
+            }
```
jamezp, line 393: Actually, I don't think this is correct. The `PersistenceUnitMetadata.getTransformers()` is a list so we should allow more than one I would think.

scottmarlow: It looks right  as the Hibernate ORM/EclipseLink persistence provider is calling PersistenceUnitInfo#addTransformer(ClassTransformer classTransformer) to add a single `classTransformer` to our internal list of transformers.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: fixed

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
jamezp: That may be, but from the API it appears to allow for more than one so we're breaking the contract IMO.

I think "one time only" check needs to be on registering the `JPADelegatingClassFileTransformer` which is a different type of transformer. The latter being a JBoss Modules one, this one being a JPA one, e.g. `jakarta.persistence.spi.ClassTransformer`.

scottmarlow: The API only allows one transformer to be added at a time but many transformers can be added by calling `addTransformer` multiple times which is still supported by the current code unless you see a problem that I don't :).

From https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/spi/persistenceunitinfo#addTransformer(jakarta.persistence.spi.ClassTransformer):

```
void addTransformer​(ClassTransformer transformer)

Add a transformer supplied by the provider that will be called for every new class definition or class redefinition that gets loaded by the loader returned by the getClassLoader() method. The transformer has no effect on the result returned by the getNewTempClassLoader() method. Classes are only transformed once within the same classloading scope, regardless of how many persistence units they may be a part of.

Parameters:
    transformer - provider-supplied transformer that the container invokes at class-(re)definition time 
```

Here is the transformer list from a few lines above in PersistenceUnitMetadataImpl.java:
```
// transformers will be written to when the Jakarta Persistence persistence provider adds their transformer.
    // there should be very few calls to add transformers but potentially many calls to get the
    // transformer list (once per application class loaded).
    private final List<ClassTransformer> transformers = new CopyOnWriteArrayList<ClassTransformer>();
```

scottmarlow: Back to your first comment:

> Actually, I don't think this is correct. The PersistenceUnitMetadata.getTransformers() is a list so we should allow more than one I would think.

We cannot change the Persistence API which only allows one transformer to be added per `addTransformer` call but Hibernate call `addTransformer` multiple times,

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
```diff
@@ -38,21 +41,26 @@ private void setClassLoaderTransformer(DeploymentUnit deploymentUnit) {
...
+        if (transformer != null) {
+            final List<PersistenceUnitMetadata> persistentUnits = new ArrayList<>();
             for (ResourceRoot resourceRoot : DeploymentUtils.allResourceRoots(deploymentUnit)) {
                 PersistenceUnitMetadataHolder holder = resourceRoot.getAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS);
                 if (holder != null) {
                     for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
-                        if (Configuration.needClassFileTransformer(pu)) {
-                            transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
+                        if (!Configuration.needClassFileTransformer(pu)) {
+                            // WFLY-20394 Ensure that bytecode enhancement only happens if all persistence units allow bytecode enhancement
```
Sanne, line 52: This comment sounds misleading? What if the current PU allows enhancement, but another doesn't?

scottmarlow: For each application deployment, we are detecting a mix of enhanced + non-enhanced and will disable enhancement for all (EE container managed) persistence units in mixed case.

More concretely, if there are 160 persistence units and 159 have enhancement enabled but 1 doesn't, all 160 persistence units will not be enhanced.  We can revise this logic in the future as needed but IMO I think it matches what we agreed to in https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/ORM.20bytecode.20enhancement.20.2B.20WildFly

@Sanne does ^ make sense?

scottmarlow: I'm going to add a warning for the mixed case so users might be informed of it.

scottmarlow: I made a change to show a warning and correct the check to ensure there is actually a mix of enhanced + not enhanced persistence units.  

Example warning: 
`WARN  [org.jboss.as.jpa] (MSC service thread 1-6) WFLYJPA0076: Some persistence units are configured to be enhanced but persistence unit(s) { 2lcmixedenhancement.jar#second_PU, 2lcmixedenhancement.jar#third_PU } disabled enhancement, none of the persistence units will be enhanced for this application deployment
`

Note that we show the list of persistence units that might need to be changed to allow enhancement so that all of the persistence units will be enhanced.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
jamezp: That is what I'm saying :) We are changing the API for this. We're only allowing one to be added and we should not be doing that. In the `PersistenceUnitMetadataImple.addTransformer()` you're only allowing for one transformer to be added. Note this is a `jakarta.persistence.spi.ClassTransformer` _not_ a `org.jboss.modules.ClassTransformer`. We want to limit the number of times a `org.jboss.modules.ClassTransformer` is executed.

The issue is we're issuing the "has the transformer been added" check in the wrong place. It needs to happen in the `org.jboss.as.jpa.processor.JPAClassFileTransformerProcessor`. Because we may have multiple PU's, we need to have something like `PersistenceUnitMetadata.compareAndSetTransformersRegsiter(boolean, boolean)`.

scottmarlow: Thanks James, I'm looking again at this!

scottmarlow: Correction made!  :-)

jamezp (approved):

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -757,4 +757,14 @@ public interface JpaLogger extends BasicLogger {
...
     IllegalStateException illegalCallOnCloseMethod();
 
+    /**
+     * Logs warning detected a mix of bytecode enhanced + non-enhanced, will disable enhancement for all persistence units in this deployment.
+     * Also see discussion on https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/ORM.20bytecode.20enhancement.20.2B.20WildFly if possible.
+     *
+     * @param scopedPuName name of the persistence unit
+     */
+    @LogMessage(level = WARN)
+    @Message(id = 76, value = "Some persistence units are configured to be enhanced but persistence unit(s) { %s } disabled enhancement, none of the persistence units will be enhanced for this application deployment")
```
bstansberry, line 767: s/enhancement, none/enhancement; therefore none/g

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -757,4 +757,14 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 75, value="Illegal to call this method from injected, managed EntityManager")
     IllegalStateException illegalCallOnCloseMethod();
 
+    /**
+     * Logs warning detected a mix of bytecode enhanced + non-enhanced, will disable enhancement for all persistence units in this deployment.
+     * Also see discussion on https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/ORM.20bytecode.20enhancement.20.2B.20WildFly if possible.
+     *
+     * @param scopedPuName name of the persistence unit
```
bstansberry, line 764: I think the param should be scopedPuNames. So then

@param scopedPuNames comma-delimited names of the persistence units

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
```diff
@@ -38,21 +44,37 @@ private void setClassLoaderTransformer(DeploymentUnit deploymentUnit) {
...
-                            transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
+                        if (!Configuration.needClassFileTransformer(pu)) {
+                            notEnhancedPersistenceUnits.add(pu);
                         }
+                        persistentUnits.add(pu);
                     }
                 }
             }
+
+            if (notEnhancedPersistenceUnits.size() > 0 && persistentUnits.size() > 0 ) {
```
bstansberry, line 63: This guard will return true in the case where *no* PU has enhancement enabled. There shouldn't be a WARN in such a case.

It might be cleaner to rename 'persistentUnits' to 'enhancedPersistenceUnits' and use an if/else for L55-58. The if test here would then only be true if both sets have entries.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
scottmarlow: @Sanne I'm going to mark [this](https://github.com/wildfly/wildfly/pull/18757) feedback as resolved.  Please comment again if there are any concerns.

bstansberry (changes requested):
Beyond my comments below, this WARN behavior seems like a good thing to cover in an integration test. Testing for the presence of log statements is a bit fussy, but we have testsuite utilities that help.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
Sanne: > More concretely, if there are 160 persistence units and 159 have enhancement enabled but 1 doesn't, all 160 persistence units will not be enhanced.

Sorry I tried to re-read the thread, but I still don't understand the rationale for disabling enhancement. When it's required, it doesn't seem safe to me to not enhance them. Either we do enhance them all and log a warning about this inconsistency, or you throw an error requiring the inconsistency to be addressed.

Clearly the warning would be less disruptive for existing applications so I expect that is what we'd do here, but perhaps in the future it's worth making it stricter as the side-effects can be rather tricky.

Sanne:
I had added another comment, but realizing it's on a "resolved concern" it's not made visible by github, so calling it out here.
  - https://github.com/wildfly/wildfly/pull/18757#discussion_r1960570509

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
scottmarlow: Change made locally and will be included in next push.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
scottmarlow: Hmm, already updated locally to the following (single loop and using if/else), will look for the ^ problem in :

```
    private void setClassLoaderTransformer(DeploymentUnit deploymentUnit) {
        // (AS7-2233) each persistence unit can use a persistence provider, that might need
        // to use ClassTransformers.  Providers that need class transformers will add them
        // during the call to CreateContainerEntityManagerFactory.

        DelegatingClassTransformer transformer = deploymentUnit.getAttachment(DelegatingClassTransformer.ATTACHMENT_KEY);
        if ( transformer != null) {
            final List<PersistenceUnitMetadata> notEnhancedPersistenceUnits = new ArrayList<>();
            int enhancedPersistenceUnitsCount = 0;
            for (ResourceRoot resourceRoot : DeploymentUtils.allResourceRoots(deploymentUnit)) {
                PersistenceUnitMetadataHolder holder = resourceRoot.getAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS);
                if (holder != null) {
                    for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
                        if (!Configuration.needClassFileTransformer(pu)) {
                            notEnhancedPersistenceUnits.add(pu);
                        } else if ( pu.needsJPADelegatingClassFileTransformer()) {
                            enhancedPersistenceUnitsCount++;
                            transformer.addTransformer(new JPADelegatingClassFileTransformer(pu));
                        }
                    }
                }
            }

            if (notEnhancedPersistenceUnits.size() > 0 && enhancedPersistenceUnitsCount > 0 ) {
                // WFLY-20394 Log warning if applications have a mix of some persistence units with "jboss.as.jpa.classtransformer" set to false and some set to true
                String persistenceUnitsConfiguredNotEnhanced = notEnhancedPersistenceUnits.
                        stream().map(PersistenceUnitMetadata::getScopedPersistenceUnitName).
                        collect(Collectors.joining(", "));
                ROOT_LOGGER.mixedEnhancedAndNotEnhanced(persistenceUnitsConfiguredNotEnhanced);
            }
       }
    }
```

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
scottmarlow: Addressed as per discussion today.

scottmarlow:
> I had added another comment, but realizing it's on a "resolved concern" it's not made visible by github, so calling it out here.
> 
>     * [[WFLY-20393] Ensure that only one bytecode transformer is added for each Persistence Unit (also includes [WFLY-20394]) #18757 (comment)](https://github.com/wildfly/wildfly/pull/18757#discussion_r1960570509)

Thank you Sanne!  The linked concern was updated to be visible and should be now addressed by change to only show the warning.

scottmarlow:
> testsuite utilities

Is https://github.com/wildfly/wildfly/blob/main/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/NoAuditLogTestCase.java#L61 an example of reading the log that we should follow here?

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/JPAClassFileTransformerProcessor.java
```diff
@@ -40,19 +46,30 @@ private void setClassLoaderTransformer(DeploymentUnit deploymentUnit) {
...
+            int enhancedPersistenceUnitsCount = 0;
             for (ResourceRoot resourceRoot : DeploymentUtils.allResourceRoots(deploymentUnit)) {
                 PersistenceUnitMetadataHolder holder = resourceRoot.getAttachment(PersistenceUnitMetadataHolder.PERSISTENCE_UNITS);
                 if (holder != null) {
                     for (PersistenceUnitMetadata pu : holder.getPersistenceUnits()) {
-                        if (Configuration.needClassFileTransformer(pu)) {
+                        if (!Configuration.needClassFileTransformer(pu)) {
+                            notEnhancedPersistenceUnits.add(pu);
+                        } else if ( pu.needsJPADelegatingClassFileTransformer()) {
+                            enhancedPersistenceUnitsCount++;
```
jamezp, line 59: Not critical, but this could be a simple `boolean`.

bstansberry:
> > Beyond my comments below, this WARN behavior seems like a good thing to cover in an integration test. Testing for the presence of log statements is a bit fussy, but we have testsuite utilities that help.
> 
> Thanks @bstansberry, I'll look to use wildfly/testsuite /shared/src/main/java/org/jboss/as/test/shared/util/LoggingUtil.java

@scottmarlow See also org.jboss.as.test.shared.TestLogHandlerSetupTask. 

There's similar stuff in org.wildfly.core:wildfly-core-testsuite-shared too: org.jboss.as.test.shared.logging.LoggingUtil and org.jboss.as.test.shared.logging.TestLogHandlerSetupTask. But WF Core doesn't use Arquillian so those aren't adapted to the WildFly Arquillian APIs.

scottmarlow:
> > > Beyond my comments below, this WARN behavior seems like a good thing to cover in an integration test. Testing for the presence of log statements is a bit fussy, but we have testsuite utilities that help.
> > 
> > 
> > Thanks @bstansberry, I'll look to use wildfly/testsuite /shared/src/main/java/org/jboss/as/test/shared/util/LoggingUtil.java
> 
> @scottmarlow See also org.jboss.as.test.shared.TestLogHandlerSetupTask.
> 
> There's similar stuff in org.wildfly.core:wildfly-core-testsuite-shared too: org.jboss.as.test.shared.logging.LoggingUtil and org.jboss.as.test.shared.logging.TestLogHandlerSetupTask. But WF Core doesn't use Arquillian so those aren't adapted to the WildFly Arquillian APIs.

Actually, it might make sense to not log a warning but use TRACE logging.  Will think some more about this.

scottmarlow:
> Should we perhaps separate the component upgrades from the rest of this? If they work properly without the other changes, then there is value in getting them in and getting the bake from all the CI we run.

@bstansberry 
https://github.com/wildfly/wildfly/pull/18748 only contains the component upgrades now.

https://github.com/scottmarlow/wildfly/tree/WFLY-20393_addonetransformer_per_persistenceunit has the other TBD changes now that corresponds to this pr.  Will see which of the `WFLY-20393_addonetransformer_per_persistenceunit` branch changes we need.

PR #18761, opened 2025-02-18
[WFLY-19915] Make MEMORY_STORAGE_THRESHOLD configurable

xjusko:
https://issues.redhat.com/browse/WFLY-19915

### DIFF
--- a/undertow/src/main/resources/schema/wildfly-undertow_preview_14_0.xsd
+++ b/undertow/src/main/resources/schema/wildfly-undertow_preview_14_0.xsd
@@ -95,6 +95,7 @@
                 <xs:attribute name="secure" type="xs:boolean" default="false"/>
                 <xs:attribute name="rfc6265-cookie-validation" type="xs:boolean" default="false"/>
                 <xs:attribute name="allow-unescaped-characters-in-url" type="xs:boolean" default="false"/>
+                <xs:attribute name="memory-storage-threshold" type="xs:long" default="16384"/>
             </xs:extension>
         </xs:complexContent>
     </xs:complexType>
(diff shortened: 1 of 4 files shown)


xjusko:
The build will probably fail, as Undertow has not yet released the feature.

rhusar:
Converting to draft as this is not available for review at the moment.

fl4via:
This one can be picked up as soon as WildFly upgrades to 2.4.0.Final @xjusko :-)

undertow/src/main/resources/schema/wildfly-undertow_preview_14_0.xsd
```diff
@@ -95,6 +95,7 @@
                 <xs:attribute name="secure" type="xs:boolean" default="false"/>
                 <xs:attribute name="rfc6265-cookie-validation" type="xs:boolean" default="false"/>
                 <xs:attribute name="allow-unescaped-characters-in-url" type="xs:boolean" default="false"/>
+                <xs:attribute name="memory-storage-threshold" type="xs:long" default="16384"/>
```
bstansberry, line 98: This schema is released and can't be changed.

The following is general info. Your PR is the random lucky one where I write this down. (I have before but it needs repeating.)

How to deal with an updated xml schema and a possible management API version bump will depend on when this gets integrated and whether there is already a new (i.e. unreleased) schema and API version for that WildFly release.

There are a lot of people adding new features to the undertow subsystem these days and the people maintaining this subsystem are going to need to come up with some good coordination.

The basic way to do this is:

1) If there is a need for a new xsd or a management API bump, there must be a separate commit that does all the boilerplate to introduce that. That commit should introduce no new functionality, e.g. the schema should be identical to the one it is derived from except a different xmlns value. This makes it easy for folks familiar with the boilerplate to review the boilerplate and it also makes the next point possible.

2) If that commit is useful to several different features (e.g. more than one feature team is working on something that would use that new schema) then that #1 commit should be published in a separate branch and all the separate teams should build their work off that same branch.  This is the team coordination part. The undertow subsystem maintainers should self-organize this. I suggest using zulip as lots of people can help provide guidance.

3) Then the developers of the individual features have their own branches where they add commits that relate to their new feature. Those commits are also easy to review because they don't have all the boilerplate; they just focus on the new thing.

PR #18766, opened 2025-02-19
[WFLY-20266] Permissions for MP Reactive Messaging TCK to run with se…

kabir:
…curity manager

https://issues.redhat.com/browse/WFLY-20266


kabir:
Unrelated failures. I'm re-running the WildFly Preview job again since I thought it could be affected, but thinking about it it isn't related at all.

bstansberry:
Thanks @kabir

PR #18770, opened 2025-02-20
[WFLY-20400] Add the org.jboss.weld:weld-api to MicroProfile REST Cli…

jamezp:
…ent TCK.

https://issues.redhat.com/browse/WFLY-20400

This is a failure on Java 24 and 25 only. I"m not entirely sure how it passes on older JVM's, but this dependency seems to be needed.

### DIFF
--- a/testsuite/integration/microprofile-tck/rest-client/pom.xml
+++ b/testsuite/integration/microprofile-tck/rest-client/pom.xml
@@ -156,6 +156,11 @@
         <!-- End permission requirement -->
 
         <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.weld.se</groupId>
             <artifactId>weld-se-core</artifactId>
             <scope>test</scope>


jamezp:
A triggered JDK 25 run https://ci.wildfly.org/buildConfiguration/WF_MainNightlyJobs_MicroProfileLinuxJdk21/488619

bstansberry:
@jamezp Before doing this I think it would be good to better understand why this broke. I know @ropalka found the JDK commit that did it, but unfortunately it's a big one.

jamezp:
@bstansberry Agreed. It's the big security manager removal one. I'll do some debugging as it might be easier to see it that way.

jamezp:
@bstansberry Okay, I found the issue. See my [comment](https://issues.redhat.com/browse/WFLY-20400?focusedId=26641555&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26641555) on the JIRA. FWIW this is the correct fix as it is somewhat broken as-is. The dependency should be there.

Note that if this behavior change in the `ServiceLoader` is expected, then we will need a change in the RESTEasy MicroProfile as well. However, I don't want to just do that to fix this yet.

bstansberry:
Thanks @jamezp

PR #18776, opened 2025-02-24
Tie the TS build-demander modules to their associated aggregator modules

bstansberry:
Currently testsuite/pom.xml executes the build-demander modules before the aggregator modules, but the aggregator modules don't demand on them, so if they fail the tests try and run (and fail.)

The intent here is if the production code build fails for some reason, then the build-demander will skipped, and in turn the aggregator will be skipped, preventing execution of the testsuite modules.

Initially a draft to see what blows up from this.

### DIFF
--- a/testsuite/aggregator-base/pom.xml
+++ b/testsuite/aggregator-base/pom.xml
@@ -34,6 +34,15 @@
     <description>Controls execution of testsuite modules relevant to testing functionality
                  available via the wildfly-ee-galleon-pack feature pack</description>
 
+    <dependencies>
+        <dependency>
+            <groupId>${ee.maven.groupId}</groupId>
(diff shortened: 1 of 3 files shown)


bstansberry:
@tomjenkinson FYI, re the Narayana testing issue you mentioned.

I haven't tried this so I don't know if this will break the WF build in some way. Hence the lack of a JIRA and my using a draft PR here.

If CI passes here, perhaps you could check if it solves the Narayana issue?

bstansberry:
Closing.

1) This defeats a design goal of not requiring a full build in order to run the TS (although that may be solvable or not an issue)
2) Tom reports it doesn't work anyway; the TS modules still try and execute.

PR #18777, opened 2025-02-24
[WFLY-20420] Make the undertow extension dependency on org.wildfly.ev…

bstansberry:
…ent.logger non-optional

ls tells me this will only add ~ 15.5k bytes to an installation, while doing it allows a slimmed server to configure the console access log

https://issues.redhat.com/browse/WFLY-20420


jamezp (approved):

bstansberry:
Thanks @jamezp

PR #18784, opened 2025-02-27
WFLY-20436 Messaging test XML does not conform to its subsystem schema

pferraro:
https://issues.redhat.com/browse/WFLY-20436

Fixes element ordering that does not comply with the xs:sequence of the subsystem XSD.


ehsavoie (approved):

bstansberry:
Thanks @pferraro and @ehsavoie

PR #18787, opened 2025-03-02
WFLY-20440 Fix domain-standard.xml to comply with the messaging-activemq subsystem schema.

pferraro:
https://issues.redhat.com/browse/WFLY-20440

As seen in https://ci.wildfly.org/buildConfiguration/WildFlyCore_PullRequest_WildFlyCorePreviewIntegrationLinuxJdk21/490069

Elements of a <server/> are ordered, and this test XML does not comply with that order.

### DIFF
--- a/testsuite/domain/src/test/resources/domain-configs/domain-standard.xml
+++ b/testsuite/domain/src/test/resources/domain-configs/domain-standard.xml
@@ -300,8 +300,8 @@
             <subsystem xmlns="urn:jboss:domain:messaging-activemq:13.0">
                 <server name="default"
                         persistence-enabled="false">
-                    <journal file-size="1024" />
                     <security elytron-domain="ApplicationDomain"/>
+                    <journal file-size="1024" />
                     <security-setting name="#">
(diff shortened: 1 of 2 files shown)


testsuite/domain/src/test/resources/domain-configs/domain-standard.xml
```diff
@@ -300,8 +300,8 @@
             <subsystem xmlns="urn:jboss:domain:messaging-activemq:13.0">
                 <server name="default"
                         persistence-enabled="false">
-                    <journal file-size="1024" />
                     <security elytron-domain="ApplicationDomain"/>
+                    <journal file-size="1024" />
```
yersan, line 304: Could you apply the same fix L660 and also at domain-standard-ee.xml:589?

yersan (changes requested):

testsuite/domain/src/test/resources/domain-configs/domain-standard.xml
pferraro: Done

yersan (approved):

bstansberry:
Thanks @pferraro and @yersan

PR #18819, opened 2025-03-14
[WFLY-20475] JCA: workmanager elytron-enabled attribute should be wri…

tadamski:
…table

https://issues.redhat.com/browse/WFLY-20475

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaWorkManagerDefinition.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaWorkManagerDefinition.java
@@ -5,6 +5,7 @@
 
 package org.jboss.as.connector.subsystems.jca;
 
+import static org.jboss.as.connector.subsystems.jca.Constants.ELYTRON_BY_DEFAULT_VERSION;
 import static org.jboss.as.connector.subsystems.jca.Constants.ELYTRON_ENABLED_NAME;
 import static org.jboss.as.connector.subsystems.jca.Constants.ELYTRON_MANAGED_SECURITY;
 import static org.jboss.as.connector.subsystems.jca.Constants.WORKMANAGER;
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaWorkManagerDefinition.java
```diff
@@ -64,7 +65,18 @@ public void registerAttributes(ManagementResourceRegistration resourceRegistrati
         super.registerAttributes(resourceRegistration);
 
         resourceRegistration.registerReadOnlyAttribute(WmParameters.NAME.getAttribute(), ReadResourceNameOperationStepHandler.INSTANCE);
-        resourceRegistration.registerReadOnlyAttribute(WmParameters.ELYTRON_ENABLED.getAttribute(), null);
+        resourceRegistration.registerReadWriteAttribute(WmParameters.ELYTRON_ENABLED.getAttribute(), null, new AbstractWriteAttributeHandler<Void>() {
```
pferraro, line 68: Use ModelOnlyWriteAttributeHandler.INSTANCE.

pferraro (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/jca/JcaWorkManagerDefinition.java
tadamski: @pferraro fixed

bstansberry (commented):
This change is ok, but looking at this I see that the 'setDeprecated' setting in JcaDistributedWorkManagerDefinition.DWmParameter.ELYTRON_ENABLED isn't applied to this JcaWorkManagerDefinition.WmParameter.ELYTRON_ENABLED. The LocalDescriptions.properties has a '.deprecated' message, but that won't be used anywhere without the 'setDeprecated'. It's that message that describes how this attribute isn't used.

I think that needs to be added.

tadamski:
@bstansberry fixed

bstansberry (approved):

bstansberry:

PR #18799, opened 2025-03-10
[WFLY-20392] Upgrade to Hibernate ORM 7.0.0.Beta5

scottmarlow:
https://issues.redhat.com/browse/WFLY-20392

PR #18834, opened 2025-03-24
Pull request18799 ee11

bstansberry:
A variant of #18799 against the EE11 branch, to get the WFP-focused CI.


bstansberry:
To be continued when the other PR is ready.

PR #18842, opened 2025-03-27
[WFLY-20487] Remove Unused Jandex Index from Persistence Unit Metadata

xjusko:
https://issues.redhat.com/browse/WFLY-20487

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
@@ -10,7 +10,6 @@ import static org.jboss.as.jpa.messages.JpaLogger.ROOT_LOGGER;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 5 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -201,16 +199,6 @@ public URL getPersistenceUnitRootUrl() {
         return persistenceUnitRootUrl;
     }
 
-    @Override
-    public void setAnnotationIndex(Map<URL, Index> indexes) {
-        annotationIndex = indexes;
-    }
-
-    @Override
-    public Map<URL, Index> getAnnotationIndex() {
```
scottmarlow, line 210: I guess we could also remove the org.jipijapa.plugin.spi.PersistenceUnitMetadata.setAnnotationIndex + getAnnotationIndex methods as that is an SPI (for WildFly components) that can change if needed.  

Or we could avod the change by adding empty PersistenceUnitMetadata.setAnnotationIndex + getAnnotationIndex methods to the PersistenceUnitMetadata interface.  

Adding methods means we keep our technical debt longer.  Removing the methods is better but that changes the spi.  

I'm now leaning towards removing the technical debt by removing the PersistenceUnitMetadata.setAnnotationIndex + getAnnotationIndex methods.

@bstansberry do you have a preference?

scottmarlow (approved):

bstansberry:
Thanks @xjusko and @scottmarlow

PR #18850, opened 2025-04-01
WFLY-19694 Add prototype for the equivalent of having a PersistenceProvider.getClassTransformer() to allow entity enhancement and bootstrap of the persistence unit to occur in two phases

scottmarlow:
https://issues.redhat.com/browse/WFLY-19694 + implements provides a prototype of the Persistence 4.0 feature https://github.com/jakartaee/persistence/issues/650

Idea is to replace the internal WildFly support for bootstrapping Hibernate ORM via two different WildFly services which was previously identified as something we should replace (see https://issues.redhat.com/browse/WFLY-2982 more details regarding switching from two persistence unit services to one service).


jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -379,6 +379,10 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        if(classTransformer.getClass().getName().contains("org.hibernate")) {
+            ROOT_LOGGER.infof("ignoring entity class transformer from Hibernate ORM as we will instead use a WildFly prototype");
```
jamezp, line 383: This needs to either log a a different level, likely `DEBUG`, or it needs o be internationalized.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceProviderAdaptor.java
```diff
@@ -94,5 +94,7 @@ public interface PersistenceProviderAdaptor {
     Object beanManagerLifeCycle(BeanManager beanManager);
 
     void markPersistenceUnitAvailable(Object wrapperBeanManagerLifeCycle);
+
+    void addClassFileTransformer(PersistenceUnitMetadata pu);
```
jamezp, line 98: Given the default in most other adaptors is a no-op, this could just be a default no-op method:

```java
default void addClassFileTransformer(final PersistenceUnitMetadata pu) {
}
```

jamezp (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Agreed, that slipped through :)

PR #18852, opened 2025-04-02
[WFLY-20510] Remove the jboss.org link from the welcome page

mskacelik:
https://issues.redhat.com/browse/WFLY-20510

### DIFF
--- a/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/web/ListenerTestCase.java
+++ b/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/web/ListenerTestCase.java
@@ -199,7 +199,7 @@ public class ListenerTestCase extends ContainerResourceMgmtTestBase {
         // check that the connector is live
         String cURL = "http://" + url.getHost() + ":8181";
         String response = HttpRequest.get(cURL, 10, TimeUnit.SECONDS);
-        assertTrue("Invalid response: " + response, response.indexOf("JBoss") >= 0);
+        assertTrue("Invalid response: " + response, response.contains("Welcome"));
         removeListener(Listener.HTTP, 5000);
     }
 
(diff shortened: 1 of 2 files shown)


bstansberry:
Thanks @mskacelik.

Wow, it turns out we have a test that is asserting "JBoss" is on that page. That goes back to 2011 when WildFly was "JBoss AS", but all these years we've been WildFly it still works because of the jboss.org link!

For this to pass CI, you'll need to modify these two lines:

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/web/ListenerTestCase.java#L202

https://github.com/wildfly/wildfly/blob/main/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/web/ListenerTestCase.java#L218

In both use:

```
assertTrue("Invalid response: " + response, response.indexOf("Welcome") >= 0);
```

This test gets run against different servers, some of which use 'WildFly' on the welcome page, some not. But they all include "Welcome" so that's a safe string to assert.

mskacelik:
Wow, that is really interesting. Will push the new changes. Hopefully, the tests will pass now.

bstansberry (approved):

testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/web/ListenerTestCase.java
```diff
@@ -199,7 +199,7 @@ public void testHttpConnector() throws Exception {
         // check that the connector is live
         String cURL = "http://" + url.getHost() + ":8181";
         String response = HttpRequest.get(cURL, 10, TimeUnit.SECONDS);
-        assertTrue("Invalid response: " + response, response.indexOf("JBoss") >= 0);
+        assertTrue("Invalid response: " + response, response.indexOf("Welcome") >= 0);
```
rhusar, line 202: Since you are touching these lines, this is really from the '90, this should be `response.contains(...)`... just sayin'.

rhusar: 2x :)

rhusar (approved):
LGTM now 👍🏼

bstansberry:
Thanks @mskacelik!!! and @rhusar too :)

PR #18848, opened 2025-03-31
[WFLY-20521] Cannot use CDI with Jakarta Persistence + application defined datasources without explicitly disabling byte code enahacement

scottmarlow:
https://issues.redhat.com/browse/WFLY-20521

This reverts WFLY-19393 commit (2caa7d617b148be19a5f2c70830fc28dc592c3c2) to address https://issues.redhat.com/browse/WFLY-20521

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> WildFly issue links:
> * [WFLY-19393](https://issues.redhat.com/browse/WFLY-19393)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #18854, opened 2025-04-02
[WFLY-20521] Added a test to ensure CDI integration works with Jakart…

jamezp:
…a Persistence listeners and the default data source. There is one disabled test which can be enabled once a fix is done to enable CDI support when bytecode enhancement is enabled.

https://issues.redhat.com/browse/WFLY-20521

This is a follow for #18848 to ensure there is no future regression.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/injection/support/jpa/cdi/EnabledEnhancementManagedInjectionTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/injection/support/jpa/cdi/EnabledEnhancementManagedInjectionTestCase.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.ee.injection.support.jpa.cdi;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Ignore;
+import org.junit.runner.RunWith;
+
+/**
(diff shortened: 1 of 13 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/injection/support/jpa/cdi/EnabledEnhancementManagedInjectionTestCase.java
```diff
@@ -0,0 +1,31 @@
...
+ * Tests that an entity callbacks, e.g. {@link jakarta.persistence.PostPersist}, work with Jakarta Context and Dependency
+ * Injection when the {@code jboss.as.jpa.classtransformer} property is set to {@code true}.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@ApplicationScoped
+@RunWith(Arquillian.class)
+@Ignore("This test can be enabled once the default data source can be used. Currently the test methods will throw a " +
+        "NullPointerException due the jpa subsystem explicitly disabling CDI support.")
+public class EnabledEnhancementManagedInjectionTestCase extends AbstractManagedInjectionTestCase {
```
jamezp, line 25: We don't actually need this one as currently CDI is explicitly disabled if bytecode enhancement is enabled. However, I left it as a placeholder for when this might be fixed. If we feel more comfortable deleting this, then adding it later that is fine too.

jamezp:
The test failures are unrelated.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/injection/support/jpa/cdi/EnabledEnhancementManagedInjectionTestCase.java
scottmarlow: Sounds like this will be enabled to recreate the failure previously seen with the RESTEasy Contact example.  Hopefully I will be ready soon to see if this test passes with the design changes I have made to improve CDI/Persistence/app datasource/default datasource integration so users don't have to choose a subset of features to use.

scottmarlow (approved):

PR #18863, opened 2025-04-04
WFLY-19694 Add prototype for the equivalent of having a PersistenceProvider.getClassTransformer() to allow entity enhancement and bootstrap of the persistence unit to occur in two phases

scottmarlow:
https://issues.redhat.com/browse/WFLY-19694 + implements provides a prototype of the Persistence 4.0 feature https://github.com/jakartaee/persistence/issues/650

The idea is for WildFly to (very early in deployment) register the Hibernate ORM bytecode enhancement to be done for the application deployment.  This ensures that the bytecode enhancement can be done for the deployment.

This change also removes the Persistence deployment code that previously disabled CDI for deployments that use Persistence bytecode enhancement along with application datasources or default datasource.

This pull request replaces https://github.com/wildfly/wildfly/pull/18850

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.jpa.hibernate;
+
+import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
+import org.hibernate.bytecode.enhance.spi.UnloadedClass;
+import org.hibernate.bytecode.enhance.spi.UnloadedField;
+import org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl;
+
+/**
+ * WildFlyClassTransformer is simple wrapper of Hibernate ORM EnhancingClassTransformerImpl
+ *
+ * @author Scott Marlow
+ */
+public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {
+
+    public WildFlyClassTransformer() {
+        super(new DefaultEnhancementContext() {
+
+            public boolean isEntityClass(UnloadedClass classDescriptor) {
+                return super.isEntityClass(classDescriptor);
+            }
+
+            public boolean isCompositeClass(UnloadedClass classDescriptor) {
+                return super.isCompositeClass(classDescriptor);
+            }
+
+            public boolean doBiDirectionalAssociationManagement(UnloadedField field) {
+                return super.doBiDirectionalAssociationManagement(field);
+            }
+
+            public boolean doDirtyCheckingInline(UnloadedClass classDescriptor) {
+                return super.doDirtyCheckingInline(classDescriptor);
+            }
+
+            public boolean hasLazyLoadableAttributes(UnloadedClass classDescriptor) {
+                return super.hasLazyLoadableAttributes(classDescriptor);
+            }
+
+            public boolean isLazyLoadable(UnloadedField field) {
+                return super.isLazyLoadable(field);
+            }
+
+            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
+                // doesn't make any sense to have extended enhancement enabled at runtime. we only enhance entities anyway.
+                return false;
+            }
+
+        });
+    }
+}
+
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
@@ -1,0 +1,50 @@
+package org.wildfly.persistence.jipijapa.hibernate7;
+
+import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
+import org.hibernate.bytecode.enhance.spi.UnloadedClass;
+import org.hibernate.bytecode.enhance.spi.UnloadedField;
+import org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl;
+
+/**
+ * WildFlyClassTransformer is simple wrapper of Hibernate ORM EnhancingClassTransformerImpl
+ *
+ * @author Scott Marlow
+ */
+public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {
+
+    public WildFlyClassTransformer() {
+        super(new DefaultEnhancementContext() {
+
+            public boolean isEntityClass(UnloadedClass classDescriptor) {
+                return super.isEntityClass(classDescriptor);
+            }
+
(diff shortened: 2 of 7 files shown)


jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,50 @@
+package org.wildfly.persistence.jipijapa.hibernate7;
```
scottmarlow, line 1: Will add the missing top comment after CI completes...

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -379,6 +379,9 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        if(classTransformer.getClass().getName().contains("org.hibernate")) {
```
rhusar, line 382: This should be `startsWith` rather than `contains`. My custom `com.acme.integration.org.hibernate.transformers` would be inadvertently skipped. :-)

rhusar (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -379,6 +379,11 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        if(classTransformer.getClass().getName().startsWith("org.hibernate")) {
```
rhusar, line 384: Should have mentioned in the first review, but if you want to check for the package prefix, the string to check for should really be `""org.hibernate."` - notice the trailing dot. Because `org.hibernate` would match something like `org.hibernateX`.

rhusar (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Good catch.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -379,6 +381,11 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        if(classTransformer.getClass().getName().startsWith(ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_PACKAGE)) {
+            return;
+        }
```
jamezp, line 388: Why do we not want to do this?

scottmarlow: Good question, I'm now thinking the comment should answer that.

scottmarlow: Pushed a comment change with the answer:
> WFLY-19694 Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl) which were already added earlier via org.jboss.as.jpa.hibernate.WildFlyClassTransformer or org.wildfly.persistence.jipijapa.hibernate7.WildFlyClassTransformer

scottmarlow: Will mark this resolved by ^ comment change.

scottmarlow:
The one test failure in `org.wildfly.test.integration.microprofile.faulttolerance.opentelemetry.FaultToleranceOpenTelemetryIntegrationTestCase.test` doesn't sound related.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+
+            public boolean hasLazyLoadableAttributes(UnloadedClass classDescriptor) {
+                return super.hasLazyLoadableAttributes(classDescriptor);
+            }
+
+            public boolean isLazyLoadable(UnloadedField field) {
+                return super.isLazyLoadable(field);
+            }
+
+            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
```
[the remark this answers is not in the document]
scottmarlow: 100% agree that we should use @Override here!

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+                return super.doDirtyCheckingInline(classDescriptor);
+            }
+
+            public boolean hasLazyLoadableAttributes(UnloadedClass classDescriptor) {
+                return super.hasLazyLoadableAttributes(classDescriptor);
+            }
+
+            public boolean isLazyLoadable(UnloadedField field) {
+                return super.isLazyLoadable(field);
+            }
```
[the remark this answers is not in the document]
scottmarlow: Good question, we are using an Hibernate internal API until Persistence 4.0 when we will switch to use PersistenceProvider.getClassTransformer().  Since we are using an internal API, I'd like to at least see a compile failure if the current internal SPI method signatures change.  I 100% agree that we should use @Override here as well.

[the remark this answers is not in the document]
jamezp: Ah, no worries. I was more curious than anything.

darranl:
I see prototype in the title, does this need mapping based on stability levels?

scottmarlow:
> I see prototype in the title, does this need mapping based on stability levels?

I'm going to update the branch and change this pr to draft.  I think if we do find this change is needed for WildFly 40+, we do need mapping based on stability levels.

This is a prototype for a Jakarta EE 12/Persistence 4.0 that proves https://github.com/jakartaee/persistence/issues/650 can be done.  In the WildFly EE 12 implementation we will use the updated Persistence spec interfaces as mentioned in https://github.com/jakartaee/persistence/pull/671 which will look different than this early WildFly change.

scottmarlow:
Replaced by https://github.com/wildfly/wildfly/pull/19569

PR #18864, opened 2025-04-04
[WFLY-20542] EJB: Add SocketPermission to DatabaseTimerServiceMultiNo…

tadamski:
…deExecutionDisabledTestCase

https://issues.redhat.com/browse/WFLY-20542

The test was failing intermittently because the in most cases connection created by server code is reused. If connection is closed and deployment code has to create a new one, then the SocketPermission was lacking.


bstansberry:
Thanks @tadamski

PR #18869, opened 2025-04-05
[WFLY-20545] At Server.lookupSecurePort, prevent the NPE when unboxin…

fl4via:
…g the return result by checking if it is null

Jira: https://issues.redhat.com/browse/WFLY-20545


bstansberry (approved):

bstansberry:
Thanks @fl4via

PR #18870, opened 2025-04-05
[WFLY-20546] At ListenerService.start and stop, make sure the port is op…

fl4via:
…en only when the service is fully configured and ready to accept requests

Jira: https://issues.redhat.com/browse/WFLY-20546

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/ListenerService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/ListenerService.java
@@ -5,25 +5,14 @@
 
 package org.wildfly.extension.undertow;
 
-import java.io.IOException;
-import java.net.BindException;
-import java.net.InetSocketAddress;
-import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/ListenerService.java
```diff
@@ -43,6 +30,14 @@
 import org.xnio.XnioWorker;
 import org.xnio.channels.AcceptingChannel;
 
+import javax.net.ssl.SSLContext;
+import java.io.IOException;
+import java.net.BindException;
+import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
```
bstansberry, line 39: The original import ordering was correct.

bstansberry:
Other than the import order thing, LGTM, Good catch!

bstansberry:
Thanks @fl4via

PR #18876, opened 2025-04-09
[WFLY-20557] Application deployment with container-interceptors fails…

jbaesner:
Jira: https://issues.redhat.com/browse/WFLY-20557


tadamski (approved):
LGTM. @jbaesner thanks for finding and fixing the bug.

tadamski (changes requested):
There is a a checkstyle type in InterceptorOrder. Also I have extended the integration test already and verified that without Joerg's fix the test fails and after the fix it is green.

The checkstyle fix and the test are here: https://github.com/tadamski/wildfly/tree/WFLY-20557 but I can't open a PR to Joerg's Widfly - have no idea why.

@jbaesner maybe you could just cherry-pick the last two commits? After they are applied this PR LGTM.

bstansberry:
@tadamski Is this ok now?

tadamski (approved):
@bstansberry yes, now it lgtm

PR #18880, opened 2025-04-10
[WFLY-20553] Make MailTesrCase more resilant when running with a podman container.

darranl:
https://issues.redhat.com/browse/WFLY-20553

The underlying issue was that the configuration files were not getting copied into a directory also mapped as a volume.


jasondlee (approved):
lgtm

bstansberry (approved):

PR #18883, opened 2025-04-14
WFLY-20559 Concurrent build is broken because modules do not depend on feature-packs they provision that are built in other modules

rhusar:
https://issues.redhat.com/browse/WFLY-20559

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -401,6 +401,17 @@
                 <module>aggregator-base</module>
                 <module>integration</module>
             </modules>
+            <dependencies>
+                <!-- Include a dependency on the build demander because in concurrent builds the order
+                     of items in the 'modules' element doesn't ensure earlier modules are
+                     built before later ones. So use a dependency make sure it is executed. -->
+                <dependency>
+                    <groupId>${ee.maven.groupId}</groupId>
+                    <artifactId>wildfly-ts-build-demander-base</artifactId>
+                    <version>${ee.maven.version}</version>
+                    <type>pom</type>
+                </dependency>
+            </dependencies>
         </profile>
         <profile>
             <id>expansion-server-tests</id>
@@ -411,6 +422,17 @@
                 <module>aggregator-expansion</module>
                 <module>integration</module>
             </modules>
+            <dependencies>
+                <!-- Include a dependency on the build demander because in concurrent builds the order
+                     of items in the 'modules' element doesn't ensure earlier modules are
+                     built before later ones. So use a dependency make sure it is executed. -->
+                <dependency>
+                    <groupId>${full.maven.groupId}</groupId>
+                    <artifactId>wildfly-ts-build-demander-expansion</artifactId>
+                    <version>${full.maven.version}</version>
+                    <type>pom</type>
+                </dependency>
+            </dependencies>
         </profile>
         <profile>
             <id>preview-server-tests</id>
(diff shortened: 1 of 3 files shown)


rhusar:
@jfdenise @jamezp Let me guys know if you can thing of a better solution to this. More details on the [WFLY-20559](https://issues.redhat.com/browse/WFLY-20559).

jfdenise:
There could be more feature-packs that could be not yet build when running the tests I am thinking at:
* testsuite.full.galleon.pack.groupId/testsuite.full.galleon.pack.artifactId
* testsuite.test.feature.pack.artifactId
* wildfly-preview-feature-pack
They should be added as dependency too I guess.

bstansberry:
I don't get why testsuite/build-demander-base isn't triggering this. With 'mvn clean install -Dquickly -T 3C -DskipTests' it should have a dependency on org.wildfly:wildfly-build and org.wildfly:wildfly-dist, which in turn should depend on the feature packs.

Without the -T 3C those modules execute before testsuite/aggregator-base and testsuite/integration.

I don't think testsuite/integration is the place for this. It doesn't encapsulate the entire TS.

bstansberry:
So the problem is there's no dependency on the build-demander-xxx modules. The build is relying on the specified module order for controlling the order of module execution.

bstansberry:
Adding such dependencies doesn't result in the build-demander-xxx modules executing before the aggregator-xxx modules that now have the dep though.

Ok, so the issue is testsuite/integration has no deps on the aggregators. It relies on the module order listing.  That will take more thought.

bstansberry:
@rhusar https://github.com/rhusar/wildfly/pull/11 seems to work and is more in-line with the overall design.

rhusar:
> @rhusar [rhusar#11](https://github.com/rhusar/wildfly/pull/11) seems to work and is more in-line with the overall design.

Thanks Brian, after your last comment I realized this can be much simpler with the use of build-demander-xyz. Honestly, I didn't realize that we have a such a module to do that! 

Tested locally and this looks good.

Nevertheless, this still doesn't solve the problem completely but that's out of scope for this missing dependency fix. The other issue is that the bom builder doesn't yet support concurrent builds, [comment on Jira](https://issues.redhat.com/browse/WFLY-20559?focusedId=26987798&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26987798).

jamezp (approved):

bstansberry:
Thanks @rhusar and @jamezp

testsuite/pom.xml
```diff
@@ -411,6 +422,17 @@
...
+            <dependencies>
+                <!-- Include a dependency on the build demander because in concurrent builds the order
+                     of items in the 'modules' element doesn't ensure earlier modules are
+                     built before later ones. So use a dependency make sure it is executed. -->
+                <dependency>
+                    <groupId>${full.maven.groupId}</groupId>
+                    <artifactId>wildfly-ts-build-demander-expansion</artifactId>
+                    <version>${full.maven.version}</version>
+                    <type>pom</type>
+                </dependency>
```
yersan, line 434: @bstansberry  This is now a source of trouble for the product branches, there we do not really want to have the `wildfly-ts-build-demander-expansion` under the `ull.maven.groupId` coordinates.

 Our requirements have changed regarding how we are now producing our product, so we should reconsider this change and move it back to:

```
            <dependency>
                    <groupId>${ee.maven.groupId}</groupId>
                    <artifactId>wildfly-ts-build-demander-expansion</artifactId>
                    <version>${ee.maven.groupId}</version>
                    <type>pom</type>
            </dependency>
```

I was tempted to create a WFLY Jira to apply this change, but it will be confusing to talk about an expansion module that is under the ee.maven.groupId, so for now, I'll fix it at the conversion branches stage, but let me know your thoughts if you wants to manage it differently. The point is not all the artifacts we have under the expansions are now required to be on `full.maven.groupId`

bstansberry: @yersan I'm fine with you sending a fix for this, if you want. We need to completely revisit ee|full.maven.groupId etc as they are of little use upstream and the way upstream uses them no longer matches how downstream does things. So I'm not much bothered if this one is non-intuitive. Maybe just add an xml comment like "We use ee.maven.groupId on purpose" to help stop people deciding to 'fix it'.

rhusar: @yersan BTW Is this documented upstream anywhere?

yersan: @rhusar It is not. We have these maven properties available because, at the product level, we switch the GAV for a set of dependencies. These dependencies used to be those distributed by the WildFly galleon pack. However, our requirements on the product side have changed, and keeping the current relationship of these GAV upstream creates some difficulties downstream.

PR #18885, opened 2025-04-15
[WFLY-20563] HAPROXY configuration: add "check" keyword

tommaso-borgato:
https://issues.redhat.com/browse/WFLY-20563

In e.g. https://www.haproxy.com/blog/enable-sticky-sessions-in-haproxy the check cookie is used: @rhusar should we add it here?

Btw, is https://github.com/wildfly/wildfly.github.io/pull/88 needed?


rhusar (approved):
Looks good, thanks for the fix @tommaso-borgato !

bstansberry:
Thanks @tommaso-borgato and @rhusar

PR #18886, opened 2025-04-15
[WFLY-20564] Deploying postgresql-42.7.5.jar causes NPE when upgrading from wildfly 35->36

scottmarlow:
https://issues.redhat.com/browse/WFLY-20564

Thanks for the bug report Stephen Sill!  Very much appreciated!


bstansberry:
Thanks @scottmarlow.

Please send up a PR for this against the 36.x branch so it's fixed in WF 36.0.1. @kabir FYI

PR #18889, opened 2025-04-17
WFLY-20567 MicroProfile OpenAPI subsystem does not allow multiple deployments to use the same endpoint

pferraro:
https://issues.redhat.com/browse/WFLY-20567

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -428,7 +428,7 @@
         <version.org.apache.cxf.xjcplugins>4.1.0</version.org.apache.cxf.xjcplugins>
         <version.org.apache.httpcomponents.httpasyncclient>4.1.5</version.org.apache.httpcomponents.httpasyncclient>
         <version.org.apache.james.apache-mime4j>0.8.12</version.org.apache.james.apache-mime4j>
-        <version.org.apache.kafka>3.9.0</version.org.apache.kafka>
+        <version.org.apache.kafka>3.9.1</version.org.apache.kafka>
         <version.org.apache.kafka.test-image>3.8.0</version.org.apache.kafka.test-image>
         <version.org.apache.lucene>9.11.1</version.org.apache.lucene>
         <version.org.apache.neethi>3.2.1</version.org.apache.neethi>


undertow/extension/src/main/java/org/wildfly/extension/undertow/HostRemove.java
```diff
@@ -1,49 +0,0 @@
...
-import org.wildfly.extension.undertow.deployment.DefaultDeploymentMappingProvider;
-
-/**
- * @author <a href="mailto:tomaz.cerar@redhat.com">Tomaz Cerar</a> (c) 2013 Red Hat Inc.
- */
-class HostRemove extends AbstractRemoveStepHandler {
-
-    @Override
-    protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
-        if (context.isResourceServiceRestartAllowed()) {
```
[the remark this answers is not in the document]
pferraro: It does. The generic remove resource operation handler in wildfly-subsystem implements service removal by conditionally invoking ResourceOperationRuntimeHandler.removeRuntime(...) [1] and service recovery by conditionally invoking ResourceOperationRuntimeHandler.addRuntime(...) [2].
The rollback handling for the add resource and write attribute operation handlers behave in a similar fashion.

[1] https://github.com/wildfly/wildfly-core/blob/main/subsystem/src/main/java/org/wildfly/subsystem/resource/operation/RemoveResourceOperationStepHandler.java#L90
[2] https://github.com/wildfly/wildfly-core/blob/main/subsystem/src/main/java/org/wildfly/subsystem/resource/operation/RemoveResourceOperationStepHandler.java#L105

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
```diff
@@ -38,49 +42,55 @@ class HostDefinition extends SimpleResourceDefinition {
 
     static final StringListAttributeDefinition ALIAS = new StringListAttributeDefinition.Builder(Constants.ALIAS)
             .setRequired(false)
-            .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
             .setElementValidator(new StringLengthValidator(1))
             .setAllowExpression(true)
             .setAttributeParser(AttributeParser.COMMA_DELIMITED_STRING_LIST)
             .setAttributeMarshaller(AttributeMarshaller.COMMA_STRING_LIST)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
```
bstansberry, line 49: Why did you change these?

AIUI this commit means a remove op no longer supports resource service restart, but now these attributes do for write-attribute.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc
```diff
@@ -51,6 +51,54 @@ In addition to the https://download.eclipse.org/microprofile/microprofile-open-a
...
+|_None_
+|Defines the summary of the OpenAPI document provided by this host.
+
+|mp.openapi.extensions._server-name_._host-name_.info.termsOfService
+|_None_
+|Defines the terms of service of the OpenAPI document provided by this host.
+
+|mp.openapi.extensions._server-name_._host-name_.info.openapi
+|{OpenAPIVersion}.0
+|Defines the version of the OpenAPI document provided by this host.
```
bstansberry, line 100: Do all these apply in the non-discordant case? If not they may all need a few words and a link to the NOTE text.

testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java
```diff
@@ -200,9 +200,6 @@ public abstract class LayersTestBase {
             "org.eclipse.microprofile.lra.api",
             // Extension not included in the default config
             "org.wildfly.extension.microprofile.openapi-smallrye",
-            "org.eclipse.microprofile.openapi.api",
-            "io.smallrye.openapi",
-            "com.fasterxml.jackson.dataformat.jackson-dataformat-yaml",
```
bstansberry, line 205: Why are these now referenced?

pferraro: Because of this additional optional dependency: https://github.com/pferraro/wildfly/blob/openapi/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/undertow/main/module.xml#L47

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
pferraro: > Why did you change these?

Because this correpsonds to the actual runtime behavior of this resource.  Only the services installed by this resource need to be restarted in the event of a write-attribute operation.

> AIUI this commit means a remove op no longer supports resource service restart, but now these attributes do for write-attribute.

It does.  The default restart level for remove operations created via wildfly-subsystem is OperationEntry.Flag.RESTART_NONE [1], unless a runtime handler is specified, in which case the default becomes OperationEntry.Flag.RESTART_RESOURCE_SERVICES [2].

[1] https://github.com/wildfly/wildfly-core/blob/main/subsystem/src/main/java/org/wildfly/subsystem/resource/ResourceDescriptor.java#L643
[2] https://github.com/wildfly/wildfly-core/blob/main/subsystem/src/main/java/org/wildfly/subsystem/resource/ResourceDescriptor.java#L668

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc
pferraro: Yes.

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
bstansberry: If I'm understanding this correctly, this requires an API version bump. It doesn't sound like a case of the API not matching what the handlers did; it's an actual change of behavior.

Has there been a lot of this kind of thing in all these migrations?

TBH, I'm not thrilled with that default behavior. I kind of regret implementing restart-resource-services, as users have little ability to understand the implications of using it.

pferraro: > If I'm understanding this correctly, this requires an API version bump. It doesn't sound like a case of the API not matching what the handlers did; it's an actual change of behavior.

If a change does not affect how a user interacts with the subsystem, i.e. the same commands have the same end state, does this, in fact, necessitate an API version bump?
e.g. /subsystem=undertow/server=foo/host=bar:write-attribute(name=aliases, value=baz) yields reload-required.
e.g. /subsystem=undertow/server=foo/host=bar:remove yields reload-required.

> Has there been a lot of this kind of thing in all these migrations?

No.  The modifications to the operation/attribute flags are unrelated to the use of wildfly-subsystem.  The operation handlers in wildfly-subsystem do not prescribe specific runtime flags.

> TBH, I'm not thrilled with that default behavior. I kind of regret implementing restart-resource-services, as users have little ability to understand the implications of using it.

Given that restart-resource-services requires the use of the allow-resource-service-restart operation header (making it effectively opt-in behaviour), have you received negative feedback about this behaviour?

I assume your reservations about this behaviour are due to the lack of feedback about affected dependent services?

It would be nice if we had an "interactive" mode, where operations that would result in the stop or restart of dependent services were communicated and authorised via a callback.
e.g.
```
$ /subsystem=undertow/server=foo/host=bar:remove{interactive=true}
This operation will affect the following running services:
* ServiceA
* ServiceB
Allow service restart? (y/N)
```
This could then add the allow-resource-service-restart=true/false header to the operation context depending on the callback response.
To generate the list of affected services, we would need to query the service container for services that require (directly or transitively) any running services that were installed by the resource.

Alternatively, we could communicate this more abstractly by listing the addresses of resources containing references to the restarting resource.  While not as specific as the above, it should be more comprehensible to users as well as easier to implement:
e.g.
```
$ /subsystem=undertow/server=foo/host=bar:write-attribute(name=aliases, value=baz){interactive=true}
This operation will affect runtime services installed by the following resources:
* /subsystem=a/resource=b
* /subsystem=c/resource=d
Allow service restart? (y/N)
```
To generate this list, we would query the capability registry for resources containing capability references to the associated resource.

If this were feasible, would something like this address your concerns?

pferraro:
Blech - I inadvertently force pushed this branch causing it to close.  Will reopen in a new pull request.

PR #18892, opened 2025-04-17
[WFLY-20569] define schema that accurately imposes the rule for eithe…

rsearls:
…r realm or provider in secure-deployment-type definition.

https://issues.redhat.com/browse/WFLY-20569

### DIFF
--- a/elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_2_0.xsd
+++ b/elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_2_0.xsd
@@ -799,21 +799,15 @@
                     </xs:documentation>
                 </xs:annotation>
             </xs:element>
-            <xs:element name="realm" type="xs:string" minOccurs="0" maxOccurs="1">
+            <xs:element ref="realm-or-provider" minOccurs="0" maxOccurs="1">
                 <xs:annotation>
                     <xs:documentation>
                         <![CDATA[
-                        The Keycloak realm to use for authentication.
-                        ]]>
-                    </xs:documentation>
-                </xs:annotation>
-            </xs:element>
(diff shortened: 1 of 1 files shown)


elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_2_0.xsd
```diff
@@ -799,21 +799,15 @@
                     </xs:documentation>
                 </xs:annotation>
             </xs:element>
-            <xs:element name="realm" type="xs:string" minOccurs="0" maxOccurs="1">
+            <xs:element ref="realm-or-provider" minOccurs="0" maxOccurs="1">
```
pferraro, line 802: This would be better modeled as an optional xs:choice containing the realm and provider elements.

pferraro (changes requested):
The 2.0 version of this subsystem schema was published several releases ago.  You'll want to create a new 3.0 version containing your changes.
Additionally, I believe the latest preview version of the schema is also affected?
It seems to me that we should also enforce this constraint in the management model.  If the "realm" and "provider" attributes are mutally exclusive, then the attribute definitions for each should specify that they are alternatives of the other.

rsearls:
@pferraro   Thank you for the review.
Unfortunately xs:alternative is an XSD 1.1 feature and wildfly
is XSD 1.0 based.   I don't foresee that changing any time soon.
xs:choice is not a valid element inside an xs:all, thus I was not
able to use it for this scenario.

It was my understanding that if a XSD change did not change the basic
function of the schema, it could be made.  In this case the change was
made more for user clarification.  I determined it does not alter the
under laying processing of the schema.  However I will move this change
to a more current preview version.

rsearls:
removed as needs more changes and moved into a newer schema version.

PR #18903, opened 2025-04-23
[WFLY-19970][WFLY-20588] WeldProvide bean manager lookup improvements

bstansberry:
https://issues.redhat.com/browse/WFLY-20588 -- drop out of date workaround

https://issues.redhat.com/browse/WFLY-19970 -- limit the workaround to the ear use case where it's needed


manovotn (approved):
Thanks, LGTM

PR #18909, opened 2025-04-26
[WFLY-20594] Remove use of org.jboss.threads.BlockingExecutor

bstansberry:
https://issues.redhat.com/browse/WFLY-20594
https://issues.redhat.com/browse/WFLY-20597

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/services/workmanager/DistributedWorkManagerService.java
+++ b/connector/src/main/java/org/jboss/as/connector/services/workmanager/DistributedWorkManagerService.java
@@ -22,7 +22,6 @@ import org.jboss.msc.service.StartContext;
 import org.jboss.msc.service.StartException;
 import org.jboss.msc.service.StopContext;
 import org.jboss.msc.value.InjectedValue;
-import org.jboss.threads.BlockingExecutor;
 import org.wildfly.clustering.server.dispatcher.CommandDispatcherFactory;
 
 /**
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/jboss/as/connector/services/workmanager/DistributedWorkManagerService.java
```diff
@@ -67,13 +66,13 @@ public void start(StartContext context) throws StartException {
 
         this.value.setTransport(transport);
 
-        BlockingExecutor longRunning = (BlockingExecutor) executorLong.getOptionalValue();
+        Executor longRunning = executorLong.getOptionalValue();
         if (longRunning != null) {
             this.value.setLongRunningThreadPool(longRunning);
```
bstansberry, line 71: @tadamski Is there a reason why this one isn't StatisticsExecutorImpl?

bstansberry: @tadamski The github ui doesn't show the context for my question, but in case it wasn't clear: I wasn't wondering about the use of BlockingExecutor here. It's about how all the other lines in the PR where the executor is used it's wrapper in StatisticsExecutorImpl, but this one doesn't wrap it.  Perhaps it was missed?

tadamski: @bstansberry sorry, I misunderstood the question. Looking at the code I don't see any reason why long running pool should be treated differently from short term one. I think this is an omission and longRunning executor should be wrapper as well (also in non-distributed WorkManagerService).

bstansberry: @tadamski I filed https://issues.redhat.com/browse/WFLY-20597 for this and will add a commmit to this PR for it.

bstansberry: Done!

bstansberry:
@tadamski Is this ok?

tadamski:
@bstansberry yes, lgtm, approved

tadamski (approved):

bstansberry:
Thanks, @tadamski

PR #18920, opened 2025-05-05
[WFLY-20609]: ActiveMQ Artemis fails to launch if system property with prefix "brokerconfig." is present.

ehsavoie:
* Adding java.xml as a module dependency for `org.apache.activemq.artemis.client`.

Jira: https://issues.redhat.com/browse/WFLY-20609


bstansberry:
Thanks @ehsavoie

PR #18935, opened 2025-05-13
[WFLY-17742] Support Transactions during WildFly’s graceful shutdown

jmfinelli:
This PR addresses [WFLY-17742](https://issues.redhat.com/browse/WFLY-17742)

____

> [!WARNING]

> * [WFLY-17742](https://redhat.atlassian.net/browse/WFLY-17742)


transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
```diff
@@ -30,6 +26,8 @@
 import org.jboss.logging.annotations.Param;
 import org.jboss.msc.service.StartException;
 
+import static org.jboss.logging.Logger.Level.*;
+
```
bstansberry: This should be reverted.

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
```diff
@@ -254,4 +252,12 @@ public interface TransactionLogger extends BasicLogger {
 
     @Message(id = 45, value = "Could not register initial reference for InboundTransactionCurrent implementation")
     RuntimeException cannotRegister(@Cause Exception e);
+
+    @LogMessage(level = INFO)
+    @Message(id = 46, value = "Transactions subsystem: starting transactions recovery suspension")
+    void scan_suspension_initiated();
```
bstansberry: This is kind of up to the subsystem maintainers, but this kind of method name is not used elsewhere in this file and if it's used elsewhere in this code base in production code it's very rare.

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
```diff
@@ -254,4 +252,12 @@ public interface TransactionLogger extends BasicLogger {
...
+
+    @LogMessage(level = INFO)
+    @Message(id = 47, value = "Transactions subsystem: transactions recovery suspension completed")
+    void scan_suspension_completed();
```
rhusar: Java method names are camelCase.

rhusar (changes requested):

jmfinelli:
`Cloud Tests Remote Run` failures don't seem related to this PR.
EDIT: re-run it after the related repo was updated and it passed

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
```diff
@@ -187,6 +187,11 @@ private static ModelNode textifyMap(ModelNode modelNode) {
             }
             return new ModelNode(sb.toString());
         }
+
+        @Override
+        protected long timeout() {
+            return 20L;
+        }
```
bstansberry: Nit: this override is redundant. I assume it predates switching to ExtendedSnapshotServerSetupTask

testsuite/shared/src/main/java/org/jboss/as/test/shared/ExtendedSnapshotServerSetupTask.java
```diff
@@ -0,0 +1,11 @@
+package org.jboss.as.test.shared;
+
+import org.jboss.as.arquillian.setup.SnapshotServerSetupTask;
+
+public class ExtendedSnapshotServerSetupTask extends SnapshotServerSetupTask {
```
bstansberry: Unless @jamezp disagrees please file a WFARQ and a PR to get SnapshotServerSetupTask changed so we can later remove the use of this subclass.

AFAICT this PR changes all uses of SnapshotServerSetupTask in this repo, and I suspect that means all or nearly all uses anywhere.

jamezp: +1 and if we want this, I can get a release out pretty quickly for it so we don't even need to introduce it here.

bstansberry (commented):
What makes the @ReloadIfRequired and SnapshotServerSetupTasks cases different from all the other ways servers are reloaded in the testsuite? Is it just that the timeouts involved with any use of ServerReload are already longer than 20 seconds?

testsuite/shared/src/main/java/org/jboss/as/test/shared/ExtendedSnapshotServerSetupTask.java
jamezp: I think a simple solution is to allow this to be configured either via a system property or via the `arquillian.xml`, the latter I'd have to see how feasible that is. A system property would be simple though.

bstansberry:
The proposal discusses community documentation changes but I don't see any.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ExtendedSnapshotServerSetupTask.java
bstansberry: Great. Let's not block on this though. My impression is this is nearing completion and it will be a few minutes work to remove this class in the future.

transactions/src/main/java/org/jboss/as/txn/suspend/RecoverySuspendController.java
```diff
@@ -63,13 +77,15 @@ public void resume() {
         if (doResume) {
             resumeRecovery();
```
pferraro: Does this call block the thread?  If so, you will want to have this method return a stage that completes when resumeRecovery() completes.

transactions/src/main/java/org/jboss/as/txn/suspend/RecoverySuspendController.java
```diff
@@ -63,13 +77,15 @@ public void resume() {
...
 
     /**
      * Receives notifications from a {@code ProcessStateNotifier} to detect when the process has reached a
      * {@link ControlledProcessState.State#isRunning()}  running state}, reacting to them by
      * {@link RecoveryManagerService#resume() resuming} the {@link RecoveryManagerService} if we haven't been
-     * {@link #preSuspend(ServerActivityCallback) suspended}.
+     * {@link #suspend(ServerSuspendContext) suspended}.
      */
     @Override
     public void propertyChange(PropertyChangeEvent evt) {
```
pferraro, line 112: Is this logic actually needed anymore?

bstansberry: Why wouldn't it be? I don't think this RFE changes the use case that led to this logic.

This kind of thing is due to the legal ability to start a server non-gracefully, i.e. where users specifically want things like like listeners listening even before all deployment is done. That's implemented by resuming things before the server is fully running, but some things don't want to start then so they employ this pattern.

pferraro: I did not mean to question whether this was newly redundant, just wondering if this logic pre-dated the suspend controller.

transactions/src/main/java/org/jboss/as/txn/suspend/RecoverySuspendController.java
jmfinelli: Thanks Paul. I completely forgot about this!

testsuite/pom.xml
```diff
@@ -131,8 +131,11 @@
         <!-- jboss.args is used in the arquillian.xml files, but wildfly-core DomainLifecycleUtil expects jboss.domain.server.args -->
         <jboss.domain.server.args></jboss.domain.server.args>
 
+        <!-- Additional JVM args for Narayana's recovery manager -->
+        <narayana.server.jvm.args>-Dcom.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.periodicRecoveryPeriod=5 -Dcom.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.recoveryBackoffPeriod=1</narayana.server.jvm.args>
```
bstansberry: @jmfinelli We chatted a bit about this PR and you're going to start a Zulip thread. That's important as it looks like the PR will have a significant impact on how CI works, so it needs broader awareness / input than PR review is good for.

This change seems to be really significant in effect so we need to discuss exactly what it does. The production code changes here AFAICT will have little impact on all the reloading that the TS does, as 'reload' is not graceful. The SuspendController is not invoked in a reload. So all the many changes to TS files and the timing of reload I don't think are related to the production code changes. So I think they must be related to the effect of these settings on server behavior.

bstansberry:
The 'Linux SM - JDK 17 (Pull Request)' failures are related.

bstansberry:
This needs to go through a few retest cycles, both because the various timeout setting changes imply potentially instability, and because I want to get a sense of the impact of this change on CI execution time. The fork timeout change from 3600 to 5400 implies it could be signficant.

CI execution times vary a decent amount though so it's hard to treat a single run as meaningful data.

I'm not kicking off a round now though because CI is quite busy. I'd like to see what happens when it has calmed down a bit.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
```diff
@@ -182,6 +182,7 @@ public void beforeTestMethod() throws Exception {
     public void afterTestMethod() throws Exception {
         this.start();
         this.undeploy();
+        this.stop();
```
rhusar: Absolutely not, that would slow down tests to ridiculous extent.

rhusar (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/AbstractClusteringTestCase.java
rhusar: Also, that would violate the contract defined in https://github.com/wildfly/wildfly/pull/18935/changes#diff-d112757c7b4b0a027bf324f57b32ea46366b1b9216500dffbecf728b7b0c29c9R27-R38

docs/src/main/asciidoc/_admin-guide/management-tasks/Suspend,_Resume_and_Graceful_shutdown.adoc
```diff
@@ -117,12 +117,23 @@ is suspending. They will be restarted from that checkpoint when the
...
+
+[NOTE]
+====
+Heuristic transactions do not delay a graceful shutdown.
+However, they are reported through the standard logging mechanisms and
+require manual reconciliation after the server has shut down.
+====
+
+WARNING: In cloud environments, where pods are ephemeral and logs are not retained, specify a persistent and secure location for saving the object store and report logs.
+--
```
rhusar: Remove this line - that will break all rendering in the file.

rhusar (changes requested):

docs/src/main/asciidoc/_admin-guide/management-tasks/Suspend,_Resume_and_Graceful_shutdown.adoc
jmfinelli: Done. Thanks @rhusar

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
```diff
@@ -103,6 +103,12 @@ public class TransactionSubsystemRootResourceDefinition extends SimpleResourceDe
             .setXmlName(Attribute.RECOVERY_LISTENER.getLocalName())
             .setAllowExpression(true).build();
 
+    public static final SimpleAttributeDefinition TRANSACTIONS_RECOVERY_GRACEFUL_SHUTDOWN = new SimpleAttributeDefinitionBuilder(CommonAttributes.TRANSACTIONS_RECOVERY_GRACEFUL_SHUTDOWN, ModelType.BOOLEAN, true)
```
rhusar: Very rarely it is a good design to model anything as a boolean. You can never then evolve that value in any way. Are you 100% sure there will never be another strategy? Example how this could be modeled e.g. `transaction-recovery-shutdown-strategy="IGNORE|WAIT"`

jmfinelli: Good point. And I like your suggestion (i.e. `transaction-recovery-shutdown-strategy="IGNORE|WAIT"`). I'll try to implement it

jmfinelli: Done. Thanks

Note: when the PR is approved, I will squeeze all commits together

jmfinelli: BTW, do you think my choice for the attribute name (i.e.`transactions-recovery-graceful-shutdown`) is good enough?

rhusar: @jmfinelli I did not mean a String type, but a Enum type :-)

jmfinelli: OK. I'll modify it.

jmfinelli: Done.

Note: the last commit was generated partially by Claude. It seems good though

transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
```diff
@@ -155,15 +166,15 @@ public void start(final StartContext context) throws StartException {
                 throw TransactionLogger.ROOT_LOGGER.managerStartFailure(e, "Recovery");
             }
         }
-        recoverySuspendController = new RecoverySuspendController(recoveryManagerService);
+        recoverySuspendController = new RecoverySuspendController(recoveryManagerService, gracefulRecoveryShutdown, executorSupplier.get());
         processStateSupplier.get().addPropertyChangeListener(recoverySuspendController);
-        suspendControllerSupplier.get().registerActivity(recoverySuspendController);
+        activityRegistration = suspendableActivityRegistrarSupplier.get().register(recoverySuspendController, SuspendPriority.TEN);
```
pferraro: Just be aware that there is a semantic difference between SuspendableActivityRegistrar.register(...) and the registration methods on the legacy interfaces.
Previously, the suspend controller would auto-invoke the SuspendableActivity.suspend(...) method invoked while the server is suspended, blocking until complete.  The new registration method does not. For details, see https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/suspend/SuspendableActivityRegistry.java#L63

For readability, I would using SuspendPriority.LAST.

jmfinelli: >Previously, the suspend controller would auto-invoke the SuspendableActivity.suspend(...) method invoked while the server is suspended, blocking until complete. The new registration method does not

Thanks for highlighting that semantic difference, Paul! Since the new logic in `RecoverySuspendController.suspend()` should be invoked only when the server is stopping, I expect that the semantic difference between `SuspendableActivityRegistrar.register(...)` and the registration methods on the legacy interface shouldn't be an issue here.

> For readability, I would using SuspendPriority.LAST.

Done. Thanks!

docs/src/main/asciidoc/_admin-guide/management-tasks/Suspend,_Resume_and_Graceful_shutdown.adoc
```diff
@@ -117,12 +117,44 @@ is suspending. They will be restarted from that checkpoint when the
...
-incoming remote call which carries the context of a transaction already
-started at the suspending server.
+
+* *Transactions* - While the server is suspending, the transaction subsystem
+waits for all _in-flight_ transactions to reach a terminal state.
+During that time the server refuses to start any new transaction.
+However, transactions already in progress continue to be serviced;
+e.g. the server will accept any incoming remote call
+which carries the context of a transaction already started at the suspending server.
++
```
bstansberry: Perhaps a sub-heading

"Transaction Recovery and Server Suspension"

docs/src/main/asciidoc/_admin-guide/management-tasks/Suspend,_Resume_and_Graceful_shutdown.adoc
```diff
@@ -117,12 +117,44 @@ is suspending. They will be restarted from that checkpoint when the
...
+If the `transactions-recovery-graceful-shutdown` attribute is set to `wait` in the
+transactions subsystem, the server will also wait for _in-doubt_ transactions that
+require recovery to be resolved before fully suspending. By *default* this attribute
+is `ignore`, meaning the recovery manager will not delay the suspension of the server
```
bstansberry: s/is `ignore`/is set to `ignore`/g

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
```diff
@@ -0,0 +1,50 @@
+package org.jboss.as.test.manualmode.transaction;
```
bstansberry: Please add the copyright/license header

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.PathAddress;
+import org.jboss.as.controller.client.ModelControllerClient;
+import org.jboss.as.test.integration.management.base.AbstractExpressionSupportTestCase;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
```
bstansberry: java imports before other non-static imports

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+
+import static org.jboss.as.controller.operations.common.Util.getWriteAttributeOperation;
```
bstansberry: static imports first

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+
+@RunWith(Arquillian.class)
+public class TransactionsRecoveryGracefulShutdownExpressionSupportTestCase extends AbstractExpressionSupportTestCase {
```
bstansberry: I don't see the point of this test. This attribute is in the standard config so ExpressionSupportSmokeTestCase covers this attribute.

bstansberry (commented):
This needs an ExtensionTransformerRegistration implementation to handle the new attribute. This is what makes mixed domain work.

This subsystem had one but we dropped it in https://github.com/wildfly/wildfly/pull/14676/changes because we no longer needed to support transforming to versions before 6.0. But now we need it again. But it doesn't need to cover the versions before 6.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
jmfinelli: TBH, it was added after I reviewed an internal document about adding attribute testing. I'll happily remove it. cc @OndrejGerzicak

jmfinelli:
> This needs an ExtensionTransformerRegistration implementation to handle the new attribute. This is what makes mixed domain work.
> 
> This subsystem had one but we dropped it in https://github.com/wildfly/wildfly/pull/14676/changes because we no longer needed to support transforming to versions before 6.0. But now we need it again. But it doesn't need to cover the versions before 6.

I hope I addressed this correctly 🤞 Please let me know if I need to modify the Transformer further. Thanks 😄

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionTransformers.java
```diff
@@ -0,0 +1,42 @@
...
+import org.jboss.as.controller.transform.SubsystemTransformerRegistration;
+import org.jboss.as.controller.transform.description.ChainedTransformationDescriptionBuilder;
+import org.jboss.as.controller.transform.description.DiscardAttributeChecker;
+import org.jboss.as.controller.transform.description.RejectAttributeChecker;
+import org.jboss.as.controller.transform.description.ResourceTransformationDescriptionBuilder;
+import org.jboss.as.controller.transform.description.TransformationDescriptionBuilder;
+
+public class TransactionTransformers implements ExtensionTransformerRegistration {
+
+    static final ModelVersion MODEL_VERSION_EAP80 = ModelVersion.create(6, 0);
```
bstansberry: @jmfinelli EAP81 is a better name as that is the latest release using version 6

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -101,7 +112,12 @@ public void testParser_EAP_7_3() throws Exception {
     }
 
     @Test
-    public void testParser_EAP_7_4() throws Exception {
+    public void testParser_EAP_7_4_and_EAP_8_0() throws Exception {
```
bstansberry: This actually should be more like 'testParser_EAP_7_4_to_EAP_8_1()' as version 6 was used in EAP 8.1.

We should refactor this class after this RFE is done to extend AbstractSubsystemSchemaTest, use @RunWith(Parameterized.class) and eliminate these testParserEAPXXX methods, which introduce problems with names. But for now just tweaking method names is enough.

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -101,7 +112,12 @@ public void testParser_EAP_7_3() throws Exception {
...
+        standardSubsystemTest("full-6.0.0.xml");
+    }
+
+    @Test
+    public void testParser_EAP_8_1() throws Exception {
```
bstansberry: This method name is wrong as EAP 8.1 does not support urn:jboss:domain:transactions:7.0 and it never will. More accurate is 'testParserWildFly40'.

This is why basing on AbstractSubsystemSchemaTest and using @RunWith(Parameterized.class) is better; no need to mess with versions in the method name. But to get this done tomorrow it's simpler to just change this method name.

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -129,6 +145,59 @@ public void testAsyncIOExpressions() throws Exception {
         standardSubsystemTest("async-io-expressions.xml");
     }
 
+    @Test
+    public void testTransformersFullEAP800() throws Exception {
```
bstansberry: testTransformersFullEAP810

We should test against the most recent version with the old management API version.

Same comment about the future benefit of using AbstractSubsystemSchemaTest and @RunWith(Parameterized.class).

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -129,6 +145,59 @@ public void testAsyncIOExpressions() throws Exception {
...
+        testTransformersFull(ModelTestControllerVersion.EAP_8_0_0, MODEL_VERSION_EAP80);
```
bstansberry: ModelTestControllerVersion.EAP_8_1_0

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -129,6 +145,59 @@ public void testAsyncIOExpressions() throws Exception {
...
+    }
+
+    @Test
+    public void testRejectTransformersEAP800() throws Exception {
```
bstansberry: testRejectTransformersEAP810

We should test against the most recent version with the old management API version.

Same comment about the future benefit of using AbstractSubsystemSchemaTest and @RunWith(Parameterized.class).

transactions/src/test/java/org/jboss/as/txn/subsystem/TransactionSubsystemTestCase.java
```diff
@@ -129,6 +145,59 @@ public void testAsyncIOExpressions() throws Exception {
...
+        testRejectTransformers(ModelTestControllerVersion.EAP_8_0_0, MODEL_VERSION_EAP80,
```
bstansberry: ModelTestControllerVersion.EAP_8_1_0

bstansberry (changes requested):
This LGTM except a lot of comments related to constant and method names to reflect the fact that the primary target for mixed domain transformation in EAP is EAP 8.1, not 8.0.

jmfinelli:
> This LGTM except a lot of comments related to constant and method names to reflect the fact that the primary target for mixed domain transformation in EAP is EAP 8.1, not 8.0.

Done! Thanks for your review, @bstansberry! If you agree, I'll squash all commits together

bstansberry (approved):
LGTM @jmfinelli

jmfinelli:
> LGTM @jmfinelli

I will need to update the version of Narayana to include a little improvement that the feature team proposed.

jmfinelli:
Hi @bstansberry, I assume my last update didn’t affect your approval, did it?

testsuite/shared/src/main/java/org/jboss/as/test/shared/ExtendedSnapshotServerSetupTask.java
bstansberry: This class was removed.

docs/src/main/asciidoc/_admin-guide/management-tasks/Suspend,_Resume_and_Graceful_shutdown.adoc
bstansberry: Line was removed

transactions/src/main/java/org/jboss/as/txn/suspend/RecoverySuspendController.java
bstansberry: This now returns a CompletionStage.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemRootResourceDefinition.java
bstansberry: This now uses a RecoveryGracefulShutdown enum.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
bstansberry: Class was removed.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
bstansberry: Class was removed.

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/transaction/TransactionsRecoveryGracefulShutdownExpressionSupportTestCase.java
bstansberry: Class was removed.

bstansberry:
> Hi @bstansberry, I assume my last update didn’t affect your approval, did it?

No, it did not.

jbliznak:
is this waiting for something (review from someone, delete missing-reqs label,..) or how can we move this forward?

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
```diff
@@ -254,4 +254,12 @@ public interface TransactionLogger extends BasicLogger {
 
     @Message(id = 45, value = "Could not register initial reference for InboundTransactionCurrent implementation")
     RuntimeException cannotRegister(@Cause Exception e);
+
+    @LogMessage()
```
rhusar: Technically this is correct, but we always make the level explicit in WF codebase

```
[rhusar@puglia wildfly]$ git grep @LogMessage | grep -v level | wc -l
       0
```

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
```diff
@@ -254,4 +254,12 @@ public interface TransactionLogger extends BasicLogger {
...
+    @Message(id = 46, value = "Transactions subsystem: starting transactions recovery suspension")
+    void scanSuspensionInitiated();
+
+    @LogMessage()
```
rhusar: Technically this is correct, but we always make the level explicit in WF codebase

```
[rhusar@puglia wildfly]$ git grep @LogMessage | grep -v level | wc -l
       0
```

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/transaction/cmt/fail/TransactionFirstPhaseErrorTestCase.java
```diff
@@ -53,15 +55,46 @@ public static Archive<?> createDeployment() {
         .addAsManifestResource(new StringAsset("Dependencies: org.jboss.jboss-transaction-spi \n"), "MANIFEST.MF")
         // grant necessary permissions for -Dsecurity.manager
         .addAsResource(createPermissionsXmlAsset(
-            new PropertyPermission("ts.timeout.factor", "read")), "META-INF/jboss-permissions.xml");
+            new PropertyPermission("ts.timeout.factor", "read"),
+            new PropertyPermission("jboss.home.dir", "read"),
+            // covers the directory itself
+            new FilePermission(System.getProperty("jboss.inst") + "/standalone/data/tx-object-store", "read,delete"),
+            // covers contents under the directory
+            new FilePermission(System.getProperty("jboss.inst") + "/standalone/data/tx-object-store/-", "read,delete")), "META-INF/jboss-permissions.xml");
```
rhusar: This is fishy - this needs to be removed. Deployments should not be having these permissions and it's tainting the test. Also, too much hardcoding.

You can probably fix this by using a ServerSetupTask to do the wiping without having to grant the deployment unnecessary permissions for the sake of a test.

rhusar (changes requested):

rhusar:
> is this waiting for something (review from someone, delete missing-reqs label,..) or how can we move this forward?

@jbliznak Given the extend of the change, I would prefer to see more eyes on this; we cannot put the onus solely on Brian to review everything. Nevertheless, it's always a good thing to have people who requested changes before respon after they were done. Also the PR owner should resolve conversations that were resolved.

rhusar:
@pferraro Can you please review the suspension logic changes you previously requested? Thanks

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/transaction/cmt/fail/TransactionFirstPhaseErrorTestCase.java
jmfinelli: Done

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
jmfinelli: Done

transactions/src/main/java/org/jboss/as/txn/logging/TransactionLogger.java
jmfinelli: Done

jmfinelli:
> Also the PR owner should resolve conversations that were resolved.

I wasn't aware that this is the policy in WildFly. In other projects, the person who reviews the PR is responsible for resolving comments. From my point of view, this makes sense, as the PR author doesn't know whether the fix they proposed to address a comment is satisfactory.

rhusar:
> > Also the PR owner should resolve conversations that were resolved.
> 
> I wasn't aware that this is the policy in WildFly. In other projects, the person who reviews the PR is responsible for resolving comments. From my point of view, this makes sense, as the PR author doesn't know whether the fix they proposed to address a comment is satisfactory.

No, not a policy – I forgot to add the word 'ideally'. Just think about how much work goes into reviewing from reviewers perspective which can easily take hours every day. Anything low-hanging (you already fixed the comment = you can resolve it = I as a reviewer don't have to revisit that) that makes that easier for the reviewer is the step in the right direction.

jmfinelli:
Last commit addresses: [WFLY-19647](https://redhat.atlassian.net/browse/WFLY-19647)
I will squash the two commits together before merging, if that works for you.

transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
```diff
@@ -53,32 +57,39 @@ public class ArjunaRecoveryManagerService implements Service {
...
     private final Supplier<SocketBinding> recoveryBindingSupplier;
     private final Supplier<SocketBinding> statusBindingSupplier;
-    private final Supplier<SuspendController> suspendControllerSupplier;
+    private final Supplier<SuspendableActivityRegistrar> suspendableActivityRegistrarSupplier;
     private final Supplier<ProcessStateNotifier> processStateSupplier;
+    private final Supplier<Executor> executorSupplier;
 
     private RecoveryManagerService recoveryManagerService;
     private RecoverySuspendController recoverySuspendController;
+    private SuspendableActivityRegistration activityRegistration;
```
pferraro: This reference should be volatile (or a final AtomicReference manipulated via methods with volatile semantics) since Service.start(...) and Service.stop(...) will likely be called by different threads.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem70Parser.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.jboss.as.controller.parsing.ParseUtils.missingRequired;
+import static org.jboss.as.controller.parsing.ParseUtils.requireNoContent;
+import static org.jboss.as.controller.parsing.ParseUtils.requireNoNamespaceAttribute;
+import static org.jboss.as.controller.parsing.ParseUtils.unexpectedAttribute;
+
+/**
+ * The {@link org.jboss.staxmapper.XMLElementReader} that handles the version 7.0 of Transaction subsystem xml.
+ */
+class TransactionSubsystem70Parser extends TransactionSubsystem60Parser {
```
pferraro, line 23: Not in scope for this change, but please consider replacing the "versioning via polymorphism" strategy at some point in the future.

By looking at this class, I have no idea which attributes were added and/or removed.

pferraro (changes requested):

transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
jmfinelli: Thanks Paul! 👍

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystem70Parser.java
jmfinelli: Noted. Thanks

transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
```diff
@@ -155,15 +167,15 @@ public void start(final StartContext context) throws StartException {
...
         processStateSupplier.get().addPropertyChangeListener(recoverySuspendController);
-        suspendControllerSupplier.get().registerActivity(recoverySuspendController);
+        activityRegistration.set(suspendableActivityRegistrarSupplier.get().register(recoverySuspendController, SuspendPriority.LAST));
         consumer.accept(recoveryManagerService);
     }
 
     public void stop(final StopContext context) {
         consumer.accept(null);
-        suspendControllerSupplier.get().unRegisterActivity(recoverySuspendController);
+        activityRegistration.get().close();
```
pferraro: Nit:
```suggestion
        activityRegistration.getAndSet(null).close();
```

pferraro (approved):

transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java
jmfinelli: Done! Thanks 🙏

rhusar (approved):
LGTM; my feedback has been address and I don't see any outstanding issues 👍🏼

bstansberry:
Thanks @jmfinelli, @pferraro and @rhusar

rhusar:
@jmfinelli @bstansberry As mentioned on Zulip, this has caused a major regression since TxControl.disable() is called before MSC begins stopping services, which causes `WFLYEE0006: Failed to destroy component instance ... WFTXN0063: Provider created a null transaction`.

Full analysis on https://redhat.atlassian.net/browse/WFLY-22035

I have prepared a fix in the meantime that I will submit, that moves the problematic code to proper MSC phase, but I am not the SME here, so that will need proper review from TX team.

rhusar:
Opened https://github.com/wildfly/wildfly/pull/20210

PR #18959, opened 2025-05-27
[WFLY-20652] Upgrade WildFly to Hibernate ORM 6.6.17

scottmarlow:
https://issues.redhat.com/browse/WFLY-20652

Incorporates https://issues.redhat.com/browse/WFLY-20646


scottmarlow:
> Not sure how the timing is going got work out on this, but hoping that 6.6.17 when released can make it it into the 37 next Beta or Final.
> 
> The problem covered by https://hibernate.atlassian.net/browse/HHH-19476 has been keeping us from upgrading past wildfly 33.0.2

We still need to solve https://issues.redhat.com/browse/WFLY-20572 for WildFly 37 as well so I think this upgrade is fine to merge.

scottmarlow:
> and looks like 6.6.17 just got tagged a few minutes ago too.

Thanks @codylerum!  Just switched the WFLY-20652 issue to upgrade to 6.6.17 instead!

bstansberry:
Thanks @scottmarlow and @codylerum!

PR #18960, opened 2025-05-27
WFLY-20647 SessionManager.getSession(...) should return null if session is invalid or expired.

pferraro:
https://issues.redhat.com/browse/WFLY-20647


bstansberry:
@pferraro I'm closing this as it was part of #18972

PR #18962, opened 2025-05-28
[WFLY-20650] allow configuration of reuse-x-forwarded and rewrite-host in mod_cluster

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-20650
This will require: https://issues.redhat.com/browse/UNDERTOW-2553 https://github.com/undertow-io/undertow/pull/1725

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemModel.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemModel.java
@@ -17,9 +17,10 @@ public enum UndertowSubsystemModel implements SubsystemModel {
     VERSION_11_0_0(11), // WildFly 23-26.x, EAP 7.4.x
     VERSION_12_0_0(12), // WildFly 27
     VERSION_13_0_0(13), // WildFly 28
-    VERSION_14_0_0(14), // WildFly 32-present
+    VERSION_14_0_0(14), // WildFly 32-36
+    VERSION_15_0_0(15), // WildFly 37-present
     ;
-    static final UndertowSubsystemModel CURRENT = VERSION_14_0_0;
+    static final UndertowSubsystemModel CURRENT = VERSION_15_0_0;
 
     private final ModelVersion version;
 
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
@@ -66,10 +66,12 @@ public enum UndertowSubsystemSchema implements PersistentSubsystemSchema<Underto
     VERSION_13_0(13),   // WildFly 27       N.B. There were no schema changes between 12.0 and 13.0!
     VERSION_14_0(14),   // WildFly 28-32
     VERSION_14_0_PREVIEW(14, 0, Stability.PREVIEW),   // WildFly 33-35
-    VERSION_14_0_COMMUNITY(14, 0, Stability.COMMUNITY)   // WildFly 36 - present
+    VERSION_14_0_COMMUNITY(14, 0, Stability.COMMUNITY),   // WildFly 36 - present
+    VERSION_15_0(15),
+    VERSION_15_0_PREVIEW(15, 0, Stability.PREVIEW)
     ;
 
-    static final Map<Stability, UndertowSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_14_0, VERSION_14_0_PREVIEW, VERSION_14_0_COMMUNITY));
+    static final Map<Stability, UndertowSubsystemSchema> CURRENT = Feature.map(EnumSet.of(VERSION_15_0, VERSION_15_0_PREVIEW, VERSION_14_0_COMMUNITY));
     private final VersionedNamespace<IntVersion, UndertowSubsystemSchema> namespace;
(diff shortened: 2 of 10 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemModel.java
```diff
@@ -18,8 +18,9 @@ public enum UndertowSubsystemModel implements SubsystemModel {
     VERSION_12_0_0(12), // WildFly 27
     VERSION_13_0_0(13), // WildFly 28
     VERSION_14_0_0(14), // WildFly 32-present
```
rhusar, line 20: Please update this line as well.

```suggestion
    VERSION_14_0_0(14), // WildFly 32-36
```

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterDefinition.java
```diff
@@ -306,10 +306,25 @@ public RuntimeCapability<Void> getDefinition() {
...
+            .setAllowExpression(true)
+            .setDefaultValue(ModelNode.FALSE)
+            .build();
+
+    public static final AttributeDefinition REWRITE_HOST_HEADER = new SimpleAttributeDefinitionBuilder(Constants.REWRITE_HOST_HEADER, ModelType.BOOLEAN)
+            .setRequired(false)
+            .setRestartAllServices()
+            .setAllowExpression(true)
+            .setDefaultValue(ModelNode.FALSE)
+            .build();
```
rhusar, line 321: Did you intend to add these in `VERSION_15_0_PREVIEW`? Because you are adding them to default.

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
```diff
@@ -66,10 +66,12 @@ public enum UndertowSubsystemSchema implements PersistentSubsystemSchema<Underto
     VERSION_13_0(13),   // WildFly 27       N.B. There were no schema changes between 12.0 and 13.0!
     VERSION_14_0(14),   // WildFly 28-32
     VERSION_14_0_PREVIEW(14, 0, Stability.PREVIEW),   // WildFly 33-35
-    VERSION_14_0_COMMUNITY(14, 0, Stability.COMMUNITY)   // WildFly 36 - present
+    VERSION_14_0_COMMUNITY(14, 0, Stability.COMMUNITY),   // WildFly 36 - present
+    VERSION_15_0(15),
+    VERSION_15_0_PREVIEW(15, 0, Stability.PREVIEW)
```
pferraro, line 71: Are we actually introducing any schema changes between 14.0 and 15.0?
If not, we should not introduce a redundant 15.0 version for Stability.DEFAULT.
If so, we need to introduce a 15.0 for Stability.COMMUNITY as well (as it would include changes introduced in 15.0 Stability.DEFAULT).

pferraro (changes requested):

rhusar:
@baranowb Anyway, perhaps this should have been opened as a draft.

undertow/src/main/java/org/wildfly/extension/undertow/filters/ModClusterDefinition.java
baranowb: oh, right, haste on cleanup it seems.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowSubsystemSchema.java
baranowb: Ok this sends mixed signals. Ive asked about it and minor versions are supposedly frowned upon and version bump is prefered option.

baranowb: hmm, I might be missing something here. Did not consider having a COMMUNITY stability as base as a valid option.
About your question - I asked about such case - ie there is 14.0_PREVIEW already and there is another RFE with PREVIEW target stability. Answer was that 14.1_PREVIEW is frowned upon and version bump is preffered option.

pferraro: Sorry if I was unclear.  My question about had nothing to do with major vs minor versions.

Rather, my question had to do with the 15.0 (DEFAULT stability) schema version being added here.  From a cursory inspection, it seemed liked it was identical to 14.0 version.  Is that the case?
If so, then it is redundant and should be removed.
If not, then you would also need to introduce a COMMUNITY version of this schema that includes these changes as well.

baranowb: np. I most likely just miss something.
yes its identical. Though this might be caused by misconception on my side. My assumption was that versions work more or less like a level and there should equilibrium  - ie if there  is new X+1_COMMUNITY, there should be X+1_DEFAULT at least.  That is version bump happens on default and desired stability for RFE.

baranowb:
Nach, it was just regular case of not being on the same page as to how treat version bumps given different inputs.

rhusar:
The required https://github.com/undertow-io/undertow/pull/1725 is still open. I am converting this to draft as it's not mergeable at this point.

fl4via (approved):

PR #18965, opened 2025-05-29
[WFLY-20664] Prometheus related test cases should be executed only if…

yersan:
… the server supports community stability level

https://issues.redhat.com/browse/WFLY-20664


jasondlee (approved):

bstansberry:
Thanks @yersan and @jasondlee

PR #18982, opened 2025-06-06
[WFLY-19761] Check that multiple MP Config deployments work as expected

kabir:
https://issues.redhat.com/browse/WFLY-19761

Part of the Hackathon exercise

### DIFF
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/multideployment/ConfigResource.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/multideployment/ConfigResource.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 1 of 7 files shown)


testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/multideployment/ConfigResource.java
```diff
@@ -0,0 +1,28 @@
+package org.wildfly.test.integration.microprofile.config.smallrye.multideployment;
```
yersan, line 6: They would need a header with the license

kabir: Weird, I got it to add those but you are right, it is not there. Will add on Monday!

kabir: @yersan Copyright headers added

bstansberry (approved):

bstansberry:
Thanks @kabir and @yersan

PR #18984, opened 2025-06-09
[WFLY-20695] Naming: ServiceBasedNamingStore - don't skip delegating …

tadamski:
…to boundServices if initial lookup fails

https://issues.redhat.com/browse/WFLY-20695

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/ServiceBasedNamingStore.java
+++ b/naming/src/main/java/org/jboss/as/naming/ServiceBasedNamingStore.java
@@ -63,7 +63,12 @@ public class ServiceBasedNamingStore implements NamingStore {
             return new NamingContext(EMPTY_NAME, this, null);
         }
         final ServiceName lookupName = buildServiceName(name);
-        Object obj = lookup(name.toString(), lookupName, dereference);
+        Object obj;
+        try {
+            obj = lookup(name.toString(), lookupName, dereference);
+        } catch (NamingException e) {
+            obj = null;
+        }
         if (obj == null) {
             final ServiceName lower = boundServices.lower(lookupName);
             if (lower != null && lower.isParentOf(lookupName)) {


naming/src/main/java/org/jboss/as/naming/ServiceBasedNamingStore.java
```diff
@@ -63,7 +63,12 @@ public Object lookup(final Name name, boolean dereference) throws NamingExceptio
             return new NamingContext(EMPTY_NAME, this, null);
         }
         final ServiceName lookupName = buildServiceName(name);
-        Object obj = lookup(name.toString(), lookupName, dereference);
+        Object obj;
+        try {
+            obj = lookup(name.toString(), lookupName, dereference);
+        } catch (NamingException e) {
+            obj = null;
```
jbaesner, line 70: Shouldn't there be a TRACE level log statement indicating about this NamingException? This might be useful in case of troubleshooting...

tadamski: hmm, I guess adding it won't hurt

tadamski: otoh, there is no other debug in this class, and hitting this exception is normal execution scenario if lookup has to be delegated further (no indication of the bug) - because of that I guess I would prefer to leave it as it is 
@jbaesner wdyt?

bstansberry:
Hi @tadamski 

Can you add a test that reasonably reproduces the original issue? Ideally somewhere in the testsuite/integration/basic org.jboss.as.test.integration.ee.appclient tree. AIUI something that worked before our EE 10 impl now fails so it seems we need a regression test.

This change appears to break several tests.

I think this method needs quite a bit of commenting, including the catch block in your change. This code is largely unchanged since 2011 and the thinking behind its design is not obvious.

AFAICT the change here is to deal with a case where the initial Service lookup finds a ServiceController but its not started so ServiceController.getValue() throws an ISE. Yes? So the idea is to return either a ResolveResult or a NamingContext because....???

tadamski:
@bstansberry ok, I have a reproducer, should be simple to transform it into the test

emmartins:
I am pretty sure this will break a lot of stuff, can you please add an issue reproducer so I understand what's to fix?

tadamski:
@bstansberry @emmartins @jbaesner  I debugged further into the problem and indeed this solution is invalid. I'm closing this one and will comment in the JIRA.

PR #18987, opened 2025-06-10
[WFLY-20698] Upgrading WildFly HTTP Client to 2.1.1.Final

ropalka:
https://issues.redhat.com/browse/WFLY-20698


bstansberry:
Thanks @ropalka

PR #18991, opened 2025-06-10
[WFLY-20703] Upgrade org.jboss.arquillian:arquillian-testcontainers from 1.0.0.Alpha1 to 1.0.0.Alpha4

bstansberry:
This requires changing 'DockerRequired' to 'TestcontainersRequired'

https://issues.redhat.com/browse/WFLY-20703

### DIFF
--- a/testsuite/shared/pom.xml
+++ b/testsuite/shared/pom.xml
@@ -268,7 +268,7 @@
         <dependency>
             <groupId>org.jboss.arquillian</groupId>
             <artifactId>arquillian-testcontainers</artifactId>
-            <version>1.0.0.Alpha1</version>
+            <version>${version.org.jboss.arquillian.testcontainers}</version>
             <scope>compile</scope>
         </dependency>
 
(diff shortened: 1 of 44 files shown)


testsuite/shared/pom.xml
```diff
@@ -268,7 +268,7 @@
         <dependency>
             <groupId>org.jboss.arquillian</groupId>
             <artifactId>arquillian-testcontainers</artifactId>
-            <version>1.0.0.Alpha1</version>
+			<version>${version.org.jboss.arquillian.testcontainers}</version>
```
jasondlee: Weird formatting. :) Probably doesn't really need another commit unless you just really want to, though.

jasondlee (approved):
One take it or leave it comment. That aside, this looks good to me.

testsuite/shared/pom.xml
bstansberry: tabs must have gotten in there.

If CI goes ok I'll fix it, push and immediately merge.

jasondlee: That works!

bstansberry:
Thanks @jasondlee

PR #18993, opened 2025-06-11
[WFLY-20708] Suppress historic XStream CVEs not a part of artemis-wildfly-integration.jar

darranl:
https://issues.redhat.com/browse/WFLY-20708


darranl:
@ehsavoie Could you please take a look at these?

The dependency check report that let to me supressing these is attached to the Jira issue, I don't believe this module does shade in XStream so this is a false match and these reported CVEs can be ignored.

ehsavoie (approved):
All the related issues are for XStream which we don't ship

darranl:
Thank you @ehsavoie

PR #19002, opened 2025-06-12
WFLY-20715 Upgrade wildfly-clustering to 7.0.5.Final

pferraro:
WFLY-20716 Upgrade Infinispan to 15.2.4.Final
WFLY-20710 Auto-enable transport-based failure detection if protocol stack does not contain a socket-based failure detection protocol.

https://issues.redhat.com/browse/WFLY-20715
https://issues.redhat.com/browse/WFLY-20716
https://issues.redhat.com/browse/WFLY-20710


ee-feature-pack/galleon-shared/src/main/resources/feature_groups/jgroups-sockets.xml
```diff
@@ -29,7 +24,7 @@
         <param name="multicast-port" value="45688"/>
     </feature>
     <feature spec="socket-binding-group.socket-binding">
-        <param name="socket-binding" value="jgroups-udp-fd"/>
+        <param name="socket-binding" value="jgroups-fd"/>
```
bstansberry: @pferraro  FYI I added this to the 'incompatibilities' section of https://github.com/wildfly/wildfly.org/issues/770 (my notes-to-self for the WF 37 release announcement).

@jfdenise Any concerns about this for the cloud feature pack?

rhusar (changes requested):
The method in `org.jboss.as.test.integration.domain.mixed.eap740.DomainAdjuster740#adjustJGroups` needs to be adjusted with the FD_SOCK2 removal in `ee-feature-pack/galleon-shared/src/main/resources/feature_groups/jgroups.xml`.

ee-feature-pack/galleon-shared/src/main/resources/feature_groups/jgroups-sockets.xml
pferraro: @bstansberry Is this an incompatibility?  The name of the socket binding is arbitrary.  A server configuration from WF36 should work in WF37 without issue.

pferraro: That I renamed "jgroups-udp-fd" to "jgroups-fd" was purely cosmetic. If this introduces an incompatibility, I can revert it.

bstansberry: @pferraro It can break CLI scripts, yaml overrides etc that assume the existing structure. With Galleon we're increasingly moving users toward a provisioning workflow of 'Galleon generates the stock config; you use CLI or yaml to tweak'.

I don't think it's a problem; I'm just trying to mention things like that. It's impossible to catch them all, and probably not worthwhile. A socket-binding though is more likely the kind of thing that people might be tweaking via CLI scripts or yaml.

pferraro:
> The method in `org.jboss.as.test.integration.domain.mixed.eap740.DomainAdjuster740#adjustJGroups` needs to be adjusted with the FD_SOCK2 removal in `ee-feature-pack/galleon-shared/src/main/resources/feature_groups/jgroups.xml`.

Right.  Good catch.

ee-feature-pack/galleon-shared/src/main/resources/feature_groups/jgroups-sockets.xml
bstansberry: @pferraro I was writing my last comment when you posted yours so I didn't see it. 

I'd note this change in the release announcement regardless, due to the removal of jgroups-tcp-fd. I don't have strong feelings about calling it jgroups-udp-fd vs jgroups-fd. I can vaguely imagine somehow the latter is better for the long term; if so fine. You say 'purely cosmetic' though, so if it's just the 4 extra chars, there are benefits to not changing.

rhusar: @pferraro I also believe this change should be reverted, while I do appreciate the intention to clean this up, this will be problematic for script incompatibility - since the scripts aren't versioned... We have always been strong proponents of using CLI scripts for tuning profiles and this is one of the use cases that I think is relatively common: elect the stack to use, remove the other one and it's bindings that would no longer be used.

pferraro: > this will be problematic for script incompatibility - since the scripts aren't versioned.

Yes, that was Brian's argument above.  Just to say, the removal of FD_SOCK2 and the obsolete jgroups-tcp-fd socket binding already introduces the potential of script compability breakage.

> and this is one of the use cases that I think is relatively common: elect the stack to use, remove the other one and it's bindings that would no longer be used.

In my experience (reviewing user configuration files), this is not actually all that common.  There is very little cost to an unused protocol stack.

rhusar: Which reminds me, that we don't remove those in our test suite scripts, we might as well https://github.com/wildfly/wildfly/pull/19011 for less potential confusion when debugging.

bstansberry:
@pferraro I'd approve this, but I don't want to seemingly overrule any continuing discussion about the socket binding name. I'm fine with whatever you decide.

I do think this should be in the Beta, so to make @darranl 's life easier please try to get any decision wrapped up tomorrow.

rhusar (changes requested):
Not sure what's changed in the last push (thanks GitHub..) but this now causes an NPE in places so we seem to need to mock the SuspendedBatch to be able to create the Map.Entry, e.g. fixes this:

```
SuspendedBatch suspendedBatch = mock(SuspendedBatch.class);
...
when(batch.suspend()).thenReturn(suspendedBatch);
 ```

pferraro:
> Not sure what's changed in the last push (thanks GitHub..) but this now causes an NPE in places so we seem to need to mock the SuspendedBatch to be able to create the Map.Entry, e.g. fixes this:
> 
> ```
> SuspendedBatch suspendedBatch = mock(SuspendedBatch.class);
> ...
> when(batch.suspend()).thenReturn(suspendedBatch);
> ```

Sorry about that. I ported some batch lifecycle fixes from upstream and forgot to include the corresponding unit test changes.

darranl:
@pferraro / @rhusar Can I check did you reach the agreement on the socket binding name Brian mentioned - are we ready to merge?

rhusar (approved):
Yes, good to go now. Review done.

darranl:
Thank you everyone.

PR #19003, opened 2025-06-12
WFLY-20567 MicroProfile OpenAPI subsystem does not allow multiple deployments to use the same endpoint

pferraro:
https://issues.redhat.com/browse/WFLY-20567

This a reincarnation of https://github.com/wildfly/wildfly/pull/18889

Proposal: https://github.com/wildfly/wildfly-proposals/pull/723


undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
```diff
@@ -34,56 +38,59 @@ class HostDefinition extends SimpleResourceDefinition {
...
+    static final RuntimeCapability<Void> HOST_CAPABILITY = RuntimeCapability.Builder.of(Host.SERVICE_DESCRIPTOR).build();
 
     static final StringListAttributeDefinition ALIAS = new StringListAttributeDefinition.Builder(Constants.ALIAS)
             .setRequired(false)
-            .setFlags(AttributeAccess.Flag.RESTART_ALL_SERVICES)
             .setElementValidator(new StringLengthValidator(1))
             .setAllowExpression(true)
             .setAttributeParser(AttributeParser.COMMA_DELIMITED_STRING_LIST)
             .setAttributeMarshaller(AttributeMarshaller.COMMA_STRING_LIST)
+            .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
```
bstansberry: This means an API version bump.

It should also involve a kind of reject transformer, since if the request header is set the current version server will restart the services while the legacy server will go into reload-required. But AFAICT the transformer API doesn't provides facilities for inspecting request headers.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc
```diff
@@ -67,8 +115,8 @@ The MicroProfile OpenAPI specification defines an HTTP endpoint that serves an O
 The OpenAPI endpoint is registered using the configured path (e.g. http://localhost:8080/openapi) local to the root of the host associated with a given deployment.
 
 [NOTE]
-Currently, the OpenAPI endpoint for a given virtual host can only document a single JAX-RS deployment.
-To use OpenAPI with multiple JAX-RS deployments registered with different context paths on the same virtual host, each deployment should use a distinct endpoint path.
+When the OpenAPI endpoint for a given host documents _a single_ JAX-RS deployment, the title and description of the OpenAPI document are obtained from the <display-name/> and <description/> elements, respectively, from the web application deployment descriptor (e.g. web.xml), if otherwise unspecified.
+When the OpenAPI endpoint for a given host documents _multiple_ JAX-RS deployments, any discordant 'info' or 'externalDocs' OpenAPI document properties will be sourced from the MicroProfile config properties prefixed by 'mp.openapi.extensions._server-name_._host-name_'.
```
bstansberry: AIUI these are new MP Config keys which makes this a Feature not an Enhancement.

Is mp.openapi.extensions._server-name_._host-name_ sufficiently namespaced? If smallrye or some other/future thing we integrate uses extension keys we can easily have conflicts between user defined servers names and keys used by our integrated libraries.

pferraro: > AIUI these are new MP Config keys which makes this a Feature not an Enhancement.

Yes - these are new.

> Is mp.openapi.extensions.server-name.host-name sufficiently namespaced? If smallrye or some other/future thing we integrate uses extension keys we can easily have conflicts between user defined servers names and keys used by our integrated libraries.

What would it mean to be "sufficiently namespaced"?  SmallRye extensions are prefixed with `mp.openapi.extensions.smallrye.`, so, yes, if the name of the target server is "smallrye", and the name of the target host aligns with some smallrye extension name key, and the other child names also align, then there would be a conflict.  As soon as you introduce dynamic config property names, this will always be a possibility.  One possible/easy solution would be to reject "smallrye" as a viable server name before attempting to process such properties - and require any "other/future thing we integrate" to be similarly namespace (and handled like above).

undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java
pferraro: Unless the operation headers are stripped from the original operation before the transformation descriptions are applied (maybe @kabir knows for sure?), could this not be achieved via something like:
```java
builder.setCustomerOperationTransformer(new OperationTransformer() {
    @Override
    public TransformedOperation transformOperation(TransformationContext context, PathAddress address, ModelNode operation) throws OperationFailedException {
        ModelNode transformedOperation = operation.clone();
        if (operation.hasDefined(ModelDescriptionConstants.OPERATION_HEADERS, ModelDescriptionConstants.ALLOW_RESOURCE_SERVICE_RESTART)) {
            if (operation.get(ModelDescriptionConstants.OPERATION_HEADERS, ModelDescriptionConstants.ALLOW_RESOURCE_SERVICE_RESTART).asBoolean()) {
                transformedOperation.get(ModelDescriptionConstants.OPERATION_HEADERS).remove(ModelDescriptionConstants.ALLOW_RESOURCE_SERVICE_RESTART);
            }
        }
        return new TransformedOperation(operation, OperationResultTransformer.ORIGINAL_RESULT);
    }
});

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc (hunk 2 above)
Comment (bstansberry) [reply docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc]: @pferraro Why not 'mp.openapi.extensions.wildfly'? We use the namespace that we reasonably should control, the same as smallrye does.

I assume that was the design intent at MP OpenAPI.

File: undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java (hunk 1 above)
Comment (bstansberry) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: That might work. Both write-attribute and undefine-attribute would need to be covered. The transformation APIs have different stuff for resources vs operations, but the resource stuff may just be about initial config transfer.

How hard is to just not do this? It's a big waste of time trying to work around this behavior change.

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc (hunk 2 above)
Comment (pferraro) [reply docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc]: > Why not 'mp.openapi.extensions.wildfly'?

1. The current naming scheme is what I inherited from the prototype integration created by @MikeEdgar
2. The spec suggests that the vendor (i.e. WF) has full control over the "mp.openapi.extensions." prefix:

> Vendors that wish to provide vendor-specific configuration via MP Config (instead of another native configuration framework) must use the prefix mp.openapi.extensions.

I can certainly move the new configuration properties under mp.openapi.extensions.wildfly.*, though that suggests re-locating existing WF-specific configuration properties (which we can do by suppling default values that look for the old property names), but, AFAIR, MP config does not have a deprecation mechanism for property names.

File: undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java (hunk 1 above)
Comment (pferraro) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: > Both write-attribute and undefine-attribute would need to be covered.

Naturally.

> How hard is to just not do this?

I requires no work to not do this :)

The point was that write-attribute operations for these attributes do _not_ require restarting all services.
For resources that install on-demand services (like this one), requiring a full restart is overkill, and often undesirable.

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc (hunk 2 above)
Comment (bstansberry) [reply docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc]: Ok; good point; we are already in this situation, and this is just a matter of not using further initial elements that may conflict with a reasonable server-name. I don't see 'smallrye' as a likely issue.

File: undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java (hunk 1 above)
Comment (bstansberry) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: What concerns me about all this is every refactor to these APIs, to be done correctly, requires new API versions and transformers. All to support a use case that IMHO is an anti-pattern in a production system.

I'm trying to understand how hard it would be to use these APIs without introducing this behavior change, so we can make informed cost-benefit decisions. By "not do this" I mean "not change the semantic of write-attribute, remove, etc".

Comment (pferraro) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: Oh, sorry if that wasn't clear.  There is nothing in wildfly-subsystem that mandates the use of AttributeAccess.Flag.RESTART_RESOURCE_SERVICES.  These attributes can keep using RESTART_ALL_SERVICES - and the operation handlers in wildfly-subsystem will respect this.

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc (hunk 2 above)
Comment (pferraro) [reply docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc]: OK, I see your point.  Perhaps we use something like:
mp.openapi.extensions.server._server-name_.host._host-name_.foo.bar

File: undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java (hunk 1 above)
Comment (bstansberry) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: oh; perfect! Thanks; I should have figured that out.

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc (hunk 2 above)
Comment (pferraro) [reply docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc]: Fixed.

File: undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java (hunk 1 above)
Comment (pferraro) [reply undertow/src/main/java/org/wildfly/extension/undertow/HostDefinition.java]: Reverted to old behavior.

File: docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc [hunk 3]
```diff
@@ -67,8 +115,8 @@ The MicroProfile OpenAPI specification defines an HTTP endpoint that serves an O
 The OpenAPI endpoint is registered using the configured path (e.g. http://localhost:8080/openapi) local to the root of the host associated with a given deployment.
 
 [NOTE]
-Currently, the OpenAPI endpoint for a given virtual host can only document a single JAX-RS deployment.
-To use OpenAPI with multiple JAX-RS deployments registered with different context paths on the same virtual host, each deployment should use a distinct endpoint path.
+When the OpenAPI endpoint for a given host documents _a single_ JAX-RS deployment, the title and description of the OpenAPI document are obtained from the <display-name/> and <description/> elements, respectively, from the web application deployment descriptor (e.g. web.xml), if otherwise unspecified.
```
fabiobrz, line 169: @pferraro wouldn't it be good to have a quick example block here? Something showing how a few global and per-deployment related properties would be sourced into the final OpenAPI document, as you've done for the supported MicroProfile Config properties above, and for the JSON format below.

Maybe just improving such existing blocks and their description would be enough.

I think this would be really helpful for a user to immediately grab what the output would be.

pferraro: Good idea. Will add.

pferraro: Fixed.

pferraro:
Rebased against main.

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_OpenAPI.adoc
fabiobrz: Thanks!

fabiobrz: Although I can only see a more detailed _Note_ while nothing was added to the properties/expected output code block examples. Is this intentional @pferraro ?

pferraro: The note includes an example of a globally provided OpenAPI property (that should apply to all applications), whereas the sample /META-INF/microprofile-config.properties only include properties specific to the application.  That this would be provided by the same config source is intentional.

fabiobrz: Ok, let me provide an example. Two applications are deployed. 
The first one, is using a static file where `info.title` is set to `static-openapi-app`, e.g.:
```
---
openapi: ${version.openapi}
info:
  title: static-openapi-app
  version: "1.0"
  ...
```
The other application is instead leveraging OASFactory to programmatically set the title to `dynamic-openapi-app`, e.g.:
```
public class OpenApiModelReader implements OASModelReader {
    @Override
    public OpenAPI buildModel() {
        return OASFactory.createObject(Info.class).title("dynamic-openapi-app");
```

If I understand correctly, in this example `info.title` is what the community proposal calls a "global" discordant OpenAPI property.

According to the community documentation in this PR, such property will then be sourced from mp.openapi.extensions.server._server-name_.host._host-name_.info.title which, in order to avoid additional conflicts, should be defined at a global level from a MicroProfile Config source perspective, e.g.: via a system property. Is this correct?

In such a case, the documentation note is a good thing, since it emphasizes that using the microprofile-config.properties file is not a way to provide such information.

As an additional question, what would happen if instead both the deployments would define two, again discordant, values for mp.openapi.extensions.server._server-name_.host._host-name_.info.title?

This is why I originally asked for adding something more  to the example code in the documentation, although it is not strictly needed, provided that we clarify the behavior with respect to the above questions.

pferraro: > If I understand correctly, in this example info.title is what the community proposal calls a "global" discordant OpenAPI property.

Yes.

> According to the community documentation in this PR, such property will then be sourced from mp.openapi.extensions.server.server-name.host.host-name.info.title which, in order to avoid additional conflicts, should be defined at a global level from a MicroProfile Config source perspective, e.g.: via a system property. Is this correct?

Correct.  If the resulting OpenAPI document for a given virtual host is to contain any info.** or externalDocs.** content, either a given global property must be defined via a global MP Config source, or each application will need to define the same value.
In your above example, assuming both applications use the same virtual host, the resulting OpenAPI document will not contain an info.title.
Instead, if each application provided the same value for info.title, then the resulting OpenAPI document will contain an info.title, using the concordant value.

> As an additional question, what would happen if instead both the deployments would define two, again discordant, values for mp.openapi.extensions.server.server-name.host.host-name.info.title?

Host-specific properties are not sourced from application-specific MP Config sources, only from global MP Config sources.

pferraro: N.B. Ideally, were there such a thing as host-scoped MP Config sources, there would be no need for these custom properties.  Alas, we only have application and global scope, hence the use of server/host property namespaces.

fabiobrz: > In your above example, assuming both applications use the same virtual host, the resulting OpenAPI document will not contain an info.title.
Instead, if each application provided the same value for info.title, then the resulting OpenAPI document will contain an info.title, using the concordant value

Will the server log any messages about that? 
Anyway I think this behavior should definitely be mentioned in the community proposal.

> Ideally, were there such a thing as host-scoped MP Config sources, there would be no need for these custom properties. Alas, we only have application and global scope, hence the use of server/host property namespaces.

This too, is an implementation detail which IMO _could_ be mentioned in the analysis document.

pferraro: > Will the server log any messages about that?

Conflicts are logged during deployment.

microprofile/openapi-smallrye/pom.xml
```diff
@@ -17,119 +19,25 @@
...
-        </dependency>
-        <dependency>
-            <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron-util</artifactId>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
+    <packaging>pom</packaging>
+
+    <name>WildFly: MicroProfile modules</name>
```
fabiobrz: s/MicroProfile modules/MicroProfile OpenAPI modules/ ?

fabiobrz (commented):
Just a minor comment as of now @pferraro - I'll review again tomorrow.

microprofile/openapi-smallrye/pom.xml
pferraro: Fixed.

microprofile/openapi-smallrye/spi/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-microprofile-openapi-smallrye</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>38.0.0.Beta1-SNAPSHOT</version>
```
rhusar:
```suggestion
        <version>38.0.0.Final-SNAPSHOT</version>
```

pferraro: Oops.

microprofile/openapi-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/openapi/deployment/OpenAPIDocumentProcessor.java
```diff
@@ -0,0 +1,185 @@
...
+                                    return registration;
+                                }
+                            });
+                            ServiceInstaller.builder(registration)
+                                    .startWhen(StartWhen.INSTALLED)
+                                    .onStop(OpenAPIRegistry.Registration::close)
+                                    .build()
+                                    .install(context);
+                        }
+                    } catch (DuplicateServiceException e) {
```
fabiobrz, line 156: Does this case still need to be handled? Multiple deployments using the same endpoint are now a valid use case. See the comment below

microprofile/openapi-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/openapi/logging/MicroProfileOpenAPILogger.java
```diff
@@ -57,4 +58,8 @@ public interface MicroProfileOpenAPILogger extends BasicLogger {
     @LogMessage(level = INFO)
     @Message(id = 8, value = "MicroProfile OpenAPI documentation disabled for '%s'")
     void disabled(String deploymentName);
+
+    @LogMessage(level = WARN)
+    @Message(id = 9, value = "Deployment-specific property values for %s will be ignored due to conflicts: %s")
```
fabiobrz: Is:

```
@Message(id = 3, value = "MicroProfile OpenAPI endpoint already registered for host '%s'. Skipping OpenAPI documentation of '%s'.")
```
still used since we introduce support for multiple deployments?

fabiobrz (commented):
@pferraro hi. I reviewed some parts while testing and dropped a couple of comments.

microprofile/openapi-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/openapi/logging/MicroProfileOpenAPILogger.java
pferraro: Yes - see above.

microprofile/openapi-smallrye/extension/src/main/java/org/wildfly/extension/microprofile/openapi/deployment/OpenAPIDocumentProcessor.java
pferraro: Yes - though it only applies to the "non-standard endpoint" use case.

fabiobrz: @pferraro, do I understand correctly that if a non-standard endpoint is set by multiple applications, then OpenAPI generation would be happening just for one of them, while the others would be skipped? 
If so, this seems to be clashing with the "supporting multiple MicroProfile applications" requirement, right?

pferraro: Not quite.  Multiple application can use separate non-standard endpoints, same as they can do now.

pferraro: However, if 2 applications try to use the same non-standard endpoint, then the second will fail to deploy.

fabiobrz: > if 2 applications try to use the same non-standard endpoint, then the second will fail to deploy

@pferraro That is what I meant, which seems weird since it defines a difference on behavior between applications using the "standard" endpoint, in which case multiple deployments are supported, and those which don't.

BTW, when you say:
>  will fail to deploy

do you mean actually a deployment failure, or just a warning that the deployment won't be scanned to generate OpenAPI documentation? 

IIUC and if the former case is true, then we have incompatibility with previous version of WildFLy and JBoss EAP XP too, which would require proper documentation, at least.

pferraro: > That is what I meant, which seems weird since it defines a difference on behavior between applications using the "standard" endpoint, in which case multiple deployments are supported, and those which don't.

It is less weird once you consider that the only reason we introduced a "non-standard" endpoints option was to enable the use of MP Open API by multiple deployments.
Non-standard endpoints are not defined by the spec and thus are not in scope for this feature.

> do you mean actually a deployment failure, or just a warning that the deployment won't be scanned to generate OpenAPI documentation?

Sorry, I meant "fail to deploy the OpenAPI model", which results in a log warning.

fabiobrz (approved):

darranl:
@pferraro The downstream issue for this is pre-checked, I don't think there is anything to stop this going to WildFly now?

rhusar:
> @pferraro The downstream issue for this is pre-checked, I don't think there is anything to stop this going to WildFly now?

That's my understanding as well plus some approvals and needs a fresh CI runs.

pferraro:
> The downstream issue for this is pre-checked, I don't think there is anything to stop this going to WildFly now?

Agreed.

rhusar (approved):
Our bot pinged my for the review of the modcluster parts, and those LGTM

bstansberry:
I'm requested for a review on this but it's far beyond what I can look at in detail. I'm happy to defer to the feature team on this, so +1 to getting merging it. My fundamental point above that this is a feature has obviously been addressed, as has the point about the namespacing of the new MP Config properties.

fabiobrz:
From the feature team perspective, and specifically about tests and MicroProfile OpenAPI extension, I've already reviewed the changes and approved them.

PR #19006, opened 2025-06-12
[WFLY-20717] Remove type checks for legacy JBoss Threads classes that…

bstansberry:
… are never injected

https://issues.redhat.com/browse/WFLY-20717


bstansberry:
@tadamski Please review. I missed this one before. This blocks https://github.com/wildfly/wildfly-core/pull/6440, which I'd like to get in pretty soon so it gets bake before WF 37 Beta.

tadamski (approved):
LGTM

bstansberry:
Thanks @tadamski

PR #19008, opened 2025-06-13
[WFLY-20682 & WFLY-20718] Remove exported module dependencies in favor of adding dependencies in a DUP

jamezp:
Issues:
- https://issues.redhat.com/browse/WFLY-20682
- https://issues.redhat.com/browse/WFLY-20718

The first commit simply removes the support for Jackson 1 annotations which is no longer supported.

The second commit removes the exported dependencies and replaces them with added dependencies in a DUP.


bstansberry:
@jamezp Is it the case that the first commit isn't removing any support? IOW the support is already gone in RESTEasy, so this is just cleaning out cruft?

jamezp:
> @jamezp Is it the case that the first commit isn't removing any support? IOW the support is already gone in RESTEasy, so this is just cleaning out cruft?

@bstansberry Correct. It enabled support for Jackson as the JSON binding serlializer based on Jackson 1 annotations or types. We no longer support or ship Jackson 1.

bstansberry (approved):

galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/lra-coordinator/main/module.xml
```diff
@@ -20,6 +20,7 @@
     <module name="org.jboss.logging" />
     <module name="org.jboss.narayana.lra.lra-service-base" />
     <!-- lra-coordinator-jar -->
+    <module name="com.fasterxml.jackson.core.jackson-core"/>
```
bstansberry, line 23: @jamezp Should this be moved to the section below (not shown in this diff, but with a "category" comment explaining the use case)?

jamezp: @bstansberry I put it here because it's needed by the LRA coordinator https://github.com/jbosstm/lra/blob/61e9ed35c7c66acad4af11092476000a210fcca9/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java#L29-L30. I'm happy to move elsewhere though.

bstansberry:
Thanks @jamezp

PR #19010, opened 2025-06-16
[WFLY-20723] Json processing failed with error java.lang.UnsupportedOperationException at java.sql/java.sql.Date.toInstant

lvydra:
https://issues.redhat.com/browse/WFLY-20723

Missing java.sql dependency results in Yasson SqlDateSerializer not being initialized and fallback to DateSerializer, which fails with java.lang.UnsupportedOperationException at java.sql/java.sql.Date.toInstant

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/eclipse/yasson/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/eclipse/yasson/main/module.xml
@@ -17,6 +17,10 @@
 
     <dependencies>
         <module name="java.logging"/>
+        <module name="java.desktop"/>
+        <module name="java.naming"/>
+        <module name="java.sql"/>
+        <module name="java.xml"/>
         <module name="jakarta.enterprise.api"/>
         <module name="jakarta.json.api"/>
         <module name="jakarta.json.bind.api"/>


bstansberry:
@lvydra @jamezp 

The module-info.class for this artifact shows other things that are not present in module.xml:

```
module org.eclipse.yasson {
    requires jakarta.json;
    requires jakarta.json.bind;
    requires java.logging;
    requires static java.xml;
    requires static java.naming;
    requires static java.sql;
    requires static java.desktop;
    requires static jakarta.cdi;

    exports org.eclipse.yasson;
    exports org.eclipse.yasson.spi;
    provides jakarta.json.bind.spi.JsonbProvider with org.eclipse.yasson.JsonBindingProvider;
    uses org.eclipse.yasson.spi.JsonbComponentInstanceCreator;
}
```

bstansberry:
The java.naming dep allows it to use JNDI to find java:comp/BeanManager if it can't find it using CDI. I don't think we need this, but OTOH the KISS / more robust approach is to match our module.xml with module-info.class

The java.desktop dep allows use of the java.beans.ConstructorProperties annotation, which seems relevant.

The java.xml dep allows use of javax.xml.datatype.XMLGregorianCalendar which seems relevant.

lvydra:
Thanks @bstansberry, I have added mentioned dependencies.

jamezp (approved):

PR #19011, opened 2025-06-16
Clustering test suite: remove associated socket-bindings when removin…

rhusar:
…g the unused stack.

No Jira required.


rhusar:
Marking as draft so this doesn't cause conflict with https://github.com/wildfly/wildfly/pull/19002.

PR #19017, opened 2025-06-17
WFLY-20722 log that the app client persistence.xml will not be deployed in server mode

scottmarlow:
https://issues.redhat.com/browse/WFLY-20722

Log that the persistence.xml contained in the application client archive will not be deployed by the WildFly server:
> 2025-06-17 15:12:22,475 INFO  [org.jboss.as.jpa] (MSC service thread 1-8) WFLYJPA0076: persistence.xml in application client database-ejb-1.0.12-SNAPSHOT.jar deployment will not be deployed in server mode


bstansberry (approved):

scottmarlow:
It doesn't look like the 3 failing checks are caused by this pull request change.

PR #19028, opened 2025-06-20
[WFLY-18960] StatefulTimeoutTestCase: make sfsb timeout higher becaus…

tadamski:
…e of intermittent test failures

https://issues.redhat.com/browse/WFLY-18960

@pferraro please review

I can reproduce this very intermittent error by dividing this timeout by two, so maybe when we multiply it by two it will dissappear forever


pferraro (approved):
This fix is fine.  Ideally, the @StatefulTimeout value could be scaled to account for slow environments, but the intention here is to validate annotation-based configuration which cannot be made dynamic.

rhusar (approved):

bstansberry:
Thanks @tadamski @pferraro and @rhusar

PR #19034, opened 2025-06-24
[WFLY-16968]: WildFly High Availability Guide: Messaging write up

ehsavoie:
Adding some initial documentation on HA with Artemis. Based on Artemis documentation

Jira: https://issues.redhat.com/browse/WFLY-16968
      https://issues.redhat.com/browse/WFLY-20749

### DIFF
--- a/docs/src/main/asciidoc/_high-availability/Messaging.adoc
+++ b/docs/src/main/asciidoc/_high-availability/Messaging.adoc
@@ -1,4 +1,6 @@
+[[Messaging]]
 = Messaging
+:experimental:
 
 ifdef::env-github[]
 :tip-caption: :bulb:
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,460 @@ ifdef::env-github[]
...
+
+Both primary & secondary servers must configure the location of journal directories to the _same shared location_ (as explained in https://activemq.apache.org/components/artemis/documentation/latest/persistence.html#persistence[persistence documentation]).
+
+===== Primary
+
+The primary broker needs this basic configuration in `standalone.xml`:
+
+[,xml]
+----
+<shared-store-primary >
```
rhusar: Extra space here `>`. Is this supposed to be also closed `\>`?

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,460 @@ ifdef::env-github[]
...
+====
+In general, synchronization occurs in parallel with current network traffic so this won't cause any blocking for current clients.
+However, there is a critical moment at the end of this process where the replicating server must complete the synchronization and ensure the secondary acknowledges this completion.
+This exchange between the replicating server and secondary will block any journal related operations.
+The maximum length of time that this exchange will block is controlled by the `initial-replication-sync-timeout` configuration element.
+====
+
+Since replication will create a copy of the data at the secondary then in case of a successful fail-over, the secondary's data will be newer than the primary's data.
+If you configure your secondary to allow failback to the primary then when the primary is restarted it will be passive and the active secondary will synchronize its data with the passive primary before stopping to allow the passive primary to become active again.
+If both servers are shutdown then the administrator will have to determine which one has the latest data.
```
bstansberry: s/shutdown/shut down/g

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,460 @@ ifdef::env-github[]
...
+* `backup1`, `backup2`, `backup3`, `backup4`: with `group-name=fish`
+* `backup5`, `backup6`: with `group-name=bird`
+
+After joining the cluster the backups with `group-name=fish` will search for primary servers with `group-name=fish` to pair with.
+Since there is one secondary too many, the `fish` will remain with one spare secondary.
+
+The 2 backups with `group-name=bird` (`backup5` and `backup6`) will pair with primary servers `primary4` and `primary5`.
+====
+
+If `group-name` is not configured then the secondary will search for any primary that it can find the cluster.
```
bstansberry: s/find the cluster/find in the cluster/g

I think ^^^ is the intended meaning.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,460 @@ ifdef::env-github[]
...
+----
+<replication-primary/>
+----
+
+====== Additional parameters
+
+group-name::
+If set, secondary servers will only pair with primary servers with matching group-name.
+See <<replication-configuration,above>> for more details.
+Valid for both quorum voting and pluggable lock manager.
```
bstansberry: Is "pluggable lock manager" a thing in WildFly?

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,460 @@ ifdef::env-github[]
...
++
+NOTE: during this interval any journal related operations will be blocked.
+
+== Failing Back to Primary Server
+
+After a primary server has failed and a secondary taken has taken over its duties, you may want to restart the primary server and have clients fail back.
+
+=== Failback with Shared Store
+
+In case of shared storage you have a couple of options:
```
bstansberry, line 352: We should recommend one of these.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: Nope, that's something I forget to remove

rhusar (changes requested):

rhusar:
@ehsavoie This does looks like almost a verbatim copy from the  Artemis documentation. Is there any reason why not just link artemis documentation for things that are not WildFly-specific?

ehsavoie:
@rhusar it is , except that I removed most (as I might have missed some) the parts that don't apply to WildFly. The issue is that Artemis documentation targets their own configuration file and has more options that we don't support/expose.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
rhusar: +1 and when to use which.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+
+=== All Shared Store Configuration
+
+==== Primary
+
+The following lists all the `ha-policy` configuration elements for HA strategy shared store for `primary`:
+
+failover-on-shutdown::
+The xref:wildscribe/subsystem/messaging-activemq/server/ha-policy/shared-store-primary/index.html[failover-on-shutdown] attribute describes whether the _graceful_ shutdown of this broker will cause the backup to activate.
+If `false` then the backup server will remain passive if this broker is shutdown gracefully (e.g. using kbd:[Ctrl+C]).
```
rhusar: `kbd` is not a thing. Is this some macro that doesn't translate?

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+If set to false then server startup will be done in the background.
+Default is `true`.
+
+==== Secondary
+
+The following lists all the `ha-policy` configuration elements for HA strategy Shared Store for `secondary`:
+
+failover-on-shutdown::
+The xref:wildscribe/subsystem/messaging-activemq/server/ha-policy/shared-store-primary/index.html[failover-on-shutdown] attribute describes whether the _graceful_ shutdown of this broker will cause the backup to activate.
+If `false` then the backup server will remain passive if this broker is shutdown gracefully (e.g. using kbd:[Ctrl+C]).
```
rhusar, line 424: `kbd` is not a thing. Is this some macro that doesn't translate?

rhusar (changes requested):
Plus still missing shared storage recommendation for users.

rhusar:
> @rhusar it is , except that I removed most (as I might have missed some) the parts that don't apply to WildFly. The issue is that Artemis documentation targets their own configuration file and has more options that we don't support/expose.

@ehsavoie OK That makes sense, to summarize, we can't just reference upstream documentation since there are many idiosyncrasies of using it within WildFly. The downside is having to synchronize those two.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+=== Failback with Replication
+
+As with shared storage the `allow-failback` option can be set in replication configurations.
+
+[,xml]
+----
+<replication-secondary allow-failback="true"/>
+----
+
+If the attribute `check-for-active-server` is set to `true` in the `primary` configuration then during start-up the primary server will first search the cluster for another active server using its nodeID.
```
bstansberry: What's a nodeID?

This is the only use of this string in the PR.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
bstansberry: Perhaps we can't recommend just one, as the choice depends on factors where there isn't really a right or wrong. But we can describe more.

AIUI, the downside of failing back is there is another failover affecting the system/clients. I presume the upside is if the primary is better able to handle the workload than the secondary, switching back to it is a good thing. The better the tradeoff and the fewer the admin resources to manually do things, the better the automated options look.

Beyond that, I think some discussion of the 'kill' option is needed. From a naive POV, that just seems like a bad idea, given there are alternatives that allow failback following a clean shutdown. The fact that it's first on the list with no discussion of pros/cons leaves me wondering what important info I've not been told.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+
+* `primary1`, `primary2`, `primary3`: with `group-name=fish`
+* `primary4`, `primary5`: with `group-name=bird`
+* `backup1`, `backup2`, `backup3`, `backup4`: with `group-name=fish`
+* `backup5`, `backup6`: with `group-name=bird`
+
+After joining the cluster the backups with `group-name=fish` will search for primary servers with `group-name=fish` to pair with.
+Since there is one secondary too many, the `fish` will remain with one spare secondary.
+
+The 2 backups with `group-name=bird` (`backup5` and `backup6`) will pair with primary servers `primary4` and `primary5`.
```
bstansberry, line 260: This examples leaves me with lots of questions. I've not seen anything elsewhere in the docs that provides answers. If I've missed something we should link to it.

This describes, for each group, multiple primary/secondary pairs. Is there anywhere where we document the client semantics of sending and consuming messages in such a setup?

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+As with shared storage the `allow-failback` option can be set in replication configurations.
+
+[,xml]
+----
+<replication-secondary allow-failback="true"/>
+----
+
+If the attribute `check-for-active-server` is set to `true` in the `primary` configuration then during start-up the primary server will first search the cluster for another active server using its nodeID.
+If it finds one it will contact this server and try to "failback".
+Since this is a remote replication scenario the primary will have to synchronize its data with the secondary server running with its ID. Once they are in sync it will request the other server (which it assumes it is a secondary that has assumed its duties) to shutdown in order for it to take over.
```
bstansberry: Another, slightly different reference to an otherwise undescribed ID.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +8,448 @@ ifdef::env-github[]
...
+allow-failback::
+Whether a server will automatically stop when another places a request to take over its place.
+The use case is when the secondary has failed over.
+
+== Scaling Down
+
+An alternative to using primary/secondary groups is to configure _scaledown_.
+When configured for scale down a server can copy all its messages and transaction state to another active server.
+The advantage of this is that you don't need full backups to provide some form of HA, however there are disadvantages with this approach the first being that it only deals with a server being stopped and not a server crash.
+The caveat here is if you configure a secondary to scale down.
```
bstansberry: This is not a complete sentence.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: yes , it needs the experimental header

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: Added

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,470 @@ ifdef::env-github[]
 :warning-caption: :warning:
 endif::[]
 
-This section is under development intending to describe high availability features and configuration pertaining to Jakarta Messaging (JMS).
\ No newline at end of file
+We define high availability (HA) as the _ability for the system to continue functioning after failure of one or more of the servers_.
```
rhusar: We already defined the HA concepts in the first section of the guide. 

> 1.1. What are High Availability services?

This might be a bit confusing to readers to have the definitions restated.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
rhusar: This is now rendering correctly 👍🏼

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,470 @@ ifdef::env-github[]
...
+If set, secondary servers will only pair with primary servers with matching group-name.
+See <<replication-configuration,above>> for more details.
+
+cluster-name::
+Name of the `cluster-connection` to use for replication.
+This setting is only necessary if you configure multiple cluster connections.
+If configured then the connector configuration of the cluster configuration with this name will be used when connecting to the cluster to discover if an active server is already running, see `check-for-active-server`.
+If unset then the default cluster connections configuration is used (i.e. the first one configured).
+
+max-saved-replicated-journals-size::
```
rhusar: This is incorrect.

The attribute name in WF is different - https://github.com/wildfly/wildfly/blob/main/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/ha/HAAttributes.java#L133

It's missing 's'. Is that intentional actually?

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: Ah sorry didn't look at the rest of the 'book'

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,470 @@ ifdef::env-github[]
...
+failover-on-shutdown::
+The xref:wildscribe/subsystem/messaging-activemq/server/ha-policy/shared-store-primary/index.html[failover-on-shutdown] attribute describes whether the _graceful_ shutdown of this broker will cause the backup to activate.
+If `false` then the backup server will remain passive if this broker is shutdown gracefully (e.g. using kbd:[Ctrl+C]) or using the shutdown command with the `jboss-cli`.
++
+If `true` then when this server is stopped the backup will activate.
+
+wait-for-activation::
+If set to true then server startup will wait until it is activated.
+If set to false then server startup will be done in the background.
+Default is `true`.
```
rhusar: Where is this configured? I don't see it anywhere in the code, model or https://docs.wildfly.org/37/wildscribe/subsystem/messaging-activemq/server/index.html

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,470 @@ ifdef::env-github[]
...
+See <<replication-configuration,above>> for more details.
+
+cluster-name::
+Name of the `cluster-connection` to use for replication.
+This setting is only necessary if you configure multiple cluster connections.
+If configured then the connector configuration of the cluster configuration with this name will be used when connecting to the cluster to discover if an active server is already running, see `check-for-active-server`.
+If unset then the default cluster connections configuration is used (i.e. the first one configured).
+
+check-for-live-server::
+Whether to check the cluster for an active server using our own server ID when starting up.
```
rhusar, line 289: As Brian mentioned in the different comment, this is referencing an 'ID' but I don't think we defined it anywhere.

rhusar (changes requested):

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: I think it is a typo in wildfly where we lost the 's' (cf. https://github.com/apache/activemq-artemis/blob/feb92af7cc16f0a240b59678ecacdae25ec493b7/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java#L1811) Fixing it by removing that 's'

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: Each broker has an ID that is attributed on start up

docs/src/main/asciidoc/_high-availability/Messaging.adoc
bstansberry: I still don't see discussion of the client side in this scenario.

I don't regard that as a blocker for merging this, but it should be improved.

Section 4.1 here describes configuring 2 connectors in the subsystem so the client is aware of both the primary and the secondary. But here I think there would need to be 7 connectors for a 'fish' group server and 4 for a 'bird' group one. I can kind of guess that from thinking about it, although I'm not confident I'm correct.

But even if I'm confident about it and I'm correct, I still don't have any sense how the client works. My *guess* is it knows how to connect to all 7 'fish' but when it interacts with the servers it learns it is meant to work with, say, 'primary2' and it's current secondary 'backup4'.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,462 @@ ifdef::env-github[]
...
+----
+
+By default this is set to false.
+
+You can also force the active secondary to shutdown when the primary comes back up allowing the primary to take over automatically by setting the following property in the `standalone.xml` configuration file as follows:
+
+[,xml]
+----
+<shared-store-secondary allow-failback="true"/>
+----
```
bstansberry, line 378: L373-378 are redundant with L356-362.

docs/src/main/asciidoc/_high-availability/Messaging.adoc
```diff
@@ -8,4 +10,462 @@ ifdef::env-github[]
...
+<replication-primary check-for-active-server="true"/>
+----
+
+[WARNING]
+====
+Be aware that if you restart a primary server after failover has occurred then `check-for-active-server` *must* be `true`.
+If not then the primary server will restart and serve the same messages that the secondary has already handled causing duplicates.
+====
+
+=== All Shared Store Configuration
```
bstansberry, line 406: This is section 4.4.3, after 4.4.1 (about shared store) and 4.4.2 (about replication). It would be better as a subsection of 4.4.1, since it's just a further discussion what's said there.

bstansberry (approved):
I still have things I'd like to see improved here, but this provides a lot of info about something that's not covered at all otherwise, and none of my concerns are about this saying something wrong; they are just about more/better explanation. So thanks and +1 for getting this into the WF 38 docs.

bstansberry:
Thanks @ehsavoie and @rhusar

docs/src/main/asciidoc/_high-availability/Messaging.adoc
ehsavoie: I depends a lot on how you access the cluster and how the cluster is set up. A proxy in front of it can have an impact for example. Also the client can discover the topology of the cluster once it connects successfully to a node. So

PR #19038, opened 2025-06-25
[WFLY-16653] Use a state for the LongRunningBatchlet to be thread-saf…

jamezp:
…e. The previous version failed intermittently. Using better response names for the error message as well.

https://issues.redhat.com/browse/WFLY-16653

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/suspend/LongRunningBatchlet.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/batch/suspend/LongRunningBatchlet.java
@@ -5,9 +5,9 @@
 
 package org.jboss.as.test.integration.batch.suspend;
 
-import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


bstansberry:
@jamezp Could you comment on the JIRA why this fixes the current problem? I don't see how this relates to the failures being seen now, shown in https://issues.redhat.com/browse/WFLY-16653?focusedId=23203144&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23203144

This change seems more relevant to the initial bug report description.

I admit I don't really understand this though. A future maintainer might not either, hence the request for a comment to help that person out.

jamezp:
Yes, I'll take a closer look as well because I completely forgot about the other error when I was looking at this. For that reason, I'm going to place a hold on this PR until that is done.

jamezp:
I'm glad you made me look as this wouldn't really fix it. This slightly makes the concurrency better, but it doesn't fix the potential issue. For this reason, I'm simply going to close this.

PR #19039, opened 2025-06-25
[WFLY-20666] Remove wildfly-elytron dependency

theashiot:
https://issues.redhat.com/browse/WFLY-20666

### DIFF
--- a/testsuite/integration/elytron-oidc-client/pom.xml
+++ b/testsuite/integration/elytron-oidc-client/pom.xml
@@ -90,12 +90,7 @@
         </dependency>
         <dependency>
             <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.wildfly.security</groupId>
-            <artifactId>wildfly-elytron-jose-util</artifactId>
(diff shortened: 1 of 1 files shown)


bstansberry:
@theashiot This doesn't compile.

emmartins:
@bstansberry right, seems that we need replacements for this one too

bstansberry:
@theashiot I put the 'hold' label on this to avoid people looking at it. Please remove the label when you update.

bstansberry (changes requested):
@theashiot This isn't a bom or client lib or something where what's listed is relevant to some other use case. So its dependencies should be limited to what's absolutely necessary for the testsuite module to execute correctly. When I played with this I found the only dependencies added here that are needed for this module are:

```
        <dependency>
            <artifactId>wildfly-elytron-http-oidc</artifactId>
            <scope>test</scope>
        </dependency>
```

and

```
        <dependency>
            <artifactId>wildfly-elytron-tests-common</artifactId>
            <version>${version.org.wildfly.security.elytron-tests-common}</version>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
````

theashiot:
Thanks, @bstansberry for looking into this, i've updated the PR!

best,
ashwin

bstansberry (approved):

bstansberry:
Thanks @theashiot

PR #19042, opened 2025-06-26
[WFLY-18585] SimpleTimerMDBTestCase - attempt to deal with intermitte…

tadamski:
…nt failure

https://issues.redhat.com/browse/WFLY-18585

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/AnnotationTimerServiceMDB.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/AnnotationTimerServiceMDB.java
@@ -36,9 +36,9 @@ public class AnnotationTimerServiceMDB implements MessageListener {
         latch.countDown();
     }
 
-    public static boolean awaitTimerCall() {
+    public static boolean awaitTimerCall(int timeout) {
         try {
-            latch.await(2, TimeUnit.SECONDS);
+            latch.await(timeout, TimeUnit.MILLISECONDS);
         } catch (InterruptedException e) {
             throw new RuntimeException(e);
         }
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/SimpleTimerMDBTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/SimpleTimerMDBTestCase.java
@@ -4,10 +4,6 @@
  */
 package org.jboss.as.test.integration.ejb.mdb.timerservice;
 
-import static org.jboss.as.test.shared.PermissionUtils.createPermissionsXmlAsset;
-
-import java.util.PropertyPermission;
-
 import jakarta.jms.Destination;
 import jakarta.jms.Message;
 import jakarta.jms.MessageProducer;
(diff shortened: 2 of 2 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/mdb/timerservice/TimedObjectTimerServiceMDB.java
```diff
@@ -26,7 +26,7 @@ public class TimedObjectTimerServiceMDB implements TimedObject , MessageListener
 
     private static final CountDownLatch latch = new CountDownLatch(1);
 
-    private static boolean timerServiceCalled = false;
+    private static volatile boolean timerServiceCalled = false;
```
bstansberry: The latch should provide a sufficient memory fence such that volatile isn't needed. But it does no harm. My concern is it won't help.

Looking at the discussion on https://github.com/wildfly/wildfly/pull/17263 it seems your original idea of increasing the timeout to 5 seconds got dropped. Using TimeoutUtil.adjust was good best practice, but it has no effect on the results from ci.wildfly.org, as the system props that drive any 'adjustment' are not set there (EAP QE tends to use them, but ci.wildfly.org is the baseline env for tests). So the TeamCity results we see are still based on a 2 second timeout. That seems way too low for sure, since AIUI the test sends a message and then waits for two secs for async processing of the message, call to MDB.onMessage, creating a timer and the timer expiring. 30 seconds seems more reasonable. The test might validly want to validate that ejbTimeout gets called promptly 100 ms after the timer is started, but the test can't really validly test how long it takes from sending the message until MDB.onMessage is called.

tadamski: @bstansberry thanks for the feedback

I had one additional doubt - why always interface test fails and annotation one never - but I think I got solution for that. The sendMessage method triggers both MDBs and testTimedObjectTimeoutMethod is always chosen as the first test (at least on my env) - so if there is a race condition this test will always suffer.

My second doubt is that on my laptop virtually all the tests pass in <200ms time so 2second is order of magnitude bigger. Can those differences be as dramatic in test envs? 

Taking above in the consideration, I can prepare a fix that will allow us to configure timeout value to 30 seconds (maybe I will also merge those tests cause those two sendMessages doesn't seem to make sense to me), but do you think it is reasonable to dowgrade the test from critical status (and potential release blocker) with this modified timeout level? Since this test is very intermittent, I guess it will take weeks to assert whether failure count dropped to 0 after it.

bstansberry: Re the env, ci.wildfly.org runs about 60 vms on 3 hosts, so it's quite different than a local run. And my impression is this isn't a frequent failure.

Re the priority, +1 to a reduction to Major. AFAIK we've not seen any indication of these failures indicating production code problems, while we do have reasonable theories to explain why the test itself isn't robust. You're also working the issue, so to the extent any high priority is a way of saying "pay attention to these intermittent failures", that goal has been achieved.

bstansberry: @tadamski Do you plan further changes or will we just go with what is here?

tadamski: @bstansberry  I have merged the tests to avoid reduntant sendMessage call, increased timeout to 30 seconds without using TimeoutUtil.adjust and reverted adding volatile since it probably won't help anyway.

bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #19054, opened 2025-07-03
[WFLY-20767] DatabaseTimerServiceMultiNodeExecutionDisabledTestCase a…

tadamski:
…nd DatabaseTimerServiceMultiNodeTestCase possible interaction - fix attempt

https://issues.redhat.com/browse/WFLY-20767

Those test failed together intermittently, but this failure is suspicous:
* they both use the same h2 server
* they use the same timer name "timer1"
* 0 invocations of timer1 in first test
* 2 invocations of timer1 in second test (1 expected)
* DatabaseTimerServiceMultiNodeExecutionDisabledTestCase didn't manage to refresh timers

Looks like an interaction between the tests.
In this fix attempt I change the timer name so that it is not shared between the tests and I'm also giving more time for the first test to check the timer.


bstansberry (approved):

bstansberry:
Thanks @tadamski

PR #19064, opened 2025-07-08
[WFLY-20771] org.jboss.as.naming.deployment.ContextNames: make causal…

tadamski:
… exception visible for optional injections

https://issues.redhat.com/browse/WFLY-20771

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
+++ b/naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
@@ -11,9 +11,9 @@ import org.jboss.as.naming.ManagedReferenceFactory;
 import org.jboss.as.naming.NamingContext;
 import org.jboss.as.naming.NamingStore;
 import org.jboss.as.naming.context.external.ExternalContexts;
+import org.jboss.as.naming.logging.NamingLogger;
 import org.jboss.as.server.CurrentServiceContainer;
 import org.jboss.as.server.deployment.DeploymentUnit;
-import org.jboss.as.naming.logging.NamingLogger;
 import org.jboss.msc.inject.InjectionException;
 import org.jboss.msc.inject.Injector;
 import org.jboss.msc.service.ServiceBuilder;
@@ -307,7 +307,7 @@ public class ContextNames {
                                 if(!optional) {
                                     throw NamingLogger.ROOT_LOGGER.resourceLookupForInjectionFailed(getAbsoluteJndiName(), e);
                                 } else {
-                                    NamingLogger.ROOT_LOGGER.tracef(e,"failed to lookup %s", getAbsoluteJndiName());
+                                    NamingLogger.ROOT_LOGGER.failedToLookup(getAbsoluteJndiName());
                                 }
                             }
                             return null;
--- a/naming/src/main/java/org/jboss/as/naming/logging/NamingLogger.java
+++ b/naming/src/main/java/org/jboss/as/naming/logging/NamingLogger.java
@@ -5,6 +5,7 @@
 
 package org.jboss.as.naming.logging;
 
+import static org.jboss.logging.Logger.Level.DEBUG;
 import static org.jboss.logging.Logger.Level.ERROR;
 import static org.jboss.logging.Logger.Level.INFO;
 import static org.jboss.logging.Logger.Level.WARN;
@@ -585,4 +586,8 @@ public interface NamingLogger extends BasicLogger {
 
     @Message(id = 67, value = "Failed to stop remote naming service")
     IllegalStateException failedToStopRemoteNamingService(@Cause Throwable cause);
+
+    @LogMessage(level = DEBUG)
+    @Message(id = 68, value = "Failed to lookup %s")
+    void failedToLookup(String name);
 }


naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
```diff
@@ -307,7 +307,7 @@ public ManagedReference getReference() {
                                 if(!optional) {
                                     throw NamingLogger.ROOT_LOGGER.resourceLookupForInjectionFailed(getAbsoluteJndiName(), e);
                                 } else {
-                                    NamingLogger.ROOT_LOGGER.tracef(e,"failed to lookup %s", getAbsoluteJndiName());
+                                    NamingLogger.ROOT_LOGGER.infof(e,"failed to lookup %s", getAbsoluteJndiName());
```
bstansberry: If this is changed to info it must be handled via a method in NamingLogger.

emmartins:
I would only move it to debug, and improve the message, to something like ("Optional reference %s lookup failed", getAbsoluteJndiName()).

naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
rhusar: If changing this line, run a formatter while at it (missing space between parameters).

And ofc, needs i18n...

rhusar (changes requested):

bstansberry:
Sorry, I thought I posted this yesterday but I didn't hit the button.

I believe the following captures the cases where 'optional' would be true and thus this change would have an effect:

```
Constructor
    LookupInjectionSource(String, boolean)
Usages or usages of base method in Project and Libraries  (6 usages found)
    New instance creation  (5 usages found)
        wildfly-ee  (4 usages found)
            org.jboss.as.ee.component.deployers  (4 usages found)
                ResourceInjectionAnnotationParsingProcessor  (2 usages found)
                    process(DeploymentPhaseContext, EEModuleClassDescription, AnnotationInstance, String, String, InjectionTarget, EEModuleDescription, ...)  (2 usages found)
                        201 valueSource = new LookupInjectionSource(lookup, JAVAX_NAMING_CONTEXT.equals(injectionType));
                        230 final LookupInjectionSource optionalInjection = new LookupInjectionSource(localContextName, true);
                ResourceReferenceProcessor  (2 usages found)
                    getResourceEnvRefEntries(DeploymentUnit, DeploymentDescriptorEnvironment, ClassLoader, DeploymentReflectionIndex, ResourceInjectionTarget)  (1 usage found)
                        91 injectionSource = new LookupInjectionSource(resourceEnvRef.getLookupName(), classType != null && JAVAX_NAMING_CONTEXT.equals(classType.getName()));
                    getResourceRefEntries(DeploymentUnit, DeploymentDescriptorEnvironment, ClassLoader, DeploymentReflectionIndex, ResourceInjectionTarget)  (1 usage found)
                        149 injectionSource = new LookupInjectionSource(resourceRef.getLookupName(), classType != null && JAVAX_NAMING_CONTEXT.equals(classType.getName()));
        wildfly-ejb3  (1 usage found)
            org.jboss.as.ejb3.deployment.processors  (1 usage found)
                EjbResourceInjectionAnnotationProcessor  (1 usage found)
                    createLookup(String, boolean)  (1 usage found)
                        168 return new LookupInjectionSource(localContextName, appclient);
```

This one is hard to reason about as what's going on there is subtle and hard to understand:

230 final LookupInjectionSource optionalInjection = new LookupInjectionSource(localContextName, true);

But anyway, Eduardo is the leads the naming component and he recommends debug.

darranl:
@tadamski FYI we may be able to include in WildFly 39 but I think this is still waiting for a change from info to debug.

emmartins:
@bstansberry IIRC it's optional and thus it will not fail when missing, if it is a java context, a "directory", cause we only have concrete elements in the naming stores, not intermediate contexts; or - which I believe was this use case - when it's an ejb  lookup from an appclient. For the latter I don't know more, just saw the code comment at https://github.com/wildfly/wildfly/blob/1f68b0ad54b7899650a782a8efdde0cec196865c/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EjbResourceInjectionAnnotationProcessor.java#L166

The reason for me to suggest DEBUG is cause we could have apps looking up such "implicit" contexts, for instance to check child entries, and this change would then start trowing exceptions on INFO. IMHO both TRACE and INFO are not good levels for this kind of information, looks a better fit for DEBUG.

naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
tadamski: fixed

tadamski: fixed

tadamski:
@bstansberry @emmartins ok, I have changed the log level to debug

emmartins (approved):

naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
```diff
@@ -307,7 +307,7 @@ public ManagedReference getReference() {
                                 if(!optional) {
                                     throw NamingLogger.ROOT_LOGGER.resourceLookupForInjectionFailed(getAbsoluteJndiName(), e);
                                 } else {
-                                    NamingLogger.ROOT_LOGGER.tracef(e,"failed to lookup %s", getAbsoluteJndiName());
+                                    NamingLogger.ROOT_LOGGER.failedToLookup(getAbsoluteJndiName());
```
rhusar, line 310: While the PR title promises to make the causal exception visible, the new failedToLookup method has no @Cause Throwable parameter  so the exception is silently swallowed! So the change is worsening rather than improving.

rhusar (changes requested):

naming/src/main/java/org/jboss/as/naming/deployment/ContextNames.java
emmartins: Good catch.

PR #19066, opened 2025-07-09
Fix for WFLY-20748, Potential mismatch between provisioning of HAL bits and config enablement of the console

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-20748

* This Fix should depend on the fix WFCORE-7284 but an override of it has been made in this PR (in an additional commit). Such commit will have to be removed when we upgrade wildfly-core.
* Enable the management-console in the web-console layer
* Enable the web-console in all configs that require it.

In the core override part we have: 
* A copy of the management layer that disables the console.
* A new management-console feature-group that enables the console.


bstansberry (approved):

bstansberry:
Thanks @jfdenise

PR #19069, opened 2025-07-10
WFLY-20779 Infinispan subsystem doesn't use the latest available mode…

rhusar:
…l version

https://issues.redhat.com/browse/WFLY-20779

This https://github.com/wildfly/wildfly/blob/37.0.0.Beta1/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemModel.java#L43 needs to use model 20.

Also descriptions are out of date since https://github.com/wildfly/wildfly/blob/36.0.1.Final/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/subsystem/InfinispanSubsystemModel.java#L40-L41 still only used 19.


pferraro (approved):
Good catch.

rhusar:
(CI failures are unrelated)

bstansberry:
Thanks @rhusar

PR #19072, opened 2025-07-10
[WFLY-20754]  change to how Persistence container checks if a transaction is active.

scottmarlow:
https://issues.redhat.com/browse/WFLY-20754

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/transaction/TransactionUtil.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/transaction/TransactionUtil.java
@@ -11,6 +11,7 @@ import static org.jboss.as.jpa.messages.JpaLogger.ROOT_LOGGER;
 import java.security.PrivilegedAction;
 
 import jakarta.persistence.EntityManager;
+import jakarta.transaction.Status;
 import jakarta.transaction.Synchronization;
 import jakarta.transaction.SystemException;
 import jakarta.transaction.Transaction;
@@ -19,7 +20,6 @@ import jakarta.transaction.TransactionSynchronizationRegistry;
 
 import org.jboss.as.jpa.container.ExtendedEntityManager;
 import org.jboss.as.jpa.messages.JpaLogger;
-import org.jboss.tm.TxUtils;
 import org.wildfly.transaction.client.AbstractTransaction;
 import org.wildfly.transaction.client.AssociationListener;
 import org.wildfly.transaction.client.ContextTransactionManager;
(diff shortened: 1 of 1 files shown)


scottmarlow:
Another reference is https://jakarta.ee/specifications/transactions/2.0/apidocs/jakarta/transaction/transactionscoped:

```
Annotation Type TransactionScoped

    [@Retention](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Retention.html?is-external=true)([value](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Retention.html?is-external=true#value--)=[RUNTIME](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/RetentionPolicy.html?is-external=true#RUNTIME))
     [@Target](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Target.html?is-external=true)([value](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Target.html?is-external=true#value--)={[TYPE](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/ElementType.html?is-external=true#TYPE),[METHOD](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/ElementType.html?is-external=true#METHOD),[FIELD](http://docs.oracle.com/javase/7/docs/api/java/lang/annotation/ElementType.html?is-external=true#FIELD)})
     @NormalScope(passivating=true)
    public @interface TransactionScoped

    The jakarta.transaction.TransactionScoped annotation provides the ability to specify a standard CDI scope to define bean instances whose lifecycle is scoped to the currently active Jakarta Transactions transaction. This annotation has no effect on classes which have non-contextual references such those defined as managed beans by the Jakarta EE specification.
    The transaction scope is active when the return from a call to UserTransaction.getStatus or TransactionManager.getStatus is one of the following states:
        Status.STATUS_ACTIVE
        Status.STATUS_MARKED_ROLLBACK
        Status.STATUS_PREPARED
        Status.STATUS_UNKNOWN
        Status.STATUS_PREPARING
        Status.STATUS_COMMITTING
        Status.STATUS_ROLLING_BACK 

    It is not intended that the term "active" as defined here in relation to the TransactionScoped annotation should also apply to its use in relation to transaction context, lifecycle, etc. mentioned elsewhere in this specification. The object with this annotation will be associated with the current active Jakarta Transactions transaction when the object is used. This association must be retained through any transaction suspend or resume calls as well as any Synchronization.beforeCompletion callbacks. Any Synchronization.afterCompletion methods will be invoked in an undefined context. The way in which the Jakarta Transactions transaction is begun and completed (for example via UserTransaction, Transactional interceptor, etc.) is of no consequence. The contextual references used across different Jakarta Transactions transactions are distinct. Refer to the CDI specification for more details on contextual references. A jakarta.enterprise.context.ContextNotActiveException will be thrown if an object with this annotation is used when the transaction context is not active.

    Since:
        JTA 1.2
    Version:
        Jakarta Transactions 2.0
```

bstansberry (approved):

bstansberry:
Thanks @scottmarlow

PR #18299, opened 2024-10-15
[WFLY-19370] refined debugging data collection

rsearls:
https://issues.redhat.com/browse/WFLY-19370

This PR supersedes https://github.com/wildfly/wildfly/pull/18055

PR #19076, opened 2025-07-14
Pull request18999 ee11

bstansberry:
This is an experiment to test a conflicts-resolved variant of #18299 against WildFly Preview, in order to have a look at the interaction of MP 7.1 and the EE 11 Core APIs.

I'm pretty sure when #18299 was 'fresh' that the 1 PR job against WF Preview ran as well as it did against standard WildFly, but that CI history is gone.

### DIFF
--- a/boms/preview-test-expansion/pom.xml
+++ b/boms/preview-test-expansion/pom.xml
@@ -26,10 +26,6 @@
 
     <name>WildFly Preview: Dependency Management (Expansion Test Dependencies)</name>
 
-    <properties>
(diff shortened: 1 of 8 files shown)


bstansberry:
Closing; I got the info I was looking for. The testsuite/integration/expansion and microprofile-tck modules get the same results when using EE 11 as a base as they did in std WF on an EE 10 base. There are KafkaReactiveMessagingAndOtelTestCase failures, but we saw those with std WF as well.

PR #19079, opened 2025-07-15
WFLY-20778 ServletRequestListener events should be guarded by GlobalRequestControllerHandler

pferraro:
https://issues.redhat.com/browse/WFLY-20778

N.B. This change requires moving the ControlPoint logic to the "initial" handler chain, rather than the "outer" handler chain.  This means that any Predicates (used to allow specific requests to be processed, though the server is suspended) will be evaluated before the request is dispatched to a servlet.


asoldano (approved):
The proposed changes look good to me. I've been in touch with @tomjenkinson , who is going to take care of testing this on XTS side.

pferraro:
> That testing did pass, but it seems to be that the thought is that testing
> there is not including testing XTS graceful shutdown (it does run some more
> XTS tests so it's still helpful I would say). I would suggest to reach out
> to @mmusgrov <https://github.com/mmusgrov> as he has been doing some
> research where testing for this feature may be. FYI the WFLY that added XTS
> graceful shutdown into WildFly is
> https://issues.redhat.com/browse/WFLY-7514

There does seem to be some level of test coverage in the xts integration
test suite, which, AFAICT, are all passing.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/suspend/WebSuspendTestCase.java
```diff
@@ -4,113 +4,205 @@
...
-                new PropertyPermission("node0", "read"),
-                // executorService.shutdown() needs the following permission
-                new RuntimePermission("modifyThread"),
-                // ManagementClient needs the following permissions and a dependency on 'org.jboss.remoting3' module
-                new RemotingPermission("createEndpoint"),
-                new RemotingPermission("connect"),
-                // HttpClient needs the following permission
-                new SocketPermission(TestSuiteEnvironment.getServerAddress() + ":" + TestSuiteEnvironment.getHttpPort(), "connect,resolve"),
-                new FilePermission(System.getProperty("jboss.inst") + "/standalone/tmp/auth/*", "read")
-        ), "permissions.xml");
```
bstansberry, line 69: @pferraro Your change to using the MBean nicely eliminated all these, but you still need a createMBeanServer permission, hence the SM CI job failure.

pferraro: Ah, right.  Fixing...

mmusgrov:
> That testing did pass, but it seems to be that the thought is that testing there is not including testing XTS graceful shutdown (it does run some more XTS tests so it's still helpful I would say). I would suggest to reach out to @mmusgrov <https://github.com/mmusgrov> as he has been doing some research where testing for this feature may be. FYI the WFLY that added XTS graceful shutdown into WildFly is https://issues.redhat.com/browse/WFLY-7514
> There does seem to be some level of test coverage in the xts integration test suite, which, AFAICT, are all passing.

The [XTS suspend/resume tests](https://github.com/wildfly/wildfly/blob/main/testsuite/integration/xts/src/test/java/org/jboss/as/test/xts/suspend/AbstractTestCase.java#L102-L132) are in the wildfly integration part of the test suite and they did indeed pass with Tom's PR that included the changes proposed by Paul's PR. 

However, as far I can tell, the tests only verify suspend and resume and don't actually test the shutdown part of suspend and resume. From my reading of the literature people, including the author of those tests and the engineer verifying the fix for [Implement graceful shutdown support for XTS endpoints](https://issues.redhat.com/browse/WFLY-7514), seem to regard suspend and resume and graceful shutdown as being synonymous which I think is wrong. The tests do things like:

```
@Test
public void testCommitAfterSuspend() throws Exception {
    executorService.begin();
    suspendServer();
    executorService.commit();
}
```
But I would have thought `suspendServer();` would need to wait for the transaction to finish in order to test graceful shutdown, in other words the tests seem to be coded incorrectly.

Do people think my analysis/understanding of the XTS tests is flawed?

mmusgrov:
> @Test
> public void testCommitAfterSuspend() throws Exception {
>     executorService.begin();
>     suspendServer();
>     executorService.commit();
> }

I think I misread the test and it does do what we need - it starts a transaction and suspends the server and then checks that existing transactions are allowed to complete when the server is in the suspended state.

@pferraro I am happy that Tom has successfully verified your fix against the XTS subsystem.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/suspend/WebSuspendTestCase.java
pferraro: Fixed. SM-enabled execution should pass now.

bstansberry:
https://ci.wildfly.org/viewLog.html?buildId=511170& is a custom run of Linux - JDK 17 failures are unrelated. A quick history check doesn't show other failures. It uses -Dts.manualmode instead of -DallTests so it should be relatively quick.

The failures are due to an 'address already in use' situation, which doesn't seem relevant to this PR. This PR doesn't touch manualmode tests, so the address in use is unlikely to be some cruft left by the test changes here. The actual production code changes in this PR would not trigger an "address in use failure".

bstansberry (approved):

bstansberry:
Thanks @pferraro @asoldano @tomjenkinson and @mmusgrov

PR #19084, opened 2025-07-18
[WFLY-20796] Upgrade com.google.code.gson:gson to 2.13.1

bstansberry:
https://issues.redhat.com/browse/WFLY-20796

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -345,7 +345,7 @@
         <version.com.github.fge.msg-simple>1.1</version.com.github.fge.msg-simple>
         <version.com.github.luben.zstd-jni>1.5.7-3</version.com.github.luben.zstd-jni>
         <version.com.google.api.grpc>2.0.1</version.com.google.api.grpc>
-        <version.com.google.code.gson>2.8.9</version.com.google.code.gson>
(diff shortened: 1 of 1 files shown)


bstansberry:
@marko-bekhta Any concerns with this? The only use is Hibernate Search. I looked and Hibernate Search 8 uses this version, so I have no concerns there. 7.2.4 though uses 2.11. 2.11.0 is being flagged up by a scanner as tied to https://nvd.nist.gov/vuln/detail/CVE-2025-53864. The relevant issue there was fixed in 2.12. Moving to 2.12 or later should help us avoid users asking about the scanner result.

We could use 2.12 in standard WildFly and 2.13.1 in WF Preview but if it's not a problem to use 2.13.1 everywhere that's cleaner.

marko-bekhta:
Hey @bstansberry, 
I think it should be fine; I've started the test run against the 7.2 branch here https://github.com/hibernate/hibernate-search/pull/4715 just in case. I'll add a comment here once it finishes. 🤞🏻🙂

marko-bekhta (approved):

marko-bekhta:
The tests passed 🟢

bstansberry:
Thanks @marko-bekhta

PR #19088, opened 2025-07-23
[WFLY-19554] Implement EE 11 JPA / CDI integration in WildFly Preview

scottmarlow:
https://issues.redhat.com/browse/WFLY-19554


jpa/subsystem/src/main/java/org/jboss/as/jpa/beanmanager/PersistenceIntegrationWithCDI.java
```diff
@@ -0,0 +1,351 @@
...
+        // * may be duplicated within different EE modules/submodules.  https://github.com/jakartaee/platform/issues/1135 is the
+        // * EE 12 tracker.
+        // Uncomment the following block and update when this TODO is addressed.
+        /**
+         String beanName = persistenceUnitMetadata.getScopedPersistenceUnitName().replace("#",".");
+         JpaLogger.ROOT_LOGGER.infof("Creating EntityManagerFactory CDI bean named %s for accessing persistence unit %s",
+         beanName, persistenceUnitMetadata.getPersistenceUnitName());
+         beanConfigurator.name(beanName);
+         **/
+        beanConfigurator.name(persistenceUnitMetadata.getPersistenceUnitName());
```
scottmarlow: Commenting out the call to **`beanConfigurator.name(persistenceUnitMetadata.getPersistenceUnitName())`** makes the test failures go away.  To be discussed in WildFly dev ml discussion.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
```diff
@@ -17,5 +17,9 @@
         <module name="java.logging"/>
         <!-- PersistenceUnitInfo needs javax.sql.DataSource -->
         <module name="java.sql" export="true"/>
+        <!-- Persistence integration with CDI needs the following three modules --> 
+         <module name="org.jboss.weld.core"/>
+         <module name="org.jboss.weld.spi"/>
+         <module name="org.jboss.weld.api"/>
```
rhusar: Nitpick, please run the formatter here as they are off.

jpa/schemamanagerbeancreator/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-jpa-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>37.0.0.Final-SNAPSHOT</version>
```
rhusar:
```suggestion
        <version>38.0.0.Beta1-SNAPSHOT</version>
```

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -761,4 +761,12 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 76, value = "persistence.xml in application client %s deployment will not be deployed in server mode")
     void ignoreAppclientPersistenceUnitsInServer(String deploymentName);
 
+    @Message(id = 77, value = "Cannot setup Persistence/CDI integration for %s")
+    RuntimeException cannotAddBeans(@Cause ReflectiveOperationException cause, String scopedPersistenceUnitName);
+
+    @Message(id = 78, value = "Cannot setup Persistence/CDI integration for %s")
+    IllegalAccessException classNotFound(@Cause ClassNotFoundException cause, String scopedPersistenceUnitName);
```
rhusar: These two messages have the same text message; they should be more specific.

jpa/subsystem/src/main/java/org/jboss/as/jpa/puparser/Element.java
```diff
@@ -32,7 +32,10 @@ public enum Element {
     PROVIDER("provider"),
     SHAREDCACHEMODE("shared-cache-mode"),
     VALIDATIONMODE("validation-mode"),
-    VERSION("version");
+    VERSION("version"),
+    QUALIFIER("qualifier"),
+    SCOPE( "scope")
```
rhusar: Nitpick: Extra space.

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
scottmarlow: Thanks @rhusar for all of the feedback!  Will address each one!  :-)

scottmarlow: I removed the ReflectiveOperationException method as that wasn't really needed anymore.

jpa/subsystem/src/main/java/org/jboss/as/jpa/puparser/Element.java
scottmarlow: Fixed!

jpa/schemamanagerbeancreator/pom.xml
scottmarlow: Fixed!

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
scottmarlow: Just ran the formatter!  Thank you!

scottmarlow:
I have a local change to address the `Check non-i18n logging / check-logging (pull_request)` failure (just need to remove a commented out line of code.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
```diff
@@ -17,5 +17,9 @@
         <module name="java.logging"/>
         <!-- PersistenceUnitInfo needs javax.sql.DataSource -->
         <module name="java.sql" export="true"/>
+        <!-- Persistence integration with CDI needs the following three modules -->
+        <module name="org.jboss.weld.core"/>
+        <module name="org.jboss.weld.spi"/>
+        <module name="org.jboss.weld.api"/>
```
jamezp: What do we need these 3 modules for? It should likely only need `jakarta.enterprise.api`. It shouldn't need dependencies on the implementation, just the API.

Also, this should be done in Preview.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+        try {
+            addBeans(event);
+        } catch (RuntimeException e) {
+            event.addDefinitionError(e);
+        } finally {
+            copyOnWriteArrayList.clear();
+        }
+    }
+
+    public IntegrationWithCDIBagImpl register(final PersistenceUnitMetadata persistenceUnitMetadata) {
```
jamezp, line 93: Nit, we should add `@Override` here to self-document this is a required method coming from something else.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+    private static final List defaultQualifier = List.of("jakarta.enterprise.inject.Default");
+    private static final String transactionScoped = "jakarta.transaction.TransactionScoped";
+    private static final String applicationScoped = "jakarta.enterprise.context.ApplicationScoped";
+    protected static final String dependentScoped = "jakarta.enterprise.context.Dependent";
+    private static final SchemaManagerBeanCreator schemaManagerBeanCreator;
+
+    static {
+        Class schemaManagerCreatorClass1;
+        try {
+            schemaManagerCreatorClass1 = IntegratePersistenceAfterBeanDiscovery.class.getClassLoader().loadClass("org.jboss.as.jpa.beanmanager.Persistence32");
```
jamezp: Is there a reason we're using a hard-coded class name instead of something like a `ServiceLoader`?

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+        } catch (ClassNotFoundException ignore) {
+            schemaManagerCreatorClass1 = null;
+        }
+        schemaManagerBeanCreator = SchemaManagerBeanCreator.getImplementation(schemaManagerCreatorClass1);
+    }
+
+    public void addBeans(AfterBeanDiscovery afterBeanDiscovery) {
```
jamezp, line 109: Same note on the `@Override` here.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+        if (scope == null || scope.isEmpty()) {
+            scope = transactionScoped;
+        }
+
+        // setup
+        BeanConfigurator<EntityManager> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(EntityManager.class);
+
+
+        Class<? extends Annotation> scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
```
jamezp: I think we should create a little utility for this which would throw a good exception if the class is not found or something like that. Not critical, but just a thought.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> entityManagerClass = EntityManager.class;
+        beanConfigurator.beanClass(entityManagerClass);
+        beanConfigurator.produceWith(c -> {
+                    return new TransactionScopedEntityManager(
```
jamezp, line 175: Is this right even if the requested scope is not `@TransactionScoped`? Reading the JavaDoc it seems like maybe it's okay, but I just want to verify. Looking at http://github.com/wildfly/wildfly/blob/d1b6dd703da585d60d174885ab55b9e39d422c88/weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java#L55-L84 it handles it differently and that is what we effectively want to mimic.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,370 @@
...
+        String scope = applicationScoped;
+
+        // EntityManagerFactory setup
+        BeanConfigurator<EntityManagerFactory> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(EntityManagerFactory.class);
+
+        // Ensure that we do not get "WELD-001414: Bean name is ambiguous" deployment failure with duplicate persistence units in a deployment.
+        if (!persistenceUnitMetadata.isDuplicate()) {
+            beanConfigurator.name(persistenceUnitMetadata.getPersistenceUnitName());
+            beanConfigurator.addQualifier(NamedLiteral.of(persistenceUnitMetadata.getPersistenceUnitName()));
```
jamezp: This shouldn't use the `@Named` pseudo qualifier here. We could only need that if we have more than one persistence unit defined and more than one does not have a qualifier defined.

jpa/subsystem/src/main/java/org/jboss/as/jpa/beanmanager/PersistenceCdiExtension.java
```diff
@@ -0,0 +1,20 @@
...
+import org.jipijapa.plugin.spi.PersistenceUnitMetadata;
+
+/**
+ * PersistenceCdiExtension
+ *
+ * @author Scott Marlow
+ */
+public interface PersistenceCdiExtension extends Extension {
+    IntegrationWithCDIBagImpl register(final PersistenceUnitMetadata persistenceUnitMetadata);
+    void addBeans(AfterBeanDiscovery afterBeanDiscovery);
```
jamezp: I don't know that this method is really needed as it's only used in the implementation.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
```diff
@@ -441,4 +441,8 @@ else if(emf.getProperties() != null && emf.getProperties().containsKey(ALLOWJOIN
         return result;
     }
 
+    public static boolean isDefaultPersistenceUnit(PersistenceUnitMetadata persistenceUnit) {
+        String defaultPU = persistenceUnit.getProperties().getProperty(Configuration.JPA_DEFAULT_PERSISTENCE_UNIT);
+        return (Boolean.TRUE.toString().equals(defaultPU));
```
jamezp: Should this simply be "true".equalsIgnoreCase(defaultPU))`?

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -416,4 +419,35 @@ public SharedCacheMode getSharedCacheMode() {
...
+    }
+
+    @Override
+    public String getScopeAnnotationName() {
+        return scopeAnnotationName;
+    }
+
+    @Override
+    public void setQualifierAnnotationNames(List<String> qualifierAnnotationNames) {
+        this.qualifierAnnotationNames = qualifierAnnotationNames;
```
jamezp: Should this be:
```java
this,qualifierAnnotationNames = List.copyOf(qualifierAnnotationNames)`
```

to make it immutable?

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -761,4 +761,16 @@ public interface JpaLogger extends BasicLogger {
...
+    RuntimeException classNotFound(@Cause ClassNotFoundException cause, String scopedPersistenceUnitName);
+
+    @Message(id = 78, value = "IntegratePersistenceAfterBeanDiscovery cannot register persistence unit %s as jakarta.enterprise.inject.spi.AfterBeanDiscovery event already ran for %s.")
+    IllegalStateException afterBeanDiscoveryEventRanAlready(String persistenceUnitName, String scopedPersistenceUnitName);
+
+    @LogMessage(level = INFO)
+    @Message(id = 79, value = "EntityManagerFactory CDI bean (for %s) will not have persistence unit name.  In order for the EntityManagerFactory CDI bean to be named remove duplicate copies of persistence unit %s.")
+    void willNotNameEntityManagerFactoryBean(String scopedPersistenceUnitName, String persistenceUnitName);
+
+    @Message(id = 80, value = "found more than one %s service for application deployment %s")
```
jamezp: Minor, but we should capitalize the "f" in "Found".

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -116,6 +120,33 @@ public static void deploy(DeploymentPhaseContext phaseContext, boolean startEarl
...
+            checkForDuplicates(phaseContext, platform);
+        }
+    }
+
+    // mark all duplicate persistence unit definitions
+    private static void checkForDuplicates(DeploymentPhaseContext phaseContext, Platform platform) {
+        final DeploymentUnit topDeploymentUnit = DeploymentUtils.getTopDeploymentUnit(phaseContext.getDeploymentUnit());
+        final PersistenceUnitsInApplication persistenceUnitsInApplication = topDeploymentUnit.getAttachment(PersistenceUnitsInApplication.PERSISTENCE_UNITS_IN_APPLICATION);
+        ArrayList<String> duplicateList = new ArrayList<>();
+        HashSet<String> uniqueList = new HashSet(persistenceUnitsInApplication.getCount());
```
jamezp: This should be: `new HashSet<>(persistenceUnitsInApplication.getCount());` otherwise the compiler will creating warnings.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -116,6 +120,33 @@ public static void deploy(DeploymentPhaseContext phaseContext, boolean startEarl
 
         if( startEarly) {
             nextPhaseDependsOnPersistenceUnit(phaseContext, platform);
+            checkForDuplicates(phaseContext, platform);
+        }
+    }
+
+    // mark all duplicate persistence unit definitions
+    private static void checkForDuplicates(DeploymentPhaseContext phaseContext, Platform platform) {
```
jamezp: Not critical, but `Platform` is never used.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -1141,6 +1181,46 @@ private static PersistenceProviderDeploymentHolder getPersistenceProviderDeploym
...
     }
 
+    private static IntegrationWithCDIBagImpl registerPersistenceAfterBeanDiscovery(DeploymentUnit deploymentUnit, CapabilityServiceSupport support, PersistenceUnitMetadata pu, TransactionManager transactionManager, TransactionSynchronizationRegistry transactionSynchronizationRegistry) {
+        deploymentUnit = DeploymentUtils.getTopDeploymentUnit(deploymentUnit);
+        IntegrationWithCDIBagImpl result = null;
+        if (support.hasCapability(WELD_CAPABILITY_NAME)) {
+            Optional<WeldCapability> weldCapability = support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class);
+
+            if (weldCapability.get().isPartOfWeldDeployment(deploymentUnit)) {
+                synchronized (deploymentUnit) {
```
jamezp: Is there a reason we're synchronizing here? It looks odd to me.

jpa/persistencecdiextension/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-jpa-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>38.0.0.Beta1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>persistencecdiextension</artifactId>
```
jamezp: I think I'd prefer the name `persistence-cdi-extension` to make it a little more readable.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/jboss/as/jpa/main/module.xml
```diff
@@ -0,0 +1,66 @@
...
+        <property name="jboss.api" value="private"/>
+    </properties>
+
+    <exports>
+        <exclude path="org/jboss/as/jpa/messages"/>
+    </exports>
+
+    <resources>
+        <artifact name="${org.wildfly:wildfly-jpa}"/>
+        <artifact name="${org.wildfly:persistencecdiextension}"/>
```
jamezp: This should not be here if there is a separate module for it.

preview/galleon-local/src/main/resources/modules/system/layers/base/org/jboss/as/jpa/persistencecdiextension/main/module.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<module xmlns="urn:jboss:module:1.9" name="org.jboss.as.jpa.persistencecdiextension">
```
jamezp: Given this is new, it should probably use a name like `org.wildfly.persistence.cdi.extension`.

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/beans.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
+       bean-discovery-mode="all">
```
jamezp: We should use the newer schema and `bean-discovery-mode` should be set to `annotated` to ensure we work without scanning all beans.

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/CDIPersistenceSchemaManagerTestCase.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Test
+    public void testEntityManagerFactory() throws Exception {
+
+            EntityManagerFactory emf = cmtBean.injectedEntityManagerFactory();
+            assertNotNull("expected nonnull EntityManagerFactory ", emf);
+        }
+
+    @Ignore
```
jamezp: Is there a reason this is not working?

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/CDIPersistenceSchemaManagerTestCase.java
```diff
@@ -0,0 +1,126 @@
...
+        return jar;
+    }
+
+    @Inject
+    RequestScopedTestBean cmtBean;
+
+    @Test
+    public void TestSchemaManager() throws Exception {
+        SchemaManager schemaManager = cmtBean.testSchemaManager();
+        assertNotNull("SchemaManager should of been returned", schemaManager);
```
jamezp: The wording here is confusing to me :)

jamezp (changes requested):

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/CDIPersistenceSchemaManagerTestCase.java
scottmarlow: The ignored test gets failure:

> CDIPersistenceSchemaManagerTestCase.testEMFOfEntityManagerEqualEntityManagerFactory:77 expected that EntityManagerFactory is same as EntityManager.getEntityManagerFactory expected: org.hibernate.internal.SessionFactoryImpl<org.hibernate.internal.SessionFactoryImpl@27bdff17> but was: org.jboss.weld.generated.proxies.persistence.EntityManagerFactory$AutoCloseable$1641163629$Proxy$_$$_WeldClientProxy<org.hibernate.internal.SessionFactoryImpl@27bdff17>

So `@Inject  @Pu1Qualifier  EntityManagerFactory emf` is actually a https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java generated proxy and is not the same object as I expected when I added the test.  

We could change the test to instead verify that the same persistence unit name is for both instead of testing object equivalence on Weld proxy.  Or we could obtain an EntityManager from the (Weld proxy) emf and validate that the EntityManager#getEntityManagerFactory returns the expected instance.  I don't see this as a big problem that we need to fix.  

I'm now thinking our action could be to either change the test to pass, remove it or add a comment with details from here as to why it will fail.

ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
scottmarlow: Addressed both changes.  Thanks James!

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Good catch as this was mostly for eager classloader issues in WildFly.  I'll update this as I think this is covered already via a ServiceLoader change already made.

scottmarlow: Removed ^ as we already are using a ServiceLoader.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: > Is this right even if the requested scope is not `@TransactionScoped`? Reading the JavaDoc it seems like maybe it's okay, but I just want to verify. Looking at [http://github.com/wildfly/wildfly/blob/d1b6dd703da585d60d174885ab55b9e39d422c88/weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java#L55-L84](https://github.com/wildfly/wildfly/blob/d1b6dd703da585d60d174885ab55b9e39d422c88/weld/jpa/src/main/java/org/jboss/as/weld/services/bootstrap/WeldJpaInjectionServices.java#L55-L84) it handles it differently and that is what we effectively want to mimic.



The mentioned WeldJpaInjectionServices code always indirectly returns a org.jboss.as.jpa.container.TransactionScopedEntityManager.  Applications can expect TransactionScopedEntityManager to work with both @TransactionScoped as well as without an active transaction correctly as per the Jakarta Persistence specification requirements going back several years.  

I named the TransactionScopedEntityManager class based on the JPA/Persistence spec description of `[transaction-scoped persistence context](https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#a11805)` which works with and without transactions.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: > This shouldn't use the `@Named` pseudo qualifier here. We could only need that if we have more than one persistence unit defined and more than one does not have a qualifier defined.

Isn't it better for app developers to consistently always be able to use the `@Named` pseudo qualifier in their code even if the persistence unit in question has other qualifiers specified?  Can the other qualifiers be used by Jakarta Pages?  

For reference https://jakarta.ee/specifications/pages/4.0/jakarta-server-pages-spec-4.0#resource-injection mentions resource injection (e.g. @PersistenceContext + @PersistenceUnit) but doesn't mention qualifiers.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
scottmarlow: > Should this simply be "true".equalsIgnoreCase(defaultPU))`?

:+1: 

Updated to the following which does the same internally:

```
    public static boolean isDefaultPersistenceUnit(PersistenceUnitMetadata persistenceUnit) {
        String defaultPU = persistenceUnit.getProperties().getProperty(Configuration.JPA_DEFAULT_PERSISTENCE_UNIT);
        return true == Boolean.parseBoolean(defaultPU);
    }
```

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
scottmarlow: Good question!  Perhaps a comment could explain that:

```
    synchronized (deploymentUnit) { // protect against multiple subdeployment threads registering the persistenceCdiExtension at same time
```

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/RequestScopedTestBean.java
```diff
@@ -0,0 +1,122 @@
...
+        return em.find(Employee.class, id);
+    }
+
+    @Transactional
+    public EntityManagerFactory entityManagerFactoryOfEntityManager() {
+        return em.getEntityManagerFactory();
+    }
+
+    public EntityManagerFactory injectedEntityManagerFactory() {
+            return emf;
```
rhusar: Nit: run a formatter on this line as it's off :)

scottmarlow: Hmm, looks like the random tabs insertion problem is back again.

scottmarlow: looking at https://bugs.kde.org/show_bug.cgi?id=484992 which was discussed before on https://wildfly.zulipchat.com/#narrow/channel/151151-general/topic/Fedora.2040.20.2B.20Intellij.20.2B.20alt-tab.20entering.20spaces.20into.20sources/with/459489796

scottmarlow: Hmm, not sure that I can blame KDE as the cause as I'm running GNOME...  Must be caused by something else like running KVM + switching between VMs (just a guess).

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
jamezp: Okay. That makes sense, I just wanted to make sure it was the same since it was a bit different than  the other service. One thing we should check is that this works with an `@ApplicationScoped` defined `EntityManager`. Granted, that is probably a terrible idea, but it seems allowed :)

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
jamezp: No, we don't actually want them to use `@Named`. We were only going to use that qualifier if there was more than one PU with the same name. If they're using CDI and more than one PU, they need to define a qualifier in the `persistence.xml`.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        Class<? extends Annotation> scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> entityManagerFactoryClass = EntityManagerFactory.class;
```
jamezp: Minor nit, but this variable is redundant.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            // beanConfigurator.addQualifier(qualifierType);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> criteriaBuilderClass = CriteriaBuilder.class;
```
jamezp: Same here with the redundant variable

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        Class<? extends Annotation> scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> persistenceUnitUtilClass = PersistenceUnitUtil.class;
```
jamezp: Same here with the redundant variable

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        Class<? extends Annotation> scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> cacheClass = Cache.class;
```
jamezp: Same here with the redundant variable

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        Class<? extends Annotation> scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            beanConfigurator.addQualifier(ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> metamodelClass = Metamodel.class;
```
jamezp: Same here with the redundant variable

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,385 @@
...
+        beanConfigurator.scope(scopeAnnotation);
+
+        for (String qualifier : qualifiers) {
+            final Class<? extends Annotation> qualifierType = persistenceUnitMetadata.getClassLoader()
+                    .loadClass(qualifier)
+                    .asSubclass(Annotation.class);
+            // beanConfigurator.addQualifier(qualifierType);
+            beanConfigurator.addQualifier(IntegratePersistenceAfterBeanDiscovery.ScopeProxy.createProxy(qualifierType));
+        }
+        Class<?> schemaManagerClass = SchemaManager.class;
```
jamezp: Same here with the redundant variable

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -1141,6 +1181,46 @@ private static PersistenceProviderDeploymentHolder getPersistenceProviderDeploym
...
+                synchronized (deploymentUnit) { // protect against multiple subdeployment threads registering the persistenceCdiExtension at same time
+                    final Module module = deploymentUnit.getAttachment(Attachments.MODULE);
+
+                    PersistenceCdiExtension persistenceCdiExtension = deploymentUnit.getAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY);
+                    if (null == persistenceCdiExtension) {
+                        // For WildFly Preview try loading a PersistenceCdiExtensionService (WildFly will not have this service).
+                        ServiceLoader<PersistenceCdiExtension> persistenceCdiExtensionService = module.loadService(PersistenceCdiExtension.class);
+                        if (persistenceCdiExtensionService == null) {
+                            return null;
+                        }
```
jamezp: This `null` check is not needed. The `ServiceLoader` will never be `null`.

jpa/subsystem/src/main/java/org/jboss/as/jpa/puparser/PersistenceUnitXmlParser.java
```diff
@@ -37,11 +38,15 @@ public class PersistenceUnitXmlParser extends MetaDataElementParser {
     static {
         Version defaultVersion;
         try {
-            // Try and load a jakarta namespace Jakarta Persistence API class to see if we're EE 8 or a later EE
-            PersistenceUnitXmlParser.class.getClassLoader().loadClass("jakarta.persistence.SharedCacheMode");
-            defaultVersion = Version.JPA_3_0;
+            // Try and load a jakarta namespace Jakarta Persistence API class method that is only in EE 11 to see if we're using EE 11+ (e.g. WildFly Preview).
+            if (Arrays.stream(PersistenceUnitXmlParser.class.getClassLoader().loadClass("jakarta.persistence.spi.PersistenceUnitInfo").
```
jamezp: Instead of iterating methods, could we just see if the `jakarta.persistence.SchemaManager` exists since it's new?

testsuite/preview/basic/src/test/java/org/wildfly/test/preview/persistence/cdi/CDIPersistenceSchemaManagerTestCase.java
jamezp: Couldn't we check the `EntityManagerFactory.getName()` though?

scottmarlow: Sure we could do that.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
scottmarlow: It won't be null in `WildFly Preview` but I think it will be null for `WildFly` (unless we later back port it to `WildFly`).

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Wow good catch as per the requirements any scope can be used:

> The scope specified by the scope XML element in persistence.xml, or jakarta.transaction.TransactionScoped if no scope is explicitly specified,

Use of @ApplicationScoped to inject an EntityManager bean is not going to work as EntityManager's are not thread safe.

Updated:  We should fail application deployment if we detect @ApplicationScoped is used with EntityManager.  Are there any other CDI Scopes that may also have problems like this?

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Thanks, I'll remove the @Named.  We can always add it back in later if we want to.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
jamezp: No, `ServiceLoader.load()` never returns `null`. It always returns a `ServiceLoader`, but it might just be empty.

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
jamezp: Scopes can be user defined. They're not limited to the available ones in the specification. It's not a bean defining annotation, but `@Singleton` is a scope too.

What would happen now if I have a bean that is `@ApplicationScoped` and I used a `@PersistenceContext` to inject an `EntityManager`?

scottmarlow: > Use of @ApplicationScoped to inject an EntityManager bean is not going to work as EntityManager's are not thread safe.

I thought more about this and I now disagree with what I said.  The injected org.jboss.as.jpa.container.TransactionScopedEntityManager instance is really a smart proxy that obtains a new EntityManager for every (with inactive transaction) call.  For every call within an active transaction the EntityManager instance to be used is kept in the TransactionSynchronationRegistry (per transaction) storage.  So there are no thread safety issues expected for using @ApplicationScoped.

jamezp: Excellent. Thank you @scottmarlow

jpa/persistencecdiextension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Improved the exception message a bit with some local changes.  Good idea to create a utility for throwing a good exception!

Example error with a purposely bad scope classname added:

```
ERROR] Errors: 
[ERROR]   CDIPersistenceSchemaManagerTestCase » Deployment Cannot deploy CDIPersistenceSchemaManagerTestCase.jar: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"CDIPersistenceSchemaManagerTestCase.jar\".WeldStartService" => "Failed to start service
    Caused by: org.jboss.weld.exceptions.DefinitionException: Exception List with 1 exceptions:
Exception 0 :
java.lang.RuntimeException: WFLYJPA0077: Jakarta Persistence/CDI Integration cannot be setup for bean type EntityManager, persistence unit pu1 (CDIPersistenceSchemaManagerTestCase.jar#pu1)
        at org.jboss.as.jpa@38.0.0.Beta1-SNAPSHOT//org.jboss.as.jpa.beanmanager.IntegratePersistenceAfterBeanDiscovery.entityManager(IntegratePersistenceAfterBeanDiscovery.java:157)
        at org.jboss.as.jpa@38.0.0.Beta1-SNAPSHOT//org.jboss.as.jpa.beanmanager.IntegratePersistenceAfterBeanDiscovery.addBeans(IntegratePersistenceAfterBeanDiscovery.java:126)
        at org.jboss.as.jpa@38.0.0.Beta1-SNAPSHOT//org.jboss.as.jpa.beanmanager.IntegratePersistenceAfterBeanDiscovery.afterBeanDiscovery(IntegratePersistenceAfterBeanDiscovery.java:82)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:569)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.injection.StaticMethodInjectionPoint.invoke(StaticMethodInjectionPoint.java:100)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.injection.MethodInvocationStrategy$SpecialParamPlusBeanManagerStrategy.invoke(MethodInvocationStrategy.java:199)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverMethodImpl.sendEvent(ObserverMethodImpl.java:350)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ExtensionObserverMethodImpl.sendEvent(ExtensionObserverMethodImpl.java:136)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverMethodImpl.sendEvent(ObserverMethodImpl.java:328)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverMethodImpl.notify(ObserverMethodImpl.java:298)
        at jakarta.enterprise.api@4.1.0//jakarta.enterprise.inject.spi.ObserverMethod.notify(ObserverMethod.java:140)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.util.Observers.notify(Observers.java:174)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverNotifier.notifySyncObservers(ObserverNotifier.java:307)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverNotifier.notify(ObserverNotifier.java:293)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverNotifier.fireEvent(ObserverNotifier.java:185)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.event.ObserverNotifier.fireEvent(ObserverNotifier.java:179)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.events.AbstractContainerEvent.fire(AbstractContainerEvent.java:53)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.events.AbstractDefinitionContainerEvent.fire(AbstractDefinitionContainerEvent.java:45)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.events.AfterBeanDiscoveryImpl.fire(AfterBeanDiscoveryImpl.java:75)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.WeldStartup.deployBeans(WeldStartup.java:462)
        at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.WeldBootstrap.deployBeans(WeldBootstrap.java:87)
        at org.jboss.as.weld@38.0.0.Beta1-SNAPSHOT//org.jboss.as.weld.WeldStartService.start(WeldStartService.java:80)
        at org.jboss.msc@1.5.6.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:1553)
        at org.jboss.msc@1.5.6.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1516)
        at org.jboss.msc@1.5.6.Final//org.jboss.msc.service.ServiceControllerImpl$ControllerTask.run(ServiceControllerImpl.java:1374)
        at org.jboss.threads@3.9.1//org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
        at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
        at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
        at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
        at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
        at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.ClassNotFoundException: org.wildfly.test.preview.persistence.cdi.MyBadScope from [Module \"deployment.CDIPersistenceSchemaManagerTestCase.jar\" from Service Module Loader]
        at org.jboss.modules.ModuleClassLoader.findClass(ModuleClassLoader.java:201)
        at org.jboss.modules.ConcurrentClassLoader.performLoadClassUnchecked(ConcurrentClassLoader.java:408)
        at org.jboss.modules.ConcurrentClassLoader.performLoadClass(ConcurrentClassLoader.java:396)
        at org.jboss.modules.ConcurrentClassLoader.loadClass(ConcurrentClassLoader.java:115)
        at org.jboss.as.jpa@38.0.0.Beta1-SNAPSHOT//org.jboss.as.jpa.beanmanager.IntegratePersistenceAfterBeanDiscovery.entityManager(IntegratePersistenceAfterBeanDiscovery.java:155)
        ... 33 more
"}}}}
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0

```

scottmarlow:
@jamezp @rhusar  I think all feedback has been addressed.  Thanks for the careful review!

bstansberry:
@jamezp or @rhusar Please approve.

scottmarlow:
I'll sync up with main and squash as well

jamezp:
I'll have another look at this today and send up a EE11 Draft PR to get a run in there as well.

bstansberry:
Note that the status box for this PR is currently also showing results for #19215. I think this happens when two branches are testing the same PR branch, but merged into different target branches. It's harmless, just confusing.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+            // this should never happen but still check just in case
+            throw JpaLogger.ROOT_LOGGER.afterBeanDiscoveryEventRanAlready(persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        final IntegrationWithCDIBagImpl integrationWithCDIBag = new IntegrationWithCDIBagImpl();
+        integrationWithCDIBag.setPersistenceUnitMetadata(persistenceUnitMetadata);
+        copyOnWriteArrayList.add(integrationWithCDIBag);
+        return integrationWithCDIBag;
+    }
+
+    private static final List defaultQualifier = List.of("jakarta.enterprise.inject.Default");
```
jamezp: This should be `List<String>`. We shouldn't leave off the generic type.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        BeanConfigurator<CriteriaBuilder> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(CriteriaBuilder.class);
+
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "CriteriaBuilder", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfiguator.scope(Dependent.class)`. This would save the class loading as the scope from this producer cannot be changed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        beanConfigurator.addTransitiveTypeClosure(PersistenceUnitUtil.class);
+
+
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "PersistenceUnitUtil", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfiguator.scope(Dependent.class)`. This would save the class loading as the scope from this producer cannot be changed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        BeanConfigurator<Cache> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(Cache.class);
+
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "Cache", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfiguator.scope(Dependent.class)`. This would save the class loading as the scope from this producer cannot be changed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        BeanConfigurator<Metamodel> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(Metamodel.class);
+
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "Metamodel", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfiguator.scope(Dependent.class)`. This would save the class loading as the scope from this producer cannot be changed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        BeanConfigurator<SchemaManager> beanConfigurator = afterBeanDiscovery.addBean();
+        beanConfigurator.addTransitiveTypeClosure(SchemaManager.class);
+
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "SchemaManager", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfiguator.scope(Dependent.class)`. This would save the class loading as the scope from this producer cannot be changed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        final IntegrationWithCDIBagImpl integrationWithCDIBag = new IntegrationWithCDIBagImpl();
+        integrationWithCDIBag.setPersistenceUnitMetadata(persistenceUnitMetadata);
+        copyOnWriteArrayList.add(integrationWithCDIBag);
+        return integrationWithCDIBag;
+    }
+
+    private static final List defaultQualifier = List.of("jakarta.enterprise.inject.Default");
+    private static final String transactionScoped = "jakarta.transaction.TransactionScoped";
+    private static final String applicationScoped = "jakarta.enterprise.context.ApplicationScoped";
+    protected static final String dependentScoped = "jakarta.enterprise.context.Dependent";
```
jamezp: This could go away if the comments below are addressed.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        } else {
+            JpaLogger.ROOT_LOGGER.willNotNameEntityManagerFactoryBean(persistenceUnitMetadata.getScopedPersistenceUnitName(), persistenceUnitMetadata.getPersistenceUnitName());
+        }
+        Class<? extends Annotation> scopeAnnotation = null;
+        try {
+            scopeAnnotation = persistenceUnitMetadata.getClassLoader().loadClass(scope).asSubclass(Annotation.class);
+        } catch (ClassNotFoundException e) {
+            throw JpaLogger.ROOT_LOGGER.classNotFound(e, "EntityManagerFactory", persistenceUnitMetadata.getPersistenceUnitName(), persistenceUnitMetadata.getScopedPersistenceUnitName());
+        }
+        beanConfigurator.scope(scopeAnnotation);
```
jamezp: This could simply be `beanConfigurator.scope(ApplicationScoped.class)`.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+        }
+        final IntegrationWithCDIBagImpl integrationWithCDIBag = new IntegrationWithCDIBagImpl();
+        integrationWithCDIBag.setPersistenceUnitMetadata(persistenceUnitMetadata);
+        copyOnWriteArrayList.add(integrationWithCDIBag);
+        return integrationWithCDIBag;
+    }
+
+    private static final List defaultQualifier = List.of("jakarta.enterprise.inject.Default");
+    private static final String transactionScoped = "jakarta.transaction.TransactionScoped";
+    private static final String applicationScoped = "jakarta.enterprise.context.ApplicationScoped";
```
jamezp, line 107: This could go away if the single usage below is updated.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,444 @@
...
+ * - no interceptor bindings,
+ * - a bean implementation which simply obtains the instance of the bean type by calling the appropriate getter method of the EntityManagerFactory bean.
+ * - To access these bean types (CriteriaBuilder, PersistenceUnitUtil, Cache, SchemaManager, and Metamodel) from callsites that use @Resource or JNDI lookup,
+ * users must first obtain the EntityManagerFactory and then use the appropriate getter methods.
+ *
+ * @author Scott Marlow
+ */
+
+public class IntegratePersistenceAfterBeanDiscovery implements PersistenceCdiExtension {
+    private final CopyOnWriteArrayList<IntegrationWithCDIBagImpl> copyOnWriteArrayList = new CopyOnWriteArrayList();
```
jamezp: This should be `new CopyOnWriteArrayList<>()`. Otherwise we'll get compiler warnings.

Minor nit, something more description should be used for the name here IMO. The name `copyOnWriteArrayList` just tells the type, not what it's used for.

This is not a critical fix though :)

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -605,7 +638,11 @@ private static void deployPersistenceUnitPhaseTwo(
                     adaptor, integratorAdaptors, provider, PersistenceUnitRegistryImpl.INSTANCE,
                     deploymentUnit.getServiceName(), validatorFactory,
                     deploymentUnit.getAttachment(org.jboss.as.ee.naming.Attachments.JAVA_NAMESPACE_SETUP_ACTION),
-                    beanManagerAfterDeploymentValidation);
+                    beanManagerAfterDeploymentValidation,
+                    deploymentUnit.getAttachment(JpaAttachments.TRANSACTION_SYNCHRONIZATION_REGISTRY),
```
jamezp: This looks like we already get this above in the `transactionSynchronizationRegistry`,

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -1141,6 +1181,43 @@ private static PersistenceProviderDeploymentHolder getPersistenceProviderDeploym
...
     }
 
+    private static IntegrationWithCDIBagImpl registerPersistenceAfterBeanDiscovery(DeploymentUnit deploymentUnit, CapabilityServiceSupport support, PersistenceUnitMetadata pu, TransactionManager transactionManager, TransactionSynchronizationRegistry transactionSynchronizationRegistry) {
+        deploymentUnit = DeploymentUtils.getTopDeploymentUnit(deploymentUnit);
+        IntegrationWithCDIBagImpl result = null;
+        if (support.hasCapability(WELD_CAPABILITY_NAME)) {
+            Optional<WeldCapability> weldCapability = support.getOptionalCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class);
+
+            if (weldCapability.get().isPartOfWeldDeployment(deploymentUnit)) {
+                synchronized (deploymentUnit) { // protect against multiple subdeployment threads registering the persistenceCdiExtension at same time
```
jamezp: This still feels weird to me.

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
```diff
@@ -0,0 +1,23 @@
...
+    <resources>
+        <artifact name="${jakarta.persistence:jakarta.persistence-api}"/>
+    </resources>
+
+    <dependencies>
+        <!-- java.instrument export due to jdeps result -->
+        <module name="java.instrument" export="true"/>
+        <module name="java.logging"/>
+        <!-- PersistenceUnitInfo needs javax.sql.DataSource -->
+        <module name="java.sql" export="true"/>
```
jamezp, line 19: Why do we export this?

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
```diff
@@ -0,0 +1,23 @@
...
+
+<module xmlns="urn:jboss:module:1.9" name="jakarta.persistence.api">
+
+    <resources>
+        <artifact name="${jakarta.persistence:jakarta.persistence-api}"/>
+    </resources>
+
+    <dependencies>
+        <!-- java.instrument export due to jdeps result -->
+        <module name="java.instrument" export="true"/>
```
jamezp: Why do we export this?

jamezp (changes requested):

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Changed variable name to `persistenceUnitIntegrationStuff`

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
scottmarlow: I'll update it...

scottmarlow: I do see the weirdness now and removed the synchronization.  I missed before that `putAttachment` is atomic already and returns null only to the thread that actually does the put.

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
scottmarlow: Persistence SPEC API needs the  JDBC classes from the JDK like javax.sql.DataSource in https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/spi/persistenceunitinfo

scottmarlow: So we export it so that applications will have access to DataSource and friends.

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
scottmarlow: Good question on this one.  From what I recall I think the `javassist` bytecode enhancement library [needed it](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html) for application enhancement.  I'm not sure if we need it still...

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
jamezp: Typically this would be done in a DUP to add the dependency to a deployment. I guess in this specific case it's okay as it really is required and the user can't package their own.

preview/galleon-local/src/main/resources/modules/system/layers/base/jakarta/persistence/api/main/module.xml
jamezp: IMO it should be added as a dependency to the `javassist` module then. It feels wrong to export it here.

Please note if this is just a copy of what is already done it's not something we need to change now :)

scottmarlow: I'll remove the export.  Good time to try this for Preview!

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: I'll do the same for the extra loading of the ApplicationScoped class as well.

scottmarlow:
All tests passed which is great to see!  I'm pushing one more change to remove an extra class load that we don't need.

bstansberry:
@darranl I put a 38.x label on this because if it's approved in time for 38 Beta that would be great. If not. it's ok.

It's WFP so we could always put in into 38 Final if we want, but of course having changes in the beta is better.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+        final IntegrationWithCDIBagImpl integrationWithCDIBag = new IntegrationWithCDIBagImpl();
+        integrationWithCDIBag.setPersistenceUnitMetadata(persistenceUnitMetadata);
+        persistenceUnitIntegrationStuff.add(integrationWithCDIBag);
+        return integrationWithCDIBag;
+    }
+
+    private static final List<String> defaultQualifier = List.of("jakarta.enterprise.inject.Default");
+    private static final String transactionScoped = "jakarta.transaction.TransactionScoped";
+    private static final String applicationScoped = "jakarta.enterprise.context.ApplicationScoped";
```
jamezp, line 107: Not blocking merging on this, but this is not used and can be removed. Once merged, please file a followup PR for that @scottmarlow.

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -1141,6 +1181,42 @@ private static PersistenceProviderDeploymentHolder getPersistenceProviderDeploym
...
+                        // putAttachment will return null if the attachment was not set yet, in which case we will also register the CDI extension with Weld
+                        if (null == deploymentUnit.putAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY, persistenceCdiExtension))
+                            weldCapability.get().registerExtensionInstance(persistenceCdiExtension, deploymentUnit);
+                    }
+                }
+                if ( persistenceCdiExtension != null) {
+                    result = persistenceCdiExtension.register(pu);
+                    result.setTransactionManager(transactionManager);
+                    result.setTransactionSynchronizationRegistry(transactionSynchronizationRegistry);
+                }
```
jamezp, line 1214: We should fix this in a follow up, but I'm not sure this is correct. I need to look at how sub-deployments are processed though.

The issue though is that if `deploymentUnit.putAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY, persistenceCdiExtension)` returns a value, you'd be executing the `register()` on the wrong instance.

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PersistenceUnitServiceImpl.java
```diff
@@ -170,18 +188,24 @@ public Void run() {
...
+                                            integrationWithCDIBag.setEntityManagerFactory(entityManagerFactory);
+                                        }
                                     }
                                     persistenceUnitRegistry.add(getScopedPersistenceUnitName(), getValue());
                                     if(wrapperBeanManagerLifeCycle != null) {
                                         beanManagerAfterDeploymentValidation.register(persistenceProviderAdaptor, wrapperBeanManagerLifeCycle);
                                     }
+                                    if (proxyBeanManager != null && proxyBeanManager.delegate() != null) {
+                                        createCDIBeansForPersistence(proxyBeanManager.delegate(), entityManagerFactory, pu, classLoader);
+                                    }
```
jamezp, line 208: Again for a follow up, I don't understand this. It invokes the `createCDIBeansForPersistence()` which does nothing.

jamezp (approved):
@scottmarlow I think we should merge this, but I've got 3 requests for a follow up PR before final. Let's discuss these.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
scottmarlow: Thanks for the review and catching these issues!  https://github.com/wildfly/wildfly/pull/19232 is the followup

jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
scottmarlow: Thanks @jamezp!  This is fixed in https://github.com/wildfly/wildfly/pull/19232

PR #19095, opened 2025-07-30
[WFLY-20812] Add test case for ear-exclusions-cascaded-to-subdeployments and its implicit sub-deployments

mskacelik:
https://issues.redhat.com/browse/WFLY-20812

/cc @yersan 
PR for WFCORE: https://github.com/wildfly/wildfly-core/pull/6467
These are the integration tests for WFCORE-7310 https://issues.redhat.com/browse/WFCORE-7310

> [!IMPORTANT]
>  EarJbossStructureCascadeExclusionsImplicitSubdeploymentTestCase is expected to fail with the current implementation of WildFly Core.

There might be a discussion about whether `EarJbossStructureCascadeExclusionsExplicitSubdeploymentTestCase` is needed at all, since having a test-case for sub-deployment is redundant.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/classloading/ear/EarJbossStructureCascadeExclusionsExplicitSubdeploymentTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/deployment/classloading/ear/EarJbossStructureCascadeExclusionsExplicitSubdeploymentTestCase.java
@@ -15,15 +15,15 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 @RunWith(Arquillian.class)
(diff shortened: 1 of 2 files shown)


mskacelik:
Should I create a separate WildFly JIRA issue for tests, or is that not necessary?

yersan:
@mskacelik We need a WildFly JIRA to track this under WildFly project, could you create one with a short description describing its purpose? thanks

yersan:
> Should I create a separate WildFly JIRA issue for tests, or is that not necessary?

I've just read this, yes, we need one. any change that will end up in WildFly repo should have its own WildFly JIRA

mskacelik:
rebasing

yersan (approved):

bstansberry:
Thanks @mskacelik and @yersan

PR #19104, opened 2025-07-31
WFLY-20853 - Upgrade to Hibernate ORM 7.1.0 and Hibernate Search 8.1.1 and register a new internal cache based on Caffeine

Sanne:
https://issues.redhat.com/browse/WFLY-20853

As being discussed on:
 - https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Hibernate.20ORM.207.2E1/with/532023215

ORM 7.1 introduces the ability to customize the cache implementation which we use internally for a number of reasons, primarily query plans.
The default implementation provided in Hibernate ORM is not very efficient and has some scalability limitations; it's not bad but the Caffeine based implementation is far better: I've had very successful results experimenting with this in Quarkus and I believe we should do the same for WildFly.

Essentially, we need to inject a very trivial wrapper around Caffeine. Caffeine is already among the modules as it's a core dependency of Infinispan, so we can reuse the existing module, improving efficiency without adding additional dependencies. Also, I think eventually we'll need to find a way to deprecate the legacy one included in Hibernate ORM, so this would prepare for that.

N.B. this internal cache is not related to the 2nd level cache implementations.

(no longer a draft)

cc/ @scottmarlow @yrodiere @marko-bekhta

____

> WildFly issue links:
> * [WFLY-20853](https://issues.redhat.com/browse/WFLY-20853)

### DIFF
--- a/boms/preview-ee/pom.xml
+++ b/boms/preview-ee/pom.xml
@@ -50,16 +50,16 @@
         <version.jakarta.websocket.jakarta-websocket-api>2.2.0</version.jakarta.websocket.jakarta-websocket-api>
         <version.jakarta.ws.rs.jakarta-ws-rs-api>4.0.0</version.jakarta.ws.rs.jakarta-ws-rs-api>
         <!-- Required for Hibernate Search -->
-        <version.org.apache.lucene>9.12.1</version.org.apache.lucene>
-        <version.org.bytebuddy>1.15.11</version.org.bytebuddy>
+        <version.org.apache.lucene>9.12.2</version.org.apache.lucene>
+        <version.org.bytebuddy>1.17.6</version.org.bytebuddy>
         <version.org.glassfish.expressly>6.0.0</version.org.glassfish.expressly>
         <version.org.glassfish.jakarta.faces>4.1.3</version.org.glassfish.jakarta.faces>
         <version.org.jboss.resteasy>7.0.0.Beta1</version.org.jboss.resteasy>
         <version.org.jboss.weld.weld>6.0.3.Final</version.org.jboss.weld.weld>
         <version.org.jboss.weld.weld-api>6.0.Final</version.org.jboss.weld.weld-api>
-        <version.org.hibernate>7.0.2.Final</version.org.hibernate>
-        <version.org.hibernate.models>1.0.0</version.org.hibernate.models>
-        <version.org.hibernate.search>8.0.0.Final</version.org.hibernate.search>
+        <version.org.hibernate>7.1.0.Final</version.org.hibernate>
+        <version.org.hibernate.models>1.0.1</version.org.hibernate.models>
+        <version.org.hibernate.search>8.1.1.Final</version.org.hibernate.search>
         <version.org.hibernate.validator>9.0.1.Final</version.org.hibernate.validator>
         <version.org.wildfly.security.jakarta.elytron-ee>4.0.0.Beta1</version.org.wildfly.security.jakarta.elytron-ee>
     </properties>
--- a/jpa/hibernate7/pom.xml
+++ b/jpa/hibernate7/pom.xml
@@ -85,6 +85,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-hibernate-cache-spi</artifactId>
             <scope>provided</scope>
--- a/jpa/hibernate7/pom.xml
+++ b/jpa/hibernate7/pom.xml
@@ -75,12 +75,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.hibernate.common</groupId>
-            <artifactId>hibernate-commons-annotations</artifactId>
-            <scope>provided</scope>
-        </dependency>
-
-        <dependency>
             <groupId>org.hibernate.orm</groupId>
             <artifactId>hibernate-core</artifactId>
             <scope>provided</scope>
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/ServiceContributorImpl.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/ServiceContributorImpl.java
@@ -7,6 +7,7 @@ package org.wildfly.persistence.jipijapa.hibernate7.service;
 import static org.wildfly.persistence.jipijapa.hibernate7.JpaLogger.JPA_LOGGER;
 
 import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
+import org.hibernate.internal.util.cache.InternalCacheFactory;
 import org.hibernate.service.spi.ServiceContributor;
 import org.kohsuke.MetaInfServices;
 
@@ -61,5 +62,13 @@ public class ServiceContributorImpl implements ServiceContributor {
             JPA_LOGGER.tracef("ServiceContributorImpl#contribute adding ORM initiator for 2lc region factory");
             serviceRegistryBuilder.addInitiator(new WildFlyCustomRegionFactoryInitiator());
         }
+
+        //Now customize the Hibernate "internal cache": this is a component used for internal needs
+        //(it's NOT the second level cache). ORM integrators such as WildFly are encouraged to inject
+        //a better implementation than the one included by default.
+        //WildFly happens to already pack Caffeine, which is an excellent choice so that's what it will use.
+        //This setting is not user configurable.
+        serviceRegistryBuilder.addService(InternalCacheFactory.class, new WildFlyCustomInternalCacheFactory());
     }
+
 }
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/WildFlyCustomInternalCache.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/WildFlyCustomInternalCache.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.persistence.jipijapa.hibernate7.service;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import org.hibernate.internal.util.cache.InternalCache;
+
+import java.util.function.Function;
+
+/**
+ * A specialized implementation of {@link InternalCache} that uses a Caffeine-based cache for
+ * improved performance and caching capabilities. Designed for internal use within WildFly to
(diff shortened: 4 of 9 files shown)


scottmarlow:
org.wildfly.test.integration.microprofile.reactive.messaging.otel.KafkaReactiveMessagingAndOtelTestCase
testNoOpenTelemetryTracing test failure doesn't seem likely to be caused by this change.

boms/preview-ee/pom.xml
```diff
@@ -50,15 +50,15 @@
...
         <version.org.glassfish.expressly>6.0.0</version.org.glassfish.expressly>
         <version.org.glassfish.jakarta.faces>4.1.3</version.org.glassfish.jakarta.faces>
         <version.org.jboss.resteasy>7.0.0.Beta1</version.org.jboss.resteasy>
         <version.org.jboss.weld.weld>6.0.3.Final</version.org.jboss.weld.weld>
         <version.org.jboss.weld.weld-api>6.0.Final</version.org.jboss.weld.weld-api>
-        <version.org.hibernate>7.0.2.Final</version.org.hibernate>
-        <version.org.hibernate.models>1.0.0</version.org.hibernate.models>
+        <version.org.hibernate>7.1.0.CR1</version.org.hibernate>
+        <version.org.hibernate.models>1.0.1</version.org.hibernate.models>
         <version.org.hibernate.search>8.0.0.Final</version.org.hibernate.search>
```
scottmarlow: We will also want to upgrade Hibernate Search to 8.1.0.CR1 which can be separate from this pr.

jpa/hibernate7/pom.xml
```diff
@@ -90,6 +90,12 @@
             <artifactId>hibernate-scan-jandex</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
```
scottmarlow, line 88: I don't recall if WildFly has any other dependencies that depend on a personal github repo like https://github.com/ben-manes/caffeine.  I'm not sure if this is a concern for WildFly but pointing it out so it won't be missed by the reviewers.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/ServiceContributorImpl.java
```diff
@@ -61,5 +62,13 @@ public void contribute(StandardServiceRegistryBuilder serviceRegistryBuilder) {
             JPA_LOGGER.tracef("ServiceContributorImpl#contribute adding ORM initiator for 2lc region factory");
             serviceRegistryBuilder.addInitiator(new WildFlyCustomRegionFactoryInitiator());
         }
+
+        //Now customize the Hibernate "internal cache": this is a component used for internal needs
+        //(it's NOT the second level cache). ORM integrators such as WildFly are encouraged to inject
+        //a better implementation than the one included by default.
+        //WildFly happens to already pack Caffeine, which is an excellent choice so that's what it will use.
+        //This setting is not user configurable.
+        serviceRegistryBuilder.addService(InternalCacheFactory.class, new WildFlyCustomInternalCacheFactory());
```
scottmarlow, line 71: For reviewers reference, InternalCacheFactory link is https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/internal/util/cache/InternalCacheFactory.java

scottmarlow: Might also be worth including part of https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Hibernate.20ORM.207.2E1/near/532073676  conversation with a bit more detail about this change and why:

> Yoann Rodière
> Uh just to clarify, because I'm not sure that's clear from your message... we're not talking about second-level caching. This is not about caching entities or query results.
> We're talking about a different, internal cache, which is used to cache e.g. HQL query plans. It used to be implemented by a custom Map, and Sanne found it to be quite sub-optimal, which is why he's moving to a Caffeine cache instead in Quarkus (and wants to do it in Wildfly as well).
> So... WildFly caching, statistics, and transations would not, I think, be relevant... ? 

> Sanne Grinovero
> Incidentally, Hibernate ORM's statistics are affected as well, as it does use this internal cache too. It's not a use case I endorse, and we should probably talk about removing it, but it does use it today.
> Anyhow, it won't change semantics of the statistics, just less so of an overhead as Caffeine is more efficient and scalable.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/WildFlyCustomInternalCache.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public void put(K key, V value) {
+        cache.put(key, value);
+    }
+
+    @Override
+    public void clear() {
+        cache.invalidateAll();
```
scottmarlow, line 46: Just to note that since this change doesn't impact how we cluster the Hibernate second level cache so that there is no concern about invalidating this case on other cluster nodes when a local cache.clear() method is called.  

@Sanne @yrodiere is the `cache` instance per deployment or meant to be shared across all deployments?  I'd expect per deployment but want to be sure I understand.  Thanks!

yrodiere: It's per deployment, per persistence unit even.

Completely local, but that's fine because the cache value is fully determined from the cache key + Hibernate version + Hibernate mapping -- the database content is irrelevant.

boms/preview-ee/pom.xml
Sanne: sure, I kept the commit which updates versions separate so you can cherry pick as you prefer.

jpa/hibernate7/pom.xml
Sanne: N.B. it's not a new dependency, as mentioned also on zulip it already is a hard dependency of both the Infinispan and the Artemis modules.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/service/WildFlyCustomInternalCache.java
Sanne: One ORM instance might need multiple caches, for various different internal purposes: please note we're not injecting a _cache instance_ but a _cache factory_.  Hibernate ORM will manage their lifecycle as necessary.

So yes they are scoped within the persistence unit - just pointing out that the number of them is undefined, there might be multiple of them within a single PU, depending on the ORM's configuration.

scottmarlow (approved):

Sanne:
Updated it to the final versions and rebased. @marko-bekhta could you have a look to check if I nailed all dependencies and alignment needs?

marko-bekhta (approved):
nice 🎉 🙂 

>  Updated it to the final versions and rebased. @marko-bekhta could you have a look to check if I nailed all dependencies and alignment needs?

yes, models, Lucene and ORM/Search versions here are all aligned 👍🏻 🙂 

I noticed we are missing an override for the elasticsearch-client in the preview.
It's not a problem in the sense that the client hasn't really changed, probably for a long time now, and all the versions are sort of interchangeable... I'll try to submit a follow-up to override the client version in the preview.

Sanne:
Rebased to add JIRA references
 - https://issues.redhat.com/browse/WFLY-20853

scottmarlow:
Should we add https://github.com/scottmarlow/wildfly/commit/bf6f219200088cc758dbfe82fe0859ee1ced0aa2 test change to insure the internal cache is used for one of the existing WildFly unit tests that I think will also be used for WildFly Preview testing?

```
diff --git a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/transaction/persistence.xml b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/transaction/persistence.xml
index 136142b9f5..33691b4b7e 100644
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/transaction/persistence.xml
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/transaction/persistence.xml
@@ -15,6 +15,9 @@
             <property name="hibernate.enhancer.enableDirtyTracking" value="true"/>
             <property name="hibernate.enhancer.enableLazyInitialization" value="true"/>
             <property name="hibernate.enhancer.enableAssociationManagement" value="true"/>
+            <property name="hibernate.query.plan_cache_max_size" value="100"/>
+            <property name="hibernate.query.plan_cache_enabled" value="true"/>
+
         </properties>
     </persistence-unit>
     <persistence-unit name="mypc">
@@ -26,6 +29,9 @@
             <property name="hibernate.enhancer.enableDirtyTracking" value="true"/>
             <property name="hibernate.enhancer.enableLazyInitialization" value="true"/>
             <property name="hibernate.enhancer.enableAssociationManagement" value="true"/>
+            <property name="hibernate.query.plan_cache_max_size" value="100"/>
+            <property name="hibernate.query.plan_cache_enabled" value="true"/>
+
         </properties>
     </persistence-unit>
     <persistence-unit name="unsynchronized">
@@ -37,6 +43,9 @@
             <property name="hibernate.enhancer.enableDirtyTracking" value="true"/>
             <property name="hibernate.enhancer.enableLazyInitialization" value="true"/>
             <property name="hibernate.enhancer.enableAssociationManagement" value="true"/>
+            <property name="hibernate.query.plan_cache_max_size" value="100"/>
+            <property name="hibernate.query.plan_cache_enabled" value="true"/>
+
         </properties>
     </persistence-unit>

```

Updated:

The various WildFly Preview CI Persistence tests should already be using the query plan cache so no need for ^ change.

scottmarlow (approved):

jamezp (approved):

Sanne:
I'm not sure why `org.wildfly.test.integration.elytron.oidc.client.propagation.OidcIdentityPropagationTestCase` would possibly be affected by an Hibernate upgrade.

TeamCity reports it's failing for the first time, however it seems to have included many other changes as well in a batch.

It's probably worth testing again, but isolating individual changes? I'm not sure how to trigger that.
cc/ @darranl ?

jamezp:
@Sanne this is a known intermittent failure and will not block merging. I'm going to kick off a CI run on the EE11 branch which will get a better test run in, #19136 

Once that is done and looks okay, I'll likely merge this.

jamezp:
A test seems to be hanging on the EE11 run. I've ran it twice and it has hung twice. However, it's not hanging for me locally. I'll have to dig into it more tomorrow.

jamezp:
This is not failing for me locally. I'm not sure why it's failing CI as these tests have nothing to do with Hibernate or Jakarta Persistence.

jamezp:
Closed the wrong PR :)

scottmarlow:
Thanks @Sanne + @jamezp + @marko-bekhta !!!

PR #19106, opened 2025-08-01
[WFLY-20804] Add a test case to demonstrate that two deployments secured using MicroProfile JWT can be deployed in parallel.

darranl:
https://issues.redhat.com/browse/WFLY-20804


kabir (approved):

bstansberry:
Thanks @darranl and @kabir

PR #19118, opened 2025-08-08
WFLY-20836 Replace non-breaking spaces with regular spaces in our documentation

rhusar:
https://issues.redhat.com/browse/WFLY-20836

This is a remnant of the migration from hosted to AsciiDoc which was never addressed. This is generally an annoyance, since our development tooling/IDEs won't work all too well, e.g. won't reformat correctly, nor clear extra spaces trailing a line like we do in the Java code, and other things.


bstansberry (approved):
LGTM; thanks @rhusar. I'd say 99% of the changes here are in source blocks, where the rendering is going to represent the source faithfully and not insert random line breaks. So no benefit to a nbsp. The other 1% is trailing spaces, which should be removed, or placeholder spaces in table structures, where the layout is defined by the table and there won't be random line breaks.

bstansberry:
Thanks @rhusar

rhusar:
> LGTM; thanks @rhusar. I'd say 99% of the changes here are in source blocks, where the rendering is going to represent the source faithfully and not insert random line breaks. So no benefit to a nbsp. The other 1% is trailing spaces, which should be removed, or placeholder spaces in table structures, where the layout is defined by the table and there won't be random line breaks.

Great assessment, yes, I got to the same conclusion.

The rendering needs to be driven be AsciiDoc, not by us. The e.g. `[source,xml]` is a literal block in AsciiDoc meaning they already preserve indentation and whitespace exactly as written in the document. Which is also why I am fixing this, since if you want to e.g. reformat an XML code within that block from an IDE/tooling, it will get confused by the NBSP instead of a whitespace.

PR #19120, opened 2025-08-11
WFLY-20837 EJB3 subsystem doesn't have any transformer/reject tests

rhusar:
https://issues.redhat.com/browse/WFLY-20837

This PR adds transform and reject tests for all supported EAP versions for the EJB 3 subsystem.

Currently, the EJB3 subsystems is untested for transformers. It gives out false impression since there is a test class called org.jboss.as.ejb3.subsystem.Ejb3TransformersTestCase which contains some of the testing infrastructure pieces, but those are NEVER used and thus never tested, e.g. https://github.com/wildfly/wildfly/blob/37.0.0.Final/ejb3/src/test/java/org/jboss/as/ejb3/subsystem/Ejb3TransformersTestCase.java#L92


rhusar:
@pferraro and @tadamski please review and help me confirm that this is testing sufficiently the changes between the models. Specifically please review the `EJB3TransformersTestCase.createFailedOperationConfig(..)` and the `subsystem-ejb3-reject.xml` and `subsystem-ejb3-transform.xml` test files. I am wondering whether we want to have multiple transform files, since the timer-service is singleton. Thanks!

ejb3/src/test/resources/org/jboss/as/ejb3/subsystem/subsystem-ejb3-reject.xml
```diff
@@ -37,12 +37,13 @@
         <passivation-store name="infinispan" cache-container="ejb" bean-cache="default" max-size="10"/>
     </passivation-stores>
     <async thread-pool-name="default"/>
-    <timer-service thread-pool-name="default">
+    <timer-service thread-pool-name="timer-service-thread-pool" default-data-store="file-data-store">
         <data-stores>
             <file-data-store name="file-data-store" path="${prop.timer-service.path:timer-service-data}" relative-to="jboss.server.data.dir"/>
+            <database-data-store name="database-data-store" datasource-jndi-name="${prop.timer-service-database:java:global/DataSource}" database="hsql" partition="mypartition" allow-execution="true" refresh-interval="100"/>
         </data-stores>
     </timer-service>
```
pferraro: This XML should validate rejection of the timer-management attributes.

pferraro (changes requested):

ejb3/src/test/resources/org/jboss/as/ejb3/subsystem/subsystem-ejb3-reject.xml
rhusar: Great point. TBH I had this structured like this originally, but was missing the step of rejecting the whole resource. Fixed now, thanks!

pferraro (approved):

rhusar:
Rebased to rerun CI - no code changes.

rhusar:
Thanks @pferraro !

PR #19135, opened 2025-08-15
[WFLY-20854] Override the version of the Elasticsearch client in the WildFly Preview

marko-bekhta:
https://issues.redhat.com/browse/WFLY-20854

We should match the version of ES client libs that Search is expecting, and since preview has gone much further in versions ... Search now is based on ES 9.1.2 (for Hibernate Search 8.1.1). This should go on top of the ORM/Search upgrade and will probably need a rebase once the other pr is merged, so I'll open it as a draft for now, just so we won't forget about doing this).

- Follows up on: https://github.com/wildfly/wildfly/pull/19104


bstansberry (approved):

bstansberry:
Thanks @marko-bekhta

PR #19158, opened 2025-09-01
[WFLY-19125] WFLY CWE-611 - proof parsing against exploit

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-19125

Im just filing PR so CI does its job. In few places Util class caused weird failures. Im going to give it another sping, maybe its something insignificant(other than module dep).

### DIFF
--- a/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFSharedTldsProcessor.java
+++ b/jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFSharedTldsProcessor.java
@@ -25,6 +25,7 @@ import org.jboss.metadata.web.spec.TldMetaData;
 import org.jboss.modules.Module;
 import org.jboss.modules.ModuleClassLoader;
 import org.jboss.modules.ModuleLoadException;
+import org.wildfly.common.xml.XMLInputFactoryUtil;
 
 /**
  * Cache the TLDs for JSF and add them to deployments as needed.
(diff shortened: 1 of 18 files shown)


jsf/subsystem/src/main/java/org/jboss/as/jsf/deployment/JSFSharedTldsProcessor.java
```diff
@@ -67,7 +68,7 @@ private Map<String, List<TldMetaData>> getMap() {
 
     private TldMetaData parseTLD(InputStream is) throws Exception {
         try {
-            final XMLInputFactory inputFactory = XMLInputFactory.newInstance();
+            final XMLInputFactory inputFactory = XMLInputFactoryUtil.create();
```
bstansberry, line 71: We must not have tests that exercise this in the testsuite because the org.jboss.as.jsf module has no dependency on org.wildfly.common. So this would not work.

So that dep needs to be added.

bstansberry (dismissed):
@baranowb This looks good except for the one module dep problem I noted.

@pferraro Can you have a look at the two clustering/ tree changes. They seem fine but they are a processing documents a bit different from the 'ee' documents the others touch, so maybe (unlikely) something is different.

bstansberry (changes requested):
See comments above.

darranl:
@baranowb checking the history I don't think we have any update since Brian's review - please let me know if this is incorrect.

bstansberry:
Superseded by #20369

PR #19168, opened 2025-09-08
[WFLY-20889] improve jdb driver add/remove failure message

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-20889

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
+++ b/connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
@@ -372,8 +372,8 @@ public interface ConnectorLogger extends BasicLogger {
      * @param moduleName the module name.
      * @return the message.
      */
-    @Message(id = 41, value = "Failed to load module for driver [%s]")
-    String failedToLoadModuleDriver(String moduleName);
+    @Message(id = 41, value = "Failed to load driver module [%s]. %s, cause of failure [%s]")
+    String failedToLoadModuleDriver(String moduleName, String moduleLocation, String cause);
 
     /**
      * Creates an exception indicating a failure to match the pool.
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/JdbcDriverAdd.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/datasources/JdbcDriverAdd.java
@@ -87,7 +87,8 @@ public class JdbcDriverAdd extends AbstractAddStepHandler {
         } catch (ModuleNotFoundException e) {
             throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.missingDependencyInModuleDriver(moduleName, e.getMessage()), e);
         } catch (ModuleLoadException e) {
-            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName), e);
+            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName, e.getLocalizedMessage(),
+                    e.getCause() != null ? e.getCause().getLocalizedMessage() : "-"), e);
         }
 
         if (dataSourceClassName != null) {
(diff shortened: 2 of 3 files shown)


tadamski (approved):
LGTM

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/JdbcDriverRemove.java
```diff
@@ -72,7 +72,7 @@ protected void recoverServices(OperationContext context, ModelNode operation, Mo
             context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.missingDependencyInModuleDriver(moduleName, e.getMessage()));
             return;
         } catch (ModuleLoadException e) {
-            context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName));
+            context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName, e.getMessage(), e.getCause()!=null ? e.getCause().getMessage() :"-"));
```
rhusar: The spaces are off in this line. It's generally a good practice to run a formatter before submitting a PR.

rhusar (changes requested):

connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
```diff
@@ -372,8 +372,8 @@ public interface ConnectorLogger extends BasicLogger {
      * @param moduleName the module name.
      * @return the message.
      */
-    @Message(id = 41, value = "Failed to load module for driver [%s]")
-    String failedToLoadModuleDriver(String moduleName);
+    @Message(id = 41, value = "Failed to load module for driver [%s]. Location [%s], cause of failure [%s]")
```
bstansberry: 1) The original message wasn't great even for what it said. 'Failed to load module for driver xxx' implies xxx is the name of the driver. But looking at the code it's the name of the module. So, it should be

"Failed to load driver module [%s]."

2) Why "Location"? The calls to the ModuleLoadException constructors aren't passing in anything that particularly like a location.

baranowb: ok, will take a look once I have spare second.

baranowb: In this case [Error loading module from /home/baranowb/redhat/git/wildfly/ee-dist/target/wildfly-38.0.0.Beta1-SNAPSHOT/modules/system/layers/base/com/mysql/main/module.xml]. I removed [] and lead string, all its left is the message.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/JdbcDriverAdd.java
```diff
@@ -87,7 +87,8 @@ protected void performRuntime(OperationContext context, ModelNode operation, Mod
         } catch (ModuleNotFoundException e) {
             throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.missingDependencyInModuleDriver(moduleName, e.getMessage()), e);
         } catch (ModuleLoadException e) {
-            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName), e);
+            throw new OperationFailedException(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName, e.getMessage(),
```
rhusar: Unless there is some specific reason that is unknown to me, this is feeding the message to a localized string already, so this should always use `Throwable.getLocalizedMessage(..)`.

connector/src/main/java/org/jboss/as/connector/subsystems/datasources/JdbcDriverRemove.java
```diff
@@ -72,7 +72,8 @@ protected void recoverServices(OperationContext context, ModelNode operation, Mo
             context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.missingDependencyInModuleDriver(moduleName, e.getMessage()));
             return;
         } catch (ModuleLoadException e) {
-            context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName));
+            context.getFailureDescription().set(ConnectorLogger.ROOT_LOGGER.failedToLoadModuleDriver(moduleName, e.getMessage(),
```
rhusar: Same here.

rhusar (changes requested):

bstansberry (approved):

bstansberry:
Thanks @baranowb and @rhusar and @tadamski

PR #19180, opened 2025-09-10
[WFLY-20911] Upgrade Netty to 4.1.127.Final

ivassile:
https://issues.redhat.com/browse/WFLY-20911

Diff: https://github.com/netty/netty/compare/netty-4.1.124.Final...netty-4.1.127.Final

PR #19181, opened 2025-09-10
[WFLY-20911] Upgrade Netty to 4.1.127.Final

ivassile:
https://issues.redhat.com/browse/WFLY-20911

Diff: https://github.com/netty/netty/compare/netty-4.1.124.Final...netty-4.1.127.Final

PR to main: https://github.com/wildfly/wildfly/pull/19180

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -383,7 +383,7 @@
         <version.io.github.resilience4j>2.2.0</version.io.github.resilience4j>
         <version.io.grpc>1.70.0</version.io.grpc>
         <version.io.micrometer>1.15.0</version.io.micrometer>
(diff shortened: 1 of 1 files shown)


bstansberry:
If this passes CI and we merge the one in main, I'll likely go ahead and merge this, but please note we have no plans to release a 37.0.2.

bstansberry:
@ivassile I'm going to close this one as we have no plans for a 37.0.2 and I don't want to add a JIRA version for it to resolve against and mislead people that one might be coming. If we did decide to do one we'd scan what we've fixed in main since 37.0.1, plus run owasp scans and we would pick up this update then.

PR #19183, opened 2025-09-11
[WFLY-20916] Fix dep-diff-workflow_run.yml dependency on wildfly/dep-…

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-20916


jamezp (approved):

bstansberry:
Thanks @emmartins and @jamezp

PR #19185, opened 2025-09-11
[WFLY-20917] Upgrade Hibernate ORM to 6.6.28

ivassile:
https://issues.redhat.com/browse/WFLY-20917

Diff: https://github.com/hibernate/hibernate-orm/compare/6.6.26...6.6.28


scottmarlow:
I think 6.6.29 might be released in the next couple of days so could make sense to upgrade to .29 or .28 is also fine now.

ivassile:
> I think 6.6.29 might be released in the next couple of days so could make sense to upgrade to .29 or .28 is also fine now.

We have to upgrade to 6.6.28 in EAP 8.1 Update 1 and we need this version to be accepted upstream.

scottmarlow (approved):

bstansberry:
Thanks @ivassile and @scottmarlow

PR #18999, opened 2025-06-12
[WFLY-20684] Microprofile Platform 7.1

kabir:
https://issues.redhat.com/browse/WFLY-20684
Includes 
* MP Telemetry 2.1 https://issues.redhat.com/browse/WFLY-20685
* MP OpenAPI 4.1 https://issues.redhat.com/browse/WFLY-20686

Proposal https://github.com/wildfly/wildfly-proposals/pull/736

____

<--- THIS SECTION IS AUTOMATICALLY GENERATED BY WILDFLY GITHUB BOT. ANY MANUAL CHANGES WILL BE LOST. --->

> WildFly issue links:
> * [WFLY-20923](https://issues.redhat.com/browse/WFLY-20923)

<--- END OF WILDFLY GITHUB BOT REPORT --->

More information about the [wildfly-bot[bot]](https://github.com/wildfly/wildfly-github-bot)

PR #19190, opened 2025-09-15
[WFLY-20923] Upgrade to SmallRye Reactive Messaging 4.30.0

kabir:
https://issues.redhat.com/browse/WFLY-20923


bstansberry:
@kabir This failed CI and has a merge conflict so I put the hold label on it just to prevent people looking at it before it's good to go. Please remove it whenever you want.

kabir:
@bstansberry  I'll close this one since it was an attempt at doing the SR RM upgrade without the Mp Platform 7.1 bits.

MP Platform 7.1 passes here https://github.com/wildfly/wildfly/pull/18999, so I think we can go with that one

PR #19199, opened 2025-09-17
[WFLY-20934]: Add a comment in Wild…Fly pom as to why we have artemis-cli as a dependency

ehsavoie:
* Comment added
* Excluding picoli from the transitive dependencies of artemis-cli

Jira: https://issues.redhat.com/browse/WFLY-20934


bstansberry (approved):
Thanks @ehsavoie

PR #19202, opened 2025-09-19
[WFLY-20959] Improve the performance of the org.jboss.as.jpa.hibernate.management.QueryName class

scottmarlow:
https://issues.redhat.com/browse/WFLY-20959

____

> WildFly issue links:
> * [WFLY-20959](https://issues.redhat.com/browse/WFLY-20959)

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/management/QueryName.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/management/QueryName.java
@@ -21,40 +21,6 @@ public class QueryName {
     // query name transformed for display use (allowed to be ugly but should be unique)
     private final String displayQuery;
 
-    // HQL symbol or operators
-    private static final String SQL_NE = "<>";
-    private static final String NE_BANG = "!=";
-    private static final String NE_HAT = "^=";
-    private static final String LE = "<=";
-    private static final String GE = ">=";
-    private static final String CONCAT = "||";
-    private static final String LT = "<";
-    private static final String EQ = "=";
-    private static final String GT = ">";
-    private static final String OPEN = "(";
-    private static final String CLOSE = ")";
-    private static final String OPEN_BRACKET = "[";
-    private static final String CLOSE_BRACKET = "]";
-    private static final String PLUS = "+";
-    private static final String MINUS = "-";
-    private static final String STAR = "*";
-    private static final String DIV = "/";
-    private static final String MOD = "%";
-    private static final String COLON = ":";
-    private static final String PARAM = "?";
-    private static final String COMMA = ",";
-    private static final String SPACE = " ";
-    private static final String TAB = "\t";
-    private static final String NEWLINE = "\n";
-    private static final String LINEFEED = "\r";
-    private static final String QUOTE = "'";
-    private static final String DQUOTE = "\"";
-    private static final String TICK = "`";
-    private static final String OPEN_BRACE = "{";
-    private static final String CLOSE_BRACE = "}";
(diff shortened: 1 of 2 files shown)


scottmarlow:
I think the next test to try is using regular expressions but didn't try that yet.  I created this pr to share the initial change.  Running with a reasonable large query (around 81,920 characters), test results:

old way:

real    0m9.217s
user    0m9.205s
sys     0m0.049s

Updated to use String.replace:

real    0m0.257s
user    0m0.222s
sys     0m0.074s

Previously I tried running the String.replace way with a Query of size 83,886,080 characters took close to a second but the old way took several minutes so I cancelled the run.

scottmarlow:
> I think the next test to try is using regular expressions but didn't try that yet.

Actually, will try a single loop version next and if that isn't fast enough will try a lambda variation of that.

scottmarlow:
Single pass through the input string version (with use of StringBuilder):


real    0m0.132s
user    0m0.139s
sys     0m0.057s

Previous change with multiple calls to String#replace:
real    0m0.262s
user    0m0.222s
sys     0m0.084s

Note that input string query length is 83968 characters which is probably longer than typical queries.

Main test driver:
```
    public static void main(String[] args) {
        String testvalue =
                "query name select * from mytable where mytable.id <> != ^= = >= , , , , , , ,\"\" {}";
        for (int loop = 0; loop < 10; loop++){
            testvalue = testvalue + testvalue;
        }
        System.out.println("input string length= " + testvalue.length());

        for (int loop = 0; loop < 100; loop++) {
            QueryName queryName = new QueryName(testvalue);
        }
}
```

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/management/QueryName.java
```diff
@@ -138,66 +103,146 @@ public int hashCode() {
...
+        final StringBuilder out = new StringBuilder(queryLength);
+        for (int index = 0; index < queryLength; index++) {
+            char current = query.charAt(index);
+            char next = index + 1 < queryLength ? query.charAt(index + 1) : 0;
+            switch (current) {
+                case '<':
+                    if (next == '>') {
+                        index++;
+                        out.append(NOT_EQUAL__);
+                    } else if (next == '=') {
```
bstansberry, line 119: there's a missing index++ in this block

bstansberry (changes requested):

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/management/QueryName.java
scottmarlow: Thanks Brian!  Fixed!

bstansberry (approved):

bstansberry:
@scottmarlow It sounds like the JIRA may have a greater scope than this. OTOH there's no reason to block merging this while waiting for further improvement. Perhaps create a JIRA subtask and change the issue of this. Or clone WFLY-20799 and convert WFLY-20799 into a subtask or 'incorporated by' the new 'parent'?

scottmarlow:
> @scottmarlow It sounds like the JIRA may have a greater scope than this. OTOH there's no reason to block merging this while waiting for further improvement. Perhaps create a JIRA subtask and change the issue of this. Or clone WFLY-20799 and convert WFLY-20799 into a subtask or 'incorporated by' the new 'parent'?

Thanks @bstansberry, great idea!  I squashed this pr as well and updated it to use the new subtask issue.

bstansberry:
Thanks @scottmarlow

PR #19204, opened 2025-09-19
Ignore intermittent container start failures tracked by WFLY-20945

bstansberry:
No JIRA; this disables a test the fix for which is tracked by a JIRA mentioned in the diff.  

Ignore the link added by wildfly-bot below.

____

> WildFly issue links:
> * [WFLY-20945](https://issues.redhat.com/browse/WFLY-20945)

### DIFF
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/RunArtemisAmqpSetupTask.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/RunArtemisAmqpSetupTask.java
@@ -20,7 +20,9 @@ import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.PathElement;
 import org.jboss.as.controller.operations.common.Util;
 import org.jboss.as.model.test.ModelTestUtils;
+import org.jboss.as.test.shared.IntermittentFailure;
 import org.jboss.dmr.ModelNode;
+import org.junit.AssumptionViolatedException;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.MountableFile;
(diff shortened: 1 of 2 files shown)


kabir (approved):

bstansberry:
It seems this doesn't work; the test still fails.

@jamezp Didn't we change things to be able to "ingore" test classes via AssumptionViolatedException thrown from a ServerSetup? I suspect that here I've forgotten some of the ingredients of the special sauce.

jamezp:
I'll need to dig into this because yes we do propagate it up, which is what is happening here, but it's not being caught and handled elsewhere.

jamezp:
@bstansberry I see it. It's in the setup itself https://github.com/bstansberry/wildfly/blob/e857f9e09f530205b4a4d1bbb30817658f8b7ac4/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/RunArtemisAmqpSetupTask.java#L109. It catches the `AssumptionViolatedException` and wraps it in a `RuntimeException`. The setup is trying to be too clever :)

bstansberry:
Doh! Thanks, James!

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/RunArtemisAmqpSetupTask.java
```diff
@@ -97,6 +107,9 @@ public void setup(ManagementClient managementClient, String containerId) throws
             } catch (Exception ex) {
                 e.printStackTrace();
             }
+            if (e instanceof AssumptionViolatedException ave) {
+                throw ave;
+            }
```
jamezp, line 112: I'll file a WFARQ issue, but it looks like an issue that `tearDown()` is not invoked by WildFly Arquillian itself.

jamezp (approved):

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/reactive/RunArtemisAmqpSetupTask.java
bstansberry: @jamezp Changing that could be problematic as I expect many impls are written where doing that would cause problems.

I suppose if any exception from the teardown was caught and logged but not allowed to propagate, it could be ok.

jamezp: Yeah. I need to take a closer look, but it seems in some cases the `tearDown()` would happen, but not in all. I just want to make sure it's consistent.

PR #19205, opened 2025-09-22
[WFLY-20951] Take into account the new Jakarta Authorization capability.

darranl:
I am going to start as a draft but this is independently mergable without the rest.

Longer term we may want to introduce a Jakarta Authorization layer / extension - for now the Elytron extension will register Jakarta Authorization 3.0 for Jakarta EE 11 if the integration supports it and if so will record a capability:

https://github.com/darranl/wildfly-core/blob/WFCORE-7322/elytron/src/main/java/org/wildfly/extension/elytron/ElytronDefinition.java#L490-L497

This means we have a capability we can detect but we don't need to depend on a Service for EE 11.

https://issues.redhat.com/browse/WFLY-20951

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/security/JaccEarDeploymentProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/security/JaccEarDeploymentProcessor.java
@@ -5,6 +5,9 @@
 
 package org.jboss.as.ee.security;
 
+import static org.jboss.as.ee.subsystem.EeCapabilities.ELYTRON_JACC_CAPABILITY;
+import static org.jboss.as.ee.subsystem.EeCapabilities.ELYTRON_JAKARTA_AUTHORIZATION;
+
 import jakarta.security.jacc.PolicyConfiguration;
 
 import org.jboss.as.controller.capability.CapabilityServiceSupport;
@@ -27,10 +30,9 @@ import org.jboss.msc.service.ServiceTarget;
  */
 public class JaccEarDeploymentProcessor implements DeploymentUnitProcessor {
 
-    private final String jaccCapabilityName;
+    private volatile boolean deployed = false;
 
-    public JaccEarDeploymentProcessor(final String jaccCapabilityName) {
-        this.jaccCapabilityName = jaccCapabilityName;
+    public JaccEarDeploymentProcessor() {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


bstansberry:
@darranl I suggest a few words in the DUPs and in EeCapabilities noting the distinction between the 2 capabilities.

If that would be too wordy, then please spell it out clearly in the JIRA description.

darranl:
Thanks @bstansberry yes I can add that.

ee/src/main/java/org/jboss/as/ee/security/JaccEarDeploymentProcessor.java
```diff
@@ -27,10 +30,9 @@
  */
 public class JaccEarDeploymentProcessor implements DeploymentUnitProcessor {
 
-    private final String jaccCapabilityName;
+    private volatile boolean deployed = false;
```
bstansberry: This needs a usage comment, because it's fragile. Maybe it needs a rename?

The instance of this class is not bound to a deployment, so any state it holds is globally applicable. But 'deployed' intuitively means something applicable to a particular deployment.

Maybe something like

// Tracks whether we've detected that one of the capabilities indicating a need for Jakarta Authorization is present. 
// Volatile because we can only detect that the first time we process an ear in deploy(...)
private volatile boolean detectedJakartaAuthentication = false;

The presence of absence of the capability won't change without a reload (otherwise this field isn't valid at all), so you could also change the 'deploy' impl to just skip looking for the capabilities if this is 'true'.

bstansberry (changes requested):

ee/src/main/java/org/jboss/as/ee/security/JaccEarDeploymentProcessor.java
darranl: Just checking this, the undeploy method on this one is no-op - can probably get rid of this boolean for now

bstansberry: +1. 

AbstractSecurityDeployer and subclasses should be renamed. They are AbstractJaccServiceFactory etc and have a 'createJaccService" method, not a 'deploy' method. They don't deploy anything in any sense other than adding an object to the heap.

That doesn't need to be in scope for this PR though.

darranl: +1 I would actually like to see what we can pull out here, there is a little but too much code IMO in the WildFly repo.

If we could refactor and pull some out it would makeit easier to use with Undertow standalone as well which wouild also make developing and testing this a lot easier.

bstansberry (approved):
@darranl LGTM. If you can put a bit more detail in the WFLY description that would be good.

darranl:
Thank you @bstansberry FYI I have copied the javadoc text into the issue, let me know if you think we need anything more.

PR #19206, opened 2025-09-22
WFLY-20947, WFLY-20948 Fix memory leak of expirable distributed state following suspend/resume

pferraro:
https://issues.redhat.com/browse/WFLY-20947
https://issues.redhat.com/browse/WFLY-20948

### DIFF
--- a/clustering/common/pom.xml
+++ b/clustering/common/pom.xml
@@ -66,6 +66,10 @@
             <artifactId>wildfly-clustering-function</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.wildfly.clustering</groupId>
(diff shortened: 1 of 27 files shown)


bstansberry:
@pferraro This LGTM but these "trying to archive a 3+ GB server.log" CI failures are odd. I don't see why this PR would change what's being logged.

pferraro:
Should be sorted now.

bstansberry (approved):

bstansberry:
Thanks @pferraro 

@tadamski This looked good to me and I wanted to get this into WF 38 Beta so it can bake, and into main ASAP so if it caused issue we might notice before the Beta. If you have any concerns with the change, @pferraro can address those in a follow-up.

PR #19213, opened 2025-09-24
[WFLY-20960] Default to a lower hibernate.statistics.query_max_size value of 200 unless application overrides it

scottmarlow:
https://issues.redhat.com/browse/WFLY-20960

### DIFF
--- a/docs/src/main/asciidoc/_developer-guide/Jakarta_Persistence_Reference_Guide.adoc
+++ b/docs/src/main/asciidoc/_developer-guide/Jakarta_Persistence_Reference_Guide.adoc
@@ -678,6 +678,9 @@ not already set in persistence unit definition):
 |=======================================================================
 |Property |Purpose
 
+|hibernate.statistics.query_max_size = 200 |only set if the
+application didn't specify a value for hibernate.statistics.query_max_size.
+
 |hibernate.id.new_generator_mappings| Can no longer be set to false by applications as per the Hibernate 6 change to only support hibernate.id.new_generator_mappings=true.
 
 |hibernate.transaction.jta.platform= instance of
@@ -699,15 +702,11 @@ name |Is set to the application name + persistence unit name
 across all application deployments on the AS instance).  
 This replaces the `hibernate.ejb.entitymanager_factory_name` property which is no longer supported.
 
-|hibernate.query.jpaql_strict_compliance=true |
-
 |hibernate.auto_quote_keyword=false |
 
 |hibernate.implicit_naming_strategy=org.hibernate.boot.model.naming.ImplicitNamingStrategyJpaCompliantImpl |
 
-|hibernate.model.generator_name_as_sequence_name=true |
-
-|hibernate.jpa.compliance=true | Sets all of the various hibernate.jpa.* properties to true (hibernate.jpa.compliance.transaction, hibernate.jpa.compliance.closed,hibernate.jpa.compliance.query, hibernate.jpa.compliance.list, hibernate.jpa.compliance.caching, hibernate.jpa.compliance.proxy
+|hibernate.jpa.compliance=true | Sets all of the various hibernate.jpa.* properties to true (including hibernate.jpa.compliance.transaction, hibernate.jpa.compliance.closed,hibernate.jpa.compliance.query, hibernate.jpa.compliance.list, hibernate.jpa.compliance.caching, hibernate.jpa.compliance.proxy, hibernate.jpa.compliance.global_id_generators, hibernate.jpa.compliance.load_by_id, 
 
 |hibernate.enable_lazy_load_no_trans=false |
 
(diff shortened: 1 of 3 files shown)


bstansberry:
@yrodiere or @sebersole Please review. You guys and Scott are the SMEs who'd know best about correct defaults!

yrodiere:
Makes sense to me, but this strikes me as something that we'd better fix upstream as well? WDYT @sebersole?

scottmarlow:
Starter zulip discussion about ^ via https://hibernate.zulipchat.com/#narrow/channel/132094-hibernate-orm-dev/topic/hibernate.2Estatistics.2Equery_max_size.20defaults.20to.205000/with/541256274

bstansberry:
@scottmarlow Even if ORM changes their default, I'm guessing they wouldn't in a 6.6 micro, so for quite a while WF will have a different default. Is there any place in the WF docs to note this?

scottmarlow:
> @scottmarlow Even if ORM changes their default, I'm guessing they wouldn't in a 6.6 micro, so for quite a while WF will have a different default. Is there any place in the WF docs to note this?

Good point, yes, we should update https://docs.wildfly.org/37/Developer_Guide.html#hibernate-properties to include `hibernate.statistics.query_max` and mention that WildFly will only set it if the application didn’t specify a value for `hibernate.statistics.query_max`.

scottmarlow:
WF docs are updated to note this.

bstansberry:
I'm going to put the 'hold' label on this, given the very interesting discussion in the Hibernate zulip thread Scott linked above.

The label basically helps take it off the radar screen of mergers. Any of the dozens of folks with GitHub triage perms can remove it whenever that's no longer appropriate.

This needs further discussion but I don't think comments on this PR are the best place. WildFly zulip or WildFly-specific information on the Hibernate zulip thread (e.g. about how the WildFly management API is how the Hibernate stats are exposed to metrics consumers) are more appropriate.

@jasondlee FYI re this topic, particular the Hibernate discussion. It's fundamentally about observability.

bstansberry:
@scottmarlow Is this still relevant now that we are on Hibernate 7.4? That is, did anything change in ORM to remove the need for this? If not, please close it.

scottmarlow:
> Makes sense to me, but this strikes me as something that we'd better fix upstream as well? 

I'm not exactly sure of what the upstream change would be other than requiring the application to indicate a default setting instead of expecting the DEFAULT_QUERY_STATISTICS_MAX_SIZE default of `5000`.  Failing application deployments that enable statistics but do not specify DEFAULT_QUERY_STATISTICS_MAX_SIZE doesn't sound like much of an improvement (not that anyone suggested that we should fail such deployments).

> @scottmarlow Is this still relevant now that we are on Hibernate 7.4? 

I think this change is still relevant not that we have heard any other complaints as users can override the default by including the `hibernate.statistics.query_max_size` property set to smaller size in their persistence.xml.  

This setting controls the number of {@link org.hibernate.stat.QueryStatistics} entries that will be stored by the Hibernate {@link org.hibernate.stat.Statistics} object.  Some possible use cases 1 + 2:

### Current use case with application deployments that have more than 5000 Hibernate queries:
By default there will be 5000 entries returned when viewing the top (hibernate.statistics.query_max_size) slowest Hibernate queries.  There will be 5000 entries in the UI for viewing the slowest queries.

### Use case with the proposed change (hibernate.statistics.query_max_size will be changed to `200`) and application deployments that have more than 5000 Hibernate queries:
By default there will be 200 entries returned when viewing the top (hibernate.statistics.query_max_size) slowest Hibernate queries.  There will be 200 entries in the UI for viewing the slowest queries.

> That is, did anything change in ORM to remove the need for this? If not, please close it.

Impact of merging this change is WildFly uses a smaller default 200 instead of 5000.  Applications would still be able to override the WildFly setting for `hibernate.statistics.query_max_size` by including `hibernate.statistics.query_max_size` in the persistence.xml.

scottmarlow:
> ### Current use case with application deployments that have more than 5000 Hibernate queries:
> 
> By default there will be 5000 entries returned when viewing the top (hibernate.statistics.query_max_size) slowest Hibernate queries. There will be 5000 entries in the UI for viewing the slowest queries.

^ has been the behavior for a long time and is still unchanged upstream on Hibernate ORM `main` branch.  The initial choice of 5000 or whatever it was in some Hibernate ORM 3.* version originally was an arbitrary limit that mostly impacted the typical application console ui (e.g. JMX console typically with JBoss AS 4).  If we lower the default choice to 200, that could confuse those looking at the top 5000 (or N greater than 200) slowest Hibernate queries in their application.

> ### Use case with the proposed change (hibernate.statistics.query_max_size will be changed to `200`) and application deployments that have more than 5000 Hibernate queries:
> 
> By default there will be 200 entries returned when viewing the top (hibernate.statistics.query_max_size) slowest Hibernate queries. There will be 200 entries in the UI for viewing the slowest queries.

^ as mentioned above could lead to some users asking why their management console is not showing as many of the slowest queries but honestly I wanted the feature years ago and do not recall much feedback on it.  If we change the default, we might get feedback from users asking how to show a large number of slow queries again which would be good as we we could tell them to update their persistence.xml to choose a different `hibernate.statistics.query_max_size` setting as the application settings will override the WildFly default for `hibernate.statistics.query_max_size`.

scottmarlow:
Having added a small wall of text above, I'm also now reading the long zulip conversation to catch up on that. :-)

After reviewing the linked zulip conversation I am thinking the more interesting part to solve is in Hibernate ORM.  So will close this as the max size of 200 will not lead to just the top 200 slowest queries being displayed (instead it will reduce the accuracy of displayed queries).

PR #19222, opened 2025-09-25
[WFLY-20967] Upgrade Jakarta Mail from 2.1.3 to 2.1.4

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-20967

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -429,7 +429,7 @@
         <version.jakarta.inject.jakarta.inject-api>2.0.1</version.jakarta.inject.jakarta.inject-api>
         <version.jakarta.jms.jakarta-jms-api>3.1.0</version.jakarta.jms.jakarta-jms-api>
         <version.jakarta.json.bind.api>3.0.1</version.jakarta.json.bind.api>
-        <version.jakarta.mail-api>2.1.3</version.jakarta.mail-api>
+        <version.jakarta.mail-api>2.1.4</version.jakarta.mail-api>
         <version.jakarta.mvc.jakarta-mvc-api>2.1.0</version.jakarta.mvc.jakarta-mvc-api>
         <version.jakarta.persistence>3.1.0</version.jakarta.persistence>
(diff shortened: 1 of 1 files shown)


yersan:
Failures here are unexpected to me, a reproducer outside of WildFly with Jakarta mail `2.1.4` and Angus Mail `2.0.4` passes. It looks like there are other details behind the scenes that are currently hidden to me, I'll take a closer look

jamezp:
I don't know much about the setup in WildFly, but these two PR's may be of interest:
* https://github.com/jakartaee/mail-api/pull/778
* https://github.com/jakartaee/mail-api/pull/779

My gut tells me something in https://github.com/jakartaee/mail-api/pull/779 is broken. The `MailSessionCustomProviderTestCase` sets a custom provider in the `META-INF/javamail.providers`. I bet changing it to `META-INF/jakarta.providers` may fix it.

OT, but `jakarta.providers` probably should have been `jakarta.mail.providers` as "providers" are very generic. Probably too late now though.

bstansberry:
If that's the problem, we should hold off on merging this, file an issue and wait for a fix.

I'm suspicious of https://github.com/jakartaee/mail-api/pull/779/files#diff-5cbee3986400643f15064c1c111133ea02d864c75ed3ad777d192159f8f6ea89R1268

If you look at the diff with whitespace changes ignored, it stands out:

https://github.com/jakartaee/mail-api/pull/779/files?w=1#diff-5cbee3986400643f15064c1c111133ea02d864c75ed3ad777d192159f8f6ea89R1268

Before, for a given 'name' it would use ClassLoader.getResources(name) to find all URLs and then iterate over them all. Now if a given name results in > 1 URL it will stop after the first.

This seems unrelated to the basic change of that commit which was to search for more names. Instead that line changes how individual names are handled.

I suspect when the test is running there is more than 1 "META-INF/javamail.providers" file visible to the TCCL, and the one our test is expecting to be processed is being skipped.

bstansberry:
@jamezp -- I've been commenting on this, but @yersan has already identified the problem and filed an issue:

https://github.com/jakartaee/mail-api/issues/812

PR #19227, opened 2025-09-26
Fix for WFLY-20979, microprofile-reactive-messaging layer has a wrong name in its spec

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-20979

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-reactive-messaging/layer-spec.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-reactive-messaging/layer-spec.xml
@@ -4,7 +4,7 @@
   ~ SPDX-License-Identifier: Apache-2.0
   -->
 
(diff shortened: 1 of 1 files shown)


bstansberry:
@jfdenise What is the impact of this bug? I suspect very little because the microprofile-reactive-messaging-amqp layer depends on microprofile-reactive-messaging and our docs for MP RM describe using the microprofile-reactive-messaging layer. So if the layer didn't work I'd expect we'd have heard that. So maybe things work because of the directory name?

Of course what you're fixing is wrong. I'm just asking because if the 'wrongness' has little effect, then correcting it is unlikely to break anything too.

jfdenise:
@bstansberry , the impact is seen in the future catalog where we read layer specs. This exposes a wrong layer. The wrongness has no impact at provisioning time. I have added a check to Galleon at feature-pack build time to identify such disconnect in the future.

bstansberry (approved):

bstansberry:
Thanks @jfdenise

PR #19232, opened 2025-09-29
[WFLY-20988] Followup improvements of Jakarta EE 11 JPA / CDI integration for WildFly Preview

scottmarlow:
https://issues.redhat.com/browse/WFLY-20988

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
@@ -770,7 +770,4 @@ public interface JpaLogger extends BasicLogger {
     @LogMessage(level = INFO)
     @Message(id = 79, value = "EntityManagerFactory CDI bean (for %s) will not have persistence unit name.  In order for the EntityManagerFactory CDI bean to be named remove duplicate copies of persistence unit %s.")
     void willNotNameEntityManagerFactoryBean(String scopedPersistenceUnitName, String persistenceUnitName);
-
-    @Message(id = 80, value = "Found more than one %s service for application deployment %s")
-    RuntimeException foundMoreThanOneCdiExtensionService(Class<?> serviceClassName, ClassLoader module);
 }
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
@@ -16,7 +16,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
-import java.util.ServiceLoader;
 
 import jakarta.persistence.SynchronizationType;
 import jakarta.persistence.ValidationMode;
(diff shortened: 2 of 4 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
```diff
@@ -1205,6 +1205,10 @@ private static IntegrationWithCDIBagImpl registerPersistenceAfterBeanDiscovery(D
                         // putAttachment will return null if the attachment was not set yet, in which case we will also register the CDI extension with Weld
                         if (null == deploymentUnit.putAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY, persistenceCdiExtension))
                             weldCapability.get().registerExtensionInstance(persistenceCdiExtension, deploymentUnit);
+                        else {
+                            // ensure that other threads use the service that was loaded in a different thread.
+                            persistenceCdiExtension = deploymentUnit.getAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY);
```
scottmarlow: This change ensures that the other threads register the persistence units with the persistenceCdiExtension (service) actually registered as a CDI extension (e.g. via the `weldCapability.get().registerExtensionInstance(persistenceCdiExtension, deploymentUnit)` call in a different thread)

jamezp: Sorry, it doesn't look like I finished my review on this.

This still is quite correct. When you have time to discuss it, ping me in chat. I need to understand better. If concurrency is a concern we need have some kind of shared lock per-deployment. I'm not sure that concurrency is a concern here though as I see other places in the source which don't seem concerned about it.

scottmarlow: For reference the previous feedback was:
> The issue though is that if deploymentUnit.putAttachment(JpaAttachments.AFTER_BEAN_DISCOVERY_ATTACHMENT_KEY, persistenceCdiExtension) returns a value, you'd be executing the register() on the wrong instance.

Yes, I see it now!  Originally, we had a synchronize on the top level deployment unit (e.g. ear, standalone war,  standalone jar):

```
synchronized (deploymentUnit) { // protect against multiple subdeployment threads registering the persistenceCdiExtension at same time
```

We could use a different object to lock on than the (top most level) DeploymentUnit but yes having a shared lock per (top level) deployment is the right idea.

jpa/subsystem/src/main/java/org/jboss/as/jpa/messages/JpaLogger.java
```diff
@@ -771,6 +771,6 @@ public interface JpaLogger extends BasicLogger {
     @Message(id = 79, value = "EntityManagerFactory CDI bean (for %s) will not have persistence unit name.  In order for the EntityManagerFactory CDI bean to be named remove duplicate copies of persistence unit %s.")
     void willNotNameEntityManagerFactoryBean(String scopedPersistenceUnitName, String persistenceUnitName);
 
-    @Message(id = 80, value = "Found more than one %s service for application deployment %s")
-    RuntimeException foundMoreThanOneCdiExtensionService(Class<?> serviceClassName, ClassLoader module);
+//     @Message(id = 80, value = "Found more than one %s service for application deployment %s")
+//     RuntimeException foundMoreThanOneCdiExtensionService(Class<?> serviceClassName, ClassLoader module);
```
jamezp: In this specific case we can actually delete this since it has never been in a final release.

jamezp (approved):

bstansberry:
@jamezp @scottmarlow  I removed the ready-for-merge label for this because yesterday I ran two other WFP+JPA PRs in a combo PR against the EE 11 branch, and this wasn't part of it, since it wasn't approved. Those tests passed.

So I'll merge those two and then get more testing of this to ensure the combo isn't a problem

jamezp:
Ah, sorry. I should have thought about kicking off an EE11 job.

bstansberry:
That's ok; I'm mostly wanting to be sure the combo of the 3 related PRs is ok. I'll do that job now.

bstansberry:
https://github.com/wildfly/wildfly/pull/19269

bstansberry:
Thanks @scottmarlow and @jamezp

PR #19238, opened 2025-09-30
[WFLY-20992] Add a property to skip the installation Manager service …

yersan:
…installation on the testsuite

Jira issue: https://issues.redhat.com/browse/WFLY-20992


yersan:
@honza-kasik Could you approve this if that resolves the RPM-based installation tests?

You can skip the Installation Manager service installation by adding `-Dinstallation.manager.services.install.skip` to your maven goals.

We discussed skipping specific tests, but that approach will force us to duplicate the existing ones, so just do not install the expected module that enables the issue.

honza-kasik:
@yersan I was able to verify that with `-Dinstallation.manager.services.install.skip` all tests in `org.jboss.as.test.integration.mgmt.access.ReadFullModelTestCase` pass. Thank you!

bstansberry:
Thanks @yersan and @honza-kasik

PR #19239, opened 2025-09-30
[WFLY-20993] Upgrade Jakarta Mail from 2.1.3 to 2.1.5

yersan:
Jira issue: https://issues.redhat.com/browse/WFLY-20993

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -429,7 +429,7 @@
         <version.jakarta.inject.jakarta.inject-api>2.0.1</version.jakarta.inject.jakarta.inject-api>
         <version.jakarta.jms.jakarta-jms-api>3.1.0</version.jakarta.jms.jakarta-jms-api>
         <version.jakarta.json.bind.api>3.0.1</version.jakarta.json.bind.api>
-        <version.jakarta.mail-api>2.1.3</version.jakarta.mail-api>
+        <version.jakarta.mail-api>2.1.5</version.jakarta.mail-api>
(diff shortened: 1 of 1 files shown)


bstansberry:
@yersan It would be good to run a build of https://github.com/jakartaee/jaf-api/pull/187 + current Angus + our test that showed the provider loading problem before we merge this.

AIUI this puts us in a situation where we can't update Angus. And of course not merging this puts us in a situation where we can't update the mail API jar. :) If we're confident that PR ^^^ will fix providing loader problem, that's more comfortable.

ropalka (approved):
Approved from JBossWS point of view

yersan:
Just for the record, this is still on hold since the backward compatibility issue has not been fixed yet. I'm moving this as a draft by now.

@bstansberry There have not been any CVEs in Angus Mail yet, but due to the tradeoff, I would be more inclined to consume fixes even if that would introduce this backward compatibility

yersan:
@bstansberry The final argument for moving forward with this is the expectation that users will migrate their custom mail resources from the javaee to the jakartaee namespace as part of the overall Java EE to Jakarta EE transition. See: https://github.com/jakartaee/mail-api/issues/789#issuecomment-3945485087

So, I guess it is up to us to follow that guideline. What are your thoughts on this? I'm not sure if it's possible for us to somehow transition those files on our end during the deployment phase to get rid of this backward compatibility issue. However, from my point of view, the argument above should be enough to justify expecting users to do it.

PR #19240, opened 2025-10-01
[WFLY-20995] Account for the inject env entry in the test assertion

bstansberry:
https://issues.redhat.com/browse/WFLY-20995

@jasondlee Perhaps this env entry could just be removed as I'm not sure it's adding anything, unless the injection was somehow a factor in triggering duplicate PostConstruct calls. But I didn't want to think hard about that so I just made use of the value.


jasondlee (approved):
I'm not sure it adds anything either, but, to be honest, I don't remember this test at all. :) Git blames me for it, but it must have come from some sort or reorganization. That said, it seems harmless enough, and does the make test actually test what is purports to do. Does it add value? I'm with you: I'm not sure. :)

PR #19243, opened 2025-10-01
 [WFLY-20998] Update the producers to only be added if there was not already a producer available provided by the application.

jamezp:
Issue: https://issues.redhat.com/browse/WFLY-20998

This also adds more testing for the Jakarta Persistence and CDI integration as well as some documentation in a separate commit.

### DIFF
--- a/docs/src/main/asciidoc/WildFly_and_WildFly_Preview.adoc
+++ b/docs/src/main/asciidoc/WildFly_and_WildFly_Preview.adoc
@@ -238,3 +238,57 @@ embedded broker is still supported. We've added to the $WILDFLY_HOME/docs/exampl
 * WildFly Preview includes the `jakarta-data` subsystem in its out-of-the-box `standalone.xml`, `standalone-ha.xml`, `standalone-full.xml`, `standalone-full-ha.xml`, `standalone-microprofile.xml` and `standalone-microprofile-ha.xml` configuration files. It also include the subsystem in the out-of-the-box `domain.xml` configuration file's `default`, `ha`, `full` and `full-ha` profiles. Standard WildFly includes support for the `jakarta-data` subsystem but does not include it in any out-of-the-box-configuration file. link:https://jakarta.ee/specifications/data/[Jakarta Data] is a new Jakarta specification that will be part of Jakarta EE 11.
 * WildFly Preview includes a new xref:Admin_Guide.adoc#Feature_stability_levels[`preview` stability] `org.wildfly.extension.vertx` extension and its `vertx` subsystem, along with a new `vertx` Galleon layer so you can provision it in a slimmed server. This subsystem can be used to configure the Vert.x instance used by our OpenTelemetry integration.
 * The Hibernate ORM integration used by the link:Developer_Guide{outfilesuffix}#JPA_Reference_Guide[JPA subsystem's] Hibernate Search feature supports using outbox polling as a coordination strategy for automatic indexing.
+
+== Jakarta Persistence 3.2+
+
+In https://jakarta.ee/specifications/persistence/3.2/[Jakarta Persistence 3.2] and https://jakarta.ee/specifications/platform/11/[Jakarta EE 11+],
+there is a requirement to produce the following types for CDI injection:
+
+* `CriteriaBuilder`
+* `EntityManager`
+* `EntityManagerFactory`
+* `Metamodel`
+* `PersistenceUnitUtil`
+* `SchemaManager`
+
+As noted in the https://jakarta.ee/specifications/platform/11/jakarta-platform-spec-11.0#a441[specification], the
+`EntityManager` has a default scope of `@TransactionScoped`. The `EntityManagerFactory` has a scope of `@ApplicationScoped`.
+All other types have a scope of `@Dependent`.
+
+In the `persistence.xml`, each persistence unit can define their own scope and own qualifiers. It's important to note
+if you choose to use CDI for your Jakarta Persistence application, you must create CDI qualifiers if you have more than
+one persistence unit defined in your `persistence.xml` file. Without this you will likely see deployment failures due
+to ambiguous beans.
+
+.Example persistence.xml
+[source,xml]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<persistence xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+             xmlns="https://jakarta.ee/xml/ns/persistence"
+             xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence
+        https://jakarta.ee/xml/ns/persistence/persistence_3_2.xsd"
+             version="3.2">
+    <persistence-unit name="auth">
+        <jta-data-source>java:jboss/datasources/AuthDS</jta-data-source>
+        <qualifier>org.wildfly.example.AuthDb</qualifier>
+    </persistence-unit>
+    <persistence-unit name="default">
+        <jta-data-source>java:jboss/datasources/defaultDS</jta-data-source>
+    </persistence-unit>
+</persistence>
+----
+
+At most one persistence unit is allowed to have no qualifier defined in the `persistence.xml`. This will be the default
+scoped `EntityManagerFactory` and `EntityManager`.
+
+NOTE: If you already have CDI producers defined in your application to inject any of these types, they should continue
+to work. However, you may be able to remove these producers and rely on the default behavior.
+
+=== Special Notes
+
+It should be noted that there is some ambiguity in the Jakarta EE 11 specification when defining beans for the
+`EntityManagerFactory`. It's noted that bean name is given by the name of the persistence unit. In most cases this
+should be fine. However, if multiple persistence units with the same name exist in a deployment,
+we've opted to not give the bean a name. This will only affect you if you want to use Jakarta Expression Language to
+get the `EntityManagerFactory` and you've got more than one persistence unit with the same name.
--- a/jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
+++ b/jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
@@ -9,16 +9,26 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Collectors;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.Dependent;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.spi.AfterBeanDiscovery;
+import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
+import jakarta.enterprise.inject.spi.ProcessProducerField;
+import jakarta.enterprise.inject.spi.ProcessProducerMethod;
 import jakarta.enterprise.inject.spi.configurator.BeanConfigurator;
 import jakarta.persistence.Cache;
 import jakarta.persistence.EntityManager;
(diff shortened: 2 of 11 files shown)


jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -371,6 +431,66 @@ private void schemaManager(
...
+     * @param qualifiers the qualifiers for the producer
+     *
+     * @return {@code true} if a producer already exists, otherwise {@code false}
+     */
+    private boolean producerExists(final Class<?> type, final Collection<String> qualifiers) {
+        final var allQualifiers = foundProducers.get(type);
+        if (allQualifiers != null) {
+            for (final var q : allQualifiers) {
+                if (qualifiers.isEmpty()) {
+                    if (DEFAULT_QUALIFIERS.containsAll(q)) {
```
scottmarlow, line 482: So `Any` is expected but we are really checking for `Default`?

jamezp: `@Any` is on every bean, https://jakarta.ee/specifications/cdi/4.1/apidocs/jakarta/enterprise/inject/any and https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#builtin_qualifiers.

Maybe on the other check I should explicitly add `@Any` for cases it might not be there. I've wondered if I have that check correct. I almost wonder if it should be `qualifers.containsAll(q)` instead of the other way around.

jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java
```diff
@@ -371,6 +431,66 @@ private void schemaManager(
...
+                        return true;
+                    }
+                }
+                if (q.containsAll(qualifiers)) {
```
scottmarlow, line 486: So only if the qualifiers specified in the persistence unit are all specified in the existing producer would we return true meaning the producer already exists?  

What if there are no qualifiers specified in the persistence unit?  Do we need a separate check for the producer existing?

jamezp: > So only if the qualifiers specified in the persistence unit are all specified in the existing producer would we return true meaning the producer already exists?

Correct. We only want to create a producer if one was not specified already.

> What if there are no qualifiers specified in the persistence unit? Do we need a separate check for the producer existing?

Yes, that is were the `qualifiers.isEmpty()` check is done against the `DEFAULT_QUALIFIERS`.

scottmarlow: This has been a learning experience for me working on the CDI Persistence integration.  I now (after this pr + discussion) better understand some of what you said previously about qualifiers + default handling.

jamezp: Feel free to ask anything else as well. It means I either have to know it or learn it and I'm more than willing to learn it :)

docs/src/main/asciidoc/WildFly_and_WildFly_Preview.adoc
```diff
@@ -238,3 +238,57 @@ embedded broker is still supported. We've added to the $WILDFLY_HOME/docs/exampl
...
+* `PersistenceUnitUtil`
+* `SchemaManager`
+
+As noted in the https://jakarta.ee/specifications/platform/11/jakarta-platform-spec-11.0#a441[specification], the
+`EntityManager` has a default scope of `@TransactionScoped`. The `EntityManagerFactory` has a scope of `@ApplicationScoped`.
+All other types have a scope of `@Dependent`.
+
+In the `persistence.xml`, each persistence unit can define their own scope and own qualifiers. It's important to note
+if you choose to use CDI for your Jakarta Persistence application, you must create CDI qualifiers if you have more than
+one persistence unit defined in your `persistence.xml` file. Without this you will likely see deployment failures due
```
scottmarlow, line 260: For several years applications have been able to specify persistence unit hint `wildfly.jpa.default-unit` set to true to choose a default persistence unit which resolved deployment failures (e.g. @PersistenceContext without the pu name) and that hint is checked if there are no qualifiers specified in the application persistence unit definition via https://github.com/wildfly/wildfly/blob/main/jpa/persistence-cdi-extension/src/main/java/org/jboss/as/jpa/beanmanager/IntegratePersistenceAfterBeanDiscovery.java#L122.

I'm not sure if we should point this out though as it is better for the application to use qualifiers but we could restate that if the application sees ambiguous beans deployment errors, be sure to add CDI qualifiers and specify those in the application persistence.xml files.

jamezp: My personal preference is it should be done the CDI way and qualifiers should be used. That said, it's ultimately up to you :)

What might be useful if it's not already done somewhere is to document all the hints available in WildFly, maybe not all the ones that are explicitly in Hibernate, and list them in a table with a description of what they do.

scottmarlow: I'm fine with pushing this new CDI Persistence way.  Here are a few variations on the text:

> In the `persistence.xml`, each persistence unit can define their own scope and own qualifiers. It's important to note if you choose to use CDI injection for Jakarta Persistence injection , you should create CDI qualifiers if you have more than one persistence unit defined in your `persistence.xml` file. Without qualifiers specified you may see deployment failures due to ambiguous beans.

A variation that is closer to the original:

> In the `persistence.xml`, each persistence unit can define their own scope and own qualifiers. It's important to note if you choose to use CDI for your Jakarta Persistence application, you should create CDI qualifiers if you have more than one persistence unit defined in your `persistence.xml` file. Without this you may see deployment failures due to ambiguous beans.

Another variation:

> In the `persistence.xml`, each persistence unit can define their own scope and own qualifiers. If you use CDI injection for Jakarta Persistence injection , you should create CDI qualifier classes for each persistence unit. Without qualifiers specified you may see deployment failures due to ambiguous beans.

The three variations basically make room for the fact that some applications that have multiple persistence units may not get deployment failures due to ambiguous beans if they have wildfly.jpa.default-unit set to true on one of the persistence units.

> What might be useful if it's not already done somewhere is to document all the hints available in WildFly, maybe not all the ones that are explicitly in Hibernate, and list them in a table with a description of what they do.

https://docs.wildfly.org/37/Developer_Guide.html#persistence-unit-properties has:

> wildfly.jpa.default-unit
> set to true to choose the default persistence unit in an application. This is useful if you inject a persistence context without specifying the unitName (@PersistenceContext EntityManager em) but have multiple persistence units specified in your persistence.xml.

scottmarlow: More to the point, the incorrect parts of the existing text are:
1.  `you must create CDI qualifiers if you have more than one persistence unit defined in your persistence.xml file.` is incorrect if one of the persistence units already have `wildfly.jpa.default-unit`.  While we don't need to mention the hint in this part of the documentation, I think we should reduce the `must` to `should` because the hint could be in use.
2. `Without this you will likely see deployment failures due to ambiguous beans.` is incorrect for the same reason as 1.  I think we should reduce the  `likely` to `might` because the hint could be in use.

jamezp: Setting the `wildfly.jpa.default-unit` doesn't seem to work. I still get the following:
```
10:41:20,503 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-8) MSC000001: Failed to start service jboss.deployment.unit."contacts.war".WeldStartService: org.jboss.msc.service.StartException in service jboss.deployment.unit."contacts.war".WeldStartService: Failed to start service
	at org.jboss.msc@1.5.6.Final//org.jboss.msc.service.ServiceControllerImpl$StartTask.execute(ServiceControllerImpl.java:1545)
	at org.jboss.msc@1.5.6.Final//org.jboss.msc.service.ServiceControllerImpl$ControllerTask.run(ServiceControllerImpl.java:1374)
	at org.jboss.threads@3.9.1//org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
	at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
	at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2630)
	at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor.runThreadBody(EnhancedQueueExecutor.java:1622)
	at org.jboss.threads@3.9.1//org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1589)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001409: Ambiguous dependencies for type EntityManager with qualifiers @Default
  at injection point [UnbackedAnnotatedField] @Inject private dev.resteasy.examples.data.ContactRegistry.em
  at dev.resteasy.examples.data.ContactRegistry.em(ContactRegistry.java:0)
  Possible dependencies: 
  - Configurator Bean [interface jakarta.persistence.EntityManager, types: Object, AutoCloseable, EntityManager, qualifiers: @Any @Default],
  - Configurator Bean [interface jakarta.persistence.EntityManager, types: Object, AutoCloseable, EntityManager, qualifiers: @Any @Default]

	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.Validator.validateInjectionPointForDeploymentProblems(Validator.java:405)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.Validator.validateInjectionPoint(Validator.java:312)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.Validator.validateGeneralBean(Validator.java:159)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.Validator.validateRIBean(Validator.java:181)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.Validator.validateBean(Validator.java:552)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.ConcurrentValidator$1.doWork(ConcurrentValidator.java:65)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.bootstrap.ConcurrentValidator$1.doWork(ConcurrentValidator.java:63)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.executor.IterativeWorkerTaskFactory$1.call(IterativeWorkerTaskFactory.java:62)
	at org.jboss.weld.core@6.0.3.Final//org.jboss.weld.executor.IterativeWorkerTaskFactory$1.call(IterativeWorkerTaskFactory.java:55)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
	at org.jboss.threads@3.9.1//org.jboss.threads.JBossThread.run(JBossThread.java:501
```

This means you've got to use CDI qualifiers.

scottmarlow: > Setting the `wildfly.jpa.default-unit` doesn't seem to work. I still get the following:
> ...
> This means you've got to use CDI qualifiers.

Lol, that is awesome to understand now.  Thank you finding that out so I retract my claim now.  Thanks for actually testing this!  I think that is it.  I'll approve and merge soon.

scottmarlow (approved):

docs/src/main/asciidoc/WildFly_and_WildFly_Preview.adoc
scottmarlow: Just in terms of the why, a paste from existing code:
```
            List<String> qualifiers = Collections.emptyList();
            if (persistenceUnitMetadata.getQualifierAnnotationNames().size() > 0) {
                qualifiers = persistenceUnitMetadata.getQualifierAnnotationNames();
            } else {
                // mark the only default persistence unit as default.
                // With multiple persistence units, mark the one with hint "wildfly.jpa.default-unit" set to true.
                if (onePersistenceUnit || Configuration.isDefaultPersistenceUnit(persistenceUnitMetadata)) {
                    qualifiers = defaultQualifier;
                }
            }
```
Clearly the above pasted code is only adding the `jakarta.enterprise.inject.Default` qualifier if there are no qualifiers specified in the persistence unit and the hint is specified as well.

scottmarlow: ^ isn't the why actually.

jamezp: I don't think that's the issue. The issue is that there are two persistence units and at least one needs a qualifier. If with the `wildfly.jpa.default-unit`, the secondary one needs some kind of a qualifier otherwise CDI can't determine which producer to use.

By default, `@Default` and `@Any` are the qualifiers, so with two PU's and neither of them defining a CDI qualifier, they both become `@Default`.

scottmarlow: Thanks @jamezp that sounds right based on what you said earlier and what I read in the CDI spec that you pointed me to.  I don't currently see a need to check for the `wildfly.jpa.default-unit` now.

scottmarlow: Thanks @jamezp for the fix + changes!

PR #19245, opened 2025-10-02
[WFLY-20989] Don't use System.getProperties() in ts util code that ma…

bstansberry:
…y run in the deployment stack; add new PropertyPermission to PooledEJBLifecycleTestCase

https://issues.redhat.com/browse/WFLY-20989


jamezp (approved):

bstansberry:
Thanks @jamezp

PR #19247, opened 2025-10-02
[WFLY-20280] Disable the ResteasyDocumentSecureProcessingFeatureTestC…

jamezp:
…ase.checkFalse() test method on Java 24+.

https://issues.redhat.com/browse/WFLY-20280


jamezp:
FWIW I've added a PR in RESTEasy for some updated documentation as well https://github.com/resteasy/resteasy/pull/4698

bstansberry (approved):
Perhaps this test could check some other aspect of secure processing other than the limit.

I assume RE doesn't expose its internal settings to the app; otherwise the test could just be a servlet that checks those settings. The fundamental tests of this belong in RestEasy; a WF test could just confirm the mgmt API config ends up in the RE config.

jamezp:
Yes. I need to deep dive into the documentation and see what else the FSP setting enables. It seems like in older JVM's, it enabled those limits and in 24+, when disabled, it doesn't disable the limits. I tried to think of a different way to test. 

I'll definitely read the documentation in more detail to see if there are other options though.

bstansberry:
@jamezp I poked and AFAICT the management model attribute ends up surfacing as a ServletContext initParam. So any testing that that attribute or any others like it ends up being correctly passed into the runtime services can be done a simple Servlet using Jakarta Servlet APIs.

bstansberry:
Thanks @jamezp 

I'm happy to continue to discuss this general topic here (but @ ping me) but the PR itself is fine, so I merged this.

PR #19249, opened 2025-10-03
[WFLY-20605] Clarify that MP RM should be used from .war files

kabir:
https://issues.redhat.com/browse/WFLY-20605


bstansberry (approved):

bstansberry:
Thanks @kabir

PR #19255, opened 2025-10-07
[WFLY-21005] NullPointerException if an RA module is missing META-INF/ra.xml

lvydra:
https://issues.redhat.com/browse/WFLY-21005

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
+++ b/connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
@@ -977,4 +977,7 @@ public interface ConnectorLogger extends BasicLogger {
 
     @Message(id = 135, value = "The jdbc driver: %s is not installed")
     OperationFailedException jdbcDriverNotInstalled(String jdbcDriver);
+
+    @Message(id = 136, value = "The %s module does not export a META-INF/ra.xml file.")
+    String raXmlNotFound(String moduleId);
 }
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/jboss/as/connector/logging/ConnectorLogger.java
```diff
@@ -977,4 +977,7 @@ public interface ConnectorLogger extends BasicLogger {
 
     @Message(id = 135, value = "The jdbc driver: %s is not installed")
     OperationFailedException jdbcDriverNotInstalled(String jdbcDriver);
+
+    @Message(id = 136, value = "META-INF/ra.xml file hasn't been found.")
```
bstansberry: This should provide more information.

"The %s module does not export a META-INF/ra.xml file"

Ideally the current L452 is moved out of the try/catch block where it's declared now and then its 'moduleId' var is used in the message here and in the two OFEs in the catch blocks, instead of 'moduleName'.  Basically 'moduleId' is the more correct way to refer to the module by name, so use it. L452 doesn't need to be in the try/catch as it will not throw either of the caught exception types.

bstansberry (changes requested):

lvydra:
Thanks for the review @bstansberry, PR should be updated.

bstansberry (approved):

tadamski (approved):

bstansberry:
Thanks @lvydra and @tadamski

PR #19257, opened 2025-10-07
[WFLY-20964] WildFly Application Server latest major addition

akil-krsna:
Fixes: [https://issues.redhat.com/browse/WFLY-20964](https://issues.redhat.com/browse/WFLY-20964)

This PR is an addition to [https://github.com/wildfly/wildfly.github.io/pull/97](https://github.com/wildfly/wildfly.github.io/pull/97), differentiating the docs' landing page from the Application Server page.

### DIFF
--- a/docs/src/main/asciidoc/index.adoc
+++ b/docs/src/main/asciidoc/index.adoc
@@ -1,6 +1,7 @@
 [[index]]
-= WildFly Documentation
+= WildFly {wildflyVersion} Documentation
 :ext-relative: {outfilesuffix}
+:wildflyVersion: 37
 
 ifdef::env-github[]
 :tip-caption: :bulb:


bstansberry (changes requested):
Thanks @akil-krsna. 

Please use 'wildflyVersion' as the attribute here. The value of that attribute is driven by the pom and is thus automagically changed with each release. You need to declare the attribute with a static value (e.g. 37) as you do on line 4, but that static value gets overridden when we build the docs.

For more on this, see https://github.com/wildfly/wildfly/tree/main/docs#configuring-the-pom

bstansberry:
Note I'm not doing the ok-to-test thing suggested by the wildfly-ci bot because that would waste energy by triggering a bunch of CI jobs that don't test the docs. I'd certainly do it if this was a PR that would benefit from that testing.

bstansberry (approved):

bstansberry:
Thanks @akil-krsna

PR #19259, opened 2025-10-08
WFLY-20366 JGroups protocols with defined module=".." should attempt to load with 'org.jgroups.protocols' prefix

rhusar:
Jira
https://issues.redhat.com/browse/WFLY-20366

### DIFF
--- a/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/AbstractChannelResourceDefinitionRegistrar.java
+++ b/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/AbstractChannelResourceDefinitionRegistrar.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -41,6 +42,7 @@ import org.jboss.as.controller.registry.PlaceholderResource;
 import org.jboss.as.controller.registry.Resource;
 import org.jboss.dmr.ModelNode;
 import org.jboss.modules.Module;
+import org.jboss.modules.ModuleClassLoader;
 import org.jboss.modules.ModuleLoadException;
 import org.jgroups.Global;
 import org.jgroups.JChannel;
@@ -97,7 +99,7 @@ public abstract class AbstractChannelResourceDefinitionRegistrar<C extends Chann
                 @Override
                 public void accept(ForkChannelFactoryConfiguration configuration) {
                     ProtocolStack stack = configuration.getChannel().getProtocolStack();
-                    FORK fork = (FORK) stack.findProtocol(FORK.class);
+                    FORK fork = stack.findProtocol(FORK.class);
                     fork.remove(name);
                 }
             };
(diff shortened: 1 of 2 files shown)


clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/AbstractChannelResourceDefinitionRegistrar.java
```diff
@@ -393,14 +394,40 @@ public Map<String, ModelNode> getChildTypeOverrideDescriptions(Locale locale) {
...
+        } catch (ClassNotFoundException e) {
+            if (!isDefaultModule) {
+                // In this case try again with the prefix even if it's not from the default module,
+                // but it's using a shorthand name, e.g. "raft.RAFT", akin to standalone jgroups class-loading.
+                className = Global.PREFIX + protocolName;
+
+                try {
+                    return tryLoadClass(classLoader, className);
+                } catch (ClassNotFoundException ex) {
+                    throw JGroupsLogger.ROOT_LOGGER.unableToLoadProtocolClass(className);
```
bstansberry: This will be a confusing message because at this point in the code className starts with the Global.PREFIX that the user didn't specify.

bstansberry (changes requested):

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/AbstractChannelResourceDefinitionRegistrar.java
pferraro: Good point.
Also, this retry is unnecessary if the protocol already starts with Global.PREFIX, but is not loadable with the associated module.
Perhaps something like this would be simpler (and less verbose):
```java
List<String> candidateClassNames = new ArrayList<>(2);
if (protocolName.startsWith(Global.PREFIX)) {
   // Protocol name is a class name
   candidateClassNames.add(protocolName);
} else {
   // If using non-default module, try loading protocol name as class name
   if (!isDefaultModule) {
      candidateClassNames.add(protocolName);
   }
   // Compose class name using standard prefix
   candidateClassNames.add(Global.PREFIX + protocolName);
}

Iterator<String> classNames = candidateClassNames.iterator();
while (classNames.hasNext()) {
    try {
        return tryLoadClass(classLoader, classNames.next());
    } catch (ClassNotFoundException e) {
        // Retry with next
    }
}
throw JGroupsLogger.ROOT_LOGGER.unableToLoadProtocolClass(protocolName);

pferraro (changes requested):

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/AbstractChannelResourceDefinitionRegistrar.java
rhusar: @bstansberry Good point.

@pferraro OK, as I said earlier, I also did not like the nesting. However, I was trying to avoid creating a new collection objects for every protocol load.

rhusar:
Synced now. 

~~Changes from the branch https://github.com/rhusar/wildfly/tree/WFLY-20366 are not propagating to the PR, convert to draft temporarily until GH refreshes it.~~

bstansberry (approved):

pferraro (approved):

bstansberry:
Thanks @pferraro and @rhusar

PR #19267, opened 2025-10-09
WFLY-20041 - Log a warning if more than one metrics system is enabled

jasondlee:
https://issues.redhat.com/browse/WFLY-20041

Use correct capability name for WF Metrics


bstansberry:
Thanks @jasondlee

PR #19277, opened 2025-10-13
[WFLY-21037] Update documentation copyright

jmesnil:
Add a docinfo-footer that adds the same copyright footer to all the generated doc based on Commonhaus template.

The copyright blurb is a HTML snippet stored in the "wildfly.copyright" property in the root pom.xml

Consistently use "WildFly authors" as the author of the documentation

This fixes https://issues.redhat.com/browse/WFLY-21037

### DIFF
--- a/docs/src/main/asciidoc/index.adoc
+++ b/docs/src/main/asciidoc/index.adoc
@@ -1,7 +1,7 @@
 [[index]]
 = WildFly {wildflyVersion} Documentation
+WildFly authors;
 :ext-relative: {outfilesuffix}
-:wildflyVersion: 37
 
 ifdef::env-github[]
 :tip-caption: :bulb:
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -57,12 +57,14 @@
                         <revdate>${maven.build.timestamp}</revdate>
                         <organization>${project.organization.name}</organization>
                         <!-- end TODO -->
+                        <nofooter>true</nofooter>
+                        <docinfo>shared</docinfo>
                         <!-- Attributes to use in the asciidoc source files. Please leave in alphabetical order -->
                         <appservername>${full.dist.product.release.name}</appservername>
                         <oracle-javadoc>https://docs.oracle.com/en/java/javase/17/docs/api</oracle-javadoc>
                         <resteasyversion>${version.org.jboss.resteasy}</resteasyversion>
                         <wildflyversion>${product.docs.server.version}</wildflyversion>
-
+                        <copyright-blurb>${wildfly.copyright}</copyright-blurb>
                         <!-- Default asciidoc setting attributes -->
                         <linkcss>false</linkcss>
                         <sectanchors/>
--- a/docs/src/main/asciidoc/Admin_Guide.adoc
+++ b/docs/src/main/asciidoc/Admin_Guide.adoc
@@ -1,6 +1,6 @@
 [[Admin_Guide]]
 = WildFly Admin Guide
-WildFly developer team;
+WildFly authors;
 :revnumber: {version}
 :revdate: {localdate}
 :toc: macro
@@ -25,8 +25,6 @@ endif::[]
 
 
 ifndef::ebook-format[:leveloffset: 1]
-
-(C) The WildFly Authors.
 
 ifdef::basebackend-html[toc::[]]
 
--- a/docs/src/main/asciidoc/Bootable_Guide.adoc
+++ b/docs/src/main/asciidoc/Bootable_Guide.adoc
@@ -1,6 +1,6 @@
 [[Bootable_Guide]]
 = Bootable JAR Guide
-WildFly team;
+WildFly authors;
 :revnumber: {version}
 :revdate: {localdate}
 :toc: macro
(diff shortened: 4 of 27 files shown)


jmesnil:
The generated documentation can be seen at https://jmesnil.github.io/wildfly.github.io/38/ 

All pages display the same footer:

Screenshot 2025-10-13 at 10 46 48

jmesnil:
I also fixed the asciidoctor errors/warnings reported when the doc is generated.
There is also https://github.com/wildfly/quickstart/pull/1088 to fix the included quickstart guides.

jamezp (approved):

docs/src/main/asciidoc/index.adoc
```diff
@@ -1,7 +1,7 @@
 [[index]]
 = WildFly {wildflyVersion} Documentation
+WildFly authors;
 :ext-relative: {outfilesuffix}
-:wildflyVersion: 37
```
rhusar, line 4: Nope, this breaks rendering on GitHub, see:

https://github.com/wildfly/wildfly/blob/a43b29dff748ec601e1605e602b90c1f033fbc83/docs/src/main/asciidoc/index.adoc

in main this looks like this

https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/index.adoc

rhusar (changes requested):

docs/src/main/asciidoc/index.adoc
jamezp: Do we care though? This isn't meant to be displayed properly on GitHub IMO. The case is different, but the value is set here https://github.com/wildfly/wildfly/blob/2556c39331f8bab2eb0185aca787a72405402656/docs/pom.xml#L64

bstansberry: I think this was likely an inadvertent removal as this line was very recently added, on my request, by a hacktoberfest contributor. 

I was going to send a separate PR to restore it, since we include this attribute with a static value in many other files where the attribute is used, and I didn't want have this one be an exception. But a `git grep wildflyVersion` in the docs dir shows many files that use the attribute without declaring it first. Many more than do declare it, and a lot of ones that declare it don't use it.

bstansberry (approved):

docs/src/main/asciidoc/index.adoc
rhusar: @jamezp I do care, as that makes sharing/editing/reviewing on GitHub easier. If we publish snapshots then perhaps less so, but for now..

rhusar: @bstansberry Also, many of the defined wildflyVersion are referring a very old version... :-/

jamezp: That is specifically why I had introduced the variable in the POM rather than defining it in every file :)

jamezp: > @jamezp I do care, as that makes sharing/editing/reviewing on GitHub easier. If we publish snapshots then perhaps less so, but for now..

I guess I don't see how i differs :) We shouldn't be pointing people to documentation on GitHub and I think we can assume when we see it that means "current version" :)

rhusar: @jamezp Agreed, we shouldn't have to, but if the documentation is not part of any release yet, it is not published anywhere outside of GH.. we really need snapshot documentation.

jamezp: @rhusar Which also means it can't be used without building a SNAPSHOT so :shrug: I guess I don't really see the value of the overhead vs simply defining it in one place to make it display nice on GitHub. It really doesn't even display that made and feels fairly intuitive to me it's referring to the current version. 

Even restoring this value is wrong as we're not on WildFly 37 anymore, it should be WildFly 38.

rhusar: @jamezp Not exactly, the situation I had in mind and experienced is that the feature exists for a long time, but the docs do not. So, for instance, user complains about lack of documentation, I author it, but can't link it until next WF release.

Nevertheless, you are right the value is so ridiculously out of date, that it's better just un-rendered than some nonsensical 14 releases old number :-D Case in point:

```
[rhusar@puglia docs]$ git grep wildflyVersion\:
src/main/asciidoc/Admin_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Client_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Developer_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Extending_WildFly.adoc::wildflyVersion: 14
src/main/asciidoc/Galleon_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Getting_Started_Developing_Applications_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Hacking_On_WildFly.adoc::wildflyVersion: 26
src/main/asciidoc/Installation_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Migration_Guide.adoc::wildflyVersion: 14
src/main/asciidoc/Testsuite.adoc::wildflyVersion: 14
src/main/asciidoc/WildFly_Elytron_Security.adoc::wildflyVersion: 14
src/main/asciidoc/_high-availability/index.adoc::wildflyVersion: 27
src/main/asciidoc/index.adoc::wildflyVersion: 37
```

rhusar (approved):
LGTM (the `wildflyVersion` discussion is for another day and another PR)

bstansberry:
Thanks @jmesnil, @jamezp and @rhusar

PR #19278, opened 2025-10-13
[WFLY-20930] ElytronRemoteOutboundConnectionTestCase: change SSL port…

tadamski:
… between the tests

https://issues.redhat.com/browse/WFLY-20930

I have spend some time looking in this test and my conclusions are:

* There is no problem with CLI commands - everything seems to be created and destroyed correctly. Furthermore, the exception thrown is not CLI related one (in fact, it is thrown from native Linux code).
* There is no problem in socket closing - there is a close operation performed by xnio just as it should be.
* There error is intermittent but reproducible by long untilfail test loop.

Digging a bit further:
*  xnio correctly sets related socket options.
* ServerSocket is not closed immediately when the stream, to which xnio has a reference, is closed.
* The failures are related only to SSL tests, so presumably SSL configured socket need some more time to be closed.

Taking all this into consideration I'm getting to the conclusion that this problem is caused by underlying Linux socket behavior and keeping in mind how intermittent this problem is plus the fact that reopening the socket on the same port for a number of times is not a standard socket use case I wouldn't even consider it a bug.

Because of the above, I'm suggesting the workaround hack of just changing SSL port between tests.

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
@@ -166,7 +166,6 @@ public class ElytronRemoteOutboundConnectionTestCase {
 
     private static final int BARE_REMOTING_PORT = 54447;
     private static final String BARE_REMOTING_PROTOCOL = "remote";
-    private static final int SSL_REMOTING_PORT = 54448;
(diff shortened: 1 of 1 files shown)


rhusar (changes requested):
None of these ports are configured in waitForPorts in arquillian.xml. I haven't reviewed the test in detail, but if the container is started and restarted, it might help.

Nevertheless, this makes debugging unnecessarily convoluted having to look at the code execution which port is which. Just use static ports (so they can be found by scanning the code) or +1 and +2 to offset the `private static final int SSL_REMOTING_PORT`...

rhusar:
This has been resolved via https://github.com/wildfly/wildfly/pull/19411

Closing - if this is still valid, please reopen. Thanks @tadamski

PR #19290, opened 2025-10-18
[WFLY-20694] Add a smoke test of Jakarta Data that also covers ear an…

bstansberry:
…d -multiple-war scenarios

https://issues.redhat.com/browse/WFLY-20694

### DIFF
--- a/testsuite/integration/smoke/pom.xml
+++ b/testsuite/integration/smoke/pom.xml
@@ -147,6 +147,12 @@
         </dependency>
 
         <dependency>
+            <groupId>jakarta.data</groupId>
+            <artifactId>jakarta.data-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>jakarta.ejb</groupId>
             <artifactId>jakarta.ejb-api</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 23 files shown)


testsuite/integration/smoke/pom.xml
```diff
@@ -656,6 +724,11 @@
                                                     <name>standalone-microprofile.xml</name>
                                                 </config>
                                             </included-configs>
+                                            <included-packages>
+                                                <!-- Tests will add the jakarta data extension, so provision its packages too -->
+                                                <name>org.wildfly.extension.jakarta.data</name>
+                                                <name>jakarta.data.api</name>
+                                            </included-packages>
```
bstansberry, line 731: @jfdenise I had to do this to get these to provision. Setting 'stability-level' to 'preview' in 'galleon-options' didn't work.

My interpretation of it was specifying 'included-configs' has an effect similar to specifying layers; i.e. Galleon no longer provisions all packages. It provisions those that are needed to satisfy the specified configs. So to get other packages provisioned I have to explicitly say to include them.  Does that sound right to you?

jfdenise: @bstansberry , this is a bug in the WildFly Maven Plugin: https://issues.redhat.com/browse/WFMP-313 
We would need a new release of the plugin.
Otherwise, yes, that sounds right.

jamezp (approved):

bstansberry:
Thanks @jamezp

PR #19291, opened 2025-10-20
[WFLY-20983] Handle injection in websocket endpoints

baranowb:
Issue: https://issues.redhat.com/browse/WFLY-20983

### DIFF
--- a/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
+++ b/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
@@ -1,0 +1,11 @@
+package org.jboss.as.test.integration.web.websocket;
+
+import jakarta.enterprise.inject.Model;
+
+@Model //to make it a bean without beans xml https://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bean_defining_annotations
+public class SimpleBean {
+
+    public String sayHello() {
+        return "No I wont";
+    }
+}
--- a/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/WebSocketTestCase.java
+++ b/testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/WebSocketTestCase.java
@@ -80,13 +80,30 @@ public class WebSocketTestCase {
         assertWebSocket(webapp);
     }
 
+    @Test
+    @OperateOnDeployment(CLIENT_IN_DEPLOYMENT)
+    public void testResourceInjectionInDeployment(@ArquillianResource URL webapp) throws Exception {
+        assertWebSocket(webapp, "mes", AnnotatedEndpoint.MESSAGE_OK);
+    }
+
+    @Test
+    @OperateOnDeployment(CLIENT_IN_DEPLOYMENT)
+    public void testBeanInjectionInDeployment(@ArquillianResource URL webapp) throws Exception {
+        //test bean injection in case of no beans.xml
+        assertWebSocket(webapp, "bean", AnnotatedEndpoint.MESSAGE_OK);
+    }
+
     private void assertWebSocket(URL webapp) throws InterruptedException, IOException, DeploymentException, URISyntaxException {
+        assertWebSocket(webapp, "Stuart", "Hello Stuart");
+    }
+
+    private void assertWebSocket(URL webapp, final String param, final String result) throws InterruptedException, IOException, DeploymentException, URISyntaxException {
         AnnotatedClient endpoint = new AnnotatedClient();
         WebSocketContainer serverContainer = ContainerProvider.getWebSocketContainer();
         try (Session session = serverContainer.connectToServer(endpoint,
                 new URI("ws", "", TestSuiteEnvironment.getServerAddress(), TestSuiteEnvironment.getHttpPort(),
-                        webapp.getPath() + "websocket/Stuart", "", ""))) {
-            Assert.assertEquals("Hello Stuart", endpoint.getMessage());
+                        webapp.getPath() + "websocket/"+param, "", ""))) {
+            Assert.assertEquals(result, endpoint.getMessage());
         }
     }
 
@@ -98,7 +115,7 @@ public class WebSocketTestCase {
      */
     private static WebArchive createBasicDeployment(String name) {
         return ShrinkWrap.create(WebArchive.class, name + ".war")
-                .addClasses(AnnotatedEndpoint.class)
+                .addClasses(AnnotatedEndpoint.class, SimpleBean.class)
                 .addAsManifestResource(new StringAsset("io.undertow.websockets.jsr.UndertowContainerProvider"),
                         "services/jakarta.websocket.ContainerProvider");
     }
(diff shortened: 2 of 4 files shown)


testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
```diff
@@ -0,0 +1,11 @@
+package org.jboss.as.test.integration.web.websocket;
+
+import jakarta.enterprise.inject.Model;
+
+@Model //to make it a bean without beans xml https://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bean_defining_annotations
```
manovotn, line 5: It is more common to annotate the bean with just its scope - so you can add for instance `@Dependent`.

FTR `@Model` is a shortcut for adding `@RequestScoped` and `@Named` at the same time, none if which  you really need/use here.

weld/subsystem/src/main/java/org/jboss/as/weld/CdiAnnotations.java
```diff
@@ -190,7 +215,9 @@ public String getSimpleName() {
     public static final Set<AnnotationType> BEAN_DEFINING_ANNOTATIONS = ImmutableSet.of(
             new AnnotationType(DotName.createComponentized(Constants.JAVAX_INTERCEPTOR, "Interceptor"), false), asAnnotationType(DECORATOR, false),
             asAnnotationType(DEPENDENT), asAnnotationType(REQ_SCOPED), asAnnotationType(CONV_SCOPED), asAnnotationType(SESS_SCOPED),
-            asAnnotationType(APP_SCOPED));
+            asAnnotationType(APP_SCOPED),
+            //WFLY-20983
+            asAnnotationType(WS_CLIENT_ENDPOINT), asAnnotationType(WS_SERVER_ENDPOINT), asAnnotationType(CdiAnnotations.WS_ENDPOINT));
```
manovotn: `CdiAnnotations` class holds just CDI annotations.
I think it would be better to do this inside `BeanDefiningAnnotationProcessor` where we already do the same thing for some of the JSF annotations - take a look [here](https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/BeanDefiningAnnotationProcessor.java#L40-L61) and do the same.

I.e. a static `DotName` field and then an `addAnnotation` call should be all you need.

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/WebSocketTestCase.java
```diff
@@ -80,13 +80,30 @@ public void testClientInDeployment(@ArquillianResource URL webapp) throws Except
...
 
+    @Test
+    @OperateOnDeployment(CLIENT_IN_DEPLOYMENT)
+    public void testResourceInjectionInDeployment(@ArquillianResource URL webapp) throws Exception {
+        assertWebSocket(webapp, "mes", AnnotatedEndpoint.MESSAGE_OK);
+    }
+
+    @Test
+    @OperateOnDeployment(CLIENT_IN_DEPLOYMENT)
+    public void testBeanInjectionInDeployment(@ArquillianResource URL webapp) throws Exception {
```
manovotn, line 91: IIUIC, you can merge `testResourceInjectionInDeployment` into this test as well.
In both cases it is the CDI container that will handle the injection and the deployment is identical.

manovotn (changes requested):
One thing I want to clarify. It is expected that the endpoints will now be [*`@Dependent`* scoped beans](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#dependent_context)?

The CDI container will mark the classes as beans but lacking any other scope annotation, it assumes `@Dependent` scope. It may be perfectly OK, I am just double checking if you are aware of it :)

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
baranowb: I just picked first annotation I laid my eyes on.

weld/subsystem/src/main/java/org/jboss/as/weld/CdiAnnotations.java
baranowb: ok, seems reasonable.

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/WebSocketTestCase.java
baranowb: Yeah, initially it was, but this way its sort of self explanatory what is expected to work - IMHO for someone who isnt familiar with this, it will be easier to understand?

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
baranowb: As to @dependant - it did not break anything. Is there any downside of such categorization ?

manovotn: > As to [@dependant](https://github.com/dependant) - it did not break anything. Is there any downside of such categorization ?

Dependent bean instances are never shared between injection points (a new bean is created for every IP) and the lifecycle of the bean is bound to the bean it is injected into.
Which is unlike, say, request scoped which has a lifecycle bound to a request and any attempt to resolve that bean within a single request will return the same bean instance.

baranowb: hmm, yeah, seems about right, unless user specify scope, it should work like that I think ? WDYT ?

manovotn: I do not know WS specification so I tried looking through it a bit.
The relevant part seems to be [chapter 7](https://jakarta.ee/specifications/websocket/2.2/jakarta-websocket-spec-2.2#websocket-endpoints-and-dependency-injection) which does not say much apart from the ability to inject into endpoints.

There is also a mention of beans in [2.3](https://jakarta.ee/specifications/websocket/2.2/jakarta-websocket-spec-2.2#clientapi) which is a bit awkward as it indicates the beans should be non-contextual which makes no sense to me.

manovotn (approved):
LGTM from CDI integration perspective although I am only vaguely familiar with WS spec.

testsuite/integration/web/src/test/java/org/jboss/as/test/integration/web/websocket/SimpleBean.java
baranowb: Yeah, its vague at best, possibly was supposed to be solidified in TCK, but TCK tests dont  cover all scenarios( ie this one)

bstansberry:
@fl4via or @ropalka or someone quite familiar with the WS spec, please review.

darranl:
@fl4via / @ropalka I am double checking what can be included in WildFly 39 Beta, if you are able to provide a positive review I may be able to include this one.

darranl:
@fabiobrz adding you to the reviewers as well if you are able to take a look?

darranl:
Maybe ignore me @fabiobrz I am misready WS == Web Services, I think actually here WS == Web Socket.

fabiobrz:
> Maybe ignore me @fabiobrz I am misready WS == Web Services, I think actually here WS == Web Socket.

+1, it looks like this is purely Web Socket related.

ropalka (approved):
LGTM

PR #19294, opened 2025-10-21
WFLY-21061 - Ensure that Micrometer metrics are not visible between deployments

jasondlee:
https://issues.redhat.com/browse/WFLY-21061

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
@@ -12,6 +12,7 @@ import static org.wildfly.extension.micrometer.MicrometerSubsystemRegistrar.MICR
 
 import java.util.function.Supplier;
 
+import io.micrometer.core.instrument.MeterRegistry;
 import org.jboss.as.controller.PathAddress;
 import org.jboss.as.controller.capability.CapabilityServiceSupport;
 import org.jboss.as.server.deployment.AttachmentKey;
@@ -25,6 +26,7 @@ import org.jboss.as.server.deployment.DeploymentUnitProcessor;
 import org.jboss.as.weld.WeldCapability;
 import org.wildfly.extension.micrometer.api.MicrometerCdiExtension;
 import org.wildfly.extension.micrometer.metrics.MetricRegistration;
+import org.wildfly.extension.micrometer.registry.ApplicationRegistry;
 import org.wildfly.service.Installer.StartWhen;
 import org.wildfly.subsystem.service.ServiceDependency;
 import org.wildfly.subsystem.service.ServiceInstaller;
@@ -32,6 +34,7 @@ import org.wildfly.subsystem.service.ServiceInstaller;
 class MicrometerDeploymentProcessor implements DeploymentUnitProcessor {
     static final String WELD_CAPABILITY_NAME = "org.wildfly.weld";
     static final AttachmentKey<MicrometerService> CONFIG_ATTACHMENT_KEY = AttachmentKey.create(MicrometerService.class);
+    static final AttachmentKey<MeterRegistry> REGISTRY_ATTACHMENT_KEY = AttachmentKey.create(MeterRegistry.class);
 
     MicrometerDeploymentProcessor() {
     }
@@ -68,8 +71,8 @@ class MicrometerDeploymentProcessor implements DeploymentUnitProcessor {
         }
     }
 
-    private void registerCdiExtension(DeploymentPhaseContext deploymentPhaseContext, MicrometerService micrometerService) throws DeploymentUnitProcessingException {
-        DeploymentUnit deploymentUnit = deploymentPhaseContext.getDeploymentUnit();
+    private void registerCdiExtension(DeploymentPhaseContext phaseContext, MicrometerService micrometerService) throws DeploymentUnitProcessingException {
+        DeploymentUnit deploymentUnit = phaseContext.getDeploymentUnit();
         try {
             CapabilityServiceSupport support = deploymentUnit.getAttachment(Attachments.CAPABILITY_SERVICE_SUPPORT);
 
@@ -77,8 +80,18 @@ class MicrometerDeploymentProcessor implements DeploymentUnitProcessor {
             if (!weldCapability.isPartOfWeldDeployment(deploymentUnit)) {
                 MICROMETER_LOGGER.noCdiDeployment();
             } else {
-                weldCapability.registerExtensionInstance(new MicrometerCdiExtension(micrometerService.getMicrometerRegistry()),
-                    deploymentUnit);
+                MeterRegistry applicationRegistry = null;
+
+                if (deploymentUnit.getParent() != null) {
+                    applicationRegistry = deploymentUnit.getParent().getAttachment(REGISTRY_ATTACHMENT_KEY);
+                }
+
+                if (applicationRegistry == null) {
+                    applicationRegistry = new ApplicationRegistry(deploymentUnit.getName(),
+                            micrometerService.getMicrometerRegistry());
+                    deploymentUnit.putAttachment(REGISTRY_ATTACHMENT_KEY, applicationRegistry);
+                }
+                weldCapability.registerExtensionInstance(new MicrometerCdiExtension(applicationRegistry), deploymentUnit);
             }
         } catch (CapabilityServiceSupport.NoSuchCapabilityException e) {
             //We should not be here since the subsystem depends on weld capability. Just in case ...
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
@@ -1,0 +1,210 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.micrometer.registry;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToLongFunction;
+import java.util.stream.StreamSupport;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.TimeGauge;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.search.RequiredSearch;
+import io.micrometer.core.instrument.search.Search;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.wildfly.extension.micrometer.MicrometerExtensionLogger;
+
+public class ApplicationRegistry extends SimpleMeterRegistry {
+    static final String TAG_WF_DEPLOYMENT = "WF_DEPLOYMENT";
+    private final String deploymentName;
+    private final MeterRegistry parentRegistry;
+    private final More more = new ApplicationMore();
+
+    public ApplicationRegistry(String deploymentName, MeterRegistry parentRegistry) {
+        this.deploymentName = deploymentName;
+        this.parentRegistry = parentRegistry;
+    }
+
+    @Override
+    public List<Meter> getMeters() {
+        return parentRegistry.getMeters().stream().filter(m -> !containsWrongDeploymentTag(m.getId()))
+                .toList();
+    }
+
+    @Override
+    public void forEachMeter(Consumer<? super Meter> consumer) {
+        parentRegistry.forEachMeter(m -> {
+            if (!containsWrongDeploymentTag(m.getId())) {
+                consumer.accept(m);
+            }
+        });
+    }
+
+    @Override
+    public Search find(String name) {
+        return parentRegistry.find(name).tag(TAG_WF_DEPLOYMENT, deploymentName);
+    }
+
+    @Override
+    public RequiredSearch get(String name) {
+        return parentRegistry.get(name).tag(TAG_WF_DEPLOYMENT, deploymentName);
+    }
+
+    @Override
+    public Meter remove(Meter meter) {
+        if (containsWrongDeploymentTag(meter.getId())) {
+            return null;
+        }
+        return parentRegistry.remove(meter.getId());
+    }
+
+    @Override
+    public Meter removeByPreFilterId(Meter.Id id) {
+        if (containsWrongDeploymentTag(id)) {
+            return null;
+        }
+        return parentRegistry.removeByPreFilterId(id);
+    }
+
+    @Override
+    public Meter remove(Meter.Id id) {
+        if (containsWrongDeploymentTag(id)) {
+            return null;
+        }
+        return parentRegistry.remove(id);
+    }
+
+    @Override
+    public Counter counter(String name, Iterable<Tag> tags) {
+        var counter = findByNameAndTags(name, tags).counter();
+
+        if (counter == null || containsWrongDeploymentTag(counter.getId()) ) {
+
+            counter = parentRegistry.counter(name, addMissingTag(tags));
+        }
+        return counter;
+    }
+
+    @Override
+    public <T> T gauge(String name, Iterable<Tag> tags, T stateObject, ToDoubleFunction<T> valueFunction) {
+        var gauge = findByNameAndTags(name, tags).gauge();
+
+        if (gauge == null || containsWrongDeploymentTag(gauge.getId()) ) {
+            parentRegistry.gauge(name, addMissingTag(tags), stateObject, valueFunction);
+        }
+
+        return stateObject;
+    }
+
+    @Override
+    public Timer timer(String name, Iterable<Tag> tags) {
+        var timer = findByNameAndTags(name, tags).timer();
+        if (timer == null || containsWrongDeploymentTag(timer.getId()) ) {
+            timer = parentRegistry.timer(name, addMissingTag(tags));
+        }
+        return timer;
+    }
+
+    @Override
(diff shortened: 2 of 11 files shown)


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -25,13 +19,22 @@
...
 import org.wildfly.subsystem.service.ServiceDependency;
 import org.wildfly.subsystem.service.ServiceInstaller;
 
+import java.util.function.Supplier;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.DEPLOYMENT;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBDEPLOYMENT;
+import static org.wildfly.extension.micrometer.MicrometerExtensionLogger.MICROMETER_LOGGER;
+import static org.wildfly.extension.micrometer.MicrometerSubsystemRegistrar.MICROMETER_SERVICE_SERVICE_NAME;
+
```
bstansberry: The old order was correct. :)

jasondlee: Bah. Something happened to my IDE config. I'll fix that and push once the CI run finishes. I'd like to make sure CI is as happy as local. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -0,0 +1,126 @@
...
+        parentRegistry.forEachMeter(m -> {
+            if (!containsWrongDeploymentTag(m.getId())) {
+                consumer.accept(m);
+            }
+        });
+    }
+
+    @Override
+    public Search find(String name) {
+        // ???
```
bstansberry: I assume this is cruft.

jasondlee: A note to self, yes. I'll clean it up before this is ready for merge (e.g., a Jira's still needed)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -0,0 +1,126 @@
...
+        return parentRegistry.find(name).tag(TAG_WF_DEPLOYMENT, deploymentName);
```
bstansberry, line 59: Note to self and others....

AIUI

1) The tag added here cannot be removed by the caller.
2) The various Search methods that return meters from the Search all use a logical AND. So adding another tag with the same name but a different value doesn't 'override' what is done here.

So, it looks like what's done here prevents these Search objects being used as a way to extract meters that are would not be accessible via the other MeterRegistry methods.

I haven't looked much at ApplicationRegistryTest but ^^^ would be good things to cover in that kind of unit test.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -0,0 +1,126 @@
+package org.wildfly.extension.micrometer.registry;
```
bstansberry, line 6: Copyright/license header.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
bstansberry, line 1: It looks like this test class can pretty easily cover the entire ApplicationRegistry API, so it should.

jasondlee: I have expanded it to cover it all, I think. There are some public methods on the base class (i.e., MeterRegistry) that call other public methods, so I've only overridden and tested those more... terminal methods...

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
jasondlee: Bunch of tests added. It's more or less a duplication of the integration test, so we might want to revisit that test. Maybe extensive unit test, with basic sanity integration? Seems odd to test the same thing in both places, but it is a nice security blanket, no doubt...

bstansberry: Thanks. I hadn't looked at the integration test. Since it's already written, duplication is fine. :) When a class is readily unit testable, covering as much as possible there and focusing on the integration tests on proving what can't be proven via the unit test is good. At least that's how I think. I'm sure others think differently. There's no clear right answer.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java
```diff
@@ -68,17 +71,27 @@ private PathAddress createDeploymentAddressPrefix(DeploymentUnit deploymentUnit)
...
             final WeldCapability weldCapability = support.getCapabilityRuntimeAPI(WELD_CAPABILITY_NAME, WeldCapability.class);
             if (!weldCapability.isPartOfWeldDeployment(deploymentUnit)) {
                 MICROMETER_LOGGER.noCdiDeployment();
             } else {
-                weldCapability.registerExtensionInstance(new MicrometerCdiExtension(micrometerService.getMicrometerRegistry()),
-                    deploymentUnit);
+                MeterRegistry applicationRegistry = null;
+
+                if (deploymentUnit.getParent() != null) {
+                    applicationRegistry = deploymentUnit.getParent().getAttachment(REGISTRY_ATTACHMENT_KEY);
```
bstansberry, line 86: Note: this is safe because of https://github.com/wildfly/wildfly-core/blob/main/server/src/main/java/org/jboss/as/server/deployment/DeploymentUnitPhaseService.java#L207-L210

That's what ensures that for any given phase the DUPs in the phase are executed for the parent before the DUPs for subdeployments run. So the code below that attaches the registry to the parent's DU will have run.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private Iterable<Tag> addMissingTag(Iterable<Tag> tags) {
+        var hasWfTag =  StreamSupport.stream(tags.spliterator(), false)
+                .anyMatch(t -> t.getKey().equals(TAG_WF_DEPLOYMENT));
+        return hasWfTag ? tags : Tags.concat(tags, TAG_WF_DEPLOYMENT, deploymentName);
+    }
+
+    private void protectSystemMetrics(Meter.Id id) {
+        if (id.getTag(TAG_WF_DEPLOYMENT) == null) {
+            throw new IllegalStateException("Cannot remove system meters.");
```
bstansberry: This needs i18n.

bstansberry: However, I'm not sure throwing an exception is right. The javadoc for remove says it will return "null if no meter matched the provided id" with a reasonably implied meaning of "null if no meter *in this registry* matched the provided id". What you've done in this PR is created a distinct registry, and no meters in that distinct registry match the provided id. (The fact your distinct registry has access to stuff from a different registry is an impl detail.

I can image code that calls remove and then implements null checks. That could work fine with this registry if it returned null but will fail now.

bstansberry: Also, Meter.Id has a public constructor, so in theory an Id object could be passed in that was created by the deployment based on its author's understanding of what the id of its meters will be, not knowing that we assume TAG_WF_DEPLOYMENT will be present. This would fail but probably shouldn't.

Doing the containsWrongDeploymentTag check but then calling Id.withTag if there is no tag would handle that case.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    @Test
+    public void verifySystemMeters() {
+        Assert.assertNotNull(systemRegistry.find(METER_SYSTEM).meter());
+        Assert.assertNotNull(systemRegistry.find(METER_JVM_MEMORY).meter());
+    }
+
+    @Test
+    public void verifySystemMeterVisibility() {
```
bstansberry: NBD, but this is kind of a confusing method name, particularly since the test method doesn't double check that the systemRegistry meter was not affected by adding METER_SYSTEM to the appRegistry.

It seems that verifyAppMeterIsolation() is covering the same thing but in a more understandable way.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    @Test
+    public void testGetForSystemMetrics() {
+        try {
+            appRegistry1.get(METER_SYSTEM).meter();
+            appRegistry1.get(METER_JVM_MEMORY).meter();
+            appRegistry2.get(METER_SYSTEM).meter();
+            appRegistry2.get(METER_JVM_MEMORY).meter();
+            Assert.fail("System metrics should not be visible.");
```
bstansberry: L128-130 could all not fail but L131 throws an exception, and this test would incorrectly pass.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/micrometer/isolation/AbstractIsolationResource.java
```diff
@@ -0,0 +1,154 @@
...
+        return Response.ok().build();
+    }
+
+    @GET
+    @Path("getMeters")
+    public Response getMeters() {
+        List<Meter> meters = meterRegistry.getMeters();
+        if (meters.stream().anyMatch(m -> {
+            var tag = m.getId().getTag(TAG_WF_DEPLOYMENT);
+            return tag != null && !tag.equals(deploymentName);
```
bstansberry, line 115: shouldn't this be 

return tag == null || !tag.equals(deploymentName);

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/micrometer/isolation/AbstractIsolationResource.java
```diff
@@ -0,0 +1,154 @@
...
+        return Response.ok().build();
+
+    }
+
+    @GET
+    @Path("forEachMeter")
+    public Response forEachMeter() {
+        meterRegistry.forEachMeter(m -> {
+            var tag = m.getId().getTag(TAG_WF_DEPLOYMENT);
+           if (tag != null && !tag.equals(deploymentName)) {
```
bstansberry, line 129: tag == null || !tag.equals(deploymentName)

well, simpler is 

!deploymentName.equals(tag)

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/micrometer/isolation/AbstractIsolationResource.java
```diff
@@ -0,0 +1,154 @@
...
+                throw new InternalServerErrorException();
+            }
+        });
+
+        return Response.ok().build();
+    }
+
+    private boolean containsWrongDeploymentTag(Meter.Id id) {
+        String tag = id.getTag(TAG_WF_DEPLOYMENT);
+        return tag != null && !deploymentName.equals(tag);
```
bstansberry: return !deploymentName.equals(tag);

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/micrometer/isolation/AbstractIsolationResource.java
```diff
@@ -0,0 +1,154 @@
+package org.wildfly.test.integration.observability.micrometer.isolation;
```
bstansberry, line 5: Missing header on this and the other new files.

bstansberry (changes requested):

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/micrometer/isolation/AbstractIsolationResource.java
jasondlee: I think you're right. The visibility rules have gone through a few iterations, but I think we've settled on this: no meter not created by this MeterRegistry instance should be visible at all, and that covers meters from other deployments AND "system" metrics, which are not required by spec or convention to be available, so if the tag is either missing or wrong, the ID is invalid.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
jasondlee: Good catch. Much better way to test that incoming.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
jasondlee: That test morphed a bit, but the name was never updated. I'll get it cleaned up, or even removed...

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
bstansberry: I unresolved this conversation just so my initial post remains visible in the future in the GH UI. I tacked on the last sentence about the test to that post but the real goal of that comment was as a design note for people who look into this for whatever reason, either now or many years from now.

observability/micrometer/src/test/java/org/wildfly/extension/micrometer/registry/ApplicationRegistryTest.java
bstansberry: Thanks; looks like you removed it.

bstansberry:
@jasondlee The "Cloud Tests Remote Run' failures seem related.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -0,0 +1,136 @@
...
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.search.RequiredSearch;
+import io.micrometer.core.instrument.search.Search;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+
+public class ApplicationRegistry extends SimpleMeterRegistry {
```
bstansberry, line 31: @jasondlee MeterRegistry has other methods that are not accounted for here.

1) clear().
2) config(). And then the API exposed by MeterRegistry.Config.

I think clear() could be implemented, e.g forEachMeter(ApplicationRegistry::remove)

MeterRegistry.Config is more complicated. Some things, like Config.meterFilter and the other methods that call it, I don't think can be supported. Others I don't know but I think there are other problems, like reachable things that expose a 'shutdown' method. It might be better to just log a WARN if getConfig() is called.

jasondlee:
> @jasondlee The "Cloud Tests Remote Run' failures seem related.

I've been looking that, but I need to grab some of @kabir's time to (re)learn how to run those locally.

jasondlee:
> @jasondlee The "Cloud Tests Remote Run' failures seem related.

@bstansberry @kabir This change did, in fact, break the tests, but for kind of a dumb reason. I have a PR open to fix it: https://github.com/wildfly-extras/wildfly-cloud-tests/pull/324

jasondlee:
@bstansberry Cloud Test failure fixed.

rhusar:
@jasondlee FYI needs a rebase.

jasondlee:
> @jasondlee FYI needs a rebase.

@rhusar Done.

bstansberry (approved):

bstansberry:
Thanks @jasondlee

NiasSt90:
After this change all of my application registered metrics are not exposed anymore.
For this looks like the new wrapper class `ApplicationRegistry` don't wrap all necessary methods.

for example:

````java
@Singleton @Startup
public class MicrometerRegistryProducer {

	@Inject
	private MeterRegistry registry;

	@PostConstruct
	void registerDefaultMetrics() {
               new ClassLoaderMetrics().bindTo(registry);
               new JvmInfoMetrics().bindTo(registry);
...
               assert !registry.getMeters().isEmpty() : "No meters successfully registered";
        }
}
````

File: observability/micrometer/src/main/java/org/wildfly/extension/micrometer/MicrometerDeploymentProcessor.java [hunk 14]
```diff
@@ -68,17 +71,27 @@ private PathAddress createDeploymentAddressPrefix(DeploymentUnit deploymentUnit)
...
+                }
+
+                if (applicationRegistry == null) {
+                    applicationRegistry = new ApplicationRegistry(deploymentUnit.getName(),
+                            micrometerService.getMicrometerRegistry());
```
NiasSt90, line 91: Hi, @jasondlee  
How this Registry could be scraped then? 
I think this pull request introduce a regression.

This registry will not be added to the `org.wildfly.extension.micrometer.registry.WildFlyCompositeRegistry` and therefore all added metrics to this ApplicationRegistry can't be scraped from the `org.wildfly.extension.micrometer.prometheus.WildFlyPrometheusRegistry` endpoint created in `org.wildfly.extension.micrometer.prometheus.PrometheusRegistryDefinitionRegistrar#configure`

Only from the call to `io.micrometer.core.instrument.composite.CompositeMeterRegistry#updateDescendants` the metrics from one registry are shared to each other and therefore visible from the WildFlyPrometheusRegistry and the endpoint. 

Because of this all my application metrics are not exported anymore since WF 39. 
There was a (unfortunately closed) ticket WFLY-21577 for this topic.

PR #19299, opened 2025-10-22
WFLY-21060 Make OpenTelemetry tests more resilient in resource-constrained CI environments

rhusar:
Changes

* reduce export interval to 2s as to not overwhelm the collector container
* essentially disable batching in the collector
* bump the collector container version to 0.138.0
* stop the container last to avoid spurious connection errors from the GrpcExporter

https://issues.redhat.com/browse/WFLY-21060

### DIFF
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/opentelemetry/FaultToleranceOpenTelemetryIntegrationTestCase.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/faulttolerance/opentelemetry/FaultToleranceOpenTelemetryIntegrationTestCase.java
@@ -49,8 +49,8 @@ public class FaultToleranceOpenTelemetryIntegrationTestCase {
     protected OpenTelemetryCollectorContainer otelCollector;
 
     private static final String MP_CONFIG = "otel.sdk.disabled=false\n" +
-            // Lower the interval from 60 seconds to 100 millis
-            "otel.metric.export.interval=100";
+            // Lower the interval from 60 seconds to 2 seconds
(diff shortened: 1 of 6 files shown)


jasondlee (approved):

jbliznak:
the same code with interval is in preview https://github.com/wildfly/wildfly/blob/main/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BaseOpenTelemetryTest.java#L34

rhusar:
> the same code with interval is in preview https://github.com/wildfly/wildfly/blob/main/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BaseOpenTelemetryTest.java#L34

Thanks @jbliznak ! But yuck, who does that... can it just not depend on a test-jar? Anyway, changed there too. Also, the fix is mainly for downstream where there is no preview.

bstansberry:
> > the same code with interval is in preview https://github.com/wildfly/wildfly/blob/main/testsuite/preview/expansion/src/test/java/org/wildfly/test/integration/observability/opentelemetry/BaseOpenTelemetryTest.java#L34
> 
> Thanks @jbliznak ! But yuck, who does that... can it just not depend on a test-jar? Anyway, changed there too. Also, the fix is mainly for downstream where there is no preview.

bstansberry:
I have no idea how I closed this a moment ago; Didn't mean to.

rhusar:
@jasondlee FYI all 14 checks now passed, I think that's the number we were awaiting :-)

rhusar:
@jasondlee Just a friendly reminder to also resolve the linked Jira(s) when merging PRs :-)

(Already fixed for this one.)

PR #19302, opened 2025-10-23
[WFLY-14559] Add the resteasy-original-webapplicationexception-behavior attribute to the jaxrs subsystem.

jamezp:
Issues:
- https://issues.redhat.com/browse/WFLY-14559
- https://issues.redhat.com/browse/WFLY-21058

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import jakarta.ws.rs.client.Client;
@@ -64,6 +65,8 @@ public class ResteasyAttributeTestCase {
             Map.entry(ModelType.LIST, String.class),
             Map.entry(ModelType.OBJECT, String.class)
     );
+
+    private static final Set<String> IGNORED_ATTRIBUTES = Set.of("tracing-type", "tracing-threshold", "resteasy-patchfilter-disabled", "resteasy-original-webapplicationexception-behavior");
     private static Client jaxrsClient;
 
     @ArquillianResource
@@ -230,7 +233,7 @@ public class ResteasyAttributeTestCase {
         for (AttributeDefinition attribute : JaxrsAttribute.ATTRIBUTES) {
             // Ignore these attributes as they are not set in the context as the attributes are not public. These are
             // also tested elsewhere
-            if ("tracing-type".equals(attribute.getName()) || "tracing-threshold".equals(attribute.getName()) || "resteasy-patchfilter-disabled".equals(attribute.getName())) {
+            if (IGNORED_ATTRIBUTES.contains(attribute.getName())) {
                 continue;
             }
             testAttribute(target, attribute);
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
@@ -1,0 +1,227 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.jaxrs.cfg;
+
+import java.net.URI;
+
+import jakarta.json.JsonObject;
+import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.ClientBuilder;
+import jakarta.ws.rs.core.MultivaluedMap;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.core.UriBuilder;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.InSequence;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.as.arquillian.api.ServerSetup;
+import org.jboss.as.arquillian.setup.ConfigureLoggingSetupTask;
+import org.jboss.as.arquillian.setup.SnapshotServerSetupTask;
+import org.jboss.as.test.integration.jaxrs.cfg.resources.ErrorResource;
+import org.jboss.as.test.integration.jaxrs.cfg.resources.TestApplication;
+import org.jboss.dmr.ModelNode;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.EmptyAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests the {@code resteasy-original-webapplicationexception-behavior} attribute works as expected.
+ * <p>
+ * Note that order used for the methods is for performance reasons of only having to reload the server once. If a reload
+ * becomes not required, the sequence is not important.
+ * </p>
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@RunAsClient
+@ServerSetup({ConfigureLoggingSetupTask.class, SnapshotServerSetupTask.class})
+public class ResteasyOriginalWebApplicationExceptionBehaviorTestCase extends AbstractResteasyAttributeTest {
+
+    @ArquillianResource
+    private URI uri;
+
+    public ResteasyOriginalWebApplicationExceptionBehaviorTestCase() {
+        super("resteasy-original-webapplicationexception-behavior");
+    }
+
+    @Deployment(testable = false)
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, ResteasyOriginalWebApplicationExceptionBehaviorTestCase.class.getSimpleName() + ".war")
+                .addClasses(
+                        TestApplication.class,
+                        ErrorResource.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    @InSequence(1)
+    public void checkDefault() {
+        try (Client client = ClientBuilder.newClient()) {
+            try (
+                    Response response = client.target(uriBuilder().path("auth"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(401, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertEquals(String.format("Expected 401 but got %d: %s - %s", response.getStatus(), response.getStatusInfo()
+                        .getReasonPhrase(), body), 401, response.getStatus());
+                Assert.assertTrue(String.format("\"%s\" is NOT expected in the response!", ErrorResource.TEXT_TO_BE_SANITIZED),
+                        body == null || !body.contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+                final MultivaluedMap<String, String> headers = response.getStringHeaders();
+                Assert.assertFalse(String.format("Expected the response headers %s to not include the user-id header", headers), headers.containsKey("user-id"));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(2)
+    public void checkDefaultRuntimeException() {
+        try (Client client = ClientBuilder.newClient()) {
+            try (
+                    Response response = client.target(uriBuilder().path("exception"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(500, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertTrue(String.format("\"%s\" is NOT expected in the response: \"%s\"", ErrorResource.TEXT_TO_BE_SANITIZED, body),
+                        body == null || !body.contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(3)
+    public void checkDefaultException() {
+        try (Client client = ClientBuilder.newClient()) {
+            try {
+                final String body = client.target(uriBuilder().path("auth"))
+                        .request()
+                        .get(String.class);
+                Assert.fail("Expected WebApplicationException to be thrown, but got a body of: " + body);
+            } catch (WebApplicationException e) {
+                final Response response = e.getResponse();
+                final String body = response.readEntity(String.class);
+                Assert.assertEquals(String.format("Expected 401 but got %d: %s - %s", response.getStatus(), response.getStatusInfo()
+                        .getReasonPhrase(), body), 401, response.getStatus());
+                Assert.assertTrue(String.format("\"%s\" is NOT expected in the response!", ErrorResource.TEXT_TO_BE_SANITIZED),
+                        body == null || !body.contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+                final MultivaluedMap<String, String> headers = response.getStringHeaders();
+                Assert.assertFalse(String.format("Expected the response headers %s to not include the user-id header", headers), headers.containsKey("user-id"));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(4)
+    public void checkDefaultExceptionRuntimeException() {
+        try (Client client = ClientBuilder.newClient()) {
+            try {
+                final String body = client.target(uriBuilder().path("exception"))
+                        .request()
+                        .get(String.class);
+                Assert.fail("Expected WebApplicationException to be thrown, but got a body of: " + body);
+            } catch (WebApplicationException e) {
+                final Response response = e.getResponse();
+                Assert.assertEquals(500, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertTrue(String.format("\"%s\" is NOT expected in the response: \"%s\"", ErrorResource.TEXT_TO_BE_SANITIZED, body),
+                        body == null || !body.contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(5)
+    public void checkOriginalBehavior() throws Exception {
+        writeAttribute(ModelNode.TRUE);
+        try (Client client = ClientBuilder.newClient()) {
+            try (
+                    Response response = client.target(uriBuilder().path("auth"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(401, response.getStatus());
+                final JsonObject body = response.readEntity(JsonObject.class);
+                Assert.assertNotNull(body);
+                Assert.assertTrue("Expected an error body in " + body, body.containsKey("error"));
+                Assert.assertTrue(String.format("\"%s\" is expected in the response!", ErrorResource.TEXT_TO_BE_SANITIZED), body.getString("error")
+                        .contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+
+                final MultivaluedMap<String, String> headers = response.getStringHeaders();
+                Assert.assertTrue(String.format("Expected the response headers %s to include the user-id header", headers), headers.containsKey("user-id"));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(6)
+    public void checkOriginalBehaviorRuntimeException() throws Exception {
+        writeAttribute(ModelNode.TRUE);
+        try (Client client = ClientBuilder.newClient()) {
+            try (
+                    Response response = client.target(uriBuilder().path("exception"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(500, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertTrue(String.format("\"%s\" is expected in the response!", ErrorResource.TEXT_TO_BE_SANITIZED),
+                        body != null && body.contains(ErrorResource.TEXT_TO_BE_SANITIZED));
+            }
+        }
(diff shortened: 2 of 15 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +45,39 @@ public Response html(final String html) {
...
+            );
+        }
+        throw new WebApplicationException(
+                Response.status(errorCode)
+                        .entity("Failed on purpose")
+                        .type("text/plain")
+                        .build()
+        );
+    }
+    @GET
```
tommaso-borgato, line 83: @jamezp I was wondering if it could be possible to add some javadoc: IIUC `callErrorEndpoint` invokes itself on a different endpoint twice:

1. resteasy.original.webapplicationexception.behavior=false --> the "Failed on purpose" text is supposed to be removed from the repsonse
2. resteasy.original.webapplicationexception.behavior=true --> the "Failed on purpose" text is supposed to be present in the respose

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +45,39 @@ public Response html(final String html) {
...
+    public Response getError(@PathParam("errorCode") final int errorCode) {
+        if (errorCode == 400) {
+            throw new BadRequestException(
+                    Response.status(Response.Status.BAD_REQUEST)
+                            .entity("Failed on purpose")
+                            .type("text/plain")
+                            .build()
+            );
+        }
+        throw new WebApplicationException(
```
tommaso-borgato, line 64: @jamezp why is this needed? isn't the endpoint always invoked with "/test/error/client/400" and raise the `BadRequestException` exception?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +45,39 @@ public Response html(final String html) {
...
+    @GET
+    @Path("/client/{errorCode}")
+    public String callErrorEndpoint(@PathParam("errorCode") final int errorCode) {
+        try {
+            return client.target(uriInfo.getBaseUri())
+                    .path("/error/" + errorCode)
+                    .request()
+                    .get(String.class);
+        } catch (WebApplicationException e) {
+            final Response response = e.getResponse();
```
tommaso-borgato: @jamezp I can't see how this block is reached in the scope of the test ("/test/error/client/400"): IIUC, the client always throws a `WebApplicationException`; the difference would be:

resteasy.original.webapplicationexception.behavior=false --> the "Failed on purpose" text is supposed to be removed from the repsonse
resteasy.original.webapplicationexception.behavior=true --> the "Failed on purpose" text is supposed to be present in the respose

It seems to me no "Sanitized" string is ever returned back.... most probably I am missing something ....

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: Yes. I can do that. I'm going to update the test in general to give a better idea of what we're trying to guard against as well.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: It's just cruft TBH. I'm going to change the test to throw a different error and only the single error.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: Yes, this is wrong. See my latest updates.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
```diff
@@ -64,6 +65,8 @@ public class ResteasyAttributeTestCase {
             Map.entry(ModelType.LIST, String.class),
             Map.entry(ModelType.OBJECT, String.class)
     );
+
+    private static final Set<String> IGNORED_ATTRIBUTES = Set.of("tracing-type", "tracing-threshold", "resteasy-patchfilter-disabled", "resteasy-original-webapplicationexception-behavior");
```
tommaso-borgato, line 69: @jamezp I am not sure I understood why the new attribute `resteasy-original-webapplicationexception-behavior` was excluded here ... is it because it now has its own dedicated `testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java` test?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +46,55 @@ public Response html(final String html) {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public JsonObject noAuth() {
+        try {
+            // This always throws an exception, the exception should always be a WebApplicationException
+            return client.target(uriInfo.getBaseUri())
+                    .path("/error/auth")
+                    .request(MediaType.APPLICATION_JSON_TYPE)
+                    // We use the get(Class<?>) here to ensure the error is thrown
+                    .get(JsonObject.class);
+        } catch (WebApplicationException e) {
```
tommaso-borgato: @jamezp why don't we let the exception to pop up, up until the test code itself and check the content there to be sanitized / not sanitized? this way we could avoid caching the exception on the server side deployment ....

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
```diff
@@ -0,0 +1,102 @@
...
+        return ShrinkWrap.create(WebArchive.class, ResteasyOriginalWebApplicationExceptionBehaviorTestCase.class.getSimpleName() + ".war")
+                .addClasses(
+                        TestApplication.class,
+                        ErrorResource.class)
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    @InSequence(1)
+    public void checkDefault() {
```
tommaso-borgato, line 64: @jamezp as discussed in https://github.com/wildfly/wildfly/pull/19302/files#r2465047424, why don't we change the test to actually expect an exception and just verify nothing from `public Response failedAuth()` is leaked?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
```diff
@@ -0,0 +1,102 @@
...
+                final JsonObject headers = json.getJsonObject("headers");
+                Assert.assertFalse(String.format("Expected the response %s to not include the user-id header", headers), headers.containsKey("user-id"));
+                Assert.assertFalse(String.format("Expected the response %s to not include an error body", json), json.containsKey("error"));
+            }
+        }
+    }
+
+    @Test
+    @InSequence(2)
+    public void checkOriginalBehavior() throws Exception {
```
tommaso-borgato, line 143: @jamezp as discussed in https://github.com/wildfly/wildfly/pull/19302/files#r2465047424, why don't we change the test to actually expect an exception and just verify content produced by `public Response failedAuth()` is actually leaked?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
jamezp: It's because I don't want to expose the attribute outside the subsystem. However, I guess I could instead modify the test to not require the use of an `org.jboss.as.controller.AttributeDefinition` and test the other attributes too.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: You're absolutely right. That is much cleaner, thank you!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyAttributeTestCase.java
jamezp: Actually, the way the test is written it would likely be a larger refactor of the test. This does seem reasonable, but I feel it should be a separate commit as it would be a bit invasive for this change IMO. I've filed https://issues.redhat.com/browse/WFLY-21078 for that.

tommaso-borgato: @jamezp Thank you for the explanation

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +46,35 @@ public Response html(final String html) {
...
+     *
+     * @return the error response
+     */
+    @GET
+    @Path("/auth")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response failedAuth() {
+        final String userId = "anonymous";
+        final JsonObjectBuilder builder = Json.createObjectBuilder();
+        throw new WebApplicationException(
```
tommaso-borgato, line 64: @jamezp isn't `WebApplicationException` it the kind of exception that the client is expected to use to wrap the original exception? I mean, the original exception can be any exception and still, the client has to wrap it in a `WebApplicationException` ... can we generalize the test and throw an exception other than `WebApplicationException`?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
```diff
@@ -33,4 +46,35 @@ public Response html(final String html) {
...
+                Response.status(Response.Status.UNAUTHORIZED)
+                        .entity(builder.add("error", String.format("User %s is not authorized", userId)).build())
```
tommaso-borgato: @jamezp "error" seems like a very common world ... can we check sensitization using some other string like e.g. "Text to be expected to be removed by sensitization" ?

tommaso-borgato (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: Without throwing a `WebApplicationException` here with a response, I can't add the header.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
jamezp: I'm not sure I follow this. Do you mean `@Test(expected=WebApplicationException.class)`? If so, I'd prefer not to do that as I'd like to inspect the response.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
jamezp: We can really check for anything I guess. I was just using JSON so I could look for a specific key in the returned body.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
tommaso-borgato: I was actually thinking about try & catch .... IIUC the final goal of the test would be to verify weather the exception caught in the catch block is sanitized/un-sanitized

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
tommaso-borgato: I suppose this is the same comment I made https://github.com/wildfly/wildfly/pull/19302#discussion_r2465111784 and hence, the discussion in that thread is valid for this one as well

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
tommaso-borgato: I see... but IIUC technically the header isn't necessary to test the `resteasy.original.webapplicationexception.behavior` attribute... why are we adding it here?

jamezp: Yes. The original RESTEasy issue was filed over the header so this is kind of a regression test to ensure a RESTEasy upgrade will not break that.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
jamezp: Okay. I've added new tests. I kept the old one and added a new one to catch the exception so we know we get the expected results either way.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
tommaso-borgato: Then my suggestion would be to have distinct endpoints and test methods: 

1. one couple **just** for checking the header (using `WebApplicationException`) 
2. the other couple **just** for checking response sanitizing (not using `WebApplicationException`)

To me, it would be easier to spot the specific issue in case anything breaks, if we keep them separated ... for sure it would be much easier for me when checking the testing pipeline

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
tommaso-borgato: Sure, my point is just that a string like "Text which is expected to be removed by sensitization" would make the test super readable

tommaso-borgato (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
tommaso-borgato: @jamezp see https://github.com/jamezp/wildfly/pull/9

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
tommaso-borgato: @jamezp see https://github.com/jamezp/wildfly/pull/9

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
tommaso-borgato: @jamezp see https://github.com/jamezp/wildfly/pull/9

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/resources/ErrorResource.java
tommaso-borgato: @jamezp see https://github.com/jamezp/wildfly/pull/9

tommaso-borgato (changes requested):
@jamezp I've tried and explained my observations with @jamezp see https://github.com/jamezp/wildfly/pull/9 : can you have a look?

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
```diff
@@ -0,0 +1,227 @@
...
+            try (
+                    Response response = client.target(uriBuilder().path("auth"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(401, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertEquals(String.format("Expected 401 but got %d: %s - %s", response.getStatus(), response.getStatusInfo()
+                        .getReasonPhrase(), body), 401, response.getStatus());
+                Assert.assertTrue(String.format("\"%s\" is expected in NOT the response!", ErrorResource.TEXT_TO_BE_SANITIZED),
```
tommaso-borgato: /is expected in NOT the response/is NOT expected in the response

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
```diff
@@ -0,0 +1,227 @@
...
+                final String body = client.target(uriBuilder().path("auth"))
+                        .request()
+                        .get(String.class);
+                Assert.fail("Expected WebApplicationException to be thrown, but got a body of: " + body);
+            } catch (WebApplicationException e) {
+                final Response response = e.getResponse();
+                final String body = response.readEntity(String.class);
+                Assert.assertEquals(String.format("Expected 401 but got %d: %s - %s", response.getStatus(), response.getStatusInfo()
+                        .getReasonPhrase(), body), 401, response.getStatus());
+                Assert.assertTrue(String.format("\"%s\" is expected in NOT the response!", ErrorResource.TEXT_TO_BE_SANITIZED),
```
tommaso-borgato: /is expected in NOT the response/is NOT expected in the response/

tommaso-borgato (changes requested):

tommaso-borgato:
@jamezp Thank you for adding tests with RuntimeException ... I just added a couple of minor comments

tommaso-borgato (approved):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
```diff
@@ -0,0 +1,227 @@
...
+        try (Client client = ClientBuilder.newClient()) {
+            try (
+                    Response response = client.target(uriBuilder().path("auth"))
+                            .request()
+                            .get()
+            ) {
+                Assert.assertEquals(401, response.getStatus());
+                final String body = response.readEntity(String.class);
+                Assert.assertEquals(String.format("Expected 401 but got %d: %s - %s", response.getStatus(), response.getStatusInfo()
+                        .getReasonPhrase(), body), 401, response.getStatus());
```
rhusar, line 74: Just a nitpick, but this is unnecessarily confusing, since JUnit is already going to print out actual/expected values, such as:

> java.lang.AssertionError: [Custom Message] expected:<[Expected Value]> but was:<[Actual Value]>

rhusar (approved):
In attempt to speed things up, I had a brief look and this looks good to me. 👍🏼 

Pointed out that we need new CI runs for this to be mergeable so James rebased this and test results will be current.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jaxrs/cfg/ResteasyOriginalWebApplicationExceptionBehaviorTestCase.java
jamezp: I think this was copied from @tommaso-borgato suggestions and I believe the idea is to get the response phrase as well.

rhusar: That part is fine, the problematic part is 'Expected 401 but got %d' as that will be confusingly duplicated.

PR #19303, opened 2025-10-23
[WFLY-21044]: Not recovered transactions after crash of server with MDB.

ehsavoie:
* Fixing how WildFly provides the recovery manager to Artemis
 * Removing the old MSC API usage to get the RecoveryService
 * Upgrading Artemis from 2.42.0 to 2.43.0

Jira: https://issues.redhat.com/browse/WFLY-21044
      https://issues.redhat.com/browse/WFLY-21063

### DIFF
--- a/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyRecoveryRegistry.java
+++ b/messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyRecoveryRegistry.java
@@ -5,13 +5,9 @@
 package org.wildfly.extension.messaging.activemq.jms;
 
 
+import java.util.function.Supplier;
 import org.jboss.activemq.artemis.wildfly.integration.recovery.WildFlyActiveMQRegistry;
-import org.jboss.as.controller.ServiceNameFactory;
-import org.jboss.msc.service.ServiceContainer;
-import org.jboss.msc.service.ServiceController;
-import org.jboss.msc.service.ServiceName;
 import org.jboss.tm.XAResourceRecoveryRegistry;
-import org.wildfly.extension.messaging.activemq.MessagingServices;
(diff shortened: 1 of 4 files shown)


mnovak:
I can confirm this PR is fixing WFLY-21044 and does not cause any regression in messaging.

messaging-activemq/subsystem/src/main/java/org/wildfly/extension/messaging/activemq/jms/WildFlyRecoveryRegistry.java
```diff
@@ -38,9 +34,6 @@ private static XAResourceRecoveryRegistry getXAResourceRecoveryRegistry() {
         // This parsing isn't 100% ideal as it's somewhat 'internal' knowledge of the relationship between
         // capability names and service names. But at this point that relationship really needs to become
         // a contract anyway
```
bstansberry: The comment lines should be removed since they are no longer relevant.

bstansberry:
Is the Artemis update required for the bug fix? If not it's better if they are decoupled.

bstansberry (changes requested):
After chatting with @ehsavoie I don't object to combining the upgrade with this; I don't think it will delay merging.

But, I suspect the MSC bits need to be ported to 38.x while a minor update or Artemis needs more thought there. So there I suggest separating things.


I've got a minor code review comment though. Sorry; I meant to post that yesterday.

jmesnil:
> But, I suspect the MSC bits need to be ported to 38.x while a minor update or Artemis needs more thought there. So there I suggest separating things.

#19304 is the backport to 38.x (without the artemis upgrade)

bstansberry (approved):
I added a commit to get rid of the comment.

I don't care about that nit in the 38.x PR; it's just cruft cleanup.

bstansberry:
Thanks @ehsavoie

PR #19305, opened 2025-10-23
WFLY-21065 - Micrometer subsystem does not register system metrics

jasondlee:
https://issues.redhat.com/browse/WFLY-21065

Register system metrics using Micrometer's API
Add a sampling of system metrics to the test
Update test utility to log last know state when checks fail


bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #19322, opened 2025-10-30
Fix for WFLY-21092, Shared GithubAction workflow can fail with multiple executions in the same workflow

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-21092

### DIFF
--- a/.github/workflows/shared-wildfly-build-and-test.yml
+++ b/.github/workflows/shared-wildfly-build-and-test.yml
@@ -88,6 +88,11 @@ on:
         required: false
         default: ""
         type: string
+      id:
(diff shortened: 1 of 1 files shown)


jamezp:
Instead of adding an additional ID parameter, couldn't we just rename the artifacts? Something like `wildfly-shared-surefire-reports-${{ matrix.os }}-${{ matrix.java }}`? I suppose given this has a default that is okay too, but we should use something to indicate where it came from IMO.

jfdenise:
@jamezp , renaming would be not enough, all executions of the shared workflow would still use the same name wildfly-shared-...

jamezp (approved):

PR #19091, opened 2025-07-24
[WFLY-19560] introduce EE Concurrency 3.1 implementation (preview only)

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19560

PR #19333, opened 2025-11-04
[EE11][WFLY-19560] introduce EE Concurrency 3.1 implementation (preview only)

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19560
Main PR: #19091


bstansberry:
This will have an unrelated failure in the WildFly Preview Galleon Linux - JDK 17 job -- see discussion at. [#Jakarta EE > EL API groupId @ 💬](https://wildfly.zulipchat.com/#narrow/channel/242838-Jakarta-EE/topic/EL.20API.20groupId/near/553720259)

I'll sort that tomorrow morning and re-run that job here.

bstansberry:
The https://ci.wildfly.org/buildConfiguration/WF_PullRequestEe11_GalleonLinuxJdk17/527358 failure is due to a problem in main that is now resolved. I don't want to do a complete retest just to get a new run of that job against a new merge commit. (It seems TC remembers the target branch head sha and the PR branch head sha and if you use the TC UI to 'Run' the job again it uses those and doesn't figure out the current ones??)

I ran the equivalent of that CI job locally and it passes.

PR #19334, opened 2025-11-04
Fix for WFLY-21110, WildFly Preview not well configured for doc generation

jfdenise:
Issue: https://issues.redhat.com/browse/WFLY-21110


jmesnil (approved):
LGTM

PR #19336, opened 2025-11-04
[WFLY-21098] Narayana updated to 7.3.2.Final

jmfinelli:
[WFLY-21098](https://issues.redhat.com/browse/WFLY-21098)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -518,7 +518,7 @@
         <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
         <version.org.jboss.metadata>17.0.0.Beta1</version.org.jboss.metadata>
         <version.org.jboss.mod_cluster>2.1.0.Final</version.org.jboss.mod_cluster>
-        <version.org.jboss.narayana>7.3.0.Final</version.org.jboss.narayana>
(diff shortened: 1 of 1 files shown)


bstansberry:
@jmfinelli This fails because the narayana-jts-idlj-7.3.2.Final jar no longer contains the JBoss Logging annotation processor-generated com.arjuna.common.logging.commonI18NLogger_$logger.class and commonI18NLogger.i18n.properties files. Something has changed in the Naryana build such that they are no longer being generated.

Are Narayana releases now being built with SE 25? If so note that annotation processing needs to be explictily turned on with SE 23+. For example:

https://github.com/wildfly/wildfly/blob/main/pom.xml#L1569-L1582

jmfinelli:
Thanks Brian! cc @marcosgopen

marcosgopen:
Thanks @bstansberry . We haven't built narayana with JDK25 but we committed some changes to move the logging-processor to the maven-plugin-compiler to be JDK25 compliant. Our test were successful so I need some time to understand why the generated logging implementation is not present in the jar of the release. I will soon revert the changes so that we can release soon a functioning release. That would be 7.3.3.Final. So I think we can close this PR. Thanks for your patience.

bstansberry:
I'm closing this based on the comment above. Thanks!

PR #19341, opened 2025-11-05
[WFLY-20695] Naming: ServiceBasedNamingStore - don't skip delegating to boundServices if initial lookup fails

elguardian:
based on @bstansberry
comments https://github.com/wildfly/wildfly/pull/18984#issuecomment-2960501152

issue https://issues.redhat.com/browse/WFLY-20695

probably it would be better to rename the issue is resolving as ResolveResult while the service is still starting.
It does exists but it did not produce any value yet.

### DIFF
--- a/naming/src/main/java/org/jboss/as/naming/ServiceBasedNamingStore.java
+++ b/naming/src/main/java/org/jboss/as/naming/ServiceBasedNamingStore.java
@@ -31,6 +31,7 @@ import javax.naming.spi.ResolveResult;
 import org.jboss.as.naming.deployment.ContextNames;
 import org.jboss.as.naming.logging.NamingLogger;
 import org.jboss.msc.service.ServiceController;
+import org.jboss.msc.service.ServiceController.State;
 import org.jboss.msc.service.ServiceName;
 import org.jboss.msc.service.ServiceRegistry;
 import org.wildfly.security.manager.WildFlySecurityManager;
@@ -63,12 +64,12 @@ public class ServiceBasedNamingStore implements NamingStore {
             return new NamingContext(EMPTY_NAME, this, null);
         }
         final ServiceName lookupName = buildServiceName(name);
-        Object obj = lookup(name.toString(), lookupName, dereference);
+        Object obj = lookup(name, lookupName, dereference);
         if (obj == null) {
             final ServiceName lower = boundServices.lower(lookupName);
             if (lower != null && lower.isParentOf(lookupName)) {
                 // Parent might be a reference or a link
-                obj = lookup(name.toString(), lower, dereference);
+                obj = lookup(name, lower, dereference);
                 //if the lower is a context that has been explicitly bound then
                 //we do not return a resolve result, as this will result in an
                 //infinite loop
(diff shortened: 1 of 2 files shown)


ropalka (approved):

bstansberry:
@emmartins Please review.

bstansberry (changes requested):
@elguardian +1 to some kind of change about the JIRA. Some future maintainer is going to be very confused about this.

How does this change relate to the original customer issue that led to the JIRA? It was not clear to me before how the existing JIRA wording etc related to that case.

If this change doesn't really relate to that case but is still a valid thing to do, perhaps it's better to create a new issue that describes the problem this solves and the expected impact and use new issue that for this commit / PR, leaveing the existing one as is. The existing one is tied to the support case so we shouldn't repurpose it to something that wouldn't fix the support case issue.

It's also possible the existing WFLY-20695 and JBEAP-30336 aren't truly 'fixes' for the original support case and are just side issues that came up when trying to work the support case. If so the JBEAP needs cleaning up because reading its details one would assume it is fixing a customer issue. If it's not a fix for a customer issue and it's not a regression perhaps there shouldn't be a JBEAP. Let it come into EAP via WildFly.

elguardian:
@bstansberry 
Hi, Your point is valid and I will try to clarify with Tomasz and try to get the reproducer to see if this fixes it or not. 

Given some background about this:

1. Jiras was on z-stream so I did not know about the history, it was sitting there so I tried to tackle it. I think this jira is coming from an engineer. Either if it lands in downstream or not does not depend on me, but as community first policy applies, I tried to do it in upstream first to at least trigger discussion to see what to do with the jira.
2. The initial fix proposed by the Tomaz was catching an illegalState exception wrapped by Naming not found. 
3. the naming not found only was reached by the ISE throw by controller.getValue.
4. That only happens due to a race condition as the service controller is still not up and the value not built yet
5. falling back in this scenario did not make sense as the controller was bind to the name and therefore is the one that should provided the value. 
6. The above is more likely to be the root cause of the orginal issue (the race condition) but I don't have the origin reproducer so I cannot confirm this end

I will try to clarify this ends with Tomaz in the jira.

bstansberry:
Great; thanks @elguardian. And sorry for not saying that in my first post! I was focused on paperwork but I definitely appreciate you for looking into this.

emmartins:
I am confused, what does this solve? Assuming that we really need this special lookup use case, I would prefer to do the state check only when the exception is thrown, to not add a performance penalty to all lookups. But again, I really need to understand the big picture and thus need some time to go back on links and JIRAs, any tips? :)

elguardian:
@tadamski Maybe you can add your reproducer to the jira in upstream. So we can move forward this.

elguardian:
@emmartins I got access finally to the logs. The problems was when the remote ejb are injected and there is a  IOexception (failed connection to the server) causing the failed step. There is more info the issue in 8.1.z

"Unable to discover destination for request for EJB StatelessEJBLocator for "<redacted>", view is interface <redacted>, affinity is None" caused by the remoting component.

This pr does not fix that issue.

emmartins:
@elguardian ok, thanks for the update

PR #19343, opened 2025-11-05
[WFLY-21052] Fix Non-Idempotent-Outcome flaky test in AddDataSourceOperationsUnitTestCase

bee0511:
Issue: https://issues.redhat.com/browse/WFLY-21052

Use timestamp-based deployment name to avoid conflicts between test classes:

```java
private static final String JDBC_DRIVER_NAME = "test-jdbc-" + System.currentTimeMillis() + ".jar";
```

### DIFF
--- a/testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/mgmt/datasource/AddDataSourceOperationsUnitTestCase.java
+++ b/testsuite/integration/smoke/src/test/java/org/jboss/as/test/smoke/mgmt/datasource/AddDataSourceOperationsUnitTestCase.java
@@ -17,6 +17,7 @@ import org.jboss.arquillian.junit5.ArquillianExtension;
 import org.jboss.as.connector.subsystems.datasources.DataSourcesExtension;
 import org.jboss.as.connector.subsystems.datasources.Namespace;
 import org.jboss.as.test.integration.management.jca.DsMgmtTestBase;
+import org.jboss.as.test.integration.management.util.ServerReload;
 import org.jboss.dmr.ModelNode;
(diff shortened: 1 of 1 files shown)


bstansberry:
@bee0511 Thanks. This fix seems to be hiding the problem though. When AddDataSourceOperationsUnitTestCase completes there should be no deployment=test-jdbc.jar resource in the configuration. If there is, that needs to be fixed.

TBH I don't see why that would happen though. The Deployment annotation has its 'managed' attribute left at the default 'true' so Arquillian should remove the deployment when the test class is complete.

Once we understand that, assuming there's some good reason why that's not happening, a better approach is to make the test class name part of the deployment name, not a timestamp. This makes things unique between tests. It also makes analyzing the server.log from the testsuite run easier as the test class name appears in the log.

bstansberry:
@bee0511 See the discussion at [#wildfly-developers > Specifying a single test &#42;method&#42; and Arquillian](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Specifying.20a.20single.20test.20*method*.20and.20Arquillian/with/553968019)

bstansberry:
@bee0511 To fix this you'll need to add this right after the test methods's call to 'remove(address)':

```
ServerReload.executeReloadAndWaitForCompletion(getModelControllerClient());
```

That and ```import org.jboss.as.test.integration.management.util.ServerReload;```

bee0511:
@bstansberry I've updated the PR, please take a look.

bstansberry (approved):

bstansberry:
Thanks @bee0511 !

PR #19350, opened 2025-11-06
WFLY-21117 Remove invisible character U+FEFF / ZWNBSP from the source…

rhusar:
… code

Resolves
https://issues.redhat.com/browse/WFLY-21117

Note to reviewers: This will render on GH as no change. It will render in ncurses-based environments as removed space (since it renders everything fixed-width including zero-width characters). Commit was generated by replace. There will be no manual changes.


bstansberry (approved):

rhusar:
n.b. this is also byte-order mark (BOM) – this PR is not removing that from binary files we have under SCM :-)

bstansberry:
Thanks @rhusar

PR #19354, opened 2025-11-07
WFLY-276 Add support for running the clustering integration tests with the JPDA debugger

rhusar:
* setup jpda properties per node since they must be unique
* remove passing the same jpda properties to each node resulting in port conflict in multi-node tests
* fixes in the testsuite documentation section and explain caveats of running multi-node test with the debugger
* organizes properties to be grouped logically
* remove noexistent properties used in arquillian-*.xml
* cleanup duplicated ${server.jvm.arg} property override
* cleanup redundant combine children

Resolves
https://issues.redhat.com/browse/WFLY-276


rhusar:
Here you go @rachmatowicz :)

pferraro (approved):

PR #19360, opened 2025-11-10
WFLY-21151 Temporarily use reflection while calling BM methods that were moved in CDI 5.0

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-21151

I spent some time delving into what is needed to be able to run [Weld 7 (CDI 5.0) with WFLY preview](https://issues.redhat.com/browse/WELD-2822) and figured I can do most in a script that just changes versions but there is the following change which I couldn't think of doing in any better way.

So the goal of this PR is to see if such ugly change would be temporarily acceptable in WFLY code @bstansberry @jamezp?

A little but of why:
* CDI 4.1 added CDI EL API jar (`ELAwareBeanManager`) and deprecated the two EL related methods on classes `BeanManager`
* CDI 5.0 removes these deprecations which causes the problem with method not found
* In so far as Weld testing goes, there seems to be only the two classes shown in this PR that need this ugly workaround
* Soon as WFLY switches to next EE version, this workaround can be replaced by proper code

If no is the answer, I suppose I can just wait until WFLY upgrades but since I already figured it, I thought I might as well ask :)

### DIFF
--- a/jsf/injection/src/main/java/org/jboss/as/jsf/injection/weld/WeldApplication.java
+++ b/jsf/injection/src/main/java/org/jboss/as/jsf/injection/weld/WeldApplication.java
@@ -14,6 +14,9 @@ import javax.naming.NamingException;
 
 import org.jboss.weld.module.web.el.WeldELContextListener;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
 /**
  * @author pmuir
  */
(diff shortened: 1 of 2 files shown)


bstansberry:
@manovotn My impression is this is needed with CDI 4.1 as well, as the code affects that scenario. Perhaps it's not needed there because calling the deprecated method works and has the same effect?

I'm asking because it might be a while before all use of EE 11 is gone from standard WildFly.

In any case this kind of thing doesn't bother me, although I'd ask for a WFLY to remove this, with an 'EE12' label on it (and on the issue you'd file to make this change.) That label should be a sufficient reminder to remove it. (I'd forget although you probably wouldn't!)

@jasondlee  @ropalka FYI

manovotn:
> @manovotn My impression is this is needed with CDI 4.1 as well, as the code affects that scenario. Perhaps it's not needed there because calling the deprecated method works and has the same effect?

It does effect 4.1 but it's not needed as invoking the deprecated methods work just as well.

> In any case this kind of thing doesn't bother me, although I'd ask for a WFLY to remove this, with an 'EE12' label on it (and on the issue you'd file to make this change.) That label should be a sufficient reminder to remove it. (I'd forget although you probably wouldn't!)

Oh, don't underestimate my forgetfulness :)
I'll create an issue tomorrow and move this out of draft then.

manovotn:
@bstansberry I have created https://issues.redhat.com/browse/WFLY-21134 to keep track of the removal.

Marking as ready for review.
The CI failures, as far as I can tell, don't seem related.

manovotn:
I have created https://issues.redhat.com/browse/WFLY-21151 to track this change alone and edited the PR and commit to reference it. That should make the bot happy.

If there are any other changes needed, let me know please.

jamezp:
I created an EE11 PR for the CI run #19368

jamezp (approved):

jasondlee (approved):

PR #19363, opened 2025-11-11
[WFLY-21131] Hibernate configured with hibernate.jndi.class no longer loads classes correctly due to deployment failure

scottmarlow:
https://issues.redhat.com/browse/WFLY-21131

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
@@ -94,6 +94,8 @@ public class PhaseOnePersistenceUnitServiceImpl implements Service<PhaseOnePersi
                             // run as security privileged action
                             @Override
                             public Void run() {
+                                ClassLoader oldTCCL = Thread.currentThread().getContextClassLoader();
+                                Thread.currentThread().setContextClassLoader(classLoader);
                                 try {
                                     ROOT_LOGGER.startingPersistenceUnitService(1, pu.getScopedPersistenceUnitName());
                                     pu.setTempClassLoaderFactory(new TempClassLoaderFactoryImpl(classLoader));
@@ -116,6 +118,7 @@ public class PhaseOnePersistenceUnitServiceImpl implements Service<PhaseOnePersi
                                 } catch (Throwable t) {
                                     context.failed(new StartException(t));
                                 } finally {
+                                    Thread.currentThread().setContextClassLoader(oldTCCL);
                                     pu.setTempClassLoaderFactory(null);    // release the temp classloader factory (only needed when creating the EMF)
                                     WritableServiceBasedNamingStore.popOwner();
                                 }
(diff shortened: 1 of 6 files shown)


jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
```diff
@@ -94,6 +94,8 @@ public void run() {
                             // run as security privileged action
                             @Override
                             public Void run() {
+                                ClassLoader oldTCCL = Thread.currentThread().getContextClassLoader();
+                                Thread.currentThread().setContextClassLoader(classLoader);
```
jamezp, line 98: These should probably use the `WildFlySecurityManager` to get and set the TCCL.

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
```diff
@@ -116,6 +118,7 @@ public Void run() {
                                 } catch (Throwable t) {
                                     context.failed(new StartException(t));
                                 } finally {
+                                    Thread.currentThread().setContextClassLoader(oldTCCL);
```
jamezp, line 121: Same note here, this should use `WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(oldTCCL)`.

jamezp (changes requested):

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
scottmarlow: Is it enough that we are running in a  `java.security.PrivilegedAction` block via `WildFlySecurityManager.doChecked(privilegedAction, accessControlContext)`?

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
scottmarlow: Is it enough that we are running in a  `java.security.PrivilegedAction` block via `WildFlySecurityManager.doChecked(privilegedAction, accessControlContext)`?

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
jamezp: Sorry, I missed that.

jpa/subsystem/src/main/java/org/jboss/as/jpa/service/PhaseOnePersistenceUnitServiceImpl.java
jamezp: Sorry I missed that

jamezp (approved):

PR #19370, opened 2025-11-13
[WFLY-20930] ElytronRemoteOutboundConnectionTestCase intermittently fails

elguardian:
Issue: https://issues.redhat.com/browse/WFLY-20930

start the container in @before annotation could cause race conditions and bad update operations.
clean up properly using snapshots insteado manual removing one by one.

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
@@ -1579,22 +1579,14 @@ public class ElytronRemoteOutboundConnectionTestCase {
     }
 
     private static ModelNode executeWithThrow(final ModelControllerClient client, ModelNode operation) {
-        long start = System.currentTimeMillis();
-        while (System.currentTimeMillis() - start < 15000) {
-            try {
-                try {
-                    ModelNode outcome = client.execute(operation);
-                    if (isSuccefulExecution(outcome)) {
-                        log.debugf("executeWithThrow outcome:\n%s" + outcome.toString());
-                        return outcome;
-                    }
-                } catch (Exception e) {
-                    // do nothing
-                }
-                Thread.sleep(100);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
+        try {
+            ModelNode outcome = client.execute(operation);
+            if (isSuccefulExecution(outcome)) {
+                log.debugf("executeWithThrow outcome:\n%s" + outcome.toString());
+                return outcome;
             }
+        } catch (Exception e) {
+            throw new RuntimeException("error during execute with throw", e);
         }
         throw new RuntimeException ("Failed to executed operation\n" + operation.toString());
     }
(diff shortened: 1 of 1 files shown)


rhusar (changes requested):
Tests failed.

```
java.lang.RuntimeException: {
    "outcome" => "failed",
    "failure-description" => "WFLYCTL0091: No files beginning with 'ElytronRemoteOutboundConnectionTestCase' found in /opt/buildAgent/work/e34a6f994de9f7c6/testsuite/integration/manualmode/target/outbound-server/standalone/configuration/standalone_xml_history/snapshot",
    "rolled-back" => true
}
```

rhusar:
@elguardian the frequent pushes are telling me this is not ready for review. Converting to draft. Please convert to regular PR once ready for review, thanks!

elguardian:
@rhusar the failure is unrelated. The problem in this case was that applyUpdate at some point was causing a model controller client losing connection against the server. This was a race conditions as the reconnection is automatic but when it did not happen in time was causing failures in 1 or more tests. retrying the operations of reloading did work.

bstansberry (commented):
This does too much locally maintained management utility code work instead of relying on battle-tested utilities.

I think the key reason for that is the type of serverSideMCC and clientSideMCC is ModelControllerClient instead of org.jboss.as.arquillian.container.ManagementClient. The ManagementClient constructor is public, and you have access to the data needed for its parameters, so I suggest using that.

Once you have ManagementClient instances, our standard utilities are more accessible. The org.jboss.as.test.shared.ServerSnapshot  class should be used for dealing with snapshots. org.jboss.as.test.shared.ServerReload should be used for reloads.

I don't think there's a need to delete snapshots. They are just a file on disk that is removed the next time the CI server agent moves on to another job or the local dev builds the manualmode module again.

I'm not clear why the existing test was capturing and trace logging the server state, and now the PR is checking for multiple legal states. When the container starts, its state is running. At points where the test code knows it has put the server into reload-required but needs it to be running, reload it. ServerReload.reloadIfRequired() is an option if you're not sure, but I suspect in most cases here we should be sure. And in case it seems the only downside to ServerReload.reloadIfRequired() is it doesn't let the caller control the timeout. But it's pretty unusual for the caller to need to control the timeout. The default timeout is fine.

elguardian:
@bstansberry 
To brief a bit

After applyUpdates to the server side the remoting client side component was getting a java.util.concurrent.Cancellation exception due to close all request from server side remoting component. (I did presume that is normal as you are changing configurations on some endpoints.

That was causing a race condition. At least, in model client controller, if you try to operate (even a reload operation) it is failing causing the test to fail and causing a cascade of failures (as there is no connection at that point in time). If we had a test not messing with the remoting configuration, that would be ok as you don't have to deal with this remoting close connection problem).

Reusing ServerReaload is fine but does not check this sort of problem. For instance if you try to check the status of the server while is still reconnecting it will fail. And It will be only useful in the case of default values as this is set on stone in the logic (and that is the reason this operates this was duplicated previously in this test).

I did not see any way to check if the client is connected or how many endpoint address are available that is the reason I had to rely on polling (the same way other operations does).

bstansberry:
How do the configuration changes being done here relate to the management endpoint? It uses a separate http-authentication-factory, security-realm and security-domain from application endpoints. AIUI the test is only touching the configuration of the endpoint used by applications.

If that is the case, if the management client is failing something is going wrong. It's understood that a MCC can fail if the connection closes while the client is executing a request. That's why ServerReload.waitForLiveServerToReload uses a separate MCC. Used properly the MCCs in this test should not be being used when a connection closes.

elguardian:
Hi @bstansberry 

"That's why ServerReload.waitForLiveServerToReload uses a separate MCC. Used properly the MCCs in this test should not be being used when a connection closes."

And I think this is a problem. If you reload you are ensuring that the server is up and running but you don't ensure that your other client (object) are connected and in sync (AFAIK it takes time till the reconnection logic kicks in) so while that ServerReload works it does not mean it works in combination with your current client being opened. Maybe I am missing something but even if Xnio is static they mantain separate endpoints at client leve (remoting component)l.

AFAIK if you reload the server you lose connection at mgmt level as well. Am I wrong ?

Let me know your thoughts on this.

elguardian:
superseeded by https://github.com/wildfly/wildfly/pull/19411

PR #19371, opened 2025-11-13
[WFLY-20562] WildFly can eat exceptions

HliasMpGH:
Issue: https://issues.redhat.com/browse/WFLY-20562

### DIFF
--- a/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
+++ b/batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
@@ -488,6 +488,8 @@ public class JobOperatorService extends AbstractJobOperator implements WildFlyJo
                             }
                         }
                     }
+                } catch (IllegalStateException e) {
+                    BatchLogger.LOGGER.error(e);
                 } finally {
                     WildFlySecurityManager.setCurrentContextClassLoaderPrivileged(current);
                     // Reset the stopped state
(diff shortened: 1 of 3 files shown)


rhusar:
IIRC @jamezp for review

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
```diff
@@ -488,6 +488,8 @@ private void stopRunningJobs(final boolean queueForRestart) {
                             }
                         }
                     }
+                } catch (IllegalStateException e) {
+                    BatchLogger.LOGGER.error(e);
```
jamezp: We probably want more information about the job here. We should add a new logger message like you did above with some information about the job.

jamezp (commented):
I need to look again to see if there is a way we can create a dependency between the services. While this fix is still valid IMO, it should be avoided if we can set the relationship up between the two services.

batch-jberet/src/main/java/org/wildfly/extension/batch/jberet/deployment/JobOperatorService.java
HliasMpGH: If the repository has indeed stopped, the exception is thrown in this call: `final Collection<String> jobNames = getJobRepository().getJobNames();` (line 469), before any attempt to stop a specific job, and thus there is no job name to use for the exception message. Maybe a change  to the exception message to "Unable to retrieve jobs because the service JobRepositoryService has been stopped and cannot execute operations." is more appropriate. What do you think?

jamezp: Ah, I see what you mean. Yes, I think something descriptive would be good even if it is "Failed to stop running jobs" and then pass in the exception.

jamezp (approved):

PR #19380, opened 2025-11-14
WFLY-21166 Intermittent failures in [ProtoStream]JSFFailoverTestCase with ViewExpiredException

rhusar:
The viewStatePattern regex is matching the 'id' attribute instead of the standard 'name' attribute for jakarta.faces.ViewState
Additionally, it couldn't handle different HTML attribute ordering.
Moreover, it could not handle line breaks or additional whitspace. 

These **may** be the cause of the test flakiness we have been observing – nevertheless, this PR adds proper logging to have more information for subsequent failures.

Other changes:
- Move Pattern objects to static fields for better performance
- Add comprehensive logging to help diagnose future issues
- Add toString() method to NumberGuessState for better log output

Resolves
https://issues.redhat.com/browse/WFLY-21166


pferraro (approved):
Not ideal, but let's get this merged so we can better see the root cause of the failures.

PR #19385, opened 2025-11-18
WFLY-21156 Create clustering test case for creating a custom DefaultCacheManager (not using WF's manager), configured from an XML file and containing a <file-store>

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-21156

The primary intention of this test (apart from detecting future regressions) is to be backported downstream.

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/infinispan/xml/deployment/infinispan.xml
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/infinispan/xml/deployment/infinispan.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<infinispan xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="urn:infinispan:config:15.2 https://infinispan.org/schemas/infinispan-config-15.2.xsd"
+            xmlns="urn:infinispan:config:15.2">
+    <cache-container name="test" default-cache="test">
+        <local-cache name="test">
+            <persistence>
+                <file-store path="target/infinispan-xml-test-case"/>
+            </persistence>
+        </local-cache>
+    </cache-container>
+</infinispan>
+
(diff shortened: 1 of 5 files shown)


testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/infinispan/xml/deployment/infinispan.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<infinispan xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="urn:infinispan:config:15.0 https://infinispan.org/schemas/infinispan-config-15.0.xsd"
```
pferraro: 15.2?

pferraro (approved):
Comment is optional.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/infinispan/xml/deployment/infinispan.xml
rhusar: Hm, I guess that's the version from the original reproducer, but there is no reason to not use the latest. Fixing.

rhusar: @pferraro FWIW that infinispan schema is broken:

    <xs:attribute name="timeout" type="xs:long" default="4m">

```
Error:(2298, 62) a-props-correct.2: Invalid value constraint value '4m' in attribute 'timeout'.
```

🤦🏼

rhusar: Nevertheless, updated to latest now.

rhusar: n.b. the schema issue is already fixed upstream.

pferraro: Nice. :/  Though, I don't understand why they didn't just go with ISO 8601 format.
Either way, it would have been easy enought to create a custom XSD type for TimeQuantity values using a pattern, but it looks like they generally use xs:string everywhere else.  Not sure why/how this one was skipped.

rhusar:
Thanks @pferraro

PR #19387, opened 2025-11-19
WFLY-13853 Upgrade download-maven-plugin post 1.6.7 to version managed by jboss-parent

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-13853


jamezp (approved):

rhusar:
n.b. I reran the jobs to make sure there is no longer the problem of bad caching as there used to be.

rhusar:
Thanks for the review @jamezp!

PR #19395, opened 2025-11-20
[WFLY-21190] ensure the microprofile-certification.sh is executed by …

istudens:
…Bash

Issue: https://issues.redhat.com/browse/WFLY-21190


rhusar (approved):
Good catch, this was never a posix compatible shell script as it's using pipefail and += concatenations.

rhusar:
Thanks @istudens !

PR #19399, opened 2025-11-21
[WFLY-20485] Integrate app dev guide

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-20485

Please see the JIRA details for the changes description.


bstansberry:
@emmartins This integration approach seems fine; it's what we talked about.

pom.xml
```diff
@@ -841,6 +842,12 @@
                 <type>pom</type>
             </dependency>
 
+            <dependency>
+                <groupId>org.wildfly</groupId>
```
rhusar, line 846: I don't think we want to be using 'org.wildfly' namespace for things that are not directly in this repo. Each repo should have it's own groupId.

rhusar (changes requested):

pom.xml
emmartins: @rhusar I am not aware of such limitation, but if that's the case I will change... @bstansberry can you please confirm?

rhusar: We haven't historically paid much attention to the groupIds but the nexus migration shown us that we should...

emmartins: Sorry but I forgot about this comment, and the guide was released already with the `org.wildfly` groupId :(

emmartins:
This is now ready to review and merge.

pom.xml
rhusar: So next time an intern is going to release the guide he/she will have to have permissions to release entire org.wildfly... 

😢

darranl: I agree we do need to think about how we use nested namespaces, as we have worked closely with the Nexus repositories we can see a lot of room for improvement but I don't think there is a lot we can do when we reach the PR review stage.

darranl:
@bstansberry Do we want this in for Beta? I know we don't publish docs for Beta but we do want it in any to test for release process side effects?

bstansberry (approved):
@darranl I'll put the 39.x label on this. If getting it into Beta works for you, that's fine with me. But if it doesn't it's not a big deal.

pom.xml
bstansberry: We should change the groupId for 40 but it's not an urgent problem. No one but this maven module is going to care about the GA of this, so we can change it at will.

bstansberry: Note there are other 'problematic' projects re groupId, including my very own mvc-krazo-feature-pack. :-( Not the end of the world, just a kind of housekeeping.

emmartins:
@theashiot FYI :)

pom.xml
emmartins: @bstansberry WFLY-21278 created for that

theashiot:
Thanks, all!

PR #19400, opened 2025-11-21
WFLY-21166 Make JSF parameters explicit for clustering in [ProtoStream]JSFFailoverTestCase and temporarily increase number of logical views to keep

rhusar:
More investigation towards
https://issues.redhat.com/browse/WFLY-21166

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<web-app xmlns="https://jakarta.ee/xml/ns/jakartaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
+         version="6.0">
+
+    <distributable/>
+
+    <!-- Make JSF parameters explicit and optimized for clustering -->
+
+    <!-- Reference: https://github.com/eclipse-ee4j/mojarra/blob/4.0/impl/src/main/java/jakarta/faces/application/StateManager.java -->
+    <context-param>
+        <param-name>jakarta.faces.STATE_SAVING_METHOD</param-name>
+        <param-value>server</param-value>
+    </context-param>
+    <!-- Reference: https://github.com/eclipse-ee4j/mojarra/blob/4.0/impl/src/main/java/com/sun/faces/config/WebConfiguration.java#L752-L786 -->
+    <context-param>
+        <param-name>com.sun.faces.serializeServerState</param-name>
+        <!-- n.b. always use 'false' to let the session manager control when session attributes are serialized and to not force JDK serialization of JSF state -->
+        <param-value>false</param-value>
+    </context-param>
+    <context-param>
+        <param-name>com.sun.faces.compressViewState</param-name>
+        <!-- n.b. defaults to 'true' in mojarra 4.0 -->
+        <param-value>false</param-value>
+    </context-param>
+    <!-- Temporarily bump the view limits high enough to definitely not hit any of the LRU size limits -->
+    <context-param>
+        <param-name>com.sun.faces.numberOfViewsInSession</param-name>
+        <param-value>100</param-value>
+    </context-param>
+    <context-param>
+        <param-name>com.sun.faces.numberOfLogicalViews</param-name>
+        <param-value>100</param-value>
+    </context-param>
+
+</web-app>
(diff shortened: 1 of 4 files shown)


rhusar:
@pferraro This bumps the view limits high enough to definitely not hit any of the LRU size plus makes all clustering options explicit according to what seem to be Morraja recommendations (which also serves as a demonstration for how this could be setup for clustering).  Let me know how you want to proceed with this.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
```diff
@@ -0,0 +1,36 @@
...
+        <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
+        <param-value>server</param-value>
+    </context-param>
+    <context-param>
+        <param-name>com.sun.faces.serializeServerState</param-name>
+        <param-value>true</param-value>
+    </context-param>
+    <context-param>
+        <param-name>com.sun.faces.compressViewState</param-name>
+        <param-value>true</param-value>
```
pferraro: Disable this.  This is intended for use when STATE_SAVING_METHOD is "client".
Also, this does not do anything unless com.sun.faces.serializeServerState is enabled, which, again, we do not want enabled.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <!-- Make JSF parameters explicit for clustering and overly graceful -->
+
+    <context-param>
+        <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
+        <param-value>server</param-value>
+    </context-param>
+    <context-param>
+        <param-name>com.sun.faces.serializeServerState</param-name>
+        <param-value>true</param-value>
```
pferraro: This setting is not safe to enable.  The session manager should control when session attributes are serialized.  It also forces JDK serialization of JSF state - which we also do not want.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
```diff
@@ -0,0 +1,36 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/web-app_6_0.xsd"
+         version="6.0">
+
+    <distributable/>
+
+    <!-- Make JSF parameters explicit for clustering and overly graceful -->
+
+    <context-param>
+        <param-name>javax.faces.STATE_SAVING_METHOD</param-name>
```
pferraro: This parameter name is incorrect. See:
https://github.com/eclipse-ee4j/mojarra/blob/4.0/impl/src/main/java/jakarta/faces/application/StateManager.java

pferraro (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
rhusar: So they decided to rename javax but kept com.sun prefix. Interesting decision. 

Fixed.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
rhusar: OK, disabled since this is enabled by default. Though, this is theoretically usable for both client and server methods.

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/jsf/web.xml
rhusar: @pferraro OK I see I forgot to explain the motivation for this: this is meant to further isolate the problem being investigated. If this continues to fail, we can (mostly) rule out protostream and LRU view cache.

Nevertheless, added the a comment note about this, that this should be disabled.

rhusar: OK discussed over chat that this isn't helpful for investigation, reverted the flipping.

rhusar:
n.b. opened a Jira to document the naunces of configuring this for HA - see https://issues.redhat.com/browse/WFLY-21201

rhusar:
Thread safety issue has been discovered upstream and I have submitted fix for it, so we might not need this one after all

https://issues.redhat.com/browse/WFLY-21166?focusedId=28537349&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28537349

Making this a draft then.

rhusar:
Investigation is completed and issue has been found, reported and submitted fix for - https://github.com/eclipse-ee4j/mojarra/issues/5617

This change is no longer necessary – closing.

rhusar:
Upgrade required to fix the test - https://issues.redhat.com/browse/WFLY-21235

PR #19403, opened 2025-11-23
Add more assertions to the Micrometer subsystem transformer test

bstansberry:
No JIRA

I was investigating #17211 and wanted to better understand how ResourceTransformationDescriptionBuilder.setCustomResourceTransformer works so I added some asserts to this test, which tests the only use of that in WF. They're valid asserts so I figured why not commit them and send up a PR.


rhusar (approved):
LGTM

jasondlee (approved):

rhusar:
Thanks @bstansberry and @jasondlee

PR #19410, opened 2025-11-26
[WFLY-21208] Switch from docs.jboss.org/hibernate to docs.hibernate.org in the documentation

marko-bekhta:
- https://issues.redhat.com/browse/WFLY-21208

I only found these few references in the jakarta data doc. Let me know if there are any other docs that we should update, or let the corresponding people know 🙂 
thanks!


bstansberry (approved):

bstansberry:
Thanks @marko-bekhta.

@emmartins @theashiot FYI about this in case the same change is needed in https://github.com/wildfly/wildfly-application-development-book

PR #19411, opened 2025-11-27
[WFLY-20930] ElytronRemoteOutboundConnectionTestCase intermittently fails

elguardian:
Issue: https://issues.redhat.com/browse/WFLY-20930

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
@@ -18,7 +18,6 @@ import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.PRO
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.READ_RESOURCE_OPERATION;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REALM;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.REMOTE_DESTINATION_OUTBOUND_SOCKET_BINDING;
-import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.RESULT;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SASL_AUTHENTICATION_FACTORY;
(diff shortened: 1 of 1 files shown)


bstansberry (approved):

rhusar:
/retest

(hash problem again..)

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
```diff
@@ -1063,6 +1024,8 @@ public void testElytronWithBareRemotingWithSecurityRealmRole() {
      */
     @Test
     public void testElytronWithBareRemotingWithSourceAddressRoleDecoderInvalidCredentials() {
+        ModelControllerClient serverSideMCC = getInboundConnectionServerMCC();
+        ModelControllerClient clientSideMCC = getOutboundConnectionServerMCC();
```
rhusar: Same two lines at the beginning of each test = `@org.junit.Before`

https://junit.org/junit4/javadoc/4.12/org/junit/Before.html

rhusar (changes requested):

testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/client/outbound/connection/security/ElytronRemoteOutboundConnectionTestCase.java
elguardian: moved. Still this could lead to cascade of failures of the mcc fails in one test. Anyhow it should not fail.
I also reload the logic using ServerReload and arquillian Managmentclient class that holds inside and mcc already.

rhusar: ok, resolving discussion

rhusar (approved):
LGTM now

rhusar:
@elguardian For future, squash those commits, there is no need to have 2 separate commits for single issue like this. But since you will be back-porting this, you will have to deal with it yourself :)

Given where we are in the cycle, I will merge this as is since our merge policies do not allow squash merging and I don't want to put more load on CI since we need it for other work.

rhusar:
Thanks @elguardian @bstansberry 

@elguardian You can proceed with the back-port now.

PR #19412, opened 2025-11-27
[WFLY-20659] Make use of ModuleIdentifierUtil.MODULE_NAME_CORRECTOR on the attributes that could require it

mskacelik:
issue: https://issues.redhat.com/browse/WFLY-20659

Hopefully I applied correctly all the correctors.

### DIFF
--- a/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
+++ b/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
@@ -13,6 +13,7 @@ import org.jboss.as.connector.metadata.api.resourceadapter.WorkManagerSecurity;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.AttributeMarshaller;
 import org.jboss.as.controller.ModelVersion;
+import org.jboss.as.controller.ModuleIdentifierUtil;
 import org.jboss.as.controller.ObjectListAttributeDefinition;
 import org.jboss.as.controller.ObjectTypeAttributeDefinition;
 import org.jboss.as.controller.PropertiesAttributeDefinition;
@@ -236,6 +237,7 @@ public class Constants {
             .setRequired(false)
             .setAllowExpression(false)
             .setMeasurementUnit(MeasurementUnit.NONE)
+            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
             .setAttributeMarshaller(new AttributeMarshaller() {
                 @Override
                 public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceModel, boolean marshallDefault, XMLStreamWriter writer) throws XMLStreamException {
--- a/naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
+++ b/naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
@@ -6,6 +6,7 @@
 package org.jboss.as.naming.subsystem;
 
 import org.jboss.as.controller.AttributeDefinition;
+import org.jboss.as.controller.ModuleIdentifierUtil;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
 import org.jboss.as.controller.OperationStepHandler;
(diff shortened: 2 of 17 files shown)


connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
```diff
@@ -236,6 +237,7 @@ public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceM
             .setRequired(false)
             .setAllowExpression(false)
             .setMeasurementUnit(MeasurementUnit.NONE)
+            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```
rhusar:
```suggestion
            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```

Line has trailing spaces, this won't pass checkstyle.

rhusar (changes requested):

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
mskacelik: thanks, did not notice that.

rhusar:
@yersan could you please review (since you are the reporter)?

yersan:
@rhusar on it ... but requires a careful review since the ultimate goal by moving the module name corrector to the attributes, is also to avoid parsing to the canonize form twice

darranl (commented):
I have the same question that I had in the Core PR - does this change mean that values that were previously accepted as valid and converted by canonicalisation will now be rejected?

If that is the case I don't think we should make a change like that.

yersan:
> I have the same question that I had in the Core PR - does this change mean that values that were previously accepted as valid and converted by canonicalisation will now be rejected?
> 
> If that is the case I don't think we should make a change like that.

@darranl  Values should not be rejected. We are only adding a parameter corrector to the attributes themselves to always assume that the input for module names is always in its canonicalized form, that is, if the input contains a slot with the name `main`, that slot is removed from the input.

The idea has always been to place this sort of conversions/adjustments as close as possible to the external inputs, hence on the attribute definitions via [ModuleIdentifierUtil.MODULE_NAME_CORRECTOR](https://github.com/wildfly/wildfly-core/blob/b5ab64888b0df01259762f0b9d575ef2fb08b750/controller/src/main/java/org/jboss/as/controller/ModuleIdentifierUtil.java#L98)

connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/Constants.java
```diff
@@ -236,6 +237,7 @@ public void marshallAsElement(AttributeDefinition attribute, ModelNode resourceM
...
```
yersan, line 240: I have my doubts about this one. The module name is later used as an identifier to get the deployment name here: https://github.com/wildfly/wildfly/blob/1f68b0ad54b7899650a782a8efdde0cec196865c/connector/src/main/java/org/jboss/as/connector/subsystems/resourceadapters/AdminObjectAdd.java#L94

And the attribute itself is always being marshalled with ":main" included as a slot if there is none.

naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingResourceDefinition.java
```diff
@@ -66,6 +67,7 @@ public class NamingBindingResourceDefinition extends SimpleResourceDefinition {
     static final SimpleAttributeDefinition MODULE = new SimpleAttributeDefinitionBuilder(NamingSubsystemModel.MODULE, ModelType.STRING, true)
             .setAllowExpression(true)
             .setFlags(AttributeAccess.Flag.RESTART_RESOURCE_SERVICES)
+            .setCorrector(ModuleIdentifierUtil.MODULE_NAME_CORRECTOR)
```
yersan, line 70: I'm still unsure about this one, since it is later used at https://github.com/wildfly/wildfly/blob/e3420eff307e23c0ce85ae6c04e4ddea57f3e6e7/naming/src/main/java/org/jboss/as/naming/subsystem/NamingBindingAdd.java#L195

I don't know yet whether the one that reads the INITIAL_CONTEXT_MODULE is prepared for canonicalized module names.

yersan (commented):
I still have some doubts, so added an initial review, it would be nice another pair of eyes here

darranl (changes requested):
Unless we have a valid reason to require it I don't plan to dig into the management models of legacy components - I would propose we leave PicketLink and security out of any changes.

Although it generates more PRs I would still recommend PRs like this are split up by component to enable individual review - when submitted like this some ends up with the task of cross checking have all the relevent components provided a review and the last component to review tends to set the time to merge when I suspect many could probably have been merged last year.

PR #19418, opened 2025-11-28
WFLY-21223 Upgrade com.h2database:h2 from 2.2.224 to 2.4.240 in preview

rhusar:
Fix parameter duplication in testsuite/pom.xml

Resolves
https://issues.redhat.com/browse/WFLY-21223

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml
@@ -6,7 +6,7 @@
 <subsystem xmlns="urn:jboss:domain:datasources:7.2">
     <datasources>
         <drivers>
-            <driver name="name" module="com.h2database.h2" major-version="2" minor-version="2">
+            <driver name="name" module="com.h2database.h2" major-version="2" minor-version="4">
                 <driver-class>org.h2.Driver</driver-class>
                 <xa-datasource-class>org.h2.jdbcx.JdbcDataSource</xa-datasource-class>
             </driver>
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml
@@ -6,7 +6,8 @@
 <subsystem xmlns="urn:jboss:domain:datasources:7.2">
     <datasources>
         <drivers>
-            <driver name="name" module="com.h2database.h2" major-version="2" minor-version="4">
+            <!-- TODO this test driver configuration used to also specify minor-version="2" which would however fail against preview which uses 2.4 already -->
+            <driver name="name" module="com.h2database.h2" major-version="2">
                 <driver-class>org.h2.Driver</driver-class>
                 <xa-datasource-class>org.h2.jdbcx.JdbcDataSource</xa-datasource-class>
             </driver>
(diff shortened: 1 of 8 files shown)


rhusar:
No idea why GH closed this.

bstansberry:
Our H2 dep is driven by what the ORM version we integrate uses/tests as part of testing its H2 dialect. So this should come into WF Preview along with the upgrade to ORM 7.2, which uses it. ORM 6.6 still use 2.2.224.

rhusar:
> Our H2 dep is driven by what the ORM version we integrate uses/tests as part of testing its H2 dialect. So this should come into WF Preview along with the upgrade to ORM 7.2, which uses it. ORM 6.6 still use 2.2.224.

@bstansberry Now that is quite unfortunate that we need to keep this dependency back because ORM has not adapted to newer versions as ORM is not the sole user of h2.

I started testing how ORM is going to be compliant with this new version of H2 and it does indeed seem to generate a queries that are no longer compatible; e.g. https://gist.github.com/rhusar/8883f2fd8787d9b2e637af0e216f1bf5

I will add a revertable commit to only upgrade in preview as you suggested.

n.b. none of this information is captured anywhere in our pom, so this is an easy trap to fall into.

bstansberry:
ORM has adapted to newer versions. ORM 6.6 is a maintenance branch.

Standard WildFly is somewhat behind the ecosystem as we wait to complete EE 11 compliance and catch up. (ORM 7.x is a Jakarta Persistence 3.2 / EE 11 impl.) The sooner we get std WF to EE 11 the better.

Fair point about the pom.xml not having a comment by the version prop discussing how managed that dep.

rhusar:
@bstansberry OK, thanks for pointing out and discussing the idiosyncrasies of H2 version management in standard vs preview. I have added comments in both places so nobody has to go through these hoops again.

I have structured this change set to two commits where the 2nd commit is revertable; reverting of which will upgrade this properly in standard and pull out of preview.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/metrics/data-sources/complex-driver.xml
```diff
@@ -6,7 +6,7 @@
 <subsystem xmlns="urn:jboss:domain:datasources:7.2">
     <datasources>
         <drivers>
-            <driver name="name" module="com.h2database.h2" major-version="2" minor-version="4">
+            <driver name="name" module="com.h2database.h2" major-version="2" minor-version="2">
```
bstansberry: We'll probably have to use an expression for this and drive the system property based on what's being tested.

Unless the uses of this config aren't really deriving any value from setting this, in which case we can just remove this. AIUI these two version attributes are used to compare the config against what the driver itself reports, failing if there's a mismatch. I assume that's why the test fails in the WildFly Preview job. If an attribute isn't set it doesn't do the comparison. I don't see any other use.

rhusar: @bstansberry I am actually familiar with this test and somewhat did expect it to fail, I just didn't have time to test the preview combination of the build yesterday.

I put in in a change that reducing verification of the major version and leave out the minor version with explanatory TODO comment - which would also be reverted in the revertable commit as described above.

If you think that's a problem, let me know, I can start looking for different solutions, but personally, I don't think it's worth the effort.

bstansberry: +1. I filed https://issues.redhat.com/browse/WFLY-21226 for this.

bstansberry (approved):

bstansberry:
Thanks @rhusar

scottmarlow:
Thanks @rhusar + @bstansberry !

PR #19419, opened 2025-11-28
WFLY-21206 Restore FD_SOCK2 to default TCP-based stacks

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-21206
https://issues.redhat.com/browse/WFLY-21180

WFLY-21206 Add back configuration parts of WFLY-20710 while keeping auto-enabling suspect events if the stack does not otherwise contain a socket-based failure-detection protocol
Revert "WFLY-20710 Auto-enable transport-based failure detection if protocol stack does not contain a socket-based failure detection protocol. Drop FD_SOCK2 from default TCP protocol stack."
This reverts commit 6e8e451ecc0eec60e53521e691c68042984beeb4.

Opening as draft to confirm whether this does actually fix the problem.

____

> [!WARNING]

> * [WFLY-20710](https://issues.redhat.com/browse/WFLY-20710)

### DIFF
--- a/ee-feature-pack/galleon-local/src/main/resources/feature_groups/domain-ec2-full-ha.xml
+++ b/ee-feature-pack/galleon-local/src/main/resources/feature_groups/domain-ec2-full-ha.xml
@@ -55,6 +55,9 @@
                     <feature spec="subsystem.jgroups.stack.protocol">
                         <param name="protocol" value="MERGE3"/>
                     </feature>
+                    <feature spec="subsystem.jgroups.stack.protocol.FD_SOCK2">
+                        <param name="socket-binding" value="jgroups-tcp-fd"/>
+                    </feature>
                     <feature spec="subsystem.jgroups.stack.protocol">
                         <param name="protocol" value="FD_ALL3"/>
                     </feature>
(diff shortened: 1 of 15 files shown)


rhusar:
First set of CI has passed, now I am triggering more runs to see whether this fixes the problem:

https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk17?branch=pull%2F19419&buildTypeTab=overview
https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21?branch=pull%2F19419&buildTypeTab=overview

@pferraro FYI

Update Nov 29: so far no failures regarding topology, triggering another 20 runs over the weekend when CI is available to gather more data

rhusar:
@pferraro OK so with the results above, this investigation confirms that this does fix the issue and stabilize the test suite. There were dozens of runs, each IIRC runs the web failover test case ~17 times, so that's way past the previous failure rate.

I propose going ahead with this right now and keep looking for solution for getting transport-based failure detection reliable.

pferraro (changes requested):
I'm fine with temporarily restoring FD_SOCK2 to testsuite protocol stack, but please revert changes to default configuration.

rhusar:
@pferraro OK, this PR is intended for situation that we are not able to find and fix the bug within the release time. I admit it's on the pessimistic spectrum. Fair enough.

In that case, I will make it a draft if it's needed and only change the clustering test suite so we can get CI stable again: https://github.com/wildfly/wildfly/pull/19424

rhusar:
This is not needed, closing unmerged.

PR #19421, opened 2025-12-01
WFLY-21225 OidcIdentityPropagationTestCase and related tests are flaky because getAdminAccessToken(..) is not resilient

rhusar:
https://issues.redhat.com/browse/WFLY-21225

The code obtaining the admin token is poorly written and not resilient enough. 

While the container is booting it keeps hammering it with requests it busy waits without backoff interval, it fails even if proper token is returned but after the deadline (unnecessary failure), it does have any http timeouts set, it returns null instead of propagating the correct failure cause, among other things like code duplication.


bstansberry (approved):

rhusar:
CI failure unrelated - flaky test trackes as https://issues.redhat.com/browse/WFLY-16653

pferraro (approved):

rhusar:
Thanks @bstansberry and @pferraro !

PR #19423, opened 2025-12-02
[WFLY-21216] Broken docs.jboss.org links in Elytron migration documentation

mskacelik:
issue: https://issues.redhat.com/browse/WFLY-21216

Proposal for the link fixes, if aprooved I might take a look on other subtasks or creating another issues, since I believe there might be more invalid links (e.g., some links in https://docs.wildfly.org/wildfly-proposals/console/HAL-1471_Support_for_Keycloak_SSO.html)

### DIFF
--- a/docs/src/main/asciidoc/_elytron/migrate/Kerberos_Based_Authentication_Migration.adoc
+++ b/docs/src/main/asciidoc/_elytron/migrate/Kerberos_Based_Authentication_Migration.adoc
@@ -201,16 +201,12 @@ Overall this results in the following configuration: -
 
 Now, to enable SPNEGO authentication for the HTTP management interface,
 update this interface to reference the `http-authentication-factory`
-defined above, as described in the
-https://docs.jboss.org/author/display/WFLY/Migrate+Legacy+Security+to+Elytron+Security#MigrateLegacySecuritytoElytronSecurity-MigratedConfiguration[properties
-authentication section].
+defined above, as described in the <<Migrated Configuration, properties authentication section>> of this documentation.
 
 Alternatively, to secure an application using SPNEGO authentication, an
 application security domain can be defined in the Undertow subsystem to
 map security domains to the `http-authentication-factory` defined above,
-as described in the
-https://docs.jboss.org/author/display/WFLY/Migrate+Legacy+Security+to+Elytron+Security#MigrateLegacySecuritytoElytronSecurity-FullyMigratedConfiguration[properties
-authentication section].
+as described in the <<Fully Migrated Configuration, properties authentication section>> of this documentation.
 
 [[remoting-sasl-authentication]]
 == Remoting / SASL Authentication


docs/src/main/asciidoc/_elytron/migrate/Kerberos_Based_Authentication_Migration.adoc
```diff
@@ -202,14 +202,14 @@ Overall this results in the following configuration: -
...
-https://docs.jboss.org/author/display/WFLY/Migrate+Legacy+Security+to+Elytron+Security#MigrateLegacySecuritytoElytronSecurity-MigratedConfiguration[properties
+https://docs.jboss.org/author/display/WFLY/Migrate%20Legacy%20Security%20to%20Elytron%20Security.html#119144556_MigrateLegacySecuritytoElytronSecurity-MigratedConfiguration[properties
 authentication section].
 
 Alternatively, to secure an application using SPNEGO authentication, an
 application security domain can be defined in the Undertow subsystem to
 map security domains to the `http-authentication-factory` defined above,
 as described in the
-https://docs.jboss.org/author/display/WFLY/Migrate+Legacy+Security+to+Elytron+Security#MigrateLegacySecuritytoElytronSecurity-FullyMigratedConfiguration[properties
+https://docs.jboss.org/author/display/WFLY/Migrate%20Legacy%20Security%20to%20Elytron%20Security.html#119144556_MigrateLegacySecuritytoElytronSecurity-FullyMigratedConfiguration[properties
```
rhusar: Nope! Look at the URL, this is linking deprecated jboss.org content. We don't want that.

This needs to link to docs in this repository; probably https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/_elytron/migrate/Properties_File_Based_Authentication_Migration.adoc

rhusar (changes requested):

docs/src/main/asciidoc/_elytron/migrate/Kerberos_Based_Authentication_Migration.adoc
mskacelik: @rhusar Ok, I reckon as a [cross reference](https://docs.asciidoctor.org/asciidoc/latest/macros/xref/)? Since it is part of the same adoc file.

rhusar: I believe so, yes.

rhusar:
> issue: https://issues.redhat.com/browse/WFLY-21216
> 
> Proposal for the link fixes, if aprooved I might take a look on other subtasks or creating another issues, since I believe there might be more invalid links (e.g., some links in https://docs.wildfly.org/wildfly-proposals/console/HAL-1471_Support_for_Keycloak_SSO.html)

@mskacelik Since you are offering, could you please do a quick scan and fix the other instances of this if you see any? e.g. searching for 'https://docs.jboss.org/author/display'

If not this is good as is 👍🏼 

n.b. that proposals example is from a different repo, located here https://github.com/wildfly/wildfly-proposals

bstansberry (approved):

rhusar (approved):

bstansberry:
Thanks @mskacelik! 

If you do more of these, cross references between other docs in this source tree folder also work (if that's applicable to any of the problems.)

Please be cautious about dealing with wildfly-proposals.  We publish those, so in a perfect world links would work. But they are fundamentally not live documentation; they are a historical record of what we decided to do about a feature. It's not necessarily the case that some replacing a link to some new URL would result in the same underlying meaning. And even if it did the change would require review by the SMEs for the feature, and TBH that doesn't seem like a great use of peoples' time.

rhusar:
Looking at the Jira I now see that https://issues.redhat.com/browse/WFLY-21209 is a subtask of the what I was asking for to review the whole docs section for these that Brian already filed.

This fixes all cases in _elytron guide 👍🏼

rhusar:
Thanks @mskacelik and @bstansberry

mskacelik:
> Looking at the Jira I now see that https://issues.redhat.com/browse/WFLY-21209 is a subtask of the what I was asking for to review the whole docs section for these that Brian already filed.
> 
> This fixes all cases in _elytron guide 👍🏼

Was about to ask -- since this is just an `_elytron` subtask -- but that solves it :). Thank you.

PR #19426, opened 2025-12-02
[WFLY-21229] Persistence trace logging in AbstractEntityManager class should check for null entity object

scottmarlow:
https://issues.redhat.com/browse/WFLY-21229

I got lucky with local TCK testing and noticed we can get a NullPointerException when org.jboss.as.jpa TRACE logging is enabled and the application code passes a null entity class to EntityManager.lock or other similar methods.

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/AbstractEntityManager.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/AbstractEntityManager.java
@@ -134,7 +134,7 @@ public abstract class AbstractEntityManager implements EntityManager {
         } finally {
             if (isTraceEnabled) {
                 long elapsed = System.currentTimeMillis() - start;
-                ROOT_LOGGER.tracef("detach entityClass '%s' took %dms", entity.getClass().getName(), elapsed);
+                ROOT_LOGGER.tracef("detach entityClass '%s' took %dms", getClassName(entity), elapsed);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


scottmarlow:
The test failures are not related to this pull request.

jamezp (approved):

jpa/subsystem/src/main/java/org/jboss/as/jpa/container/AbstractEntityManager.java
```diff
@@ -289,11 +289,15 @@ public void lock(Object entity, LockModeType lockMode, Map<String, Object> prope
         } finally {
             if (isTraceEnabled) {
                 long elapsed = System.currentTimeMillis() - start;
-                ROOT_LOGGER.tracef("lock entityClass '%s', lockMode '%s'  took %dms", entity.getClass().getName(), getLockModeAsString(lockMode), elapsed);
+                ROOT_LOGGER.tracef("lock entityClass '%s', lockMode '%s'  took %dms", getClassName(entity), getLockModeAsString(lockMode), elapsed);
```
rhusar: Since editing this, you might want to fix the double space here (and above)...

```suggestion
                ROOT_LOGGER.tracef("lock entityClass '%s', lockMode '%s' took %dms", getClassName(entity), getLockModeAsString(lockMode), elapsed);
```

scottmarlow: Thanks for catching that @rhusar !  I pushed a change to remote the double space.

rhusar: Looks good now. The only reason I brought it up is because when you see a logging with extra spaces, you start to wonder if there wasn't a " %s " somewhere that was empty. Then you look at the source code and realize it's just an extra space in the log message :) (been there done it, not here but in general)

rhusar (approved):

rhusar:
Thanks @scottmarlow and @jamezp !

PR #19433, opened 2025-12-05
[DO NOT MERGE - CI ONLY] Add temporary TRACE logging for org.jgroups to facilitate investigatation of WFLY-21236

rhusar:
Investigating https://issues.redhat.com/browse/WFLY-21236

### DIFF
--- a/testsuite/integration/clustering/clustering-ha-bootable-jar.cli
+++ b/testsuite/integration/clustering/clustering-ha-bootable-jar.cli
@@ -37,12 +37,6 @@ batch
 /subsystem=jgroups/stack=tcp-bridge/protocol=FRAG4:add
 run-batch
 
-# TODO Remove workaround - WFLY-21180 Failures in variants of *WebFailoverTestCase showing inconsistent topology
-/socket-binding-group=standard-sockets/socket-binding=jgroups-tcp-fd:add(interface=private,port=57600)
-/subsystem=jgroups/stack=tcp/protocol=FD_SOCK2:add(add-index=3,socket-binding=jgroups-tcp-fd)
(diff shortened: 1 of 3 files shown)


rhusar:
@pferraro FYI - for jgroups trace logs for this issue

rhusar:
@pferraro FYI rebased to pick up the JGroups 5.5 upgrade – I will be re-running until we hit the issue.

rhusar:
~~Reproduced here https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21/535821  (unfortunately its from the TLS adaptation of the test, which brings in additional context, but it appears to be the same cause)~~

This has expired since... pushed again to get some new runs and new reproductions. New runs:

https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21?branch=pull%2F19433&buildTypeTab=overview

rhusar:
> ~Reproduced here https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21/535821 (unfortunately its from the TLS adaptation of the test, which brings in additional context, but it appears to be the same cause)~
> 
> This has expired since... pushed again to get some new runs and new reproductions. New runs:
> 
> https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21?branch=pull%2F19433&buildTypeTab=overview

@pferraro FYI ^^^ There are 2 runs that probably reproduced the issue (and 2 additional ones but from TLS cases which bring in unnecessary complication). Unfortunately I can't open it atm since TC is not responding.

rhusar:
I have created a reproducer in upstream, so this should no longer be needed within WF

FYI https://github.com/belaban/JGroups/blob/master/tests/junit-functional/org/jgroups/tests/EnableSuspectEventsTest.java (just lower the timeout and increase interations to reproduce)

PR #19434, opened 2025-12-05
WFLY-21243 mod_cluster ssl-context and advertise-socket references allow expressions but they cannot since they are capability references + WFLY-21242 Update deprecated references for org.jboss.as.clustering.controller.CommonUnaryRequirement in mod_cluster subsystem

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-21243
https://issues.redhat.com/browse/WFLY-21242


rhusar:
@bstansberry If you have a minute, could you please check that this is the correct way to disallow expressions that were incorrectly enabled for capability references?

Moreover, I am trying to understand why doesn't `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setCapabilityReference(java.lang.String)` imply `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setAllowExpression(false)`? I am failing to see the case where this would be allowable.

darranl (commented):
Although expressions are not allowed could they have worked?

rhusar:
> Although expressions are not allowed could they have worked?

@darranl Well, that's nuanced. In standalone, the model is resolved before capability is referenced, so theoretically this "does work". This is problem for domain, because domain cannot resolve expressions and thus cannot validate the model. That's why we don't allow expressions for capability references.

darranl:
If " theoretically this "does work"" I think we need to consider how we handle server migration with @emmartins before we make a change like this.

pferraro:
@darranl We only support the migratation of valid legacy configuration. Rejecting invalid legacy configuration does not require, nor has it ever required, server migration rules.

darranl:
However to the user who's configuration is now broken, their perspective is that it was a valid configuration - it may have been working for them for a long time!

rhusar:
> However to the user who's configuration is now broken, their perspective is that it was a valid configuration - it may have been working for them for a long time!

@darranl False.

We have been actually warning our users that we will fix that for years now, specifically since 2017.

> 18:23:24,082 WARN  [org.jboss.as.controller.management-deprecated] (ServerService Thread Pool -- 33) WFLYCTL0447: Attribute 'advertise-socket' in the resource at address '/subsystem=modcluster/proxy=default' has been configured with an expression, but support for use of expressions in this attribute's value may be removed in a future version. This attribute configures whether a capability that can be required by other parts of the configuration is present or itself configures a requirement for a capability provided by another part of the configuration. Full support for this kind of configuration cannot be provided when an expression is used.

darranl:
That is great, thank you @rhusar

pferraro:
@darranl Invalid configuration that does not fail validation rules != valid configuration.

pferraro (approved):

darranl:
I think @rhusar has provided the context I was looking for, it sounds like we have been giving users an opportunituy to address this for quite a while now.

bstansberry (approved):
What a nice log WARN message. Buy whoever wrote that a beer!

I'd forgotten all about that.

bstansberry:
> @bstansberry If you have a minute, could you please check that this is the correct way to disallow expressions that were incorrectly enabled for capability references?
> 
> Moreover, I am trying to understand why doesn't `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setCapabilityReference(java.lang.String)` imply `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setAllowExpression(false)`? I am failing to see the case where this would be allowable.

It's allowable because of various legacy cases that still support it.

It can actually work if the expression is resolvable via system properties or env vars, which, TBH is what someone who might put an expression in one of these would probably use for resolution. But other mechanisms for resolution, including new stuff we might do some day, won't work. So it's not reliable. Galleon also can't deal with it.

Maybe we can get rid of the remaining ones that allow it. That log message I forgot about has been around for 8 years now!

Blindly changing the behavior of the builder isn't good though. We'd need to identify what attributes that would change so we could update the model versions.

rhusar:
> What a nice log WARN message. Buy whoever wrote that a beer!
> 
> I'd forgotten all about that.

More than happy to buy you a drink next time I see YOU. :-)

rhusar:
> > @bstansberry If you have a minute, could you please check that this is the correct way to disallow expressions that were incorrectly enabled for capability references?
> > Moreover, I am trying to understand why doesn't `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setCapabilityReference(java.lang.String)` imply `org.jboss.as.controller.AbstractAttributeDefinitionBuilder#setAllowExpression(false)`? I am failing to see the case where this would be allowable.
> 
> It's allowable because of various legacy cases that still support it.
> 
> It can actually work if the expression is resolvable via system properties or env vars, which, TBH is what someone who might put an expression in one of these would probably use for resolution. But other mechanisms for resolution, including new stuff we might do some day, won't work. So it's not reliable. Galleon also can't deal with it.
> 
> Maybe we can get rid of the remaining ones that allow it. That log message I forgot about has been around for 8 years now!
> 
> Blindly changing the behavior of the builder isn't good though. We'd need to identify what attributes that would change so we could update the model versions.

OK @bstansberry I am calling the shots then! Filed https://issues.redhat.com/browse/WFLY-21280 but no promises as to when I will prioritize this, so free for taking.

PR #19437, opened 2025-12-06
WFLY-16653 Fix race condition in SuspendBatchletTestCase: JobOperator…

rhusar:
….start() is asynchronous, so wait for the Batchlet to actually begin executing before suspending the server to avoid the race condition that caused "The job execution has no steps" failures

Included suggestions for future improvements of the test logic.

Resolves
https://issues.redhat.com/browse/WFLY-16653


bstansberry (approved):
@jamezp  This LGTM. See discussion starting at https://issues.redhat.com/browse/WFLY-16653?focusedId=27501876&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27501876 why this is the change.

Thanks, @rhusar

jamezp (approved):
This looks reasonable to me.

PR #19441, opened 2025-12-08
[WFLY-21247] Don't import standard boms into preview boms; use common

bstansberry:
This requires some refactoring of how the testsuite controls which boms to use and how ts/int/basic declares some deps that vary between WFP and std WF

https://issues.redhat.com/browse/WFLY-21247

### DIFF
--- a/boms/preview-ee/pom.xml
+++ b/boms/preview-ee/pom.xml
@@ -76,11 +76,11 @@
         <dependencies>
 
             <!--
-                Re-expose the standard-ee deps.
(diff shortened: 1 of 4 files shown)


bstansberry:
#19442 tested this using the EE11 branch to get WFP coverage. I'll close that one as its purpose has been achieved.

jamezp (approved):

darranl (approved):
Adding an approval, the hierarchy of boms confused me so good to get this cleaned up.

PR #19444, opened 2025-12-08
[WFLY-21154] Default Hibernate ORM compatibility setting for UUID to char

scottmarlow:
https://issues.redhat.com/browse/WFLY-21154

Pass EE 11 Persistence tests:
- ee.jakarta.tck.persistence.core.types.field.ClientPmservletTest#testCreateUUIDType
- ee.jakarta.tck.persistence.core.types.field.ClientPuservletTest#testCreateUUIDType

I believe that we will soon pass the other Persistence tests mentioned in https://issues.redhat.com/browse/WFLY-21154 pending  TCK challenge https://github.com/jakartaee/platform-tck/issues/2593 (e.g. needs to be approved or rejected or questioned).

### DIFF
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/HibernatePersistenceProviderAdaptor.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/HibernatePersistenceProviderAdaptor.java
@@ -12,6 +12,7 @@ import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.persistence.SharedCacheMode;
 import jakarta.persistence.spi.PersistenceUnitInfo;
 import org.hibernate.cfg.AvailableSettings;
+import org.hibernate.cfg.MappingSettings;
 import org.jipijapa.cache.spi.Classification;
 import org.jipijapa.event.impl.internal.Notification;
 import org.jipijapa.plugin.spi.EntityManagerFactoryBuilder;
@@ -80,6 +81,9 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
 
         // Enable JPA Compliance mode
         putPropertyIfAbsent(pu, properties, AvailableSettings.JPA_COMPLIANCE, true);
+        // Default to CHAR type for UUID column mapping.
+        // Applications can override by setting "hibernate.type.preferred_uuid_jdbc_type" to "uuid"
+        putPropertyIfAbsent(pu, properties, MappingSettings.PREFERRED_UUID_JDBC_TYPE, "CHAR");
     }
 
     private void failOnIncompatibleSetting(PersistenceUnitMetadata pu, Map properties) {


scottmarlow:
The `org.wildfly.test.integration.microprofile.jwt.propagation.JWTIdentityPropagationTestCase(standalone-enabled-expansion-test).testInvokeEJBWithinSingleDeploymentOutflowAnonymousConfigured` + `org.wildfly.test.integration.microprofile.jwt.propagation.JWTIdentityPropagationTestCase(standalone-enabled-expansion-test).testInvokeEJBWithinSingleDeploymentOutflowNotConfigured` test failures do not look related to this change.

rhusar:
> The `org.wildfly.test.integration.microprofile.jwt.propagation.JWTIdentityPropagationTestCase(standalone-enabled-expansion-test).testInvokeEJBWithinSingleDeploymentOutflowAnonymousConfigured` + `org.wildfly.test.integration.microprofile.jwt.propagation.JWTIdentityPropagationTestCase(standalone-enabled-expansion-test).testInvokeEJBWithinSingleDeploymentOutflowNotConfigured` test failures do not look related to this change.

Right, those are unrelated (I see you have rerun them anyway, all green)

jamezp:
@scottmarlow Should we be changing the default here? Will this match what we're doing for Jakarta EE 10 as well or will this change the default behavior?

rhusar:
> @scottmarlow Should we be changing the default here? Will this match what we're doing for Jakarta EE 10 as well or will this change the default behavior?

I am wondering the same thing, this appears to be a workaround for the TCK that uses a database that cannot use proper UUID but needs to revert to converting this to a string, but changes the whole default in WF? But I am no expert here.

scottmarlow:
> > @scottmarlow Should we be changing the default here? Will this match what we're doing for Jakarta EE 10 as well or will this change the default behavior?
> 
> I am wondering the same thing, this appears to be a workaround for the TCK that uses a database that cannot use proper UUID but needs to revert to converting this to a string, but changes the whole default in WF? But I am no expert here.

These are good questions.  I'm going to look a bit further into this.

jamezp:
We can make changes like this in a deployment descriptor if we need.

scottmarlow:
> We can make changes like this in a deployment descriptor if we need.

For this change I prefer to keep using the out of the box WildFly Preview default configuration.

jamezp:
> > We can make changes like this in a deployment descriptor if we need.
> 
> For this change I prefer to keep using the out of the box WildFly Preview default configuration.

I think we need to understand first if this would be considered backwards incompatible and decide if it's an okay breaking change or not.

scottmarlow:
> > > We can make changes like this in a deployment descriptor if we need.
> > 
> > 
> > For this change I prefer to keep using the out of the box WildFly Preview default configuration.
> 
> I think we need to understand first if this would be considered backwards incompatible and decide if it's an okay breaking change or not.

I have a better fix which is allowed by updating the database DDL script to use native UUID PostGresSQL support:
drop table IF EXISTS UUIDTYPE;
CREATE TABLE UUIDTYPE (ID UUID NOT NULL, NAME VARCHAR(255), PRIMARY KEY (ID));

Previously the ID column was `ID VARCHAR(96) NOT NULL` which Hibernate ORM also supports via the change made by this pull request.  Applications normally choose their preferred settings via hints they add in persistence.xml which override the WildFly default settings.

PR #19453, opened 2025-12-09
[WFLY-21235] Upgrade Mojarra to 4.0.13 and 4.1.5

bstansberry:
https://issues.redhat.com/browse/WFLY-21235


jasondlee (approved):

rhusar (approved):
Excellent, I was going to ask them about the release, but didn't even get to that.

Verified that these contain my fix for the serialization isolation problem.

This will also resolve https://issues.redhat.com/browse/WFLY-21166

rhusar:
Thanks @bstansberry and @jasondlee 

I am merging this as we need more runs to verify this fixes the intermittent failures and we are working with a very limited time here before 39 release 👍🏼

PR #19454, opened 2025-12-09
[WFLY-21218] From the Jakarta WebSocket 2.2 specification section 6.2…

jamezp:
… only mentions scanning for classes with @ServerEndpoint and not classes with @ClientEndpoint. WildFly has been scanning for these endpoints and failing during deployment if an invalid endpoint is found. Instead of a breaking change of no longer scanning for endpoints, we'll do some simple validation and log a warning instead of failing for client endpoints.

https://issues.redhat.com/browse/WFLY-21218

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
@@ -99,6 +99,16 @@ public class UndertowJSRWebSocketDeploymentProcessor implements DeploymentUnitPr
                 for (AnnotationInstance endpoint : clientEndpoints) {
                     if (endpoint.target() instanceof ClassInfo) {
                         ClassInfo clazz = (ClassInfo) endpoint.target();
+                        // Check for a default no-arg constructor as per the Jakarta WebSocket 2.2 specification
+                        // section 6.2. It only mentions scanning for classes with @ServerEndpoint and not classes
+                        // with @ClientEndpoint. However, we've been adding deployed client endpoints since this
+                        // specification was implemented in WildFly. If there is not a no-arg constructor, we'll
+                        // log a warning and not process the type.
+                        if (!clazz.hasNoArgsConstructor()) {
+                            UndertowLogger.ROOT_LOGGER.debugf("The client endpoint '%s' annotated with @ClientEndpoint does not have a no-arg constructor. " +
+                                    "The endpoint will not be available for use without manual instantiation.", clazz.name());
+                            continue;
+                        }
                         try {
                             Class<?> moduleClass = ClassLoadingUtils.loadClass(clazz.name().toString(), module);
                             if (!Modifier.isAbstract(moduleClass.getModifiers())) {


undertow/src/main/java/org/wildfly/extension/undertow/logging/UndertowLogger.java
```diff
@@ -429,4 +430,8 @@ public interface UndertowLogger extends BasicLogger {
     @Message(id = 111, value = "The annotation: '%s' will have no effect on Servlet: '%s'")
     void badAnnotationOnServlet(String annotation, String servlet);
 
+    @LogMessage(level = WARN)
```
bstansberry: As mentioned in Zulip, I think INFO. DEBUG can work too; I don't have a strong opinion. There can't be many of these cases or we'd have been getting reports of failures. So if someone adds one they might not be expecting us to register it like we do with the no-arg ones. Or maybe they will. So ???

NBD anyway as I'm pretty confident there won't be many cases of this getting logged.

undertow/src/main/java/org/wildfly/extension/undertow/logging/UndertowLogger.java
```diff
@@ -429,4 +430,8 @@ public interface UndertowLogger extends BasicLogger {
...
+    @Message(id = 112, value = "The client endpoint '%s' annotated with @ClientEndpoint does not have a no-arg constructor. This endpoint will be ignored.")
```
bstansberry: s/ignored/something else that better describes what won't happen/g

AIUI the class still exists and be used by the app; WF's just not doing something with it. The user reading this will have no idea what the thing logging this was trying to do.

bstansberry (changes requested):

jamezp:
This fix is somewhat for a TCK challenge https://github.com/jakartaee/websocket/issues/578. This is likely to get rejected due to the fact that the specification indicates we are not required to look for `@ClientEndpoint` annotated types. WildFly has scanned for these since Websocket was integrated into WildFly.

IMO the fact the TCK packages the an invalid class is wrong. Regardless of whether we should be scanning for annotated `@ClientEndpoint`'s, the spec packages a class that violates the [specification](https://github.com/jakartaee/websocket/issues/578):

> ## 4.2. @ClientEndpoint
> This class level annotation signifies that the Java class it decorates is to be deployed as a WebSocket client endpoint that will connect to a WebSocket endpoint residing on a WebSocket server. The class must have a public no-args constructor, and additionally may conform to one of the types listed in [Chapter 7](https://jakarta.ee/specifications/websocket/2.2/jakarta-websocket-spec-2.2#jakartaee).

For this reason, I've not added a test case for this.

bstansberry:
I put the 39.x label on this as it fixes a TCK problem. Any TCK fixes we have that seem safe I'd like to get into 39. This seems harmless.

undertow/src/main/java/org/wildfly/extension/undertow/logging/UndertowLogger.java
bstansberry: I'm unresolving this not because I consider it not resolved, but just to make this thread visible in thge GH UI. It's about a main discussion point for this.

The PR now logs at debug which is fine with me.

bstansberry (approved):
This LGTM but a review from @fl4via or @ropalka would be good.

The point about not adding a test because it would involve a class that violates the spec seems reasonable to me. We can always add one later if we see a need.

ropalka (approved):

rhusar:
Thanks @jamezp @bstansberry and @ropalka !

PR #19459, opened 2025-12-10
[WFLY-21188] Fix Hibernate runtime bytecode enhancement not working in EJB JAR module in EAR deployment.  This reverts the WFLY-20393 change that previously ensured that only one bytecode transformer is added for each Persistence Unit.

scottmarlow:
https://issues.redhat.com/browse/WFLY-21188
The previous change we made to reduce the number of threads trying to enhance the same class from multiple threads caused other problems.  Will open a new issue to replace how we solve WFLY-20393.

____

> [!WARNING]

> * [WFLY-20393](https://issues.redhat.com/browse/WFLY-20393)
> * [WFLY-21260](https://issues.redhat.com/browse/WFLY-21260)

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EJBJarPackagingTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EJBJarPackagingTestCase.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.jpa.earbytecodeenhancement;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import javax.naming.InitialContext;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.EnterpriseArchive;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+/**
+ * Test that an EJB jar can reference the persistence provider classes
+ */
+@RunWith(Arquillian.class)
+public class EJBJarPackagingTestCase {
+
+    @Deployment
+    public static Archive<?> deploy() {
+
+        EnterpriseArchive ear = ShrinkWrap.create(EnterpriseArchive.class, "scopedToEar.ear");
+
+        JavaArchive ejbjar = ShrinkWrap.create(JavaArchive.class, "ejbjar.jar");
+        ejbjar.addAsManifestResource(emptyEjbJar(), "ejb-jar.xml");
+        ejbjar.addAsManifestResource(EJBJarPackagingTestCase.class.getPackage(), "persistence.xml", "persistence.xml");
+        ejbjar.addClasses(EmployeeBean.class, EJBJarPackagingTestCase.class);
+        ejbjar.addClasses(Employee.class);
+        ear.addAsModule(ejbjar);        // add ejbjar to root of ear
+
+        JavaArchive lib = ShrinkWrap.create(JavaArchive.class, "lib.jar");
+        lib.addClasses(Organisation.class);
+        ear.addAsLibrary(lib);          // add entity jar to ear/lib
+
+
+
+
+        return ear;
+    }
+
+    @ArquillianResource
+    private static InitialContext iniCtx;
+
+    /**
+     * Test that bean in ejbjar can access persistence provider class
+     */
+    @Test
+    public void testBeanInEJBJarCanAccessPersistenceProviderClass() throws Exception {
+        EmployeeBean bean = (EmployeeBean) iniCtx.lookup("java:app/ejbjar/EmployeeBean");
+        Class sessionClass = bean.getPersistenceProviderClass("org.hibernate.Session");
+        assertNotNull("was able to load 'org.hibernate.Session' class from persistence provider", sessionClass);
+    }
+
+    @Test
+    public void testEntityByteCodeIsEnhanced() throws Exception {
+            EmployeeBean bean = (EmployeeBean) iniCtx.lookup("java:app/ejbjar/EmployeeBean");
+            assertTrue("Employee entity class needs to be bytecode enhanced",
+                    bean.isEmployeeClassByteCodeEnhanced());
+            assertTrue("Organisation entity class (that resides in ear/lib/lib.jar) needs to be bytecode enhanced",
+                    bean.isOrganisationClassByteCodeEnhanced());
+        }
+
+    private static StringAsset emptyEjbJar() {
+        return new StringAsset(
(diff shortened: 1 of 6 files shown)


scottmarlow:
https://issues.redhat.com/browse/WFLY-21260 is the new issue to address WFLY-20393 in a better way.

scottmarlow:
We don't have to merge https://github.com/wildfly/wildfly/pull/19459 just yet as when we fix https://issues.redhat.com/browse/WFLY-21260 I think that will also address the regression fixed by this pull request.

If for some reason we don't fix WFLY-21260 for WF 39. then we could merge wf/pull/19459 for 39 final and then revert that as part of the WFLY-21260 fix.

jamezp:
Does the issue seem to be that it's not enhancing EJB JAR's because the same PU is used in a WAR? I'm trying to understand how running it once per PU doesn't work, but a "run it multiple times" approach does.

FWIW I do think it's okay to do this, but if we do we should have a followup issue to look at it again.

scottmarlow:
Firstly the optimization was wrong because we need more than one instance of the bytecode transformer to be added per persistence unit for the EAR case.  The EAR case relevant for this fix is that the EAR top level deployment needs a transformer and the EJB subdeployment also needs a transformer.

> Does the issue seem to be that it's not enhancing EJB JAR's because the same PU is used in a WAR? I'm trying to understand how running it once per PU doesn't work, but a "run it multiple times" approach does.

A WAR subdeployment in an EAR case would be the same problem but this with an EJB.  Basically, the bug was that the EAR deployment unit processor is run first and then the subdeployment but regardless of the order the optimization we made was wrong because both top level deployment + subdeployments need the transformers added.  

> 
> FWIW I do think it's okay to do this, but if we do we should have a followup issue to look at it again.

+1

I do want to add a new WFLY issue to do the optimization completely different in a way that is driven by EE 12/Persistence 4.0 requirements.  We can do that in an EE 10/11/12 compatible way as we already started that previously, just need more.  However, we should identify if we want to only solve this for EE 12 or to also solve it for EE 10/11 I think.

scottmarlow:
> FWIW I do think it's okay to do this, but if we do we should have a followup issue to look at it again.

https://issues.redhat.com/browse/WFLY-21270 is a followup issue to look at again.

jamezp:
The test failures seem related

scottmarlow:
> The test failures seem related

I fixed the failures locally.  Moved the new test to a separate test package to address failures in:

> org.jboss.as.test.integration.jpa.packaging.PersistenceUnitPackagingTestCase(basic-integration-default-web)
> New
> testUserOfOveriddenSubDeploymentUsingExplicitPath
> New
> testUsingEarLibPuInSubdeployment
>

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EJBJarPackagingTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+    @ArquillianResource
+    private static InitialContext iniCtx;
+
+    /**
+     * Test that bean in ejbjar can access persistence provider class
+     */
+    @Test
+    public void testBeanInEJBJarCanAccessPersistenceProviderClass() throws Exception {
+        EmployeeBean bean = (EmployeeBean) iniCtx.lookup("java:app/ejbjar/EmployeeBean");
+        Class sessionClass = bean.getPersistenceProviderClass("org.hibernate.Session");
```
jamezp: This should be:
```java
Class<?> sessionClass = bean.getPersistenceProviderClass("org.hibernate.Session");
```

If that can't work, then we need to suppress the compiler warning with a comment IMO.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
```diff
@@ -0,0 +1,67 @@
...
+    public EntityManagerFactory getEntityManagerFactory() {
+        return entityManagerFactory;
+    }
+
+    public EntityManagerFactory getDefaultEntityManagerFactory() {
+        return defaultEntityManagerFactory;
+    }
+
+    // AS7-2829 bean in ejbjar should be able to access class in persistence provider
+    public Class getPersistenceProviderClass(String classname) {
```
jamezp: We should avoid the raw type usage here and return `Class<?>`,

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EJBJarPackagingTestCase.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.jpa.earbytecodeenhancement;
+
+import static org.junit.Assert.assertNotNull;
```
jamezp: This will fail the checkstyle as it's an unused import.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
```diff
@@ -0,0 +1,67 @@
...
+        Class result = null;
```
jamezp: Same note here.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
jamezp: Actually, maybe this method is never used and should just be deleted.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+@Stateless
+public class EmployeeBean {
+
+    @PersistenceUnit(unitName = "mainPu")
+    EntityManagerFactory entityManagerFactory;
+
+    // AS7-2275 requires each PU reference to specify a persistence unit name, if there are
+    // multiple persistence unit definitions.
+    // as a workaround, specified the pu name
```
jamezp: Are these comments still relevant to this or should we simply just be reusing the other types?

scottmarlow: No, the comments are not relevant.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EJBJarPackagingTestCase.java
scottmarlow: The test doesn't really need to use the old school lookup, I'll inject the bean instead.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
scottmarlow: It is now deleted.  Thanks!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/earbytecodeenhancement/EmployeeBean.java
scottmarlow: deleted that also...

jamezp (approved):

PR #19464, opened 2025-12-11
[WFLY-21210] Fix broken docs.jboss.org/ironjacamar links in documenta…

tadamski:
…tion

https://issues.redhat.com/browse/WFLY-21210

### DIFF
--- a/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
+++ b/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
@@ -67,7 +67,7 @@ A resource adapter archive can be automatically activated with a
 configuration by including an META-INF/ironjacamar.xml in the archive.
 
 The schema can be found at
-http://docs.jboss.org/ironjacamar/schema/ironjacamar_1_0.xsd
+https://www.ironjacamar.org/doc/schema/ironjacamar_1_2.xsd
 
 [[ra-component-reference]]
 == Component Reference
@@ -80,4 +80,4 @@ the project documentation.
 * IronJacamar homepage: http://www.jboss.org/ironjacamar
 * Project Documentation: http://www.jboss.org/ironjacamar/docs
 * Schema description:
-http://docs.jboss.org/ironjacamar/userguide/1.0/en-US/html/deployment.html#deployingra_descriptor
+https://www.ironjacamar.org/doc/userguide/1.2/en-US/html/ch05.html#deployingra_descriptor
--- a/docs/src/main/asciidoc/Getting_Started_Guide.adoc
+++ b/docs/src/main/asciidoc/Getting_Started_Guide.adoc
@@ -727,7 +727,7 @@ the project documentation.
 * IronJacamar homepage: http://www.jboss.org/ironjacamar
 * Project Documentation: http://www.jboss.org/ironjacamar/docs
 * Schema description:
-http://docs.jboss.org/ironjacamar/userguide/1.0/en-US/html/deployment.html#deployingds_descriptor
+https://www.ironjacamar.org/doc/userguide/1.2/en-US/html/ch05.html#deployingds_descriptor
 
 [[configure-logging-in-wildfly]]
 === Configure Logging in WildFly


docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
```diff
@@ -67,7 +67,7 @@ A resource adapter archive can be automatically activated with a
 configuration by including an META-INF/ironjacamar.xml in the archive.
 
 The schema can be found at
-http://docs.jboss.org/ironjacamar/schema/ironjacamar_1_0.xsd
+https://www.ironjacamar.org/doc/userguide/1.2/en-US/html/apa.html#ironjacamar12
```
rhusar: I believe the intention for this link is the XSD schema that can be used by tooling to validate XML. Having it written out is not overly helpful. Can you link the schema directly?

rhusar (changes requested):

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
tadamski: fixed

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
```diff
@@ -67,7 +67,7 @@ A resource adapter archive can be automatically activated with a
 configuration by including an META-INF/ironjacamar.xml in the archive.
 
 The schema can be found at
-http://docs.jboss.org/ironjacamar/schema/ironjacamar_1_0.xsd
+https://www.ironjacamar.org/doc/schema/datasources_1_3.xsd
```
rhusar: IIUC the schema you linked does not contain `element name="datasources"` as the XML example above, so that's not the right XSD for this?

rhusar (changes requested):

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
tadamski: true, this was a mistake, now it should be fixed

bstansberry (dismissed):

docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc
bstansberry: I don't understand this thread. :)

What XML example above that uses 'datasources'? 

https://docs.wildfly.org/38/Admin_Guide.html#Resource_adapters

And I see 'element name="datasources"' in https://www.ironjacamar.org/doc/schema/datasources_1_3.xsd

NBD but it seems now the link is to https://www.ironjacamar.org/doc/schema/datasources_1_2.xsd which presumably is not the best one to use if 1_3 exists.

rhusar: From docs:

> A resource adapter archive can be automatically activated with a configuration by including an META-INF/ironjacamar.xml in the archive.

I dono, basically my question is "which is the correct XSD for the META-INF/ironjacamar.xml file because it doesn't seem to be datasources_1_3.xsd?" (but maybe it is?)

rhusar: Looking at existing ironjacamar.xml files we have in our testsuite for reference, and they are a mess: they do not define a schema.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/archive/ironjacamar.xml
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/beanvalidation/ironjacamar.xml
testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jca/security/ironjacamar.xml

Can @tadamski fix that at some point too?

rhusar: > I don't understand this thread. :)
> 
> What XML example above that uses 'datasources'?
> 
> https://docs.wildfly.org/38/Admin_Guide.html#Resource_adapters
> 
> And I see 'element name="datasources"' in https://www.ironjacamar.org/doc/schema/datasources_1_3.xsd
> 
> NBD but it seems now the link is to https://www.ironjacamar.org/doc/schema/datasources_1_2.xsd which presumably is not the best one to use if 1_3 exists.

@bstansberry  Typo - I meant 'resource-adapter' in the XML above. The linked XSD contains only datasources, not resource adapters. 

https://github.com/tadamski/wildfly/blob/693f838430608749d4ca7a4eb0bf2dbf36ebbb57/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/Resource_adapters.adoc?plain=1#L26

tadamski: @bstansberry @rhusar tbh my only intention was to refer to a working version of a link that is already present in a docs. This issue reminded me that I finally need to do something with legacy IJ page + now I see that schemas require refactor as well. Maybe lets put this PR on hold - I will figure out those two issues and return to it.

rhusar: @tadamski @bstansberry When looking at the existing ironjacamar.xml in our testsuite, to figure out which schema is actually correct, I noticed they do NOT even define a schema!!! How anyone creates an XML without defining a namespace is an enigma to me... but let's assume these are from early days before schemas were published.

Opened https://issues.redhat.com/browse/WFLY-21341 for that. 

Of course defining the schema uncovers that the XML files do violate that schema, so that needs fixing too.

rhusar: Submitted a fix for these: https://github.com/wildfly/wildfly/pull/19514

But it gets worse: the https://www.ironjacamar.org/doc/schema/ironjacamar_1_3.xsd is not even published! Opened https://issues.redhat.com/browse/WFLY-21342

PR #19469, opened 2025-12-12
[WFLY-21267] Create a transaction scoped StatelessSession; use a CDI Extension to …

bstansberry:
…replace the session in Hibernate-generated Repository impls

https://issues.redhat.com/browse/WFLY-21267


jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernateStatelessSessionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+@MetaInfServices(StatelessSessionFactory.class)
+public class HibernateStatelessSessionFactory implements StatelessSessionFactory<StatelessSession> {
+
+    @Override
+    public StatelessSession getTransactionScopedSession(Function<Function<EntityManagerFactory, AutoCloseable>, ScopedStatelessSessionSupplier> supplierFactory) {
+        ScopedStatelessSessionSupplier sessionSupplier = supplierFactory.apply(getSessionFactory());
+        return new TransactionScopedStatelessSession(sessionSupplier);
+    }
+    private Function<EntityManagerFactory, AutoCloseable> getSessionFactory() {
+        return entityManagerFactory -> entityManagerFactory.unwrap(SessionFactory.class).openStatelessSession();
```
scottmarlow: IMO this call to openStatelessSession could be considered similar to calling entityManagerFactory.createEntityManager() which returns an application managed PersistenceContext that the application closes.  I have yet to understand if any `EE container` management is needed for the returned StatelessSession.

ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
```diff
@@ -26,6 +26,7 @@ public static final class Component {
         public static final int ENTITY_BEAN_REMOVE_INTERCEPTOR = 0x550;
         public static final int JPA_SFSB_INTERCEPTOR = 0x560;
         public static final int JPA_SESSION_BEAN_INTERCEPTOR = 0x600;
+        public static final int JAKARTA_DATA_SESSION_BEAN_INTERCEPTOR = 0x600;
```
rhusar: I believe the interceptor order is supposed to be deterministic and having 2 interceptors with the same number is not. Unless this is intentional?

rhusar (changes requested):

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernateStatelessSessionFactory.java
bstansberry: @scottmarlow At this stage the only use of this is the field Hibernate Data Repositories generates. There they do call close() on the StatelessSession (in a @PreDestroy method they generate). So the 'application author' doesn't close the session but internally 'we' do. (This 'us' vs 'them' topic is complex with Jakarta Data since 'we' are generating application code.) But anyway, right now this is very tailored to the single use case that can use it.

This PR results in that 'close()' call being ignored. I think that's ok for the current specific use of this, which is quite tied to the specific classes ORM generates. Right now TransactionScopedStatelessSession just ignores the close call and relies on transaction completion or NoTxEmCloser to close the underlying Hibernate StatelessSessionImpl instances.

TransactionScopedEntityManager throws an ISE if close() is called. AIUI that is spec-defined behavior. My very-much noob sense is it would be better behavior for this as well. But doing that would break the exact ORM-generated code that this is meant to work with. In the future when we have more time to coordinate with the ORM folks to get generated code that does what we all agree is best, then we can update. We might even need to add a field to TransactionScopedStatelessSession to tell it what to do, so if we are creating it via one call path (say to handle code generated by a newer ORM release) we configure the TSSS to throw ISE. For call paths dealing with code from an older ORM (like this) we configure the TSSS to ignore the failure.  Over time the need to deal with the older code will disappear. For example, assuming ORM 6.6.x never changes what it writes, our need to deal with what it writes will disappear when we stop supporting ORM 6.6-backed Jakarta Data. (Note: I don't see EAP every supporting that; EAP support of Jakarta Data should be limited to EE 11 / ORM 7.)

Now, why doesn't this PR just close the underlying StatelessSessionImpls when TSSS.close() is called? In theory it could but:

1) It's harder to implement. The underlying StatelessSessionImpls (and SessionImpls in the EM case) are stored in ThreadLocals. To implement close(), TSSS would need to have a mechanism to track all the created StatelessSessionImpls and close them. (As I write I'm thinking how to do this and it might be possible if we wanted, although probably not something I could do today.)

2) We'd need to think through all the impacts of closing StatelessSessionImpls in the repository impl's @PreDestroy method. Particularly the impact to any in-flight transactions. Things like that. Perhaps it would be OK, but what's here right now instead is like what happens with EM. The app code can't close the EM. We understand very well the semantics of that. I think we should think harder before doing something different for StatelessSession.

I'm about to link the JIRA for this, and fix the SecurityManager problem. But after that I will file another JIRA to track future developments. Resolving that can be where we nail down the semantics we want for use cases other than the current very narrow one.

ee/src/main/java/org/jboss/as/ee/component/interceptors/InterceptorOrder.java
bstansberry: No, that was a copy-past mistake; I intended to change the number. Thanks @rhusar

bstansberry: Actually this is cruft; an earlier iteration used a separate interceptor but I reworked this to use the same one.

rhusar: Gotcha! Lets remove then.

bstansberry: Done! Thanks!

bstansberry:
@darranl @scottmarlow @rhusar  I've moved this out of Draft state.

I've updated it from what it was earlier by:

1) Adding lots of comments/javadoc.
2) Removing the cruft line Rado pointed out.
3) Adding a doPrivileged in JakartaDataDeploymentProcessor to fix the SecurityManager CI job failure.
4) In the two HibernateStatelessSessionFactory classes I replaced a lamba with a class that "implements Function, Serializable" as the function instance is stored in a field that may need to be serialized.

95%+ of the change was the added comments/javadoc.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernateStatelessSessionFactory.java
scottmarlow: The overall change looks really good, thank you!  I do not see any `logic` problems.  I really like the idea of having a managed stateless session!  Thanks for the response about apps closing the StatelessSession versus the container!

> TransactionScopedEntityManager throws an ISE if close() is called. AIUI that is spec-defined behavior. 

Yes, that is spec-defined behavior!  

Here is a funny question, if a Stateless Session Bean has a StatelessSession reference, I wonder if the StatelessSession should be serialized when the Stateless Session Bean is serialized (for clustering + passivation/activation)?  On the surface,  the StatelessSession has no state but internally it might have some state.  I just checked and the org.hibernate.internal.StatelessSessionImpl class extends AbstractSharedSessionContract which indeed has serialization handling via read/write.  This all works because the WildFly persistence unit fully scoped name should still be used as the key.  I suppose we could update our clustering tests to also cover this I guess without an extended persistence context but I'm not really sure if it is worth adding that to our CI for repeated testing if there is no state to assert being available in different cluster nodes (probably ever).  I guess we would find out if a minimal StatelessSession impacts clustering though, so not a bad test actually.

scottmarlow (approved):

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernateStatelessSessionFactory.java
scottmarlow: Thinking more out loud, I wonder if the Jakarta EE Persistence container could help eliminate the need for the ORM internal StatelessSession to have state by delegating to an api/spi that deals with database connections + current threads transaction state kind of thing.

bstansberry: Re the SFSB or SessionCase case, if we added a feature where the app code had a direct ref to one of these and then we serialized/deserialized the bean, all the JDBC stuff would be left behind. TSSS is a StatelessSession impl whose only internal state is what's needed to reestablish access to an EntityManagerFactory so it can create/manage StatelessSessionImpl instances. It doesn't itself hold any refs to those StatelessSessionImpl instances.

I *think* that's actually good, as naively 'StatelessSession' is meant to be 'stateless' so if you serialize/deserialize one you'd not expect a bunch of state to come along.

Right now the only ref to a TSSS is going to be an internal detail of a repository implementation Hibernate Data Repositories generates. There's some discussion of that in [#wildfly-developers > WFLY-21267](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/WFLY-21267/with/563566879)

A repository impl using what's in main now fails to serialize because the StatelessSessionImpl the repo has a ref to has non-serializable state. With this PR you can actually serialize. Deserializing failed because i was trying to do it in the app code, and that code doesn't have access to the TSSS class. Maybe it could work if we were doing the deserialization in a more controlled environment (controlled by the container) where the needed classes were in place. I don't know how that kind of thing works with deserializing aTransactionScopedEntityManager.

darranl:
@rhusar you are still shoing as changes requested, I think this related to things already addressed by Brian - did you want to perform another check or just go with Scott's review?

rhusar (approved):
@darranl To the extend I can help review this, it looks good to me. 👍🏼 

Nevertheless, I can't avoid mentioning there is no test coverage added to this change and I the temporary nature of this change is not thoroughly documented – but I believe the intention is to address this eventually, so this is fine.

bstansberry:
+1 @rhusar regarding tests. I'm trying to figure out how to test the bug this fixes. The StatelessSession objects themselves are hidden internal details of the repository impls. And, they are largely stateless. So how to have two clients access the session and produce testable effects is not obvious! :) 

It passes the WF Jakarta Data tests and the Jakarta Data TCK, but that's evidence that it didn't break something.

rhusar:
Should be enough validation we needed 👍🏼

bstansberry:
The follow-up JIRA for this is a Feature Request: https://issues.redhat.com/browse/WFLY-21272

That would be much more testable as we could inject StatelessSession into any of the many bean types that are easily accessible in tests, e.g. Jakarta REST resources, so we can directly access the session instance.

PR #19470, opened 2025-12-12
WFLY-21158 Model descriptions for expiration parameters are out-of-date.

pferraro:
https://issues.redhat.com/browse/WFLY-21158

Also, remove out-dated documentation WRT max-idle, which has been safe for use with clustered caches since Infinispan 13.


rhusar (approved):
👍🏼

PR #19478, opened 2025-12-15
WFLY-21279 IdleThresholdTimerPassivationTestCase fails intermittently

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-21279

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/ejb/timer/passivation/bean/TimerTrackerBean.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/ejb/timer/passivation/bean/TimerTrackerBean.java
@@ -5,18 +5,20 @@
 package org.jboss.as.test.clustering.single.ejb.timer.passivation.bean;
 
 import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.stream.Stream;
 
-import org.jboss.as.test.clustering.PassivationEventTrackerBean;
-
+import jakarta.annotation.PreDestroy;
 import jakarta.annotation.Resource;
 import jakarta.ejb.Remote;
-import jakarta.ejb.Remove;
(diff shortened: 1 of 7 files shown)


rhusar:
Reword as per Paul's feedback: removed any calls to TimerService.getTimers().

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/ejb/timer/passivation/bean/TimerTrackerBean.java
```diff
@@ -32,41 +35,36 @@ public class TimerTrackerBean extends PassivationEventTrackerBean implements Tim
...
-    public void remove() {
-        System.out.println("Called @Remove");
+    @PreDestroy
+    public void preDestroy() {
+        System.out.println("Called @PreDestroy - removing created timers");
+
         // Cancel any remaining timers before removal
-        this.cancelAllTimers();
+        Collection<Timer> collection = new LinkedHashSet<>();
+        timerQueue.drainTo(collection);
```
pferraro: Why create allocate and populate a hash table if only to collect and iterate?
Why not just iterate over the queue directly?

pferraro (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/single/ejb/timer/passivation/bean/TimerTrackerBean.java
rhusar: Good question – because drainTo() is the only method in this API that achieves a transfer of multiple elements from a queue to another collection in a single call. Nevertheless, you are right of course right about the inefficiency.

Anyway, since I don't want to use the idiomatic for loop, it's a great opportunity to use Stream#generate to maintain the destructive properties of the iteration.

pferraro (approved):

rhusar:

PR #19491, opened 2025-12-17
WFLY-21293 Session.requestDone(...) invoked outside of ControlPoint boundary

pferraro:
https://issues.redhat.com/browse/WFLY-21293

Should resolve the vast majority of intermittent distributable web test failures in clustering testsuite.


undertow/src/main/java/org/wildfly/extension/undertow/deployment/SuspendedServerRequestListener.java
```diff
@@ -35,18 +44,29 @@ public SuspendedServerRequestListener(ControlPoint entryPoint) {
...
+        HttpServerExchange exchange = context.getExchange();
         if (exchange.removeAttachment(SuspendedServerHandlerWrapper.RUN_RESULT_KEY) == RunResult.RUN) {
-            // Defer request completion until exchange completion, if necessary.
-            if (exchange.isComplete()) {
-                this.entryPoint.requestComplete();
-            } else {
-                exchange.addExchangeCompleteListener(this);
+            // If a session is associated with the request context, defer ControlPoint completion until Session#requestDone(...)
+            // See: io.undertow.servlet.spec.HttpServletResponseImpl#responseDone()
+            if (context.getSession() == null) {
```
bstansberry: Should this be != null?

bstansberry: ok, ic. it's a little confusing because the comment describes what happens in an 'else' case but precedes the if case

pferraro: I'll clarify the comments.

pferraro:
Let me convert to a draft while I sort out the corner cases that are clearly not being handled.

pferraro:
N.B. I don't like this idea of deferring ControlPoint.requestDone() to the Session.requestDone(...).
I'm exploring an alternate solution that effectively moves Session.requestDone(...) handling to a ExchangeCompletionListener.

pferraro:
OK - this should be ready for review now.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/ServletRequestCompletionListener.java
```diff
@@ -0,0 +1,377 @@
...
+
+        @Override
+        public Session getSession(HttpServerExchange exchange, SessionConfig config) {
+            Session session = !exchange.isComplete() ? this.manager.getSession(exchange, config) : null;
+            return (session != null) ? this.decorate(exchange, session) : null;
+        }
+
+        private Session decorate(HttpServerExchange exchange, Session session) {
+            AtomicReference<Session> reference = new AtomicReference<>(session);
+            ExchangeCompletedSession completed = new ExchangeCompletedSession(this, session.getId());
```
bstansberry: Suggestion:

Can this go inside the exchangeEvent callback? My impression is this line is on the main request handling hot path while the callback is after the exchange is completed.

This is not a huge deal but this PR is adding a fair amount to the critical path.

pferraro: We cannot.
We are effectively overriding the behavior of undertow-servlet so that it does not invoke Session.requestDone(...) outside of scope of the available listeners.
Ultimately, this is really a temporary workaround, and should be removed once https://issues.redhat.com/browse/UNDERTOW-2673 is addressed and available to WildFly.

bstansberry: To be clear, I'm referring to this one line that constructs an object that is never used other than at L135 of the callback. The session.getId() call might need to happen outside the callback if that call might fail inside, but the ExchangeCompletedSession doesn't need to be instantiated here.

But, NBD.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/ServletRequestCompletionListener.java
```diff
@@ -0,0 +1,377 @@
...
+    @Override
+    public DeploymentInfo apply(DeploymentInfo deployment) {
+        // Registered first
+        deployment.addListener(new ListenerInfo(ServletRequestListener.class, new ImmediateInstanceFactory<>(this)));
+        SessionManagerFactory factory = deployment.getSessionManagerFactory();
+        if (factory != null) {
+            // Decorate sessions so that Session.requestDone(...) is called when the exchange completes
+            // This effectively overrides logic within {@link io.undertow.servlet.spec.HttpServletResponseImpl#responseDone()}
+            // and ensures that Session lifecycle does extend beyond request boundary, as determined by the ControlPoint.
+            deployment.setSessionManagerFactory(new HttpServerExchangeSessionManagerFactory(factory));
```
rhusar, line 55: As we discussed in December, wrapping the session manager factory is going to break the fragile logic in mod_cluster subsystem in org.wildfly.mod_cluster.undertow.UndertowContext#isDistributable, which looks like as follows to determine status of distributability:

```
    @Override
    public boolean isDistributable() {
        return !(this.deployment.getDeploymentInfo().getSessionManagerFactory() instanceof InMemorySessionManagerFactory);
    }
```

If I am reading this right, this will make all contexts be handled as distributable. We need to find a better way to implement this method, ideas?

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/deployment/ServletRequestCompletionListener.java
pferraro: Funny - I have no recollection of that conversation :)

I'll add another commit to this PR to fix the above.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/ServletRequestCompletionListener.java
pferraro: Oh, I see what you mean now.  Yes, this object can be created within the listener.

fl4via (approved):

bstansberry:
Thanks @pferraro @rhusar and @fl4via

PR #19492, opened 2025-12-17
WFLY-21294 - Add test for multiple metrics systems

jasondlee:
https://issues.redhat.com/browse/WFLY-21294

Add test
Modify visibility of values in shared utilities for reuse

### DIFF
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.test.integration.observability;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.SUBSYSTEM;
+
+import java.io.IOException;
+import java.nio.file.Files;
+
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.junit.InSequence;
+import org.jboss.as.arquillian.api.ContainerResource;
+import org.jboss.as.arquillian.container.ManagementClient;
+import org.jboss.as.controller.client.Operation;
+import org.jboss.as.controller.client.helpers.Operations;
+import org.jboss.as.test.shared.ServerReload;
+import org.jboss.as.test.shared.observability.setuptasks.MicrometerSetupTask;
+import org.jboss.as.test.shared.observability.setuptasks.OpenTelemetrySetupTask;
+import org.jboss.as.test.shared.util.LoggingUtil;
+import org.jboss.dmr.ModelNode;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * This test examines the logging produced when multiple metrics systems are enabled. The first subsystem will log nothing,
+ * but each subsequent system should log an INFO message alerting the user of the duplication. The number of messages logged
+ * is the number of metrics subsystems enabled minus 1.
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class MultipleMetricsTestCase {
+    private static final String MESSAGE_PREAMBLE = "Additional metrics systems discovered";
+
+    private static final ModelNode ADDRESS_ROOT_LOGGER = Operations.createAddress("subsystem", "logging", "root-logger", "ROOT");
+
+    private static final ModelNode ADDRESS_METRICS_EXTENSION = Operations.createAddress("extension", "org.wildfly.extension.metrics");
+
+    private static final ModelNode ADDRESS_MICROMETER_EXTENSION = MicrometerSetupTask.MICROMETER_EXTENSION;
+    private static final ModelNode ADDRESS_MICROMETER_SUBSYSTEM = MicrometerSetupTask.MICROMETER_SUBSYSTEM;
+
+    private static final ModelNode ADDRESS_OPENTELEMETRY_EXTENSION = OpenTelemetrySetupTask.OPENTELEMETRY_EXTENSION;
+    private static final ModelNode ADDRESS_OPENTELEMETRY_SUBSYSTEM = OpenTelemetrySetupTask.OPENTELEMETRY_ADDRESS;
+
+    private static final ModelNode ADDRESS_MPTELEMETRY_EXTENSION = Operations.createAddress("extension", "org.wildfly.extension.microprofile.telemetry");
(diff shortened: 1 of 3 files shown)


testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
jamezp, line 1: This file is missing the copyright header.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
```diff
@@ -0,0 +1,225 @@
...
+            Assert.assertEquals(expected, lines.filter(line -> line.contains(MESSAGE_PREAMBLE)).count());
+        }
+    }
+
+    private void reloadServer() {
+        ServerReload.executeReloadAndWaitForCompletion(managementClient);
+    }
+
+    private void addLogHandler(String loggerName) {
+        System.out.println("***** [blargh] Adding logger: " + loggerName);
```
jamezp: I'm sure this is left over, but we should remove this before merging :)

jamezp (commented):
The failed tests seem to be because of the type of server provisioned. I didn't look much more than that :)

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
jasondlee: I thought I got 'em all. :p

jasondlee:
> The failed tests seem to be because of the type of server provisioned. I didn't look much more than that :)

I kind of expected these types of failures. Not sure I like the if-this-then-that number approach, but I'm not sure there's a better way. I guess I could just make sure that _a_ log message (or none!) is present. I'll have to think that through and see if we lose anything from checking the count...

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
```diff
@@ -0,0 +1,237 @@
...
+        ModelNode file = new ModelNode();
+        file.get("relative-to").set("jboss.server.log.dir");
+        file.get("path").set(loggerName + ".log");
+        addTestLogOp.get("file").set(file);
+        addTestLogOp.get("autoflush").set("true");
+        addTestLogOp.get("suffix").set(".yyyy-MM-dd");
+        addTestLogOp.get("formatter").set("%-5p [%c] (%t) %s%e%n");
+        builder.addStep(addTestLogOp);
+
+        executeOp(managementClient, builder.build());
```
jamezp, line 163: Minor, but if you're using a composite builder like this, you should just execute the op at the end to do it in one call.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
```diff
@@ -0,0 +1,237 @@
...
+    private ManagementClient managementClient;
+
+    // WildFly Metrics is not enabled for bootable jars, so that changes the number of messages logged.
+    private static boolean wildflyMetricsAvailable = true;
+
+    @Test
+    @InSequence
+    public void setup() throws Exception {
+        wildflyMetricsAvailable = Operations.isSuccessfulOutcome(executeRead(managementClient, ADDRESS_METRICS_EXTENSION));
+    }
```
jamezp, line 61: Could this not be done in a `ServerSetupTask`? If not that, a `@BeforeClass`?

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
```diff
@@ -0,0 +1,237 @@
...
+            removeLogHandler(loggerName);
+        }
+    }
+
+    @Test
+    @InSequence(Integer.MAX_VALUE)
+    public void shutdown() {
+        disableOpenTelemetry();
+        disableMicrometer();
+    }
```
jamezp, line 132: Same note here about a `ServerSetupTask` and/or `@AfterClass`.

jamezp (commented):
Note we don't need to block merging based on these comments.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
jasondlee: I don't know that a ServerSetupTask makes sense, as it's only reading something to determine if it exists, and I don't think `@BeforeClass` would work either, as the `managementClient` needs to be injected, which I don't think will happen before then?

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
rhusar: Next time just do "git grep blargh"? :-D

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
rhusar: You can't really share any context from ServerSetupTask to the test class, so I don't think that's possible.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
jasondlee: You scared me. I thought I pushed up another one. :P

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
jamezp: FWIW in a client test you should be able to with static variables like this, but it's not critical.

jamezp (approved):

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
rhusar: Ah, so a static field in the ServerSetupTask accessed from the test class.. true, possible, but kinda yuck.

bstansberry (approved):
LGTM.

No need to do anything re this, but reviewing this led me to file https://issues.redhat.com/browse/WFLY-21333.

rhusar (approved):
LGTM

bstansberry:
Thanks @jasondlee, @jamezp and @rhusar

PR #19496, opened 2025-12-19
WFLY-21304 AppClientScriptTestCase fails if environment defines _JAVA_OPTIONS environment variable

pferraro:
Revision of original WFLY-14488 fix, which otherwise neglected the "NOTE: " prefix (as well as ANSI color escape sequences).

https://issues.redhat.com/browse/WFLY-21304


bstansberry (dismissed):

pferraro:
@bstansberry Ah - I had only tested this with JDK25.  In that case, this "additionalLines" stuff can just go away.

bstansberry (approved):

bstansberry:
Thanks @pferraro

PR #19497, opened 2025-12-19
[WFLY-20864] Prefer using CDI, if available, for session beans instea…

jamezp:
…d of JNDI lookups. If CDI is not available, allow more than one component to be bound. The latter is uncommon, but should be allowed.

https://issues.redhat.com/browse/WFLY-20864


jamezp:
I'm closing this for now as it will require a RESTEasy upgrade.

PR #19499, opened 2025-12-22
[WFLY-20301] removes plugin management in Tools User BOMs

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-20301

### DIFF
--- a/boms/user/pom.xml
+++ b/boms/user/pom.xml
@@ -58,14 +58,6 @@
                         </includeRepositories>
                     </configuration>
                 </plugin>
-                <!-- The WildFly plugin deploys your war to a local WildFly container -->
-                <!-- To use, set the JBOSS_HOME environment variable and
-                    run: mvn package wildfly:deploy -->
-                <plugin>
-                    <groupId>org.wildfly.plugins</groupId>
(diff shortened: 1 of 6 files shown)


rhusar (changes requested):
Is there any particular reason why these remaining bom builds are not updated with the removal?

If not, these need to be updated too

```
boms/user/server/expansion/preview/with-tools/pom.xml:                                <includePlugin>org.wildfly.plugins:wildfly-maven-plugin</includePlugin>
boms/user/server/expansion/standard/with-tools/pom.xml:                                <includePlugin>org.wildfly.plugins:wildfly-maven-plugin</includePlugin>
```

rhusar:
@emmartins Looks better now. Do you want this for 39.x? If so, stick a label here.

Given that this can be considered a bit of a bug fix, since we give false impression to users that we are managing plugins for them which BOMs simply cannot do in Maven 3.x and nobody is using this as a parent, this is well mergeable into 39.

rhusar (approved):
LGTM now, issues addressed.

I also agree to have this for 39.x

rhusar:
Thanks @emmartins !

PR #19502, opened 2025-12-23
WFLY-21312 Upgrade org.wildfly.glow:wildfly-glow-core from 1.5.1.Final to 1.5.2.Final

rhusar:
https://issues.redhat.com/browse/WFLY-21312


rhusar:
Stumbled upon this when looking at component versions for 39 Final, do we want this @jfdenise ?

jfdenise (approved):

jfdenise:
@rhusar , that is not strictly required. We could have it, it affects tests.

rhusar:
@jfdenise Better sooner than later and less confusion. Lets have it then!

rhusar:
Thanks @jfdenise !

PR #19507, opened 2026-01-06
[WFLY-21326] Applying workaround to resolve JDK26 regression

ropalka:
https://issues.redhat.com/browse/WFLY-21326

### DIFF
--- a/testsuite/integration/manualmode/pom.xml
+++ b/testsuite/integration/manualmode/pom.xml
@@ -597,6 +597,7 @@
                         <oidc.jboss.home>${oidc.jboss.home}</oidc.jboss.home>
                         <ee.dist.output.dir>ee-dist/target/${server.output.dir.prefix}-${server.output.dir.version}</ee.dist.output.dir>
                         <build.output.dir>${wildfly.build.output.dir}</build.output.dir>
+                        <jdk.tls.SunX509KeyManager.certChecking>false</jdk.tls.SunX509KeyManager.certChecking>
                     </systemPropertyVariables>
                 </configuration>
             </plugin>


rhusar:
@ropalka Can you link the custom runs with JDK 26 to verify this as they are not yet run as part of regular CI?

bstansberry:
@ropalka @rhusar https://ci.wildfly.org/viewLog.html?buildId=539680 runs the new SE 26 CI job against the branch, with just -Dts.manualmode instead of -DallTests.

Note that that job in its normal run currently results in 595 failures, with many due to what looks like byte-buddy incompatibility. I don't know how many of those are in ts/int/manualmode. I also don't know if the tests that this is meant to fix are affected by that.

ropalka:
Thanks @bstansberry for scheduling the run! I see just 1 failure (JakartaDataEarTestCase) affected by that bytebuddy incompatibility issue in that run. Anyway SSLEJBRemoteClientTestCase & SSLEJBRemoteRemotingClientTestCase are passing - confirming this PR worked around the WFLY-21326 issue.

rhusar:
You need bytebuddy 1.17.8 or newer AND pass -Dnet.bytebuddy.experimental=true to support JDK 26.

rhusar:
Anyway, if this PR is acceptable as part of the complete JDK 26 still awaiting approval from @darranl

darranl:
What is the error being encountered and how does setting this property fix it?

bstansberry:
@darranl @ropalka In case it helps answer Darran's question, we have a nightly ci.wildfly.org job that tests with SE 26.

Here's a build log link from the most recent run to the start of logging for the relevant tests:

https://ci.wildfly.org/viewLog.html?buildTypeId=WF_MainNightlyJobs_StandardLinuxJdk26&buildId=542494&tab=buildLog&logTab=tree&filter=debug&expand=all&_focus=114250

Note that that job gets 595 failures, so it may be a bit difficult to wade through noise.

My *impression* when I poked at this a week or so ago was EJB clients were failing to connect, presumably because the JDK was rejecting the certificates.

darranl (changes requested):
I have insufficient context to provide a PR preview.

ropalka:
I have added failure.txt to the associated [JIRA](https://issues.redhat.com/browse/WFLY-21326) @darranl for more information.

bstansberry:
I believe our night Linux JDK 26 job is clean except for the failures related to this PR.

https://ci.wildfly.org/viewLog.html?buildId=558043&tab=buildResultsDiv&buildTypeId=WF_MainNightlyJobs_StandardLinuxJdk26

darranl:
I will reproduce and see what I can do to fix the failure.

darranl:
Superceded by https://github.com/wildfly/wildfly/pull/19929

PR #19508, opened 2026-01-08
[WFLY-21317] Allow misc services to declare they should start before …

bstansberry:
…Weld emits application started events

https://issues.redhat.com/browse/WFLY-21317

Note there may be other services where the same thing should happen; the CdiValidatorFactoryService touched here is just the one that's observable via a test failure. AIUI the basic idea is services where ExtensionObserverMethodImpl is going to be delivering extension lifecycle events should be started before WeldStartCompletionServices begins emitting lifecycle events to applications.

I don't know if this is right so it definitely needs a review from @manovotn

### DIFF
--- a/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentCleanupProcessor.java
+++ b/weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentCleanupProcessor.java
@@ -4,7 +4,10 @@
  */
 package org.jboss.as.weld.deployment.processors;
 
+import static org.jboss.as.weld.deployment.AttachmentKeys.START_COMPLETION_DEPENDENCIES;
+
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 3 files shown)


weld/subsystem/src/main/java/org/jboss/as/weld/deployment/processors/WeldDeploymentCleanupProcessor.java
```diff
@@ -85,12 +99,17 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
     private List<ServiceName> getComponentStartServiceNames(DeploymentUnit deploymentUnit) {
         List<ServiceName> serviceNames = new ArrayList<>();
         final EEModuleDescription eeModuleDescription = deploymentUnit.getAttachment(org.jboss.as.ee.component.Attachments.EE_MODULE_DESCRIPTION);
-        if (eeModuleDescription == null) {
+        if (eeModuleDescription != null) {
```
manovotn: @bstansberry why change this? It would now blow up at L105 with NPE if `eeModuleDescription` is `null`.

bstansberry: Thank! Oops; sorry, that is cruft left over from an earlier idea to do as part of this method what my new getMiscServiceDependencies does.

bstansberry: @manovotn Fixed

manovotn (approved):
I've added my sum up of the investigation into [a JIRA comment](https://issues.redhat.com/browse/WFLY-21317?focusedId=28774992&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28774992).

I think this approach is the best way to fix it as changing Weld behavior might cause existing apps to break with little to no benefit as the conditions to encounter this problem are very specific.

Huge thanks to @bstansberry for proposing this fix :)

bstansberry:
Thanks @manovotn

PR #19513, opened 2026-01-10
WFLY-21211 Fix parallel build race conditions in feature pack modules due to missing dependencies

rhusar:
Resolves
https://issues.redhat.com/browse/WFLY-21211

### DIFF
--- a/ee-feature-pack/galleon-feature-pack/pom.xml
+++ b/ee-feature-pack/galleon-feature-pack/pom.xml
@@ -27,12 +27,13 @@
     <packaging>pom</packaging>
 
     <properties>
+        <!-- Some paths below reference target/ directories from other Maven modules, bypassing Maven's dependency management. -->
+        <!-- Any module referenced by these paths MUST be explicitly declared as a <dependency> in this POM to ensure the reactor schedules it to complete BEFORE this module's build starts! -->
         <galleon.shared.resources.directory>${basedir}/../galleon-shared/src/main/resources</galleon.shared.resources.directory>
         <galleon.shared.generated.resources.directory>${basedir}/../galleon-shared/target/index</galleon.shared.generated.resources.directory>
         <galleon.local.resources.directory>${basedir}/../galleon-local/src/main/resources</galleon.local.resources.directory>
         <galleon.client.resources.directory>${basedir}/../../client/shade/target/resources</galleon.client.resources.directory>
         <license.directory>${project.build.directory}/resources/content/docs/licenses</license.directory>
-        <pruned.resources.directory>${basedir}/../pruned/src/main/resources</pruned.resources.directory>
     </properties>
 
     <build>
@@ -301,6 +302,14 @@
             <version>${ee.maven.version}</version>
             <type>pom</type>
             <scope>provided</scope>
+        </dependency>
+
+        <!-- n.b. Add at 'test' scope to ensure wildfly-client-all is built first but is NOT included in the artifact-list.txt or manifest. -->
+        <dependency>
+            <groupId>${ee.maven.groupId}</groupId>
+            <artifactId>wildfly-client-all</artifactId>
+            <version>${ee.maven.version}</version>
+            <scope>test</scope>
         </dependency>
 
     </dependencies>
--- a/preview/feature-pack/pom.xml
+++ b/preview/feature-pack/pom.xml
@@ -26,6 +26,8 @@
     <packaging>pom</packaging>
 
     <properties>
+        <!-- Some paths below reference target/ directories from other Maven modules, bypassing Maven's dependency management. -->
+        <!-- Any module referenced by these paths MUST be explicitly declared as a <dependency> in this POM to ensure the reactor schedules it to complete BEFORE this module's build starts! -->
         <license.directory>${project.build.directory}/resources/content/docs/licenses</license.directory>
 
         <!-- WildFly and ee galleon contents -->
@@ -34,7 +36,6 @@
         <galleon.client.resources.directory>${basedir}/../../client/shade/target/resources</galleon.client.resources.directory>
         <!-- WildFly Preview modules -->
         <galleon.local.resources.directory>${basedir}/../galleon-local/src/main/resources</galleon.local.resources.directory>
-
     </properties>
 
     <build>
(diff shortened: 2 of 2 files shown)


rhusar:
~~n.b. that test in preview fails consistently, started [#wildfly-developers > ReactiveMessagingChannelsTestCase fails in preview](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/ReactiveMessagingChannelsTestCase.20fails.20in.20preview/with/567474390)~~

Resolved since.

ee-feature-pack/galleon-feature-pack/pom.xml
```diff
@@ -303,6 +304,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
```
jfdenise, line 308: Tthis pom depends on wildfly-ee-feature-pack-galleon-shared that in turns depends on wildfly-client-all, do we really need this explicit dependency?

rhusar: @jfdenise That depends how much you want to follow the Maven standards. Strictly, not required. However, in Maven, the industry-standard best practice is to explicitly declare any dependency that the code/module directly uses. Which is the case in this module. Relying on transitive dependencies *that are used directly* is generally considered a bad practice.

preview/feature-pack/pom.xml
```diff
@@ -359,6 +360,21 @@
...
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>${ee.maven.groupId}</groupId>
+            <artifactId>wildfly-ee-galleon-pack</artifactId>
+            <version>${ee.maven.version}</version>
+            <type>zip</type>
+            <scope>provided</scope>
+        </dependency>
```
bstansberry, line 383: What's this about?

rhusar: @bstansberry Good question! Well spotted, I did not include an explanation for this; so I added extensive comment above to explain this:

> Workaround for https://issues.redhat.com/browse/WFGP-308 to support building WildFly with `mvn -T`
>              The wildfly-galleon-maven-plugin is not thread-safe.
>              Both this module and wildfly-ee-galleon-pack use the plugin, and when built in parallel they concurrently access the same ZIP artifacts,
>              causing ClosedFileSystemException due to thread-safety issues in ZIP FileSystem access.
>              This dependency forces Maven's reactor to serialize the builds: wildfly-ee-galleon-pack completes first,
>              then this module builds, preventing concurrent plugin execution.

bstansberry: This adds cruft to the FP's artifact list:

40b21778019d238923e889e9b7cf3017e272d1fd519e7e25410c3eebc2cc01f5,/org/wildfly/wildfly-ee-galleon-pack/39.0.0.Final-SNAPSHOT/wildfly-ee-galleon-pack-39.0.0.Final-SNAPSHOT.pom
2f0a313696b84d34c139688ac824573598ad38bb057e54355b2538f231feb7c9,/org/wildfly/wildfly-ee-galleon-pack/39.0.0.Final-SNAPSHOT/wildfly-ee-galleon-pack-39.0.0.Final-SNAPSHOT.zip

I believe that downstream that ends up driving things that get included in the maven repos we give to customers and what gets into maven.repository.redhat.com. I think we should get WFGP-308 fixed or work out how to exclude deps like these from the feature pack.

ee-feature-pack/galleon-feature-pack/pom.xml
bstansberry: @jfdenise Should wildfly-client-all appear in the artifact-list and manifest? IIRC we are generating the client at provisioning time now.

This is a semi-tangent to this PR, as wildfly-client-all already appears in the artifact-list and manifest without this change. That's because wildfly-ee-feature-pack-galleon-shared declares a dep on it in its pom. The relevance to this PR is if it shouldn't appear there, this change will cause the dep to be used regardless.

preview/feature-pack/pom.xml
rhusar: That is not good. How do I inspect the "FP's artifact list"? 

> or work out how to exclude deps like these from the feature pack.

@jfdenise If you have an idea for ^^^

I will ramp up work on the https://issues.redhat.com/browse/WFGP-308 as the fix should be quite simple semaphoring with reference counting when galleon is opening/reading/closing the zip file system.

bstansberry: The artifact list is one of the artifacts in target, e.g. target/wildfly-ee-galleon-pack-39.0.0.Final-SNAPSHOT-artifact-list.txt

bstansberry: A hack workaround would be to use 'test' scope for these, with a comment explaining the hack. The wildfly-galleon-plugin ignores test scope deps.

rhusar: Great idea, that does the trick for sure.

```
[rhusar@puglia wildfly]$ cat preview/feature-pack/target/wildfly-preview-feature-pack-39.0.0.Final-SNAPSHOT-artifact-list.txt | grep wildfly-ee-galleon-pack | wc -l
       0
```

Changed to test scope and updated the description.

ee-feature-pack/galleon-feature-pack/pom.xml
jfdenise: @bstansberry @rhusar , we need the dependency to make sure that the client-all is built before we build the feature-pack but we don't need it in the manifest. As @bstansberry you suggested, it should have a scope of test here and in  wildfly-ee-feature-pack-galleon-shared to not be included in the manifest.

rhusar: @jfdenise @bstansberry OK thanks for the discussion, I have updated the PR accordingly and added a note.

preview/feature-pack/pom.xml
rhusar: FWIW I started exploring fixing the galleon plugins to support concurrency, while a simple exclusive lock per archive does the trick for the build, I find it very fragile (can result in a deadlock if there is nested locking on archives). The proper way to fix this is in galleon itself, so WIP but just wanted to say it's not exactly trivial to fix right.

bstansberry (approved):

bstansberry:
Thanks @rhusar and @jfdenise

ee-feature-pack/galleon-feature-pack/pom.xml
bstansberry: https://github.com/wildfly/wildfly/pull/19538 cleans up ee-feature-pack-galleon-shared. No test scope dep is needed there; that module does nothing with the output of the client/shaded module.

PR #19529, opened 2026-01-15
WFLY-21354 Upgrade Weld to 5.1.7.Final and 6.0.4.Final

manovotn:
JIRA - https://issues.redhat.com/browse/WFLY-21354

### DIFF
--- a/boms/preview-ee/pom.xml
+++ b/boms/preview-ee/pom.xml
@@ -64,7 +64,7 @@
         <version.org.jboss.metadata>17.0.0.Beta1</version.org.jboss.metadata>
         <version.org.jboss.resteasy>7.0.1.Final</version.org.jboss.resteasy>
         <version.org.jboss.spec.jakarta.el.jakarta.el-api>6.0.2.Final</version.org.jboss.spec.jakarta.el.jakarta.el-api>
-        <version.org.jboss.weld.weld>6.0.3.Final</version.org.jboss.weld.weld>
+        <version.org.jboss.weld.weld>6.0.4.Final</version.org.jboss.weld.weld>
         <version.org.jboss.weld.weld-api>6.0.Final</version.org.jboss.weld.weld-api>
         <version.org.hibernate>7.1.11.Final</version.org.hibernate>
         <version.org.hibernate.models>1.0.1</version.org.hibernate.models>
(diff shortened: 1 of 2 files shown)


jamezp (approved):
This looks good to me. One PR did catch my eye, https://github.com/weld/core/pull/3271. I do not think that is wrong or we should be concerned. I'm merely pointing it out in case we see any test issues or user issues relying on this wrong behavior.

jamezp:
This is the EE11 variant #19530

manovotn:
> This looks good to me. One PR did catch my eye, [weld/core#3271](https://github.com/weld/core/pull/3271). I do not think that is wrong or we should be concerned. I'm merely pointing it out in case we see any test issues or user issues relying on this wrong behavior.

@jamezp  I don't think you can encounter this in WFLY though.
This only concerns situations where there is no Jandex present and Weld falls back to reflection for discovery.

And at least from Weld code, I only saw `DiscoveryStrategyFactory#create` being invoked for Weld SE and Weld Servlet, none of which are used in WFLY. AFAIK, WFLY does its own discovery process - see for instance  org.jboss.as.weld.deployment.processors.`BeanArchiveProcessor`.

jamezp:
> > This looks good to me. One PR did catch my eye, [weld/core#3271](https://github.com/weld/core/pull/3271). I do not think that is wrong or we should be concerned. I'm merely pointing it out in case we see any test issues or user issues relying on this wrong behavior.
> 
> @jamezp I don't think you can encounter this in WFLY though. This only concerns situations where there is no Jandex present and Weld falls back to reflection for discovery.
> 
> And at least from Weld code, I only saw `DiscoveryStrategyFactory#create` being invoked for Weld SE and Weld Servlet, none of which are used in WFLY. AFAIK, WFLY does its own discovery process - see for instance org.jboss.as.weld.deployment.processors.`BeanArchiveProcessor`.

Ah, got it. That makes sense, thank you for the clarification. I'll double check it in RESTEasy then as we do use Weld Servlet there for the `SeBootstrap` implementation.

manovotn:
To be fair I was at first hesitant to include this fix in older Weld versions such as 5 for fear of breaking existing apps. 
However the bug directly breaks specification expectations and would break even TCKs but the test there just doesn't happen to run against SE/servlet. So in the end I decided to do that.

If you find any issues in RESTEasy ping me, I'll be happy to help :)

bstansberry:
Thanks @manovotn and @jamezp

PR #19540, opened 2026-01-22
WFLY-21376 SessionManagerStatistics.getHighestSessionCount() always returns -1 for distributable applications

rhusar:
Also add unit test for DistributableSessionManagerStatistics

Resolves
https://issues.redhat.com/browse/WFLY-21376


pferraro (changes requested):
This implementation would require a full cache key scan for every new session.  Tbis would be prohibitively expensive for the remote Infinispan session manager.

rhusar:
Thats correct as noted in the included comments.

We will have to lower the precision and use some sort of time/invocation snapshot approach or other ideas?

rhusar:
I recon time-based snapshot of active sessions is a reasonable balance between cost and usefulness of the metric.

rhusar:
Actually, there is a smarter way to do this, use local counting and only check the cache when necessary.

PR #19542, opened 2026-01-23
[WFLY-19808] Bump AJP_ALLOWED_REQUEST_ATTRIBUTES_PATTERN to Default s…

baranowb:
…tability
Issue: https://issues.redhat.com/browse/WFLY-19808
Obsoletes: https://github.com/wildfly/wildfly/pull/18284
RFE https://issues.redhat.com/browse/EAP7-1836
Proposal: https://github.com/wildfly/wildfly-proposals/pull/615

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/AjpListenerResourceDefinition.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/AjpListenerResourceDefinition.java
@@ -22,7 +22,6 @@ import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.client.helpers.MeasurementUnit;
 import org.jboss.as.controller.operations.validation.IntRangeValidator;
 import org.jboss.as.controller.registry.AttributeAccess;
-import org.jboss.as.version.Stability;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
 import org.wildfly.extension.io.OptionAttributeDefinition;
@@ -52,7 +51,6 @@ public class AjpListenerResourceDefinition extends ListenerResourceDefinition {
             .setRequired(false)
             .setAllowExpression(true)
             .setDefaultValue(null)
-            .setStability(Stability.COMMUNITY)
             .build();
 
     static final List<AttributeDefinition> ATTRIBUTES = List.of(SCHEME, REDIRECT_SOCKET, MAX_AJP_PACKET_SIZE, ALLOWED_REQUEST_ATTRIBUTES_PATTERN);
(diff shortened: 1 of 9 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/AjpListenerResourceDefinition.java
```diff
@@ -52,7 +52,7 @@ public class AjpListenerResourceDefinition extends ListenerResourceDefinition {
             .setRequired(false)
             .setAllowExpression(true)
             .setDefaultValue(null)
-            .setStability(Stability.COMMUNITY)
+            .setStability(Stability.DEFAULT)
```
rhusar: No need to explicitly set DEFAULT stability as this is the default stability (see org.jboss.as.controller.AbstractAttributeDefinitionBuilder#AbstractAttributeDefinitionBuilder(java.lang.String, org.jboss.dmr.ModelType, boolean)), just remove this line.

rhusar (changes requested):

rhusar (changes requested):
I think this is missing a bunch of process stuff?

proposal is not linked from the PR
proposal not approved by outside-perspective
PR not approved by outside-perspective

istraka:
No issues discovered in QE testing, LGTM.

fl4via (commented):
I am wondering if we do we have to bump to 15, or if we shouldn't we just move from 14 Stability.COMMUNITY, to plain, default, 14?

baranowb:
@bstansberry  ^^ I have no idea about above question.

bstansberry:
@fl4via @baranowb The version needs to be 15. Version 14 at default stability  (i.e. urn:jboss:domain:undertow:14.0) exists since WildFly 28.

Keep in mind that that the normal pattern of these schemas, for a given ##.# version (e.g. 14.0), a urn:jboss:domain:undertow:community:##.# schema is ADDITIVE to urn:jboss:domain:undertow:##.# and a urn:jboss:domain:undertow:preview:##.# schema is ADDITIVE to urn:jboss:domain:undertow:community:##.#.

The schemas for the lower stability levels include new things that are not present in the schemas for the higher stability levels.

NOTE: if we introduce a new thing in 15.0 default that isn't in 14 community or 14 preview, we are going to have to publish a 15 community and 15 preview so we have a schemas that include both the new thing and the existing unpromoted 'preview' and 'community' stuff.

We don't have to introduce those *right now* though because this 'allowed-request-attributes-pattern' xml attribute already exists in 14.community and 14.preview. Someone with a config document that uses those namespaces can validly use the attribute.

bstansberry (changes requested):
@baranowb OTOH UndertowSubsystemModel needs to be updated with a new VERSION_15_0_0 and UndertowSubsystemTransformerTestCase needs to be updated to test transformation to the 14 version. 

I sent you https://github.com/baranowb/wildfly/pull/2 to implement that.

bstansberry (approved):

bstansberry:
Thanks @baranowb @istraka @fl4via

PR #19549, opened 2026-01-23
Update wildfly-bot PR review request settings to align with current c…

bstansberry:
…omponents/leads

No JIRA as it doesn't affect production code or tests.

@emmartins @scottmarlow @fabiobrz @pferraro Please review. Note that it's not required that all areas of the code have someone assigned, so leaving 'notify' as an empty list is ok if no one finds widfly-bot's review requests for that area to be helpful. There can also be more than one person if you want to suggest an additional name(s).

### DIFF
--- a/.github/wildfly-bot.yml
+++ b/.github/wildfly-bot.yml
@@ -20,6 +20,10 @@ wildfly:
         - pferraro
         - rhusar
 
+    - id: concurrency
+      directories: [ concurrency ]
+      notify: [ emmartins ]
+
     - id: connector
       directories: [ connector ]
       notify: [ tadamski ]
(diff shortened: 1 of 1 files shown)


pferraro (approved):

fabiobrz (approved):

emmartins (approved):

rhusar (approved):
LGTM (nevertheless so NOT a fan of Flow Sequence ("arrays"), I much rather prefer Block Sequences as they work much better with SCM)

bstansberry:
tbh i didn't particularly think about flow sequence vs block sequence (terms i didn't know) -- I just kept things consistent with how other multi-element arrays in the file were done, which it seems is block sequences

scottmarlow (approved):

rhusar:
Thanks everyone but especially @bstansberry !

PR #19551, opened 2026-01-26
WFLY-21390 Fix memory leak of ComponentConfiguration after deployment being retained due to lambdas

rhusar:
ComponentConfiguration objects were being retained after deployment due to lambda expressions capturing entire configuration objects when they only needed specific values.
Since these lambdas were stored in long-lived MSC services (e.g., BinderService), the deployment-time configuration remained in memory indefinitely.

Fix lambda captures in ViewDescription and EjbJndiBindingsDeploymentUnitProcessor
Refactor StatefulSessionBeanCacheProvider SPI to avoid retaining configuration.

Resolves
https://issues.redhat.com/browse/WFLY-21390


jamezp (dismissed):

bstansberry (dismissed):

bstansberry (approved):

jamezp (approved):

bstansberry:
Thanks @rhusar and @jamezp 

I'm going to leave the WFLY unresolved for now as it's configured to block 39.0.1 (which I agree with) and that part is still to come.

rhusar:
Thats correct. I will handle the rest in the morning.

PR #19569, opened 2026-01-28
[WFLY-19393][WFLY-19694] default bytecode enhancement on and implement equivalent of having a PersistenceProvider.getClassTransformer() to allow entity enhancement and bootstrap of the persistence unit to occur in two phases without well known chicken + egg initialization problems.

scottmarlow:
https://issues.redhat.com/browse/WFLY-19393 - `Persistence container bytecode enhancement must be enabled by default to ensure better performance.`
https://issues.redhat.com/browse/WFLY-19694 - `Implement equivalent of having a PersistenceProvider.getClassTransformer() to allow entity enhancement and bootstrap of the persistence unit to occur in two phases without well known chicken + egg initialization problems.`

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.jpa.hibernate;
+
+import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
+import org.hibernate.bytecode.enhance.spi.UnloadedClass;
+import org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl;
+
+/**
+ * WildFlyClassTransformer is simple wrapper of Hibernate ORM EnhancingClassTransformerImpl that provides
+ * bytecode enhancing of application deployment classes (e.g. entity classes).
+ *
+ * @author Scott Marlow
+ */
+public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {
+
+    public WildFlyClassTransformer() {
+        super(new DefaultEnhancementContext() {
+
+            @Override
+            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
+                // return false to disable extended enhancement so that non-entity classes are not enhanced.
+                return false;
+            }
+
+        });
+    }
+}
+
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.persistence.jipijapa.hibernate7;
+
+import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
+import org.hibernate.bytecode.enhance.spi.UnloadedClass;
+import org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl;
+
+/**
+ * WildFlyClassTransformer is simple wrapper of Hibernate ORM EnhancingClassTransformerImpl that provides
+ * bytecode enhancing of application deployment classes (e.g. entity classes).
+ *
+ * @author Scott Marlow
+ */
+public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {
+
+    public WildFlyClassTransformer() {
+        super(new DefaultEnhancementContext() {
+
+            @Override
+            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
+                // return false to disable extended enhancement so that non-entity classes are not enhanced.
+                return false;
+            }
+
+        });
+    }
+}
+
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
@@ -100,6 +100,8 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
     private volatile List<String> qualifierAnnotationNames = List.of();
     private volatile boolean isDuplicate;
 
+    private static final String ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_CLASS = "org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl";
+
     @Override
     public void setPersistenceUnitName(String name) {
         this.name = name;
@@ -376,6 +378,13 @@ public class PersistenceUnitMetadataImpl implements PersistenceUnitMetadata {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // WFLY-19694 Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        // which were already added earlier via org.jboss.as.jpa.hibernate.WildFlyClassTransformer
+        // or org.wildfly.persistence.jipijapa.hibernate7.WildFlyClassTransformer
+        if (classTransformer.getClass().getName().equals(ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_CLASS)) {
+            return;
+        }
         transformers.add(classTransformer);
         if (ROOT_LOGGER.isTraceEnabled()) {
             ROOT_LOGGER.tracef("added entity class transformer '%s' for '%s'",
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernatePersistenceProviderAdaptor.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernatePersistenceProviderAdaptor.java
@@ -214,6 +214,11 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
         hibernateExtendedBeanManager.beanManagerIsAvailableForUse();
     }
 
+    @Override
+    public void addClassFileTransformer(PersistenceUnitMetadata pu) {
+        pu.addTransformer(new WildFlyClassTransformer());
+    }
+
     /* start of TwoPhaseBootstrapCapable methods */
 
     public EntityManagerFactoryBuilder getBootstrap(final PersistenceUnitInfo info, final Map map) {
@@ -221,5 +226,6 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
     }
 
     /* end of TwoPhaseBootstrapCapable methods */
+
 }
 
--- a/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/HibernatePersistenceProviderAdaptor.java
+++ b/jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/HibernatePersistenceProviderAdaptor.java
@@ -218,6 +218,11 @@ public class HibernatePersistenceProviderAdaptor implements PersistenceProviderA
         hibernateExtendedBeanManager.beanManagerIsAvailableForUse();
     }
 
+    @Override
+    public void addClassFileTransformer(PersistenceUnitMetadata pu) {
+        pu.addTransformer(new WildFlyClassTransformer());
+    }
+
     /* start of TwoPhaseBootstrapCapable methods */
 
     public EntityManagerFactoryBuilder getBootstrap(final PersistenceUnitInfo info, final Map map) {
--- a/jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceProviderAdaptor.java
+++ b/jpa/spi/src/main/java/org/jipijapa/plugin/spi/PersistenceProviderAdaptor.java
@@ -94,5 +94,8 @@ public interface PersistenceProviderAdaptor {
     Object beanManagerLifeCycle(BeanManager beanManager);
 
     void markPersistenceUnitAvailable(Object wrapperBeanManagerLifeCycle);
+
+    default void addClassFileTransformer(PersistenceUnitMetadata pu) {
+    }
 }
 
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/config/Configuration.java
@@ -7,7 +7,6 @@ package org.jboss.as.jpa.config;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 
 import jakarta.persistence.EntityManagerFactory;
 
@@ -272,34 +271,7 @@ public class Configuration {
         if (pu.getProperties().containsKey(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER)) {
             return Boolean.parseBoolean(pu.getProperties().getProperty(Configuration.JPA_CONTAINER_CLASS_TRANSFORMER));
         }
-        if (isHibernateProvider(pu.getPersistenceProviderClassName())) {
-
-            return isAnyHibernateEnhancerPropertySpecified(pu);
-        }
         return true;
-    }
-
-    /**
-     * Consider jboss.as.jpa.classtransformer to be set to true if any hibernate.enhancer properties are set to true.
-     *
-     * @param pu
-     * @return true if any hibernate.enhancer property is detected otherwise return false
-     */
-    private static boolean isAnyHibernateEnhancerPropertySpecified(PersistenceUnitMetadata pu) {
-
-        Set<String> set = pu.getProperties().stringPropertyNames();
-        for (String key : set) {
-            if (key.startsWith(HIBERNATE_ENHANCER) &&
-                    (true == Boolean.parseBoolean(pu.getProperties().getProperty(key)))) {
-                return true;
-            }
-        }
-        return false;
-    }
-
-    private static boolean isHibernateProvider(String provider) {
-        // If the persistence provider is not specified (null case), Hibernate ORM will be used as the persistence provider.
-        return provider == null || provider.contains(HIBERNATE);
     }
 
     // key = provider class name, value = adapter module name
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
@@ -1062,9 +1062,9 @@ public class PersistenceUnitServiceHandler {
                     final PersistenceProviderAdaptor adaptor = getPersistenceProviderAdaptor(pu, persistenceProviderDeploymentHolder, phaseContext.getDeploymentUnit(), provider, platform);
                     final boolean twoPhaseBootStrapCapable = (adaptor instanceof TwoPhaseBootstrapCapable) && Configuration.allowTwoPhaseBootstrap(pu);
                     // only add the next phase dependency, if the persistence unit service is starting early.
-                    if( !Configuration.allowApplicationDefinedDatasource(pu) && !Configuration.allowDefaultDataSourceUse(pu) && twoPhaseBootStrapCapable ) {
+                    if( Configuration.needClassFileTransformer(pu) && !Configuration.allowApplicationDefinedDatasource(pu) && !Configuration.allowDefaultDataSourceUse(pu) ) {
                         // wait until the persistence unit service is started before starting the next deployment phase
-                        phaseContext.addToAttachmentList(Attachments.NEXT_PHASE_DEPS, puServiceName.append(FIRST_PHASE) );
+                        phaseContext.addToAttachmentList(Attachments.NEXT_PHASE_DEPS, twoPhaseBootStrapCapable ? puServiceName.append(FIRST_PHASE) : puServiceName);
                     }
                 }
             }
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/processor/PersistenceUnitServiceHandler.java
@@ -105,7 +105,6 @@ public class PersistenceUnitServiceHandler {
 
     private static final String ENTITYMANAGERFACTORY_JNDI_PROPERTY = "jboss.entity.manager.factory.jndi.name";
     private static final String ENTITYMANAGER_JNDI_PROPERTY = "jboss.entity.manager.jndi.name";
-    public static final ServiceName BEANMANAGER_NAME = ServiceName.of("beanmanager");
 
     private static final AttachmentKey<Map<String,PersistenceProviderAdaptor>> providerAdaptorMapKey = AttachmentKey.create(Map.class);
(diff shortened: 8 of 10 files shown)


scottmarlow:
@yrodiere could you please check that the jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java + jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java files look correct to you?

scottmarlow:
Adding `@Override` + a few more methods...

yrodiere:
I will have a look tomorrow, but I understand you've been working on this with Steve or Gavin? If so you'll probably want them to check.

scottmarlow:
> I will have a look tomorrow, but I understand you've been working on this with Steve or Gavin? If so you'll probably want them to check.

Thank you!  I'll push more changes later today anyway to add missing ` @Override` and other (super) DefaultEnhancementContext methods.  I'm adding all of the DefaultEnhancementContext methods mostly for documentation purposes to make it clear that WildFly is overriding the super class methods and also get a warning if any of the super class methods are removed (only action would be to acknowledge the change mostly).

I'll reach out to Gavin + Steve as well after the next push.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,93 @@
...
+
+            @Override
+            public boolean isDiscoveredType(UnloadedClass classDescriptor) {
+                return super.isDiscoveredType(classDescriptor);
+            }
+
+            @Override
+            public void registerDiscoveredType(UnloadedClass classDescriptor, Type.PersistenceType type) {
+                super.registerDiscoveredType(classDescriptor, type);
+            }
```
jamezp: Do we need to override these if they're just invoking the super method?

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+
+            @Override
+            public boolean isDiscoveredType(UnloadedClass classDescriptor) {
+                return super.isDiscoveredType(classDescriptor);
+            }
+
+            @Override
+            public void registerDiscoveredType(UnloadedClass classDescriptor, Type.PersistenceType type) {
+                super.registerDiscoveredType(classDescriptor, type);
+            }
```
jamezp: Same note here.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -376,6 +378,13 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // WFLY-19694 Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        // which were already added earlier via org.jboss.as.jpa.hibernate.WildFlyClassTransformer
+        // or org.wildfly.persistence.jipijapa.hibernate7.WildFlyClassTransformer
+        if(classTransformer.getClass().getName().startsWith(ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_PACKAGE)) {
+            return;
+        }
```
jamezp: This would skip **any** `org.hibernate` transformers. Is that intended?

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
scottmarlow: We could remove the @Override actually.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
scottmarlow: We can remove the override here also good point.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Yes that is intended as the org.hibernate ORM transformer has already been added by the time this call is reached.

Technically, we could still register the org.hibernate ORM transformer also but that will make it even more difficult to debug Hibernate bytecode enhancement problems with complicated EAR deployments as there would be double the number of calls to transform each application entity class * the number of WildFly threads that are deploying/starting Persistence Units.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
scottmarlow: Addressed!  Thanks!

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
scottmarlow: Addressed!  Thanks!

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
jamezp: No problem. Just the comment variable is called `ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_PACKAGE`,  but `org.hibernate` felt pretty generic. For example if something like Hibernate Search had a transformer, it would also be ignored.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+                return super.order(persistentFields);
+            }
+
+            public boolean isDiscoveredType(UnloadedClass classDescriptor) {
+                return super.isDiscoveredType(classDescriptor);
+            }
+
+            public void registerDiscoveredType(UnloadedClass classDescriptor, Type.PersistenceType type) {
+                super.registerDiscoveredType(classDescriptor, type);
+            }
```
jamezp: Did something weird happen in the commit? I'm still seeing these here.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,80 @@
...
+                return super.order(persistentFields);
+            }
+
+            public boolean isDiscoveredType(UnloadedClass classDescriptor) {
+                return super.isDiscoveredType(classDescriptor);
+            }
+
+            public void registerDiscoveredType(UnloadedClass classDescriptor, Type.PersistenceType type) {
+                super.registerDiscoveredType(classDescriptor, type);
+            }
```
jamezp: Same note about these.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
scottmarlow: I saw saw the @Overrides for a few minutes after I pushed an updated commit.  I did leave one @Override in for the one method that doesn't call the super method.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
scottmarlow: Maybe refresh?

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Hmm, good point, I think we should get rid of the variable and just reference the class directly.

scottmarlow: Well maybe not the actual class but more of the name would be better.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
jamezp: These are definitely still showing up. I pulled down the PR and diff'ed it and I'm still seeing them in both files.

The one override definitely make sense as it always returns `false`. The others won't hurt, but it makes it harder to read IMO.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
jamezp: Looks better to me. Thank you.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
scottmarlow: Here is what I see locally and on github (when I view changed files view):

```
public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {

    public WildFlyClassTransformer() {
        super(new DefaultEnhancementContext() {

            public boolean isEntityClass(UnloadedClass classDescriptor) {
                return super.isEntityClass(classDescriptor);
            }

            public boolean isCompositeClass(UnloadedClass classDescriptor) {
                return super.isCompositeClass(classDescriptor);
            }

            public boolean isMappedSuperclassClass(UnloadedClass classDescriptor) {
                return super.isMappedSuperclassClass(classDescriptor);
            }

            public boolean doBiDirectionalAssociationManagement(UnloadedField field) {
                return super.doBiDirectionalAssociationManagement(field);
            }

            public boolean doDirtyCheckingInline(UnloadedClass classDescriptor) {
                return super.doDirtyCheckingInline(classDescriptor);
            }

            public boolean hasLazyLoadableAttributes(UnloadedClass classDescriptor) {
                return super.hasLazyLoadableAttributes(classDescriptor);
            }

            public boolean isLazyLoadable(UnloadedField field) {
                return super.isLazyLoadable(field);
            }

            public boolean isPersistentField(UnloadedField ctField) {
                return super.isPersistentField(ctField);
            }

            public boolean isMappedCollection(UnloadedField field) {
                return super.isMappedCollection(field);
            }

            public UnloadedField[] order(UnloadedField[] persistentFields) {
                return super.order(persistentFields);
            }

            public boolean isDiscoveredType(UnloadedClass classDescriptor) {
                return super.isDiscoveredType(classDescriptor);
            }

            public void registerDiscoveredType(UnloadedClass classDescriptor, Type.PersistenceType type) {
                super.registerDiscoveredType(classDescriptor, type);
            }

            @Override
            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
                // doesn't make any sense to have extended enhancement enabled at runtime. we only enhance entities anyway.
                return false;
            }

        });
    }
}

```

scottmarlow: If your still seeing the @Override on each method I probably need to create a new pull request because something is very odd.

jamezp: Wait, sorry. I think there is a misunderstanding. I think all these methods should be removed. We don't gain anything from having them. The `@Override` annotation should be used IMO when overriding a method. I just don't see any reason to override these methods.

scottmarlow: As an example, if we remove the `isDiscoveredType` method and the super class happened to remove the `isDiscoveredType` method that isn't a big deal but with the current code we would get a WildFly build error due to the missing super `isDiscoveredType` method:

```
public boolean isDiscoveredType(UnloadedClass classDescriptor) {
    return super.isDiscoveredType(classDescriptor);
}
```

Is ^ a big deal?  Maybe yes or maybe no as it depends on what changed.  Perhaps this is a false safety illusion so I will remove these methods because we have no protection against other changes like a new super method being added.  Of course the new super method would be called fine but we really don't have a reasonable way of detecting when methods have been added/removed in the super class, so better to follow your advice I think.  I'll push another change and thanks for review!

scottmarlow: I cleaned up my local changes and pushed again with a few updated comments as well.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
scottmarlow: I cleaned up my local changes and pushed again with a few updated comments as well.

jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
jamezp: > As an example, if we remove the `isDiscoveredType` method and the super class happened to remove the `isDiscoveredType` method that isn't a big deal but with the current code we would get a WildFly build error due to the missing super `isDiscoveredType` method:
> 
> ```
> public boolean isDiscoveredType(UnloadedClass classDescriptor) {
>     return super.isDiscoveredType(classDescriptor);
> }
> ```
> 
> Is ^ a big deal? Maybe yes or maybe no as it depends on what changed. Perhaps this is a false safety illusion so I will remove these methods because we have no protection against other changes like a new super method being added. Of course the new super method would be called fine but we really don't have a reasonable way of detecting when methods have been added/removed in the super class, so better to follow your advice I think. I'll push another change and thanks for review!

I wouldn't worry too much about that TBH. That shouldn't happen in a patch release and something we should catch in a minor release. If they removed it though, then we wouldn't need the override anyway. If they make it `abstract` instead, building fails anyway because we don't implement. 

IoW, I don't think we need to worry about it :)

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
jamezp, line 387: One more thing on this. My understanding is we're trying to not register `org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl` because we have the `WildFlyClassTransformer`. With this, we exclude **all** transformers in this package. Is that okay?

Could we check if the class transformer is an instance of `org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl` and just ignore those? Maybe not as that might exclude some custom user transformer I guess... ....anyway just kind of wondering if we really want to ignore all transformers in that package.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/hibernate/classfiletransformertest/ClassFileTransformerTestCase.java
```diff
@@ -75,10 +74,10 @@ public void testhibernate_ejb_use_class_enhancer() throws Exception {
     }
 
     @Test
-    public void testHibernateByteCodeEnhancementIsDisabledByDefault() {
+    public void testHibernateByteCodeEnhancementIsEnabledByDefault() {
         // Note: ManagedTypeHelper is an internal Hibernate ORM class, if it is removed or renamed then this test can be updated
         // accordingly.
-        assertFalse("Employee class is not bytecode enhanced", org.hibernate.engine.internal.ManagedTypeHelper.isManagedType(Employee.class));
+        assertTrue("Employee class is bytecode enhanced by default", org.hibernate.engine.internal.ManagedTypeHelper.isManagedType(Employee.class));
```
jamezp, line 80: Is there a way to keep this test, but add a new one?

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Yeah, I like that idea.  Thanks!

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * WildFlyClassTransformer is simple wrapper of Hibernate ORM EnhancingClassTransformerImpl that provides
+ * bytecode enhancing of application deployment classes (e.g. entity classes).
+ *
+ * @author Scott Marlow
+ */
+public class WildFlyClassTransformer extends EnhancingClassTransformerImpl {
+
+    public WildFlyClassTransformer() {
+        super(new DefaultEnhancementContext() {
```
scottmarlow, line 20: Need to remove unneeded imports.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/hibernate/classfiletransformertest/ClassFileTransformerTestCase.java
scottmarlow: Will add a new test that has a new name but tests the functionality of the old `testHibernateByteCodeEnhancementIsDisabledByDefault`.

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
jamezp: That's the one I miss all the time :)

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/jpa/hibernate/classfiletransformertest/ClassFileTransformerTestCase.java
scottmarlow: Hmm, actually we need a separate test which is fine.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -376,6 +378,14 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // WFLY-19694 Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        // which were already added earlier via org.jboss.as.jpa.hibernate.WildFlyClassTransformer
+        // or org.wildfly.persistence.jipijapa.hibernate7.WildFlyClassTransformer
+        if(classTransformer.getClass().getName().equals(ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_CLASS)) {
+            ROOT_LOGGER.infof("Skipping add of Hibernate ORM transformer %s", classTransformer.getClass().getName());
```
scottmarlow: I forgot to remove this print output.

scottmarlow: For reference the following output is from a local unit test run:

> 2026-02-06 12:19:44,739 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 5) Skipping add of Hibernate ORM transformer org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl
> 2026-02-06 12:19:44,739 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 23) Skipping add of Hibernate ORM transformer org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl
> 2026-02-06 12:19:44,740 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 23) WFLYJPA0010: Starting Persistence Unit (phase 2 of 2) Service 'jpa_ClassFileTransformerTestCase.jar#SecondPU'
> 2026-02-06 12:19:44,740 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 5) WFLYJPA0010: Starting Persistence Unit (phase 2 of 2) Service 'jpa_ClassFileTransformerTestCase.jar#mypc'
> 2026-02-06 12:19:44,742 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 86) Skipping add of Hibernate ORM transformer org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl
> 2026-02-06 12:19:44,742 INFO  [org.jboss.as.jpa] (ServerService Thread Pool -- 86) WFLYJPA0010: Starting Persistence Unit (phase 2 of 2) Service 'jpa_ClassFileTransformerTestCase.jar#ThirdPU'
> 202

jpa/hibernate7/src/main/java/org/wildfly/persistence/jipijapa/hibernate7/WildFlyClassTransformer.java
```diff
@@ -0,0 +1,31 @@
...
+
+            @Override
+            public boolean doExtendedEnhancement(UnloadedClass classDescriptor) {
+                // return false to disable extended enhancement so that non-entity classes are not enhanced.
```
scottmarlow, line 24: FYI, think of enabling extended enhancement for an entity class field would mean that non-entity (application) classes that reference the fields will also be enhanced for lazy loading and similar optimizations.

jamezp:
I've manually kicked off new runs of the jobs that failed because the commit couldn't be found. This is a known issue when pushes happen too fast. There is a large queue on CI right now though, so they will take a bit to get started.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
```diff
@@ -376,6 +378,13 @@ public boolean needsJPADelegatingClassFileTransformer() {
 
     @Override
     public void addTransformer(ClassTransformer classTransformer) {
+        // WFLY-19694 Do not add Hibernate ORM 6.x/7.x bytecode enhancement class transformers
+        // (e.g. org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl)
+        // which were already added earlier via org.jboss.as.jpa.hibernate.WildFlyClassTransformer
+        // or org.wildfly.persistence.jipijapa.hibernate7.WildFlyClassTransformer
+        if(classTransformer.getClass().getName().equals(ORG_HIBERNATE_ORM_PROVIDER_CLASS_ENHANCER_CLASS)) {
```
jamezp: This is minor and should not block this PR from being merged. There should be a space between `if(`. It should be `if (`.

jamezp (commented):
This looks good to me. I've got one minor comment, but as I said in the comment this should not block this PR. If CI passes I'll approve it, but I'd like to wait for the CI runs to complete before I approve it.

jpa/subsystem/src/main/java/org/jboss/as/jpa/config/PersistenceUnitMetadataImpl.java
scottmarlow: Good catch, I'll push a fix, thanks!

scottmarlow: I pushed a change for ^.  Fyi test checks from yesterday passed.

scottmarlow:
Thanks James for the review + approval!

scottmarlow:
Thanks @jamezp again for the review!   Thanks @gavinking for the idea behind this change and adding it to Persistence 4.0!  Thanks @yrodiere @sebersole for reviewing this change and feedback elsewhere!

PR #19576, opened 2026-01-29
[WFLY-21407][WFLY-21464] Upgrade WildFly Maven Plugin to 6.0.0.Beta2

mchoma:
https://issues.redhat.com/browse/WFLY-21407

____

> [!WARNING]

> * [WFLY-21464](https://issues.redhat.com/browse/WFLY-21464)
> * [WFLY-21406](https://issues.redhat.com/browse/WFLY-21406)

### DIFF
--- a/testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/management/ProvisioningConsistencyTestCase.java
+++ b/testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/management/ProvisioningConsistencyTestCase.java
@@ -5,7 +5,6 @@
 
 package org.wildfly.test.manual.management;
 
-
 import org.wildfly.test.distribution.validation.ProvisioningConsistencyBaseTest;
 
 /**
--- a/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
+++ b/ee-feature-pack/galleon-local/src/main/resources/configs/domain/model.xml
@@ -25,6 +25,7 @@
         <package name="cleanup.domain.servers.dir" optional="true"/>
         <package name="cleanup.domain.log.dir" optional="true"/>
         <package name="cleanup.domain.data.dir" optional="true"/>
+        <package name="cleanup.domain.tmp.dir" optional="true"/>
         <!-- Unstable API Annotation Index -->
         <package name="unstable-api-annotation-index.wildfly-ee-feature-pack" optional="true" valid-for-stability="preview"/>
         <package name="org.wildfly.unstable.annotation.api.indexer" optional="true" valid-for-stability="preview" />
--- a/ee-feature-pack/galleon-local/src/main/resources/configs/standalone/model.xml
+++ b/ee-feature-pack/galleon-local/src/main/resources/configs/standalone/model.xml
@@ -28,7 +28,7 @@
         <package name="cleanup.standalone.config.history.dir" optional="true"/>
         <package name="cleanup.standalone.log.dir" optional="true"/>
         <package name="cleanup.standalone.data.dir" optional="true"/>
-        <package name="cleanup.standalone.tmp.vfs" optional="true"/>
+        <package name="cleanup.standalone.tmp" optional="true"/>
 
         <package name="org.wildfly.bootable-jar" optional="true"/>
         <package name="org.wildfly.embedded"/>
--- a/pom.xml
+++ b/pom.xml
@@ -299,7 +299,7 @@
         <version.asciidoctor.plugin>2.2.6</version.asciidoctor.plugin>
         <version.central.publishing.maven.plugin>0.9.0</version.central.publishing.maven.plugin>
         <version.org.jacoco>0.8.14</version.org.jacoco>
-        <version.org.jboss.galleon>7.0.0.Final</version.org.jboss.galleon>
+        <version.org.jboss.galleon>7.0.2.Final</version.org.jboss.galleon>
         <version.org.owasp.dependency-check>12.2.0</version.org.owasp.dependency-check>
         <version.org.wildfly.bom-builder-plugin>2.0.10.Final</version.org.wildfly.bom-builder-plugin>
         <version.org.wildfly.checkstyle-config>1.0.8.Final</version.org.wildfly.checkstyle-config>
@@ -307,7 +307,7 @@
         <version.org.wildfly.galleon-plugins>8.1.0.Final</version.org.wildfly.galleon-plugins>
         <version.org.wildfly.glow>1.5.2.Final</version.org.wildfly.glow>
         <version.org.wildfly.licenses.plugin>2.4.3.Final</version.org.wildfly.licenses.plugin>
-        <version.org.wildfly.plugin>5.1.0.Final</version.org.wildfly.plugin>
+        <version.org.wildfly.plugin>6.0.0.Beta2</version.org.wildfly.plugin>
         <version.org.wildfly.unstable.annotation.api>1.0.2.Final</version.org.wildfly.unstable.annotation.api>
         <version.org.wildfly.wildfly-channel-plugin>1.0.30</version.org.wildfly.wildfly-channel-plugin>
         <version.org.wildfly.wildfly-maven-gpg-plugin>3.2.8.SP1</version.org.wildfly.wildfly-maven-gpg-plugin>
--- a/preview/galleon-local/src/main/resources/configs/domain/model.xml
+++ b/preview/galleon-local/src/main/resources/configs/domain/model.xml
@@ -25,6 +25,7 @@
         <package name="cleanup.domain.servers.dir" optional="true"/>
         <package name="cleanup.domain.log.dir" optional="true"/>
         <package name="cleanup.domain.data.dir" optional="true"/>
+        <package name="cleanup.domain.tmp.dir" optional="true"/>
 
         <package name="org.wildfly.deployment.transformation"/>
 
--- a/preview/galleon-local/src/main/resources/configs/standalone/model.xml
+++ b/preview/galleon-local/src/main/resources/configs/standalone/model.xml
@@ -25,7 +25,7 @@
         <package name="cleanup.standalone.config.history.dir" optional="true"/>
         <package name="cleanup.standalone.log.dir" optional="true"/>
         <package name="cleanup.standalone.data.dir" optional="true"/>
-        <package name="cleanup.standalone.tmp.vfs" optional="true"/>
+        <package name="cleanup.standalone.tmp" optional="true"/>
 
         <package name="org.wildfly.bootable-jar" optional="true"/>
         <package name="org.wildfly.embedded"/>


jamezp (approved):

rhusar (dismissed):
LGTM

rhusar:
FYI This should have come via dependabot, but it hit an error:

```
updater | 2026/01/28 18:54:05 INFO <job_1224724717> Updating org.wildfly.plugins:wildfly-maven-plugin from 5.1.0.Final to 5.1.5.Final
  proxy | 2026/01/28 18:54:05 [320] POST /update_jobs/1224724717/record_update_job_error
2026/01/28 18:54:05 [320] 204 /update_jobs/1224724717/record_update_job_error (cached)
updater | 2026/01/28 18:54:05 INFO <job_1224724717> Handled error whilst updating org.wildfly.plugins:wildfly-maven-plugin: dependency_file_not_evaluatable {message: "ERROR: Invalid expression: /project/groupId}.channel"}
```

I am putting in a workaround for the above ^

rhusar:
@jamezp let me know if you think it's worth back-porting to 39.x.

mchoma:
@rhusar +1 for this beeing automatic upgrade via dependabot

jamezp:
> @jamezp let me know if you think it's worth back-porting to 39.x.

It seems fine for 39.x too, but to me it's not critical unless it ends up in the BOM's which I don't know the answer to :)

jamezp:
> FYI This should have come via dependabot, but it hit an error:
> 
> ```
> updater | 2026/01/28 18:54:05 INFO <job_1224724717> Updating org.wildfly.plugins:wildfly-maven-plugin from 5.1.0.Final to 5.1.5.Final
>   proxy | 2026/01/28 18:54:05 [320] POST /update_jobs/1224724717/record_update_job_error
> 2026/01/28 18:54:05 [320] 204 /update_jobs/1224724717/record_update_job_error (cached)
> updater | 2026/01/28 18:54:05 INFO <job_1224724717> Handled error whilst updating org.wildfly.plugins:wildfly-maven-plugin: dependency_file_not_evaluatable {message: "ERROR: Invalid expression: /project/groupId}.channel"}
> ```
> 
> I am putting in a workaround for the above ^

I'm thinking we may need to update dependabot to force it to look at Maven Central. Part of the issue will be it won't pick up dependencies that live both in JBoss Nexus and Maven Central. It will only pick the latest from JBoss Nexus.

jfdenise (approved):

rhusar:
> @rhusar +1 for this beeing automatic upgrade via dependabot

Opened https://github.com/wildfly/wildfly/pull/19577 with a "fix".

rhusar:
> > @jamezp let me know if you think it's worth back-porting to 39.x.
> 
> It seems fine for 39.x too, but to me it's not critical unless it ends up in the BOM's which I don't know the answer to :)

Just checked - it does not (with-tools). So, this is  moot for 39.x. 👍🏼

rhusar:
> I'm thinking we may need to update dependabot to force it to look at Maven Central. Part of the issue will be it won't pick up dependencies that live both in JBoss Nexus and Maven Central. It will only pick the latest from JBoss Nexus.

OK, I suggest open a Jira / Zulip with evidence so we can have a look at what's really going on. 

I had already done that in different repository, but for different reason really e.g. here https://github.com/modcluster/mod_cluster/blob/main/.github/dependabot.yml#L2-L7

jbliznak:
> unless it ends up in the BOM's

JFYI BOMs don't / can't manage plugin versions so even if it end up there it would have no effect

jamezp:
> > unless it ends up in the BOM's
> 
> JFYI BOMs don't / can't manage plugin versions so even if it end up there it would have no effect

Oh, of course. Sorry, not sure what I was thinking there :)

mchoma:
@jamezp @jfdenise ProvisioningConsistencyTestCase is failing, so maybe there is some behaviour change in new version of plugin.

jfdenise:
@mchoma , the culprit is this commit that appeared in 5.1.1.Final of the plugin: https://github.com/wildfly/wildfly-maven-plugin/commit/d7c75267a400605c34d929c0c3304159a7d86c7c#diff-0b2f1e793aa9ad4aef151dba847084ec8fce8d49fe5ae1925313e2f66f018962
This dependency has a side effect on the XML writer used by Galleon.
Galleon naked:
```
<?xml version="1.0" ?>
```
Starting 5.1.1.FInal Maven plugin
```
<?xml version='1.0' encoding='UTF-8'?>
```
So the encoding is now explicit. UTF-8, being the default that has no effect.

I am investigating more to see what is the exact XML API that is impacted.

rhusar:
However, being explicit with the encoding is generally considered a "best practice."

Nevertheless, should not be a reason to fail the test.

jfdenise:
So the introduced transitive woodstock dependency defines a ServiceLoader that is used to retrieve the XML writer. instead of the default JVM implementation: com.sun.xml.internal.stream.writers.XMLStreamWriterImpl, com.ctc.wstx.sw.SimpleNsStreamWriter is used and handles differently the header.

I think that we shouldn't fix anything in the WildFly Maven Plugin, the output is correct.
I don't see why we should fix Galleon to now output the encoding. It never did it.

@mchoma the test does a diff of the file size blindly, we could create a special case for JBOSS_HOME/.galleon/*.xml files (not invest in complex semantic XML diffing). Then read all the lines after the first one (the header) and diff.

jbliznak:
> I think that we shouldn't fix anything in the WildFly Maven Plugin, the output is correct.
> I don't see why we should fix Galleon to now output the encoding. It never did it.

One could argue that it would be good to produce the same output across all tooling to keep consistency (regardless of the naivity of the test here)

mchoma:
> I don't see why we should fix Galleon to now output the encoding. It never did it.

Well, if explicit encoding is considered better (which I would agree), that sounds to me as good reason.

So I would propose to fill the Galleon issue and fix it. I can't say if we need to wait for Galleon fix to continue with this.

Overall this is getting too complicated for me to finish. I was just hoping of easy upgrade. I do not feel comfortable with relaxing test in scope of this MR.

rhusar:
>  I do not feel comfortable with relaxing test in scope of this MR.

@mchoma You could just fix the test to use proper XML comparison rather than file equality.

bstansberry:
We shouldn't make ProvisioningConsistencyTestCase more complicated. It's not really testing WildFly, it is a helper for testing tooling in the WF ecosystem. The tool we tell our users to use is wildfly-maven-plugin -- testing galleon-maven-plugin in this way is helpful to the Galleon project but isn't really relevant to WildFly users. If we're not going to maintain our tooling such that wildfly-maven-plugin and galleon-maven-plugin both produce the same result we should drop the test. 

We already added complexity to this test to deal with an issue with s390 and at that time we considered dropping the test.

Historical note: when I added this test we were still using galleon-maven-plugin to produce the zips/tars that we published on the GH release page and linked from widlfly.org/downloads. At that point it really was a way of testing what we produced for users. But we stopped using galleon-maven-plugin for that purpose quite a while ago.

So, I think we either file an issue to get a galleon-maven-plugin version that matches and we @Ignore the test referencing that, or we delete the test.

jamezp:
FWIW we don't have to upgrade. We're going to need to upgrade, I think, to 6.0.0 once I get it released. We'll need to fix it then, but we have time. We only use the plugin for provisioning.

mchoma:
> I think we either file an issue to get a galleon-maven-plugin version that matches and we @ignore the test referencing that

@bstansberry I have created https://issues.redhat.com/browse/WFGP-309 and Ignored ProvisioningConsistencyTestCase. Let me know if I understand you correctly and if project WFGP is correct one (I am not sure).

jfdenise:
@mchoma , I have moved the issue to https://issues.redhat.com/browse/GAL-379, the encoding is missing when provisioning using Galleon Maven Plugin.
I am fixing the issue, will release a new Galleon 7.0.1.Final.

rhusar:
> @rhusar +1 for this beeing automatic upgrade via dependabot

OK I have fixed this, and we got the upgrade here - https://github.com/wildfly/wildfly/pull/19595 - closed it to continue discussion and action items on this PR.

jfdenise:
There is a side effect of the WildFly Maven Plugin new XML dep. The syntax for the XML header is using single quote, Galleon uses double quote. Both are valid. I don't think that we should align on a side effect of a dependency.

The test check the size of the file, both character are of the same size, so the test will pass.

jamezp:
@jfdenise Would a better solution be to exclude the transitive `com.fasterxml.woodstox:woodstox-core` dependency? To be fair, maybe we need to downgrade the Plexus dependency anyway as I just noticed it seems to depend on Maven 4.
```
[INFO] org.wildfly.plugins:wildfly-maven-plugin:maven-plugin:6.0.0.Beta1-SNAPSHOT
[INFO] \- org.codehaus.plexus:plexus-xml:jar:4.1.1:compile
[INFO]    \- org.apache.maven:maven-xml:jar:4.0.0-rc-5:compile
[INFO]       \- com.fasterxml.woodstox:woodstox-core:jar:7.1.1:compile
```

jfdenise:
@jamezp , I didn't analyze the actual need of the transitive. I was under the assumption that it would be needed. But if we are going to remove the `plexus-xml` dep, then yes, the fix for Galleon seems still relevant (explicit encoding).

jfdenise:
@mchoma Galleon 7.01.Final has been released.

mchoma:
> @mchoma Galleon 7.01.Final has been released.

I have updated galleon to 7.0.1.Final and enabled ProvisioningConsistencyTestCase, which is passing locally.

mchoma:
Test ProvisioningConsistencyTestCase keep failing on windows :crying_cat_face: 

@jfdenise are you able to diagnose what is wrong this time? 
```
dist file for C:\BuildAgent\work\dc9a11b6a14b0d86\ee-dist\target\wildfly-40.0.0.Beta1-SNAPSHOT\.galleon\provisioning.xml has an unexpected length: 438 not 474
```

36 chars difference this time

mchoma:
On galleon upgrade PR https://github.com/wildfly/wildfly/pull/19605 I see 
```
dist file for C:\BuildAgent\work\dc9a11b6a14b0d86\ee-dist\target\wildfly-40.0.0.Beta1-SNAPSHOT\.galleon\provisioning.xml has an unexpected length: 438 not 422]
```
So I guess what is happening is maven plugin and its output brings something more to provisioning.xml on windows.

jfdenise:
@mchoma , the windows failure before the Galleon upgrade was already showing this different failure.
Would you be able to instrument your test to dump the content of each file in the console?
I don't have a Windows Box.

My initial thinking is that we could have an encoding issue. Until JDK17 JVM is not defaulting to UTF-8, so somewhere we have the files written to CP-1252.
I checked on the Galleon side (that instantiate the FileWriter and uses UTF-8 as default for Files.newBufferedWriter) so so we should be good. Same encoding ruled by Galleon in the 2 cases. And encoding shouldn't be touched by woodstox (when in the WildFly Maven Plugin).

Another possible cause would be formatting (using tab vs space, something like that). Although we are calling into an API that should abstract us from that.

mchoma:
I believe this MR is ready to be merged. @jfdenise @rhusar @jamezp please review.

I am going offline for a week, so I postponed investigation of Wildfly problem to https://issues.redhat.com/browse/WFLY-21445.

I rebased and there are some new failures, but I think they are not related (please check):

* Linux with security manager - JDK 17 - Some missing class. PR directly not touching this. I cant rule out some upgraded dependencies could affect this. But I do not have such deep insight. So I would ask someone more experienced with CI results if this is safe to ignore for purpose of this PR
* WildFly Preview - Linux - JDK 21 - maven central timeout
* Windows - JDK 17 - flaky test LocalIdleThresholdFineSessionPassivationTestCase

rhusar:
Thanks @mchoma we can just retest.

jfdenise:
@mchoma , I found the root cause. Galleon uses OS dependent line separator : https://github.com/jfdenise/galleon/blob/4bad3812306eda6c51199582d1ae0fc195ee65b4/common-api/src/main/java/org/jboss/galleon/xml/util/FormattingXmlStreamWriter.java#L51

With JDK XML library, no encoding occurs, that doesn't cause any problem. With woodstox, `\r\n` is encoded to be `&#xd;` So each line contains extra characters.

I am wandering if Galleon is right? Shouldn't we use `\n` on all platforms. @jamezp @mchoma , any feedback?

testsuite/shared/src/main/java/org/wildfly/test/distribution/validation/ProvisioningConsistencyBaseTest.java
```diff
@@ -92,6 +94,7 @@ private static void assumeJbossDistIsNotExternallySet() throws IOException {
     @SuppressWarnings("JUnit3StyleTestMethodInJUnit4Class")
     @Test
     public void testInstallationEquivalence() throws IOException {
+        assumeFalse(TestSuiteEnvironment.isWindows()); // Remove once WFLY-21445 fixed
```
rhusar: BTW For future, I would add "TODO" so that it gets pickup by IDE/tooling. But no need to change this now.

mchoma:
@jfdenise  seems to me galleon using just `\n` would help here. But I do not know whole picture and consequences.  According to https://www.w3.org/TR/xml/#sec-line-ends `\r` is adviced to be stripped anyway. And woodstock's `&#xd;` is way to enforce `\r` preservation - apparently from SOAP world.  

But I am afraid what else can be broken elsewhere :sweat_smile: But that is discussion for https://issues.redhat.com/browse/WFLY-21445. Not blocking this PR to be merged.

rhusar:
@mchoma Agreed, I am trying to get this some clean CI runs...

jamezp:
> @mchoma , I found the root cause. Galleon uses OS dependent line separator : https://github.com/jfdenise/galleon/blob/4bad3812306eda6c51199582d1ae0fc195ee65b4/common-api/src/main/java/org/jboss/galleon/xml/util/FormattingXmlStreamWriter.java#L51
> 
> With JDK XML library, no encoding occurs, that doesn't cause any problem. With woodstox, `\r\n` is encoded to be `&#xd;` So each line contains extra characters.
> 
> I am wandering if Galleon is right? Shouldn't we use `\n` on all platforms. @jamezp @mchoma , any feedback?

I've got no strong opinion TBH. I'm starting to think we should close this PR for now. I think we'll need to migrate to 6.0 anyway and we can get these issues sorted. They're all very valid issues.

I've also filed [WFLY-21443](https://issues.redhat.com/browse/WFLY-21443) and [WFMP-328](https://issues.redhat.com/browse/WFMP-328) to move these tests to the wildfly-maven-plugin where they likely should be. These tests seem to just be testing that the two plugins produce the same thing.

martinchoma:
If CI failures are not caused by this PR , I would rather continue with this PR, so we propagate new version to further testing.
That is my main motivation. And we can continue with work on followup jiras..

jamezp:
> If CI failures are not caused by this PR , I would rather continue with this PR, so we propagate new version to further testing. That is my main motivation. And we can continue with work on followup jiras..

That works for me too. I've got no strong opinion either way.

jfdenise:
I agree that merging the upgrade will allow for more testing. I made the https://issues.redhat.com/browse/WFLY-21445 Blocker. The generated XML are very wrong on Windows.

jfdenise:
An interesting finding, is that prospero enforces Galleon to use `\n` on all platforms (it appears that it was broken in latest Galleon GAL-380, I am fixing it).

jfdenise:
@jamezp , I have done the following exclusion, I can see that XML writer is no more leaking in Galleon (so not what you observed).

```
<dependency>
  <groupId>org.codehaus.plexus</groupId>
  <artifactId>plexus-xml</artifactId>
  <version>${version.org.codehaus.plexus-xml}</version>
  <exclusions>
    <exclusion>
      <groupId>com.fasterxml.woodstox</groupId>
      <artifactId>*</artifactId>
    </exclusion>
  </exclusions>
</dependency>
```
but when executing the plugin tests we have exceptions (the classes are expected). Exclusion doesn't feel right, it seems fragile.

I think I have found the right approach. Galleon is not prepared for discovery of XML factory, it relies implicitly on the default one.
I have logged: https://issues.redhat.com/browse/GAL-381 that is going to protect Galleon against discovering another Xml factory.
The fix is straightforward, I will release Galleon 7.0.2.FInal that we would then use in the next Maven Plugin release.

@mchoma @jamezp so in the end we shouldn't merge this PR I think. Wait for the Maven plugin release that will behave properly. 

https://issues.redhat.com/browse/WFLY-21445 should be bound to the Jira that will track the WildFly Maven Plugin upgrade to 6.0.0.Beta1.

jfdenise (changes requested):
As I commented, I don't think that we should merge this PR with an excluded test.

jamezp:
Sounds reasonable to me @jfdenise. Thank you! I was hoping that exclusion would have just been a simple option. We should explore excluding the `org.codehaus.plexus` dependencies too as Maven should include those, but that's for the wildfly-maven-plugin in general.

bstansberry:
I believe WFLY-21445 captures the fundamental problem here, so I'm trying to move the discussion of how to fix it there. 

https://issues.redhat.com/browse/WFLY-21445?focusedId=29018982&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29018982

This PR may or may not end up being part of the fix, so I think it's good to discuss the solution in the JIRA.

mchoma:
I am not sure what fate of this PR is. I have updated PR and JIRA about fact of upgrading to 6.0.0.Beta2.

Locally `ProvisioningConsistencyTestCase` is passing. Lets see what CI thinks. 

@bstansberry @jfdenise feel free to route faith of this PR where you wish.

bstansberry:
@mchoma If you can remove the ProvisioningConsistencyTestCase changes I'll get this merged. The import changes in one of them fails checkstyle. This PR does step #3 in https://issues.redhat.com/browse/WFLY-21445?focusedId=29028102&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-29028102 -- steps 1 and 2 were already done.

Thanks for your patience with this. It's given me a renewed appreciation for ProvisioningConsistencyTestCase.

mchoma:
> @mchoma If you can remove the ProvisioningConsistencyTestCase changes I'll get this merged. The import changes in one of them fails checkstyle. 

Sure, sorry, I was in hurry.  Second attempt ...

> Thanks for your patience with this. It's given me a renewed appreciation for ProvisioningConsistencyTestCase.

At the beginning I thought it will be simple micro upgrade, I haven't imagine what it will trigger :sweat_smile:

rhusar (approved):
LGTM now

rhusar:
Actually, org.wildfly.test.manual.management.ProvisioningConsistencyTestCase.testInstallationEquivalence is still failing..

bstansberry:
Sigh. The galleon-maven-plugin installation has a standalone/tmp dir; the wildfly-maven-plugin one does not.

I know something changed recently re cleaning up a spurious tmp dir. Maybe it was done in the plugin to remove stuff that is done as part of the operation of the plugin, not as part of the basic provisioning. @jfdenise will know.

mchoma:
> Sigh. The galleon-maven-plugin installation has a standalone/tmp dir; the wildfly-maven-plugin one does not.
> 
> I know something changed recently re cleaning up a spurious tmp dir. Maybe it was done in the plugin to remove stuff that is done as part of the operation of the plugin, not as part of the basic provisioning. @jfdenise will know.

Seems we need https://issues.redhat.com/browse/WFLY-21406 to be resolved.

jfdenise:
To have it pass we need:
* https://issues.redhat.com/browse/WFLY-21406 PR opened: https://github.com/wildfly/wildfly/pull/19656 to be merged
* https://issues.redhat.com/browse/WFLY-21481 To be added in your PR @mchoma to upgrade Galleon to 7.0.2.Final. We can't upgrade Galleon to 7.0.2.Final alone, that would create a diff between what the 5.1.0.Final plugin produces and the latest Galleon (the UTF-8 encoding that is now generated by Galleon).

bstansberry:
@jfdenise WFLY-21481 seems to be a duplicate of WFLY-21464. This PR already handles WFLY-21464.

mchoma:
To summarize:
-To merge this PR (have green runs), I need to wait for https://github.com/wildfly/wildfly/pull/19656 to be merged.

Feel free to correct me if I am wrong.

rhusar:
> To summarize: -To merge this PR (have green runs), I need to wait for #19656 to be merged.
> 
> Feel free to correct me if I am wrong.

@mchoma FWIW you can always `git cherry-pick 89e7db625b14084a2748bebba232746b20d91f38`; if that is the commit that ends up getting in `main`, GH will recognize it and your branch will be mergeable witout rebase and you can get clean CI run early.

bstansberry:
I merged #19656 and kicked off a re-test of this

mchoma:
We are green on latest run :partying_face: I believe this PR is ready to be merged.

Failing  JWTIdentityPropagationTestCase.testInvokeEJBWithinSingleDeploymentOutflowNotConfigured is not related to this PR and occures intermittently in history and is covered by https://issues.redhat.com/browse/WFLY-19370

rhusar:
Excellent, yes, that's unrelated, @jfdenise please review/approve.

testsuite/integration/manualmode-expansion/src/test/java/org/wildfly/test/manual/management/ProvisioningConsistencyTestCase.java
```diff
@@ -5,7 +5,6 @@
 
 package org.wildfly.test.manual.management;
 
-
```
jfdenise, line 8: @mchoma those formatting changes should be removed.
JF

rhusar: These aren't necessary but lets keep these - it fixes double space and no line ending at the end of file.

mchoma: As Rado say, I agree they are not necessary, but I left it because I found it correct.

jfdenise (approved):

rhusar:
CI failure unrelated, lets go.

rhusar:
Thanks @mchoma @jfdenise @jamezp @bstansberry !

PR #19577, opened 2026-01-29
[WFLY-21408] Apply workaround for dependabot pom.xml parsing issue

rhusar:
Replace ${project.groupId}.channels with hardcoded org.wildfly.channels to work around dependabot's bug to parse this specific expression correctly.

Manifests as "ERROR: Invalid expression: /project/groupId}.channel"

Jira
https://issues.redhat.com/browse/WFLY-21408 (created a Jira as a placeholder for the context why we are doing it and what needs to happen)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -232,7 +232,9 @@
              By default, if channels drive the provisioning we use the 'wildfly' channel.
              Profiles that focus on wildfly-preview or wildfly-ee can change that.
          -->
-        <channels.maven.groupId>${project.groupId}.channels</channels.maven.groupId>
+        <!-- TODO WFLY-21408 Apply workaround for dependabot pom.xml parsing issue "ERROR: Invalid expression: /project/groupId}.channel" -->
+        <!-- <channels.maven.groupId>${project.groupId}.channels</channels.maven.groupId> -->
+        <channels.maven.groupId>org.wildfly.channels</channels.maven.groupId>
         <channels.maven.artifactId>wildfly</channels.maven.artifactId>
         <channels.maven.version>${full.maven.version}</channels.maven.version>
 


rhusar:
@yersan Would this be OK for you? You would need to include the 'revert' in the downstream repo as discussed on the email thread.

pom.xml
```diff
@@ -232,7 +232,9 @@
              By default, if channels drive the provisioning we use the 'wildfly' channel.
              Profiles that focus on wildfly-preview or wildfly-ee can change that.
          -->
-        <channels.maven.groupId>${project.groupId}.channels</channels.maven.groupId>
+        <!-- TODO WFLY-21408 Apply workaround for dependabot pom.xml parsing issue "ERROR: Invalid expression: /project/groupId}.channel" -->
+        <!-- <channels.maven.groupId>${project.groupId}.channels</channels.maven.groupId> -->
```
yersan, line 236: Feel free even to remove all of this.

We are indeed overwriting this on the conversion branches, hence even making no use of `${project.groupId}` facility to convert this commit.

yersan (approved):
That's fine from the conversion branches' point of view. Thanks for asking @rhusar

pom.xml
rhusar: I would keep it. The next guy refactoring is going to break dependabot without knowing it. 

This way we keep track and revert.

rhusar:
Thanks @yersan I ll wait for @bstansberry for final ack.

bstansberry (approved):

bstansberry:
The Windows failure is https://issues.apache.org/jira/browse/ARTEMIS-5873 / https://issues.redhat.com/browse/WFLY-21414

I went ahead and re-ran it since there were a lot of failures making the result not so useful. Highly unlikely this would break anything though.

bstansberry:
Thanks @rhusar and @yersan

rhusar:
@bstansberry Hm, not sure why dependabot job didn't get run in the last 3 days and this commit doesn't touch the dependabot description so it wasn't treated as dependabot config.

So, I had to trigger the job manually https://github.com/wildfly/wildfly/network/updates to see if this is effective.

PR #19578, opened 2026-01-29
[WFLY-21409] Update org.jboss.as.jpa.container.NonTxEmCloser.popCall to check for null container managed stateless session/entity manager

scottmarlow:
https://issues.redhat.com/browse/WFLY-21409

This is to address a minor (e.g. ignored) NPE that only occurs with TRACE logging enabled for Persistence:

> 2026-01-29 11:02:32,150 TRACE [org.jboss.as.jpa] (default task-2) Could not close (non-transactional) container managed stateless session. This shouldn't impact application functionality (only read operations occur in non-transactional mode): java.lang.NullPointerException: Cannot invoke "java.lang.AutoCloseable.close()" because "statelessSession" is null
> at org.jboss.as.jpa@40.0.0.Beta1-SNAPSHOT//org.jboss.as.jpa.container.NonTxEmCloser.popCall(NonTxEmCloser.java:61)

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/NonTxEmCloser.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/NonTxEmCloser.java
@@ -46,7 +46,7 @@ public class NonTxEmCloser {
             for (ScopedObjects scopedObjects : emStack.values()) {
                 try {
                     EntityManager entityManager = scopedObjects.getEntityManager();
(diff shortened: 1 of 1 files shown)


scottmarlow:
I happened to notice the problem while testing locally with TRACE logging enabled for `org.jboss.as.jpa`.

bstansberry (approved):

scottmarlow:
Thanks for the review Brian!

PR #19579, opened 2026-01-29
[WFLY-20788] Update docs to reflect that dynamic client SSLContext is in default stability

skyllarr:
https://issues.redhat.com/browse/WFLY-20788


kstekovi:
Based on change in the https://github.com/wildfly/wildfly-core/pull/6623 the stability level is not community anymore. it is default.

LGTM.

yersan (approved):

yersan:
EAP7 paperwork is not fully completed yet, so I've added missing-reqs label ... I guess it should not block merging it, but just in case

PR #19580, opened 2026-01-29
[WFLY-21411] Upgrade the Maven wrapper to the latest version.

jamezp:
https://issues.redhat.com/browse/WFLY-21411


bstansberry (approved):

rhusar (approved):
Coincidentally I was also working on this, mainly as another effort which is to remove all binary non-image files from this repository which is on the backburner.

I did not submit it because the way you have done this, there is no integrity checking mandated by the script.  For that, there should be ability to pass `mvn wrapper:wrapper -DdistributionSha256Sum=305773a68d6ddfd413df58c82b3f8050e89778e777f3a745c8e5b8cbea4018e` however, I was not able to get this to verify.

Docs for it: https://maven.apache.org/tools/wrapper/maven-wrapper-plugin/wrapper-mojo.html#distributionSha256Sum

Anyway, approving given it's better to what we have anyway.

bstansberry:
Thanks @jamezp and @rhusar

rhusar:
> Coincidentally I was also working on this, mainly as another effort which is to remove all binary non-image files from this repository which is on the backburner.
> 
> I did not submit it because the way you have done this, there is no integrity checking mandated by the script. For that, there should be ability to pass `mvn wrapper:wrapper -DdistributionSha256Sum=305773a68d6ddfd413df58c82b3f8050e89778e777f3a745c8e5b8cbea4018e` however, I was not able to get this to verify.
> 
> Docs for it: https://maven.apache.org/tools/wrapper/maven-wrapper-plugin/wrapper-mojo.html#distributionSha256Sum
> 
> Anyway, approving given it's better to what we have anyway.

@jamezp This ^^^

What I was trying to do to validate this works was: setup wrapper with the correct sha of the release.

Then, intentionally change the SHA to not match, delete all .m2/wrapper, repo and uninstall maven  => wrapper downloaded maven and did no complaining about sha not matching. Then I ran out of time :)

rhusar:
Opened https://github.com/wildfly/wildfly/pull/19624 to address this.

PR #19581, opened 2026-01-30
[WFLY-21339] Fix ApplicationRegistry to support Builder pattern registry

kstekovi:
https://issues.redhat.com/browse/WFLY-21339

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
@@ -16,13 +16,17 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.FunctionCounter;
 import io.micrometer.core.instrument.FunctionTimer;
+import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Measurement;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.TimeGauge;
 import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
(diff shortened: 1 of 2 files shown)


observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
```diff
@@ -159,6 +163,132 @@ private boolean containsWrongDeploymentTag(Meter.Id id) {
...
+        Gauge.Builder<T> builder = Gauge.builder(taggedId.getName(), obj, valueFunction)
+                .tags(taggedId.getTags());
+        if (taggedId.getDescription() != null) {
+            builder.description(taggedId.getDescription());
+        }
+        if (taggedId.getBaseUnit() != null) {
+            builder.baseUnit(taggedId.getBaseUnit());
+        }
+        return builder.register(parentRegistry);
+    }
```
jasondlee: I think these new methods can be simplified a bit so they're more readable. I don't think the null checks are necessary, and the tag check/manipulation can be inlined, so to speak, to make the code more concise. For example:

```
    private Iterable<Tag> addMissingTag(Meter.Id id) {
        return (id.getTag(TAG_WF_DEPLOYMENT) != null) ? id.getTags() :
                Tags.concat(id.getTags(), TAG_WF_DEPLOYMENT, deploymentName);
    }

    @Override
    protected <T> Gauge newGauge(Meter.Id id, T obj, ToDoubleFunction<T> valueFunction) {
        return Gauge.builder(id.getName(), obj, valueFunction)
                .tags(addMissingTag(id))
                .description(id.getDescription())
                .baseUnit(id.getBaseUnit())
                .register(parentRegistry);
    }
```

jasondlee (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
kstekovi: Good point, implementation improved. Thanks.

jasondlee (approved):
Looks good to me. Thanks for the work on this!

bstansberry:
I kicked off a re-run of the Cloud Tests since the issue that was causing them to always fail looks to be fixed.

bstansberry:
I managed to get the Cloud Tests to run on #19630, which is this on an updated main. Passed.

bstansberry:
Thanks @kstekovi and @jasondlee

PR #19598, opened 2026-02-02
[WFLY-21430] Try to work around Guice injection issue with maven-checkstyle-plugin and plexus-velocity

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-21430


bstansberry:
Thanks @RanabirChakraborty

PR #19602, opened 2026-02-03
[WFLY-21433] Add microprofile-rest-api to lra-participant

marcosgopen:
https://issues.redhat.com/browse/WFLY-21433

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/lra-participant/main/module.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/narayana/lra/lra-participant/main/module.xml
@@ -24,6 +24,10 @@
     <module name="org.jboss.narayana.lra.lra-service-base" />
 
     <!-- lra-client -->
+<!--     org.eclipse.microprofile.restclient stands for microprofile-rest-client-api -->
+    <module name="org.eclipse.microprofile.restclient"/>
+    <module name="org.jboss.resteasy.resteasy-client-microprofile"/>
+    <module name="org.jboss.resteasy.resteasy-client"/>
     <module name="jakarta.enterprise.api"/>
     <!-- Setting the cuncurrent.api optional prevents it being provisioned in a slimmed server whose provisioning config doesn't include the actual Jakarta Concurrency integration -->
     <module name="jakarta.enterprise.concurrent.api" optional="true"/>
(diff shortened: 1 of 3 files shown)


rhusar (changes requested):
According to the description 'The new NarayanaLraClient will use the resteasy microprofile-rest-api implementation so we need to include those dependency into the module.xml and the modules using it.' this change appears that  it should be done in the same change set as the NarayanaLraClient ugrade, not as a separate PR. Or am I missing something?

marcosgopen:
> According to the description 'The new NarayanaLraClient will use the resteasy microprofile-rest-api implementation so we need to include those dependency into the module.xml and the modules using it.' this change appears that it should be done in the same change set as the NarayanaLraClient ugrade, not as a separate PR. Or am I missing something?

Thanks @rhusar for your review. The narayana lra change (https://github.com/jbosstm/lra/pull/211) is going to be merged soon (tested with this change) and I am going to release lra this week. So if you prefer I can wait for the new release to be available and add the lra update to this PR. WDYT?

rhusar:
@marcosgopen IIUC this change does not bring anything, the newly added module dependency cannot be used until the client lib upgrades. BUT should you e.g. delay that change to later release, this change would be unnecessary - so this makes this PR not mergeable on it's own.

You can make this a draft so people can have a look and then update it, but I don't see much value in that either.

marcosgopen:
> @marcosgopen IIUC this change does not bring anything, the newly added module dependency cannot be used until the client lib upgrades. BUT should you e.g. delay that change to later release, this change would be unnecessary - so this makes this PR not mergeable on it's own.
> 
> You can make this a draft so people can have a look and then update it, but I don't see much value in that either.

I agree @rhusar , as this change to WildFly is needed by lra integration tests I wanted to make sure this PR is 'approved' as the new lra release will depend on it. As now I don't need this change to be merged, but I wanted to make sure the change is fine and I can proceed with the lra release as I have already tested lra main with it.
I see your point and I can close this PR to reopen it with the lra update. My only concern is that I am going to test the new lra release with this change so I'd like to be sure this change will be in WildFly together with the new lra release. Does it make sense? Thanks

rhusar:
You are starting to lose me now, basically a PR cannot be approved if it cannot be merged.

Basically if the idea is to test this only on it's own, you can just use draft for that (done that for you now).

marcosgopen:
Making it a draft is fine. Thanks @rhusar

marcosgopen:
Closing in favor of https://github.com/wildfly/wildfly/pull/19621

PR #19606, opened 2026-02-03
WFLY-21437 Disable FILE_PING and its subclasses' register_shutdown_hook as its leaking resources

rhusar:
Resolves two issues – shutdown hook leak https://issues.redhat.com/browse/JGRP-2976 workaround and the fact that these should never be registered in WF https://issues.redhat.com/browse/WFLY-21437

WFLY-21437 Disable FILE_PING and its subclasses' register_shutdown_hook as its leaking resources

WildFly already handles JVM shutdown gracefully so it should be default disable register_shutdown_hook which defaults to true in vanilla JGroups.

This impl is problematic as it cases memory leak.

Follow up on JGRP-2976:

FILE_PING registers a JVM shutdown hook in init() but doesn't keep a reference to it, so the hook cannot be removed when the protocol stops.

The shutdown hook to remain registered even after the channel is closed
Memory leaks when channels are repeatedly created and destroyed - Thread is anonymous class and keeps reference to the stack via up/down protocol references
Errors when closing gracefully and then shutdown hook operates on closed resources
This also affects subclasses such as JDBC_PING, S3_PING, etc.

Also submitted a fix to upstream https://github.com/belaban/JGroups/pull/986

### DIFF
--- a/clustering/jgroups/extension/src/main/resources/jgroups-defaults.xml
+++ b/clustering/jgroups/extension/src/main/resources/jgroups-defaults.xml
@@ -27,6 +27,16 @@
         port_range="0"
         use_virtual_threads="false"
     />
+    <!-- Disable register_shutdown_hook for all *loadable* FILE_PING-based discovery protocols -->
+    <FILE_PING
+        register_shutdown_hook="false"
+    />
+    <JDBC_PING
+        register_shutdown_hook="false"
+    />
+    <JDBC_PING2
+        register_shutdown_hook="false"
+    />
     <TCPPING
         port_range="0"
         num_discovery_runs="3"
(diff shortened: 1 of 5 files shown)


rhusar:
@pferraro please review; I don't like it but the other option is even worse which is to pollute subsystem class path with the external protocols...

pferraro (approved):
I would suggest disabling shutdown hook registration by default in upstream S3/AZURE_PING protocol implementations.

rhusar:
> I would suggest disabling shutdown hook registration by default in upstream S3/AZURE_PING protocol implementations.

Agreed, already [merged](https://github.com/jgroups-extras/jgroups-aws/pull/544) into jgroups-aws main branch (release imminent)and jgroups-azure is wip. This will be done within WF 40, so this PR can go as is now.

bstansberry:
Thanks @rhusar and @pferraro

PR #19611, opened 2026-02-04
WFLY-21370 TLS support to JGroups TCP-based transports test cases fai…

rhusar:
…l when running on a container that does not support 'community' stability

Resolves
https://issues.redhat.com/browse/WFLY-21370


rhusar:
@yersan Please review if this fixes your issue with downstream.

n.b. only rebased since I can't access teamcity on monday morning :)

rhusar:
@yersan Kind reminder after 2 weeks to review this. There are other WFLY issues you created for the same pattern I want to make sure  this works for you before I open fixes for others. Thanks!

yersan (approved):

PR #19633, opened 2026-02-08
WFLY-21456 Stabilize CachingFilesystemRealmTestCase on Windows

rhusar:
https://issues.redhat.com/browse/WFLY-21456

The test case makes assumption that FS lock is released by the WF process on reload and that the file is deletable, which is not a reliable assumption on Windows OS.


darranl (approved):
Thank you @rhusar

bstansberry (approved):
This is a good short term workaround until we fix the path the test sends to the server. (And it's not wrong for the long run.)

rhusar:
Thanks @darranl and @bstansberry 

Merging so we can merge 39.x https://github.com/wildfly/wildfly/pull/19634 in anticipation of the core upgrade.

PR #19637, opened 2026-02-09
[WFLY-16047]: Allow a shared single connection in the Resource Adaptor.

ehsavoie:
Simple testing of the number of connections for a single MDB with maxSessions=15 compared between shared-connection=true and false.

Issue: https://issues.redhat.com/browse/WFLY-16047

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ServerManagementOperationsTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ServerManagementOperationsTestCase.java
@@ -1,0 +1,397 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.test.integration.messaging.mgmt;
+
+import org.jboss.as.test.integration.messaging.jms.deployment.MultipleConnectionsQueueMDB;
+
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.INCLUDE_DEFAULTS;
+import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.INCLUDE_RUNTIME;
+import static org.jboss.shrinkwrap.api.ShrinkWrap.create;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import jakarta.json.Json;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonObject;
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ServerManagementOperationsTestCase.java
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    @Test
+    public void testListConnectionsAsJsonDetailed() throws Exception {
+        deployer.deploy("MULTIPLE");
+        try {
+            // Wait for the MDB pool to initialize
+            // MDB instances are created on-demand as messages arrive
+            // The pool creates instances up to the configured size (default 15)
+            Thread.sleep(2000);
```
mnovak: Instead of sleep() it will better to add condition (to wait for 15 connections being created) with timeout.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ServerManagementOperationsTestCase.java
```diff
@@ -0,0 +1,422 @@
...
+            deployer.undeploy("MULTIPLE");
+        }
+    }
+
+    @Test
+    public void testSingleListConnectionsAsJsonDetailed() throws Exception {
+        deployer.deploy("SHARED");
+        try {
+            // Wait a bit for the MDB to initialize and create connections
+            Thread.sleep(1000);
```
mnovak: Same here.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/mgmt/ServerManagementOperationsTestCase.java
```diff
@@ -0,0 +1,422 @@
...
+            int connectionCount = countOccurrences(jsonString, "\"connectionID\"");
+
+            // MultipleConnectionsQueueMDB is configured with maxSession=15
+            // This creates 15 concurrent sessions, each with its own connection
+            // Additionally, 1 connection may be created during MDB initialization
+            // Validate we have at least 15 remoting connections (typically 15 or 16)
+            assertTrue("MDB with maxSession=15 should create at least 15 connections, found: " + connectionCount,
+                    connectionCount >= 15);
+            assertTrue("Connection count should not exceed 16 (15 sessions + 1 initialization), found: " + connectionCount,
+                    connectionCount <= 16);
```
mnovak: I'm afraid this assert `<=16` might be unstable. Problem is with random connections created for example by recovery manager which are short lived and might break it randomly. Ideally we would like to identify the connections created by MDB but I guess there are no a metadata on connection to do it.

ehsavoie:
Superseded by https://github.com/wildfly/wildfly/pull/19680

PR #19642, opened 2026-02-10
[WFLY-21413] promote (WFLY-16532 additional scope for auth request) f…

rsearls:
…rom preview to default

https://issues.redhat.com/browse/WFLY-21413

____

> [!WARNING]

> * [WFLY-16532](https://issues.redhat.com/browse/WFLY-16532)


elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -91,7 +91,6 @@ public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<Elyt
 
     VERSION_1_0(1, Stability.DEFAULT),
     VERSION_2_0(2, Stability.DEFAULT),
-    VERSION_2_0_PREVIEW(2, 0, Stability.PREVIEW), // WildFly 32
```
pferraro, line 94: Never delete schema versions, unless explicitly dropping support for very old releases.  This will otherwise break any migrated user configuration based on this version.

elytron-oidc-client/src/main/java/org/wildfly/extension/elytron/oidc/ElytronOidcSubsystemSchema.java
```diff
@@ -91,7 +91,6 @@ public enum ElytronOidcSubsystemSchema implements PersistentSubsystemSchema<Elyt
...
     VERSION_3_0_PREVIEW(3, 0, Stability.PREVIEW), // WildFly 33-present
     ;
```
pferraro, line 95: I think you mean to create a new VERSION_3_0 version?

elytron-oidc-client/src/main/resources/schema/wildfly-elytron-oidc-client_2_0.xsd
```diff
@@ -1194,6 +1194,16 @@
                     </xs:documentation>
                 </xs:annotation>
             </xs:element>
+            <xs:element name="scope" type="xs:string" minOccurs="0" maxOccurs="1">
```
pferraro, line 1197: We should never modify a XSD for released versions (unless it is an explicit correction).
Rather, you want to create a new `urn:wildfly:elytron-oidc-client:3.0` schema.

pferraro (changes requested):

rhusar (changes requested):
Personally I would block all development in this subsystem UNTIL the subsystem testing is actually in place – https://issues.redhat.com/browse/WFLY-21298

cc @bstansberry

PR #19652, opened 2026-02-13
WFLY-21381 - Add support for SECURE_PROTOCOL tag to ExchangeAttributeDefinitions

jasondlee:
https://redhat.atlassian.net/browse/WFLY-21381

Add new exchange attribute
Update test

____

> [!WARNING]

> * [WFLY-21381](https://issues.redhat.com/browse/WFLY-21381)

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/ExchangeAttributeDefinitions.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/ExchangeAttributeDefinitions.java
@@ -52,6 +52,7 @@ import io.undertow.attribute.ResponseHeaderAttribute;
 import io.undertow.attribute.ResponseReasonPhraseAttribute;
 import io.undertow.attribute.ResponseTimeAttribute;
 import io.undertow.attribute.SecureExchangeAttribute;
+import io.undertow.attribute.SecureProtocolAttribute;
 import io.undertow.attribute.SslCipherAttribute;
 import io.undertow.attribute.SslClientCertAttribute;
 import io.undertow.attribute.SslSessionIdAttribute;
@@ -536,6 +537,16 @@ class ExchangeAttributeDefinitions {
                 }
             });
 
+    private static final SimpleAttributeDefinition SECURE_PROTOCOL_KEY = createKey("secureProtocol");
+    public static final ObjectTypeAttributeDefinition SECURE_PROTOCOL = create(
+            ObjectTypeAttributeDefinition.create("secure-protocol", SECURE_PROTOCOL_KEY),
+            new ExceptionBiFunction<>() {
+                @Override
+                public Collection<AccessLogAttribute> apply(final OperationContext context, final ModelNode model) throws OperationFailedException {
+                    return createSingleton(SECURE_PROTOCOL_KEY, context, model, SecureProtocolAttribute.INSTANCE);
+                }
+            });
+
     private static final SimpleAttributeDefinition SSL_CIPHER_KEY = createKey("sslCipher");
     private static final ObjectTypeAttributeDefinition SSL_CIPHER = create(
             ObjectTypeAttributeDefinition.create("ssl-cipher", SSL_CIPHER_KEY),
@@ -626,6 +637,7 @@ class ExchangeAttributeDefinitions {
             RESPONSE_TIME,
             SECURE_EXCHANGE,
             SSL_CIPHER,
+            SECURE_PROTOCOL,
             SSL_CLIENT_CERT,
             SSL_SESSION_ID,
             STORED_RESPONSE,
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
@@ -6,6 +6,7 @@
 package org.wildfly.extension.undertow;
 
 import java.util.EnumSet;
+import java.util.Map;
 import java.util.Set;
 
 import org.jboss.as.controller.ModelVersion;
@@ -66,6 +67,15 @@ public class UndertowExtensionTransformerRegistration implements ExtensionTransf
                 ajpListenerAttributeTransformationDescriptionBuilder.setDiscard(DiscardAttributeChecker.UNDEFINED, AjpListenerResourceDefinition.ALLOWED_REQUEST_ATTRIBUTES_PATTERN)
                 .addRejectCheck(RejectAttributeChecker.DEFINED, AjpListenerResourceDefinition.ALLOWED_REQUEST_ATTRIBUTES_PATTERN)
                 .end();
+
+                RejectAttributeChecker.ObjectFieldsRejectAttributeChecker attributeChecker = new RejectAttributeChecker.ObjectFieldsRejectAttributeChecker(
+                    Map.of(ExchangeAttributeDefinitions.SECURE_PROTOCOL.getName(), RejectAttributeChecker.DEFINED)
+                );
+                server.addChildResource(HostDefinition.PATH_ELEMENT)
+                    .addChildResource(ConsoleAccessLogDefinition.PATH_ELEMENT)
+                    .getAttributeBuilder()
+                    .addRejectCheck(attributeChecker, ExchangeAttributeDefinitions.ATTRIBUTES)
+                ;
 
                 if (UndertowSubsystemModel.VERSION_13_0_0.requiresTransformation(version)) {
                     final ResourceTransformationDescriptionBuilder servletContainer = subsystem.addChildResource(ServletContainerDefinition.PATH_ELEMENT);
--- a/undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
+++ b/undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
@@ -149,6 +149,7 @@ undertow.console-access-log.attributes.response-reason-phrase=The text reason fo
 undertow.console-access-log.attributes.response-time=The time used to process the request.
 undertow.console-access-log.attributes.time-unit=The unit of time used for the response time.
 undertow.console-access-log.attributes.secure-exchange=Indicates whether or not the exchange was secure.
+undertow.console-access-log.attributes.secure-protocol=The name of the secure protocol used for the request.
 undertow.console-access-log.attributes.ssl-cipher=The SSL cipher.
 undertow.console-access-log.attributes.ssl-client-cert=The SSL client certificate.
 undertow.console-access-log.attributes.ssl-session-id=The SSL session id.
--- a/undertow/src/test/java/org/wildfly/extension/undertow/UndertowSubsystemTransformerTestCase.java
+++ b/undertow/src/test/java/org/wildfly/extension/undertow/UndertowSubsystemTransformerTestCase.java
@@ -142,10 +142,14 @@ public class UndertowSubsystemTransformerTestCase extends AbstractSubsystemTest 
         PathAddress subsystemAddress = PathAddress.pathAddress(UndertowRootDefinition.PATH_ELEMENT);
 
         if (UndertowSubsystemModel.VERSION_15_0_0.requiresTransformation(this.modelVersion)) {
-            PathAddress ajpListenerAddress = subsystemAddress.append(PathElement.pathElement(ServerDefinition.PATH_ELEMENT.getKey(), "default-server"))
-                    .append(PathElement.pathElement(AjpListenerResourceDefinition.PATH_ELEMENT.getKey(), "ajp"));
+            PathAddress serverAddress = subsystemAddress.append(PathElement.pathElement(ServerDefinition.PATH_ELEMENT.getKey(), "default-server"));
 
+            PathAddress ajpListenerAddress = serverAddress.append(PathElement.pathElement(AjpListenerResourceDefinition.PATH_ELEMENT.getKey(), "ajp"));
             config.addFailedAttribute(ajpListenerAddress, new FailedOperationTransformationConfig.NewAttributesConfig(AjpListenerResourceDefinition.ALLOWED_REQUEST_ATTRIBUTES_PATTERN));
+
+            PathAddress consoleAccessLogAddress = serverAddress.append(PathElement.pathElement(HostDefinition.PATH_ELEMENT.getKey(), "default-host"))
+                .append(PathElement.pathElement(ConsoleAccessLogDefinition.PATH_ELEMENT.getKey(), "console-access-log"));
+            config.addFailedAttribute(consoleAccessLogAddress, new SecureProtocolAttributeConfig(ExchangeAttributeDefinitions.ATTRIBUTES.getName()));
         }
         if (UndertowSubsystemModel.VERSION_13_0_0.requiresTransformation(this.modelVersion)) {
             PathAddress servletContainerAddress = subsystemAddress.append(PathElement.pathElement(ServletContainerDefinition.PATH_ELEMENT.getKey(), "rejected-container"));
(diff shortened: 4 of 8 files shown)


bstansberry:
I know this is a draft, but please keep separate commits for the Undertow upgrade vs the new feature. The odds are very high that we are going to want to deal with the upgrade before we are able to deal with the feature. If you keep them separate we can likely fork the upgrade part out into a separate branch, which will be very helpful.

Also, we are almost certainly going to get version 15 of the undertow subsystem stuff coming in soon via https://github.com/wildfly/wildfly/pull/19542. So be prepared to rebase.

jasondlee:
> I know this is a draft, but please keep separate commits for the Undertow upgrade vs the new feature. The odds are very high that we are going to want to deal with the upgrade before we are able to deal with the feature. If you keep them separate we can likely fork the upgrade part out into a separate branch, which will be very helpful.
> 
> Also, we are almost certainly going to get version 15 of the undertow subsystem stuff coming in soon via #19542. So be prepared to rebase.

Once we have a proper integration of the Undertow upgrade, I plan to rebase my PR on top of that, so all of those changes will go away in the final PR.

jasondlee:
@fl4via @bstansberry @ropalka This should be ready for review now.

bstansberry (changes requested):
@jasondlee This needs:

1) Update to the xsd.
2) Inclusion in the src/test/resources/undertow-15.0.xml file.
3) Inclusion in UndertowExtensionTransformerRegistration and UndertowSubsystemTransformerTestCase

The last one will be a bit tricky as it's a new field in an existing attribute. Probably that's been done somewhere before.

jasondlee:
> @jasondlee This needs:
> 
>     1. Update to the xsd.
> 
>     2. Inclusion in the src/test/resources/undertow-15.0.xml file.
> 
>     3. Inclusion in UndertowExtensionTransformerRegistration and UndertowSubsystemTransformerTestCase
> 
> 
> The last one will be a bit tricky as it's a new field in an existing attribute. Probably that's been done somewhere before.

Ah, shoot. I had at least the schema changes done, but I overwrote it when I updated to James' PR. I'll get that fixed tomorrow.

undertow/src/test/resources/org/wildfly/extension/undertow/undertow-15.0.xml
```diff
@@ -37,6 +37,7 @@
                </request-header>
                <response-code/>
                <response-time time-unit="MICROSECONDS"/>
+               <secure-protocol/>
```
rhusar, line 40: Is the omission of `key=".."` attribute intentional? In general, the test files are supposed to have fully-defined model.

jasondlee: That looks like `authentication-type` above it. The schema definition, the test entry, and the `ObjectTypeAttributeDefinition` all look structurally similar to me. I tried to emulate what was already happening in the code, so it's at least consistent, but should be correct.

rhusar (changes requested):
Looks like a stray commit made it here 834193f2493cd69d6cf231a95394afa18267b115 - please squash! :)

undertow/src/test/resources/org/wildfly/extension/undertow/undertow-15.0.xml
rhusar: I see, but I find that also wrong, so being consistent with wrong approach is suboptimal ;) 

Basically there might then no testing coverage for parsing that attribute.

jasondlee: I can't argue with that, but I don't understand this part of the code base to "fix" this, so I will be consistent with the rest of the code.

jasondlee:
> Looks like a stray commit made it here [834193f](https://github.com/wildfly/wildfly/commit/834193f2493cd69d6cf231a95394afa18267b115) - please squash! :)

There are more changes coming (especially for testing), so this will get squashed before the final PR state.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
```diff
@@ -56,6 +57,18 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
...
 
+
+                ////////////////////////////////////////////////////////////////////////////////////////////////////////
+                RejectAttributeChecker.ObjectFieldsRejectAttributeChecker attributeChecker = new RejectAttributeChecker.ObjectFieldsRejectAttributeChecker(
+                        Map.of(ExchangeAttributeDefinitions.SECURE_PROTOCOL.getName(), RejectAttributeChecker.DEFINED)
+                );
+                server.addChildResource(HostDefinition.PATH_ELEMENT)
+                        .addChildResource(ConsoleAccessLogDefinition.PATH_ELEMENT)
+                        .getAttributeBuilder()
+                        .addRejectCheck(attributeChecker, ExchangeAttributeDefinitions.SECURE_PROTOCOL)
```
kabir: Shouldn't this be for the attribute? i.e. 

````
.addRejectCheck(attributeChecker, ExchangeAttributeDefinitions.ATTRIBUTES)
````

Comment (jasondlee) [reply undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java]: Done.

**Comment by jasondlee** [comment]
@bstansberry I think I have addressed all of your concerns. I know it's a busy week, but if you could take another look, we'd really like to get this into WF 40. :)

File: undertow/src/test/resources/org/wildfly/extension/undertow/undertow-transform-reject.xml [hunk 3]
```diff
@@ -7,7 +7,13 @@
     <byte-buffer-pool name="default" thread-local-cache-size="45" buffer-size="1000" direct="false" leak-detection-percent="50" max-pool-size="1000"/>
     <server name="default-server" default-host="default-host">
         <ajp-listener name="ajp" allowed-request-attributes-pattern="test" socket-binding="ajp"/>
-        <host name="default-host"/>
+        <host name="default-host">
+            <console-access-log >
```
rhusar: Nitpick: run a formatter (extra space)

undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
```diff
@@ -56,6 +57,18 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
                 final AttributeTransformationDescriptionBuilder reverseProxyAttributeTransformationDescriptionBuilder = reverseProxy.getAttributeBuilder();
                 final ResourceTransformationDescriptionBuilder ajpListener = server.addChildResource(AjpListenerResourceDefinition.PATH_ELEMENT);
 
+
+                ////////////////////////////////////////////////////////////////////////////////////////////////////////
```
rhusar: Nitpick: stray comment separator?

undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
```diff
@@ -56,6 +57,18 @@ public void registerTransformers(SubsystemTransformerRegistration registration)
...
+                RejectAttributeChecker.ObjectFieldsRejectAttributeChecker attributeChecker = new RejectAttributeChecker.ObjectFieldsRejectAttributeChecker(
+                        Map.of(ExchangeAttributeDefinitions.SECURE_PROTOCOL.getName(), RejectAttributeChecker.DEFINED)
+                );
+                server.addChildResource(HostDefinition.PATH_ELEMENT)
+                        .addChildResource(ConsoleAccessLogDefinition.PATH_ELEMENT)
+                        .getAttributeBuilder()
+                        .addRejectCheck(attributeChecker, ExchangeAttributeDefinitions.ATTRIBUTES)
+                        ;
+                ////////////////////////////////////////////////////////////////////////////////////////////////////////
```
rhusar: Nitpick: stray comment separator?

undertow/src/test/resources/org/wildfly/extension/undertow/undertow-15.0.xml
rhusar: If you just add the key attribute and tests pass, it will be correct.

```
               <secure-protocol key="abc"/>
```

rhusar (changes requested):

fl4via (approved):

undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
bstansberry: Quasi-nitpick: This whole block should either move above L55 or below L81. Right now it's unrelated stuff inserted in the middle of a chunk of related code. L55-58 creates some vars and the L71-88 uses those vars.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowExtensionTransformerRegistration.java
jasondlee: Whoops. I was using that to block off my changes visually so I could spot them more easily. Deletion incoming.

undertow/src/main/resources/org/wildfly/extension/undertow/LocalDescriptions.properties
```diff
@@ -149,6 +149,7 @@ undertow.console-access-log.attributes.response-reason-phrase=The text reason fo
 undertow.console-access-log.attributes.response-time=The time used to process the request.
 undertow.console-access-log.attributes.time-unit=The unit of time used for the response time.
 undertow.console-access-log.attributes.secure-exchange=Indicates whether or not the exchange was secure.
+undertow.console-access-log.attributes.secure-protocol=The secure protocol.
```
bstansberry: What does this mean?

AIUI this ends up being the value of javax.net.ssl.SSLSession.getProtocol() which is javadoced as being 'the standard name of the protocol used for all connections in the session'. That's still not super clear but the class javadoc says ' Sessions are created, or rejoined, as part of the SSL handshaking protocol' so from my non-expert POV I think this is 'The standard name of the SSL handshaking protocol used for the request'. Probably without 'standard' for brevity, perhaps without 'handshaking' for brevity and maybe replacing SSL with TLS or something.

But someone who knows better should confirm.

undertow/src/test/java/org/wildfly/extension/undertow/UndertowSubsystemTransformerTestCase.java
```diff
@@ -142,10 +142,14 @@ public void testRejections() throws Exception {
...
-                    .append(PathElement.pathElement(AjpListenerResourceDefinition.PATH_ELEMENT.getKey(), "ajp"));
+            PathAddress serverAddress = subsystemAddress.append(PathElement.pathElement(ServerDefinition.PATH_ELEMENT.getKey(), "default-server"));
 
+            PathAddress ajpListenerAddress = serverAddress.append(PathElement.pathElement(AjpListenerResourceDefinition.PATH_ELEMENT.getKey(), "ajp"));
             config.addFailedAttribute(ajpListenerAddress, new FailedOperationTransformationConfig.NewAttributesConfig(AjpListenerResourceDefinition.ALLOWED_REQUEST_ATTRIBUTES_PATTERN));
+
+            PathAddress consoleAccessLogAddress = serverAddress.append(PathElement.pathElement(HostDefinition.PATH_ELEMENT.getKey(), "default-host"))
+                .append(PathElement.pathElement(ConsoleAccessLogDefinition.PATH_ELEMENT.getKey(), "console-access-log"));
+            config.addFailedAttribute(consoleAccessLogAddress, new FailedOperationTransformationConfig.NewAttributesConfig(ExchangeAttributeDefinitions.ATTRIBUTES));
```
kabir: I don't think this is the right FailedOperationTransformationConfig implementation. 
I could be wrong,  but I think this will see an 'attributes' entry, and remove the whole thing. 

Instead you should use an implementation that inspects the value of the 'attributes' entry, and fails if 'secure-protocol' is set, and removes 'secure-protocol' in the correct step.

FWIW FailedOperationTransformationConfig.AttributesPathAddressConfig has a configureComplexAttribute() method which looks related, but no examples.
My guess would be to register an implementation of AttributesPathAddressConfig for the 'attributes' attribute, and then to call configureComplexAttribute() with an instance of NewAttributesConfig for the 'secure-protocol' entry

kabir (changes requested):

bstansberry (changes requested):
Sorry; I thought I submitted this yesterday.

Re the key=abc thing in the test config file, I'm fine with dealing with that later. That file needs a broad overhaul anyway as it should be using expressions most everywhere.

undertow/src/test/java/org/wildfly/extension/undertow/UndertowSubsystemTransformerTestCase.java
jasondlee: @kabir @bstansberry 

I think the transformer and the test are correct as they are. If I modify the test XML, adding a known valid exchange attribute, but removing `<secure-protocol/>`:

```
            <console-access-log>
                <attributes>
                    <bytes-sent/>
                </attributes>
            </console-access-log>
```

the test fails, as it was expecting a rejection, which didn't happen because the attributes are all acceptable. If I add `<secure-protocol/>` back and leave `<bytes-sent/>`, the test passes as expected. And, of course, if it's _only_ `<secure-protocol/>`, it passes with the expected rejection. Additionally, if I add the same snippet above to a transform test file that should not fail (`src/test/resources/org/wildfly/extension/undertow/undertow-transform.xml`), the transform test continues to pass as expected.

Long story short:

- With only known bad: test passes
- With only known good: test fails, expecting a rejection
- With known good _and_ known bad, test passes

So unless my verification logic is faulty, this seems to be correct.

bstansberry: @jasondlee Please file a JIRA to follow up on this.

AIUI the issue here is not what is in the test xml. It's in the sophistication of how the test setup 'repairs' the flaw and then validates that the repair works.

What you have is a setup that simply removes the offending attribute. The entire attribute.

Better would be something that fixes the offending attribute by removing the offending field -- secure-protocol -- while leaving the rest of the attribute in place (so you'd want the xml to have some other field too.

That's a more sophisticated use of the FailedOperationTransformationConfig stuff. It's the right thing to do but let's not block this PR on doing it.

jasondlee: I filed this, targeting 40.0.0.Final: https://redhat.atlassian.net/browse/WFLY-21703

undertow/src/test/resources/org/wildfly/extension/undertow/undertow-15.0.xml
bstansberry: Follow up: https://redhat.atlassian.net/browse/WFLY-21704

bstansberry (approved):

undertow/src/main/java/org/wildfly/extension/undertow/ExchangeAttributeDefinitions.java
```diff
@@ -626,6 +637,7 @@ public Collection<AccessLogAttribute> apply(final OperationContext context, fina
             RESPONSE_TIME,
             SECURE_EXCHANGE,
             SSL_CIPHER,
+            SECURE_PROTOCOL,
```
rhusar, line 640: Don't fix it now, only if you are going to push again - but this list is meant to be kept sorted alphabetically and this violates that patttern. Move it one above.

rhusar (approved):
Bunch of follow up work but this appears to me functionally usable 👍🏼

fl4via (approved):

PR #19653, opened 2026-02-13
[WFLY-21469] ContextProxyInvocationHandler should unwrap InvocationTa…

emmartins:
…rgetException

Issue: https://issues.redhat.com/browse/WFLY-21469

Note: this fix is in Concurro 3.1, see https://github.com/eclipse-ee4j/glassfish-concurro/blob/3.1.x/src/main/java/org/glassfish/concurro/internal/ContextProxyInvocationHandler.java#L84 , so no need to add it to wildfly-concurrency-impl-31

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/ContextualProxyExceptionUnwrapBean.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/ContextualProxyExceptionUnwrapBean.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.ee.concurrent;
+
+import static org.junit.Assert.fail;
+
+import jakarta.annotation.Resource;
+import jakarta.ejb.Stateless;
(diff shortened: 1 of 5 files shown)


bstansberry (approved):

bstansberry (changes requested):
The glow discovery stuff in testsuite/integration/basic/pom.xml will need to be updated to fix the CI failure with slimmed servers.

I don't know if using @EJB instead of @Inject would remove the need for that.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ee/concurrent/ContextualProxyExceptionUnwrapBean.java
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.integration.ee.concurrent;
+
+import jakarta.annotation.Resource;
+import jakarta.ejb.Stateless;
+import jakarta.enterprise.concurrent.ContextService;
+
+import static org.junit.Assert.fail;
```
bstansberry: import static goes first

emmartins:
retest please

bstansberry (approved):

bstansberry:
Thanks @emmartins

PR #19654, opened 2026-02-14
[WFLY-21477] Update WildFly Preview to Hibernate ORM 7.2.4.Final

bstansberry:
https://issues.redhat.com/browse/WFLY-21477

Needed to unblock WFLY-21476 which in turn will be needed to test any fix for WFLY-21461.


bstansberry:
Note that I looked at the deps of this ORM version as well as things that might depend on it (e.g. Hibernate Search) for the ones that we typically update to align. There are no changes to the relevant ORM deps. Hibernate Search has an micro update but it depends on something earlier than the 7.2.21 we are already using, so updating Search is valid but out of scope for this PR.

scottmarlow (approved):

scottmarlow:
Thank you Brian!

PR #19658, opened 2026-02-16
[WFLY-21482] Restore deploying of org.wildfly:wildfly-testsuite-share…

bstansberry:
…d so the testsuite can be executed against deployed artifacts

https://issues.redhat.com/browse/WFLY-21482

### DIFF
--- a/testsuite/shared/pom.xml
+++ b/testsuite/shared/pom.xml
@@ -24,13 +24,9 @@
     <packaging>jar</packaging>
 
     <name>WildFly Test Suite: Shared</name>
-
-    <properties>
-        <!-- Don't try to deploy the module -->
-        <maven.deploy.skip>true</maven.deploy.skip>
-        <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
-        <skipPublishing>${maven.deploy.skip}</skipPublishing>
-    </properties>
+    <description>
+        Test utilities for shared used across testsuite modules in the main WildFly repo. Not for use elsewhere.
+    </description>
 
     <dependencyManagement>
         <dependencies>


bstansberry:
@TomasHofman @darranl @ehsavoie FYI

TomasHofman:
Hi @bstansberry , I was talking to Ruben that we probably don't need it for the use case I was thinking - we align the deps and rebuild the sources anyway, so we don't need to have it deployed. Sorry for the noise. Not sure if anybody else came up with other reason..

bstansberry:
Thanks @TomasHofman -- I think it's good to deploy anyway. The basic idea is the testsuite can be executed from a maven repo, not requiring building the rest. A reasonable expectation is that repo would be cached output from some earlier build, where 'install' would probably be enough, but restricting to that seems like an unnecessary limitation.

The main downside is it lets people use the artifact for other purposes, but that was possible for 15+ years before this was changed last year and never caused issues. And if anyone did that and complained we then broke them we'd just say that's not our problem.

TomasHofman:
Good, let's have it back then. Thanks!

testsuite/shared/pom.xml
```diff
@@ -24,13 +24,9 @@
     <packaging>jar</packaging>
 
     <name>WildFly Test Suite: Shared</name>
-
-    <properties>
-        <!-- Don't try to deploy the module -->
-        <maven.deploy.skip>true</maven.deploy.skip>
```
jbliznak, line 30: do this work?

there is the same property inherited from parent module https://github.com/wildfly/wildfly/blob/main/testsuite/pom.xml#L169

jbliznak: ah sorry, this module doesn't inherit from testsuite/pom.xml but directly from root pom.xml

rhusar: @jbliznak The `skipPublishing` is what matters to the central plugin. The `skipNexusStagingDeployMojo` is what matters to the nexus plugin (we are no longer using). `maven.deploy.skip` is picked up by the maven default deploy plugin(s).

This is our triple way of saying we really do not want this deployed!

bstansberry: Yeah, when we moved to deploying to direct to Maven Central we were scrambling to release 39 Beta and did not spend energy removing no-longer-relevant JBoss Nexus stuff.

I just filed https://issues.redhat.com/browse/WFLY-21497

I forget whether maven.deploy.skip was needed. It seems like something that could be declared in the root pom as we don't want the maven-deploy-plugin doing anything in any module.

jamezp (approved):

rhusar (approved):

bstansberry:
Thanks @jamezp and @rhusar

PR #19664, opened 2026-02-17
WFLY-19370 Fix HTTP client resource leak in JWTIdentityPropagationTes…

rhusar:
…tCase

The CloseableHttpClient was created as an instance field but never closed, causing connection pool, socket, and thread leaks across test runs. Fixed by useing try-with-resources to properly close both the client and response.

Probably won't fix https://issues.redhat.com/browse/WFLY-19370, but at least lets cleanup the basic resource leaks

IIRC ARQ will create test class instance per test method, so 7 times the http client field is created leaking connections, connection pool, file descriptors..


rsearls (approved):

bstansberry:
Thanks @rhusar and @rsearls

PR #19666, opened 2026-02-17
WFLY-21480 - Warning WFLYMMTREXT0015 in log when registering micrometer Timer using Builder pattern

jasondlee:
https://issues.redhat.com/browse/WFLY-21480

The MeterRegistry.Config class does not expose any problematic APIs, so we need not warn about its use in ApplicationRegstiry. Removing log message.

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/registry/ApplicationRegistry.java
@@ -139,7 +139,11 @@ public class ApplicationRegistry extends SimpleMeterRegistry {
 
     @Override
     public Config config() {
-        MicrometerExtensionLogger.MICROMETER_LOGGER.configNotSupported();
+        // Micrometer calls this internally, but if application code does, the returned Config cannot be used normally
+        // since this isn't the real MeterRegistry
+        if (!StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE).getCallerClass().getName().startsWith("io.micrometer")) {
+            MicrometerExtensionLogger.MICROMETER_LOGGER.configNotSupported();
+        }
         return super.config();
     }
 


bstansberry:
The returned Config is useless, yes?  Callers can't use it to affect the operation of the underlying registry (which is good) but they also can't use it to learn anything about how the registry will behave. In the Timer.Builder.register case from the bug report it will call registry.config().pauseDetector() and pass NoPauseDetector.INSTANCE to registry.timer, regardless of what the real registry's Config was using.

Maybe -- for reads -- this is ok because the real registry's Config is just using the default values, so it doesn't matter.

For writes we'd just be silently ignoring the write with no indication to the user that we're doing something unexpected.

jasondlee:
There is also `Config.pauseDetector(PauseDetector detector)` that allows the caller to set that value, should the user wish to use that. 

Looking at both `OtlpMeterRegistry` and `PrometheusMeterRegistry` and their associated `Config` defaults, they are the same as this in `SimpleMeterRegistry`, which `ApplicationRegistry` extends, so the behavior is the same across the three `MeterRegistry` types in play in WildFly. By default, no `PauseDetector` or `HighCardinalityTagsDetector` are configured by default. These two properties have default values set in `Config`, and I see no instances where the setters are called, so this doesn't seem to be a concern for WildFly. Have I missed something that concerns you?

bstansberry:
Ok, so for reads this is (currently) fine as the Config associated with the ApplicationRegistry provides the same default values as the real one. (If we ever changed how we configure the real one that would no longer be true.)

If the user invokes any of the methods that mutate the returned Config they are probably not going to see the expected effect, because the real registry doesn't use that Config.

I have an idea; give me a couple minutes.

bstansberry:
@jasondlee How about this?

https://github.com/bstansberry/wildfly/commit/afa4869f7bf1c7c437258a79483641d1ecfe8ec0

jasondlee:
> @jasondlee How about this?
> 
> [bstansberry@afa4869](https://github.com/bstansberry/wildfly/commit/afa4869f7bf1c7c437258a79483641d1ecfe8ec0)

Ha! Nice. I'll do that.

bstansberry:
Thanks @jasondlee

PR #19655, opened 2026-02-14
[WFLY-21476] Enable JakartaDataEarTestCase testing with WildFly Preview

bstansberry:
https://issues.redhat.com/browse/WFLY-21476

Builds on #19654

PR #19667, opened 2026-02-17
[WFLY-21461] WildFly's StatelessSession implementation should implement all public interfaces implemented by Hibernate's StatelessSessionImpl

bstansberry:
https://issues.redhat.com/browse/WFLY-21461

Builds on #19655

Explanation:

In WFLY-21267 I implemented in WildFly an implementation of Hibernate's StatelessSession interface, one that can be injected into components that have a lifespan longer than a single request or transaction. This impl would delegate calls to an underlying Hibernate StatelessSessionImpl instance choosing the correct instance based on the transaction associated with the calling thread or on the thread itself. This avoids different threads mixing up internal state of the StatelessSessionImpl instance (which isn't completely stateless.)

WFLY-21461 shows a weakness with this approach, in that, at least in Hibernate 7.1.1 and later, for some call paths ORM casts the StatelessSession instance to other public interfaces the Hibernate provides, and that StatelessSessionImpl implements. The specific one involved with the bug being SharedSessionContractImplementor.

I initially fixed this by having WF's StatelessSession implementation also implement SharedSessionContractImplementor, but that was quite ugly. That interface declares dozens of methods. The implementation of them was simple enough -- just get the delegate StatelessSessionImpl instance and invoke on it, much like was done with the StatelessSession methods. But it felt very fragile.

WildFly's impl in all cases delegates to the Hibernate session, except for two methods, close and unwrap, where we want out own behavior, and then the basic Object methods. So I decided instead to reflectively analyze all the public interfaces implemented by StatelessSessionImpl and generate a proxy to provide them. Instead of implementing each method in a WF class, the WildFly class will be an InvocationHandler that handles specially close and unwrap and the Object methods, and reflectively calls on the delegate session for all the rest.

This simplified things a bit with respect to the differences between ORM 6 and 7 as it allowed the same InvocationHandler impl to be used for both cases. Now the difference in our integration of the two is limited to code that analyzes StatelessSessionImpl and determines what interfaces the proxy must implement.

I admit that this is still somewhat unclean, in that WF is analyzing an ORM implementation class to decide what interfaces to expose. It's possible this is not necessary. The WFLY-21461 bug does not affect our ORM 6.6 integration, so perhaps not analyzing StatelessSessionImpl and simply hard coding the proxy to implement StatelessSession was enough there. In ORM 7 (at least in 7.2) there is a new interface 'StatelessSessionImplementor' that combines StatelessSession and SharedSessionContractImplementor, so perhaps perhaps not analyzing StatelessSessionImpl and simply hard coding the proxy to implement StatelessSessionImplementor is sufficient there. I'm interested in @yrodiere 's thoughts here. A benefit of doing the reflective analysis is it somewhat future-proofs this against other changes in ORM.

Note: beyond the initial Jakarta Data problem that led to WFLY-21267 and this WF impl, my thinking was much of this code could be the foundation for WF support for a proper API for injecting StatelessSession or for injecting Jakarta Persistence 4 analogues to Stateless Session. The injection mechanism to solve the  WFLY-21267 problem is very much a workaround, but the core solution of having a shared object that delegates properly based on tx or thread is just another use of the infrastructure WF already uses for EntityManager.


bstansberry:
I converted this to draft while I try the Jakarta Data TCK and resolve any problems.

yrodiere:
> In ORM 7 (at least in 7.2) there is a new interface 'StatelessSessionImplementor' that combines StatelessSession and SharedSessionContractImplementor, so perhaps perhaps not analyzing StatelessSessionImpl and simply hard coding the proxy to implement `StatelessSessionImplementor`

I would be wary of anything that exposes internal API/SPI to users. `StatelessSessionImplementor` certainly has plenty of that.

> So I decided instead to reflectively analyze all the public interfaces implemented by `StatelessSessionImpl` and generate a proxy to provide them.

Why `StatelessSessionImpl` and not just `StatelessSession` and superinterfaces? That's [what we do in Quarkus](https://github.com/quarkusio/quarkus/blob/dcd0a4ca45e1dc5102a36765f20917b7e83e4893/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/session/TransactionScopedStatelessSession.java#L46). I'm not saying the implementation in Quarkus is good (it's awful, see below) but considering only the public interfaces makes sense.

`StatelessSessionImpl` probably has a few methods that, while public, are not meant to be exposed to users.

Now, I do understand the desire to not list and implement all these methods by hand... it's become a maintenance nightmare in Quarkus, and we have [plans to move away from that](https://github.com/quarkusio/quarkus/issues/47796). Perhaps we should coordinate our efforts? The idea in Quarkus was that Hibernate ORM really should be exposing a "delegating implementation" for all each public API entrypoints (`Session`, `StatelessSession`, ...), so that integrators can easily implement "proxies" -- not necessarily using Java proxies.

There is already a few such implementations: `org.hibernate.engine.spi.SessionLazyDelegator`, `org.hibernate.engine.spi.SessionDelegatorBaseImpl`, but they have flaws, and AFAICS there is none for statelessSession. There should be IMO.

yrodiere:
Looking at https://issues.redhat.com/browse/WFLY-21461 ... If you want to unwrap a `StatelessSessionImplementor`, IMO the unwrapping should return the real, transaction-scoped `StatelessSession` instance. That's what I would have expected, at least. I wouldn't expect a `StatelessSessionImplementor` delegator to be necessary.

scottmarlow:
> There is already a few such implementations: org.hibernate.engine.spi.SessionLazyDelegator, org.hibernate.engine.spi.SessionDelegatorBaseImpl, but they have flaws, and AFAICS there is none for statelessSession. There should be IMO.

Should we create a HHH issue or perhaps discuss further on Hibernate Zulip?

yrodiere:
> > There is already a few such implementations: org.hibernate.engine.spi.SessionLazyDelegator, org.hibernate.engine.spi.SessionDelegatorBaseImpl, but they have flaws, and AFAICS there is none for statelessSession. There should be IMO.
> 
> Should we create a HHH issue or perhaps discuss further on Hibernate Zulip?

https://hibernate.atlassian.net/browse/HHH-17990

Happy to discuss on Zulip!

bstansberry:
> Looking at https://issues.redhat.com/browse/WFLY-21461 ... If you want to unwrap a `StatelessSessionImplementor`, IMO the unwrapping should return the real, transaction-scoped `StatelessSession` instance. That's what I would have expected, at least. I wouldn't expect a `StatelessSessionImplementor` delegator to be necessary.

First, thanks for this; it might result in a much simpler solution. In the existing WF code the unwrap method is using the SharedSessionContract default impl of unwrap, instead of getting the delegate and invoking unwrap on it. Replacing that with an invocation on the delegate would remove the need to be able to cast the wrapper to SharedSessionContractImplementor. Relying on default method implementations was a mistake on my part in the WFLY-21267 work. (If I'd have gone with the proxy approach I use in this PR I'd have avoided that.)

The problem with doing that is unwrap is part of the StatelessSession interface, so any user code with a ref to the wrapper can call it and they would get a ref to the delegate. If they only use that ref within the appropriate call context that's ok but if they cache it somewhere where it gets reused across transactions, threads etc we are back to the original WFLY-21267 problem.

In the Jakarta Data context where this is currently being used this is probably ok. The user code in question is generated by Hibernate Data Repositories and that generated code doesn't expose the StatelessSession outside of the repository impl (except to ORM code via the stack.)

bstansberry:
> > > There is already a few such implementations: org.hibernate.engine.spi.SessionLazyDelegator, org.hibernate.engine.spi.SessionDelegatorBaseImpl, but they have flaws, and AFAICS there is none for statelessSession. There should be IMO.
> > 
> > 
> > Should we create a HHH issue or perhaps discuss further on Hibernate Zulip?
> 
> https://hibernate.atlassian.net/browse/HHH-17990
> 
> Happy to discuss on Zulip!

Great!  In a couple of the classes I deleted in this PR I'd written 'TODO: a 'DelegatingStatelessSession' in Hibernate would be nice.' Indeed it would.

bstansberry:
I added a commit limiting the proxy to exposing StatelessSession and allowing unwrap to invoke on and thus return the delegate. I left the old code in the invocation handler, commented out with added comments why its not used and when that might need to be reconsidered.

yrodiere:
> The problem with doing that is unwrap is part of the StatelessSession interface, so any user code with a ref to the wrapper can call it and they would get a ref to the delegate. If they only use that ref within the appropriate call context that's ok but if they cache it somewhere where it gets reused across transactions, threads etc we are back to the original WFLY-21267 problem.
> 
> In the Jakarta Data context where this is currently being used this is probably ok. The user code in question is generated by Hibernate Data Repositories and that generated code doesn't expose the StatelessSession outside of the repository impl (except to ORM code via the stack.)

To clarify, I'd expect your delegator/proxy to return `this` when `.unwrap(Session.class)` or `.unwrap(StatelessSession.class)` is called -- basically when any *API* interface is passed. That's what we do in `org.hibernate.engine.spi.SessionLazyDelegator#unwrap`. Only in cases where callers request a more complex type (e.g. session implementor) would you forward the call to the delegate. I *think* that would be a reasonable, and relatively safe, compromise?

bstansberry:
@yrodiere That sounds good; I've updated it to work that way.

If user code passes in some internal class to unwrap... well, they are hacking and if things don't work as they expect they should stop hacking.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    private Object getDelegate() {
+        return delegateSupplier.get();
+    }
+
+    private <T> T unwrap(Object proxy, Class<T> type) {
+        // This is a copy of the ORM 7 StatelessSessionImpl unwrap, except we unwrap the proxy and use an i18n exception message
+        checkOpen();
```
yrodiere: Does this make sense if you're unwrapping the proxy? The delegate might very well not be open, if using the proxy across transactions...

Also, checking whether the session is open is not very critical here:

1. You're effectively only allowing a cast to `StatelessSession` or superinterface, which is what the proxy already exposes.
2. Any other operation the user might call afterwards will likely also call `checkOpen` on the delegate.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
```diff
@@ -0,0 +1,150 @@
...
+
+        if ( type.isInstance( proxy ) ) {
```
yrodiere: AFAIU, this will always be true, because you check `((Class<?>) args[0]).isAssignableFrom(statelessSessionClass)` before calling `unwrap`.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
bstansberry: +1. My original thinking in this method of "do the same ORM does as they know best" doesn't really fit.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
bstansberry: Yes, it is redudant now.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
bstansberry: Fixed.

jpa/spi/src/main/java/org/jipijapa/plugin/spi/ScopedStatelessSessionInvocationHandler.java
bstansberry: Fixed.

yrodiere (approved):

scottmarlow (approved):

scottmarlow:
Thanks Brian!

scottmarlow:
Thanks @yrodiere!

PR #19668, opened 2026-02-17
WFLY-21498 Olympus WildFly testsuite running issue fix/work around

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-21498

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/license/licenses.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/license/licenses.xml
@@ -66,5 +66,29 @@
                 </license>
             </licenses>
         </dependency>
+        <dependency>
+            <groupId>jakarta.jms</groupId>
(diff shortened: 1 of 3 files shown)


RanabirChakraborty:
@bstansberry @jamezp Can you please review the PR?

bstansberry (changes requested):
I don't know what Olympus is but this problem does not affect an actual WildFly build and I'm not inclined to pollute our source with workarounds for some outside system.

The poms for these artifacts are included in the artifact jars themselves and properly declare what their license is. If there is a problem related to them my assumption is its either in the license-plugin or in whatever system is producing the problem, not in the WildFly code.

rhusar:
What is Olympus (apart from extensive massif near the Thermaic Gulf of the Aegean Sea)?

bstansberry:
Sorry; I don't see the need for this.

PR #19677, opened 2026-02-19
Use IntermittentFailure util to suppress frequently failing Clustere…

bstansberry:
…dJPA2LCTestCase


rhusar (changes requested):
This is surprisingly wrong. Looks like the events are ordered in such way that this will fail to connect to the un-started H2 before the Assumption bubbles up.

rhusar:
Closing – replaced by https://github.com/wildfly/wildfly/pull/19679

PR #19678, opened 2026-02-20
[WFLY-21508] Override the inherited organization setting and the mail…

jamezp:
…ing list setting.

https://issues.redhat.com/browse/WFLY-21508

The `mailingList` directing to the forum, I wasn't sure about. I couldn't get lists.jboss.org to come up so I could get a good link to the real mailing list. That said, on https://wildfly.org the only links are for the Google Group. If we want the mailing lists to be something different, I can change that.

WildFly Core PR: https://github.com/wildfly/wildfly-core/pull/6677

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,11 @@
     <description>WildFly: Parent Aggregator</description>
     <url>https://wildfly.org</url>
 
+    <organization>
+        <name>WildFly</name>
+        <url>https://wildfly.org</url>
+    </organization>
+
     <licenses>
         <license>
             <name>Apache License 2.0</name>
@@ -50,6 +55,15 @@
             <organizationUrl>https://wildfly.org</organizationUrl>
         </developer>
     </developers>
+
+    <mailingLists>
+        <mailingList>
+            <name>WildFly</name>
+            <subscribe>https://groups.google.com/forum/#!forum/wildfly</subscribe>
+            <unsubscribe>https://groups.google.com/forum/#!forum/wildfly</unsubscribe>
+            <archive>https://groups.google.com/forum/#!forum/wildfly</archive>
+        </mailingList>
+    </mailingLists>
 
     <properties>
         <!-- Release Info.
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
 
     <mailingLists>
         <mailingList>
-            <name>WildFly</name>
+            <name>WildFly User Forum</name>
             <subscribe>https://groups.google.com/forum/#!forum/wildfly</subscribe>
             <unsubscribe>https://groups.google.com/forum/#!forum/wildfly</unsubscribe>
             <archive>https://groups.google.com/forum/#!forum/wildfly</archive>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -28,6 +28,11 @@
     <description>WildFly: Parent Aggregator</description>
     <url>https://wildfly.org</url>
 
+    <organization>
+        <name>WildFly</name>
```
rhusar, line 32: We typically refer to this as community.

```suggestion
        <name>WildFly Community</name>
```

e.g. https://github.com/quarkusio/quarkus/blob/main/pom.xml#L25

pom.xml
```diff
@@ -51,6 +56,15 @@
         </developer>
     </developers>
 
+    <mailingLists>
+        <mailingList>
+            <name>WildFly</name>
```
rhusar: Can we explicitly say that this is the user mailing list so that we differentiate from the dev list?

rhusar (changes requested):

pom.xml
jamezp: We didn't want this here as it didn't feel right for the organization. It makes sense for the developer though which is why it's there.

pom.xml
jamezp: We could make a comment. TBH I don't think this will ever really come into play given we don't use the `maven-site-plugin` to generate a site.

pom.xml
rhusar: OK! Resolving

pom.xml
bstansberry:
```suggestion
            <name>WildFly User Forum</name>
```

pom.xml
```diff
@@ -51,6 +56,15 @@
...
+            <subscribe>https://groups.google.com/forum/#!forum/wildfly</subscribe>
+            <unsubscribe>https://groups.google.com/forum/#!forum/wildfly</unsubscribe>
+            <archive>https://groups.google.com/forum/#!forum/wildfly</archive>
+        </mailingList>
```
bstansberry, line 65:
```suggestion
        </mailingList>
        <mailingList>
            <name>WildFly Development</name>
            <subscribe>https://lists.jboss.org/admin/lists/wildfly-dev.lists.jboss.org/</subscribe>
            <unsubscribe>https://lists.jboss.org/admin/lists/wildfly-dev.lists.jboss.org/</unsubscribe>
            <archive>https://lists.jboss.org/archives/list/wildfly-dev@lists.jboss.org/</archive>
        </mailingList>
```

bstansberry (changes requested):
@jamezp We should include both lists.

jamezp:
This was my first time accepting edits and I realized too late that if you look at them in the Files Changed tab you can group them. If desired, I can pull this back down and sqash everything to one commit.

bstansberry (approved):
@jamezp I don't see much point in squashing.

pom.xml
rhusar: Fixed; resolving

rhusar (approved):
LGTM now

rhusar:
CI failures unrelated

rhusar:
Thanks @jamezp and @bstansberry

PR #19692, opened 2026-02-23
[WFLY-21515] Comment out unused JipiLogger methods

bstansberry:
https://issues.redhat.com/browse/WFLY-21515

### DIFF
--- a/jpa/spi/src/main/java/org/jipijapa/JipiLogger.java
+++ b/jpa/spi/src/main/java/org/jipijapa/JipiLogger.java
@@ -5,15 +5,11 @@
 
 package org.jipijapa;
 
-import static org.jboss.logging.Logger.Level.WARN;
 
 import java.lang.invoke.MethodHandles;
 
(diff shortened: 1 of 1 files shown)


jpa/spi/src/main/java/org/jipijapa/JipiLogger.java
```diff
@@ -30,84 +26,84 @@ public interface JipiLogger extends BasicLogger {
     JipiLogger JPA_LOGGER = Logger.getMessageLogger(MethodHandles.lookup(), JipiLogger.class, "org.jipijapa");
```
bstansberry, line 26: Note that this is still used, which is why I didn't delete the interface altogether.

rhusar (approved):
LGTM; org.jipijapa.JipiLogger#JPA_LOGGER is still used, class stays.

rhusar:
CI failure unrelated, sadly, JWTIdentityPropagationTestCase continues to be flaky.

Thanks @bstansberry

PR #19697, opened 2026-02-24
[WFLY-21521]: Upgrade Apache Artemis to 2.52.0

ehsavoie:
* Upgrading Artemis version
 * Changing the groupId from org.apache.activemq to org.apache.artemis

Issue: https://issues.redhat.com/browse/WFLY-21521

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -1637,172 +1637,84 @@
                 <version>${version.org.apache.activemq.artemis.native}</version>
                 <exclusions>
                     <exclusion>
-                        <groupId>org.jboss.logging</groupId>
-                        <artifactId>jboss-logging</artifactId>
-                    </exclusion>
-                    <exclusion>
-                        <groupId>org.jboss.logmanager</groupId>
-                        <artifactId>jboss-logmanager</artifactId>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
                     </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.apache.activemq</groupId>
+                <groupId>org.apache.artemis</groupId>
                 <artifactId>artemis-amqp-protocol</artifactId>
                 <version>${version.org.apache.activemq.artemis}</version>
                 <exclusions>
                     <exclusion>
-                        <groupId>io.netty</groupId>
+                        <groupId>*</groupId>
                         <artifactId>*</artifactId>
                     </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.apache.activemq</groupId>
+                <groupId>org.apache.artemis</groupId>
                 <artifactId>artemis-cli</artifactId>
                 <version>${version.org.apache.activemq.artemis}</version>
                 <exclusions>
                     <exclusion>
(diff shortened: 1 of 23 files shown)


boms/common-ee/pom.xml
```diff
@@ -1952,10 +1952,18 @@
...
                     </exclusion>
+                    <exclusion>
+                        <groupId>commons-logging</groupId>
+                        <artifactId>commons-logging</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.apache.commons</groupId>
+                        <artifactId>commons-dbcp2</artifactId>
+                    </exclusion>
                 </exclusions>
```
rhusar, line 1803: Pardon my ignorance, but just wondering why is this enumerating all exclusions rather than excluding everything like other artifacts do?

```xml
                <exclusions>
                    <exclusion>
                        <groupId>*</groupId>
                        <artifactId>*</artifactId>
                    </exclusion>
                </exclusions>

File: boms/common-ee/pom.xml [hunk 2]
```diff
@@ -1701,7 +1701,7 @@
...
             </dependency>
             <dependency>
-                <groupId>org.apache.activemq</groupId>
+                <groupId>org.apache.artemis</groupId>
```
pferraro, line 1668: Why not just import org.apache.artemis:artemis-bom?

ehsavoie: Good point, I would say because it didn't exist back then ;)

boms/common-ee/pom.xml
ehsavoie: I remember issues with maven then. Let's try that now

rhusar: I have never heard or seen issues with * excludes like this. I would be surprised if there is still an issue here.

boms/common-ee/pom.xml
rhusar: Just make sure the artemis-bom is correct and only manages what it should manage, as it's a common problem we have seen out there in the wild (and have done numerous fixes in our own projects..)

ehsavoie: Yes, it brings transitive dependencies

bstansberry:
@ehsavoie  This doesn't compile.

Likely the problem is testsuite/integration/basic needs a dep on JGroups added because your (good) move to exclude transitives eliminated the path by which it was being added to the classpath.

rhusar (changes requested):
java.lang.NoClassDefFoundError: org/apache/activemq/artemis/utils/sm/SecurityManagerShim

rhusar (changes requested):
java.lang.IllegalStateException: java.lang.NoClassDefFoundError: org/apache/commons/beanutils/FluentPropertyBeanIntrospector

Have you considered running a smoke test before pushing the changes? :-)If you are using upstream CI for testing, consider converting this to a draft so it doesn't appear on the mergers and reviewers list.

ehsavoie:
I don't run the full test suite locally, only the basic and one could argue that the initial PR didn't have those issues ;)

bstansberry (approved):

bstansberry:
Thanks @ehsavoie and @rhusar

PR #19698, opened 2026-02-24
WFLY-21473 Promote "TLS support to JGroups TCP-based transport" to 'default' stability

rhusar:
Jira
https://redhat.atlassian.net/browse/WFLY-21473
https://redhat.atlassian.net/browse/EAP7-1865

Proposal
https://github.com/wildfly/wildfly-proposals/pull/807 (original feature proposal was https://github.com/wildfly/wildfly-proposals/pull/552)

____

> [!WARNING]

> * [WFLY-15836](https://redhat.atlassian.net/browse/WFLY-15836)


pferraro (approved):

bstansberry (approved):

marekkopecky:
Current implementation allows to set `/subsystem=elytron/client-ssl-context=jgroupsCSC` to `/subsystem=jgroups/stack=tcp/transport=TCP - server-ssl-context` and vice versa. Is it possible / reasonable to restrict the values more?

rhusar:
> Current implementation allows to set `/subsystem=elytron/client-ssl-context=jgroupsCSC` to `/subsystem=jgroups/stack=tcp/transport=TCP - server-ssl-context` and vice versa. Is it possible / reasonable to restrict the values more?

@marekkopecky Both client and server ssl-context register the same capability `org.wildfly.extension.elytron.Capabilities#SSL_CONTEXT_RUNTIME_CAPABILITY` of the same service value `javax.net.ssl.SSLContext`. There is unfortunately no way to differentiate them as that's how this was designed. Perhaps @darranl or @skyllarr could elaborate why the design decision was done this way – because IMO this is a design bug cc @bstansberry

marekkopecky:
I'm looking to the differences between JGroups xsd files between version 9 and 10. How are the following changes related with TLS RFE please?

* maxOccurs=1 limit in default-thread-pool attribute has been removed
* min-threads, max-threads and keepalive-time attributes are no longer optional

rhusar:
> I'm looking to the differences between JGroups xsd files between version 9 and 10. How are the following changes related with TLS RFE please?
> 
> * maxOccurs=1 limit in default-thread-pool attribute has been removed
> * min-threads, max-threads and keepalive-time attributes are no longer optional

Not true. Optional and maxOccurs=1 are the defaults. Thus the cleanup to reduce confusion.

It's part of the PR given that the new schema must be introduced in this RFE. So, it is introduced without the old issues.

That's described by this commit message https://github.com/wildfly/wildfly/pull/19698/changes/0dd87f7691a599f4aabcf8136a53fbe0f1a30be9

rhusar:
> > Current implementation allows to set `/subsystem=elytron/client-ssl-context=jgroupsCSC` to `/subsystem=jgroups/stack=tcp/transport=TCP - server-ssl-context` and vice versa. Is it possible / reasonable to restrict the values more?
> 
> @marekkopecky Both client and server ssl-context register the same capability `org.wildfly.extension.elytron.Capabilities#SSL_CONTEXT_RUNTIME_CAPABILITY` of the same service value `javax.net.ssl.SSLContext`. There is unfortunately no way to differentiate them as that's how this was designed. Perhaps @darranl or @skyllarr could elaborate why the design decision was done this way – because IMO this is a design bug cc @bstansberry

We have opened a tracker for this https://redhat.atlassian.net/browse/WFCORE-7538

marekkopecky:
Thank you for your feedback!

marekkopecky:
I'm looking to TLSTCP_NIO2CommandDispatcherTestCase. It contains `TLSServerSetupTasks.SharedStoreSecureJGroupsTCP_NIO2TransportServerSetupTask_NODE_1_2.class` server setup task, that uses shared keys, location should be relative to jboss.server.config.dir - `../../../server.keystore.pkcs12` and `../../../server.truststore.pkcs12`. AFAICS from process parameters - `-Djboss.server.config.dir=/path/to/wf/testsuite/integration/clustering/target/wildfly-1/standalone/configuration` - key/trust stores should be located in target directory, but those key/trust stores are not there. 

Is it possible please that `TLSServerSetupTasks.SharedPhysicalKeyStoresServerSetupTask.class` server setup task is missing in `TLSTCP_NIO2CommandDispatcherTestCase`?

I'm also thinking why server doesn't throw the error if key store and trust stores are missing, WDYT?

Or is it possible that I'm missing something and SharedPhysicalKeyStoresServerSetupTask should not be there?

rhusar:
@marekkopecky Well spotted, this is a legitimate issue that uncovers a bigger problem that TCP_NIO2 is not reasonably supportable (requires wrapping of low level objects and directly interact with the SSLEngine). I am investigating this since yesterday, let me stick a hold on this.

n.b. the core TCP transport is unaffected by this.

rhusar:
I have added a new commit that clarifies and correctly restricts TLS support to JGroups TCP transport only as uncovered by @marekkopecky.

TLSManagedSocketFactory only overrides createSocket()/createServerSocket() (used by TCP), not createSocketChannel()/createServerSocketChannel() (used by TCP_NIO2).
TLS cannot work with TCP_NIO2 without SSLEngine-based NIO channel wrapping which does not exist in JGroups - tracked by a separate Jira.

After some ENG discussions, I have remodeled this a little bit:

- Extract SecurableSocketTransportResourceDefinitionRegistrar (TCP) from SocketTransportResourceDefinitionRegistrar (TCP_NIO2), so SSL attributes (client-ssl-context, server-ssl-context) only appear on 'securable' TCP
- Add ssl-context as optional child element in the 10.0 transport XSD type (XSD 1.0 cannot enforce per-type-attribute constraints)
- Add transformer rejection for SSL attributes against older model versions
- Rework TLS TCP_NIO2 test case as a plain TCP_NIO2 test (without TLS as per original Jira for this)

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemSchema.java
```diff
@@ -188,6 +189,9 @@ private ResourceXMLChoice transportChoice() {
         NamedResourceRegistrationXMLChoice.Builder builder = this.factory.namedElementChoice(transportElement);
 
         if (JGroupsSubsystemSchema.this.since(VERSION_7_0)) {
+            for (SecurableSocketTransportResourceDefinitionRegistrar.Transport transport : EnumSet.allOf(SecurableSocketTransportResourceDefinitionRegistrar.Transport.class)) {
+                builder.addElement(this.transportBuilder(transport).withElementLocalName(ResourceXMLElementLocalName.KEY).addAttribute(SocketTransportResourceDefinitionRegistrar.CLIENT_SOCKET_BINDING).build());
+            }
```
pferraro: DRY.  The loop below is identical.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemSchema.java
```diff
@@ -241,10 +245,10 @@ public ModelNode apply(ModelNode operation) {
             });
         }
 
-        if (this.since(JGroupsSubsystemSchema.VERSION_9_0_COMMUNITY)) {
+        if ((this.since(JGroupsSubsystemSchema.VERSION_10_0) || this.since(JGroupsSubsystemSchema.VERSION_9_0_COMMUNITY)) && registration instanceof SecurableSocketTransportResourceDefinitionRegistrar.Transport) {
```
pferraro: Requiring an instanceof test like this usually indicates that your method signature is too generic, and should be further parameterised.  In this case, consider adding a parameter containing a list of child elements.

pferraro (approved):
Functionally, this looks good.  Comments are only stylistic suggestions.

rhusar:
Rebased.

marekkopecky:
Hello, I implemented a few more tests according to the test plan in this commit: https://github.com/marekkopecky/wildfly/commit/b87b1d524fd6c3b885d6232958506bbbe78cb167

TLSTCPUniqueKeysCommandDispatcherTestCase testcase uses this scenario:

- two nodes
- each node has its own private key in keystore, but has oposite node's public key in trust store

This test is passing locally on my Fedora, but is failing intermittently (65% - quite often) on windows machines - cluster has not been established.

I see this suspicious warnings in node1 server logs:

```
2026-04-13 09:37:31,199 WARN  [org.jgroups.protocols.TCP] (TcpServer.Acceptor[7600]-1,null,null) JGRP000006: 127.0.0.1:7600: failed accepting connection from peer SSLSocket[hostname=127.0.0.1, port=50014, Session(1776073035214|SSL_NULL_WITH_NULL_NULL)]: java.net.SocketException: Socket is closed
2026-04-13 09:37:31,563 WARN  [org.jgroups.protocols.TCP] (TcpServer.Acceptor[7600]-1,null,null) JGRP000006: 127.0.0.1:7600: failed accepting connection from peer SSLSocket[hostname=127.0.0.1, port=50017, Session(1776073051199|SSL_NULL_WITH_NULL_NULL)]: java.net.SocketException: Socket is closed
2026-04-13 09:37:32,227 WARN  [org.jgroups.protocols.TCP] (TcpServer.Acceptor[7600]-1,null,null) JGRP000006: 127.0.0.1:7600: failed accepting connection from peer SSLSocket[hostname=127.0.0.1, port=50020, Session(1776073051563|SSL_NULL_WITH_NULL_NULL)]: java.net.SocketException: Socket is closed
```

@rhusar Can you please check it? Feel free to ping me if you need more logs, etc.

rhusar:
> each node has its own private key in keystore, but has oposite node's public key in trust store

That's a good addition to have for a test. Do you intend to add the test upstream?

> This test is passing locally on my Fedora, 

Excellent!

> but is failing intermittently (65% - quite often) on windows machines - cluster has not been established.

@marekkopecky Where does it fail like that? Is that on our Teamcity CI machines? IIRC we have been seeing quite a bit of those in our tests that look like problems in virtualization and how the connections are being handled in the virtualized environment. Are you able to reproduce that on a non-virtualized Windows box?

marekkopecky:
Thank you for your feedback!

> Do you intend to add the test upstream?

Yes, I can create follow-up MR once this PR will be merged, or you can add it to this PR directly - but after the new tests will be stable.

> Are you able to reproduce that on a non-virtualized Windows box?

Yes, I can see this issue on RHEL machines as well, although it's not so common as it is on windows - ~20% on RHEL. I don't see this issue at all on your TCP_NIO2CommandDispatcherTestCase without TLS, but the issue is quite common on TLS issues with keys, which seems to be suspicious, WDYT?

marekkopecky:
I noticed that default TCP stack has `<socket-protocol type="MPING" socket-binding="jgroups-mping"/>`, but TS configuration is using `<socket-discovery-protocol type="TCPPING" socket-bindings="node-1 node-2 node-3 node-4"/>`, could be SocketException caused by TCPPING layer?

rhusar:
> I noticed that default TCP stack has `<socket-protocol type="MPING" socket-binding="jgroups-mping"/>`, but TS configuration is using `<socket-discovery-protocol type="TCPPING" socket-bindings="node-1 node-2 node-3 node-4"/>`, could be SocketException caused by TCPPING layer?

@marekkopecky That is intentional, ironically, we made that change for the same reason as above – debugging IPv6 multicast on localhost in virtualized Windows OS – is simply not a valid use of anyones time, so we switched to static discovery using TCPPING. It never opens sockets directly. The actual TCP connections are created by the TCP transport at the bottom of the stack. So, I don't see how this could cause the problem.

rhusar:
> Yes, I can create follow-up MR once this PR will be merged, or you can add it to this PR directly - but after the new tests will be stable.

Normally I would add it, but given the potential CI instability, let's do this as post-merge item (you can open WFWIP if you want).

marekkopecky:
>, let's do this as post-merge item 

Sure, follow-up PR is ok for me. Anyway, as I mentioned above, unstable tests suggests that there is some bug in the code, so the bug needs to be fixed anyway first.

rhusar:
@marekkopecky OK I have been investigating and I see the problem now – the problem is not TCPPING but FD_SOCK2 which uses NIO and thus hits the same problem as we discussed with supporting TCP_NIO2 transport.

We are evaluating a couple of options how to solve this (model constraint, auto-protocol downgrade, implementing secure channel), I ll move this back to draft for now.

rhusar:
> @marekkopecky OK I have been investigating and I see the problem now – the problem is not TCPPING but FD_SOCK2 which uses NIO and thus hits the same problem as we discussed with supporting TCP_NIO2 transport.
> 
> We are evaluating a couple of options how to solve this (model constraint, auto-protocol downgrade, implementing secure channel), I ll move this back to draft for now.

@marekkopecky In the meantime, Paul reminded me that I should tell you to continue testing with TCP + FD_SOCK (not FD_SOCK2 which use NioServer) so that you aren't unnecessarily blocked while we resolve the FD_SOCK2/NIO situation.

marekkopecky:
Thank you for your investigation! I confirm that FD_SOCK (following server setup task) stabilized my tests and workarounded the issue. Please let me know once proper fix will be available.

```
public class FdSockSetupTask extends ManagementServerSetupTask {
    public FdSockSetupTask() {
        super(createContainerSetConfigurationBuilder()
                .addContainers(NODE_1_2, createContainerConfigurationBuilder()
                        .setupScript(createScriptBuilder()
                                .add("/subsystem=jgroups/stack=tcp/protocol=FD_SOCK2:remove")
                                .add("/subsystem=jgroups/stack=tcp/protocol=FD_SOCK:add(socket-binding=jgroups-tcp-fd, add-index=3)")
                                .build())
                        .build())
                .build());
    }
}
```

rhusar:
Just a quick update: the fix for the blocking issue in JGroups has been merged and release should be available today. In the next push, I will add the JGRoups 5.5.5 upgrade and the NIO TLS-based channel implementation. Just a reminder, the core TCP transport (already tested by QE) remains unchanged; it will only affect TCP_NIO2 and FD_SOCK2.

Moreover, as @marekkopecky pointed out, we do want to setup the security slightly differently in tests so that we create a chain of trust, each member will have it's own part of private/pub key and the whole cluster will share the keystore with public keys. I will also clean up the redundant `authentication-optional=true, want-client-auth=true` which do not have affect and continue to require client auth. I will also make sure to add the negative tests so that we verify things do not cluster when the setup is incorrect to prevent false posiitives.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/SocketTransportResourceDefinitionRegistrar.java
```diff
@@ -40,8 +35,6 @@
  */
 public class SocketTransportResourceDefinitionRegistrar<T extends BasicTCP> extends AbstractTransportResourceDefinitionRegistrar<T> {
     enum Transport implements ResourceRegistration {
```
marekkopecky, line 38: What is the reason for keeping enum, if both members (TCP and TCP_NIO2) has been removed?

rhusar: @marekkopecky Yes, this will be removed altogether if we confirm we want to proceed with this approach.

marekkopecky:
There is an error in bootable jar execution. I can reproduce it locally by those commands:
```
mvn clean -fae -Dmaven.test.failure.ignore=true -DskipTests -DallTests
# double-check your maven repo path in next command
mvn install -fae -Dmaven.test.failure.ignore=true -Denforcer.skip -Dtest=TCP_NIO2CommandDispatcherTestCase -Dsurefire.failIfNoSpecifiedTests=false -Dts.bootable -Dmaven.repo.local=/home/$USER/.m2/repository
```
Build ends with success (thanks to `maven.test.failure.ignore=true`, so all modules are executed), but you can see that TCP_NIO2CommandDispatcherTestCase fails with:
```
org.opentest4j.AssertionFailedError: [node-2] ==> expected: <2> but was: <1>
	at org.jboss.as.test.clustering.cluster.dispatcher.CommandDispatcherTestCase.test(CommandDispatcherTestCase.java:105)
	at org.jboss.as.test.clustering.cluster.dispatcher.CommandDispatcherTestCase.test(CommandDispatcherTestCase.java:56)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

Can you please double check?

rhusar:
@marekkopecky You can just ignore this - the fix is to exclude the new test the same way it's excluded already (`                                        <exclude>org/jboss/as/test/clustering/cluster/dispatcher/CommandDispatcherTestCase.java</exclude>`) as that's unsupportable for layers based tests. The reason is described in the pom.xml.

The CI is swamped before the release, I did not want to add more load for unimportant things.

marekkopecky:
Thank you for your feedback! Anyway, I guess that this PR can't be merged with regression in CI ...

I also prepared extra tests based on the test plan here: https://github.com/marekkopecky/wildfly/commit/85146f1873367dfd87bf660c431a478a8f6def88 

There are generally tests that I prepared a few weeks ago, but rebased / updated based on latest changes. Commit can be added here, or I can create follow-up PR once this PR will be merged.

Update: I added a few more tests based on the discussion with @tommaso-borgato , cc @rhusar

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/SocketTransportResourceDefinitionRegistrar.java
marekkopecky: What is blocking that confirmation please?

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemSchema.java
rhusar: Fixed.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/JGroupsSubsystemSchema.java
rhusar: Fixed

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/SocketTransportResourceDefinitionRegistrar.java
rhusar: OK fixed while we wait for @pferraro to review.

rhusar:
> Thank you for your feedback! Anyway, I guess that this PR can't be merged with regression in CI ...

Sure. Pushed the fixes now as CI doesn't seem to be overloaded at the moment.

> I also prepared extra tests based on the test plan here: [marekkopecky@85146f1](https://github.com/marekkopecky/wildfly/commit/85146f1873367dfd87bf660c431a478a8f6def88)
> 
> There are generally tests that I prepared a few weeks ago, but rebased / updated based on latest changes. Commit can be added here, or I can create follow-up PR once this PR will be merged.
> 
> Update: I added a few more tests based on the discussion with @tommaso-borgato , cc @rhusar

I see, OK I think we can add this as post-merge items as that might need more discussion (e.g. checking that UDP-based transports don't expose TLS configuration, is not an efficient test and we don't typically add tests like that).

marekkopecky:
https://github.com/wildfly/wildfly/pull/19698#issuecomment-4449029696 

> The reason is described in the pom.xml

Can you please point me to specific lines? I'm thinking why other new *CommandDispatcherTestCase classes are passing, but just NIO2 without TLS is failing.

marekkopecky:
> checking that UDP-based transports don't expose TLS configuration, is not an efficient test and we don't typically add tests like that

Sure, we can discuss it in follow-up PR. My UDP test uses CLI commands instead of management commands, so the test takes a few milliseconds (/microseconds?) longer, but the test is much better readable and maintainable.

PR with tests are prepared here: https://github.com/wildfly/wildfly/pull/20024 - I believe that although that PR can't be merged now, it's ready for a review

rhusar:
> [#19698 (comment)](https://github.com/wildfly/wildfly/pull/19698#issuecomment-4449029696)
> 
> > The reason is described in the pom.xml
> 
> Can you please point me to specific lines? I'm thinking why other new *CommandDispatcherTestCase classes are passing, but just NIO2 without TLS is failing.

None of these are run with bootable jar execution.

Explainer for exclusions is e.g. here https://github.com/wildfly/wildfly/blob/16f84e423576b2de03db1bc87f29e0ab7e584a05/testsuite/integration/clustering/pom.xml#L1273-L1286

marekkopecky:
@rhusar Maybe wrong link? I see one comment only in your snippet: ` <!-- Uses ClusterTopologyRetriever EJB --> `, it doesn't elaborate much how it relates with bootable jar.

rhusar:
> @rhusar Maybe wrong link? I see one comment only in your snippet: `<!-- Uses ClusterTopologyRetriever EJB -->`, it doesn't elaborate much how it relates with bootable jar.

Yes, that's the one. All of these tests use the remote EJB to verify cluster topology. This has nothing to do with any of the work here, the standard versions of this test are skipped in that profile so do ANY of the extensions done for TLS. The reason is that bootable-jar runtime doesn't support the remote EJB client connectivity that  these tests require.

rhusar:
After review by @pferraro, the approach the current verison of the PR has been deemed **not supportable** primarily for the following reasons:

  1. Security – The implementation is prone to CVEs as we would have to reimplement TLS-wrapped base JDK aspects
  2. Threading issues – Several concurrency problems were supposedly identified
  3. Test coverage – Significantly more testing would be needed before inclusion in EAP -> thus a separate Jira if that materializes

Agreed path forward:

  - Revert to the previous working approach.
  - Port the per-node certificates setup and other cleanup done for the NIO work
  - ~~Add runtime auto-swap of FD_SOCK2 for FD_SOCK when the transport is configured for TLS~~ Inform the user that FD_SOCK2 does not use secured channel or similar alternative.

FYI @marekkopecky I ll be working to get the branch into this shape immediately.

marekkopecky:
> Test coverage – Significantly more testing would be needed before inclusion in EAP -> thus a separate Jira if that materializes

What do you mean by this please?

> Add runtime auto-swap of FD_SOCK2 for FD_SOCK when the transport is configured for TLS

This effectively means that

* there will be two default stacks - possible impact on testing
* Users will be informed wrongly about protocol definition by CLI / configuration. They expect FD_SOCK2, but server uses FD_SOCK. If another user code assumes FD_SOCK2 (which will be configured, but not used), then user can see confusing errors. cc @pferraro

tommaso-borgato:
> > Test coverage – Significantly more testing would be needed before inclusion in EAP -> thus a separate Jira if that materializes
> 
> What do you mean by this please?
> 
> > Add runtime auto-swap of FD_SOCK2 for FD_SOCK when the transport is configured for TLS
> 
> This effectively means that
> 
>     * there will be two default stacks - possible impact on testing
> 
>     * Users will be informed wrongly about protocol definition by CLI / configuration. They expect FD_SOCK2, but server uses FD_SOCK. If another user code assumes FD_SOCK2 (which will be configured, but not used), then user can see confusing errors. cc @pferraro

@pferraro @rhusar please guys don't do that! "runtime auto-swap of FD_SOCK2 for FD_SOCK" really looks as terrible idea to me ... perhaps is has already been done in the past, I don't know about that, but having a configuration in standalone*.xml which is changed under the hood, sounds like something we should try not having in WF

marekkopecky:
I would like to also object the usage of FD_SOCK:

* it seems that FD_SOCK2 is better then FD_SOCK [1]. User should not be forced to choose between TLS and worse failure detection
* missing implementation on FD_SOCK2 will create technical debt for the future. We should implement new feature on new protocols (not on almost deprecated protocols)

[1] https://belaban.blogspot.com/2021/05/jgroups-517-released.html

rhusar:
> > Test coverage – Significantly more testing would be needed before inclusion in EAP -> thus a separate Jira if that materializes
> 
> What do you mean by this please?

What we mean by this is that this is a significant and very complex implementation that introduces risk. The testing required for this would not be a black-box integration test, but would need substantial unit test coverage. It is not something we want to support in this scope.

rhusar:
> > > Test coverage – Significantly more testing would be needed before inclusion in EAP -> thus a separate Jira if that materializes
> > 
> > 
> > What do you mean by this please?
> > > Add runtime auto-swap of FD_SOCK2 for FD_SOCK when the transport is configured for TLS
> > 
> > 
> > This effectively means that
> > ```
> > * there will be two default stacks - possible impact on testing

False, this is not making any changes to any default stacks.

> > * Users will be informed wrongly about protocol definition by CLI / configuration. They expect FD_SOCK2, but server uses FD_SOCK. If another user code assumes FD_SOCK2 (which will be configured, but not used), then user can see confusing errors. cc @pferraro
> > ```
> 
> @pferraro @rhusar please guys don't do that! "runtime auto-swap of FD_SOCK2 for FD_SOCK" really looks as terrible idea to me ... perhaps is has already been done in the past, I don't know about that, but having a configuration in standalone*.xml which is changed under the hood, sounds like something we should try not having in WF

That's a common pattern already; nothing new here.

rhusar:
> I would like to also object the usage of FD_SOCK:
> 
> * it seems that FD_SOCK2 is better then FD_SOCK [1]. User should not be forced to choose between TLS and worse failure detection
> * missing implementation on FD_SOCK2 will create technical debt for the future. We should implement new feature on new protocols (not on almost deprecated protocols)
> 
> [1] https://belaban.blogspot.com/2021/05/jgroups-517-released.html

We have supported the protocol for decades and there aren't any material indications of issues with it. Nevertheless, lets not jump into conclusions while all options are being researched – I only meant that comment as an information for QE to hold off testing until I update the branch with a solution that we want to go with.

marekkopecky:
> False, this is not making any changes to any default stacks.

My current understanding is that with TLS configured in JGroups Transport, the default stack contains FD_SOCK (`runtime auto-swap of FD_SOCK2 for FD_SOCK`), but without TLS configured in JGroups Transport, the default stack contains FD_SOCK2. So there seems to be two stacks. Or do I missing something?

marekkopecky:
> That's a common pattern already; nothing new here.

This seems to be an anti-pattern (`Users will be informed wrongly about protocol definition by CLI / configuration. They expect FD_SOCK2, but server uses FD_SOCK. If another user code/configuration assumes FD_SOCK2 (which will be configured, but not used), then user can see confusing errors.`). If we are using this pattern in clustering, we should rather get rid of it, not extend it.

> We have supported the protocol for decades and there aren't any material indications of issues with it. 

At least effectiveness seems to be an issue - https://belaban.blogspot.com/2021/05/jgroups-517-released.html - eg 3 thread instead of 1 threads per node, etc.

> hold off testing until I update the branch with a solution that we want to go with.

Sure, thanks! There is also an analysis PR, where architecture may be described / decided.

rhusar:
> > That's a common pattern already; nothing new here.
> 
> This seems to be an anti-pattern (`Users will be informed wrongly about protocol definition by CLI / configuration. They expect FD_SOCK2, but server uses FD_SOCK. If another user code/configuration assumes FD_SOCK2 (which will be configured, but not used), then user can see confusing errors.`). If we are using this pattern in clustering, we should rather get rid of it, not extend it.

This isn't relevant for this PR as there is no such pattern here. I was just trying to make the R&D process more transparent, but it appears to have caused more confusion than benefit.

> > We have supported the protocol for decades and there aren't any material indications of issues with it.
> 
> At least effectiveness seems to be an issue - https://belaban.blogspot.com/2021/05/jgroups-517-released.html - eg 3 thread instead of 1 threads per node, etc.

This is not really significant in practice. FD_SOCK uses 3 threads per member: pinger_thread - monitors the socket to the neighbor; ServerSocketHandler - accepts incoming connections on the server socket; ClientConnectionHandler - handles each accepted client connection and FD_SOCK2 replaces all of that with a single NioServer using one selector thread, plus it reuses the transport's shared thread pool for callbacks.

But these are idle threads: as FD_SOCK's Javadoc says so (line 39: "those threads will be idle as long as both members are running"). They sit blocked on socket reads and accept(), only consuming a stack's memory (~1MB each by default) but essentially no CPU. So the saving is ~2MB of stack memory per member, not CPU which in is negligible.

However, once virtual threads are enabled, these scales skew the other side since, ironically, NioServer's single-selector design is more complex than simple blocking I/O with virtual threads, since NIO requires readiness selection, event loops, and callback dispatch - which virtual threads were designed to make unnecessary.

> > hold off testing until I update the branch with a solution that we want to go with.
> 
> Sure, thanks! There is also an analysis PR, where architecture may be described / decided.

Updated the proposal now along with the latest update of this PR. And fixed the link at the top that still linked the original proposal.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/logging/JGroupsLogger.java
```diff
@@ -169,4 +169,16 @@ public interface JGroupsLogger extends BasicLogger {
...
+    /**
+     * Warning for when a service requests a non-blocking NIO channel from TLS-secured ManagedSocketFactory.
+     * TLS can only be supported for blocking socket-based communication and NIO channels used by protocols such as FD_SOCK2 cannot be wrapped with TLS.
+     * The channel will be created without TLS encryption, so to secure failure detection communication, use TCP transport-based failure detection (TCP.enable_suspect_events) or the socket-based FD_SOCK protocol.
+     * TCP_NIO2 cannot be configured with TLS in the model; so this does not apply to it.
+     *
+     * @param serviceName name of the service that requested the channel to be created
+     */
+    @LogMessage(level = WARN)
+    @Message(id = 37, value = "Transport of the stack is configured to create TLS sockets and service '%s' requires non-blocking channel for which TLS support is not available. Use TCP transport-based failure detection or socket-based FD_SOCK protocol for also securing the failure detection communication.")
```
pferraro: Given that this log message is not protocol specific, I would keep it generic, e.g.
```suggestion
    @Message(id = 37, value = "Service '%s' requires a non-blocking socket for which TLS support is not available.  To secure cluster communication for this service, use a blocking socket variant of the associated protocol.")
```

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/logging/JGroupsLogger.java
```diff
@@ -169,4 +169,16 @@ public interface JGroupsLogger extends BasicLogger {
...
+    void channelTlsNotAvailable(String serviceName);
```
pferraro: "channel" is too overloaded in this context.

```suggestion
    void secureSocketChannelNotAvailable(String serviceName);
```

clustering/jgroups/extension/src/main/resources/schema/jboss-as-jgroups_10_0.xsd
```diff
@@ -0,0 +1,465 @@
...
+                    </xs:annotation>
+                </xs:attribute>
+            </xs:extension>
+        </xs:complexContent>
+    </xs:complexType>
+
+    <xs:complexType name="transport">
+        <xs:complexContent>
+            <xs:extension base="tns:protocol">
+                <xs:sequence>
```
pferraro, line 344: Nit: Let's use xs:all here (and in the ResourceXMLElement content), since there is no need for ordering (or repetition).

pferraro (approved):
Remaining comments are optional.

clustering/jgroups/extension/src/main/resources/schema/jboss-as-jgroups_10_0.xsd
rhusar: ..and in the parser. Will do.

marekkopecky:
I see those two warnings in CommandDispatcherTestCase test with TLS configuration and with FD_SOCK2:

```
node-2 2026-06-05  WARN  [org.jboss.as.clustering.jgroups] (ServerService Thread Pool -- 38) WFLYCLJG0037: Transport of the stack is configured to create TLS sockets and service 'jgroups.nio.server.fd_sock' requires non-blocking channel for which TLS support is not available. Use TCP transport-based failure detection or socket-based FD_SOCK protocol for also securing the failure detection communication.

node-1 2026-06-05  WARN  [org.jboss.as.clustering.jgroups] (thread-5,ejb,node-1(v=16.0.11)) WFLYCLJG0037: Transport of the stack is configured to create TLS sockets and service 'jgroups.nio.client' requires non-blocking channel for which TLS support is not available. Use TCP transport-based failure detection or socket-based FD_SOCK protocol for also securing the failure detection communication.
```

'jgroups.nio.server.fd_sock' could make sense (not fd_sock2?), but I'm not sure about second warning about 'jgroups.nio.client'. Is this expected?

rhusar:
> I see those two warnings in CommandDispatcherTestCase test with TLS configuration and with FD_SOCK2:
> 
> ```
> node-2 2026-06-05  WARN  [org.jboss.as.clustering.jgroups] (ServerService Thread Pool -- 38) WFLYCLJG0037: Transport of the stack is configured to create TLS sockets and service 'jgroups.nio.server.fd_sock' requires non-blocking channel for which TLS support is not available. Use TCP transport-based failure detection or socket-based FD_SOCK protocol for also securing the failure detection communication.
> 
> node-1 2026-06-05  WARN  [org.jboss.as.clustering.jgroups] (thread-5,ejb,node-1(v=16.0.11)) WFLYCLJG0037: Transport of the stack is configured to create TLS sockets and service 'jgroups.nio.client' requires non-blocking channel for which TLS support is not available. Use TCP transport-based failure detection or socket-based FD_SOCK protocol for also securing the failure detection communication.
> ```
> 
> 'jgroups.nio.server.fd_sock' could make sense (not fd_sock2?), but I'm not sure about second warning about 'jgroups.nio.client'. Is this expected?

@marekkopecky Well, yes, that's how JGroups does it. Both service names come from FD_SOCK2's impl and use of NioServer (a JGroups object):

  - `jgroups.nio.server.fd_sock` is defined directly in FD_SOCK2.java:334, passed when creating the NioServer for the FD_SOCK2 ping server.
  - `jgroups.nio.client` is defined in NioConnection.java:52, used when the NioServer creates an outgoing client connection (i.e., when one node connects to another node's FD_SOCK2 server).

So this is logging as expected. I will update the message today as proposed by Paul above.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/logging/JGroupsLogger.java
rhusar: Thanks done

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/logging/JGroupsLogger.java
rhusar: True, fixed

clustering/jgroups/extension/src/main/resources/schema/jboss-as-jgroups_10_0.xsd
rhusar: Ah, no.

xs:all can't appear inside xs:extension in XSD 1.0.

Moving to XSD 1.1 would be a different effort as that needs to replace JDK validation library we use with an external one,since in JDK 17 the validation javax.xml.validation is XSD 1.0 only IIUC.

rhusar:
@marekkopecky I believe I have approved/reviewed all pending requests. Anything else holding QE pre-check that I can help with?

pferraro (approved):

tommaso-borgato:
> I have added a new commit that clarifies and correctly restricts TLS support to JGroups TCP transport only as uncovered by @marekkopecky.
> 
> TLSManagedSocketFactory only overrides createSocket()/createServerSocket() (used by TCP), not createSocketChannel()/createServerSocketChannel() (used by TCP_NIO2). TLS cannot work with TCP_NIO2 without SSLEngine-based NIO channel wrapping which does not exist in JGroups - tracked by a separate Jira.
> 
> After some ENG discussions, I have remodeled this a little bit:
> 
>     * Extract SecurableSocketTransportResourceDefinitionRegistrar (TCP) from SocketTransportResourceDefinitionRegistrar (TCP_NIO2), so SSL attributes (client-ssl-context, server-ssl-context) only appear on 'securable' TCP
> 
>     * Add ssl-context as optional child element in the 10.0 transport XSD type (XSD 1.0 cannot enforce per-type-attribute constraints)
> 
>     * Add transformer rejection for SSL attributes against older model versions
> 
>     * Rework TLS TCP_NIO2 test case as a plain TCP_NIO2 test (without TLS as per original Jira for this)

@rhusar @pferraro can you please add a few words about what the actual behavior is from the functional point of view? IIUC, since FD_SOCK2 doesn't have support for encryption and it can't be added because it is currently missing in the underlying java libraries used, then EAP7-1865 requirement becomes just about encrypting  sensitive information ... is that what the above translates to?

rhusar:
> > I have added a new commit that clarifies and correctly restricts TLS support to JGroups TCP transport only as uncovered by @marekkopecky.
> > TLSManagedSocketFactory only overrides createSocket()/createServerSocket() (used by TCP), not createSocketChannel()/createServerSocketChannel() (used by TCP_NIO2). TLS cannot work with TCP_NIO2 without SSLEngine-based NIO channel wrapping which does not exist in JGroups - tracked by a separate Jira.
> > After some ENG discussions, I have remodeled this a little bit:
> > ```
> > * Extract SecurableSocketTransportResourceDefinitionRegistrar (TCP) from SocketTransportResourceDefinitionRegistrar (TCP_NIO2), so SSL attributes (client-ssl-context, server-ssl-context) only appear on 'securable' TCP
> > 
> > * Add ssl-context as optional child element in the 10.0 transport XSD type (XSD 1.0 cannot enforce per-type-attribute constraints)
> > 
> > * Add transformer rejection for SSL attributes against older model versions
> > 
> > * Rework TLS TCP_NIO2 test case as a plain TCP_NIO2 test (without TLS as per original Jira for this)
> > ```
> 
> @rhusar @pferraro can you please add a few words about what the actual behavior is from the functional point of view? IIUC, since FD_SOCK2 doesn't have support for encryption and it can't be added because it is currently missing in the underlying java libraries used, then EAP7-1865 requirement becomes just about encrypting sensitive information ... is that what the above translates to?

@tommaso-borgato There is a lengthy section in the documentation that explains this is detail - https://github.com/wildfly/wildfly/pull/19698/changes#diff-5d9c200f017b6e806ad674b13839304dc22f98ff9d718cb3a9eb6be09b4c6d74

> "becomes just about encrypting sensitive information"

That's an incorrect oversimplification. Compared to patterns like ASYM_ENCRYPT which only encrypts the body, this feature encrypts internal  control messages, headers and entire communication, not just jgroups message body. More importantly, with required mutual authentication this also provides auth.

tommaso-borgato:
> > > I have added a new commit that clarifies and correctly restricts TLS support to JGroups TCP transport only as uncovered by @marekkopecky.
> > > TLSManagedSocketFactory only overrides createSocket()/createServerSocket() (used by TCP), not createSocketChannel()/createServerSocketChannel() (used by TCP_NIO2). TLS cannot work with TCP_NIO2 without SSLEngine-based NIO channel wrapping which does not exist in JGroups - tracked by a separate Jira.
> > > After some ENG discussions, I have remodeled this a little bit:
> > > ```
> > > * Extract SecurableSocketTransportResourceDefinitionRegistrar (TCP) from SocketTransportResourceDefinitionRegistrar (TCP_NIO2), so SSL attributes (client-ssl-context, server-ssl-context) only appear on 'securable' TCP
> > > 
> > > * Add ssl-context as optional child element in the 10.0 transport XSD type (XSD 1.0 cannot enforce per-type-attribute constraints)
> > > 
> > > * Add transformer rejection for SSL attributes against older model versions
> > > 
> > > * Rework TLS TCP_NIO2 test case as a plain TCP_NIO2 test (without TLS as per original Jira for this)
> > > ```
> > 
> > 
> > @rhusar @pferraro can you please add a few words about what the actual behavior is from the functional point of view? IIUC, since FD_SOCK2 doesn't have support for encryption and it can't be added because it is currently missing in the underlying java libraries used, then EAP7-1865 requirement becomes just about encrypting sensitive information ... is that what the above translates to?
> 
> @tommaso-borgato There is a lengthy section in the documentation that explains this is detail - https://github.com/wildfly/wildfly/pull/19698/changes#diff-5d9c200f017b6e806ad674b13839304dc22f98ff9d718cb3a9eb6be09b4c6d74
> 
> > "becomes just about encrypting sensitive information"
> 
> That's an incorrect oversimplification. Compared to patterns like ASYM_ENCRYPT which only encrypts the body, this feature encrypts internal control messages, headers and entire communication, not just jgroups message body. More importantly, with required mutual authentication this also provides auth.

Tnks @rhusar, I see .... IIUC when using an Elytron `client-ssl-context` to configure encryption on the default JGroups TPC stack, everything other than FD_SOCK2 would be encrypted; in case the user wants complete encryption, just removing FD_SOCK2, would default WF to using FD_SOCK which would be encrypted the same way the JGroups TPC stack would ... is that a correct summary?

rhusar:
> > > @rhusar @pferraro can you please add a few words about what the actual behavior is from the functional point of view? IIUC, since FD_SOCK2 doesn't have support for encryption and it can't be added because it is currently missing in the underlying java libraries used, then EAP7-1865 requirement becomes just about encrypting sensitive information ... is that what the above translates to?
> > 
> > 
> > @tommaso-borgato There is a lengthy section in the documentation that explains this is detail - https://github.com/wildfly/wildfly/pull/19698/changes#diff-5d9c200f017b6e806ad674b13839304dc22f98ff9d718cb3a9eb6be09b4c6d74
> > > "becomes just about encrypting sensitive information"
> > 
> > 
> > That's an incorrect oversimplification. Compared to patterns like ASYM_ENCRYPT which only encrypts the body, this feature encrypts internal control messages, headers and entire communication, not just jgroups message body. More importantly, with required mutual authentication this also provides auth.
> 
> Tnks @rhusar, I see .... IIUC when using an Elytron `client-ssl-context` to configure encryption on the default JGroups TPC stack, 

For starters, @tommaso-borgato, you need both, `client-ssl-context` and `server-ssl-context` as required by the model.

> everything other than FD_SOCK2 would be encrypted; in case the user wants complete encryption, 

The TCP connections created by FD_SOCK2 would not be encrypted, as those are infrastructure connections, not data connections. No data flows over those connections so there is nothing to encrypt.

> just removing FD_SOCK2, would default WF to using FD_SOCK 

Not at all (as agreed and discussed above) – it would use the transport-based (enable suspect events) FD mechanism, using the encrypted transport TCP connections.

> which would be encrypted the same way the JGroups TPC stack would ... is that a correct summary?

So no, if users want to use FD_SOCK, thy need to configure it manually.

rhusar:
n.b. last force push has no code changes, just clean rebase to get fresh CI runs while we wait for QE ack.

marekkopecky (approved):
Thank you! I'm approving this PR.

pferraro:
Thanks @rhusar, @marekkopecky, and @bstansberry!

PR #19700, opened 2026-02-25
WFLY-21524 - Replace testcontainer-based otel collector with in-memory

jasondlee:
https://issues.redhat.com/browse/WFLY-21524

Create an in-memory collector
* Add InMemoryCollector to encapsulate grpc and proto concerns
* Add Undertow server and servlets to handle OTLP over HTTP
* Add deps for grpc, otlp proto, etc
* Add models for signal data
* Update tests to reflect API changes (e.g., Jaeger -> OTLP)
* Migrate tests to the in-mem collector
* API cleanup

### DIFF
--- a/testsuite/integration/expansion/pom.xml
+++ b/testsuite/integration/expansion/pom.xml
@@ -39,6 +39,8 @@
         <!-- Use the expansion bom -->
         <dependency.management.import.artifact>wildfly-standard-expansion-bom</dependency.management.import.artifact>
         <dependency.management.import.test.artifact>wildfly-standard-test-expansion-bom</dependency.management.import.test.artifact>
+
+        <version.io.undertow>2.3.20.Final</version.io.undertow>
     </properties>
 
     <dependencyManagement>
@@ -264,6 +266,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>failureaccess</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-codec</artifactId>
             <scope>test</scope>
@@ -334,6 +341,11 @@
         </dependency>
         <dependency>
             <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+            <version>${version.io.opentelemetry.opentelemetry}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-sdk-common</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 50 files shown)


jasondlee:
This is a change I've been kicking around in my head for a while now. Over the past couple of weeks, as I've had time, I've implemented it to see how it would go, and if it would be an acceptable change. @bstansberry , @jamezp , @rhusar , please review, pick apart, etc. Even reject if you wish. I've had the entire testuite/integration/expansion testsuite running since about 5:30 last night (21 full runs so far) with 0 failures, so I think it's stable.

testsuite/integration/expansion/pom.xml
```diff
@@ -39,6 +39,8 @@
         <!-- Use the expansion bom -->
         <dependency.management.import.artifact>wildfly-standard-expansion-bom</dependency.management.import.artifact>
         <dependency.management.import.test.artifact>wildfly-standard-test-expansion-bom</dependency.management.import.test.artifact>
+
+        <version.io.undertow>2.3.20.Final</version.io.undertow>
```
jasondlee, line 43: This is already defined (in the Core pom), but that is not visible from this POM for some reason, and I'm a bit hesitant to mess with the POM relationships too much, but I'm going to dig into this $SOON. If anyone knows the answer off hand, I'm all ears. :)

rhusar:
@jasondlee @bstansberry @jamezp OK since you are asking for it, here it goes :-) This is a major shift in testing philosophy that trades true integration testing for speed and stability. I have always treated running against a real container as our way to test our integration with real endpoint. While someone might argue that is the concern of the downstream project, we have way too much wiring that we did historically found issues in, that I believe we have to continue testing here.

So, now we are using actual ecosystem components (OpenTelemetry Collector container + Jaeger container) and with this change these are no longer a true integration test. It would only test WF can emit OTLP protocol messages and that they can be parsed by the in-memory test logic. However, we would *lose* testing that WF actually interoperates with the real OpenTelemetry Collector and that data survives the full processing pipeline (that we parse from our test util). Also, as part of this exercise we (or I) learned how to tune many aspects of telemetry backends.

Don't get me wrong – I use a similar approach in my [jgroups-opentelemetry](https://github.com/rhusar/jgroups-opentelemetry/blob/main/src/test/java/org/jgroups/opentelemetry/TestMetricExporter.java) project where I implement a custom [exporter](https://github.com/rhusar/jgroups-opentelemetry/blob/main/src/test/java/org/jgroups/opentelemetry/TestMetricExporter.java) and bypass the whole chain, but that's not a container and only a library so the approach is appropriate there.

I would much rather see the issues we are having fixed or at least understood why we are having them before doing any of this. We could still then go mix & match approach, use this for most of the test for speed and reliability but keep proper tests for only subset of tests.

jasondlee:
@rhusar That's all fair. There's certainly value in at least smoke testing integration with an actual server, but we have a a lot of janky code in our tests. We only spin up the Jaeger container to get traces, as it was the first supported downstream I could find with a usable API, we're reading, iirc, exported log entries from a file, and parsing those entries into a model we can use in our tests., and we configure the Prometheus endpoint in the collector so we can read metrics from it. There's a lot of extra commotion going on just to the get the data back out of the collector so that we can verify that the WildFly-specific information we care about is correct. This gives us a much simpler approach to that data using officially-supported otel models (right up until "we" convert it to something simpler, e.g., we get rid of all of the grpc field id variables). 

That said, I think the smoke test is fine to verify that it CAN connect, but much beyond that feels like we're testing -- awkwardly -- that the library itself works, which I think we can and must assume (or we wouldn't be using it). For testing WF-specific bits, I think an in-memory approach like this should be sufficient. I'm a team player, though, so if others don't like the idea, I can close the PR and be content knowing that I learned some things along the way. :)

PR #19705, opened 2026-02-25
[WFLY-21514] As required by the EE spec, isolate appclient modules fr…

scottmarlow:
…om other ear subdeplyments but not in appclient process

https://issues.redhat.com/browse/WFLY-21514

### DIFF
--- a/ee/src/main/java/org/jboss/as/ee/structure/ApplicationClientDeploymentProcessor.java
+++ b/ee/src/main/java/org/jboss/as/ee/structure/ApplicationClientDeploymentProcessor.java
@@ -99,8 +99,12 @@ public class ApplicationClientDeploymentProcessor implements DeploymentUnitProce
                 }
             }
 
-            // TODO perhaps limit this to non-appclient processes
             if (APP_CLIENT_ISOLATED && deploymentUnit.getParent() != null
+                    // limit this to non-appclient processes
+                    // See 'Components in the application client container must have access to the following classes and resources.' discussion in:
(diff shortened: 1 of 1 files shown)


ee/src/main/java/org/jboss/as/ee/structure/ApplicationClientDeploymentProcessor.java
```diff
@@ -99,8 +99,12 @@ public void deploy(DeploymentPhaseContext phaseContext) throws DeploymentUnitPro
                 }
             }
 
-            // TODO perhaps limit this to non-appclient processes
             if (APP_CLIENT_ISOLATED && deploymentUnit.getParent() != null
+                    // limit this to non-appclient processes
+                    // See 'Components in the application client container must have access to the following classes and resources.' discussion in:
+                    // https://jakarta.ee/specifications/platform/10/jakarta-platform-spec-10.0#application-client-container-class-loading-requirements
```
bstansberry, line 105: Reading this again / thinking about it more carefully I think this spec section is clear that this should not be limited to non-appclient processes.

```
Components in the application client container must not have access to the following classes and resources, unless such classes or resources are covered by one of the rules above.

Other classes or resources in the application package. For example, the application client should not have access to the classes in other application client jar files in the same ear file....
```

The application client container doesn't exist outside of an appclient process, so that language is relevant.

scottmarlow:
Changing to draft as we get more TCK failures (didn't look into causes) with this change.

bstansberry:
@scottmarlow My WFCORE-7518 fix resolved the EE 10 TCK problem we were seeing with WFLY-21514, so I don't think this is needed.

PR #19708, opened 2026-02-26
[WFLY-21528] Disable scanning client WebSocket endpoints by default. …

jamezp:
…Added a system property to re-enable this given it has been the default since the introduction of WebSocket into the server.

https://issues.redhat.com/browse/WFLY-21528

Instead of completely deleting the scanning, I opted to disable it by default. We can definitely remove it, but since it's been like this since the introduction of WebSocket into WIldFly, I'm hesitant to remove it.

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
@@ -24,12 +24,14 @@ import org.jboss.jandex.DotName;
 import org.jboss.modules.Module;
 import org.wildfly.extension.undertow.UndertowExtension;
 import org.wildfly.extension.undertow.logging.UndertowLogger;
+import org.wildfly.security.manager.WildFlySecurityManager;
 
 import jakarta.websocket.ClientEndpoint;
 import jakarta.websocket.Endpoint;
 import jakarta.websocket.server.ServerApplicationConfig;
 import jakarta.websocket.server.ServerEndpoint;
 import jakarta.websocket.server.ServerEndpointConfig;
+
 import java.lang.reflect.Modifier;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

jamezp:
The security manager failure might be related. I'm going to look now, but it's really strange give the permission error is `"java.lang.RuntimePermission" "accessDeclaredMembers"` which I made no changes to.

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
```diff
@@ -231,4 +245,9 @@ void addManagementWebsocket(final DeploymentUnit unit, Class endpoint, String pa
             UndertowLogger.ROOT_LOGGER.failedToRegisterWebsocket(endpoint, path, e);
         }
     }
+
+    private static boolean isClientScanningEnabled() {
+        final String value = WildFlySecurityManager.getPropertyPrivileged(ENABLE_CLIENT_ENDPOINT_SCANNING, "false");
+        return value.isEmpty() || Boolean.parseBoolean(value);
```
rhusar, line 251: This logic is wrong – you resolve with a default value and then check for emptiness which will never be the case unless you explicitly set to empty string.

Easiest fix

```suggestion
        final String value = WildFlySecurityManager.getPropertyPrivileged(ENABLE_CLIENT_ENDPOINT_SCANNING, "false");
        return Boolean.parseBoolean(value);
```

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/deployment/UndertowJSRWebSocketDeploymentProcessor.java
jamezp: This is to catch things like an empty parameter passed, e.g. `-Dorg.wildfly.extension.undertow.deployment.enableClientEndpointScanning`. That would just return as an empty string, not the default value.

rhusar: Exactly, because `Boolean.parseBoolean("")` resolves to `false`. So, no need for value.isEmpty().

jamezp: Correct, which is the opposite of what I want :) I want the empty value to be `true` not `false`. In other words `org.wildfly.extension.undertow.deployment.enableClientEndpointScanning=true` and `org.wildfly.extension.undertow.deployment.enableClientEndpointScanning` should both result in `true`.

rhusar: Oh gotcha, empty or true = true. We use that pattern quite often, that's true.

rhusar (approved):
LGTM

rhusar:
Thanks @jamezp and @bstansberry

PR #19719, opened 2026-03-03
[WFLY-21548] Deployment fails with ConcurrentModificationException when using modcluster capabilities

elguardian:
Issue: https://issues.redhat.com/browse/WFLY-21548

All access to contexts is sync but resume. This might lead to concurrent modification exception when some operations are happening at the same time.

### DIFF
--- a/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
+++ b/mod_cluster/undertow/src/main/java/org/wildfly/mod_cluster/undertow/UndertowEventHandlerAdapterService.java
@@ -192,7 +192,7 @@ public class UndertowEventHandlerAdapterService implements UndertowEventListener
     }
 
     @Override
-    public void resume() {
+    public synchronized void resume() {
         for (Context context : this.contexts) {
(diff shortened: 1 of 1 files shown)


rhusar:
Let me stick a hold on this – I have an older branch that was too late for 39.x that I still have not submitted which fixes the problem.

Also, the Jira is duplicate of WFLY-21276.

elguardian:
@rhusar remember to properly link the jiras en the downstream, Otherwise info is lost unless somebody tells you.

rhusar:
> @rhusar remember to properly link the jiras en the downstream, Otherwise info is lost unless somebody tells you.

@elguardian Not sure what you are trying to tell me, but given I created the first issue, there is nothing I can link. People filing issues afterwards (e.g. you) would have to do a search and find existing issues to link. But our Jira instance doesn't make it too easy, hopefully that will improve with the new instance.

rhusar (approved):
Lets go ahead with this, so we have something right now for downstream, I will rebase and do a broader cleanup in a follow-up branch.

rhusar:
Thanks @elguardian

PR #19724, opened 2026-03-05
[WFLY-21571] Create test cases for Elytron Audit Logging using digest method

mskacelik:
issue: https://issues.redhat.com/browse/WFLY-21571

Few notes:
- `makeCallWithBasicAuthn` seem to also support [Digest](https://github.com/wildfly/wildfly/blob/841fb9b9bba20abbbe6f3cd4fd606357f68eca50/testsuite/shared/src/main/java/org/jboss/as/test/integration/security/common/Utils.java#L451), so the method name might be bit misleading - but again I am no expert.

- I noticed that BASIC and DIGEST behave differently for `runTestAuditLogFileEncoding` and `runTestFailedPermissionCheck`, mainly former returns a 401 and latter returns a 403. With that said, the events are properly logged in the `audit.log`.

- I changed some deprecated stuff too: `UndertowDomainMapper`, `new URL`.

### DIFF
--- a/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/audit/AbstractAuditLogTestCase.java
+++ b/testsuite/integration/elytron/src/test/java/org/wildfly/test/integration/elytron/audit/AbstractAuditLogTestCase.java
@@ -16,7 +16,7 @@ import org.jboss.as.test.shared.ServerReload;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.wildfly.test.security.common.elytron.SimpleSecurityDomain;
-import org.wildfly.test.security.common.elytron.UndertowDomainMapper;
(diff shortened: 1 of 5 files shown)


mskacelik:
OK, I've just cherry-picked (on my local machine) the commit next to 38ad708a2455bcb0bf7134e4429197bfb8b174fc (before the fix) to ensure the tests are valid and that `testFailedAuthWrongUserDigest` is failing as expected.
```
[ERROR] Failures:
[ERROR]   FileAuditLogTestCase.testFailedAuthWrongUserDigest:150->runTestFailedAuthWrongUser:66 Failed authentication with wrong user was not logged
[INFO]
[ERROR] Tests run: 14, Failures: 1, Errors: 0, Skipped: 0
```

Interestingly enough, `testFailedAuthWrongPasswordDigest` is working even before the fix. I have reproduced this just in case, and it was indeed working as intended; only invalid name attempts were not logged into the audit file.

PR #19733, opened 2026-03-09
Workaround for WFLY-21519 StabilityServerSetupSnapshotRestoreTasks.tearDown() fails with ConnectException

rhusar:
Instead of failing the test, catch and log the exception instead warning that the container might be tainted for future tests.

Workaround for https://issues.redhat.com/browse/WFLY-21519


rhusar:
The workaround seems to be effective! I triggered over 10 runs overnight and they are all green.

https://ci.wildfly.org/buildConfiguration/WF_PullRequest_LinuxJdk21?branch=pull%2F19733&buildTypeTab=overview

@pferraro please review.

pferraro (approved):

PR #19742, opened 2026-03-10
[WFLY-21519] Pass the server address:port info to ServerReload from S…

bstansberry:
…tabilityServerSetupSnapshotRestoreTasks

https://issues.redhat.com/browse/WFLY-21519

@rhusar Please let me know if this fixes the problem. If not I'll just file a different jira and change the commit message, because this is a valid change regardless.


rhusar (approved):
Amazing find Brian, thank you!!! I knew why I asked you :)

The parameters default to a default node which would match node1 in clustering test suite:

```java
    public static class Parameters {
        private int timeout = TIMEOUT;
        boolean adminOnly = false;
        String protocol = "remote+http";
        String serverAddress = TestSuiteEnvironment.getServerAddress();
        int serverPort = TestSuiteEnvironment.getServerPort();
```

And this worked for the original author as that was only ever tested on a single node tests. This class has been quite bug ridden, but I believe we have now fixed it entirely.

I went through the logs, there are no SST errors nor 'Failed to stop container'.

rhusar:
Thanks @bstansberry 

FYI @pferraro

PR #19747, opened 2026-03-12
[WFLY-21491] Fix for intermittent failures of OrderedChildResourcesTe…

yersan:
…stCase

Jira issue: https://issues.redhat.com/browse/WFLY-21491

### DIFF
--- a/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/OrderedChildResourcesTestCase.java
+++ b/testsuite/domain/src/test/java/org/jboss/as/test/integration/domain/OrderedChildResourcesTestCase.java
@@ -5,11 +5,25 @@
 
 package org.jboss.as.test.integration.domain;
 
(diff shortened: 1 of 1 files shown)


rhusar (approved):
The wait timeout was increased from 20 seconds to 60 seconds, giving the secondary more time to reconnect after a primary reload.

Lets do it.

bstansberry (approved):
Good call. 20 seconds is not very long and AIUI this rarely fails so I expect that's the problem.

yersan:
In addition, there were no assertion at the end of the do / while, so if it timed out, we were asserting later using stale model configuration since the secondary host could have not been connected yet, so probably there instead of in the host reconnection process

rhusar:
Clustering CI failure is unrelated

rhusar:
Thanks @yersan and @bstansberry

PR #19752, opened 2026-03-16
WFLY-21592 Fix NPE on channel disconnect

pferraro:
https://issues.redhat.com/browse/WFLY-21592


rhusar (approved):
LGTM; in the original fix, channelDisconnected() was called manually from AbstractChannelResourceDefinitionRegistrar after channel.disconnect() had already returned. At that point JChannel.getClusterName() returns null

rhusar:
Thanks @pferraro

PR #19754, opened 2026-03-16
[WFLY-21523] EJB: External key to search for related timers

### LINKED ISSUE #816: EJB: External key to search for related timers
### Description

Certain users (such as those utilizing Process Automation Managers like PAM) require the ability to query EJB timers based on custom metadata associated with the timer. While it is currently possible to rely on the `Serializable` info object provided to the `TimerConfig` during timer creation, this querying method is highly inefficient. The info object must be deserialized, forcing the system to search through timers sequentially (one by one).

This proposal introduces an indexed `external_id` column to the underlying database persistence schema and exposes an extended Java API to configure and query this identifier. This enhancement allows applications to tag timers with a specific business key upon creation, thereby enabling low-latency SQL lookups directly through the EJB Timer Service.

### Issue Contact

tadamski

tadamski:
https://redhat.atlassian.net/browse/WFLY-21523


ejb3/src/main/java/org/jboss/as/ejb3/timerservice/ExtendedTimerConfig.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * WildFly specific extension of {@link TimerConfig} that allows for an external identifier
+ * to be associated with an EJB timer.
+ * <p>
+ * This identifier is stored in a dedicated database column to allow for high-performance
+ * indexed lookups without deserializing the timer info object.
+ * </p>
+ */
+public class ExtendedTimerConfig extends TimerConfig {
```
bstansberry: Is this intended as end user API? The test treats it as such by using it in ExternalIdTimerBean.

If so, this module is private, so end user API doesn't belong in it.

tadamski: @bstansberry I have moved the API to jboss-ejb3-ext-api project: https://github.com/wildfly/jboss-ejb3-ext-api/pull/42

tadamski:
failing microprofile tests are unrelated

PR #19770, opened 2026-03-17
WFLY-21629 Server entries auto-generated from listener socket bindings should reflect port-offset.

pferraro:
https://redhat.atlassian.net/browse/WFLY-21629

____

> [!WARNING]

> * [WFLY-21629](https://issues.redhat.com/browse/WFLY-21629)


rhusar (approved):
LGTM

n.b. The existing OpenAPIAutoGenerateServersTestCase doesn't cover the port-offset scenario and no new test is added.

rhusar:
Thanks @pferraro

pferraro:
> LGTM
> 
> n.b. The existing OpenAPIAutoGenerateServersTestCase doesn't cover the port-offset scenario and no new test is added.

I did not bother because this configuration is not very practical, since the URLs it generates are not likely accessible, if even resolvable, by a client. It is more likely that users would keep relative to the host/port of the OprnAPI endpoint, or just specify manually.

PR #19772, opened 2026-03-18
WFLY-21589 Deprecate ManagementServerSetupTask.ContainerConfigurationBuilder#tearDownScript and supersede by restoring a snapshot

rhusar:
Resolves
https://redhat.atlassian.net/browse/WFLY-21589
https://redhat.atlassian.net/browse/WFLY-21587

____

> [!WARNING]

> * [WFLY-21589](https://issues.redhat.com/browse/WFLY-21589)
> * [WFLY-21587](https://issues.redhat.com/browse/WFLY-21587)


testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
```diff
@@ -181,9 +217,23 @@ public ContainerConfiguration getContainerConfiguration(String container) {
      */
     public static ContainerConfigurationBuilder createContainerConfigurationBuilder() {
         return new ContainerConfigurationBuilder() {
+            private Stability stability = Stability.DEFAULT;
+            private boolean skipStabilityRequirementAndSnapshot = false;
             private List<List<String>> setupScript = List.of();
             private List<List<String>> tearDownScript = List.of();
 
+            @Override
+            public ContainerConfigurationBuilder requiredStability(Stability stability) {
```
pferraro: Use a verb, e.g. requireStability(...)

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
```diff
@@ -40,6 +45,18 @@ public interface ContainerSetConfiguration {
...
+         * Returns {@link Stability} required for the container during the test.
+         * @return {@link Stability} required for the container during the test.
+         */
+        Stability getRequiredStability();
+
+        /**
+         * Returns whether stability handling and automatic snapshot restore are skipped for the container.
+         * @return {@code true} if stability handling and snapshot restore are skipped, {@code false} otherwise.
+         */
+        boolean isStabilityRequirementAndSnapshotSkipped();
```
rhusar: We also discussed that we want to avoid this and find a better solution for when this is 'needed'

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
```diff
@@ -40,6 +45,18 @@ public interface ContainerSetConfiguration {
     }
 
     public interface ContainerConfiguration {
+        /**
+         * Returns {@link Stability} required for the container during the test.
+         * @return {@link Stability} required for the container during the test.
+         */
+        Stability getRequiredStability();
```
jamezp, line 54: Should we default this to `Stability.DEFAULT`?

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
```diff
@@ -198,9 +248,21 @@ public ContainerConfigurationBuilder tearDownScript(List<List<String>> batches)
 
             @Override
             public ContainerConfiguration build() {
+                Stability stability = this.stability;
+                boolean skip = this.skipStabilityRequirementAndSnapshot;
                 List<List<String>> setupScript = Collections.unmodifiableList(this.setupScript);
                 List<List<String>> tearDownScript = Collections.unmodifiableList(this.tearDownScript);
```
jamezp, line 234: Not part of this PR, but this seems odd to me. We're using unmodifiable collections here, but the collection itself can be modified because there's not a defensive copy. Granted, this is a testing framework so it likely doesn't matter as much, but seems odd :)

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: @jamezp It should and it does ;-)

            private Stability stability = Stability.DEFAULT;

That's what actually uncovered the bug - https://redhat.atlassian.net/browse/WFLY-21634

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Well, this is inside build() of an anonymous builder class and nobody retains a reference to this.setupScript to mutate it... So a defensive copy would be safer, but I dont think that's worth it

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Done

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Removed completely, restore always happens

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
jamezp: Sorry, all I meant was should this be a default method which returns `Stability.DEFAULT`. I saw below it did and again maybe it doesn't matter as TBH I didn't look close enough at how this is used :)

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
jamezp: Yeah, not a big deal really. We probably don't even need the collections wrapped in an unmoodifable list :) Again, not a big deal at all, I just happened to notice when looking.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Given the org.jboss.as.test.shared.ManagementServerSetupTask.ContainerConfiguration#getRequiredStability is the only implementor I dono if we need a default impl :)

testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
```diff
@@ -69,7 +69,7 @@ public final void setup(ManagementClient managementClient, String containerId) t
         // Check the reload-enhanced operation exists in the stability level we want to load to so that
         // we can reload back to the current one
         Stability reloadOpStability = getReloadEnhancedOperationStabilityLevel(managementClient);
-        Assume.assumeTrue(desiredStability.enables(reloadOpStability));
+        Assume.assumeTrue(reloadOpStability.enables(desiredStability));
```
rhusar: @jamezp Given you are already looking at this PR, this is the bit I need reviewed the most as a fix for https://redhat.atlassian.net/browse/WFLY-21634 since I have no idea what this does and why. All I know it's wrong.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Resolving, let me know if you want me to act on this within this PR.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Resolving, let me know if you want me to act on this within this PR.

testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
rhusar: I should probably go to the source - the author of the code - @kabir if you can explain whether this is a bug or what was this code trying to do.

Basically, it's wrong because it SKIPS the test that is supportable by the given stability already without a reload. So in normal testing it won't manifest as failure, but will pass by skipping.

jamezp: I recall there being some conversation around this when it was created. I don't recall the details unfortunately.

rhusar: OK Paul pointed out that this is because the actual `reload-enhanced` operation is a separate operation (n.b. but why was this not designed as parameters at community stability of the reload op @yersan ?) that is registered at community stability, so it would not be available at default if it reloaded to it... sigh.

```java
    public static OperationDefinition getEnhancedDefinition(final LocalHostControllerInfo hostControllerInfo) {
        return new DeferredParametersOperationDefinitionBuilder(hostControllerInfo, ENHANCED_OPERATION_NAME, HostModelUtil.getResourceDescriptionResolver(), ENHANCED_PRIMARY_HC_ATTRIBUTES, ENHANCED_SECONDARY_HC_ATTRIBUTES, Stability.COMMUNITY)
                .setStability(Stability.COMMUNITY)
```

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
```diff
@@ -256,6 +291,7 @@ public ScriptBuilder endBatch() {
     }
 
     private final ContainerSetConfiguration config;
+    private final Map<String, StabilityServerSetupSnapshotRestoreTasks> snapshotRestoreTasksByContainerId = new ConcurrentHashMap<>();
```
pferraro:
```java
private final Map<String, ServerSetupTask> snapshotRestoreTasks = new ConcurrentHashMap<>();

File: testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java [hunk 7]
```diff
@@ -275,12 +311,24 @@ public ManagementServerSetupTask(Set<String> containers, ContainerConfiguration
 
     @Override
     public void setup(ManagementClient client, String containerId) throws Exception {
+        ContainerConfiguration containerConfig = this.config.getContainerConfiguration(containerId);
+        if (containerConfig != null) {
+            StabilityServerSetupSnapshotRestoreTasks snapshotRestoreTasks = new StabilityServerSetupSnapshotRestoreTasks(containerConfig.getRequiredStability());
```
pferraro:
```java
Stability requiredStability = containerConfig.getRequiredStability();
ServerSetupTask snapshotRestoreTask = Stability.DEFAULT.enables(requiredStability) ? new SnapshotServerSetupTask() : new StabilityServerSetupSnapshotRestoreTasks(requiredStability);

File: testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java [hunk 8]
```diff
@@ -275,12 +311,24 @@ public ManagementServerSetupTask(Set<String> containers, ContainerConfiguration
...
+        }
+
         this.configure(client, containerId, ContainerConfiguration::getSetupScript);
     }
 
     @Override
     public void tearDown(ManagementClient client, String containerId) throws Exception {
         this.configure(client, containerId, ContainerConfiguration::getTearDownScript);
+
+        StabilityServerSetupSnapshotRestoreTasks snapshotRestoreTasks = this.snapshotRestoreTasksByContainerId.remove(containerId);
```
pferraro:
```java
ServerSetupTask snapshotRestoreTasks = this.snapshotRestoreTasksByContainerId.remove(containerId);

pferraro (changes requested):

testsuite/shared/src/main/java/org/wildfly/test/stabilitylevel/StabilityServerSetupSnapshotRestoreTasks.java
yersan: @rhusar I also made that question some time ago: [#wildfly-developers > Stability levels and testsuite @ 💬](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Stability.20levels.20and.20testsuite/near/437420027)

Not sure if at that time we were not able to have operation parameters based on stability levels, but from the response the main reason seems to be due to the requirement to have RBAC strict permissions that can affect to the standard reload operation

pferraro: > Not sure if at that time we were not able to have operation parameters based on stability levels

I think https://github.com/wildfly/wildfly-core/pull/5897 prevented Kabir's original implementation from working at the time.

> the requirement to have RBAC strict permissions that can affect to the standard reload operation

Rereading that thread, it is unclear to me what Brian meant by:
> there's no concept of RBAC constraints based on a parameter value, or at least not ones that appear in r-r-d

This constraint would not need to be applied conditionally based on value.  AIUI, attribute access constraints are only applied to _defined_ values.
e.g.
If the `SensitiveTargetAccessConstraintDefinition.RELOAD_ENHANCED` constraint was applied to Kabir's original implementation (where an optional stability parameter was added to the reload operation) it would mean that:

`:reload(stability=FOO)` would require specific access control
whereas
`:reload` would not, since the parameter containing an access constraint is undefined.

For a server _already_ running with stability FOO, `:reload(stability=FOO)` and `:reload` are functionally equivalent, even though one requires access control but the other does not - but I don't see any problem with that.

rhusar: I have removed these changes entirely as per suggestions below; and reassigned to the original author.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Done - went with this approach.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Done.

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
rhusar: Done.

rhusar:
@pferraro All green, ready for re-review. Thx

pferraro (approved):

testsuite/shared/src/main/java/org/jboss/as/test/shared/ManagementServerSetupTask.java
bstansberry: We don't support use of this library outside this codebase so there's no need to account for other implementors.

bstansberry:
Thanks @rhusar, @jamezp and @pferraro

PR #19775, opened 2026-03-18
[WFLY-21615][WFLY-21616] Add a capability service for the MicroProfile Config created by the subsystem

kabir:
https://redhat.atlassian.net/browse/WFLY-21615
https://redhat.atlassian.net/browse/WFLY-21616

____

> [!WARNING]

> * [WFLY-21615](https://issues.redhat.com/browse/WFLY-21615)
> * [WFLY-21616](https://issues.redhat.com/browse/WFLY-21616)

### DIFF
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
@@ -6,7 +6,6 @@
 package org.wildfly.extension.microprofile.config.smallrye;
 
 import static org.jboss.as.controller.ModuleIdentifierUtil.parseCanonicalModuleIdentifier;
-
 import static org.jboss.as.controller.SimpleAttributeDefinitionBuilder.create;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.FILESYSTEM_PATH;
 import static org.jboss.as.controller.descriptions.ModelDescriptionConstants.MODULE;
@@ -17,10 +16,9 @@ import java.util.Arrays;
 import java.util.Collection;
 
 import io.smallrye.config.PropertiesConfigSource;
-
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
-import org.jboss.as.controller.AbstractAddStepHandler;
+import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.AttributeMarshaller;
 import org.jboss.as.controller.AttributeMarshallers;
@@ -34,6 +32,7 @@ import org.jboss.as.controller.ReloadRequiredRemoveStepHandler;
 import org.jboss.as.controller.SimpleAttributeDefinitionBuilder;
 import org.jboss.as.controller.SimpleResourceDefinition;
 import org.jboss.as.controller.descriptions.ModelDescriptionConstants;
+import org.jboss.as.controller.registry.OperationEntry;
 import org.jboss.as.controller.services.path.PathManager;
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
@@ -109,7 +108,9 @@ class ConfigSourceDefinition extends PersistentResourceDefinition {
         super(new SimpleResourceDefinition.Parameters(MicroProfileConfigExtension.CONFIG_SOURCE_PATH,
                 MicroProfileConfigExtension.getResourceDescriptionResolver(MicroProfileConfigExtension.CONFIG_SOURCE_PATH.getKey()))
                 .setAddHandler(new ConfigSourceDefinitionAddHandler(providers, sources))
-                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE));
+                .setAddRestartLevel(OperationEntry.Flag.RESTART_ALL_SERVICES)
+                .setRemoveHandler(ReloadRequiredRemoveStepHandler.INSTANCE)
+                .setRemoveRestartLevel(OperationEntry.Flag.RESTART_ALL_SERVICES));
     }
 
     @Override
@@ -129,7 +130,7 @@ class ConfigSourceDefinition extends PersistentResourceDefinition {
         }
     }
 
-    private static class ConfigSourceDefinitionAddHandler extends AbstractAddStepHandler {
+    private static class ConfigSourceDefinitionAddHandler extends AbstractBoottimeAddStepHandler {
         private Registry<ConfigSourceProvider> providers;
         private final Registry<ConfigSource> sources;
 
@@ -139,8 +140,9 @@ class ConfigSourceDefinition extends PersistentResourceDefinition {
             this.sources = sources;
         }
 
+
         @Override
-        protected void performRuntime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
+        public void performBoottime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
             super.performRuntime(context, operation, model);
             String name = context.getCurrentAddressValue();
             int ordinal = ORDINAL.resolveModelAttribute(context, model).asInt();
--- a/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java
+++ b/microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java
@@ -21,6 +21,7 @@ import org.jboss.as.server.AbstractDeploymentChainStep;
 import org.jboss.as.server.DeploymentProcessorTarget;
 import org.jboss.as.server.deployment.Phase;
 import org.jboss.dmr.ModelNode;
+import org.jboss.msc.Service;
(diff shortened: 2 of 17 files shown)


pferraro:
I think something went awry with your last rebase?

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/converter/MicroProfileConfigConvertersTestCase.java
```diff
@@ -32,7 +33,7 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
-@ServerSetup(SetupTask.class)
+@ServerSetup({SetupTask.class, ReloadServerSetupTask.class})
```
pferraro: Why this change? The SetupTask already reloads if required on setup/tearDown.

Rather, it would seem prudent to validate that the operation responses have the expected state.

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/app/ConfigCapabilityValidator.java
```diff
@@ -0,0 +1,86 @@
...
+            UnaryServiceDescriptor.of("org.wildfly.microprofile.config", ConfigProviderResolver.class);
+
+    // Static holder for the resolver from the capability service
+    private static volatile ConfigProviderResolver capabilityResolver;
+    private static volatile Config capabilityConfig;
+
+    @Override
+    public void activate(ServiceActivatorContext context) {
+        ServiceBuilder<?> builder = context.getServiceTarget()
+                .addService(ServiceName.of("test", "config-capability-validator"));
```
pferraro: This method is deprecated.  Why would this service need a name?

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/app/ConfigCapabilityValidator.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * ServiceActivator that validates the microprofile-config capability service is properly installed.
+ * Stores the ConfigProviderResolver from the capability in a static field so REST endpoints can access it.
+ *
+ * @author <a href="kabir.khan@jboss.com">Kabir Khan</a>
+ */
+public class ConfigCapabilityValidator implements ServiceActivator {
+
+    // The capability descriptor - matches what's defined in MicroProfileSubsystemDefinition
+    private static final UnaryServiceDescriptor<ConfigProviderResolver> CONFIG_CAPABILITY_DESCRIPTOR =
```
pferraro: What is the purpose of this here?
Normally, a service descriptor would be defined in an SPI module to be made available to consuming subsystems - and would be used to create the corresponding RuntimeCapability.
Also, this capability does not have a dynamic name, thus UnaryServiceDescriptor (i.e. a ServiceDescriptor with one dynamic name segment) is not appropriate.

pferraro (changes requested):

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
```diff
@@ -129,7 +130,7 @@ private static Class unwrapClass(ModelNode classModel) throws OperationFailedExc
         }
     }
 
-    private static class ConfigSourceDefinitionAddHandler extends AbstractAddStepHandler {
+    private static class ConfigSourceDefinitionAddHandler extends AbstractBoottimeAddStepHandler {
```
pferraro, line 133: Why change this?  Is the OperationEntry.Flag.RESTART_ALL_SERVICES operation flag not sufficient? (same question for similar changes below).

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java
```diff
@@ -68,10 +73,46 @@ public SmallRyeConfigBuilder forClassLoader(ClassLoader classLoader) {
...
+                ModelNode configSourceModel = childResource.getModel();
+
+                // Check if this is a root directory config source (uses CONFIG_SOURCE_ROOT service name)
+                // or a regular config source (uses CONFIG_SOURCE service name)
+                if (configSourceModel.hasDefined("dir", "root") &&
+                        context.resolveExpressions(configSourceModel.get("dir", "root")).asBoolean()) {
+                    sb.requires(ServiceNames.CONFIG_SOURCE_ROOT.append(name));
+                } else {
+                    sb.requires(ServiceNames.CONFIG_SOURCE.append(name));
+                }
```
pferraro, line 104: Since DirConfigSourceRegistrationService and ConfigSourceRootRegistrationService are mutually exclusive, and neither provides a value (that would otherwise require that a consumer be able to differentiate them), why use distinct ServiceNames?
e.g. Use ServiceNames.CONFIG_SOURCE.append(name) for both services and replace the above with:
```java
sb.requires(ServiceNames.CONFIG_SOURCE.append(name));
```

pferraro (changes requested):

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java
kabir: This is consuming/requiring services that are installed elsewhere.

If we want to change that, we should open a separate issue.

microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java
kabir: I've checked manually, and with my changes, it does put the server into the reload-required state.

````
[standalone@localhost:9990 /] /subsystem=microprofile-config-smallrye/config-source=test:add
{
    "outcome" => "success",
    "response-headers" => {
        "operation-requires-reload" => true,
        "process-state" => "reload-required"
    }
}
````

If I make it inherit from AbstractAddStepHandler instead, and just rely on the flags (unless I've missed something), it does not.

````
[standalone@localhost:9990 /] /subsystem=microprofile-config-smallrye/config-source=test:add
{"outcome" => "success"}

[standalone@localhost:9990 /]  /subsystem=microprofile-config-smallrye/config-source=test2:add(properties={test=123, another=456})    
{"outcome" => "success"}
````

reload-required seems to only get triggered for `:write-attribute` and `:remove`. 

I *could* rework the code to make it work, but I am missing why the AbstractBoottimeAddStepHandler change is undesirable, as it seems the easiest to me?

Comment (pferraro) [reply microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/ConfigSourceDefinition.java:133]: Whoops - I forgot that most of these abstract OperationStepHandler classes have incomplete flag handling.
All good.

File: microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java (hunk 5 above)
Comment (pferraro) [reply microprofile/config-smallrye/src/main/java/org/wildfly/extension/microprofile/config/smallrye/MicroProfileConfigSubsystemAdd.java:104]: No problem.  I opened https://redhat.atlassian.net/browse/WFLY-21730 to track this.

File: testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source_provider/ConfigSourceProviderFromClassTestCase.java [hunk 6]
```diff
@@ -35,7 +36,7 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
-@ServerSetup(SetupTask.class)
+@ServerSetup({SetupTask.class, ReloadServerSetupTask.class})
```
pferraro, line 39: Optional: This renders this line redundant:
https://github.com/wildfly/wildfly/blob/main/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/management/config_source_provider/SetupTask.java#L54

N.B. You could instead annotate SetupTask with org.jboss.as.arquillian.api.ReloadIfRequired

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/config/smallrye/multideployment/MultiWarConfigDeploymentTestCase.java
```diff
@@ -40,7 +41,7 @@
 
 @RunWith(Arquillian.class)
 @RunAsClient
-@ServerSetup(MultiWarConfigDeploymentTestCase.GlobalConfigSourceSetupTask.class)
+@ServerSetup({MultiWarConfigDeploymentTestCase.GlobalConfigSourceSetupTask.class, ReloadServerSetupTask.class})
```
pferraro, line 44: N.B. You could instead annotate SetupTask with org.jboss.as.arquillian.api.ReloadIfRequired

pferraro (approved):
Remaining comments are optional.

rhusar (approved):
LGTM with some follow up work filed as https://redhat.atlassian.net/browse/WFLY-21730

rhusar:
Thanks @kabir and @pferraro

PR #19778, opened 2026-03-19
[WFLY-21644] Bidirectional association management should be disabled if Hibernate Bytecode enhancement is used

scottmarlow:
https://redhat.atlassian.net/browse/WFLY-21644

Disable bytecode enhancement for application entities that have bidirectional associations mapped.

____

> [!WARNING]

> * [WFLY-21644](https://issues.redhat.com/browse/WFLY-21644)

### DIFF
--- a/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
+++ b/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/WildFlyClassTransformer.java
@@ -7,6 +7,7 @@ package org.jboss.as.jpa.hibernate;
 
 import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
 import org.hibernate.bytecode.enhance.spi.UnloadedClass;
+import org.hibernate.bytecode.enhance.spi.UnloadedField;
 import org.hibernate.jpa.internal.enhance.EnhancingClassTransformerImpl;
 
 /**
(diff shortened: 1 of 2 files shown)


jamezp:
@scottmarlow Can you explain a bit on why we need to return `false` here? It seems to have to do with enhanced fields if i'm understanding correctly. 

Is it because we don't know whether or not a field is enhanced?

scottmarlow:
I corrected the title to `Bidirectional association management should be disabled if Hibernate Bytecode enhancement is used`

> @scottmarlow Can you explain a bit on why we need to return `false` here? It seems to have to do with enhanced fields if i'm understanding correctly.
> 
> Is it because we don't know whether or not a field is enhanced?

https://github.com/hibernate/hibernate-orm/blob/d2d2186ec578daacff747c05a5afe69e270023e2/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/BiDirectionalAssociationHandler.java#L52 is the calling code that we return false to because we do not want the Bidirectional association management feature to be used with bytecode enhancement because it requires application code changes to have it work which isn't always possible, so it will cause problems if we do not disable the `Bidirectional association management`.

https://docs.hibernate.org/orm/7.3/userguide/html_single/#BytecodeEnhancement-dirty-tracking-bidirectional mentions an example of how application code must be written to work with this feature.

jamezp (approved):

scottmarlow:
The two failures do not look related to this change:
- The 
org.jboss.as.test.integration.messaging.jms.deployment.SingleConnectionActivationPropertyTestCase
New
testMdbRedeploymentWithSingleConnection
- org.jboss.as.test.clustering.cluster.jgroups.tls.TLSWebFailoverTestCase
New
gracefulSimpleFailover(URL, URL, URL)

PR #19784, opened 2026-03-23
[WFLY-21651] Prometheus endpoint does not return Content-Type header for Prometheus text format

beth-soptim:
https://redhat.atlassian.net/browse/WFLY-21651

____

> [!WARNING]

> * [WFLY-21651](https://issues.redhat.com/browse/WFLY-21651)

### DIFF
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
@@ -8,6 +8,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.UnaryOperator;
 
+import io.undertow.util.Headers;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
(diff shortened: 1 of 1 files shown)


jasondlee (approved):
Looks good. Thanks!

jasondlee:
Thanks @beth-soptim!

rhusar:
@beth-soptim @jasondlee @bstansberry I am confused by this (as raised on the chat last night) as there are multitude formats and not a single one.

The PR hardcodes text/plain; version=0.0.4 regardless of what the client requests. In general, a proper implementation should:

  1. Read the Accept header from the request
  2. Let Micrometer's scrape(String) handle the content negotiation (it supports both Prometheus text format and OpenMetrics application/openmetrics-text)
  3. Set the Content-Type response header to match what was actually produced

This matters because Prometheus itself sends something like `Accept: application/openmetrics-text;version=1.0.0;q=0.5,application/openmetrics-text;version=0.0.1;q=0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`, and other clients may prefer OpenMetrics format.

pferraro:
Generally, if an Accept header is present in the request, you would parse them into a collection (sorted in descending quality. i.e. 'q') and use the first acceptable media type, or return a 406 response if none are acceptable.
e.g. https://github.com/wildfly/wildfly/blob/main/microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHttpHandler.java

jasondlee:
I have filed a Jira to follow up on this, hopefully in time for WF 40, but we'll see: https://redhat.atlassian.net/browse/WFLY-21660

jasondlee:
> I have filed a Jira to follow up on this, hopefully in time for WF 40, but we'll see: https://redhat.atlassian.net/browse/WFLY-21660

I actually hadn't closed the original Jira, so I'll just do the extra work under that one.

beth-soptim:
Micrometer Content-Type logic is very simple and always tries Prometheus Protobuf format first. Also, it seems difficult to find out what format was actually returned.

I tried to find a solution that you might want to use: https://github.com/wildfly/wildfly/pull/19794

Not sure if this way is "correct" (using classes from Jakarta REST. etc.)

PR #19791, opened 2026-03-24
[IGNORE - CI ONLY] Reproducer for WFLY-21658.

rhusar:
Strives to confirm that the testsuite is broken and relies on test order.

https://redhat.atlassian.net/browse/WFLY-21658

### DIFF
--- a/testsuite/integration/expansion/pom.xml
+++ b/testsuite/integration/expansion/pom.xml
@@ -542,6 +542,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <runOrder>reversealphabetical</runOrder>
                     <!-- TODO Remove workaround for https://issues.redhat.com/browse/WFLY-20030 *ReactiveMessaging* tests fail with podman -->
                     <reuseForks>false</reuseForks>
                     <environmentVariables>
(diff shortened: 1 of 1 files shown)


rhusar:
FYI @kabir Bingo! The testsuite only passes because of how Linux sorts the files on FS by default and is picked up by surefire. This is a bug in @jasondlee 's MultipleMetricsTestCase that removes the telemetry subsystem it never added

jasondlee:
> FYI @kabir Bingo! The testsuite only passes because of how Linux sorts the files on FS by default and is picked up by surefire. This is a bug in @jasondlee 's MultipleMetricsTestCase that removes the telemetry subsystem it never added

Hrm. That test is supposed to restore the original state of the server. Maybe it should just use the snapshot server setup variant...

rhusar:
> > FYI @kabir Bingo! The testsuite only passes because of how Linux sorts the files on FS by default and is picked up by surefire. This is a bug in @jasondlee 's MultipleMetricsTestCase that removes the telemetry subsystem it never added
> 
> Hrm. That test is supposed to restore the original state of the server. Maybe it should just use the snapshot server setup variant...

See https://github.com/wildfly/wildfly/pull/19796 how I fixed this.

rhusar:
Closing this, this demonstrated the real issue. No longer needed, fix ready.

PR #19792, opened 2026-03-24
[WFLY-20093] Group webservices component upgrades for batched validation

mskacelik:
issue: https://redhat.atlassian.net/browse/WFLY-20093
Confirmed via `git grep <artifactId> | grep module.xml` (being ONLY in either the definition module or WS-related module).

**CXF core & extensions:** `org.apache.cxf`, `org.apache.cxf.services.*`, `org.apache.cxf.xjcplugins`, `org.apache.cxf.xjc-utils`
**WS-Security:** `org.apache.wss4j`, `org.apache.santuario`
**JBossWS:** `org.jboss.ws`, `org.jboss.ws.cxf`, `org.jboss.ws.projects`
**Supporting libs:** `org.apache.ws.xmlschema`, `org.apache.neethi`, `org.opensaml`, `net.shibboleth.utilities`, `org.jasypt`, `org.apache.velocity`, `wsdl4j`, `com.sun.xml.fastinfoset`.

____

> [!WARNING]

> * [WFLY-20093](https://issues.redhat.com/browse/WFLY-20093)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,21 @@ updates:
     registries:
       - maven-central
       - jboss-public-repository-group
+    groups:
+      webservices:
+        patterns:
+          - 'com.sun.xml.fastinfoset*'
+          - 'net.shibboleth.utilities*'
+          - 'org.apache.cxf*'
+          - 'org.apache.neethi*'
+          - 'org.apache.santuario*'
+          - 'org.apache.velocity*'
+          - 'org.apache.ws.xmlschema*'
+          - 'org.apache.wss4j*'
+          - 'org.jasypt*'
+          - 'org.jboss.ws*'
+          - 'org.opensaml*'
+          - 'wsdl4j*'
   - package-ecosystem: "github-actions"
     # Workflow files stored in the
     # default location of `.github/workflows`


mskacelik:
@fabiobrz Hi, can you please take a look at it, if by any chance you need some of the artifacts to still be updated by Dependabot?

rhusar (changes requested):
This makes little sense to me @bstansberry @mskacelik @fabiobrz Instead of ignoring these entirely, have you considered using dependabot groups? This would batch all webservices dependency updates into a single PR, making the side validation manageable in one pass rather than losing the automation:

```yaml
      groups:
        webservices:
          patterns:
            - "org.apache.cxf*"
            - "org.apache.wss4j*"
            - "org.apache.ws.xmlschema*"
            - "org.apache.neethi*"
            - "org.apache.santuario*"
            - "org.jboss.ws*"
            - "org.opensaml*"
            - "net.shibboleth.utilities*"
            - "org.jasypt*"
            - "org.apache.velocity*"
            - "wsdl4j*"
            - "com.sun.xml.fastinfoset*"
```

That way you still get automated upgrade PRs but as a single combined one for all WS components, instead of individual noisy ones.

You can also move these into a weekly schedule if that's what you want too.

cc @jamezp who has experience with this too

fabiobrz:
Thanks for looking after this WFLY issue @mskacelik. What @rhusar [suggested](https://github.com/wildfly/wildfly/pull/19792#pullrequestreview-4000743207) seems reasonable to me, WDYT?

jamezp:
Using groups is definitely the way to go. What it will do is group the upgrades into a single PR. It ensures that the components listed will be upgraded together when multiple are available. It also just simply cuts down on resource usage.

mskacelik:
Yea, I think it make sense, I wonder if we can in the future also split certain components into individual groups like the webservices case.

Anyway, going to change the issue's title and description appropriately to reflect the group solution.

jamezp (approved):

rhusar (approved):
LGTM now

Personally I don't understand the preference to mix single quotes and double quotes within the same file and I would have sorted that list. But that's fine.

mskacelik:
> LGTM now
> 
> Personally I don't understand the preference to mix single quotes and double quotes within the same file and I would have sorted that list. But that's fine.

Fair enough. The file was already inconsistent before my changes, so I assumed it was okay to leave as is. However, I've gone ahead and fixed both issues anyway.

fabiobrz (approved):
Thanks @mskacelik, LGTM.

.github/dependabot.yml
```diff
@@ -30,6 +30,21 @@ updates:
...
+          - 'com.sun.xml.fastinfoset:*'
+          - 'net.shibboleth.utilities:*'
+          - 'org.apache.cxf:*'
+          - 'org.apache.neethi:*'
+          - 'org.apache.santuario:*'
+          - 'org.apache.velocity:*'
+          - 'org.apache.ws.xmlschema:*'
+          - 'org.apache.wss4j:*'
+          - 'org.jasypt:*'
+          - 'org.jboss.ws:*'
```
rhusar: @mskacelik As I was about to merge this, and now that it's sorted, it's more visible - is the omission of the rest of the groupIds intentional? I don't think so?

```
  - 'org.jboss.ws.cxf:*'
  - 'org.jboss.ws.projects:*'
```

rhusar (changes requested):

.github/dependabot.yml
mskacelik: Yes, it's supposed to be there. I used your code block, in which these two dependencies were omitted, but I didn't notice the missing colon (`:`). This makes it a completely different pattern. I'm going to fix it in the next push.

Edit: I actually left it simple, removed the colon, so now `org.jboss.ws.projects*` and `org.jboss.ws.projects*` be part of the `org.jboss.ws.*` set.

rhusar: That block was meant as an example how to use groups, right...

The theoretical problem with this approach is that   org.jasypt* — could match any org.jasyptxyz... groupId.

In practice this is unlikely to cause issues since Dependabot only matches against dependencies actually declared in the project's POMs, so overly broad patterns won't pull in anything that isn't already a dependency.

So, I think that's fine. We can change it at any time.

rhusar (approved):
The broadened patterns now properly cover org.jboss.ws.cxf:* and org.jboss.ws.projects:*.

Lets see how this will perform.

rhusar:
Thanks @mskacelik and @fabiobrz @jamezp for the reviews

rhusar:
> Thanks @mskacelik and @fabiobrz @jamezp for the reviews

Damn, this may have caused us to hit the limit on dependabot timeout. Previously it took around 38 minutes, now we hit the 55 minute limit and it failed. But there is quite a big variability, so perhaps it's just maven slowness.

There are other problems with unresolvable property, I ll try to fix that.

We might need to split dependencies into multiple dependabot iterations, e.g. move this group to a separate run and exlude from main run.

Just FYI @bstansberry

rhusar:
Fix fix is https://github.com/wildfly/wildfly/pull/19805 that might help speed up by avoiding the undefined property resolution by Dependabot.

rhusar:
> Fix fix is #19805 that might help speed up by avoiding the undefined property resolution by Dependabot.

https://github.com/wildfly/wildfly/pull/19805#issuecomment-4161589855 - lets see.

PR #19794, opened 2026-03-25
[WFLY-21651] Handle different Media Types

beth-soptim:
https://redhat.atlassian.net/browse/WFLY-21651

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
@@ -15,6 +15,7 @@
 
     <dependencies>
         <module name="io.micrometer"/>
+        <module name="io.prometheus"/>
         <module name="io.undertow.core"/>
         <module name="org.wildfly.micrometer.deployment"/>
 
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.extension.micrometer.prometheus;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+record MediaType(String type, String subtype, Map<String, String> parameters) implements Comparable<MediaType> {
+
+    MediaType(String type, String subtype, Map<String, String> parameters) {
+        this.type = type;
+        this.subtype = subtype;
+        this.parameters = Map.copyOf(parameters);
+    }
+
+    private static final String WILDCARD = "*";
+
+    static MediaType parse(String input) {
+        String[] parts = input.split(";");
+        String[] types = parts[0].trim().split("/");
+
+        String type = types[0].trim();
+        type = type.isEmpty() ? WILDCARD : type;
+        String subtype = types.length > 1 ? types[1].trim() : WILDCARD;
+
+        Map<String, String> params = new HashMap<>();
+        params.put("q", "1.0");
+        for (int i = 1; i < parts.length; i++) {
+            String param = parts[i].trim();
+            if (param.isEmpty()) {
+                continue;
+            }
+
+            String[] kv = param.split("=", 2);
+            String key = kv[0].trim();
+            String value = kv.length > 1 ? kv[1].trim().replace("\"", "") : "";
+
+            if ("q".equalsIgnoreCase(key)) {
+                try {
+                    double q = Double.parseDouble(value);
+                    // https://datatracker.ietf.org/doc/html/rfc9110#section-12.4.2
+                    q = Math.max(0, Math.min(1, q));
+                    params.put(key, Double.toString(q));
+                } catch (NumberFormatException ignored) {
+                }
+            } else {
+                params.put(key, value);
+            }
+        }
+
+        return new MediaType(type, subtype, Map.copyOf(params));
+    }
+
+    boolean matches(String otherType, String otherSubtype) {
+        return (type.equalsIgnoreCase(otherType) || WILDCARD.equals(type) || WILDCARD.equals(otherType)) &&
+                (subtype.equalsIgnoreCase(otherSubtype) || WILDCARD.equals(subtype) || WILDCARD.equals(otherSubtype));
+    }
+
+    boolean hasParameters(Map<String, String> parameters, Set<String> parameterKeysToIgnore) {
+        return parameters().entrySet().containsAll(parameters.entrySet().stream().filter(e -> !parameterKeysToIgnore.contains(e.getKey())).collect(Collectors.toSet()));
+    }
+
+    private int specificity() {
+        if (WILDCARD.equals(type) && WILDCARD.equals(subtype)) {
+            return 0;
+        }
+        if (WILDCARD.equals(subtype)) {
+            return 1;
+        }
+        return 2;
+    }
+
+    @Override
+    public int compareTo(MediaType o) {
+        int qCompare = Double.compare(Double.parseDouble(o.parameters.getOrDefault("q", "1.0")),
+                Double.parseDouble(this.parameters.getOrDefault("q", "1.0")));
+        if (qCompare != 0) {
+            return qCompare;
+        }
+        return Integer.compare(o.specificity(), this.specificity());
+    }
+
+    @Override
+    public String toString() {
+        return asHeaderString() + ";q=" + parameters.get("q");
+    }
+
+    String asHeaderString() {
+        return type + "/" + subtype + parametersStringWithoutQ();
+    }
+
+    private String parametersStringWithoutQ() {
+        if (parameters.isEmpty()) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            if (entry.getKey() != null && entry.getValue() != null && !"q".equalsIgnoreCase(entry.getKey())) {
+                sb.append(";").append(entry.getKey()).append("=").append(entry.getValue());
+            }
+        }
+        return sb.toString();
+    }
+
+}
--- a/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
+++ b/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
@@ -4,13 +4,23 @@
  */
 package org.wildfly.extension.micrometer.prometheus;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
 
 import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
+import io.prometheus.metrics.expositionformats.OpenMetricsTextFormatWriter;
+import io.prometheus.metrics.expositionformats.PrometheusProtobufWriter;
+import io.prometheus.metrics.expositionformats.PrometheusTextFormatWriter;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.HeaderValues;
 import io.undertow.util.Headers;
+import io.undertow.util.StatusCodes;
 import org.jboss.as.controller.AttributeDefinition;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
@@ -68,8 +78,10 @@ public class PrometheusRegistryDefinitionRegistrar implements ChildResourceDefin
             .build();
     public static final Collection<AttributeDefinition> ATTRIBUTES = List.of(CONTEXT, SECURITY_ENABLED);
 
-    // https://prometheus.io/docs/instrumenting/exposition_formats/#prometheus-text-format
-    private static final String PROMETHEUS_CONTENT_TYPE = "text/plain; version=0.0.4; charset=utf-8";
+    // Formats supported in io.prometheus.metrics.expositionformats.ExpositionFormats
+    private static final MediaType PROMETHEUS_TEXT_MEDIA_TYPE = MediaType.parse(PrometheusTextFormatWriter.CONTENT_TYPE);
+    private static final MediaType OPENMETRICS_TEXT_MEDIA_TYPE = MediaType.parse(OpenMetricsTextFormatWriter.CONTENT_TYPE);
+    private static final MediaType PROMETHEUS_PROTOBUF_MEDIA_TYPE = MediaType.parse(PrometheusProtobufWriter.CONTENT_TYPE);
 
     @Override
     public ManagementResourceRegistration register(ManagementResourceRegistration parent, ManagementResourceRegistrationContext context) {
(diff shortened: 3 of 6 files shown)


rhusar (changes requested):
@beth-soptim Causes a failure in org.wildfly.test.integration.observability.micrometer.MicrometerPrometheusTestCase in case you did not see the results, but they should be publicly available.

galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
```diff
@@ -15,13 +15,15 @@
...
         <module name="io.micrometer"/>
+        <module name="io.prometheus"/>
         <module name="io.undertow.core"/>
         <module name="org.wildfly.micrometer.deployment"/>
 
         <module name="java.logging"/>
         <module name="java.desktop"/>
         <module name="java.management"/>
         <module name="jakarta.enterprise.api" />
+        <module name="jakarta.ws.rs.api"/>
```
bstansberry: @jasondlee Does the micrometer layer currently have a direct or transitive dependency on the jaxrs layer? Are there module deps between its modules and RESTEasy / Jakarta REST?

If not we need to think carefully before adding that kind of dependency.

@jamezp fyi

jamezp: +1 this will be good to know. I'm not really keen on adding a dependency to `org.jboss.resteasy.resteasy-core` just to use the `org.jboss.resteasy.util.MediaTypeHelper`. Especially since the method being used is not that sophisticated.

bstansberry: I don't see any such module deps, so this would bring in a lot to a slimmed server that otherwise doesn't use Jakarta REST.

This is what jakarta.ws.rs.api brings in:

https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/jakarta/ws/rs/api/main/module.xml

So, unlike many Jakarta API modules, which are no big deal, it brings in the impl.

Then resteasy-core brings a ton more: https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-core/main/module.xml

More from resteasy-client: https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/jboss/resteasy/resteasy-client/main/module.xml

Lots more if we continue down the dependency tree, including org.glassfish.jaxb.

So, I think if we need utility code we should fork it.

bstansberry: BTW, thanks @beth-soptim for this! It looks like a nice improvement.

jamezp: Agreed. The only thing that method does is:
```java
public static List<MediaType> parseHeader(String header) {
    ArrayList<MediaType> types = new ArrayList<MediaType>();
    String[] medias = header.split(",");
    for (int i = 0; i < medias.length; i++) {
        types.add(MediaType.valueOf(medias[i].trim()));
    }
    return types;
}
```

Bring in the `org.jboss.resteasy.resteasy-core` module, and all its dependencies, for that feels overkill. It could even be reduced to:
```java
public static List<MediaType> parseHeader(String header) {
    return Stream.of(header.split(","))
            .map((media) -> MediaType.valueOf(media.trim()))
            .collect(Collectors.toList());
}
```

jasondlee: > @jasondlee Does the micrometer layer currently have a direct or transitive dependency on the jaxrs layer? Are there module deps between its modules and RESTEasy / Jakarta REST?
> 
> If not we need to think carefully before adding that kind of dependency.
> 
> @jamezp fyi

I don't think there is. Adding this module dep, as discussed, does very much seem to be overkill. I think James' suggestion should work fine, if that's all that's needed.

beth-soptim:
I removed Jakarta REST dependency, Resteasy dependency and fixed the NPE.

The latest test failures seem to be unrelated?

rhusar:
> I removed Jakarta REST dependency, Resteasy dependency and fixed the NPE.
> 
> The latest test failures seem to be unrelated?

Yes, we have a bunch of CI flakiness with many/most of which have fixes already in the pipeline, so bear with us.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -102,16 +111,70 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+                sendMetrics(metricsMediaType, prometheusRegistry, exchange);
+                sent = true;
+                break;
+            }
+        }
+        if (!sent) {
+            if (headerValues== null || headerValues.isEmpty()) {
+                sendMetrics(PROMETHEUS_TEXT_MEDIA_TYPE, prometheusRegistry, exchange);
+            } else {
+                exchange.setStatusCode(StatusCodes.UNSUPPORTED_MEDIA_TYPE);
```
rhusar: Is this coming from the spec? I hope not? The StatusCodes.UNSUPPORTED_MEDIA_TYPE (415) is for rejecting the request body's content type. When the server cannot satisfy the Accept header, the correct response is 406 Not Acceptable.

> The HTTP 415 Unsupported Media Type [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#client_error_responses) status code indicates that the server refused to accept the request because the message [content](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_Content) format is not supported.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -102,16 +111,70 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+        for (MediaType mediaType : mediaTypes) {
+            MediaType metricsMediaType = getMetricsMediaType(mediaType);
+            if (metricsMediaType != null) {
+                sendMetrics(metricsMediaType, prometheusRegistry, exchange);
+                sent = true;
+                break;
+            }
+        }
+        if (!sent) {
+            if (headerValues== null || headerValues.isEmpty()) {
```
rhusar: Run formatter on this line (missing space).

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -102,16 +111,70 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+        } else if (OPENMETRICS_TEXT_MEDIA_TYPE.matches(mediaType.type(), mediaType.subtype())) {
+            metricsMediaType = OPENMETRICS_TEXT_MEDIA_TYPE;
+        } else if (PROMETHEUS_PROTOBUF_MEDIA_TYPE.matches(mediaType.type(), mediaType.subtype())) {
+            metricsMediaType = PROMETHEUS_PROTOBUF_MEDIA_TYPE;
+        }
+        return metricsMediaType;
+    }
+
+    private void sendMetrics(MediaType mediaType, WildFlyPrometheusRegistry prometheusRegistry, HttpServerExchange exchange) {
+        String metrics = prometheusRegistry.scrape(mediaType.toString());
```
rhusar: The brotobuf media format is binary and as this is populating a String the prometheusRegistry.scrape supposedly does the transformation. This should be verified this works as intended. Did we? Ideally trough a test.

rhusar (changes requested):

jasondlee (approved):
Looks much better. Thanks for the test, too. That's often overlooked. :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        if (WILDCARD.equals(subtype)) {
+            return 1;
+        }
+        return 2;
+    }
+
+    @Override
+    public int compareTo(MediaType o) {
+        int qCompare = Double.compare(Double.parseDouble(o.parameters.get("q")), Double.parseDouble(this.parameters.get("q")));
```
rhusar: If "q" is undefined, this will NPE.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+    public String asHeaderString() {
+        return type + "/" + subtype + parametersStringWithoutQ();
+    }
+
+    private String parametersStringWithoutQ() {
+        if (parameters.isEmpty()) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
```
rhusar, line 101: HashMap iteration order is non-deterministic, so the parameter order may differ from the original request – could this ever be a problem @jasondlee ? Hopefully not.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+
+    static MediaType parse(String input) {
+        String[] parts = input.split(";");
+        String[] types = parts[0].trim().split("/");
+
+        String type = types[0].trim();
+        type = type.isEmpty() ? WILDCARD : type;
+        String subtype = types.length > 1 ? types[1].trim() : WILDCARD;
+
+        Map<String, String> params = new HashMap<>();
```
rhusar, line 30: Mutable in a record class.

This should be wrapped at the end in Map.copyOf or Collections.unmodifiableMap

rhusar (changes requested):
Also, when done, you should ideally squash these into a single commit 👍🏼

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -102,16 +112,70 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+                            exchange -> handleRequest(exchange, prometheusRegistry)
                     );
                 })
                 .startWhen(StartWhen.INSTALLED)
                 .build();
     }
 
+    private void handleRequest(HttpServerExchange exchange, WildFlyPrometheusRegistry prometheusRegistry) {
+        HeaderValues headerValues = exchange.getRequestHeaders().get(Headers.ACCEPT);
+        List<MediaType> mediaTypes = headerValues != null ? getSortedMediaTypes(String.join(", ", headerValues)) : List.of();
```
pferraro, line 122: This logic (i.e. Accept header parsing, a MediaType object, etc) is already implemented in resteasy, which is already a dependency of this module. Why not just use that?

pferraro (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
jasondlee: The resteasy dep was added then removed in this PR, as it pulled in WAY too much other stuff for a slimmed server, for example.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+        return type + "/" + subtype + parametersStringWithoutQ();
+    }
+
+    private String parametersStringWithoutQ() {
+        if (parameters.isEmpty()) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            if (!"q".equals(entry.getKey())) {
```
jasondlee: Actually, a potential NPE here as well:

```
if (!entry.getKey().equals("q")) {
```

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+
+        String type = types[0].trim();
+        type = type.isEmpty() ? WILDCARD : type;
+        String subtype = types.length > 1 ? types[1].trim() : WILDCARD;
+
+        Map<String, String> params = new HashMap<>();
+        params.put("q", "1.0");
+        for (int i = 1; i < parts.length; i++) {
+            String param = parts[i].trim();
+            if (param.isEmpty()) continue;
```
jasondlee: Curlies around if block (for safety, clarity, etc and consistency with WF coding standards)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+
+            String[] kv = param.split("=", 2);
+            String key = kv[0].trim();
+            String value = kv.length > 1 ? kv[1].trim() : "";
+
+            if (value.startsWith("\"") && value.endsWith("\"") && value.length() >= 2) {
+                value = value.substring(1, value.length() - 1);
+            }
```
jasondlee: ?

```
String value = kv.length > 1 ? kv[1].trim().replace("\"", "")  : "";
```

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
jasondlee: `.getOrDefault("q", 0)` ?

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
```diff
@@ -0,0 +1,99 @@
...
+        if (qCompare != 0) {
+            return qCompare;
+        }
+
+        return Integer.compare(o.specificity(), this.specificity());
+    }
+
+    @Override
+    public String toString() {
+        return asHeaderString() + ";q=" + parameters.get("q");
```
jasondlee, line 89: Seems a little weird that "q" is stripped in `asHeaderString()` then added back, but `toString()` is mostly for logging/debugging, I guess, so any perf impact should be low and limited.

jasondlee (changes requested):
Actually... found a few things... :)

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/MediaType.java
beth-soptim: According to https://datatracker.ietf.org/doc/html/rfc2046#section-1 (referenced by RFC 9110) "the ordering of parameters is not significant."

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
pferraro: Oh, sorry.  I saw resteasy in the pom, and assumed it was used at runtime.  In that case, you can removed it from here: https://github.com/wildfly/wildfly/blob/main/observability/micrometer/pom.xml#L138
I should think that Undertow would have similar header parsers/abstractions that can be reused?

jasondlee: I do not see anything in the Undertow codebase for header parsing like this. There's `Headers`, but that's just a class with a lot of `public static final String` declarations. I haven't found a public API for parsing a header string. I'm sure there's code to do that, but it doesn't appear to be intended to be used externally.

jasondlee:
@beth-soptim Can you resolve the conflicts on this? Today is the deadline for WF 40 Beta PRs, so if this is going to make the beta, today's the day (assuming the schedule holds).

Thanks!

beth-soptim:
> @beth-soptim Can you resolve the conflicts on this? Today is the deadline for WF 40 Beta PRs, so if this is going to make the beta, today's the day (assuming the schedule holds).
> 
> Thanks!

Done!

jasondlee:
> > @beth-soptim Can you resolve the conflicts on this? Today is the deadline for WF 40 Beta PRs, so if this is going to make the beta, today's the day (assuming the schedule holds).
> > Thanks!
> 
> Done!

Thanks! :)

jasondlee (approved):
Nothing else is jumping out at me.

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -108,6 +115,63 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+        } else if (OPENMETRICS_TEXT_MEDIA_TYPE.matches(mediaType.type(), mediaType.subtype())) {
+            metricsMediaType = OPENMETRICS_TEXT_MEDIA_TYPE;
+        } else if (PROMETHEUS_PROTOBUF_MEDIA_TYPE.matches(mediaType.type(), mediaType.subtype())) {
+            metricsMediaType = PROMETHEUS_PROTOBUF_MEDIA_TYPE;
+        }
+        return metricsMediaType;
+    }
+
+    private void sendMetrics(MediaType mediaType, WildFlyPrometheusRegistry prometheusRegistry, HttpServerExchange exchange) {
+        String metrics = prometheusRegistry.scrape(mediaType.asHeaderString());
```
rhusar, line 151: The matching is wrong. After it's matched, it still needs to use the original request Accept header.

Think about this: this ignores all provided headers, e.g. `proto=io.prometheus.client.MetricFamily`

https://prometheus.io/docs/instrumenting/exposition_formats/#prometheus-protobuf-format

...and that's why I asked for the verification and test case :-)

rhusar (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
beth-soptim: Got the error. Will try to fix this after easter holidays

beth-soptim: Committed my changes.

The parameters are now checked (version and charset are ignored because Prometheus doesn't care either)

rhusar (changes requested):
@beth-soptim Can you please squash all commits? Probably 1 commit for this change is enough.

jasondlee (approved):
Thanks for all the hard work on this!

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
```diff
@@ -108,6 +117,55 @@ public ResourceServiceInstaller configure(OperationContext context, ModelNode mo
...
+            }
+        }
+    }
+
+    private static Optional<MediaType> getMetricsMediaType(MediaType mediaType) {
+        return Stream.of(PROMETHEUS_TEXT_MEDIA_TYPE, OPENMETRICS_TEXT_MEDIA_TYPE, PROMETHEUS_PROTOBUF_MEDIA_TYPE)
+                .filter(mmt -> mmt.matches(mediaType.type(), mediaType.subtype())
+                        // ignoring q (only relevant for priority), version and charset (micrometer-registry-prometheus is doing the same)
+                        && mediaType.hasParameters(mmt.parameters(), Set.of("q", "version", "charset")))
+                .findAny();
```
rhusar: Why `findAny`? This should be `findFirst` to make the priority explicit.

rhusar (changes requested):

observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java
beth-soptim: Fixed

rhusar (approved):
LGTM; cool stuff @beth-soptim 

@jasondlee should we eventually add a test case to verify the protobuf output format? It's simply too much to ask from the @beth-soptim within this PR.

jasondlee (approved):

jasondlee:
Thanks again, @beth-soptim!

jasondlee:
> LGTM; cool stuff @beth-soptim
> 
> @jasondlee should we eventually add a test case to verify the protobuf output format? It's simply too much to ask from the @beth-soptim within this PR.

It probably couldn't hurt, but feels pretty low priority, to be candid. :)

PR #19796, opened 2026-03-25
WFLY-21658 Fix the 'expansion' test suite is not portable

rhusar:
Resolves
https://redhat.atlassian.net/browse/WFLY-21658

Proved by https://github.com/wildfly/wildfly/pull/19791 that this is a test ordering issue due to tests not cleaning up.

Developers have turned to me for advice about the ‘expansion’ test suite not working on macOS nor podman.

I discovered a lot of problem with this TS in the latest tests, especially MultipleMetricsTestCase

1. containtainers have hard coded ‘localhost' which makes assumptions about containerization networking which is wrong
2. order of the test is required which is different per OS - works with order of fresh checkout on linux but does not work on others, slipping this to random or reversealphabetical will cause tests to fail
3. tests are NOT cleaning after themselves correctly - this is the core bug.

With these changes tests pass on e.g. macOS and with podman.

____

> [!WARNING]

> * [WFLY-21658](https://issues.redhat.com/browse/WFLY-21658)

### DIFF
--- a/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
+++ b/testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
@@ -24,8 +24,7 @@ import org.jboss.as.test.shared.observability.setuptasks.MicrometerSetupTask;
 import org.jboss.as.test.shared.observability.setuptasks.OpenTelemetrySetupTask;
 import org.jboss.as.test.shared.logging.LoggingUtil;
 import org.jboss.dmr.ModelNode;
-import org.jboss.shrinkwrap.api.ShrinkWrap;
-import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.as.test.shared.util.AssumeTestGroupUtil;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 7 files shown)


testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/observability/MultipleMetricsTestCase.java
```diff
@@ -32,9 +38,16 @@
...
 @RunWith(Arquillian.class)
 @RunAsClient
+@ServerSetup(SnapshotServerSetupTask.class)
 public class MultipleMetricsTestCase {
     private static final String MESSAGE_PREAMBLE = "Additional metrics systems discovered";
 
+    // Dummy deployment to trigger @ServerSetup
+    @Deployment(testable = false)
+    public static WebArchive createDummyDeployment() {
+        return ShrinkWrap.create(WebArchive.class, MultipleMetricsTestCase.class.getSimpleName() + ".war").addAsWebResource(new StringAsset(""), "index.html");
```
bstansberry: This is fine, but FYI re AssumeTestGroupUtil.emptyWar() and similar methods.

The comment at L35-45 in that class explains why they are in that class.

rhusar: Thanks @bstansberry for the tip

I just used the pattern used in this module, but I agree it's suboptimal. I have added a new comment that fixes this across this entire module 64c6e142199bd1ece0073c491cef17713bf8d0f8

jasondlee (approved):

rhusar:
Thanks @jasondlee and @bstansberry !

PR #19799, opened 2026-03-26
[WFLY-19876] Deprecate the jaxrs subsystems resteasy-2-0-request-matching attribute

Dkafetzis:
Issue link: https://redhat.atlassian.net/browse/WFLY-19876

____

> [!WARNING]

> * [WFLY-19876](https://issues.redhat.com/browse/WFLY-19876)

### DIFF
--- a/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsAttribute.java
+++ b/jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsAttribute.java
@@ -100,6 +100,8 @@ public abstract class JaxrsAttribute {
             new SensitivityClassification(JaxrsExtension.SUBSYSTEM_NAME, "tracing-management", false, false, true)
     );
 
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    @Deprecated
     public static final SimpleAttributeDefinition JAXRS_2_0_REQUEST_MATCHING = new SimpleAttributeDefinitionBuilder(JaxrsConstants.JAXRS_2_0_REQUEST_MATCHING, ModelType.BOOLEAN)
             .setRequired(false)
             .setAllowExpression(true)
@@ -108,6 +110,7 @@ public abstract class JaxrsAttribute {
             .setAttributeGroup(RESTEASY_PARAMETER_GROUP)
             .setAttributeMarshaller(AttributeMarshallers.SIMPLE_ELEMENT)
             .setAttributeParser(AttributeParsers.SIMPLE_ELEMENT)
+            .setDeprecated(JaxrsExtension.JaxrsSubsystemModel.VERSION_6_0_0.getVersion(), true)
             .setRestartAllServices()
             .build();
 
--- a/jaxrs/src/main/resources/org/jboss/as/jaxrs/LocalDescriptions.properties
+++ b/jaxrs/src/main/resources/org/jboss/as/jaxrs/LocalDescriptions.properties
@@ -36,6 +36,7 @@ jaxrs.deployment.sub-resource-locators.java-method=Java method of the Jakarta RE
 jaxrs.deployment.sub-resource-locators.resource-methods=Annotated methods for the Jakarta RESTful Web Services resource.
 
 jaxrs.jaxrs-2-0-request-matching=In searching for a matching resource method with which to respond to a request, consider only resource methods with the best match for the request path.
+jaxrs.jaxrs-2-0-request-matching.deprecated=This attribute has been deprecated and has no effect. It will be removed in a future release.
 jaxrs.resteasy-add-charset=If a resource method returns a text/* or application/xml* media type without an explicit charset, RESTEasy will add "charset=UTF-8" to the returned Content-Type header. Note that the charset defaults to UTF-8 in this case, independent of the setting of this parameter.
 jaxrs.resteasy-buffer-exception-entity=Upon receiving an exception, the client side buffers any response entity before closing the connection.
 jaxrs.resteasy-disable-html-sanitizer=Normally, a response with media type "text/html" and a status of 400 will be processed so that the characters "/", "<", ">", "&", """ (double quote), and "'" (single quote) are escaped to prevent an XSS attack. Setting this parameter to "true", escaping will not occur.


jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
```diff
@@ -6,6 +6,7 @@
 
 public class JaxrsConstants {
 
+    @Deprecated
```
jamezp:
```suggestion
    @Deprecated(forRemoval = true, since = "40")
```

jaxrs/src/main/java/org/jboss/as/jaxrs/logging/JaxrsLogger.java
```diff
@@ -199,4 +199,8 @@ void baseTypeMethodFailed(String defaultValue, String param, String method,
     @Message(id = 31, value = "Failed to load RESTEasy MicroProfile Configuration: %s")
     void failedToLoadConfigurationFactory(String msg);
 
+    @LogMessage(level = WARN)
+    @Message(id = 32, value = "The '%s' attribute is deprecated and has no effect. Please remove it from your configuration.")
```
jamezp:
```suggestion
    @Message(id = 32, value = "The '%s' attribute is deprecated and has no effect. It will be removed in a future release.")
```

jaxrs/src/main/resources/org/jboss/as/jaxrs/LocalDescriptions.properties
```diff
@@ -36,6 +36,7 @@ jaxrs.deployment.sub-resource-locators.java-method=Java method of the Jakarta RE
 jaxrs.deployment.sub-resource-locators.resource-methods=Annotated methods for the Jakarta RESTful Web Services resource.
 
 jaxrs.jaxrs-2-0-request-matching=In searching for a matching resource method with which to respond to a request, consider only resource methods with the best match for the request path.
+jaxrs.jaxrs-2-0-request-matching.deprecated=This attribute has been deprecated and has no effect. Please remove it from your configuration.
```
jamezp:
```suggestion
jaxrs.jaxrs-2-0-request-matching.deprecated=This attribute has been deprecated and has no effect. It will be removed in a future release.
```

jamezp (changes requested):

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
```diff
@@ -6,6 +6,7 @@
 
 public class JaxrsConstants {
 
+    @Deprecated(forRemoval = true, since = "40")
```
bstansberry: What @jamezp says about this should overrule me if he disagrees, but personally I don't think we should do this kind of thing with management API. The vast majority of the time these kinds of fields are internal details of functionality and the maintainers of that functionality don't benefit from being told the internal details are going to change when the functionality is removed. It just adds IDE and build noise.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
```diff
@@ -97,6 +97,9 @@ public void execute(DeploymentProcessorTarget processorTarget) {
     @SuppressWarnings("deprecation")
     private static JaxrsServerConfig createServerConfig(ModelNode configuration, OperationContext context) throws OperationFailedException {
         final JaxrsServerConfig config = new JaxrsServerConfig();
+        if (configuration.hasDefined(JaxrsAttribute.JAXRS_2_0_REQUEST_MATCHING.getName())) {
+            JaxrsLogger.JAXRS_LOGGER.jaxrs20RequestMatchingDeprecated(JaxrsAttribute.JAXRS_2_0_REQUEST_MATCHING.getName());
+        }
```
bstansberry: I don't think this is needed because the setDeprecated on the attribute triggers core functionality that logs about the deprecation if the attribute is set.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsConstants.java
jamezp: That makes sense to me and I'm fine with simply removing the `@Deprecated` annotation in that case.

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsSubsystemAdd.java
Dkafetzis: Ok, should the  `jaxrs20RequestMatchingDeprecated` log method also be removed?

jamezp: Yes, I think it makes sense to remove it. Thanks for the reminder on that @bstansberry

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsAttribute.java
```diff
@@ -108,6 +110,7 @@ public void parseElement(final AttributeDefinition attribute, final XMLExtendedS
             .setAttributeGroup(RESTEASY_PARAMETER_GROUP)
             .setAttributeMarshaller(AttributeMarshallers.SIMPLE_ELEMENT)
             .setAttributeParser(AttributeParsers.SIMPLE_ELEMENT)
+            .setDeprecated(ModelVersion.create(6, 0, 0), true)
```
rhusar: Don't create random objects that already exist

org.jboss.as.jaxrs.JaxrsExtension.JaxrsSubsystemModel#VERSION_6_0_0.getVersion()

rhusar (changes requested):

jaxrs/src/main/java/org/jboss/as/jaxrs/JaxrsAttribute.java
Dkafetzis: Nice catch, didn't know about that. Updated to fix it.

rhusar (approved):
LGTM; all review feedback has been addressed.

jamezp (approved):

bstansberry:
Thanks @Dkafetzis, @rhusar and @jamezp

PR #19820, opened 2026-03-31
[WFLY-21647] Implict security roles missing from Web App MetaData.

darranl:
https://redhat.atlassian.net/browse/WFLY-21647

The servlet specification describes @DeclareRoles handling as:

"This annotation is used to define the security roles that comprise the security model of the application. This annotation is specified on a class, and it is used to define roles that could be tested (i.e., by calling isUserInRole) from within the methods of the annotated class. Roles that are implicitly declared as a result of their use in a @RolesAllowed need not be explicitly declared using the @DeclareRoles annotaion."

The handling of implicitly defined roles was missing.

____

> [!WARNING]

> * [WFLY-21647](https://issues.redhat.com/browse/WFLY-21647)

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/deployment/WarAnnotationDeploymentProcessor.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/deployment/WarAnnotationDeploymentProcessor.java
@@ -390,19 +390,13 @@ public class WarAnnotationDeploymentProcessor implements DeploymentUnitProcessor
         // @DeclareRoles
         final List<AnnotationInstance> declareRolesAnnotations = index.getAnnotations(declareRoles);
         if (declareRolesAnnotations != null && !declareRolesAnnotations.isEmpty()) {
-            SecurityRolesMetaData securityRoles = metaData.getSecurityRoles();
-            if (securityRoles == null) {
-               securityRoles = new SecurityRolesMetaData();
-               metaData.setSecurityRoles(securityRoles);
-            }
+            SecurityRolesMetaData securityRoles = getSecurityRolesMetaData(metaData);
             for (final AnnotationInstance annotation : declareRolesAnnotations) {
                 if (annotation.value() == null) {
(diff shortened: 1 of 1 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/deployment/WarAnnotationDeploymentProcessor.java
```diff
@@ -603,4 +601,22 @@ protected DescriptionGroupMetaData getDescriptionGroup(String description, Strin
...
+        SecurityRolesMetaData securityRoles = metaData.getSecurityRoles();
+        if (securityRoles == null) {
+            securityRoles = new SecurityRolesMetaData();
+            metaData.setSecurityRoles(securityRoles);
+        }
+        return securityRoles;
+    }
+
+    private static void addSecurityRole(SecurityRolesMetaData securityRoles, String role) {
+        if (role == null || role.isEmpty() || "*".equals(role) || "**".equals(role)) {
```
bstansberry, line 614: @darranl Please add a comment explaining why wildcards are ignored.

bstansberry (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/deployment/WarAnnotationDeploymentProcessor.java
bstansberry: I talked to Darran and the answer is we're only trying to add the explicitly named roles.

bstansberry: A comment is a nice-to-have; this thread on the PR is also a form of 'documentation' since I won't resolve it / hide it in the GH UI.

bstansberry (approved):

bstansberry:
Thanks @darranl

PR #19821, opened 2026-03-31
[WFLY-21637] Ensure the Jakata Authorization Context-ID is set for ServletContextListeners.

darranl:
https://redhat.atlassian.net/browse/WFLY-21637

____

> [!WARNING]

> * [WFLY-21637](https://issues.redhat.com/browse/WFLY-21637)

### DIFF
--- a/undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/JACCContextIdHandler.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/JACCContextIdHandler.java
@@ -14,6 +14,7 @@ import io.undertow.server.HandlerWrapper;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
 import io.undertow.server.handlers.PredicateHandler;
+import io.undertow.servlet.api.ThreadSetupHandler;
 import io.undertow.servlet.predicate.DispatcherTypePredicate;
 import org.wildfly.security.manager.WildFlySecurityManager;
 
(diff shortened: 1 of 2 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/security/jacc/JACCContextIdHandler.java
```diff
@@ -74,22 +49,69 @@ public String run() {
...
+        return new ThreadSetupHandler() {
+
+            @Override
+            public <T, C> Action<T, C> create(final Action<T, C> action) {
+                return new Action<T, C>() {
+                    @Override
+                    public T call(HttpServerExchange exchange, C context) throws Exception {
+                        String previousContextID = null;
+                        try {
+                            previousContextID = setContextID(setContextIdAction);
```
bstansberry, line 102: Maybe a comment explaining why there is no SM check here? Reviewing this I find it confusing to see that in the finally block but not in the try.

bstansberry: I see this is just following the existing pattern from the HandlerWrapper case, so never mind. It would be nice to describe why no SM check for both, but whatever, we've been using this idiom for a long time already.

bstansberry (approved):

bstansberry:
This LGTM. It's just applying the existing logic for Context-Id handling used for requests to the thread setup actions used during deployment/undeployment, when a ServletContextListener would be invoked.

bstansberry:
Thanks @darranl

PR #19833, opened 2026-04-01
[WFLY-21701]: SingleConnectionActivationPropertyTestCase is flaky.

ehsavoie:
* Fixing memory leak
* Waiting for the sessions to open in 1 connection scenarios.

Jira: https://redhat.atlassian.net/browse/WFLY-21701

____

> [!WARNING]

> * [WFLY-21701](https://issues.redhat.com/browse/WFLY-21701)

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/jms/deployment/SingleConnectionActivationPropertyTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/messaging/jms/deployment/SingleConnectionActivationPropertyTestCase.java
@@ -28,6 +28,7 @@ import jakarta.jms.Queue;
 import jakarta.json.Json;
 import jakarta.json.JsonArray;
 import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
 import org.awaitility.Awaitility;
 import org.jboss.arquillian.container.test.api.Deployer;
 import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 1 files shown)


rhusar (commented):
@ehsavoie So the idea is that the fix treats it as a test timing issue (sessions are still being created). Looking at the error log — AMQ219000: ClientSession closed while creating session — that WARN from Artemis suggests some sessions fail entirely. So the idea is that the Artemis RA   retry logic will be triggered so the sessions do eventually get   created, just not instantly as hte test originally anticipated?

ehsavoie:
@rhusar yes, because the trick is that with the 'classic' scenario you create 1 connection which holds one session so when all the connections are created the sessions are ready while here once the connection is created sessions are still being asynchronously created. So we wait a bit for all of them to be ready

mnovak:
@ehsavoie the problem is that with such a low load there should be no issue/exception when creating sessions. I tried to reproduce it locally with 1000 sessions (for maxSession, test is using 15) in multiple MDB but could not hit it any problem. Might it be that CI is really slow/overloaded?

ehsavoie:
@mnovak I don't think the issue is with load but with the fact that the sessions are all created on a single connection thus when the connection is up all the sessions might not be ready. I could reproduce locally

rhusar:
@ehsavoie The explanation for the exception is still missing though — the server log shows:

```
AMQ219000: ClientSession closed while creating session
```

Is this a known issue with Artemis 2.52.0 in the `singleConnection` code path? Should we consider this PR a workaround and track the root cause upstream?

rhusar:
> @ehsavoie the problem is that with such a low load there should be no issue/exception when creating sessions. I tried to reproduce it locally with 1000 sessions (for maxSession, test is using 15) in multiple MDB but could not hit it any problem. Might it be that CI is really slow/overloaded?

If there is any asynchronicity involved, you can never make an assumption about when the actual test expectations are met and you always need to use tooling like awaitility.

The question is still whether there should be asynchronicity involved here - if we agree that this is a workaround for AMQ219000 and there is a tracker for the actual exception I am OK merging this.

mnovak:
> If there is any asynchronicity involved, you can never make an assumption about when the actual test expectations are met and you always need to use tooling like awaitility.

thats' fine, awaitility should be used here. But It's not what I meant. Throwing any WARN is unacceptable unless the server would be out of resources (CPU, Memory, threads, ...) which would prevent to create the new session. I don't believe it's happening here and we should investigate why the WARN occured. 

@ehsavoie It might be that WARN occured as a result of server being shutdown (Artemis refusing new sessions) due to the failed assertion? Test fails -> server shutdowns while sessions are being created -> WARN occur. Could you add a check that no WARN/ERROR occur in server log, please? Otherwise we hide the WARN problem.

rhusar:
> > If there is any asynchronicity involved, you can never make an assumption about when the actual test expectations are met and you always need to use tooling like awaitility.
> 
> thats' fine, awaitility should be used here. But It's not what I meant. Throwing any WARN is unacceptable unless the server would be out of resources (CPU, Memory, threads, ...) which would prevent to create the new session. I don't believe it's happening here and we should investigate why the WARN occured.
> 
> @ehsavoie It might be that WARN occured as a result of server being shutdown (Artemis refusing new sessions) due to the failed assertion? Test fails -> server shutdowns while sessions are being created -> WARN occur. Could you add a check that no WARN/ERROR occur in server log, please? Otherwise we hide the WARN problem.

That's what this comment is about - https://github.com/wildfly/wildfly/pull/19833#issuecomment-4175842538. 

Nevertheless, Emmanuel is working on confirming this is really an Artemis bug.

When we have that, the question remains how we address this in main ASAP - with ignore approach PR or this awaitility approach PR + upstream Jira.

ehsavoie:
https://redhat.atlassian.net/browse/WFLY-21701?focusedCommentId=16613854

ehsavoie:
I think this is the best fix @rhusar @mnovak

rhusar (approved):
OK then @ehsavoie lets go with this then, it certainly doesn't hurt, yet I am still not 100% convinced about the exception, but not my domain really to judge that.

rhusar:
Thanks @ehsavoie @mnovak

PR #19859, opened 2026-04-10
[WFLY-21738] [WFLY-20061] Enable OpenAPI for LRA Coordinator

marcosgopen:
https://redhat.atlassian.net/browse/WFLY-20061
https://redhat.atlassian.net/browse/WFLY-21738

Deployed services can be scanned by openapi but the  LRA coordinator is embedded in WildFly and so the OpenApi scan does not include it.
To enable the OpenApi for it I register a new service called 'LRACoordinatorOpenAPIService' to build the OpenApi model for the LRA coordinator and add it to the OpenAPIModelRegistry

### DIFF
--- a/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
+++ b/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
@@ -9,16 +9,19 @@ import org.jboss.as.controller.AbstractBoottimeAddStepHandler;
 import org.jboss.as.controller.CapabilityServiceBuilder;
 import org.jboss.as.controller.OperationContext;
 import org.jboss.as.controller.OperationFailedException;
+import org.jboss.as.controller.ServiceNameFactory;
 import org.jboss.as.server.Services;
 import org.jboss.dmr.ModelNode;
 import org.jboss.msc.service.ServiceController;
 import org.jboss.tm.XAResourceRecoveryRegistry;
 import org.wildfly.extension.microprofile.lra.coordinator._private.MicroProfileLRACoordinatorLogger;
+import org.wildfly.extension.microprofile.lra.coordinator.service.LRACoordinatorOpenAPIService;
 import org.wildfly.extension.microprofile.lra.coordinator.service.LRACoordinatorService;
 import org.wildfly.extension.microprofile.lra.coordinator.service.LRARecoveryService;
 import org.wildfly.extension.undertow.Capabilities;
 import org.wildfly.extension.undertow.Host;
 import org.wildfly.extension.undertow.UndertowService;
+import org.wildfly.microprofile.openapi.OpenAPIModelRegistry;
 
 import java.util.Arrays;
 import java.util.concurrent.ExecutorService;
@@ -27,6 +30,8 @@ import java.util.function.Supplier;
 import static org.wildfly.extension.microprofile.lra.coordinator.MicroProfileLRACoordinatorSubsystemDefinition.ATTRIBUTES;
 
 class MicroProfileLRACoordinatorAdd extends AbstractBoottimeAddStepHandler {
+
+    private static final String OPENAPI_CAPABILITY_NAME = "org.wildfly.microprofile.openapi";
 
     MicroProfileLRACoordinatorAdd() {
         super(Arrays.asList(ATTRIBUTES));
@@ -38,6 +43,14 @@ class MicroProfileLRACoordinatorAdd extends AbstractBoottimeAddStepHandler {
 
         registerRecoveryService(context);
         registerCoordinatorService(context, model);
+
+        ModelNode serverNode = MicroProfileLRACoordinatorSubsystemDefinition.SERVER.resolveModelAttribute(context, model);
+        ModelNode hostNode = MicroProfileLRACoordinatorSubsystemDefinition.HOST.resolveModelAttribute(context, model);
+        if (serverNode.isDefined() && hostNode.isDefined()
+                && context.hasOptionalCapability(OPENAPI_CAPABILITY_NAME,
+                        MicroProfileLRACoordinatorSubsystemDefinition.LRA_COORDINATOR_CAPABILITY_NAME, null)) {
+            registerOpenAPIService(context, serverNode.asString(), hostNode.asString());
+        }
 
         MicroProfileLRACoordinatorLogger.LOGGER.activatingSubsystem();
     }
@@ -64,6 +77,18 @@ class MicroProfileLRACoordinatorAdd extends AbstractBoottimeAddStepHandler {
 
         builder.setInstance(lraCoordinatorService);
         builder.setInitialMode(ServiceController.Mode.ACTIVE).install();
+    }
+
+    private void registerOpenAPIService(final OperationContext context, final String serverName, final String hostName) {
+        CapabilityServiceBuilder builder = context.getCapabilityServiceTarget().addService();
+
+        Supplier<OpenAPIModelRegistry> registrySupplier = builder.requires(
+            ServiceNameFactory.resolveServiceName(OpenAPIModelRegistry.SERVICE_DESCRIPTOR, serverName, hostName));
+
+        builder.requiresCapability(MicroProfileLRACoordinatorSubsystemDefinition.LRA_COORDINATOR_CAPABILITY_NAME, null);
+
+        builder.setInstance(new LRACoordinatorOpenAPIService(registrySupplier, LRACoordinatorService.CONTEXT_PATH));
+        builder.setInitialMode(ServiceController.Mode.PASSIVE).install();
     }
 
     private void registerRecoveryService(final OperationContext context) {
--- a/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
+++ b/microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
@@ -47,9 +47,16 @@ class MicroProfileLRACoordinatorAdd extends AbstractBoottimeAddStepHandler {
             .addCapability(MicroProfileLRACoordinatorSubsystemDefinition.LRA_COORDINATOR_CAPABILITY);
 
         builder.requiresCapability(Capabilities.CAPABILITY_UNDERTOW, UndertowService.class);
-        String serverModelValue = MicroProfileLRACoordinatorSubsystemDefinition.SERVER.resolveModelAttribute(context, model).asString();
-        String hostModelValue = MicroProfileLRACoordinatorSubsystemDefinition.HOST.resolveModelAttribute(context, model).asString();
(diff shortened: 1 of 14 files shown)


xstefank:
I think this should be reviewed by someone on WFLY side since this is not LRA specific - cc/ @bstansberry

xstefank (approved):
I'm just not sure if this is how OpenAPI is used in WFLY. But I don't see issues with the PR.

marcosgopen:
@pferraro I can see this PR is causing some OpenAPI tests to fail. This is probably because they don't expect the LRA Coordinator to register its model. I might also create a property to enable the LRA Coordinator OpenAPI model registration so that I can disable it for OpenApi tests?

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorSubsystemDefinition.java
```diff
@@ -72,6 +80,7 @@ public class MicroProfileLRACoordinatorSubsystemDefinition extends SimpleResourc
     public void registerAttributes(ManagementResourceRegistration resourceRegistration) {
         resourceRegistration.registerReadWriteAttribute(SERVER, null, new ReloadRequiredWriteAttributeHandler(SERVER));
         resourceRegistration.registerReadWriteAttribute(HOST, null, new ReloadRequiredWriteAttributeHandler(HOST));
+        resourceRegistration.registerReadWriteAttribute(ENABLE_OPENAPI, null, new ReloadRequiredWriteAttributeHandler(ENABLE_OPENAPI));
```
pferraro, line 80: This new attribute necessitates a new management model version for this subsystem, as well as a new XML schema version.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
```diff
@@ -32,13 +35,21 @@ class MicroProfileLRACoordinatorAdd extends AbstractBoottimeAddStepHandler {
...
+
     @Override
     protected void performBoottime(OperationContext context, ModelNode operation, ModelNode model) throws OperationFailedException {
         super.performBoottime(context, operation, model);
 
         registerRecoveryService(context);
         registerCoordinatorService(context, model);
 
+        boolean enableOpenAPI = MicroProfileLRACoordinatorSubsystemDefinition.ENABLE_OPENAPI.resolveModelAttribute(context, model).asBoolean();
+        if (enableOpenAPI && context.hasOptionalCapability(OPENAPI_CAPABILITY_NAME,
```
pferraro: Presumably, a value of `true` should _require_ the OpenAPI capability.
To do this, you will need to conditionally register the requirement on the OpenAPI capability during OperationContext.Stage.MODEL.

This will require that you fix the missing capability references in the "server" and "host" attributes (see below), as the presence of these capabilities is required to ensure that the associated OpenAPIModelRegistry is available at runtime.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorSubsystemDefinition.java
pferraro, line 47: This attribute should reference the capability described by Server.SERVICE_DESCRIPTOR.

pferraro: This attribute must not define a default value, as it references (albeit implicitly, currently) a capability.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorSubsystemDefinition.java
```diff
@@ -59,7 +59,15 @@ public class MicroProfileLRACoordinatorSubsystemDefinition extends SimpleResourc
             .setDefaultValue(new ModelNode(Constants.DEFAULT_HOST))
```
pferraro: This attribute must not define a default value, as it references (albeit implicitly, currently) a capability.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorSubsystemDefinition.java
pferraro, line 54: This attribute should reference the capability described by Host.SERVICE_DESCRIPTOR.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
```diff
@@ -59,6 +70,22 @@ private void registerCoordinatorService(final OperationContext context, final Mo
...
+        String hostModelValue = MicroProfileLRACoordinatorSubsystemDefinition.HOST.resolveModelAttribute(context, model).asString();
+
+        Supplier<OpenAPIModelRegistry> registrySupplier = builder.requires(
+            ServiceNameFactory.resolveServiceName(OpenAPIModelRegistry.SERVICE_DESCRIPTOR, serverModelValue, hostModelValue));
+
+        // Ensure the coordinator is deployed before registering its OpenAPI model
+        builder.requiresCapability(MicroProfileLRACoordinatorSubsystemDefinition.LRA_COORDINATOR_CAPABILITY_NAME, null);
+
+        builder.setInstance(new LRACoordinatorOpenAPIService(registrySupplier, LRACoordinatorService.CONTEXT_PATH));
+        builder.setInitialMode(ServiceController.Mode.ACTIVE).install();
```
pferraro: Why start this service actively?

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorSubsystemDefinition.java
pferraro: A capability referencing attribute must not allow expressions - as it otherwise prevents model validation.

pferraro: A capability referencing attribute must not allow expressions - as it otherwise prevents model validation.

pferraro:
> @pferraro I can see this PR is causing some OpenAPI tests to fail. This is probably because they don't expect the LRA Coordinator to register its model. I might also create a property to enable the LRA Coordinator OpenAPI model registration so that I can disable it for OpenApi tests?

Those failures are likely exacerbated the combination of: 
* https://redhat.atlassian.net/browse/WFLY-21648
* The actively installed service introduced by this PR which increases the probability of WFLY-21648.

marcosgopen:
Thank you very much @pferraro for your review. I will update the PR shortly.

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
marcosgopen: Good point, I will start it in PASSIVE mode

testsuite/integration/expansion/src/test/java/org/wildfly/test/integration/microprofile/lra/EnableLRAExtensionsSetupTask.java
```diff
@@ -31,7 +31,10 @@ protected void doSetup(final ManagementClient managementClient, final String con
             builder.addStep(Operations.createAddOperation(Operations.createAddress("extension", MODULE_LRA_COORDINATOR)));
         }
         if (!subsystems.contains(SUBSYSTEM_LRA_COORDINATOR)) {
-            builder.addStep(Operations.createAddOperation(Operations.createAddress("subsystem", SUBSYSTEM_LRA_COORDINATOR)));
+            final ModelNode addOp = Operations.createAddOperation(Operations.createAddress("subsystem", SUBSYSTEM_LRA_COORDINATOR));
+            addOp.get("server").set("default-server");
```
marcosgopen, line 35: @pferraro as I cannot define default values I should always remember to set them, is it correct?

microprofile/lra/coordinator/src/main/resources/schema/wildfly-microprofile-lra-coordinator_1_0.xsd
```diff
@@ -26,4 +26,4 @@
     <xs:attribute name="server" type="xs:string" use="optional" default="default-server"/>
     <xs:attribute name="host" type="xs:string" use="optional" default="default-host"/>
```
marcosgopen, line 27: @pferraro what about these defaults for the attributes? are them not allowed as mentioned here: https://redhat.atlassian.net/browse/WFLY-21738 as well?

marcosgopen: I can see that when I remove the default value for the attributes there is an error log that I suppose is causing many tests to fail:
```
13:23:17,988 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([("subsystem" => "microprofile-lra-coordinator")]) - Stage: (VERIFY) - failure description: {
          "WFLYCTL0412: Required services that are not installed:" => ["org.wildfly.undertow.host.undefined.undefined"],
          "WFLYCTL0180: Services with missing/unavailable dependencies" => ["service org.wildfly.microprofile.lra.coordinator is missing [org.wildfly.undertow.host.undefined.undefined]"]
      }
```
cc @pferraro

marcosgopen:
I think this PR should wait for https://github.com/wildfly/wildfly/pull/19907 to be merged first. However I have updated the PR so that the LRA Coordinator exposes OpenApi when the lra subsystem is present and the OPENAPI_CAPABILITY_NAME capability is available. This means we don't require an extra property which means a schema update. 
cc @pferraro

microprofile/lra/coordinator/src/main/java/org/wildfly/extension/microprofile/lra/coordinator/MicroProfileLRACoordinatorAdd.java
marcosgopen: Tests are passing locally for me, but I suspect in CI 'LRACoordinatorOpenAPITestCase' is not passing because of the PASSIVE mode as the service might be not up when tested. I'll try to make the test resilient to it.

marcosgopen:
@bstansberry I can see you added the label 'missing-req'. Do you think it still applies to this PR? If so, what are the requirements missing? Thanks

PR #19883, opened 2026-04-15
WFLY-21754 Move jackson-bom import to end of dependencyManagement in common-ee and common-expansion BOMs

rhusar:
https://redhat.atlassian.net/browse/WFLY-21754

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -7,7 +7,7 @@
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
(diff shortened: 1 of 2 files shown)


rhusar:
n.b. the question was whether there is a real issue here, the answer is not that I am aware of, but it's future proofing and so that people don't accidentally proliferate this mistake

pferraro:
Against what is this future-proofing?

rhusar:
> Against what is this future-proofing?

This is future-proofing against human error as someone adding an individual dependency version management entry for e.g. a Jackson artifact (e.g. to pin or override a specific module version) after the jackson-bom import. With the BOM declared inline among the regular artifacts, the BOM import would silently win over any individual `<dependency>` entry declared later in the `<dependencyManagement>` since Maven uses first-declaration-wins appach. Moving the BOM type import to the **end** ensures any individual artifact entries above it take precedence, which is the expected behavior and consistent with how the other BOM imports (like wildfly-core) are already ordered. This also synchronizes the comments with the adjacent BOMs so that people don't make this mistake again. Case in point - existence of this PR.

bstansberry (approved):

bstansberry:
Thanks @rhusar

bstansberry:
Note that I had some trouble setting up all the WildFly EE 10 stuff related to this concept (I forget the details), which is why I added the stuff Rado is adapting to.

TBH a lot of this related to our own boms, where, for example bom/standard-ee imports 3 of our other boms, basically assembling pieces of shared content together. That can go awry whether those boms come first or not, since if the same artifact appears in two pieces it means someone did something incorrect. But if we enforce a logically consistent ordering, if things go awry at least we'll get a predictable result.

External boms are kind of a different case, since if a bom says version 1.2.3 overriding one thing in it to 2.0.1 is probably wrong -- assuming said bom really should be authoritative for everything it declares. But saying some boms should go at the end but ignoring that for others is just confusing.

PR #19889, opened 2026-04-16
[WFLY-21762] Remove requirement for the wildfly channel to the wildfly-ee channel

jmesnil:
When the channel is generated (in galleon-pack/galleon-feature-pack), do not add a requirement for the org.wildfly:wildfly-ee-galleon-pack feature pack.

When the channel is edited (in galleon-pack/channel) before it is attached to the the project, do not add a requirement for the org.wildfly.channels:wildfly-ee channel.

This fixes https://redhat.atlassian.net/browse/WFLY-21762.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -65,8 +65,22 @@
         <!-- Depends on the org.wildfly.channels:wildfly artifact to ensure
              that this channel is built first before the distribution is created -->
         <dependency>
-            <groupId>org.wildfly.channels</groupId>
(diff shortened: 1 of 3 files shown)


bstansberry:
We instead went with the approach of adding a new 'wildfly-compat-ee-10' channel for use with wildfly-ee-10 FP + wildfly FP provisioning: https://github.com/bstansberry/wildfly/pull/67.

So, I'm closing this.

Thanks @jmesnil -- perhaps keep this branch around for a few days in case we find some flaw with the other approach.

PR #19890, opened 2026-04-16
[WFLY-21753] Do not rely on process state when starting the servers i…

yersan:
…n the mixed domain test suite

Jira issue: https://redhat.atlassian.net/browse/WFLY-21753


bstansberry:
Thanks @yersan

PR #19898, opened 2026-04-20
WFLY-21777 Ensure RemoteCache is configured to use the marshaller expected by the HotRodUserManagerFactory implementation.

pferraro:
https://redhat.atlassian.net/browse/WFLY-21777

### DIFF
--- a/clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/sso/hotrod/HotRodUserManagementProvider.java
+++ b/clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/sso/hotrod/HotRodUserManagementProvider.java
@@ -5,22 +5,35 @@
 
 package org.wildfly.extension.clustering.web.sso.hotrod;
 
+import java.net.URI;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
+import org.infinispan.client.hotrod.DataFormat;
 import org.infinispan.client.hotrod.RemoteCache;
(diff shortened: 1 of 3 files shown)


clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/sso/hotrod/HotRodUserManagementProvider.java
rhusar, line 46: Any reason why not to make encoding explicit as we do in `org.wildfly.extension.clustering.web.session.hotrod.HotRodSessionManagementProvider#DEFAULT_CONFIGURATION`? 

```suggestion
    "distributed-cache": {
        "encoding" : {
            "key" : {
                "media-type" : "application/octet-stream"
            },
            "value" : {
                "media-type" : "application/octet-stream"
            }
        },
```

rhusar (changes requested):

clustering/web/extension/src/main/java/org/wildfly/extension/clustering/web/sso/hotrod/HotRodUserManagementProvider.java
pferraro: Yeah - this should be fixed.  Good catch.

rhusar (approved):
Feedback addressed, LGTM now.

rhusar:
CI failure unrelated; doesn't conflict with the EE work; and the fix is important – I ll merge this now.

rhusar:
Thanks @pferraro

PR #19904, opened 2026-04-20
[WFLY-21719] Add WebParsingDeploymentProcessor support for https://ja…

bstansberry:
…karta.ee/xml/ns/jakartaee/web-app_6_1.xsd

https://redhat.atlassian.net/browse/WFLY-21719


scottmarlow (approved):

rhusar (approved):
LGTM

Though the style of the same repeated null check is pretty odd but preexisting or why not just `else if ("6.1".equals(webMetaData.getVersion()))`...

rhusar:
Thanks @bstansberry and @scottmarlow

PR #19905, opened 2026-04-21
[WFLY-21783] Make io.prometheus module dependency on org.wildfly.exte…

yersan:
…nsion.micrometer optional

Jira issue: https://redhat.atlassian.net/browse/WFLY-21783

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
@@ -15,7 +15,8 @@
 
     <dependencies>
         <module name="io.micrometer"/>
-        <module name="io.prometheus"/>
+        <!-- This is only required if the 'registry=prometheus' resource is configured -->
+        <module name="io.prometheus" optional="true"/>
         <module name="io.undertow.core"/>
         <module name="org.wildfly.micrometer.deployment"/>
 


galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
```diff
@@ -15,7 +15,7 @@
 
     <dependencies>
         <module name="io.micrometer"/>
-        <module name="io.prometheus"/>
+        <module name="io.prometheus" optional="true"/>
```
bstansberry, line 19: This should be preceding by a comment explaining why it is optional (i.e. "Only needed if 'foo' is configured".

bstansberry (changes requested):
Beyond my specific comment below, this *may* be insufficient, although maybe there is already code that does what I'll suggest.

The resource definition for the resource whose use in the config results in this no longer being optional must configure the requirement for this package, so Galleon knows it is no longer optional. Is this handled?

yersan:
@bstansberry 

> so Galleon knows it is no longer optional. Is this handled?

- At layer level is specified here: 

https://github.com/wildfly/wildfly/blob/05cda62419936a377e67596c3e3c35959b478d63/galleon-pack/galleon-shared/src/main/resources/layers/standalone/micrometer/layer-spec.xml#L24

- At management model level is here:

https://github.com/wildfly/wildfly/blob/398f8b335c145bd27e882c9982291475221a5027/observability/micrometer/src/main/java/org/wildfly/extension/micrometer/prometheus/PrometheusRegistryDefinitionRegistrar.java#L96-L98

The resource is only registered for Community stability level, returning null if it is not valid for the current one, which skip the RuntimePackageDependency requirement if it is not valid for the current stability level only.

I think that's enough to advertise how Galleon should behave, wdyt?

galleon-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/wildfly/extension/micrometer/main/module.xml
yersan: Done

bstansberry:
> I think that's enough to advertise how Galleon should behave, wdyt?

Yes, thanks. The resourceRegistration.registerAdditionalRuntimePackages thing is what I was looking for.

bstansberry (approved):

rhusar (approved):
LGTM; thanks for adding the comment for the next guy.

rhusar:
Thanks @yersan @bstansberry

PR #19907, opened 2026-04-22
WFLY-21648 Per-host services installed by OpenAPI subsystem can read MP Config before config sources are registered

pferraro:
https://redhat.atlassian.net/browse/WFLY-21648

Also, OpenAPI document may contain missing Server entries
https://redhat.atlassian.net/browse/WFLY-21736

### DIFF
--- a/microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
+++ b/microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
@@ -56,7 +56,7 @@ public class OpenAPIHostServiceInstallerProvider implements HostServiceInstaller
                 }
                 // Collect listeners of the server associated with this host, in case we require them
                 Set<String> listeners = new TreeSet<>();
-                Resource serverResource = context.readResourceFromRoot(context.getCurrentAddress().getParent());
+                Resource serverResource = context.readResourceFromRoot(context.getCurrentAddress().getParent(), false);
                 for (String childType : serverResource.getChildTypes()) {
                     for (String childName : serverResource.getChildrenNames(childType)) {
                         // Determine if this child resource is a listener
--- a/microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
+++ b/microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
@@ -5,21 +5,27 @@
 
 package org.wildfly.microprofile.openapi.host;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
+import io.smallrye.common.function.Functions;
 import io.smallrye.openapi.spi.OASFactoryResolverImpl;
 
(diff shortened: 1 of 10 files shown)


microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
```diff
@@ -41,17 +50,34 @@ public ResourceServiceInstaller getServiceInstaller(String serverName, String ho
...
-                    if (configuration.isEnabled()) {
-                        installers.add(new HostOpenAPIProviderServiceInstaller(configuration));
-
-                        installers.add(new OpenAPIHttpHandlerServiceInstaller(configuration));
+                if ((hostCapability == null) || !context.hasOptionalCapability(OpenAPIModelProvider.SUBSYSTEM_SERVICE_DESCRIPTOR, hostCapability, null)) {
+                    return Functions.discardingConsumer();
+                }
+                // Collect listeners of the server associated with this host, in case we require them
+                Set<String> listeners = new TreeSet<>();
+                Resource serverResource = context.readResourceFromRoot(context.getCurrentAddress().getParent());
```
bstansberry: This can be readResourceFromRoot(address, false) because that will provide the children. It provides placeholder Resources that have no model and no children. It does that to support this kind of use case.

This is NBD in this context but I like to avoid any calls to this method without the 'false' if they are not needed. I want to avoid people learning the wrong idiom from looking at working code.

pferraro: Thanks for the reminder.  Fixed.

undertow/src/main/java/org/wildfly/extension/undertow/UndertowListener.java
```diff
@@ -6,13 +6,16 @@
...
 import org.jboss.as.network.SocketBinding;
+import org.wildfly.service.descriptor.BinaryServiceDescriptor;
 
 /**
  * Represents the externally accessible interface provided by Undertow's listeners
  *
  * @author Stuart Douglas
  */
 public interface UndertowListener {
+    // TODO Relocate this interface to an SPI module
```
marcosgopen, line 17: @pferraro is this TODO for a follow-up PR?

pferraro: Yes. There is a Jira for creating an undertow SPI module, which is a WIP.

bstansberry:
@kabir Can you have a look at the WFLY-21648 commit, please? (Currently it's the 3rd commit.)

kabir (approved):
@bstansberry looks good

microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
bstansberry: @pferraro You didn't change this. :)

bstansberry (approved):

bstansberry:
Thanks @pferraro and @kabir

microprofile/openapi-smallrye/host/src/main/java/org/wildfly/microprofile/openapi/host/OpenAPIHostServiceInstallerProvider.java
pferraro: Hmm - did I forget to push that?  Sorry about that.  I will file a follow-up fix.

PR #19929, opened 2026-04-29
[WFLY-21326] Use a stronger algorithm and key size as enforced by Java 26

darranl:
https://redhat.atlassian.net/browse/WFLY-21326

Replaces PR https://github.com/wildfly/wildfly/pull/19507


rhusar (approved):
LGTM

JDK Commit that introuced checking: https://github.com/openjdk/jdk/commit/e544cd992099ef905266610c2c1456705cdc4587

That old fix was just setting the flag with still using old algo.

There are remaining occurences but somehow they don't cause failure?

```
testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/ejb/ssl/GenerateJBossStores.java:    private static final String SHA_1_RSA = "SHA1withRSA";
testsuite/integration/ws/src/test/java/org/jboss/as/test/integration/ws/wsse/GenerateJKSAndPropertiesFiles.java:    private static final String SHA_1_RSA = "SHA1withRSA";
testsuite/integration/ws/src/test/java/org/jboss/as/test/integration/ws/wsse/GenerateSignEncryptFiles.java:    private static final String SHA_1_RSA = "SHA1withRSA";
```

@darranl Should those be updated too at some point?

ropalka (approved):
Verified locally that algorithm change solved the problem.

PR #19934, opened 2026-04-30
[WFLY-21813] Move ejb3 subsystem services for incoming remote EJB invocations to the “remote” resource

rachmatowicz:
This PR does the following:
- moves certain classes to the "remote" resource and from the "remote" resource so that the "remote" resource may be removed from the server configuration while preserving the ability to make local invocations on EJB beans of type @Local and @Remote 
- extracts local EJB discovery from AsociationService and places it in a new service LocalEJBDiscoverProviderService
- tidy up some service installations

PR link: https://redhat.atlassian.net/browse/WFLY-21813


ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
```diff
@@ -48,12 +48,15 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
+        // only add local discovery service dependency if the context is configured to use the local EJB receiver & we are not app client
         final EJBClientDescriptorMetaData ejbClientDescriptorMetaData = deploymentUnit.getAttachment(Attachments.EJB_CLIENT_METADATA);
         final boolean useLocalReceiver = ejbClientDescriptorMetaData == null || ejbClientDescriptorMetaData.isLocalReceiverExcluded() != Boolean.TRUE;
-        final Supplier<AssociationService> associationServiceSupplier = (useLocalReceiver && !appClient) ? builder.requires(AssociationService.SERVICE_NAME) : null;
-        final DiscoveryService discoveryService = new DiscoveryService(discoveryConsumer, remotingProfileServiceSupplier, associationServiceSupplier);
+        // add in local discovery
+        final Supplier<LocalEJBDiscoveryProviderService> localDiscoveryServiceSupplier = (useLocalReceiver && !appClient) ? builder.requires(LocalEJBDiscoveryProviderService.SERVICE_NAME) : null;
+        final DiscoveryService discoveryService = new DiscoveryService(discoveryConsumer, remotingProfileServiceSupplier, localDiscoveryServiceSupplier);
```
pferraro: This class is an MSC service - did you mean to install it?
Otherwise, its lifecycle methods will never be invoked.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
```diff
@@ -0,0 +1,187 @@
...
+import org.wildfly.discovery.impl.MutableDiscoveryProvider;
+import org.wildfly.discovery.spi.DiscoveryProvider;
+import org.wildfly.discovery.spi.DiscoveryRequest;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+
+public class LocalEJBDiscoveryProviderService implements Service<LocalEJBDiscoveryProviderService> {
```
pferraro: org.jboss.msc.service.Service is deprecated.  Do not use.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public static final ServiceName SERVICE_NAME = ServiceName.JBOSS.append("ejb", "localDiscovery");
+
+    private final Supplier<ServerEnvironment> serverEnvironmentSupplier;
+    private final Supplier<DeploymentRepository> deploymentRepositorySupplier;
+    private final List<Supplier<Registry>> clientMappingsRegistriesSupplier;
+
+    private final Object serviceLock = new Object();
+    private final MutableDiscoveryProvider mutableDiscoveryProvider = new MutableDiscoveryProvider();
```
pferraro, line 50: Why use a mutable implementation?  ServiceURL is already volatile.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/noremote/LocalInvocationWithRemoteResourceRemovedTestCase.java
```diff
@@ -0,0 +1,125 @@
...
+        public void tearDown(final ManagementClient managementClient, final String containerId) throws Exception {
+            // restore the service=remote resource
+            List<String> connectors = new ArrayList<String>();
+            connectors.add("http-remoting-connector");
+            // chnanel creation options
+            Map<String, Map.Entry<String, String>> channelCreationOptions = new HashMap<String, Map.Entry<String, String>>();
+            channelCreationOptions.put("MAX_OUTBOUND_MESSAGES", Map.entry("1234", "remoting"));
+            EJBManagementUtil.createRemoteResource(managementClient, connectors, "default", true, "ejb", channelCreationOptions);
+            logger.info("The ejb subsystem remote resource has been restored");
+        }
```
bstansberry: This is a good use case for using SnapshotRestoreSetupTask. Recreating an OOTB resource in a testsuite method is vulnerable to falling out of date and affecting later tests.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -528,12 +493,34 @@ public void checkAccessAllowed() throws NamingException {
...
+    private static void addOutgoingRemoteInvocationServices(final OperationContext context,
+                                                            final ModelNode ejbSubsystemModel,
+                                                            final Resource ejbSubsystemResource,
+                                                            final boolean appclient) throws OperationFailedException {
 
         final ServiceTarget serviceTarget = context.getCapabilityServiceTarget();
-        //add the default EjbClientContext
 
+        // add in the local discovery provider
+        final ServiceBuilder<LocalEJBDiscoveryProviderService> localDiscoveryProviderBuilder = (ServiceBuilder<LocalEJBDiscoveryProviderService>) serviceTarget.addService(LocalEJBDiscoveryProviderService.SERVICE_NAME);
```
pferraro: This method is deprecated. Do not use.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -528,12 +493,34 @@ public void checkAccessAllowed() throws NamingException {
...
+            final List<ModelNode> connectors = remoteModel.get(CONNECTORS).asList();
+            for (ModelNode connector : connectors) {
+                String connectorName = connector.asString();
+                final Supplier<Registry> registrySupplier = localDiscoveryProviderBuilder.requires(ServiceNameFactory.resolveServiceName(EJB3RemoteResourceDefinition.CLIENT_MAPPINGS_REGISTRY, connectorName));
+                clientMappingsRegistrySupplierList.add(registrySupplier);
+            }
+        }
+        final LocalEJBDiscoveryProviderService localDiscoveryProviderService = new LocalEJBDiscoveryProviderService(serverEnvironmentSupplier, deploymentRepositorySupplier, clientMappingsRegistrySupplierList);
+        localDiscoveryProviderBuilder.setInstance(localDiscoveryProviderService);
+        localDiscoveryProviderBuilder.setInitialMode(ServiceController.Mode.LAZY);
```
pferraro: Why LAZY?

pferraro (changes requested):

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
rachmatowicz: Fixed.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
rachmatowicz: This useage came from the original implementation before my refactoring. I beleive it is to permit updating the discovery provider instance when the ServceURL needs to be reconstructed due to chages in deploments or client mappings.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
rachmatowicz: Fixed.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
rachmatowicz: Changed to ON_DEMAND as it's only use is as a dependency of DiscoveryService.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
rachmatowicz: Which class are you refering to? The LocalEJBDiscoveryProviderService has now been refactored.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/noremote/LocalInvocationWithRemoteResourceRemovedTestCase.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
pferraro: Basically, this service has 2 volatile references which it manipulates together.  One of these is redundant.
Let's track that clean-up in a follow-up jira.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
```diff
@@ -0,0 +1,195 @@
...
+        });
+
+        // provide the discovery provider to services that need it
+        discoveryProviderConsumer.accept(mutableDiscoveryProvider);
+    }
+
+    @Override
+    public void stop(StopContext context) {
+        // cease to provide the discovery provider
+        discoveryProviderConsumer.accept(null);
```
pferraro, line 189: N.B. This consumer has no effect within Service.stop(...), since the MSC will dereference provided values upon completion of this method.
That said, doing this is also harmless, so feel free to ignore.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private EJBModuleIdentifier toModuleIdentifier(DeploymentModuleIdentifier identifier) {
+        return new EJBModuleIdentifier(identifier.getApplicationName(), identifier.getModuleName(), identifier.getDistinctName());
+    }
```
pferraro, line 197: Can you file a follow-up jira to eliminate replace all usages of DeploymentModuleIdentifier with EJBModuleIdentifier?

ejb3/src/main/java/org/jboss/as/ejb3/remote/http/EJB3RemoteHTTPService.java
```diff
@@ -5,60 +5,55 @@
...
 
     @Override
     public void start(StartContext context) throws StartException {
-        HttpRemoteEjbService service = new HttpRemoteEjbService(associationServiceInjectedValue.getValue().getAssociation(),
-                null, localTransactionContextInjectedValue.getValue(), classResolverFilter);
-        pathHandlerInjectedValue.getValue().addPrefixPath("/ejb", service.createHttpHandler());
+        HttpRemoteEjbService service = new HttpRemoteEjbService(associationServiceSupplier.get().getAssociation(),
+                null, transactionContextSupplier.get(), classResolverFilter);
+        pathHandlerSupplier.get().addPrefixPath("/ejb", service.createHttpHandler());
+        remoteHTTPServiceConsumer.accept(this);
```
pferraro: This service does not need to provide any value (and an MSC service should never provide itself).

ejb3/src/main/java/org/jboss/as/ejb3/remote/AssociationService.java
```diff
@@ -5,177 +5,66 @@
...
-    @SuppressWarnings("rawtypes")
-    private final List<Map.Entry<Value<ProtocolSocketBinding>, Value<Registry>>> clientMappingsRegistries = new LinkedList<>();
-    private final InjectedValue<ServerEnvironment> serverEnvironmentServiceInjector = new InjectedValue<>();
-
-    private final Object serviceLock = new Object();
-    private final Set<EJBModuleIdentifier> ourModules = new HashSet<>();
-    private volatile ServiceURL cachedServiceURL;
-
-    private final MutableDiscoveryProvider mutableDiscoveryProvider = new MutableDiscoveryProvider();
+    private final Consumer<AssociationService> associationServiceConsumer;
```
pferraro, line 38: Can you file a follow-up jira to fix this encapsulation?
An MSC service should not expose the Service interface via its provided value.
At first glance, it should probably just provide an Association.

ejb3/src/main/java/org/jboss/as/ejb3/remote/AssociationService.java
```diff
@@ -5,177 +5,66 @@
...
-            result.complete();
-            return DiscoveryRequest.NULL;
-        });
+        value = new AssociationImpl(deploymentRepositorySupplier.get(), clientMappingsRegistries);
+        this.associationServiceConsumer.accept(this);
     }
 
     @Override
     public void stop(final StopContext context) {
+        this.associationServiceConsumer.accept(null);
```
pferraro, line 65: Redundant, but harmless.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3Subsystem110Parser.java
```diff
@@ -0,0 +1,131 @@
...
+ * becomes:
+ *
+ * <remote>
+ *   ...
+ * </remote>
+ * <profiles>
+ *     <profile/>
+ * </profiles>
+ *
+ * TODO Parameterize a single parser class by schema version.  Inheritence is a poor model for versioning.
```
pferraro, line 47: Can you file a jira for this if one does not already exist?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
```diff
@@ -528,12 +495,35 @@ public void checkAccessAllowed() throws NamingException {
...
+        // add in the local discovery provider
+        final ServiceBuilder<?> localDiscoveryProviderBuilder = serviceTarget.addService();
+        final Consumer<DiscoveryProvider> discoveryProviderConsumer = localDiscoveryProviderBuilder.provides(LocalEJBDiscoveryProviderService.SERVICE_NAME);
+        final Supplier<DeploymentRepository> deploymentRepositorySupplier = localDiscoveryProviderBuilder.requires(DeploymentRepositoryService.SERVICE_NAME);
+        final Supplier<ServerEnvironment> serverEnvironmentSupplier = localDiscoveryProviderBuilder.requires(ServerEnvironmentService.SERVICE_NAME);
+        // if the "remote" resource is present, pass in references to the client mappings registries installed there
+        List<Supplier<Registry>> clientMappingsRegistrySupplierList = new ArrayList<>();
+        if (ejbSubsystemResource.hasChild(EJB3SubsystemModel.REMOTE_SERVICE_PATH)) {
+            final ModelNode remoteModel = ejbSubsystemResource.getChild(REMOTE_SERVICE_PATH).getModel();
+            final List<ModelNode> connectors = remoteModel.get(CONNECTORS).asList();
```
tadamski: this probably should be:
EJB3RemoteResourceDefinition.CONNECTORS.resolveModelAttribute(context, remoteModel).asList()
so that f.e. expressions are handled correctly

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
```diff
@@ -0,0 +1,195 @@
...
+            result.complete();
+            return DiscoveryRequest.NULL;
+        });
+
+        // provide the discovery provider to services that need it
+        discoveryProviderConsumer.accept(mutableDiscoveryProvider);
+    }
+
+    @Override
+    public void stop(StopContext context) {
```
tadamski, line 187: shouldn't  deploymentRepositoryListener be unregistered?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
tadamski: also, maybe it would more more elegant if underlying EJB3RemoteServiceAdd added those when (if) it is installed? to some extent this breaks the decoupling that this PR is creating

tadamski (dismissed):

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -127,29 +172,52 @@ public ServiceController<?> install(RequirementServiceTarget target) {
...
-        final EJBRemoteConnectorService ejbRemoteConnectorService = new EJBRemoteConnectorService(serviceConsumer,
-                endpointSupplier, executorSupplier, associationServiceSupplier, remotingTransactionServiceSupplier,
+
+        final EJBRemoteConnectorService ejbRemoteConnectorService = new EJBRemoteConnectorService(connectorServiceConsumer, endpointSupplier,
+                executorSupplier, associationServiceSupplier, remotingTransactionServiceSupplier,
                 channelCreationOptions, FilterSpecClassResolverFilter.getFilterForOperationContext(context));
-        builder.setInstance(ejbRemoteConnectorService);
-        builder.install();
+        remoteConnectorServiceBuilder.setInstance(ejbRemoteConnectorService);
+        remoteConnectorServiceBuilder.addAliases(EJBRemoteConnectorService.SERVICE_NAME);
```
pferraro, line 195: Is this alias still needed?

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -127,29 +172,52 @@ public ServiceController<?> install(RequirementServiceTarget target) {
...
-        // remoting channel
         // TODO: Externalize (expose via management API if needed) the version and the marshalling strategy
-        final CapabilityServiceBuilder<?> builder = context.getCapabilityServiceTarget()
+        final CapabilityServiceBuilder<?> remoteConnectorServiceBuilder = context.getCapabilityServiceTarget()
                 .addCapability(EJB3RemoteResourceDefinition.EJB_REMOTE_CAPABILITY);
-        final Consumer<EJBRemoteConnectorService> serviceConsumer = builder
-                .provides(EJB3RemoteResourceDefinition.EJB_REMOTE_CAPABILITY);
-        final Supplier<Endpoint> endpointSupplier = builder
+        final Consumer<EJBRemoteConnectorService> connectorServiceConsumer =
+                remoteConnectorServiceBuilder.provides(EJB3RemoteResourceDefinition.EJB_REMOTE_CAPABILITY.getCapabilityServiceName());
```
pferraro:
```suggestion
        final Consumer<EJBRemoteConnectorService> connectorServiceConsumer = remoteConnectorServiceBuilder.provides(EJB3RemoteResourceDefinition.EJB_REMOTE_CAPABILITY);
```

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -127,29 +172,52 @@ public ServiceController<?> install(RequirementServiceTarget target) {
...
+
+        // Install the Jakarta Enterprise Beans connector service which will listen for client connections
+        // on the Remoting connector (i.e. the client is using the Remoting transport to send EJB invocations)
         final OptionMap channelCreationOptions = this.getChannelCreationOptions(context);
-        // Install the Jakarta Enterprise Beans remoting connector service which will listen for client connections on the
-        // remoting channel
         // TODO: Externalize (expose via management API if needed) the version and the marshalling strategy
-        final CapabilityServiceBuilder<?> builder = context.getCapabilityServiceTarget()
+        final CapabilityServiceBuilder<?> remoteConnectorServiceBuilder = context.getCapabilityServiceTarget()
                 .addCapability(EJB3RemoteResourceDefinition.EJB_REMOTE_CAPABILITY);
```
pferraro: This method is deprecated.
```suggestion
        final CapabilityServiceBuilder<?> remoteConnectorServiceBuilder = serviceTarget.addService();
```

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -98,7 +121,29 @@ void installRuntimeServices(final OperationContext context, final ModelNode mode
...
+        final Consumer<AssociationService> associationServiceConsumer = associationServiceBuilder.provides(AssociationService.SERVICE_NAME);
+        final Supplier<DeploymentRepository> deploymentRepositorySupplier = associationServiceBuilder.requires(DeploymentRepositoryService.SERVICE_NAME);
+
+        // now, for each connector, configure the relevant AssociationService dependencies
+        final List<Map.Entry<Supplier<ProtocolSocketBinding>, Supplier<Registry>>> protocolRegistryPairs = new ArrayList<>();
+        for (ModelNode connector : connectorNameNodes) {
+            String connectorName = connector.asString();
+
+            final Supplier<ProtocolSocketBinding> protocol = associationServiceBuilder.requires(context.getCapabilityServiceName(CONNECTOR_CAPABILITY_NAME, connectorName, ProtocolSocketBinding.class));
+            final Supplier<Registry> registry = associationServiceBuilder.requires(ServiceNameFactory.resolveServiceName(EJB3RemoteResourceDefinition.CLIENT_MAPPINGS_REGISTRY, connectorName));
```
pferraro:
```suggestion
            final Supplier<Registry> registry = associationServiceBuilder.requires(EJB3RemoteResourceDefinition.CLIENT_MAPPINGS_REGISTRY, connectorName);
```

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
```diff
@@ -98,7 +121,29 @@ void installRuntimeServices(final OperationContext context, final ModelNode mode
...
+        final CapabilityServiceBuilder<?> associationServiceBuilder = serviceTarget.addService();
+        final Consumer<AssociationService> associationServiceConsumer = associationServiceBuilder.provides(AssociationService.SERVICE_NAME);
+        final Supplier<DeploymentRepository> deploymentRepositorySupplier = associationServiceBuilder.requires(DeploymentRepositoryService.SERVICE_NAME);
+
+        // now, for each connector, configure the relevant AssociationService dependencies
+        final List<Map.Entry<Supplier<ProtocolSocketBinding>, Supplier<Registry>>> protocolRegistryPairs = new ArrayList<>();
+        for (ModelNode connector : connectorNameNodes) {
+            String connectorName = connector.asString();
+
+            final Supplier<ProtocolSocketBinding> protocol = associationServiceBuilder.requires(context.getCapabilityServiceName(CONNECTOR_CAPABILITY_NAME, connectorName, ProtocolSocketBinding.class));
```
pferraro:
```suggestion
            final Supplier<ProtocolSocketBinding> protocol = associationServiceBuilder.requiresCapability(CONNECTOR_CAPABILITY_NAME, ProtocolSocketBinding.class, connectorName);
```

ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemoteConnectorService.java
```diff
@@ -76,12 +76,12 @@ public void start(StartContext context) throws StartException {
         } catch (ServiceRegistrationException e) {
             throw new StartException(e);
         }
-        serviceConsumer.accept(this);
+        connectorServiceConsumer.accept(this);
```
pferraro: It seem to me that this service does not need to provide a value (and certainly should not provide itself).

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
```diff
@@ -48,12 +50,15 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
...
-        // only add association service dependency if the context is configured to use the local EJB receiver & we are not app client
+        // only add local discovery service dependency if the context is configured to use the local EJB receiver & we are not app client
         final EJBClientDescriptorMetaData ejbClientDescriptorMetaData = deploymentUnit.getAttachment(Attachments.EJB_CLIENT_METADATA);
         final boolean useLocalReceiver = ejbClientDescriptorMetaData == null || ejbClientDescriptorMetaData.isLocalReceiverExcluded() != Boolean.TRUE;
-        final Supplier<AssociationService> associationServiceSupplier = (useLocalReceiver && !appClient) ? builder.requires(AssociationService.SERVICE_NAME) : null;
-        final DiscoveryService discoveryService = new DiscoveryService(discoveryConsumer, remotingProfileServiceSupplier, associationServiceSupplier);
+        // add in local discovery
+        final Supplier<DiscoveryProvider> localDiscoveryProviderSupplier = (useLocalReceiver && !appClient) ? builder.requires(LocalEJBDiscoveryProviderService.SERVICE_NAME) : null;
```
pferraro: Simplifies null checking in service implementation:
```suggestion
        final Supplier<DiscoveryProvider> localDiscoveryProviderSupplier = (useLocalReceiver && !appClient) ? builder.requires(LocalEJBDiscoveryProviderService.SERVICE_NAME) : Functions.constantSupplier(null);
```

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
```diff
@@ -48,12 +50,15 @@ public void deploy(final DeploymentPhaseContext phaseContext) throws DeploymentU
         final ServiceBuilder<?> builder = phaseContext.getServiceTarget().addService(discoveryServiceName);
         final Consumer<Discovery> discoveryConsumer = builder.provides(discoveryServiceName);
         final Supplier<RemotingProfileService> remotingProfileServiceSupplier = profileServiceName != null ? builder.requires(profileServiceName) : null;
```
pferraro: Simplifies null checking in service implementation:
```suggestion
        final Supplier<RemotingProfileService> remotingProfileServiceSupplier = profileServiceName != null ? builder.requires(profileServiceName) : Function.constantSupplier(null);
```

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
```diff
@@ -36,31 +35,31 @@ public final class DiscoveryService implements Service {
...
         this.discoveryConsumer = discoveryConsumer;
         this.remotingProfileServiceSupplier = remotingProfileServiceSupplier;
-        this.associationServiceSupplier = associationServiceSupplier;
+        this.localEJBDiscoveryProviderSupplier = localEJBDiscoveryProviderSupplier;
     }
 
     public void start(final StartContext context) throws StartException {
         final RemotingProfileService remotingProfileService = remotingProfileServiceSupplier != null ? remotingProfileServiceSupplier.get() : null;
-        final AssociationService associationService = associationServiceSupplier != null ? associationServiceSupplier.get() : null;
+        final DiscoveryProvider localEJBDiscoveryProvider = localEJBDiscoveryProviderSupplier != null ? localEJBDiscoveryProviderSupplier.get() : null;
```
pferraro: Assuming the above suggestions were applied:
```suggestion
       final RemotingProfileService remotingProfileService = remotingProfileServiceSupplier.get();
       final DiscoveryProvider localEJBDiscoveryProvider = localEJBDiscoveryProviderSupplier.get();
```

pferraro (changes requested):
Mostly minor comments.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
pferraro: Ideally, the CONNECTORS attribute would be a CapabilityReferenceListAttributeDefinition and thus could resolve directly to a `ServiceDependency<List<ProtocolSocketBinding>>` without the need for boilerplate code.
That said, this code block was merely relocated from somewhere else, and is not part of the scope of this proposed change.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
rachmatowicz: This change has already been made in the PR for WFLY-20091 (distributed deployment repository).

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3SubsystemAdd.java
rachmatowicz: > this probably should be: EJB3RemoteResourceDefinition.CONNECTORS.resolveModelAttribute(context, remoteModel).asList() so that f.e. expressions are handled correctly

Done.

> also, maybe it would more more elegant if underlying EJB3RemoteServiceAdd added those when (if) it is installed? to some extent this breaks the decoupling that this PR is creating

The LocalEJBDiscoveryProviderService only needs the client mappings registries to find out which clusters this node belongs to and it uses them only when remote connectors are enabled (i.e.the remote resource is configured). At the same time, the LocalEJBDiscoveryProviderService needs to be started whether the remote resource is available or not.  This was the simplest way I could think of to arrange for both of those conditions to be satisifed.

> Ideally, the CONNECTORS attribute would be a CapabilityReferenceListAttributeDefinition and thus could resolve directly to a `ServiceDependency<List<ProtocolSocketBinding>>` without the need for boilerplate code. That said, this code block was merely relocated from somewhere else, and is not part of the scope of this proposed change.

I'm trying my best to stop this PR from growing!

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryService.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/DiscoveryRegistrationProcessor.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/local/LocalEJBDiscoveryProviderService.java
rachmatowicz: Ignoring.

ejb3/src/main/java/org/jboss/as/ejb3/remote/AssociationService.java
rachmatowicz: Ignoring.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3Subsystem110Parser.java
rachmatowicz: Filed: https://redhat.atlassian.net/browse/WFLY-22097

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
rachmatowicz: It is used in the RemoteConnectorChannelCreationOptionResource. I'll ckec if it can be removed.

ejb3/src/main/java/org/jboss/as/ejb3/remote/AssociationService.java
rachmatowicz: This case is complicated by the fact that the AssociationService provides methods above and beyond what is contained in the Association interface: namely setExecutor() and sendTopologyUpdateIfLastNodeToLeave(), both called by EJBRemotingConnectorService.

ejb3/src/main/java/org/jboss/as/ejb3/remote/http/EJB3RemoteHTTPService.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/remote/EJBRemoteConnectorService.java
rachmatowicz: Done.

ejb3/src/main/java/org/jboss/as/ejb3/subsystem/EJB3RemoteServiceAdd.java
rachmatowicz: The Consumer has now been removed as not referenced.

pferraro (approved):

rachmatowicz:
FYI, one test case of thousands in this build is failing: org.wildfly.test.integration.microprofile.jwt.propagation.JWTIdentityPropagationTestCase

Recently, it has failed in TeamCity builds: 20277, 20276, 20170, 20271, 20266, 19934, 20273, 20272, 20269 and others, almost all of which are not builds related to this PR. I can't reproduce it locally.

It has been reviewed by at least two reviewers and the rerquested changes were implemented in all cases where the review comments were valid. And it was approved last week.

asoldano:
@jmesnil @kabir @darranl @pferraro @bstansberry @yersan can you please either proceed merging this PR or tell what is preventing you from doing that? thanks

bstansberry:
Thanks @rachmatowicz and @pferraro. @asoldano fyi

asoldano:
thanks @bstansberry :-)

PR #19935, opened 2026-04-30
[WFLY-21826][WFLY-21828] Clean up left over 'preview' settings

bstansberry:
https://redhat.atlassian.net/browse/WFLY-21826
https://redhat.atlassian.net/browse/WFLY-21828
https://redhat.atlassian.net/browse/WFLY-21843

This fixes a few different scenarios where things are associated with a 'preview' concept that shouldn't be any more.

### DIFF
--- a/jpa/hibernate7/pom.xml
+++ b/jpa/hibernate7/pom.xml
@@ -24,16 +24,11 @@
 
     <name>WildFly: Jipijapa Hibernate 7 (JPA 3.2) integration</name>
 
-    <properties>
-        <!-- Use the standard WildFly dependency set -->
-        <dependency.management.pom.artifactId>wildfly-preview-ee-bom</dependency.management.pom.artifactId>
-    </properties>
-
     <dependencyManagement>
         <dependencies>
             <dependency>
                 <groupId>${project.groupId}</groupId>
-                <artifactId>wildfly-preview-test-bom</artifactId>
+                <artifactId>wildfly-standard-test-bom</artifactId>
                 <version>${project.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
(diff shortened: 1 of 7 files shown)


scottmarlow (approved):

jpa/hibernate7/pom.xml
yersan: This module is still requesting its parent to depend on `wildfly-preview-ee-bom`

jpa/persistence-cdi-extension/pom.xml
yersan: This module is still requesting its parent to depend on `wildfly-preview-ee-bom`

yersan (changes requested):

yersan (changes requested):
@bstansberry This one is also suspicious:

https://github.com/wildfly/wildfly/blob/main/pom.xml#L1722

If we disable preview, this module will not be built and the build will fail. I'm not sure where should be its place or even if `-Dno.preview.build` is an expected valid property, but if it does not depend on preview stuff, I would assume it should not be there

jpa/hibernate7/pom.xml
bstansberry: Thanks. Fixed by removing the property so it inherits wildfly-standard-ee-bom from its parent.

jpa/persistence-cdi-extension/pom.xml
bstansberry: Thanks. Fixed by removing the property so it inherits wildfly-standard-ee-bom from its parent.

bstansberry:
> @bstansberry This one is also suspicious:
> 
> https://github.com/wildfly/wildfly/blob/main/pom.xml#L1722
> 
> If we disable preview, this module will not be built and the build will fail. I'm not sure where should be its place or even if `-Dno.preview.build` is an expected valid property, but if it does not depend on preview stuff, I would assume it should not be there

Thanks. I added a commit making a number of related to this profiles that enable/disable modules.

yersan (approved):

PR #19936, opened 2026-04-30
WFLY-21825: fix channel fork resource operations

michpetrov:
Issue: [WFLY-21825](https://redhat.atlassian.net/browse/WFLY-21825)

As far as I can see `channel` has runtime-only protocols, while `fork` doesn't so we can skip the whole addRuntime method, rather than adding extra checks there.

Do we/should we have a test for add+remove? The OperationsTestCase wouldn't catch it since it doesn't execute the runtime methods but "can this resource be removed" feels like a test that should be run on everything.

(I will be on PTO from tomorrow until the 11th, if this needs to get into WF 40 someone take over)

### DIFF
--- a/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/ForkResourceDefinitionRegistrar.java
+++ b/clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/ForkResourceDefinitionRegistrar.java
@@ -158,4 +158,9 @@ public class ForkResourceDefinitionRegistrar extends AbstractChannelResourceDefi
 
         return registration;
     }
+
+    @Override
+    public void addRuntime(OperationContext context, ModelNode model) {
+        // skip
+    }
 }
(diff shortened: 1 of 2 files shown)


clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/subsystem/ForkResourceDefinitionRegistrar.java
```diff
@@ -158,4 +158,9 @@ public ManagementResourceRegistration register(ManagementResourceRegistration pa
 
         return registration;
     }
+
+    @Override
+    public void addRuntime(OperationContext context, ModelNode model) {
+        // skip
+    }
```
pferraro, line 165: We don't want to skip registration of runtime resources.

pferraro (changes requested):

pferraro:
Superseded by https://github.com/wildfly/wildfly/pull/19953

PR #19941, opened 2026-05-01
[WFLY-20255] Test MP Config with security manager

kabir:
https://redhat.atlassian.net/browse/WFLY-20255

### DIFF
--- a/testsuite/integration/microprofile-tck/config/src/test/java/org/wildfly/extension/microprofile/config/test/WildFlyArquillianExtension.java
+++ b/testsuite/integration/microprofile-tck/config/src/test/java/org/wildfly/extension/microprofile/config/test/WildFlyArquillianExtension.java
@@ -4,9 +4,21 @@
  */
 package org.wildfly.extension.microprofile.config.test;
 
+import static org.jboss.as.test.shared.PermissionUtils.createPermissionsXmlAsset;
+
+import java.io.FilePermission;
+import java.lang.reflect.ReflectPermission;
+import java.util.PropertyPermission;
+
 import org.jboss.arquillian.container.spi.client.container.DeploymentExceptionTransformer;
+import org.jboss.arquillian.container.spi.client.deployment.Deployment;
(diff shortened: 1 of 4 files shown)


testsuite/integration/microprofile-tck/config/src/test/java/org/wildfly/extension/microprofile/config/test/WildFlyArquillianExtension.java
```diff
@@ -16,5 +28,40 @@ public class WildFlyArquillianExtension implements LoadableExtension {
...
+
+                    manifestContainer.addAsManifestResource(createPermissionsXmlAsset(
+                            // Permissions required by test instrumentation - arquillian-core.jar and arquillian-testng.jar
+                            new ReflectPermission("suppressAccessChecks"),
+                            new PropertyPermission("*", "read,write"),
+                            new RuntimePermission("accessDeclaredMembers"),
+                            new RuntimePermission("getClassLoader"),
+                            // Permissions required by the MP Config implementation
+                            new FilePermission("<<ALL FILES>>", "read"),
+                            new RuntimePermission("getenv.*")
```
rhusar, line 60: This is a workaround for the impl sec manager handling. I remember experimenting with a fix a while ago - https://github.com/smallrye/smallrye-config/compare/main...rhusar:smallrye-config:security-manager?expand=1

Can you link a tracker for fixing this upstream?

bstansberry: +1. And please reference a WFLY in the comment. We should not be adding test workarounds for incorrect SM behavior by the server without a tracker to fix them.

kabir: @bstansberry @rhusar updated

bstansberry (approved):

bstansberry:
Thanks @kabir @rhusar

PR #19946, opened 2026-05-04
WFLY-21845 - Update to Mojarra 4.1.9

jasondlee:
https://redhat.atlassian.net/browse/WFLY-21845

Update version


bstansberry:
Thanks @jasondlee

PR #19955, opened 2026-05-06
[WFLY-21789] Export expressly from the Hibernate Validator module

bstansberry:
https://redhat.atlassian.net/browse/WFLY-21789

I'm not thrilled with how classloading for Jakarta EL works, but this kind of thing seems necessary. (I beat my head against the wall trying to find other solutions for hours.)

A bunch of other modules, including jakarta.enterprise.api, jakarta.servlet.jsp.api and jakarta.faces.api also depend on and export expressly. And the EE subsystem's JavaEEDependencyProcessor add it as a dep to every deployment. So it's already being made visible to lots of things. This just exposes it to more static modules -- mostly WF extension modules, most of which probably already have it exposed via jakarta.enterprise.api deps

### DIFF
--- a/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/validator/main/module.xml
+++ b/ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/hibernate/validator/main/module.xml
@@ -19,7 +19,9 @@
     <module name="jakarta.persistence.api"/>
     <module name="jakarta.validation.api"/>
     <module name="jakarta.xml.bind.api"/>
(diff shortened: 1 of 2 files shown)


bstansberry:
@jasondlee Can you have a look? My impression is for all the cases where server code relies on ExpressionFactory to load an impl, a relevant module needs to export it's dep on the impl (org.glassfish.expressly). Hibernate Validator 9 removed some code where it would directly instantiate the impl so now it too relies on ExpressionFactory, so it now falls into the same category.

The code that is removed from the testsuite class is a workaround I added a couple weeks ago when I discovered this problem.

jasondlee (approved):

PR #19957, opened 2026-05-06
[WFLY-21850] Fix duplicate htmlunit dependency declaration

yersan:
- From the duplicates, remove the lower version 4.10.0 and use 4.21.0 which is the latest

Jira issue: https://redhat.atlassian.net/browse/WFLY-21850


bstansberry (approved):

bstansberry:
Thanks @yersan

PR #19975, opened 2026-05-08
[WFLY-21867] Upgrade Narayana from 7.3.3.Final to 7.3.4.Final

jmfinelli:
[WFLY-21867](https://redhat.atlassian.net/browse/WFLY-21867)


bstansberry:
Thanks @jmfinelli

PR #19981, opened 2026-05-09
[WFLY-21872] - Adding com.sun.xml.bind.external* to the webservices group

fabiobrz:
Adding com.sun.xml.bind.external* to the webservices grroup in the dependabot.yml configuration, to let it be managed in batch with other related Web Services component updates.

Upstream issue: https://redhat.atlassian.net/browse/WFLY-21872


bstansberry:
Thanks @fabiobrz

PR #19986, opened 2026-05-11
[WFLY-21842] Update Maven plugin guide with feature pack

jmesnil:
* Add information about `org.wildfly:wildfly-ee-10-feature-pack` to the list of Galleon feature-packs
* Moved that list to the Galleon_feature_packs.adoc file
* Include it in the WildFly Maven Plug-in Guide.

JIRA: https://redhat.atlassian.net/browse/WFLY-21842

### DIFF
--- a/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
+++ b/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
@@ -81,7 +81,7 @@ Galleon is configured thanks to the Maven plugin ```<configuration>``` element.
 The first required piece of information that Galleon needs is a reference to at least one WildFly Galleon feature-pack.
 A WildFly Galleon feature-pack is a Maven artifact that contains everything needed to dynamically provision a server.
 
-This feature-pack, as well as the feature-packs on which it depends, are deployed in public maven repositories.
+This feature-pack, as well as any feature-packs on which it depends, are deployed in public Maven repositories.
 
 At plugin execution time the WildFly feature-packs are retrieved and their content is assembled to provision a server.
 
@@ -414,7 +414,7 @@ can be found in the https://docs.wildfly.org/quickstart[WildFly Quickstarts].
 You can check https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/README.md[this documentation] 
 for more details on the WildFly cloud feature-pack .
 
-== WildFly Galleon feature-packs & layers
+== WildFly Galleon feature-packs and layers
 
 include::_galleon/Galleon_feature_packs.adoc[]
 include::_galleon/Galleon_layers.adoc[]
--- a/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
+++ b/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
@@ -414,7 +414,49 @@ can be found in the https://docs.wildfly.org/quickstart[WildFly Quickstarts].
 You can check https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/README.md[this documentation] 
 for more details on the WildFly cloud feature-pack .
 
-== WildFly Galleon feature-packs and layers
+== Feature-packs and layers
 
 include::_galleon/Galleon_feature_packs.adoc[]
+
+The configuration to provision WildFly with EE 10 for a cloud deployment would look like this:
+
+[source,xml]
+----
+<plugin>
+  <groupId>org.wildfly.plugins</groupId>
+  <artifactId>wildfly-maven-plugin</artifactId>
+  <configuration>
+    <feature-packs>
+      <feature-pack>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ee-10-feature-pack</artifactId>
+        <version>${version.wildfly}</version>
+      </feature-pack>
+      <feature-pack>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-galleon-pack</artifactId>
+        <version>${version.wildfly}</version>
+      </feature-pack>
+      <feature-pack>
+        <groupId>org.wildfly.cloud</groupId>
+        <artifactId>wildfly-cloud-galleon-pack</artifactId>
+        <version>${version.wildfly-cloud}</version>
+      </feature-pack>
+    </feature-packs>
+  </configuration>
+  <executions>
+    <execution>
+      <goals>
+        <goal>package</goal>
+      </goals>
+    </execution>
+  </executions>
+</plugin>
+----
+
+In that example, `wildfly-ee-10-feature-pack` is listed before `wildfly-galleon-pack` to ensure that the EE 10 support is
+properly provisioned.
+Likewise, `wildfly-cloud-galleon-pack` is listed last so that its dependencies have been already been provisioned by `wildfly-ee-10-feature-pack`.
+
+
 include::_galleon/Galleon_layers.adoc[]
--- a/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
+++ b/docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
@@ -79,13 +79,16 @@ The WildFly Maven Plugin depends on https://docs.wildfly.org/galleon/[Galleon] t
 Galleon is configured thanks to the Maven plugin ```<configuration>``` element.
 
 The first required piece of information that Galleon needs is a reference to at least one WildFly Galleon feature-pack.
-A WildFly Galleon feature-pack is a maven artifact that contains everything needed to dynamically provision a server.
-This feature-pack, as well as the feature-packs on which it depends,
-are deployed in public maven repositories. 
+A WildFly Galleon feature-pack is a Maven artifact that contains everything needed to dynamically provision a server.
+
+This feature-pack, as well as the feature-packs on which it depends, are deployed in public maven repositories.
 
 At plugin execution time the WildFly feature-packs are retrieved and their content is assembled to provision a server.
 
-Once you have identified a WildFly Galleon feature-pack, you need to select a set of 
+First, you need to identify the set of feature-packs you want to use to provision your server among the available
+link:#WildFly_Galleon_feature-packs[WildFly Galleon feature-packs].
+
+Once you have identified a WildFly Galleon feature-pack, you then need to select a set of
 link:#wildfly_layers[WildFly Layers] that are used to compose the server.
 
 The set of Galleon layers to include is driven by the needs of the application you are developing.
@@ -411,4 +414,7 @@ can be found in the https://docs.wildfly.org/quickstart[WildFly Quickstarts].
 You can check https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/README.md[this documentation] 
 for more details on the WildFly cloud feature-pack .
 
+== WildFly Galleon feature-packs & layers
+
+include::_galleon/Galleon_feature_packs.adoc[]
 include::_galleon/Galleon_layers.adoc[]
--- a/docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
+++ b/docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
@@ -13,15 +13,15 @@ The main WildFly project provides the following feature-packs for use with xref:
 |=======================================================================
 |Maven GroupId:ArtifactId |Short name|Use|Requires
 
-|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.| *wildfly-ee* (by default) or *wildfly-ee-10* (for EE 10 compatibility)
+|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.| *wildfly-ee* (by default) or *wildfly-ee-10* (for EE 10 support)
 
 |*org.wildfly:wildfly-ee-galleon-pack*|*wildfly-ee*|Provisions the most long-term stable functionality of standard WildFly. Does not include MicroProfile or other rapidly evolving technologies that may frequently introduce incompatible changes.| none
 
-|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee-10*|This alternative to *wildfly-ee* provides a Jakarta EE 10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually| none
+|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee-10*|This alternative to *wildfly-ee* provides a Jakarta EE 10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually migrate to EE 11| none
 
-|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.| *wildfly* or *wildfly-preview*
+|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.| *wildfly-ee*, *wildfly-ee-10* or *wildfly-preview*
 
-|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.|*wildfly* or *wildfly-preview*
+|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.|*wildfly-ee*, *wildfly-ee-10* or *wildfly-preview*
 
 |*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces support for WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.|*wildfly* or *wildfly-preview*|
 |=======================================================================
@@ -48,5 +48,5 @@ A specific feature pack must be listed before a feature pack that would use anot
 For example, the *wildfly-ee-10* feature-pack provides a Jakarta EE 10 variant of *wildfly-ee*.
 
 To provision WildFly with that EE 10 variant, this `org.wildfly:wildfly-ee-10-feature-pack` feature pack must be listed *before* the
-`org.wildfly:wildfly-galleon-pack` feature-pack (that would require *wildfly-ee* otherwise) in the provisioning tool (such as the WildFly Maven Plug-in).
+`org.wildfly:wildfly-galleon-pack` feature-pack (that would require *wildfly-ee* otherwise) in the provisioning tool.
 ====
--- a/docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
+++ b/docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
@@ -9,16 +9,21 @@ their content is assembled to create an installation.
 
 The main WildFly project provides the following feature-packs for use with xref:WildFly_and_WildFly_Preview.adoc[standard WildFly]:
 
-[cols="~,~,~",options="header"]
+[cols="~,~,~,~",options="header"]
 |=======================================================================
-|Maven GroupId:ArtifactId |Short name|Use
+|Maven GroupId:ArtifactId |Short name|Use|Requires
 
-|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.
-|*org.wildfly:wildfly-ee-galleon-pack*|*wildfly-ee*|Provisions the most long-term stable functionality of standard WildFly. Does not include MicroProfile or other rapidly evolving technologies that may frequently introduce incompatible changes.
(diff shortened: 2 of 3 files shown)


jmesnil:
@bstansberry the key new piece of information is https://github.com/wildfly/wildfly/pull/19986/changes#diff-1f36a0dc357cf868500c13abfefd8d6ad37b038a418cbb7a426d5da240c11686R18.

The `Use` text does not read great. If you have any suggestion to improve it, I'll take it :) .

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,39 @@
...
+|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces 4.x support for standard WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.
+|=======================================================================
+
+The main WildFly project also provides feature-packs for use with xref:WildFly_and_WildFly_Preview.adoc[WildFly Preview]:
+
+[cols="~,~,~",options="header"]
+|=======================================================================
+|Maven GroupId:ArifactId |Short name|Use
+
+|*org.wildfly:wildfly-preview-feature-pack*|*wildfly-preview*|Provides a technology preview of things that may be coming in standard WildFly. Likely to have significant incompatible changes with every release.
```
jfdenise, line 35: I am wandering if we should speak abou the cloud and datasources preview. cloud and ds are now compatible with both preview, wildfly and ee-10.
FYI, the list of known by Glow feature-packs for 40.0.0.Beta1: https://github.com/wildfly/wildfly-galleon-feature-packs/tree/release/40.0.0.Beta1

jmesnil: that's a good point, we can remove them.

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,37 @@
...
+upgrade their applications to "move back" to `wildfly-ee`.|
+|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.
+|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.
+|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces 4.x support for standard WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces support for standard WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.
```

The MyFaces version will evolve over time and this will eventually be out of date.

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,37 @@
...
+|=======================================================================
+
+The main WildFly project also provide a feature-pack for xref:WildFly_and_WildFly_Preview.adoc[WildFly Preview]:
```
bstansberry:
```suggestion
The main WildFly project also provides a feature-pack for xref:WildFly_and_WildFly_Preview.adoc[WildFly Preview]:
```

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,37 @@
...
+The main WildFly project provides the following feature-packs for use with xref:WildFly_and_WildFly_Preview.adoc[standard WildFly]:
+
+[cols="~,~,~",options="header"]
+|=======================================================================
+|Maven GroupId:ArifactId |Short name|Use
+
+|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.
+|*org.wildfly:wildfly-ee-galleon-pack*|*wildfly-ee*|Provisions the most long-term stable functionality of standard WildFly. Does not include MicroProfile or other rapidly evolving technologies that may frequently introduce incompatible changes.
+|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee10*|This feature pack provides a Jakarta EE10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee10*|This feature pack provides a Jakarta EE10 variant of the most long-term stable functionality of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually
```
My idea here is to make it more clear that this is an alternative to wildfly-ee-galleon-pack. I'll post an alternative suggestion next.

bstansberry:
```suggestion
|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee10*|This alternative to the `wildfly-ee-galleon-pack` provides a Jakarta EE10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually
```
This is another approach to the above.

docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
```diff
@@ -79,13 +79,16 @@ The WildFly Maven Plugin depends on https://docs.wildfly.org/galleon/[Galleon] t
 Galleon is configured thanks to the Maven plugin ```<configuration>``` element.
 
 The first required piece of information that Galleon needs is a reference to at least one WildFly Galleon feature-pack.
-A WildFly Galleon feature-pack is a maven artifact that contains everything needed to dynamically provision a server.
-This feature-pack, as well as the feature-packs on which it depends,
-are deployed in public maven repositories. 
+A WildFly Galleon feature-pack is a Maven artifact that contains everything needed to dynamically provision a server.
+
+This feature-pack, as well as the feature-packs on which it depends, are deployed in public maven repositories.
```
bstansberry:
```suggestion
This feature-pack, as well as any feature-packs on which it depends, are deployed in public maven repositories.
```
This is just to account for the fact that some wildfly-ee / wildfly-ee-10 don't have dependencies.

docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
```diff
@@ -411,4 +414,7 @@ can be found in the https://docs.wildfly.org/quickstart[WildFly Quickstarts].
 You can check https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/README.md[this documentation] 
 for more details on the WildFly cloud feature-pack .
 
+== WildFly Galleon feature-packs & layers
```
bstansberry:
```suggestion
== WildFly Galleon feature-packs and layers
```
To me & looks odd unless all the other text is using initial caps for words.

bstansberry (changes requested):
Thanks @jmesnil! I've made some minor suggestions.

Bigger picture are a couple things we should work through.

1) An FP depends on what by default with what alternatives. The new Galleon_feature_packs.adoc seems like the place for that.

Getting into families etc might be overkill. Maybe a 'Requires' column in the table? Values being:

a) 

blank (e.g. for wildfly-ee)

b) 

short-name

c)

short-name
Alt: another-short-name
Alt: a-third-short-name

The myfaces-feature-pack case is a bit different because the right scenario is b) but which 'short-name' is correct depends on the version of the FP. (This FP provides EE so older releases require wildfly-ee, the soon-to-be-latest 2.1.x will require wildfly-ee-10, and the soon-to-be-latest 3.0 will require wildfly-ee.

2) The WildFly_Maven_Plugin_Guide.adoc should probably have a section illustrating using wildfly-ee-10 with wildfly on top, since the requirement for declaring two feature packs is a different thing.

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,37 @@
...
+These feature-packs are deployed in public maven repositories.
+
+When Galleon is used to install WildFly, the desired WildFly feature-packs are retrieved and
+their content is assembled to create an installation.
+
+The main WildFly project provides the following feature-packs for use with xref:WildFly_and_WildFly_Preview.adoc[standard WildFly]:
+
+[cols="~,~,~",options="header"]
+|=======================================================================
+|Maven GroupId:ArifactId |Short name|Use
```
jbliznak: typo (inherited from before but we may fix it) 

```suggestion
|Maven GroupId:ArtifactId |Short name|Use
```

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,37 @@
...
+|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.
+|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces 4.x support for standard WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.
+|=======================================================================
+
+The main WildFly project also provide a feature-pack for xref:WildFly_and_WildFly_Preview.adoc[WildFly Preview]:
+
+[cols="~,~,~",options="header"]
+|=======================================================================
+|Maven GroupId:ArifactId |Short name|Use
```
jbliznak: typo (inherited from before but we may fix it) 

```suggestion
|Maven GroupId:ArtifactId |Short name|Use
```

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
jmesnil: I prefer that one.

jmesnil:
@bstansberry @jfdenise @jbliznak thanks for the review, I've addressed your comments.

I've added a `Requires` column to the feature pack lists as @bstansberry suggested.
I added an `IMPORTANT` admonition note for EE 10 compatibility that the order of feature packs is significant. I put it in the Feature pack section and not the Maven Plugin section because aiui this is important regardless of the provisioning tool. The order would be significant for Galleon (and provisioning.xml) as well as Prospero.

jmesnil:
https://github.com/wildfly/wildfly.org/pull/894 is a companion guide for the EE 10 feature pack that groups different use case (distribution archives, provisioning & container images)

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+When Galleon is used to install WildFly, the desired WildFly feature-packs are retrieved and
+their content is assembled to create an installation.
+
+The main WildFly project provides the following feature-packs for use with xref:WildFly_and_WildFly_Preview.adoc[standard WildFly]:
+
+[cols="~,~,~,~",options="header"]
+|=======================================================================
+|Maven GroupId:ArtifactId |Short name|Use|Requires
+
+|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.| *wildfly-ee* (by default) or *wildfly-ee-10* (for EE 10 compatibility)
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-galleon-pack*|*wildfly*|The feature-pack most WildFly users use. Provisions a standard WildFly installation. Internally depends on the `wildfly-ee` feature-pack, but also provides integration of technologies like MicroProfile, Micrometer and Open Telemetry that are evolving rapidly, possibly in incompatible ways.| *wildfly-ee* (by default) or *wildfly-ee-10* (for EE 10 support)
```

bstansberry: Compatible has fairly legalistic implications in the context of spec, and although we plan to pass the TCKs I think it's better to stay away from using that term if not required.

Note that this is a bit complex because compatible/incompatible is an important concept in a general sense, e.g. forward/backward compatibility, and that's a different thing. So, this doc uses 'imcompatible' a lot with that sense in mind and that's fine.

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+|*org.wildfly:wildfly-ee-galleon-pack*|*wildfly-ee*|Provisions the most long-term stable functionality of standard WildFly. Does not include MicroProfile or other rapidly evolving technologies that may frequently introduce incompatible changes.| none
+
+|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee-10*|This alternative to *wildfly-ee* provides a Jakarta EE 10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually| none
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee-10*|This alternative to *wildfly-ee* provides a Jakarta EE 10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually migrate to EE 11| none
```

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.| *wildfly* or *wildfly-preview*
+
+|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.|*wildfly* or *wildfly-preview*
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.|*wildfly-ee*, *wildfly-ee-10* or *wildfly-preview*
```
See https://github.com/wildfly/wildfly-datasources-galleon-pack/blob/main/datasources-galleon-pack/wildfly-user-feature-pack-build.xml

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+|*org.wildfly:wildfly-ee-galleon-pack*|*wildfly-ee*|Provisions the most long-term stable functionality of standard WildFly. Does not include MicroProfile or other rapidly evolving technologies that may frequently introduce incompatible changes.| none
+
+|*org.wildfly:wildfly-ee-10-feature-pack*|*wildfly-ee-10*|This alternative to *wildfly-ee* provides a Jakarta EE 10 variant of WildFly for users that want to stay on that version while WildFly moves to a more recent version of the Jakarta EE specifications. This feature pack is released for a limited period of time and users should eventually| none
+
+|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.| *wildfly* or *wildfly-preview*
```
bstansberry:
```suggestion
|*org.wildfly.cloud:wildfly-cloud-galleon-pack*|-|Provisions a set of additional features allowing you to better configure a standard WildFly server to work on the cloud. It brings to the server some configuration aspects that are generally expected in a cloud context. See the feature-pack https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/doc/index.md[documentation] for more details.| *wildfly-ee*, *wildfly-ee-10* or *wildfly-preview*
```
See https://github.com/wildfly/wildfly-cloud-galleon-pack/blob/main/wildfly-cloud-galleon-pack/wildfly-feature-pack-build.xml

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+|*org.wildfly:wildfly-datasources-galleon-pack*|-|Easy integration of popular JDBC drivers and associated datasources into a standard WildFly installation. See the feature-pack project https://github.com/wildfly-extras/wildfly-datasources-galleon-pack/blob/main/README.md[README] for more details.|*wildfly* or *wildfly-preview*
+
+|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces support for WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.|*wildfly* or *wildfly-preview*|
```
bstansberry:
```suggestion
|*org.wildfly:wildfly-myfaces-feature-pack*|-|Provides MyFaces support for WildFly. See the feature-pack project https://github.com/wildfly-extras/wildfly-myfaces-feature-pack/blob/master/README.md[README] for more details.|*wildfly-ee* or *wildfly-preview* (for Jakarta Faces 4.1) or *wildfly-ee-10* (for Faces 4.0)|
```

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+[IMPORTANT]
+.Order of feature-packs.
+====
+When feature packs are used to provision WildFly, *their* order is significant if they have requirements.
+A specific feature pack must be listed before a feature pack that would use another default feature pack.
+
+For example, the *wildfly-ee-10* feature-pack provides a Jakarta EE 10 variant of *wildfly-ee*.
+
+To provision WildFly with that EE 10 variant, this `org.wildfly:wildfly-ee-10-feature-pack` feature pack must be listed *before* the
+`org.wildfly:wildfly-galleon-pack` feature-pack (that would require *wildfly-ee* otherwise) in the provisioning tool (such as the WildFly Maven Plug-in).
```
bstansberry:
```suggestion
`org.wildfly:wildfly-galleon-pack` feature-pack (that would require *wildfly-ee* otherwise) in the provisioning tool.
```
This file provides shared content included in the other tool-specific docs so the reader will have context for the relevant tool.

docs/src/main/asciidoc/_galleon/Galleon_provisioning.adoc
```diff
@@ -21,42 +21,7 @@ Releases of the Galleon command line tool are available on the Galleon link:http
...
-
-|*org.wildfly:wildfly-preview-feature-pack*|*wildfly-preview*|Provides a technology preview of things that may be coming in standard WildFly. Likely to have significant incompatible changes with every release.
-|*org.wildfly.cloud:wildfly-preview-cloud-galleon-pack*|-|WildFly Preview analogue to *wildfly-cloud-galleon-pack*.
-|*org.wildfly:wildfly-datasources-preview-galleon-pack*|-|WildFly Preview analogue to *wildfly-datasources-galleon-pack*.
-|=======================================================================
-
-There are also a number of feature-packs of varying levels of stability available from the https://github.com/orgs/wildfly-extras/repositories[*wildfly-extras* GitHub organization].
-
-In most cases, people working with WildFly are using the `wildfly` feature-pack, so that is the one used in most of the examples in this document.
+include::./Galleon_feature_packs.adoc[]
```
bstansberry, line 24: Since the included file ends with the "Order of feature-packs" admonition, perhaps include a brief example of using wildfly-ee-10, wildfly and something else?

The "and something" else is future-proofing to make it a bit easier to update the section when we drop wildfly-ee-10, since we'd still have two FPs in the example.

docs/src/main/asciidoc/WildFly_Maven_Plugin_Guide.adoc
```diff
@@ -411,4 +414,7 @@ can be found in the https://docs.wildfly.org/quickstart[WildFly Quickstarts].
 You can check https://github.com/wildfly-extras/wildfly-cloud-galleon-pack/blob/main/README.md[this documentation] 
 for more details on the WildFly cloud feature-pack .
 
+== WildFly Galleon feature-packs and layers
+
+include::_galleon/Galleon_feature_packs.adoc[]
```
bstansberry, line 419: Same comments as I had in the other file...

Since the included file ends with the "Order of feature-packs" admonition, perhaps include a brief example of using wildfly-ee-10, wildfly and something else?

The "and something" else is future-proofing to make it a bit easier to update the section when we drop wildfly-ee-10, since we'd still have two FPs in the example.

bstansberry (changes requested):
Thanks, Jeff. I've got more requests. I'll go build this now to see how the pages look.

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
jbliznak: shouldn't that be Jakarta Faces 4.1 for wildfly-ee and wildfly-preview?

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
```diff
@@ -0,0 +1,52 @@
...
+====
```
bstansberry: If you like my suggestion re adding concrete examples of ordering in the two docs that include this just after the include, I recommend adding an adoc comment here noting that that's the intended use and that people shouldn't add more content after this block without considering that.

docs/src/main/asciidoc/_galleon/Galleon_provisioning.adoc
jmesnil: That's a good suggestion. I've made a proper section for the "Order of feature-packs" and added such an example (wildfly-ee-10 + wildfly + cloud) in the Maven plug-in guide

jmesnil:
@bstansberry Thanks for another review. I've addressed almost all your points.
There is just @jbliznak comment about Jakarta Faces to resolve

docs/src/main/asciidoc/_galleon/Galleon_feature_packs.adoc
bstansberry: Thanks @jbliznak. I corrected my suggestion.

bstansberry: @jmesnil ^^^

bstansberry:
Thanks @jmesnil -- LGTM. I responded to Jan's question. If all you do is incorporate that or make other minor tweaks please feel free to merge this. TeamCity is kind of overloaded this week and it adds no value to a PR like this, so it's good to push any update and promptly merge before the pull player kicks off the TC jobs. Or, if they've kicked off cancel them after merging.

bstansberry:
Thanks @jmesnil, @jbliznak and @jfdenise

PR #19988, opened 2026-05-11
[WFLY-21717] Move hashicorp-vault layer under wildfly-ee feature pack

skyllarr:
This PR also adds testing for hashicorp-vault 

https://redhat.atlassian.net/browse/WFLY-21717

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -193,9 +193,6 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
-                                        <name>org.wildfly.extension.hashicorp-vault</name>
-                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
-                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                                 <feature-pack>
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -193,6 +193,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
+                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
+                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                                 <feature-pack>
--- a/ee-dist/pom.xml
+++ b/ee-dist/pom.xml
@@ -172,9 +172,6 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
-                                        <name>org.wildfly.extension.hashicorp-vault</name>
-                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
-                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                             </feature-packs>
--- a/ee-dist/pom.xml
+++ b/ee-dist/pom.xml
@@ -172,6 +172,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
+                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
+                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                             </feature-packs>
--- a/legacy/ee-feature-pack/ee-dist/pom.xml
+++ b/legacy/ee-feature-pack/ee-dist/pom.xml
@@ -173,9 +173,6 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
-                                        <name>org.wildfly.extension.hashicorp-vault</name>
-                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
-                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                             </feature-packs>
--- a/legacy/ee-feature-pack/ee-dist/pom.xml
+++ b/legacy/ee-feature-pack/ee-dist/pom.xml
@@ -173,6 +173,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
+                                        <name>org.wildfly.security.elytron-hashicorp-vault</name>
+                                        <name>io.github.jopenlibs.vault-java-driver</name>
                                     </included-packages>
                                 </feature-pack>
                             </feature-packs>
--- a/testsuite/aggregator-expansion/pom.xml
+++ b/testsuite/aggregator-expansion/pom.xml
@@ -50,6 +50,7 @@
             <modules>
                 <module>../integration/expansion</module>
                 <module>../integration/microprofile-tck</module>
+                <module>../integration/hashicorp-vault</module>
                 <module>../integration/manualmode-expansion</module>
             </modules>
         </profile>
@@ -61,6 +62,7 @@
                 <module>../integration/manualmode-expansion</module>
                 <module>../integration/expansion</module>
                 <module>../integration/microprofile-tck</module>
+                <module>../integration/hashicorp-vault</module>
             </modules>
         </profile>
 
@@ -77,6 +79,7 @@
             <modules>
                 <module>../integration/expansion</module>
                 <module>../integration/microprofile-tck</module>
+                <module>../integration/hashicorp-vault</module>
                 <module>../layers-expansion</module>
             </modules>
         </profile>
@@ -88,6 +91,7 @@
             <modules>
                 <module>../integration/expansion</module>
                 <module>../integration/microprofile-tck</module>
+                <module>../integration/hashicorp-vault</module>
                 <module>../integration/manualmode-expansion</module>
             </modules>
         </profile>
@@ -124,6 +128,15 @@
             </modules>
         </profile>
 
+        <!-- -Dts.hashicorp-vault -->
+        <profile>
+            <id>ts.integ.group.hashicorp-vault</id>
+            <activation><property><name>ts.hashicorp-vault</name></property></activation>
+            <modules>
+                <module>../integration/hashicorp-vault</module>
+            </modules>
+        </profile>
+
         <!-- Test against bootable jar -->
         <profile>
             <id>expansion.bootablejar.profile</id>
@@ -135,6 +148,7 @@
             <modules>
                 <module>../integration/expansion</module>
                 <module>../integration/microprofile-tck</module>
+                <module>../integration/hashicorp-vault</module>
             </modules>
         </profile>
 
--- a/testsuite/integration/hashicorp-vault/pom.xml
+++ b/testsuite/integration/hashicorp-vault/pom.xml
@@ -1,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright The WildFly Authors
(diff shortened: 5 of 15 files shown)


skyllarr:
Hi @jfdenise , I am trying to move the hashicorp-vault layer to the wildfly-ee feature pack. Do these changes look reasonable to you?

skyllarr:
@honza-kasik This is the PR, the zip dependency is required for the [layer definition](https://github.com/wildfly-security-incubator/wildfly-vault-feature-pack/blob/main/galleon-shared/src/main/resources/layers/standalone/hashicorp-vault/layer-spec.xml), I would have to change how I publish the modules to maven central to remove the full zip dependency, but I decided to leave it as is

skyllarr:
This PR will not build as it uses Final-SNAPSHOT version of feature pack. Later I'll change to Final version when published

jfdenise:
Hi @skyllarr , unzipping a built feature-pack is not the pattern that has been followed by WildFly until now (and I would not recommend it). If the goal is to keep the src of the feature-pack in the hashicorp repo, then a different approach should be taken:
A dedicated artifact in the hashicorp src repo should contain what you want to share with wildfly-ee feature-pack. I can see [https://github.com/wildfly-security-incubator/wildfly-vault-feature-pack/tree/main/galleon-shared](https://github.com/wildfly-security-incubator/wildfly-vault-feature-pack/tree/main/galleon-shared) that should be it.

I think, @bstansberry feel free to correct me, but the MVC Krazo integration could be a source of inspiration for this integration. 
The mvc krazo feature-pack is there: https://github.com/wildfly/mvc-krazo-feature-pack, in wildfly, the unzip of its shared content: https://github.com/wildfly/wildfly/blob/main/galleon-pack/galleon-feature-pack/pom.xml#L55

bstansberry:
Yes, the mvc-krazo integration is the basic pattern to follow for this kind of thing. It's the pattern we used for bringing in Galleon content from WF Core as well. The wildfly-preview FP also uses it for vertx stuff.

bstansberry:
I requested a review from myself because when the time comes I want to approve this before it goes in, to make sure various subtle things are handled.

skyllarr:
@jfdenise Thanks for review! I have made the changes based on mvc-krazo (I've also added testsuite based on microprofile-TCK) and used galleon-shared module. Do these changess look okay to you now? 

Ignore the SNAPSHOT versions, these will be updated when this feature will be promoted to default stability, so after WF40 Final

ee-feature-pack/galleon-shared/src/main/resources/packages/io.github.jopenlibs.vault-java-driver/package.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
```
jfdenise: @skyllarr , such packages should be automatically generated by Galleon plugins when building the ee-feature-pack. They are generated from the JBOSS Modules modules (that should have been copied over from the galleon-shared).

skyllarr: @jfdenise Thank you, I've removed those modules.

skyllarr:
@jfdenise Does the PR look okay now? The resulting WF should contain the hashicorp-vault modules, but default configuration doesn't include it

skyllarr:
Hi @jfdenise , does this PR looks good to you now? Thanks!

skyllarr:
@bstansberry Please review if possible, thanks!

jfdenise:
@skyllarr , the integration should be done in the wildfly-ee feature-packs (there are 2 now, wiildfly-ee and wildfly-ee-10), this PR adds it to the wildfly feature-pack (that depends on wildfly-ee feature-pack). 
krazo subsystem is added to the wildfly feature-pack, not the wildfly-ee one, I suspect that is the reason.

With the ee-10 vs ee-11 you will need to add the hashicorp content in 2 places:

* https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-feature-pack/pom.xml
* https://github.com/wildfly/wildfly/blob/main/legacy/ee-feature-pack/feature-pack/pom.xml

You also need to add the extensions in two places:

* https://github.com/wildfly/wildfly/blob/main/ee-feature-pack/galleon-feature-pack/wildfly-feature-pack-build.xml
* https://github.com/wildfly/wildfly/blob/main/legacy/ee-feature-pack/feature-pack/wildfly-feature-pack-build.xml

testsuite/integration/hashicorp-vault/pom.xml
```diff
@@ -0,0 +1,207 @@
...
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly</groupId>
+        <artifactId>wildfly-ts-integ</artifactId>
+        <version>40.0.0.Final-SNAPSHOT</version>
```
jbliznak:
```suggestion
        <version>41.0.0.Beta1-SNAPSHOT</version>
```

skyllarr:
@jfdenise Thank you! I moved it to the wildfly-ee and added extensions to the places you mentioned

testsuite/integration/hashicorp-vault/pom.xml
skyllarr: Resolved, thanks!

skyllarr:
Thanks for all reviews up to this point @jfdenise ! I've made the changes. 

@jfdenise  @bstansberry Given the deadline of 17th June is approaching, please let me know if this PR is good when you'll get some time. Thank you!

dist/pom.xml
```diff
@@ -193,6 +193,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
```
jfdenise: You shouldn't need those packages. They are automatically included as default packages in the generated modules.all package.

ee-dist/pom.xml
```diff
@@ -172,6 +172,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
```
jfdenise: You shouldn't need those packages. They are automatically included as default packages in the generated modules.all package.

legacy/ee-feature-pack/ee-dist/pom.xml
```diff
@@ -173,6 +173,9 @@
                                     <version>${ee.maven.version}</version>
                                     <included-packages>
                                         <name>docs.examples.configs</name>
+                                        <name>org.wildfly.extension.hashicorp-vault</name>
```
jfdenise: You shouldn't need those packages. They are automatically included as default packages in the generated modules.all package.

testsuite/integration/hashicorp-vault/pom.xml
```diff
@@ -0,0 +1,207 @@
...
+
+    <artifactId>wildfly-ts-integ-hashicorp-vault</artifactId>
+    <name>WildFly Test Suite: Integration - HashiCorp Vault</name>
+
+    <properties>
+        <jbossas.ts.integ.dir>${basedir}/..</jbossas.ts.integ.dir>
+        <jbossas.ts.dir>${jbossas.ts.integ.dir}/..</jbossas.ts.dir>
+        <jbossas.project.dir>${jbossas.ts.dir}/..</jbossas.project.dir>
+        <jboss.dist>${project.build.directory}/wildfly-layers</jboss.dist>
+        <server.jboss.args>${jboss.args} --stability=community</server.jboss.args>
```
jfdenise: Do you really need this stability level, that is the default one.

testsuite/integration/hashicorp-vault/pom.xml
```diff
@@ -0,0 +1,207 @@
...
+                <version>${version.org.wildfly.plugin}</version>
+                <configuration>
+                    <overwrite-provisioned-server>true</overwrite-provisioned-server>
+                    <log-provisioning-time>${galleon.log.time}</log-provisioning-time>
+                    <offline-provisioning>${galleon.offline}</offline-provisioning>
+                    <record-provisioning-state>false</record-provisioning-state>
+                    <galleon-options>
+                        <jboss-fork-embedded>${galleon.fork.embedded}</jboss-fork-embedded>
+                        <optional-packages>passive+</optional-packages>
+                        <config-stability-level>community</config-stability-level>
```
jfdenise: Do you really need to set the stability level, community is the default. Also applies to next option.

jfdenise (changes requested):

jfdenise:
@skyllarr , I looked at the failing tests, that highlight some issues:
* the subsystem has not been added to the preview feature-pack. I think (but @bstansberry could correct me) that you also need to add it to preview feature-pack. The preview feature-pack doesn't inherit from wildfly-ee, so you need to do it explictely. You can check for krazo integration in preview. That should explain the preview tests failing: https://ci.wildfly.org/buildConfiguration/WF_PullRequest_WildFlyPreviewLinuxJdk25/566643?logFilter=debug&logView=flowAware
* For the layers test failing, the simpler is that I create a PR against your branch. Working on it.

testsuite/integration/hashicorp-vault/pom.xml
skyllarr: Ah actually it should be default stability now

testsuite/integration/hashicorp-vault/pom.xml
skyllarr: Ah actually it should be default stability now

jfdenise:
@skyllarr , 
* Galleon tests are now passing. 
* The WildFly preview testing is failing because hashicorp-vault is not in the WildFly preview feature-pack.
* The ProvisioningConsistencyTestCase failure, I suspect, is due to the org.wildfly.extension.hashicorp-vault (and 2 others) packages being explicitely included, I commented on it.
* Other failing tests, are failing due to some arquillian misconfiguration but locally I can't reproduce the problem. Perhaps @jamezp can see something.

testsuite/integration/hashicorp-vault/pom.xml
skyllarr: fixed

testsuite/integration/hashicorp-vault/pom.xml
skyllarr: fixed

legacy/ee-feature-pack/ee-dist/pom.xml
skyllarr: fixed

ee-dist/pom.xml
skyllarr: fixed

dist/pom.xml
skyllarr: fixed

skyllarr:
@jfdenise I've addressed your comments and upgraded wildfly-vault-feature-pack version to be `.Final` and it seems all checks are now passing

jamezp:
I'll look at the tests once they finish, but I'm not sure they are due to Arquillian. It looks like the test is trying to use CLI through an test API to connect to the controller and it can't for some reason.

testsuite/layers-expansion/pom.xml
bstansberry, line 1: This file should be unchanged. This is for functionality provided by the wildfly feature-pack and this PR is integrating this content in wildfly-ee/wildfly-ee-10, so this should be in testsuite/layers.

Also, there should be an execution that provisions a server with just the hashicorp-vault layer.

testsuite/aggregator-expansion/pom.xml
bstansberry, line 1: This file should be unchanged. This is for functionality provided by the wildfly feature-pack and this PR is integrating this content in wildfly-ee/wildfly-ee-10, so this should be in testsuite/aggregator-base.

bstansberry (changes requested):

bstansberry:
Superceded by #20164

PR #19995, opened 2026-05-12
[WFLY-10656]: Create a test based on messaging-clustering-singleton quickstart.

ehsavoie:
Introduces ClusteredSingletonMDBTestCase, verifying that a singleton MDB is active on exactly one cluster node at a time and that delivery fails over to another node when the active node stops.

Jira: https://redhat.atlassian.net/browse/WFLY-10656

### DIFF
--- a/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ClusteredSingletonMDBTestCase.java
+++ b/testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ClusteredSingletonMDBTestCase.java
@@ -1,0 +1,317 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.test.manualmode.messaging;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
+
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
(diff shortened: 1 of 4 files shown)


testsuite/integration/manualmode/src/test/java/org/jboss/as/test/manualmode/messaging/ClusteredSingletonMDBTestCase.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    private boolean[] waitForSingletonElection(ModelControllerClient client1, ModelControllerClient client2)
+            throws Exception {
+        long start = System.currentTimeMillis();
+        do {
+            try {
+                boolean s1 = isMDBDeliveryActive(client1);
+                boolean s2 = isMDBDeliveryActive(client2);
+                if (s1 ^ s2) {
```
pferraro:
```suggestion
                if (s1 || s2) {
```

Otherwise, the test will tolerate the condition where s1 && s2, rather than triggering the assertion on line 109.

However, if the primary provider of the singleton were to change from s1 -> s2 between invocations to the calls to isMDBDeliveryActive(...), it is possible for both calls to return true, even though there was never a time where the delivery group of each server was active.

To reduce the likelihood of a false positive, we can either:
* Test for a stale s1 value by retesting isMDBDeliveryActive(client) in the event that both calls returned true.
e.g.
```java
                if ((s1 ^ s2) || (s1 && s2 && isMDBDeliveryActive(client1))) {
```
* Invoke isMDBDeliveryActive(...) per server concurrently, rather than sequentially.

Neither will completely eliminate the possibility of a false positive, but should reduce the likelihood considerably.

pferraro (changes requested):

pferraro (approved):

rhusar (approved):

rhusar:
Thanks @ehsavoie and @pferraro

PR #19997, opened 2026-05-13
[WFLY-21885] Make it easy to swap production code compilation between EE 10 and 11

bstansberry:
https://redhat.atlassian.net/browse/WFLY-21885

### DIFF
--- a/batch-jberet/pom.xml
+++ b/batch-jberet/pom.xml
@@ -62,10 +62,6 @@
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>jakarta.enterprise</groupId>
-            <artifactId>jakarta.enterprise.cdi-el-api</artifactId>
-        </dependency>
-        <dependency>
             <groupId>jakarta.transaction</groupId>
             <artifactId>jakarta.transaction-api</artifactId>
         </dependency>
@@ -184,4 +180,23 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>latest.ee.specific.dependencies</id>
+            <activation>
+                <property>
+                    <!-- This profile exists to make it is easy to stop doing what WildFly does downstream.
+                         It's not meant to be turned off/on by external callers. -->
+                    <name>!WRONGLY.disable.latest.ee.specific.dependencies</name>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>jakarta.enterprise</groupId>
+                    <artifactId>jakarta.enterprise.cdi-el-api</artifactId>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
 </project>
--- a/batch-jberet/pom.xml
+++ b/batch-jberet/pom.xml
@@ -29,7 +29,7 @@
         <dependencies>
             <dependency>
                 <groupId>${project.groupId}</groupId>
-                <artifactId>wildfly-standard-ee-bom</artifactId>
+                <artifactId>${dependency.management.import.artifact}</artifactId>
                 <version>${ee.maven.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
--- a/jpa/hibernatesearch/pom.xml
+++ b/jpa/hibernatesearch/pom.xml
@@ -24,11 +24,6 @@
 
     <name>WildFly: Jipijapa Hibernate Search integration</name>
 
-    <properties>
-        <!-- Use the standard WildFly dependency set -->
-        <dependency.management.pom.artifactId>wildfly-standard-ee-bom</dependency.management.pom.artifactId>
-    </properties>
-
     <dependencies>
         <!-- Internal -->
 
--- a/testsuite/integration/basic/pom.xml
+++ b/testsuite/integration/basic/pom.xml
@@ -797,6 +797,11 @@
             <artifactId>ironjacamar-deployers-common</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jboss.spec.jakarta.el</groupId>
+            <artifactId>${jakarta.el.api.artifactId}</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.spec.jakarta.xml.soap</groupId>
             <artifactId>jboss-saaj-api_3.0_spec</artifactId>
         </dependency>
@@ -3761,8 +3766,12 @@
         </profile>
         <!-- Dependencies only to be used when not testing installations using the legacy EE FP -->
         <profile>
-            <id>standard-ee-tests-deps</id>
+            <id>latest.ee.specific.dependencies</id>
             <activation>
+                <!-- Note: an analogous profile with the same id in some production code modules
+                     uses activeByDefault because its use is not meant to be externally controlled.
+                     But for testsuite use cases the intent is to allow this profile to be disabled
+                     when what it does is inappropriate for the test job. -->
                 <property>
                     <name>!legacy-ee-tests</name>
                 </property>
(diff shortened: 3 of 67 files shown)


weld/pom.xml
```diff
@@ -39,12 +39,27 @@
...
             </dependency>
         </dependencies>
     </dependencyManagement>
 
+    <profiles>
+        <profile>
+            <id>latest.ee.specific.dependencies</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
```
yersan, line 59: Could not the activation of these profiles be controller from the main pom root? I guess what we would be looking for is to just modify it in a single place. My understanding is that it should be deactivated by default for the product

testsuite/integration/basic/pom.xml
```diff
@@ -3773,27 +3778,6 @@
                     <artifactId>jakarta.enterprise.cdi-el-api</artifactId>
                     <scope>test</scope>
                 </dependency>
```
yersan, line 3784: Should this profile be replaced by the `<id>latest.ee.specific.dependencies</id>` one or should be completely removed from this pom?

weld/pom.xml
bstansberry: @yersan Doing it in the main pom would result in this dependency being added to every module, which I'm not willing to do.

I tried this but unfortunately it does not work:

```<activeByDefault>${some.property.set.in.the.root.pom}</activeByDefault>```

IDEA flags that as a code error and Maven just treats that as if it were 'false'. If you do a help:effective-pom and look at the profile, it actually just shows ```<activation/>``` -- i.e. it treats it as false and then since that's the default behavior of profile activation when it outputs the effective pom it just simplifies it down to the minimum.

I'm pretty sure this is because Maven needs to work out what profiles are enabled very early in its processing, before it gets to the point where it can evaluate pom properties. Looking at it, all the profile activation mechanisms involve data that is known by the VM before it calls the application's main() method. This is probably why they don't support this kind of super-useful thing:

```
<activation>
    <maven-property>
        <name>some.property.set.in.the.root.pom</name>
    </maven-property>
</activation>
```

I can see too why they don't, since a profile itself can change the value of a property, so you can end up in chicken-or-egg scenarios.

File: testsuite/integration/basic/pom.xml (hunk 2 above)
Comment (bstansberry) [reply testsuite/integration/basic/pom.xml:3784]: It can't be removed -- it's what allows WF test jobs that test EE 10 to disable the presence of this dep. For the testsuite, this is externally controllable; it's what allows us to test using EE 10 deps.

I could change the id, although I don't know what difference it makes.  I don't think we'd ever intend for people/jobs to be doing things like -Platest.ee.specific.dependencies or -P-latest.ee.specific.dependencies. The latest.ee.specific.dependencies profiles in the other modules would be statically activatedByDefault or not, and never externally changed. WF by policy uses EE 11 deps when compiling those modules; EAP by policy uses EE 10. It's not something that varies based on context, e.g. across different test jobs.

Note: here's how I envision this working downstream:

1) This profile is removed.
2) We add two new profiles that do the same thing as this one. One is activated by the presence of the -Dlatest-ee-full-server-tests property, the other by the presence of -Dlatest-ee-only-server-tests.

This, plus my addition of the jakarta.el.api.artifactId property, the default value of which downstream would change, eliminates the need for users using the "extra" -Dlatest-ee-tests property.

Unfortunately, upstream the need for -Dlegacy-ee-tests cannot be eliminated. A profile can't be activated based on "!legacy-ee-full-server-tests && !legacy-ee-only-server-tests"

Note: writing all this made me realize I forgot something re the jakarta.el.api.artifactId property. The various xxx-server-tests profiles in testsuite/pom.xml need to change the value of that property so it matches what the test job is using.

Note: since I have to touch this PR to fix ^^^ anyway, I'll change the profile id here just to eliminate using a different name.

Comment (jbliznak) [reply testsuite/integration/basic/pom.xml:3784]: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#Active_by_default

> Active by default
> Boolean flag which determines if the profile is active by default. Is false by default. **This flag is only evaluated if no other profile is explicitly activated** via command line, settings.xml or activated through some other activator. Otherwise, it has no effect.

I think we cannot rely on this. If you want some profile to be enabled always unless explicitly disabled, you must control it via property

Comment (jbliznak) [reply testsuite/integration/basic/pom.xml:3784]: this will make the profile active unless `-Ddisable.this.profile` is added to the command, regardless of other profiles being activated or deactivated
```xml
            <activation>
                <property>
                    <name>!disable.this.profile</name>
                </property>
            </activation>
```

yersan: > I could change the id, although I don't know what difference it makes.

Sorry, what I meant was to replace the profile with this one:

```
<profile>
    <id>latest.ee.specific.dependencies</id>
    <activation>
        <activeByDefault>true</activeByDefault>
    </activation>
    <dependencies>
        <dependency>
            <groupId>jakarta.enterprise</groupId>
            <artifactId>jakarta.enterprise.cdi-el-api</artifactId>
        </dependency>
    </dependencies>
</profile>
```

I got the impression by looking at the changes that the switch on downstream was going to switch those profiles off (disabling them by default) and switch the value of the new property for the boms that takes in place when we are building downstream (`dependency.management.import.artifact`) plus that defines the jpa api (`jakarta.el.api.artifactId`)

Anyway, it is a work in progress, do not have the final view so ... just suggestions from what I was understanding

jpa/hibernatesearch/pom.xml
```diff
@@ -24,11 +24,6 @@
 
     <name>WildFly: Jipijapa Hibernate Search integration</name>
 
-    <properties>
-        <!-- Use the standard WildFly dependency set -->
```
scottmarlow, line 28: I'm not really sure if this part for Hibernate Search is right.  Will try building locally and see what happens.

scottmarlow: The build does contain the expected Hibernate Search jar in the right location so I think this is right.

testsuite/integration/basic/pom.xml
bstansberry: @jbliznak  -- thanks -- I'll make it work like that. Downstream can change it to ```<name>do.not.enable.this.profile</name>``` or similar.

bstansberry: > I got the impression by looking at the changes that the switch on downstream was going to switch those profiles off (disabling them by default) and switch the value of the new property for the boms that takes in place when we are building downstream (`dependency.management.import.artifact`) plus that defines the jpa api (`jakarta.el.api.artifactId`)
> 
> Anyway, it is a work in progress, do not have the final view so ... just suggestions from what I was understanding

You are correct about all of that, but adding this dep in this module is different. Whether it happens or not needs to be externally controllable.

Note that the profiles here where I used ```<activateByDefault>true``` are only set up as profiles to make it easy to exclude adding the dep downstream by tweaking the activation so it doesn't happen. Those profiles should **never** be deliberately activated or deactivated. An alternative would be to not use a profile at all and just delete the dependency elements downstream. That's more fragile though. 

This testsuite situation is different. It's not about how production code is compiled.

jpa/hibernatesearch/pom.xml
bstansberry: @scottmarlow This module needs to compile against the EE 11 dependencies upstream and the EE 10 ones downstream. The settings from jpa/pom.xml that it inherits result in that.

yersan (approved):

jasondlee (approved):
lgtm

batch-jberet/pom.xml
```diff
@@ -184,4 +180,23 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>latest.ee.specific.dependencies</id>
+            <activation>
+                <property>
+                    <!-- This profile exists to make it is easy to stop doing what WildFly does downstream.
```
rhusar, line 189:
```suggestion
                    <!-- This profile exists to make it easy to stop doing what WildFly does downstream.
```

rhusar (approved):
Brief review LGTM.

batch-jberet/pom.xml
rhusar: 5x

scottmarlow (approved):

bstansberry:
https://ci.wildfly.org/viewLog.html?buildId=562227& is a custom run of the Windows job that's limited to -Dts.clustering and that switches to using SE 21. See [#wildfly-developers > Clustering failures on Windows](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Clustering.20failures.20on.20Windows/with/595104369) for context -- *maybe* whatever is causing the Windows jobs to fail everywhere does not affect SE 21 jobs.

emmartins (approved):
@bstansberry this is fine if those modules don't have different (other than version) dependencies, now or in the future. 

This could also be used in the following tools and client builders at boms/user. The server ones we can't cause bom builder plugin config changes too.

bstansberry:
Thanks @yersan and others :)

PR #20001, opened 2026-05-13
Fix for WFLY-21882, Allow use of a settings file in testsuite/galleon/update

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFLY-21882
In the WildFly CI configuration we would have to set: `-Dwildfly.galleon.test.settings.path=~/.m2/settings.xml`


bstansberry:
I'm going to cancel CI on this now. We had some CI disruption and have a massive job queue as a result and i want to prioritize time sensitive stuff. TC will re-queue.

bstansberry:
I kicked off https://ci.wildfly.org/viewQueued.html?itemId=561940 which just does -Dts.galleon. Right now the job configs are not setting this flag so all CI can do on this PR is check that it doesn't break the testsuite/galleon module. One job is enough for that.

bstansberry:
Thanks @jfdenise

PR #20004, opened 2026-05-13
Fix for WFLY-21893, Upgrade Galleon to 7.0.7.Final and Galleon-plugin…

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFLY-21893


bstansberry:
Thanks @jfdenise

PR #20007, opened 2026-05-13
WFLY-21894 Distributed session manager reuses session ID for recreated sessions.

pferraro:
https://redhat.atlassian.net/browse/WFLY-21894

N.B. Integration test changes proposed here creates evidence for https://redhat.atlassian.net/browse/WFLY-21895 (SimpleWebTestCase will block server shutdown until default timeout, i.e. 1 minute).

PR #20008, opened 2026-05-13
WFLY-21895 Add workaround for UNDERTOW-2764 which otherwise causes recreated sessions to never complete.

pferraro:
https://redhat.atlassian.net/browse/WFLY-21895

Includes https://redhat.atlassian.net/browse/WFLY-21894
Distributed session manager reuses session ID for recreated sessions

Updates SimpleWebTestCase to validate session invalidation followed by recreation in same request.

### DIFF
--- a/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
+++ b/clustering/web/undertow/src/main/java/org/wildfly/clustering/web/undertow/session/DistributableSessionManager.java
@@ -18,6 +18,8 @@ import io.undertow.server.session.SessionListeners;
 import io.undertow.util.AttachmentKey;
 
 import org.jboss.logging.Logger;
+import org.wildfly.clustering.function.BiConsumer;
+import org.wildfly.clustering.function.BiFunction;
 import org.wildfly.clustering.function.Consumer;
 import org.wildfly.clustering.function.Function;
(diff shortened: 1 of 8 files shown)


rhusar (dismissed):
LGTM

rhusar:
Adding the 40.x label in case we want to still include it last minute.

pferraro:
Actually, better to combine this with https://github.com/wildfly/wildfly/pull/20007 as this otherwise causes other test issues when run in isolation.

rhusar (dismissed):
LGTM including the new changes.

pferraro:
The session expiration tests also needed fixing as they were expecting events for a recreated session to be indexed via the old ID.

rhusar:
@pferraro Hm, many clustering tests are failing now, SimpleWebTestCase, DistributableSessionManagerTestCase, etc.

bstansberry:
I was going to start a fresh run of the Windows job now that the env problem affecting it should be fixed. But I suspect the current complete set of jobs that started about an hour ago may have started after the fix was put in place. We'll see. If not I'll start another run of it.

bstansberry:
@rhusar is this ok now?

rhusar:
> @rhusar is this ok now?
> 
> 

Re-re-reviewing now.

rhusar (approved):
LGTM

PR #20016, opened 2026-05-15
WFLY-21029 Move -proc:full compiler arg to pluginManagement

javideros:
Java 23 changed the default annotation processing behaviour: processors no longer run unless explicitly requested via -proc:full (or -proc:only). The existing fix used a profile activated on [23,), but profiles activated by JDK version are not applied during reactor resolution, so child modules that re-declare maven-compiler-plugin configuration could silently lose the argument.

Moving the arg to the root guarantees that every module inherits it unconditionally and that combine.children="append" semantics compose correctly with any local overrides.

Note: -proc:full was backported to JDK 17.0.11 and JDK 21.0.3; builds on older patch releases of those JDKs will need to upgrade.

Tested locally with Java 17.0.19, 21, and 25 — all 112 smoke tests pass.

Fixes: https://issues.redhat.com/browse/WFLY-21029

____

> [!WARNING]

> * [WFLY-21029](https://redhat.atlassian.net/browse/WFLY-21029)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1550,6 +1550,16 @@
                         <ignorePropertiesPrefixedWith>legacy.</ignorePropertiesPrefixedWith>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-compiler-plugin</artifactId>
+                    <configuration>
+                        <compilerArgs combine.children="append">
+                            <!-- Note: this requires Java SE 17.0.11 or later. Update your environment if this causes a problem. -->
+                            <arg>-proc:full</arg>
+                        </compilerArgs>
+                    </configuration>
+                </plugin>
             </plugins>
         </pluginManagement>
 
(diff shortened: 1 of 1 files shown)


bstansberry:
Hi @javideros. Please squash this to a single commit. Thanks!

pom.xml
```diff
@@ -1561,6 +1550,16 @@
                         <ignorePropertiesPrefixedWith>legacy.</ignorePropertiesPrefixedWith>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-compiler-plugin</artifactId>
+                    <configuration>
+                        <compilerArgs combine.children="append">
+                              <!-- Note: this requires Java SE 17+11 or later. Update your environment if this causes a problem. -->
```
rhusar: "Java SE 17+11" is not standard version notation. This should read "17.0.11".

rhusar: Indentation is off here as well.

rhusar (approved):
The core fix looks good, but should fix nits.

rhusar:
@ehsavoie for review as the original reporter.

pom.xml
javideros: Fixed both — corrected the version notation to 17.0.11 and aligned the indentation. Thank you.

ehsavoie (approved):

ehsavoie:
Thanks @javideros and @rhusar

PR #20019, opened 2026-05-16
[WFLY-21903] Upgrade Shibboleth java-support to 8.4.2

bstansberry:
https://redhat.atlassian.net/browse/WFLY-21903

This updates from 8.0.0 to 8.4.2. The only dep we have that uses this, opensaml, itself uses 8.4.2 so we are quite behind with 8.0.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -493,7 +493,7 @@
         <version.jboss.jaxbintros>2.0.1</version.jboss.jaxbintros>
         <version.joda-time>2.12.7</version.joda-time>
         <version.net.jcip>1.0</version.net.jcip>
-        <version.net.shibboleth.utilities.java-support>8.0.0</version.net.shibboleth.utilities.java-support>
+        <version.net.shibboleth.utilities.java-support>8.4.2</version.net.shibboleth.utilities.java-support>
         <version.org.apache.activemq.artemis>2.53.0</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>2.0.0</version.org.apache.activemq.artemis.native>
         <version.org.apache.avro>1.12.1</version.org.apache.avro>


rhusar (approved):

rhusar:
Sigh, this version isn't in central

https://repo1.maven.org/maven2/net/shibboleth/utilities/java-support/8.0.0/java-support-8.0.0.pom
https://repo1.maven.org/maven2/net/shibboleth/utilities/java-support/8.4.2/java-support-8.4.2.pom

rhusar:
The new version has hard depenency on slf4j-api, but given no module.xml is modified

Checking ee-feature-pack/galleon-shared/src/main/resources/modules/system/layers/base/org/opensaml/main/module.xml it's a on a flat CP which already defined slf4j, so all good.

rhusar:
> Sigh, this version isn't in central
> 
> https://repo1.maven.org/maven2/net/shibboleth/utilities/java-support/8.0.0/java-support-8.0.0.pom https://repo1.maven.org/maven2/net/shibboleth/utilities/java-support/8.4.2/java-support-8.4.2.pom

So just to state the obvious, the build obtains this from the jboss nexus which has this cached; given that we don't define the shibboleth repo in our build. Magic!

bstansberry:
Yes, I asked David Hladky to add Shibboleth proxying to JBoss Nexus a few years ago. We'd not be able to get Opensaml either otherwise, unless we added the Shibboleth repo to peoples' configs.

This is why we say you should use JBoss Nexus repo as part of provisioning. (That and -jbossorg-# versions, which we are trying to avoid.)

fabiobrz (approved):
Relevant tests in jbossws-cxf are passing as well, and Apache CXF 4.0.11 depends on 8.4.2 too.
Thanks for the discussion about the last opensaml dependency version and the JBoss repo @rhusar and @bstansberry.

bstansberry:
Thanks @fabiobrz and @rhusar

PR #20026, opened 2026-05-18
[WFLY-21909] Fix assembly source path for legacy-ee distribution modules

yersan:
Jira issue: https://redhat.atlassian.net/browse/WFLY-21909


bstansberry (approved):

bstansberry:
Thanks @yersan 

I want to investigate why we are doing this at all in WF, but since we are this is correct.

PR #20037, opened 2026-05-19
WFLY-21915 - Downgrade Mojarra until TCK issues can be resolved

jasondlee:
https://redhat.atlassian.net/browse/WFLY-21915

Downgrade Mojarra in the EE 10 FP
Downgrade Mojarra in the EE 11 FP


bstansberry:
Thanks @jasondlee

PR #20040, opened 2026-05-20
[WFLY-21919] Remove the Elytron Web override as this is controlled in WildFly Core.

darranl:
This can be merged once we have a WildFly Core with Elytron Web 4.2.1.Final or later.

https://redhat.atlassian.net/browse/WFLY-21919

IMO we don't need to backport for WildFly 40.0.1, Elytron Web upgrades to the micro release are rare and we can still apply it should it be required.

### DIFF
--- a/boms/common-ee/pom.xml
+++ b/boms/common-ee/pom.xml
@@ -3296,31 +3296,6 @@
                 </exclusions>
             </dependency>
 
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

bstansberry:
Thanks @darranl

PR #20049, opened 2026-05-22
Test jbossws-cxf-7.4.0 consolidation

fabiobrz:
This is just a test to validate the last jbossws-cxf-7.4.0 Beta, which includes updated JBossWS dependencies, aligning with Apache CXF 4.1.x and WildFly. 

- Apache CXF → 4.1.6
- jbossws-api → https://repo1.maven.org/maven2/org/jboss/ws/jbossws-api/3.1.0.Beta1/ 
- jbossws-spi → https://repo1.maven.org/maven2/org/jboss/ws/jbossws-spi/5.1.0.Beta1/  
- jbossws-common → https://repo1.maven.org/maven2/org/jboss/ws/jbossws-common/5.2.0.Beta1/ 
- jbossws-common-tools → https://repo1.maven.org/maven2/org/jboss/ws/jbossws-common-tools/2.2.0.Beta1/ 
- jaxws-undertow-httpspi → https://repo1.maven.org/maven2/org/jboss/ws/projects/jaxws-undertow-httpspi/3.0.0.Beta1/ (a major, but just because of previous configuration, could be a minor)
- jboss-jaxb-intros → https://repo1.maven.org/maven2/jboss/jaxbintros/jboss-jaxb-intros/2.1.0.Beta1/ 

- jbossws-cxf →e.g.: https://repo1.maven.org/maven2/org/jboss/ws/cxf/jbossws-cxf-server/7.4.0.Beta3/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -499,7 +499,7 @@
         <version.org.apache.activemq.artemis>2.53.0</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>2.0.0</version.org.apache.activemq.artemis.native>
         <version.org.apache.avro>1.12.1</version.org.apache.avro>
-        <version.org.apache.cxf>4.0.11</version.org.apache.cxf>
+        <version.org.apache.cxf>4.1.6</version.org.apache.cxf>
         <version.org.apache.cxf.xjcplugins>4.1.3</version.org.apache.cxf.xjcplugins>
         <version.org.apache.httpcomponents.httpasyncclient>4.1.5</version.org.apache.httpcomponents.httpasyncclient>
         <version.org.apache.james.apache-mime4j>0.8.14</version.org.apache.james.apache-mime4j>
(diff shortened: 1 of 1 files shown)


fabiobrz:
@bstansberry WDYT about the new major version? 
That is because the POM was previously configured fr the next major, instead of a minor.
Anyway, from the API compatibility perspective it can be considered as a minor, since nothing changed in the exposed APIs.

fabiobrz:
All checks passed, just one unrelated failure in one combination.

rhusar:
@fabiobrz Made this a draft since this is not meant to be merged so it doesn't show up in the mergers filters.

fabiobrz:
> @fabiobrz Made this a draft since this is not meant to be merged so it doesn't show up in the mergers filters.

Thanks @rhusar!

bstansberry:
> @bstansberry WDYT about the new major version? That is because the POM was previously configured fr the next major, instead of a minor. Anyway, from the API compatibility perspective it can be considered as a minor, since nothing changed in the exposed APIs.

@fabiobrz If it's compatible, it's fine.

Even if something is not compatible, if it's only used in a jboss-api=private module and not one that our subsystems add  as a dep to the deployment, it's fine. Of course our own use of it would need to deal with the incompatibility. (The private but added as a dep to deployments case needs more care. But only if there is some incompatibility.)

fabiobrz:
Superseded by https://github.com/wildfly/wildfly/pull/20072

PR #20050, opened 2026-05-22
[WFLY-21926] Extract properties from root pom.xml to specific bom

yersan:
Jira issue: https://redhat.atlassian.net/browse/WFLY-21926


bstansberry:
We should use #20052, which builds on this.

PR #20052, opened 2026-05-22
[WFLY-21926] Extract properties from root pom.xml to specific bom

bstansberry:
https://redhat.atlassian.net/browse/WFLY-21926

Resolves the problem breaking CI in #20050 

@yersan Hopefully this doesn't break the alignment use case you were working. I don't think there is any alternative to using a root pom property of some form for this.

This at least somewhat addresses a goal I have in mind for your change of making it more clear when a PR is mixing shared dependency changes with EE 11-specific ones. Your change does the bulk of that by splitting the properties for most deps into different poms so its clearer that there's mixing if that occurs. The Hibernate ORM version is still vulnerable to mixing concerns, but it's just one comment and it has an unusual name and a comment to help make the prop visually unusual.

Note that mixing isn't necessarily wrong; it just requires more care to ensure the shared component also works well with its EE 10 dependency.


yersan (approved):

bstansberry:
Thanks @yersan

yersan:
Kudos to @TomasHofman who found the issue we had replacing properties at the alignment phase!

PR #20058, opened 2026-05-26
[WFLY-21930] Fix EJB-over-HTTP session stickiness path mismatch

rafaeldarosagit:
JIRA Issues:
* Upstream: [WFLY-21930](https://redhat.atlassian.net/browse/WFLY-21930)
* Downstream: [JBEAP-33250](https://redhat.atlassian.net/issues?filter=51706&selectedIssue=JBEAP-33250)

This PR fixes a session stickiness issue in EJB-over-HTTP where a path mismatch was preventing the session from being correctly associated with the request.

### DIFF
--- a/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
+++ b/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.test.integration.ejb.remote.http;
+
+import java.net.URL;
+
+import org.apache.http.Header;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Tests that the HTTP invoker session cookie has the correct path attribute.
+ * This test verifies the fix for [WFLY-21930] where the cookie path was missing
+ * the leading slash, causing load balancers and HTTP clients to reject the cookie.
+ *
+ * @author Rafael Rosa
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class HttpInvokerCookiePathTestCase {
+
+    private static final String PATH_PREFIX = "path=";
+
+    @ArquillianResource
+    private URL url;
+
+    @Deployment
+    public static Archive<?> deploy() {
+        return ShrinkWrap.create(WebArchive.class, HttpInvokerCookiePathTestCase.class.getSimpleName() + ".war");
+    }
+
+     /**
+     * Verifies that the JSESSIONID cookie set by the HTTP invoker endpoint
+     * has a path attribute that starts with a forward slash.
+     */
+    @Test
+    public void testHttpInvokerCookiePathHasLeadingSlash() throws Exception {
+        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+            // Access the wildfly-services endpoint to trigger HTTP invoker session cookie
+            String wildflyServicesUrl = url.getProtocol() + "://" + url.getHost() + ":" + url.getPort() + "/wildfly-services";
+            HttpGet request = new HttpGet(wildflyServicesUrl);
+
+            try (CloseableHttpResponse response = httpClient.execute(request)) {
+                // Inspect Set-Cookie headers
+                Header[] setCookieHeaders = response.getHeaders("Set-Cookie");
+
+                boolean foundJSessionId = false;
+                String actualPath = null;
+
+                // Look for JSESSIONID cookie from HTTP invoker
+                for (Header header : setCookieHeaders) {
+                    String cookieValue = header.getValue();
+
+                    if (cookieValue.contains("JSESSIONID=")) {
+                        foundJSessionId = true;
+
+                        // Extract path attribute
+                        String[] parts = cookieValue.split(";");
+                        for (String part : parts) {
+                            String trimmed = part.trim();
+                            if (trimmed.toLowerCase().startsWith(PATH_PREFIX)) {
+                                actualPath = trimmed.substring(PATH_PREFIX.length());
+                                break;
+                            }
+                        }
+                        break;
+                    }
+                }
+
+                Assert.assertTrue(
+                    "HTTP invoker endpoint should set JSESSIONID cookie",
+                    foundJSessionId
+                );
+                Assert.assertNotNull(
+                    "JSESSIONID cookie should have a path attribute",
+                    actualPath
+                );
+                Assert.assertTrue(
+                    "HTTP invoker cookie path must start with '/', but was: " + actualPath,
+                    actualPath.startsWith("/")
+                );
+                Assert.assertEquals(
+                    "HTTP invoker cookie path should be '/wildfly-services'",
+                    "/wildfly-services",
+                    actualPath
+                );
+            }
+        }
+    }
+}
--- a/undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
@@ -64,7 +64,7 @@ final class HttpInvokerHostService implements Service {
         }
 
         SessionCookieConfig sessionConfig = new SessionCookieConfig();
-        sessionConfig.setPath(this.path);
+        sessionConfig.setPath(UndertowUtils.normalizePath(this.path));
         Server server = this.host.get().getServer();
         ServletContainerService container = server.getServletContainer();
         CookieConfig affinityCookeConfig = container.getAffinityCookieConfig();
@@ -147,3 +147,4 @@ final class HttpInvokerHostService implements Service {
         return path;
     }
 }
+
--- a/undertow/src/main/java/org/wildfly/extension/undertow/UndertowUtils.java
+++ b/undertow/src/main/java/org/wildfly/extension/undertow/UndertowUtils.java
@@ -1,0 +1,19 @@
+package org.wildfly.extension.undertow;
+
+public final class UndertowUtils {
+
+    private UndertowUtils() {
+    }
+
+    /**
+     * Normalizes the path to ensure it starts with a leading slash.
+     *
+     * @param path the path to normalize
+     * @return the normalized path with a leading slash, or {@code null} if the input path was null
+     */
+    public static String normalizePath(String path) {
+        if (path == null) return null;
+        return path.startsWith("/") ? path : "/" + path;
+    }
+
+}
--- a/undertow/src/test/java/org/wildfly/extension/undertow/UndertowUtilsTestCase.java
+++ b/undertow/src/test/java/org/wildfly/extension/undertow/UndertowUtilsTestCase.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.wildfly.extension.undertow;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+/**
+ * Tests for {@link UndertowUtils}.
+ *
+ * @author Rafael Rosa
+ */
+public class UndertowUtilsTestCase {
+
+    @Test
+    public void testNormalizePathWithoutLeadingSlash() {
+        assertEquals("/wildfly-services", UndertowUtils.normalizePath("wildfly-services"));
+    }
+
+    @Test
+    public void testNormalizePathWithLeadingSlash() {
+        assertEquals("/wildfly-services", UndertowUtils.normalizePath("/wildfly-services"));
+    }
+
+    @Test
+    public void testNormalizePathEmpty() {
+        assertEquals("/", UndertowUtils.normalizePath(""));
+    }
+
+    @Test
+    public void testNormalizePathNull() {
+        assertEquals(null, UndertowUtils.normalizePath(null));
+    }
+
+    @Test
+    public void testNormalizePathRoot() {
+        assertEquals("/", UndertowUtils.normalizePath("/"));
+    }
+
+    @Test
+    public void testNormalizePathComplex() {
+        assertEquals("/api/v1/services", UndertowUtils.normalizePath("api/v1/services"));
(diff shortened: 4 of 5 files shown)


undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
```diff
@@ -64,7 +64,7 @@ public void start(final StartContext startContext) {
         }
 
         SessionCookieConfig sessionConfig = new SessionCookieConfig();
-        sessionConfig.setPath(this.path);
+        sessionConfig.setPath("/" + this.path);
```
rhusar: The fix is fragile — it doesn't handle user-configured paths correctly. The PATH attribute is user-configurable. If someone sets path="/my-services" (with a leading slash), this change would produce "//my-services".

rhusar (changes requested):

undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
rhusar: The simple naive fix would be

```suggestion
        sessionConfig.setPath(this.path.startsWith("/") ? this.path : "/" + this.path);
```

rhusar:
Also, missing test coverage.

undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
rafaeldarosagit: @rhusar Updated the logic to handle user-configured paths correctly. Thanks for the review!

rafaeldarosagit:
@rhusar Added tests.

mskacelik:
@rafaeldarosagit while unit tests are nice, I think this bug fix would benefit from a proper Integration Test, that proofs that the bug is actually fixed. Maybe take an inspiration from:
- https://github.com/wildfly/wildfly/blob/353fcbe88cbae835c3471fef1ac198e3edbbe9a5/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/web/headers/authentication/ResponseHeaderAuthenticationTestCase.java - `Set-Cookie` header inspection using Apache `HttpClient` `response.getHeaders("Set-Cookie")` (atr. `path=` in the `JSESSIONID`)
- https://github.com/wildfly/wildfly/blob/353fcbe88cbae835c3471fef1ac198e3edbbe9a5/testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/httpobfuscatedroute/HttpObfuscatedRouteInvocationInContainerTestCase.java - client-side ARQ test targeting the http-invoker endpoint at `/wildfly-services`

@rhusar WDYT?

undertow/src/test/java/org/wildfly/extension/undertow/HttpInvokerHostServiceTestCase.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link HttpInvokerHostService}.
+ *
+ * @author Rafael Rosa
+ */
+public class HttpInvokerHostServiceTestCase {
```
rhusar: I don't find this test case particularly useful, also, it's not a HttpInvokerHostServiceTestCase but really a `normalizePath` method test case.

rhusar (changes requested):
+1 @mskacelik that's what I had in mind originally - an integration test.

undertow/src/test/java/org/wildfly/extension/undertow/HttpInvokerHostServiceTestCase.java
rafaeldarosagit: @rhusar and @mskacelik,

I have added the integration test (HttpInvokerCookiePathTestCase) using Arquillian and Apache HttpClient to inspect the cookie path as suggested.

darranl:
I have added the 41.x and Blocker labels to this PR as that is the state of the Jira issue - obviously that can be evaluated but flagging for now.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @ArquillianResource
+    private URL url;
+
+    @Deployment
+    public static Archive<?> deploy() {
+        WebArchive war = ShrinkWrap.create(WebArchive.class, "http-cookie-path-test.war");
+        war.addClasses(EchoBean.class, EchoRemote.class);
+        return war;
+    }
```
mskacelik, line 44:
```suggestion
    @Deployment
    public static Archive<?> deploy() {
        return ShrinkWrap.create(WebArchive.class, "http-cookie-path-test.war");
    }
```

I think you can remove these classes from the deployment.

undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
```diff
@@ -146,4 +146,17 @@ public void authenticationComplete(SecurityIdentity securityIdentity, String mec
...
     }
+
+    /**
+     * Normalizes the path to ensure it starts with a leading slash.
+     * This is required for proper cookie path matching in the HTTP client.
+     *
+     * @param path the path to normalize
+     * @return the normalized path with a leading slash, or {@code null} if the input path was null
+     */
+    static String normalizePath(String path) {
```
mskacelik: Maybe we should create a new utility class for this? WildFly (esp. in the undertow extension) actually uses the exact same pattern here (and there might be others): https://github.com/wildfly/wildfly/blob/f73aa1e33ededa6db146d9b930b061bc9713acb3/undertow/src/main/java/org/wildfly/extension/undertow/Host.java#L295

It's a solid excuse to decouple the API and get some proper unit test coverage on it.

undertow/src/test/java/org/wildfly/extension/undertow/HttpInvokerHostServiceTestCase.java
mskacelik: This could be Test case for the Utility class.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+
+                    if (cookieValue.contains("JSESSIONID=")) {
+                        foundJSessionId = true;
+
+                        // Extract path attribute
+                        String[] parts = cookieValue.split(";");
+                        for (String part : parts) {
+                            String trimmed = part.trim();
+                            if (trimmed.toLowerCase().startsWith("path=")) {
+                                actualPath = trimmed.substring(5);
```
mskacelik: Maybe you could have a`"path="` as a variable, and then you would just use `.length()` instead of explicit `5`, what do you think @rafaeldarosagit?

rafaeldarosagit:
@mskacelik done, thanks!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
rafaeldarosagit: Done.

undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerHostService.java
rafaeldarosagit: Done.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
rafaeldarosagit: Great idea, done.

undertow/src/test/java/org/wildfly/extension/undertow/HttpInvokerHostServiceTestCase.java
rafaeldarosagit: Done.

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.net.URL;
```
rhusar: Please follow the coding standards in this project

https://docs.wildfly.org/40/Testsuite.html#follow-coding-standards

rhusar (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+public class HttpInvokerCookiePathTestCase {
+
+    private static final String PATH_PREFIX = "path=";
+
+    @ArquillianResource
+    private URL url;
+
+    @Deployment
+    public static Archive<?> deploy() {
+        return ShrinkWrap.create(WebArchive.class, "http-cookie-path-test.war");
```
rhusar: Don't introduce arbitrary strings, just use test class name as the deployment name.

rhusar (changes requested):

undertow/src/test/java/org/wildfly/extension/undertow/UndertowUtilsTestCase.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.wildfly.extension.undertow;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
```
rhusar: Please follow the coding standards in this project

https://docs.wildfly.org/40/Testsuite.html#follow-coding-standards

rhusar (changes requested):

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
rafaeldarosagit: @rhusar Done!

testsuite/integration/basic/src/test/java/org/jboss/as/test/integration/ejb/remote/http/HttpInvokerCookiePathTestCase.java
rafaeldarosagit: @rhusar Done!

undertow/src/test/java/org/wildfly/extension/undertow/UndertowUtilsTestCase.java
rafaeldarosagit: @rhusar Done!

rhusar (dismissed):
I had a brief look but I don't think any of this approach is correct.

The starting slash should be validated by the model and wrong value fixed by the parser/value converter.

rafaeldarosagit:
@rhusar Thanks for the update! If I understood correctly, we want to handle the leading slash normalization at the model level rather than inside the runtime service logic, right?

@rhusar @pferraro Could you clarify what you have in mind for "a different fix" at the model level, or point me to a reference in the codebase? Once confirmed, I'll update the PR with the new approach.

bstansberry:
@rhusar I think adding a ParameterCorrector is out-of-scope for this. Primarily because HttpInvokerDefinition.PATH has a default value 'wildfly-services' which itself has no leading slash. To be consistent we'd need to change the default and bump the management API version.

If we were going to do that it seems like that attribute not supporting expressions is a bug that could be fixed as well.

Neither a default value (if we didn't change it) nor the value of a resolved expression would be fixed by a ParameterCorrector, so the value returned from PATH.resolveModelAttribute would still need to be adjusted.

That seems like too much, particularly since there is a downstream need for this that would not deal with any API change.


All that said my IDE tells me that in the end all values of 'path' that end up getting normalized are coming from 

https://github.com/wildfly/wildfly/blob/main/undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerDefinition.java#L99

```String path = PATH.resolveModelAttribute(context, model).asString();```

If that is correct (@rafaeldarosagit please confirm), the line after that is the place to do any adjustment. Anywhere else that is currently adjusting a path can just add an assert to confirm that a internal contract that there is a leading slash has been met.


@rafaeldarosagit The rest of this is just to answer your question about how this kind of thing could be done, _if it was the thing to do_. (I wrote all this before realizing I don't think it is the right thing to do and writing everything above.)

As noted above, my IDE tells me that in the end all values of 'path' that end up getting normalized are coming from 

https://github.com/wildfly/wildfly/blob/main/undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerDefinition.java#L99

```String path = PATH.resolveModelAttribute(context, model).asString();```

What comes out of that PATH.resolveModelAttribute call depends on the configuration of the PATH attribute which is set at https://github.com/wildfly/wildfly/blob/main/undertow/src/main/java/org/wildfly/extension/undertow/HttpInvokerDefinition.java#L65-L69

You would need to add a builder call to setCorrector(ParameterCorrector) passing in a ParameterCorrector impl.

A correct ParameterCorrector impl isn't totally trivial, e.g. it's not just a call to your util method, because it can't assume the ModelNode provided with the operation it is correcting is a logical string, i.e. of ModelType.STRING. Most commonly it might be ModelType.UNDEFINED (equivalent to null) or it might be an expression (ModelType.EXPRESSION). More uncommonly it might be ModelType.INT or some other invalid type. A ParameterCorrector should not 'correct' any of those cases; doing that is not the job of a ParameterCorrector.

See org.jboss.as.controller.CaseParameterCorrector for a couple examples of correctors that handle these things appropriately.

**Comment by bstansberry** [comment]
/ok-to-test

As noted above, this should be changed, but in the end the basic behavior of ensuring there's always a leading / will be in place, so there's value in letting CI have at it.

**Comment by rafaeldarosagit** [comment]
> **[bstansberry](/bstansberry) ** c

Thanks for the detailed explanation and analysis, @bstansberry ! 
That makes complete sense. I'll review the flow again, make the suggested changes, and run a few tests to validate everything. Once I've confirmed the behavior, I'll update the PR with the changes.

File: undertow/src/main/java/org/wildfly/extension/undertow/UndertowUtils.java [hunk 9]
```diff
@@ -0,0 +1,19 @@
+package org.wildfly.extension.undertow;
```
rhusar, line 1: Missing copyright header ;(

rhusar (changes requested):
The path normalization should happen in one place, rather than in multiple places; perhaps in HttpInvokerHostService.start

bstansberry:
Thanks @rafaeldarosagit and @rhusar. I will file a separate issue to take care of the things that were discussed here.

rhusar:
Sounds good Brian and thanks @rafaeldarosagit

PR #20059, opened 2026-05-26
[WFLY-21931] Build modules do not override the wildfly.channels when …

yersan:
…it is externally set

Jira issue: https://redhat.atlassian.net/browse/WFLY-21931

For the sake of completeness, I've also added changes to make the wildfly-maven-plugin configurations in builds* modules similar to the dist* ones


bstansberry (approved):

bstansberry:
Thanks @yersan

PR #20062, opened 2026-05-26
Disable dependabot for Hibernate ORM

bstansberry:
I'm tired of broken PRs like https://github.com/wildfly/wildfly/pull/20057.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,9 @@ updates:
         update-types: ['version-update:semver-major', 'version-update:semver-minor']
       # WildFly Clustering updates are usually non-trivial
       - dependency-name: 'org.wildfly.clustering:*'
+      # Dependabot doesn't understand how to upgrade Hibernate ORM correctly, 
+      # and when we do we want to check other components where we align with what ORM uses.
+      - dependency-name: 'org.hibernate.orm:*'
       # Dependencies exclusive to Hibernate Search:
       # we'll only upgrade those when we upgrade to a newer version of Hibernate Search.
       - dependency-name: 'org.elasticsearch.client:*'


rhusar:
Really, the right way to do this is similar to what we do in other projects where we support multiple versions of the same library; for example here - https://github.com/modcluster/mod_cluster/blob/main/.github/dependabot.yml

However, that might require not having both versions within one root pom or getting rid of the properties that are getting little out of hand.

rhusar:
@pferraro points out we could use this feature https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#exclude-paths-

.github/dependabot.yml
```diff
@@ -19,6 +19,9 @@ updates:
         update-types: ['version-update:semver-major', 'version-update:semver-minor']
       # WildFly Clustering updates are usually non-trivial
       - dependency-name: 'org.wildfly.clustering:*'
+      # Dependabot doesn't understand how to upgrade Hibernate ORM correctly, 
+	  # and when we do we want to check other components where we align with what ORM uses.
```
rhusar: n.b. stray 'tab' instead of spaces.

scottmarlow (approved):

bstansberry:
Thanks, @rhusar for the info. I'll fix the tab thing.

The ORM properties unfortunately need to be in the same pom.

Regardless of the technical aspects, IMHO this is a dep that should be human managed. Every time there's an upgrade we want to know if other components that we align with ORM can/should be updated. This means the odds are low of the dependabot PR being merged as is. @scottmarlow is very much plugged into the ORM community and knows when there are releases, as does the wildfly-dev report.

rhusar:
> Thanks, @rhusar for the info. I'll fix the tab thing.
> 
> The ORM properties unfortunately need to be in the same pom.
> 
> Regardless of the technical aspects, IMHO this is a dep that should be human managed. Every time there's an upgrade we want to know if other components that we align with ORM can/should be updated. This means the odds are low of the dependabot PR being merged as is. @scottmarlow is very much plugged into the ORM community and knows when there are releases, as does the wildfly-dev report.

@bstansberry Oh I agree – and we do the same for e.g. wildfly-clustering.

My comment was meant more of 'if we wanted to keep these updated by dependabot what would it take' and it's not just practical for the time being.

bstansberry:
Thanks @scottmarlow and @rhusar

PR #20065, opened 2026-05-27
Fix for WFLY-21943, Upgrade Galleon to 7.0.8.Final and Galleon plugins to 8.1.6.Final

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFLY-21943


bstansberry:
Thanks @jfdenise

PR #20066, opened 2026-05-27
Fix for WFLY-21939, Glow metadata tests should resolve feature-packs in offline mode

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/WFLY-21939

There is a component upgrade to 2.1.0.Final for Glow: https://github.com/wildfly/wildfly-glow/compare/2.0.0.Final...2.1.0.Final


bstansberry:
Thanks @jfdenise

PR #20071, opened 2026-05-28
[WFLY-21946] Remove maven warning and drive org.wildfly.plugins:wildf…

yersan:
…ly-maven-plugin via root pom to avoid getting the same warning in the future

Jira issue: https://redhat.atlassian.net/browse/WFLY-21946


rhusar (approved):
LGTM

I wonder why this was set up like this from the beginning rather than using pluginManagement which is a standard maven practice.

bstansberry (approved):
Thank you!

rhusar:
Thanks @yersan and @bstansberry

PR #20074, opened 2026-05-28
[WFLY-15867] Replace XiPKI based OCSP server with BouncyCastle based responder

lvydra:
https://redhat.atlassian.net/browse/WFLY-15867
https://redhat.atlassian.net/browse/WFLY-19144


testsuite/integration/elytron/pom.xml
```diff
@@ -35,15 +35,9 @@
...
-        <!-- WFLY-16155 temporary pin bouncycastle 1.69 for test only.
-             Remove pinning with WFLY-15867 upgrading org.xipki.pki tests
-             and change to org.bouncycastle:*-jdk18on -->
-        <version.org.bouncycastle>1.69</version.org.bouncycastle>
         <version.org.mock-server.mockserver>5.6.1</version.org.mock-server.mockserver>
         <version.org.mock-server.mockserver-netty>5.6.1</version.org.mock-server.mockserver-netty>
-        <version.org.xipki.pki.ocsp-server>3.0.0</version.org.xipki.pki.ocsp-server>
-        <version.javax.xml.bind.jaxb-api>2.4.0-b180830.0359</version.javax.xml.bind.jaxb-api>
-        <version.org.glassfish.jaxb.jaxb-runtime>2.4.0-b180830.0438</version.org.glassfish.jaxb.jaxb-runtime>
+        <version.org.glassfish.jaxb.jaxb-runtime>3.0.2</version.org.glassfish.jaxb.jaxb-runtime>
```
bstansberry: Does 4.0.8 work?

The org.glassfish.jaxb:jaxb-runtime artifact is already a standard server dependency, so not managing a separate version here would be nice.

lvydra: @bstansberry Thank you for the review.  The dependency should be removed now.

bstansberry:
@skyllarr or @darranl Please review. This looks ok to me; i.e. it implements a server and handles requests in what seems like a reasonable way. If it did that incorrectly I figure a test would fail. :). But I don't pretend to understand this in depth.

@lvydra If this approach is what we go with please tweak the JIRA titles a bit to better reflect the change.

bstansberry:
Thanks @lvydra 

@darranl This seemed fine to me and where we can I want to move past the use of really old deps, even in the testsuite.

PR #20079, opened 2026-05-29
WFLY-21928 Adding a fork stack protocol throws a duplicate resource ISE during runtime.

pferraro:
https://redhat.atlassian.net/browse/WFLY-21928

Unlike the channel resource, the fork resource already contains protocol child resources.


michpetrov (approved):
LGTM? I don't know enough about the internals but I see the runtime step I tried to avoid is necessary for the protocol properties.

rhusar (approved):
LGTM

rhusar:
Thanks @pferraro and @michpetrov

PR #20093, opened 2026-06-01
WFLY-21959 - [CVE-2026-45292] Upgrade SmallRye OpenTelemetry to 2.15.0

jasondlee:
https://redhat.atlassian.net/browse/WFLY-21959
https://redhat.atlassian.net/browse/WFLY-22004

Update SmallRye OpenTelemetry version
Update transitive dependency versions
Modify JBoss Modules to add new required upstream artifacts

### DIFF
--- a/boms/common-expansion/pom.xml
+++ b/boms/common-expansion/pom.xml
@@ -450,6 +450,17 @@
             </dependency>
             <dependency>
                 <groupId>io.opentelemetry</groupId>
+                <artifactId>opentelemetry-common</artifactId>
+                <version>${version.io.opentelemetry.opentelemetry}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry</groupId>
                 <artifactId>opentelemetry-context</artifactId>
                 <version>${version.io.opentelemetry.opentelemetry}</version>
                 <exclusions>
@@ -616,6 +627,17 @@
             <dependency>
                 <groupId>io.opentelemetry.instrumentation</groupId>
                 <artifactId>opentelemetry-runtime-telemetry-java8</artifactId>
+                <version>${version.io.opentelemetry.instrumentation}-alpha</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry.instrumentation</groupId>
+                <artifactId>opentelemetry-runtime-telemetry</artifactId>
                 <version>${version.io.opentelemetry.instrumentation}-alpha</version>
                 <exclusions>
                     <exclusion>
@@ -877,6 +899,17 @@
             <dependency>
                 <groupId>io.smallrye.opentelemetry</groupId>
                 <artifactId>smallrye-opentelemetry-rest</artifactId>
+                <version>${version.io.smallrye.smallrye-opentelemetry}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>io.smallrye.opentelemetry</groupId>
+                <artifactId>smallrye-opentelemetry-senders</artifactId>
                 <version>${version.io.smallrye.smallrye-opentelemetry}</version>
                 <exclusions>
                     <exclusion>
--- a/galleon-pack/galleon-shared/pom.xml
+++ b/galleon-pack/galleon-shared/pom.xml
@@ -78,11 +78,13 @@
         <dependency><groupId>io.opentelemetry.instrumentation</groupId><artifactId>opentelemetry-instrumentation-api-incubator</artifactId></dependency>
         <dependency><groupId>io.opentelemetry.instrumentation</groupId><artifactId>opentelemetry-instrumentation-api</artifactId></dependency>
         <dependency><groupId>io.opentelemetry.instrumentation</groupId><artifactId>opentelemetry-runtime-telemetry-java8</artifactId></dependency>
+        <dependency><groupId>io.opentelemetry.instrumentation</groupId><artifactId>opentelemetry-runtime-telemetry</artifactId></dependency>
         <dependency><groupId>io.opentelemetry.proto</groupId><artifactId>opentelemetry-proto</artifactId></dependency>
         <dependency><groupId>io.opentelemetry.semconv</groupId><artifactId>opentelemetry-semconv</artifactId></dependency>
         <dependency><groupId>io.opentelemetry.semconv</groupId><artifactId>opentelemetry-semconv-incubating</artifactId></dependency>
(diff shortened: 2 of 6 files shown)


bstansberry:
@jasondlee The SR Otel update includes moves to Vert.x 5 and Mutiny 3.1.1

https://github.com/smallrye/smallrye-opentelemetry/pull/540/changes

I have no idea what the broader implications of that might be.

@kabir @hpehl @darranl FYI.

jasondlee:
> @jasondlee The SR Otel update includes moves to Vert.x 5 and Mutiny 3.1.1
> 
> https://github.com/smallrye/smallrye-opentelemetry/pull/540/changes

The upgrade to Vert.x 5 was reverted in main here: https://github.com/smallrye/smallrye-opentelemetry/commit/56331a2dacaf630b68327e66f11f301640a842e5

The 3.x branch moves to Vert.x 5, but main/2.x remains on 4:
- main: https://github.com/smallrye/smallrye-opentelemetry/blob/main/pom.xml#L26
- 2.14.0: https://github.com/smallrye/smallrye-opentelemetry/blob/2.14.0/pom.xml#L26

kabir:
Right, ideally I think we should stay on VertX 4 for the time being, which it sounds like is the case.
We have the same situation in reactive messaging where the current stream is on 4, and the next major is on VertX 5

bstansberry:
@jasondlee Unfortunately this fails badly with a SecurityManager. :-(

At least a couple failures are like this:

```
Caused by: java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getenv.OTEL_INSTRUMENTATION_EXPERIMENTAL_SPAN_SUPPRESSION_STRATEGY")" in code source "(vfs:/content/deployment1.war/WEB-INF/classes <no signer certificates>)" of "ModuleClassLoader for Module "deployment.deployment1.war" from Service Module Loader")
	at org.wildfly.security.elytron-base@2.9.1.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:341)
	at org.wildfly.security.elytron-base@2.9.1.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:233)
	at java.base/java.lang.System.getenv(System.java:1116)
	at io.opentelemetry.instrumentation.api//io.opentelemetry.instrumentation.api.internal.ConfigPropertiesUtil.getString(ConfigPropertiesUtil.java:82)
	at io.opentelemetry.instrumentation.api//io.opentelemetry.instrumentation.api.internal.ConfigPropertiesUtil.getString(ConfigPropertiesUtil.java:93)
	at io.opentelemetry.instrumentation.api//io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder.getSpanSuppressionStrategy(InstrumenterBuilder.java:394)
	at io.opentelemetry.instrumentation.api//io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder.buildSpanSuppressor(InstrumenterBuilder.java:375)
```

The ```InstrumenterBuilder.buildSpanSuppressor``` implementation differs in that the 1.48 class creates a SpanSuppressionStrategy in a static init an holds it there, while 1.62 creates it as part of the method invocation. I assume we do something in our integration such that the static init ran when only WF code was on the stack, but now the code runs as part of request handling.

bstansberry:
@radcortez FYI in case there are other Smallrye OTEL users who still need to support running with a SecurityManager.

jasondlee:
I have seen that. I've reproduced locally, but haven't had a chance to dig in yet. Hopefully tomorrow. Thanks for the analysis. That'll save me a bit of effort. :)

radcortez:
I don't remember us changing anything on how the integration works... (only the senders), but this code doesn't seem related to it. It may be an issue with an OTel upgrade.

@jasondlee let me know if you need any help.

jasondlee:
> I don't remember us changing anything on how the integration works... (only the senders), but this code doesn't seem related to it. It may be an issue with an OTel upgrade.
> 
> @jasondlee let me know if you need any help.

Thanks, @radcortez . It's likely a change in the newer otel version. I hope to dig in today. We'll see how other tasks cooperate. :)

jasondlee:
> @jasondlee Unfortunately this fails badly with a SecurityManager. :-(

FTR, this line, added in March of this year, is where the things go sideways:

```
this.contextClassLoader = Thread.currentThread().getContextClassLoader();
```

bstansberry:
> > @jasondlee Unfortunately this fails badly with a SecurityManager. :-(
> 
> FTR, this line, added in March of this year, is where the things go sideways:
> 
> ```
> this.contextClassLoader = Thread.currentThread().getContextClassLoader();
> ```

Ah, that must be a different one from the one I looked at / commented on, which was a System.getEnv call. Bummer.

jasondlee:
> Ah, that must be a different one from the one I looked at / commented on, which was a System.getEnv call. Bummer.

Well, that might still be the case. I just took the first failure I found locally and dove down the hole. There may well be others... :)

bstansberry:
> > I don't remember us changing anything on how the integration works... (only the senders), but this code doesn't seem related to it. It may be an issue with an OTel upgrade.
> > @jasondlee let me know if you need any help.
> 
> Thanks, @radcortez . It's likely a change in the newer otel version. I hope to dig in today. We'll see how other tasks cooperate. :)

Yes; sorry I was unclear!  The thing I looked at was definitely an OTel issue. I pinged you @radcortez just as an FYI that this update (which AIUI SR Otel made as well) results in this issue, in case other SR OTel consumers would care but might not have noticed.

bstansberry (dismissed):

bstansberry:
@jasondlee This still has SM failures. The first failures in the EE 10 SM run may not be SM-related; I'm not sure. The TC info for the first one didn't have obvious SM-related stuff in it but I didn't look carefully.

jasondlee:
> @jasondlee This still has SM failures. The first failures in the EE 10 SM run may not be SM-related; I'm not sure. The TC info for the first one didn't have obvious SM-related stuff in it but I didn't look carefully.

It looks like Reactive Messaging needs similar fix ups (@kabir). I'll get started on a PR there.

jasondlee:
@bstansberry  I ran the entire expansion suite locally on JDK 17 with sec man enabled (`mvn -U -Dinsecure.repositories=WARN -DallTests -Dmaven.test.failure.ignore=true -DfailIfNoTests=false -Dipv6 -Dsecurity.manager=true -Dlegacy-ee-full-server-tests -Dlegacy-ee-tests clean install`, a simplified version of what's run in CI), and all the tests pass (153 total). I *do* see some odd messages in the logs that I can't explain yet, but the tests pass. Not sure what to make of that. Just posting an update here to keep everyone in the loop.

bander-1:
Hi @jasondlee @bstansberry 
when do you plan to fix this CVE?

bstansberry:
@bander-1  -- In WildFly 42. The thing preventing this going in is it breaks SecurityManager use, and we've not been able to find workarounds for all the ways it does.  The security issue the upgrade would resolve assumes no limits to request sizes at the HTTP server level, and that is not the case with WildFly, so we have not been willing to break SecurityManager use to resolve this problem.

But WildFly 42 will not support running with the SecurityManager enabled, so no longer need to block on that problem.

jasondlee:
Testing shows that WildFly is not vulnerable to this issue thanks to protections already in place from Undertow, so I'm going to close this issue as invalid/obsolete. We'll do a proper update once MP 7.2 is finalized and released.

PR #20095, opened 2026-06-02
[WFLY-21961] Building server test suite fails with no.expansion.build

yersan:
Jira issue: https://redhat.atlassian.net/browse/WFLY-21961

This adds the requirement of also disable preview builds when no.expansion.build gets build.

I guess we can find a way to disable it automatically, but I just wanted to go using the simple approach

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1650,7 +1650,10 @@
             <id>expansion-build</id>
             <activation>
                 <property>
-                    <!-- Allow users to disable building expansion stuff if they just want base artifacts -->
+                    <!--
+                    Allow users to disable building expansion stuff if they just want base artifacts.
+                    This option requires -Dno.preview.build
+                    -->
                     <name>!no.expansion.build</name>
                 </property>
             </activation>
@@ -1663,6 +1666,7 @@
                 <module>build</module>
                 <module>dist</module>
                 <module>galleon-pack</module>
+                <module>legacy/channel</module>
                 <module>legacy/opentracing-extension</module>
                 <module>microprofile</module>
                 <module>observability</module>
@@ -1703,7 +1707,6 @@
                 <module>boms/legacy-test</module>
                 <module>concurrency/impl-30</module>
                 <module>legacy/ee-feature-pack</module>
-                <module>legacy/channel</module>
             </modules>
         </profile>
 


bstansberry:
> This adds the requirement of also disable preview builds when no.expansion.build gets build.

Why? I don't see the relationship of legacy/channel to building preview.

I think the change is fine (main is clearly wrong), so I'm just interested in understanding.

yersan:
I thought there were more dependencies, but you're right it should not be necesary to skip preview.

The main problem if we do not build the galleon-pack modules it is that preview  feature pack copies the annotation index package from the galleon-pack:
https://github.com/wildfly/wildfly/blob/main/galleon-pack/galleon-shared/src/main/resources/packages/unstable-api-annotation-index.wildfly-galleon-pack/pm/wildfly/tasks.xml#L8

And this task fails when provisions the preview feature pack because the wildfly-galleon-pack-unstable-api-annotation-index.zip does not exists since the unstable-api-annotation-classpath-indexer-plugin (https://github.com/wildfly/wildfly/blob/main/galleon-pack/galleon-shared/pom.xml#L670-L685) is not built at all because we skip building the galleon-feature-pack via `-Dno.expansion.build`

I was able to exclude the [unstable-api-annotation-index.wildfly-galleon-pack](https://github.com/wildfly/wildfly/tree/main/galleon-pack/galleon-shared/src/main/resources/packages/unstable-api-annotation-index.wildfly-galleon-pack) task from the copy based on a profile activated when `-Dno.expansion.build` is present:

```
       <profile>
            <id>exclude-galleon-annotation-index</id>
            <activation>
                <property>
                       <name>no.expansion.build</name>
                </property>
            </activation>
            <build>
                <plugins>
                    <plugin>
                        <groupId>org.apache.maven.plugins</groupId>
                        <artifactId>maven-resources-plugin</artifactId>
                        <executions>
                            <execution>
                                <id>copy-full-galleon-shared-feature-pack-resources</id>
                                <phase>process-resources</phase>
                                <goals>
                                    <goal>copy-resources</goal>
                                </goals>
                                <configuration combine.self="override">
                                    <outputDirectory>${basedir}/target/resources</outputDirectory>
                                    <escapeString>\</escapeString>
                                    <resources>
                                        <resource>
                                            <directory>${full.galleon.shared.resources.directory}</directory>
                                            <excludes>
                                                <exclude>packages/unstable-api-annotation-index.wildfly-galleon-pack/**</exclude>
                                            </excludes>
                                        </resource>
                                    </resources>
                                    <overwrite>true</overwrite>
                                </configuration>
                            </execution>
                        </executions>
                    </plugin>
                </plugins>
            </build>
        </profile>
```

but that opens another bug where even if the package is optional (https://github.com/wildfly/wildfly/blob/main/preview/galleon-local/src/main/resources/configs/standalone/model.xml#L47) for the preview feature pack, it seems Galleon does not ignores it failing with the following error:

```
[ERROR] Failed to execute goal org.wildfly.plugins:wildfly-maven-plugin:6.0.0.Final:provision (server-provisioning) on project wildfly-preview-build: Provisioning failed: Failed to resolve configuration  model standalone: Failed to resolve package unstable-api-annotation-index.wildfly-galleon-pack in wildfly-preview@maven(org.jboss.universe:community-universe):current#41.0.0.Beta1-SNAPSHOT -> [Help 1]
```

It seems a bug in Galleon and I am currently investigating it.

Ideally, instead of avoiding the copy of this package when building the preview feature pack, the Galleon task could have an option to skip the task, but I have not found how to do it, so I went straight to the alternative of not copying the package when `-Dno.expansion.build` is present

yersan:
Indeed https://github.com/wildfly/wildfly/blob/main/preview/galleon-local/src/main/resources/configs/standalone/model.xml#L47 is only present for the standalone model and not for the domain one, which seems could be another issue in this area. Maybe I am getting it wrong and that annotation index coming from the galleon-pack should not be in use for preview? that annotation index is built from the dependencies found on the galleon-pack build, but preview would use a different set of dependencies so ... not sure about it, the same applies to the annotation indexes built from the ee-feature-pack, I wonder now if that is correct ... I'm not fully familiar yet with this area ...

bstansberry:
@yersan How does this block this PR?

Apologies if you've explained; I haven't been able to properly digest what you've written, but my impression from scanning it is you are talking about trying to disable building expansion in general and why building preview must be disabled as well if you do that. And that's fine, thanks for the info, but isn't that a situation that already exists and can be handled separately?

FWIW, I don't think being able to build preview while also using -Dno.expansion.build is an important use case.

yersan:
> @yersan How does this block this PR?

@bstansberry  I added on hold because I was assuming that we don't want to disable preview if we are disabling expansion.

> but isn't that a situation that already exists and can be handled separately?

Yes, we can manage it in a different Jira but we would need to assume the condition that using `-Dno.expansion.build` will required `-Dno.preview.build` not because it is logically correct, but because it could tentatively be due a bug in Galleon which it seems does not make honor, in this case, to the optional=true attribute configuration at package level.

> FWIW, I don't think being able to build preview while also using -Dno.expansion.build is an important use case.

In such a case, I remove the hold label and let you decide after/before digesting my previous points.



About whether makes sense to inject into preview an annotation index made from classes prepared at the galleon-pack level ... I am not sure about this so I'll open a Zulip discussion after digesting from myself how it works (right now I'm not familiar with this area)

bstansberry:
Thanks @yersan 

Thinking about it more, I don't see trying to build WFP while also doing -Dno.expansion.build as conceptually valid and we shouldn't try and make it work.

"Expansion" is more than just the wildfly-galleon pack and the builds, dists etc produced from it. It also covers all the components and shared galleon content that expands upon what wildfly-ee/wildfly-ee-10 provide. Those things are fundamental parts of WFP so you can't build WFP if you don't build those.

So, needing to do -Dno.preview.build if you do -Dno.expansion.build is logically correct.

With Maven 4 we could perhaps improve ease of use by having -Dno.expansion.build disable WFP builds as well, since logically that is implied. But that needs Maven 4's ability to activate profiles based on property-x || property-y.

Regarding the annotation index thing, I did look at / think about that as part setting up the WildFly EE 10 stuff and consciously put in place what's there now. That's not **at all** to say that it's right -- WildFly EE 10 was a big change and I had to make a lot of decisions and for sure some were wrong. But it's not a case of something set up years ago that has had no thought since.

yersan:
> So, needing to do -Dno.preview.build if you do -Dno.expansion.build is logically correct.

Ok, Brian ... from your first question I got the impression this was not true. Thanks for the review

PR #20096, opened 2026-06-02
[WFLY-21823] Add a maven property to specify the JBoss Modules layer used for expansion modules

christosderetzis:
Issue: https://redhat.atlassian.net/browse/WFLY-21823

### DIFF
--- a/ee-dist/src/verifier/verifications.xml
+++ b/ee-dist/src/verifier/verifications.xml
@@ -45,47 +45,47 @@
       <exists>true</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/io/smallrye/fault-tolerance/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/io/smallrye/fault-tolerance/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/eclipse/microprofile/fault-tolerance/api/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/eclipse/microprofile/fault-tolerance/api/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/wildfly/extension/microprofile/fault-tolerance-smallrye/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/wildfly/microprofile/fault-tolerance-smallrye</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/wildfly/microprofile/fault-tolerance-smallrye</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/io/smallrye/openapi/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/io/smallrye/openapi/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/eclipse/microprofile/openapi/api/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/eclipse/microprofile/openapi/api/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/wildfly/extension/microprofile/openapi-smallrye/main</location>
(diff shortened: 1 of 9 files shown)


yersan (changes requested):
Hello @christosderetzis 

There is a limited list of modules which are moved from `modules/system/base` to `modules/system/layers/microprofile`. They are only the modules related to Microprofile functionality

The list is here: https://github.com/wildfly/wildfly/blob/72cc1633265db934f80f2ece0f9702f207f26d18/ee-dist/src/verifier/verifications.xml#L47-L90

In addition to path parametrization, we could go ahead and also ensure that the same list is provisioned in the dist modules (those modules provides Microprofile).

I mean:
for `ee-dist` the verification will verify that the modules are not there
for `dist` the verification will verify that the modules **are** there

Optional, it would be also nice to expand those verifications to the legacy modules:
for `legacy/ee-feature-pack/ee-dist` the verification will verify that the modules are not there
for `legacy/ee-feature-pack/dist` the verification will verify that the modules are there

yersan:
Well, I've just checked downstream and the full list of files is much bigger than that, but with the current one I think we will be OK as a proof those expected modules are where they should be

christosderetzis:
Hello @yersan

The verifications for:
 - `legacy/ee-feature-pack/ee-dist` the verification will verify that the modules are not there
 - for `ee-dist` the verification will verify that the modules are not there
 Already exist.
 
 I added the verifications just for:
 - `dist` the verification will verify that the modules are there
 - `legacy/ee-feature-pack/dist` the verification will verify that the modules are there

rhusar (changes requested):
Please squash and skip the merge commit, we don't want that pollution in SCM log. I think one commit for a small change like this is enough.

Also, your commits are not linked with your GH account, please check you git config.

build/src/verifier/verifications.xml
```diff
@@ -12,7 +12,7 @@
        'full.dist.product.slot' property
     3) jboss-modules.jar must exist
     4) standalone/configuration/standalone.xml must exist
-    5) modules/system/layers/base/org/jboss/as/product/${preview.dist.product.slot}/module.xml exists
+    5) modules/system/layers/${expansion.module.layer}/org/jboss/as/product/${preview.dist.product.slot}/module.xml exists
```
yersan: This module is not moved to a different path.

@christosderetzis Could you please crosscheck again your changes against the original list I placed as a comment before?

It looks like you are adding the parametrized path over more modules which were not in that list. Only the modules on that list are the modules that are moved to modules/system/layers/microprofile, thanks!

yersan (changes requested):

build/src/verifier/verifications.xml
christosderetzis: @yersan done !

ee-dist/src/verifier/verifications.xml
```diff
@@ -45,51 +45,51 @@
...
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/wildfly/security/elytron-jwt/main</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/wildfly/security/elytron-jwt/main</location>
       <exists>false</exists>
     </file>
     <file>
-      <location>target/${latest.server.output.dir}/modules/system/layers/base/org/wildfly/_internal/unstable-api-annotation-index/main/content/wildfly-ee-galleon-pack-unstable-api-annotation-index.zip</location>
+      <location>target/${latest.server.output.dir}/modules/system/layers/${expansion.module.layer}/org/wildfly/_internal/unstable-api-annotation-index/main/content/wildfly-ee-galleon-pack-unstable-api-annotation-index.zip</location>
```
yersan: This on is not in the original list posted as comment. It is not moved between upstream and downstream.

pom.xml
```diff
@@ -156,6 +156,7 @@
         <skipPublishing>false</skipPublishing>
 
         <!-- Galleon -->
+        <expansion.module.layer>base</expansion.module.layer>
```
yersan: Really it is not a galleon configuration property, but I do no have any other suggestion. Jira description stated this specific property name so, it should be fine as it is ... This comment is just in case anyone has any other feeling about this.

yersan (changes requested):
@christosderetzis Could you cross check again the original list? I spotted one module that won't be moved at all.

ee-dist/src/verifier/verifications.xml
christosderetzis: Oppps, sorry, fixed!

pom.xml
bstansberry: Good point, @yersan.

@christosderetzis  Please move this here:

https://github.com/wildfly/wildfly/blob/40.0.0.Final/pom.xml#L129

Please prefix it with the comment:

`<!-- JBoss Modules layer used for modules that provide expansion functionality -->`

I'd like it in that location because the preceding properties are all used for similar purposes -- making it easier to create branches based on on WF tag that have different build requirements from what WF had.

EDIT: this comment appears in the GH UI twice; the other one is in a conversation that is marked resolved. So this is resolved.

bstansberry (dismissed):

pom.xml
bstansberry: Thanks; this bit is resolved.

yersan (approved):

bstansberry:
Thanks @christosderetzis and @yersan!

PR #20099, opened 2026-06-03
[WFLY-21962] Fix invalid error logging in StatefulSessionSynchronizat…

tadamski:
…ionInterceptor

https://redhat.atlassian.net/browse/WFLY-21962

### DIFF
--- a/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
+++ b/ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
@@ -3086,8 +3086,8 @@ public interface EjbLogger extends BasicLogger {
     void remappingCacheAttributes(String address, ModelNode defClustered, ModelNode passivationDisabled);
 
     @LogMessage(level = ERROR)
-    @Message(id = 487, value = "Unexpected invocation state %s")
-    void unexpectedInvocationState(int state);
+    @Message(id = 487, value = "Unexpected invocation state %s for stateful bean %s/%s/%s")
+    void unexpectedInvocationState(int state, String applicationName, String moduleName, String componentName);
 
 //    @Message(id = 488, value = "Unauthenticated (anonymous) access to this Jakarta Enterprise Beans method is not authorized")
 //    SecurityException ejbAuthenticationRequired();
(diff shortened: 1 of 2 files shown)


bstansberry (approved):

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
```diff
@@ -3086,8 +3086,8 @@ public interface EjbLogger extends BasicLogger {
     void remappingCacheAttributes(String address, ModelNode defClustered, ModelNode passivationDisabled);
 
     @LogMessage(level = ERROR)
-    @Message(id = 487, value = "Unexpected invocation state %s")
-    void unexpectedInvocationState(int state);
+    @Message(id = 487, value = "Unexpected invocation state %s for stateful bean %s/%s/%s")
+    void unexpectedInvocationState(int state, String applicationName, String moduleName, String componentName);
```
bstansberry, line 3090: Note: none of the following is relevant to whether we merge this PR into WF main....

@tadamski Be a bit careful about this downstream with maintenance branches where l10n has been done.  The translated strings that drive the string format won't have these 3 new parameters. The results will be that the message will be emitted in English.

That could be fine; it depends on the tradeoff between the loss of info from not getting the translation vs the extra info you've added. I don't know the right answer there; I'm curious what @bmaxwell thinks.

You sent me down a rabbit hole here to understand what happens. :) JBoss Logging generates a base implementation of EjbLogger  EjbLogger_$logger that has an impl of each method here, plus a protected method, e.g. unexpectedInvocationState$str() that returns the format string. Then there is a subclass of that per language, e.g. EjbLogger_$logger_ja, that overrides those protected methods.  Once you change this string format to one where the l10n property files that drive this don't match, it won't generate the unexpectedInvocationState$str() method in the subclasses, so things the base English method in EjbLogger_$logger is used.

It's a nicely robust design.

bstansberry (approved):

rhusar (approved):
LGTM

ejb3/src/main/java/org/jboss/as/ejb3/logging/EjbLogger.java
rhusar: +1 good point on the downstream

Nevertheless, this is fine for wildfly main so we can proceed.

rhusar:
Thanks @tadamski @bstansberry and please @bmaxwell have a look at the discussion above.

bmaxwell:
@lvydra so downstream, it means the l10n language files need to be updated? Just adding the appName/moduleName/componentName without the 'for stateful bean' would be useful, something like:

    @Message(id = 487, value = "Unexpected invocation state %s - %s/%s/%s")
    void unexpectedInvocationState(int state, String applicationName, String moduleName, String componentName);

PR #20107, opened 2026-06-04
[WFLY-20800] Change second call of getEntityManager() to underlyingEntityManager

aidentZhang:
https://redhat.atlassian.net/browse/WFLY-20800

### DIFF
--- a/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/AbstractEntityManager.java
+++ b/jpa/subsystem/src/main/java/org/jboss/as/jpa/container/AbstractEntityManager.java
@@ -196,7 +196,7 @@ public abstract class AbstractEntityManager implements EntityManager {
             start = System.currentTimeMillis();
         try {
             final EntityManager underlyingEntityManager = getEntityManager();
-            T result = getEntityManager().find(entityClass, primaryKey);
+            T result = underlyingEntityManager.find(entityClass, primaryKey);
             detachNonTxInvocation(underlyingEntityManager);
             return result;
         } finally {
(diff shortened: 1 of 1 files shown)


bstansberry:
Thanks @aidentZhang. Can you amend the commit message to reference the JIRA? Thanks.

rhusar (approved):
Nice catch, this is a bug since 2011 or even longer.

It's not just that the call to `org.jboss.as.jpa.container.AbstractEntityManager#getEntityManager` is 'unneeded', it's not a simple getter, there is much more in the underlying implementations. Depending whether this is called in transactional context or not, the impl is involving transaction state lookups or registering with NonTxEmCloser, etc. I don't see a functional issue, but certainly a perf one.

scottmarlow:
> Nice catch, this is a bug since 2011 or even longer.

The bug definitely was added in 2011 as the AS5 (EJB3 project) JPA container didn't have an equivalent way to show what is going on exactly in a JPA application.

Now that @aidentZhang has fixed this bug users that enable TRACE logging for JPA/Persistence should see fewer unexpected problems.

rhusar:
> > Nice catch, this is a bug since 2011 or even longer.
> 
> 
> 
> The bug definitely was added in 2011 as the AS5 (EJB3 project) JPA container didn't have an equivalent way to show what is going on exactly in a JPA application.
> 
> 
> 
> Now that @aidentZhang has fixed this bug users that enable TRACE logging for JPA/Persistence should see fewer unexpected problems.  
> 

Excellent!

rhusar:
CI failures unreleated

rhusar:
Thanks @aidentZhang @bstansberry @scottmarlow 

...and congratulations on the first commit in WF!

aidentZhang:
Thank you all! I look forward to contributing more in the future!

PR #20117, opened 2026-06-08
[WFLY-16586] microprofile-reactive-streams-operators layer doesn't pr…

kabir:
…ovision org.wildfly.reactive.dep.jts

https://redhat.atlassian.net/browse/WFLY-16586

### DIFF
--- a/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-reactive-streams-operators/layer-spec.xml
+++ b/galleon-pack/galleon-shared/src/main/resources/layers/standalone/microprofile-reactive-streams-operators/layer-spec.xml
@@ -18,5 +18,6 @@
 
     <packages>
         <package name="org.wildfly.extension.microprofile.reactive-streams-operators-smallrye"/>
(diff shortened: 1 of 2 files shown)


bstansberry:
@kabir This doesn't touch https://github.com/wildfly/wildfly/blob/main/testsuite/shared/src/main/java/org/jboss/as/test/shared/LayersTestBase.java#L164-L165

So either:

1) The layers tests will fail on this PR when the jobs get to them.
2) Or, the fix probably doesn't do what's expected
3) Or, it does do what's expected but the comment there needs to be update to explain why that module is still not provisioned by any layer.

bstansberry (changes requested):
I didn't use the GH review UI for my previous comment, but something will need to change here, so I'm requesting 'something'.

kabir:
Thanks @bstansberry I've made a change to LayersTestBase in testsuite/shared, and both subclasses now pass

bstansberry:
CI failures are unrelated.

bstansberry:
Thanks @kabir

PR #20134, opened 2026-06-10
WFLY-21979   jgroups-tcp-fd socket binding port 57600 is in the OS ephemeral port range + WFLY-21975   FD_SOCK2 port_range should be set to 0 for deterministic port binding

rhusar:
Review linked issues for in-depth details about the issues.

Resolves
https://redhat.atlassian.net/browse/WFLY-21975
https://redhat.atlassian.net/browse/WFLY-21979

### DIFF
--- a/clustering/jgroups/extension/src/main/resources/jgroups-defaults.xml
+++ b/clustering/jgroups/extension/src/main/resources/jgroups-defaults.xml
@@ -54,7 +54,7 @@
         timeout_check_interval="5s"
     />
     <FD_SOCK port_range="0"/>
-    <FD_SOCK2 offset="1"/>
+    <FD_SOCK2 offset="1" port_range="0"/>
     <VERIFY_SUSPECT timeout="1s"/>
     <VERIFY_SUSPECT2 timeout="1s"/>
     <pbcast.NAKACK2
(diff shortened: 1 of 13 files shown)


yersan (commented):
Just out of curiosity, for the sake of completeness, should not all the jgroups-tcp-fd ports used in the tests files be changed?

e.g https://github.com/wildfly/wildfly/blob/main/testsuite/integration/clustering/clustering-ha.cli#L25

Since they are ephemeral ports, would not they add more resilience?

rhusar:
> Just out of curiosity, for the sake of completeness, should not all the jgroups-tcp-fd ports used in the tests files be changed?
> 
> e.g https://github.com/wildfly/wildfly/blob/main/testsuite/integration/clustering/clustering-ha.cli#L25
> 
> Since they are ephemeral ports, would not they add more resilience?

Good point, we want to adjust these too. But more improtantly - the line ABOVE what you commented is a probilem in latest CI run

> node-1 2026-06-10 18:54:00,932 ERROR [org.jgroups.protocols.relay.RELAY2] (thread-5,ejb,node-1(v=16.0.13)) node-1(v=16.0.13): failed starting relayer: java.net.BindException: No available port to bind to in range [7601 .. 7601]

as that chosen port would now conflict with this one. I will use something like offset 50.

rhusar:
@yersan Updated that too now - offset the bridge stack by 50 of the main transport.

yersan (approved):
I did not spot anything else, being ports used for local machine, I do not think they will make any influence in any domain mode related aspect and should not be considered as backward incompatible. So, if CI looks green, this LGTM

pferraro (approved):

rhusar:
Why is there no CI!!!!!

rhusar:
Still waiting for CI to pick this up...

bstansberry:
Thanks @rhusar, @pferraro and @yersan

PR #20136, opened 2026-06-10
Add skip-teamcity label to Dependabot PRs

rhusar:
No Jira needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ registries:
 updates:
   - package-ecosystem: "maven"
     directory: "/"
+    labels:
+      - "dependencies"
+      - "skip-teamcity"
     schedule:
       interval: "daily"
     ignore:
@@ -62,5 +65,8 @@ updates:
     # Workflow files stored in the
     # default location of `.github/workflows`
     directory: "/"
+    labels:
+      - "dependencies"
+      - "skip-teamcity"
     schedule:
       interval: "daily"


jasondlee (approved):

bstansberry:
We are chatting about this general topic in zulip now but I'll discuss here.

The effect of this is one of the many people in the triage role will need to remove the label to initiate CI.

I think that is fine. 

There are many, many such people, and some of us (e.g. me and you) scan the queue often enough that PRs will not rot untested. Checking these is trivial, so there's no significant burden on anyone who looks at these.

There are quite a few dependabot PRs where we know they are wrong. 

In many cases we ask others via a comment for a review before merging. We can enable CI when we do so and it often will be done before those reviewers even notice the ask.

There are times when CI is overloaded so this lets us defer testing new dependabot PRs until higher priority tasks are out of the way.

rhusar:
Zulip thread [#wildfly-developers > Disabling TeamCity CI for PRs using PR labels](https://wildfly.zulipchat.com/#narrow/channel/174184-wildfly-developers/topic/Disabling.20TeamCity.20CI.20for.20PRs.20using.20PR.20labels/with/601925077)

rhusar:
> We are chatting about this general topic in zulip now but I'll discuss here.
> 
> The effect of this is one of the many people in the triage role will need to remove the label to initiate CI.
> 
> I think that is fine.
> 
> There are many, many such people, and some of us (e.g. me and you) scan the queue often enough that PRs will not rot untested. Checking these is trivial, so there's no significant burden on anyone who looks at these.
> 
> There are quite a few dependabot PRs where we know they are wrong.
> 
> In many cases we ask others via a comment for a review before merging. We can enable CI when we do so and it often will be done before those reviewers even notice the ask.
> 
> There are times when CI is overloaded so this lets us defer testing new dependabot PRs until higher priority tasks are out of the way.

Agreed on all points – and that's the workflow and aspects I had in mind. There has been a lot of 'good CI' runs that were hours completely wasted as the PR was unusable.

n.b. I did check the logic in pull player to verify that CI runs will kick in after the label is removed and that checks (as pull-player is using pull model, not push model).

PR #20137, opened 2026-06-11
WFLY-21996 Fix flaky HotRodSessionExpirationTestCase loop and tighten assertions

rhusar:
The expiration retry loop contained hard assertions (CREATED_SESSIONS, session ID consistency) that failed on the first attempt before the loop could retry, defeating the purpose of the 30-attempt retry.

Rework the loop to:
- poll for the 3 expiration events: destroyed, removed, unbound
- once an event is observed on a node, only poll that node
- verify each event is received exactly once
- verify no expiration events on the other node
- assert all events were observed after the loop

Resolves
https://redhat.atlassian.net/browse/WFLY-21996

### DIFF
--- a/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
+++ b/testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
@@ -9,7 +9,7 @@ import static org.junit.jupiter.api.Assertions.*;
 import java.net.URL;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import jakarta.servlet.http.HttpServletResponse;
@@ -26,6 +26,7 @@ import org.jboss.as.test.clustering.cluster.AbstractClusteringTestCase;
 import org.jboss.as.test.clustering.cluster.web.DistributableTestCase;
 import org.jboss.as.test.clustering.cluster.web.EnableUndertowStatisticsSetupTask;
 import org.jboss.as.test.http.util.TestHttpClientUtils;
+import org.jboss.as.test.shared.TimeoutUtil;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
```diff
@@ -419,44 +422,75 @@ public void test(@ArquillianResource(SessionOperationServlet.class) @OperateOnDe
...
+                            removed = true;
+                        }
+                        if (hasUnbound) {
+                            assertEquals("7", response.getFirstHeader(SessionOperationServlet.UNBOUND_ATTRIBUTES).getValue());
+                            unbound = true;
+                        }
+
+                        // We observed an event on one of the nodes, so we will only continue to poll that node for the remaining events
+                        if (hasDestroyed || hasRemoved || hasUnbound) {
+                            urls.retainAll(Collections.singletonList(baseURL));
```
pferraro: This is fragile.

> The iterators returned by this class's iterator() and listIterator() methods are fail-fast: if the list is structurally modified at any time after the iterator is created, in any way except through the iterator's own remove() or add(E) methods, the iterator will throw a ConcurrentModificationException.

pferraro (changes requested):

testsuite/integration/clustering/src/test/java/org/jboss/as/test/clustering/cluster/web/expiration/SessionExpirationTestCase.java
rhusar: @pferraro Not in the current impl – the loop is always `break` after the `retainAll` call.

pferraro: That is why I said "fragile" rather than "broken".  Otherwise, this becomes a trap for future maintainers of this test.
Just use a separate reference to capture the baseURL of the cluster member on which the expiration event was received.

rhusar: Good point — replaced retainAll with a separate URL eventURL reference that captures the node on which the expiration event was received. Came up with an elegant solution where the inner loop now iterates over an immutable List.of(...).

pferraro (approved):

rhusar:
Rerunning the Windows CI run

rhusar:
> Rerunning the Windows CI run

If there is something in main that's causing this, rerunning won't help – [#ci-servers > Maven artifactory dirty reads @ 💬](https://wildfly.zulipchat.com/#narrow/channel/174177-ci-servers/topic/Maven.20artifactory.20dirty.20reads/near/605181521)

Rebased only to get proper clean run.

rhusar:
Thanks for the review @pferraro – another flaky test down, and with a fix for jgroups https://github.com/belaban/JGroups/pull/1032 we will be running clean very soon!

PR #20139, opened 2026-06-11
WFLY-21998 Deployments using web cache-containers configured with LEGACY marshaller throws NotSerializableException at runtime

pferraro:
https://redhat.atlassian.net/browse/WFLY-21998


rhusar (approved):
LGTM

n.b. a brief audit shows there are zero existing integration tests exercising the LEGACY marshaller core path. Should we add some? WDYT @pferraro ?

pferraro:
> Should we add some?

No - it's not worth the trouble:
* I'm planning to drop LEGACY soon
* Any test would not prevent a similar regression in the future, given the way LEGACY is implemented.

rhusar:
OK; agreed.

bstansberry:
Thanks @pferraro and @rhusar

PR #20140, opened 2026-06-11
WFLY-21920 Remote queries fail with "IllegalArgumentException: No marshaller registered for Protobuf type" when using container-managed RemoteCacheContainer with deployment-specific types

rhusar:
Resolves
https://redhat.atlassian.net/browse/WFLY-21920

### DIFF
--- a/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/client/ManagedRemoteCacheContainer.java
+++ b/clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/client/ManagedRemoteCacheContainer.java
@@ -16,6 +16,7 @@ import org.infinispan.client.hotrod.RemoteSchemasAdmin;
 import org.infinispan.client.hotrod.RemoteSchemasAdmin.SchemaOpResult;
 import org.infinispan.client.hotrod.configuration.Configuration;
 import org.infinispan.client.hotrod.configuration.RemoteCacheConfiguration;
+import org.infinispan.client.hotrod.impl.MarshallerRegistry;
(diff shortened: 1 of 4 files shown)


clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/client/ManagedRemoteCacheContainer.java
```diff
@@ -85,6 +87,9 @@ public <K, V> RemoteCache<K, V> getCache(String cacheName) {
                                 }
                             }
                         }
+                        // Create a per-deployment marshaller registry containing the deployment-aware marshaller
+                        deploymentSpecificMarshallerRegistry = new MarshallerRegistry();
+                        deploymentSpecificMarshallerRegistry.registerMarshaller(marshaller);
```
pferraro: You're on the right track, but this override needs to be more aggressive.
This only works the first time this method is invoked.
Subsequent calls will use the wrong MarshallerRegistry, and will be unable to use query.

pferraro (changes requested):

clustering/infinispan/extension/src/main/java/org/jboss/as/clustering/infinispan/client/ManagedRemoteCacheContainer.java
rhusar: Ah, very good catch – we can cover this in a test. I was just replacing code I had in the previous fix without moving the code appropriately. Fixed now.

pferraro:
Superseded by https://github.com/wildfly/wildfly/pull/20169

PR #20142, opened 2026-06-12
[WFLY-21999] - Upgrade Apache CXF components to 4.1.7

fabiobrz:
Upstream issue: https://redhat.atlassian.net/browse/WFLY-21999

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -477,7 +477,7 @@
         <version.org.apache.activemq.artemis>2.54.0</version.org.apache.activemq.artemis>
         <version.org.apache.activemq.artemis.native>2.0.0</version.org.apache.activemq.artemis.native>
         <version.org.apache.avro>1.12.1</version.org.apache.avro>
-        <version.org.apache.cxf>4.1.6</version.org.apache.cxf>
+        <version.org.apache.cxf>4.1.7</version.org.apache.cxf>
         <version.org.apache.cxf.xjcplugins>4.1.3</version.org.apache.cxf.xjcplugins>
         <version.org.apache.httpcomponents.httpasyncclient>4.1.5</version.org.apache.httpcomponents.httpasyncclient>
         <version.org.apache.james.apache-mime4j>0.8.14</version.org.apache.james.apache-mime4j>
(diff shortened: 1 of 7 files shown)


bstansberry:
@fabiobrz It seems this broke something with the SecurityManager enabled. :-(

fabiobrz:
> @fabiobrz It seems this broke something with the SecurityManager enabled. :-(

Thanks @bstansberry - it seems they introduced a new permission check (`NetPermission("getProxySelector")`) via https://issues.apache.org/jira/browse/CXF-9218, and this makes the `ws` tests fail.

Several tests fail instead because of an existing permission check, i.e. `RuntimePermission("org.apache.cxf.permission")`, which I couldn't relate to any of the recent changes, unless it is about https://issues.apache.org/jira/browse/CXF-9212.

I'll continue investigating but I have added [one commit](https://github.com/wildfly/wildfly/pull/20142/changes/edf19512d13e92f63c054ffaab017fc8e9c7553a) already that I hope will fix the current failures.

Anyway, the [first change](https://issues.apache.org/jira/browse/CXF-9218) in Apache CXF doesn't look as something which should be done in a micro, and I'll probably log an issue to the Apache CXF Jira, also because the call should be wrapped in a call to `doPrivileged,`, consistently with what is done for similar cases.

As for the other failures, I'll have to investigate deeper.

bstansberry:
Following are CXF discussions around the SM problem that this PR currently works around by adding perms to deployments:

https://lists.apache.org/thread/3pbvn7x99fdo9g8drs6qyogzy9py95n6
https://lists.apache.org/thread/qqnh95r9no08g8zt67rsvdnf87593f40

Adding perms is not a viable end-user solution.

fabiobrz:
> Following are CXF discussions around the SM problem that this PR currently works around by adding perms to deployments:
> 
> https://lists.apache.org/thread/3pbvn7x99fdo9g8drs6qyogzy9py95n6 https://lists.apache.org/thread/qqnh95r9no08g8zt67rsvdnf87593f40
> 
> Adding perms is not a viable end-user solution.

Thanks @bstansberry - yes, this PR ended up being just a test for me to provide the APache CXF team with details about the regression, which is why I added the security permissions incrementally.
Let's wait for a 4.1.8 release :crossed_fingers:

fabiobrz:
Superseded by #20235

PR #20145, opened 2026-06-15
WFLY-22001 Deprecate statistics-enabled attribute on JGroups channel …

rhusar:
…resource

The statistics-enabled attribute on the channel resource calls JChannel.stats(boolean) which is a no-op since JGRP-2426 moved all stats collection to MsgStats in the transport protocol. The actual statistics collection is controlled by the stack's statistics-enabled attribute which propagats to individual protocols.

https://redhat.atlassian.net/browse/WFLY-22001


pferraro (approved):

bstansberry:
Thanks @rhusar and @pferraro

PR #20154, opened 2026-06-16
WFLY-22011 Update OIDC logout tests to use relative logout callback path

OndrejKotek:
https://redhat.atlassian.net/browse/WFLY-22011


darranl (approved):
None of the failures relate to logout support.

PR #20155, opened 2026-06-16
[WFLY-22010] Upgrade JBeret to 3.2.0.Beta2.

jamezp:
https://redhat.atlassian.net/browse/WFLY-22010

Tag: https://github.com/jberet/jsr352/releases/tag/3.2.0.Beta2
Diff: https://github.com/jberet/jsr352/compare/3.1.0.Final...3.2.0.Beta2

cc @theashiot


bstansberry:
Thanks @jamezp

PR #20157, opened 2026-06-17
WFLY-22012 Consolidate EJB client topology update wait timeout into AbstractClusteringTestCase (fixes ClientClusterNodeSelectorTestCase fails intermittently)

rhusar:
Move the duplicated CLIENT_TOPOLOGY_UPDATE_WAIT constant from 6 individual EJB remote test classes into a single EJB_CLIENT_TOPOLOGY_UPDATE_WAIT field. This also fixes the insufficient 2-second timeout in ClientClusterNodeSelectorTestCase (now aligned with the common 5-second value used by all other EJB testrs).

See details in https://redhat.atlassian.net/browse/WFLY-22012


pferraro (approved):

rhusar:

PR #20163, opened 2026-06-19
[WFLY-19563] Support use of virtual threads in the Jakarta Concurren…

emmartins:
…cy 3.1
Issue: https://redhat.atlassian.net/browse/WFLY-19563


emmartins:
@bstansberry honestly there is something here that is bottering me, which is the unit testing, specially the classes and resources duped in impl-30 and impl-31, but I think that is something worth an enhancement on its own. 
I even thought of the possibility of bring back to ee module all those tests, and all necessary subsystem stuff in the concurrency spi module, after all such tests are just about subsystem/parsing/models, but that would require some refactoring of the spi module, probably a better topic for WFLY 45+ roadmap, for instance it may make more sense to instead introduce a concurrency subsystem...

bstansberry:
I put the 42.x label on this because I think it is important to get into 42. And we've got a lot going on so it's easy to let it fall in a crack.

PR #20165, opened 2026-06-19
[WFLY-21273] CosServices.cfg gets created in /target

jmfinelli:
Addressed [WFLY-21273](https://redhat.atlassian.net/browse/WFLY-21273)

### DIFF
--- a/transactions/src/main/java/org/jboss/as/txn/service/OrbPortabilityEnvironmentService.java
+++ b/transactions/src/main/java/org/jboss/as/txn/service/OrbPortabilityEnvironmentService.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright The WildFly Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.jboss.as.txn.service;
+
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import com.arjuna.common.internal.util.propertyservice.BeanPopulator;
+import com.arjuna.orbportability.common.OrbPortabilityEnvironmentBean;
+
+import org.jboss.as.server.ServerEnvironment;
+import org.jboss.msc.Service;
+import org.jboss.msc.service.StartContext;
+import org.jboss.msc.service.StartException;
+import org.jboss.msc.service.StopContext;
+
+/**
+ * Sets up the {@link OrbPortabilityEnvironmentBean}
+ *
+ * @author <a href="mailto:jfinelli@redhat.com">Manuel Finelli</a>
+ */
+public final class OrbPortabilityEnvironmentService implements Service {
+
+    private final Consumer<OrbPortabilityEnvironmentBean> orbPortabilityEnvironmentBeanConsumer;
+    private final Supplier<ServerEnvironment> serverEnvironmentSupplier;
+
+    public OrbPortabilityEnvironmentService(final Consumer<OrbPortabilityEnvironmentBean> orbPortabilityEnvironmentBeanConsumer,
+                                            final Supplier<ServerEnvironment> serverEnvironmentSupplier) {
+        this.orbPortabilityEnvironmentBeanConsumer = orbPortabilityEnvironmentBeanConsumer;
+        this.serverEnvironmentSupplier = serverEnvironmentSupplier;
+    }
+
+    @Override
+    public void start(final StartContext context) throws StartException {
+        final OrbPortabilityEnvironmentBean orbPortabilityEnvironmentBean =
+                BeanPopulator.getDefaultInstance(OrbPortabilityEnvironmentBean.class);
+        orbPortabilityEnvironmentBean.setInitialReferencesRoot(
+                serverEnvironmentSupplier.get().getServerDataDir().getAbsolutePath());
+        orbPortabilityEnvironmentBeanConsumer.accept(orbPortabilityEnvironmentBean);
+    }
+
+    @Override
+    public void stop(final StopContext context) {
+        final OrbPortabilityEnvironmentBean orbPortabilityEnvironmentBean =
+                BeanPopulator.getDefaultInstance(OrbPortabilityEnvironmentBean.class);
+        orbPortabilityEnvironmentBean.setInitialReferencesRoot(null);
+        orbPortabilityEnvironmentBeanConsumer.accept(null);
+    }
+}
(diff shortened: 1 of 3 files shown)


rhusar (changes requested):
@jmfinelli The question begs to be asked – is the default value actually correct (IIRC it's user home)? Is that file meant to be shared by *all* instances running on the OS under same user account?

My hunch is the answer is no, so the true fix would then belong to the project or the integration logic, not the test suite.

rhusar:
> @jmfinelli The question begs to be asked – is the default value actually correct (IIRC it's user home)? Is that file meant to be shared by _all_ instances running on the OS under same user account?
> 
> My hunch is the answer is no, so the true fix would then belong to the project or the integration logic, not the test suite.

FWIW @tomjenkinson already investigated this a while back, so just in case he has an opinion on this :) otherwise please ignore.

jmfinelli:
[AFAIU](https://docs.oracle.com/javase/8/docs/api/java/io/File.html), `user.dir` is _[...] typically the directory in which the Java virtual machine was invoked._ IOW, it isn't the user's home directory.

rhusar:
> [AFAIU](https://docs.oracle.com/javase/8/docs/api/java/io/File.html), `user.dir` is _[...] typically the directory in which the Java virtual machine was invoked._ IOW, it isn't the user's home directory.

If so, in that case it **significantly** worse, because `standalone.sh` does not tinker with `user.dir` in any way. e.g. if my `pwd` is root `/` then the `.cos` file will completely fail to write!

I am pretty sure the file is meant to be scoped to the installation directory, typically in `data/` dir like all other subsystem do.

jmfinelli:
Could you please advise how to implement what you suggested in your comment? The purpose of this PR was only to fix the issue reported in WFLY-21273.

Our of curiosity:

> If so, in that case it significantly worse, because `standalone.sh` does not tinker with `user.dir` in any way. e.g. if my `pwd` is root `/` then the `.cos` file will completely fail to write!

I'm not sure I'm following, sorry. AFAIU, `user.dir` defaults to the directory where the JVM was started. Why would your `pwd` be `/`? Also, I didn't know WildFly could run without write permissions in the directory where it's installed.

rhusar:
> Could you please advise how to implement what you suggested in your comment? The purpose of this PR was only to fix the issue reported in WFLY-21273.

The fix is to not rely on `user.dir` as explained above. The file should be created relative to the server installation directory.

I filed that issue to highlight the problem. While your fix fixes the issue with the file being generated outside of Maven `target/` directory, I am questioning whether this should not be fixed properly for the entire server.

> Our of curiosity:
> 
> > If so, in that case it significantly worse, because `standalone.sh` does not tinker with `user.dir` in any way. e.g. if my `pwd` is root `/` then the `.cos` file will completely fail to write!
> 
> I'm not sure I'm following, sorry. AFAIU, `user.dir` defaults to the directory where the JVM was started. Why would your `pwd` be `/`? Also, I didn't know WildFly could run without write permissions in the directory where it's installed.

We do not mandate what pwd people can start WF from, that would not make any sense.

Anyway why don't you just test it? It's really simple to demonstrate what the `user.dir` resolves to. Lets start from drive root:

```
[rhusar@puglia /]$ cd /
[rhusar@puglia /]$ pwd
/
[rhusar@puglia /]$ /Users/rhusar/git/wildfly/build/target/wildfly-41.0.0.Beta1-SNAPSHOT/bin/standalone.sh -c standalone-ha.xml
=========================================================================

  JBoss Bootstrap Environment

  JBOSS_HOME: /Users/rhusar/git/wildfly/build/target/wildfly-41.0.0.Beta1-SNAPSHOT
...
```

Say I use a servlet and add `info.append("user.dir: ").append(System.getProperty("user.dir"));` to find out the location where the CosServices.cfg would be created by the subsystem as you say that's the default.

When I access the servlet I get as expected

```
user.dir: /
```

..which would fail to crate the file there.

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemAdd.java
```diff
@@ -520,6 +522,13 @@ private void performCoordinatorEnvBoottime(OperationContext context, ModelNode c
         if (jts) {
             orbSupplier = sb.requires(ServiceName.JBOSS.append("iiop-openjdk", "orb-service"));
             sb.requires(CorbaNamingService.SERVICE_NAME);
+            sb.requires(TxnServices.JBOSS_TXN_ORB_PORTABILITY_ENVIRONMENT);
+
+            // install OrbPortabilityEnvironmentBean service
+            final ServiceBuilder<?> orbPortSB = context.getCapabilityServiceTarget().addService();
+            final Consumer<OrbPortabilityEnvironmentBean> orbPortConsumer = orbPortSB.provides(TxnServices.JBOSS_TXN_ORB_PORTABILITY_ENVIRONMENT);
+            final Supplier<ServerEnvironment> serverEnvSupplier = orbPortSB.requires(ServerEnvironmentService.SERVICE_NAME);
```
bstansberry:
```suggestion
            final Supplier<ServerEnvironment> serverEnvSupplier = orbPortSB.requires(ServerEnvironment.SERVICE_DESCRIPTOR);
```

ServerEnvironmentService.SERVICE_NAME is deprecated.

This class uses it in another spot, but for new code we might as well use the non-deprecated one.

transactions/src/main/java/org/jboss/as/txn/service/OrbPortabilityEnvironmentService.java
```diff
@@ -0,0 +1,49 @@
...
+        final OrbPortabilityEnvironmentBean orbPortabilityEnvironmentBean =
+                BeanPopulator.getDefaultInstance(OrbPortabilityEnvironmentBean.class);
+        orbPortabilityEnvironmentBean.setInitialReferencesRoot(
+                serverEnvironmentSupplier.get().getServerDataDir().getAbsolutePath());
+        orbPortabilityEnvironmentBeanConsumer.accept(orbPortabilityEnvironmentBean);
+    }
+
+    @Override
+    public void stop(final StopContext context) {
+        orbPortabilityEnvironmentBeanConsumer.accept(null);
```
rhusar, line 50: The `setInitialReferencesRoot` is set in start but never cleared in stop.

Sibling services like JTAEnvironmentBeanService do clear fields correctly, so apply the same pattern here.

rhusar (changes requested):

transactions/src/main/java/org/jboss/as/txn/subsystem/TransactionSubsystemAdd.java
jmfinelli: Thanks Brian!

transactions/src/main/java/org/jboss/as/txn/service/OrbPortabilityEnvironmentService.java
jmfinelli: Thanks Rado!

bstansberry (approved):

bstansberry:
The points @rhusar raised above have been addressed, so I'm going to merge this.

bstansberry:
Thanks @jmfinelli and @rhusar

PR #20168, opened 2026-06-20
WFLY-22017 Ensure adhoc TLS sockets are registered with the socket-binding manager

pferraro:
https://redhat.atlassian.net/browse/WFLY-22017


clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
```diff
@@ -12,51 +12,76 @@
...
+        return socket;
     }
 
     @Override
     public ServerSocket createServerSocket(String name) throws IOException {
         SocketBinding binding = this.bindings.get(name);
-        org.jboss.as.network.ManagedServerSocketFactory factory = this.manager.getServerSocketFactory();
-        return (binding != null) ? factory.createServerSocket(binding.getName()) : factory.createServerSocket();
+        ServerSocket socket = (this.serverSSLContext != null) ? this.serverSSLContext.getServerSocketFactory().createServerSocket() : (binding != null) ? this.manager.getServerSocketFactory().createServerSocket(binding.getName()) : this.manager.getServerSocketFactory().createServerSocket();
+        if (this.clientSSLContext != null) {
```
rhusar:
```suggestion
        if (this.serverSSLContext != null) {
```

The socket creation correctly uses serverSSLContext, but the registration guard incorrectly checks clientSSLContext. This should never be a problem in practice, but lets make this consistent.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
```diff
@@ -81,37 +106,43 @@ public MulticastSocket createMulticastSocket(String name, SocketAddress bindAddr
 
     @Override
     public SocketChannel createSocketChannel(String name) throws IOException {
-        return this.createNetworkChannel(name, SelectorProvider::openSocketChannel, SocketBindingManager.NamedManagedBindingRegistry::registerChannel, SocketBindingManager.UnnamedBindingRegistry::registerChannel);
+        if ((this.clientSSLContext != null) && (name != null)) {
```
rhusar: The old TLSManagedSocketFactory logged this warning unconditionally (without name != null check). The new code suppresses the warning for unnamed channel request.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
```diff
@@ -81,37 +106,43 @@ public MulticastSocket createMulticastSocket(String name, SocketAddress bindAddr
...
+        SocketBinding binding = this.bindings.get(name);
+        SocketChannel channel = this.provider.openSocketChannel();
+        this.registrations.put(channel, (binding != null) ? this.manager.getNamedRegistry().registerChannel(binding.getName(), channel) : this.manager.getUnnamedRegistry().registerChannel(channel));
+        return channel;
     }
 
     @Override
     public ServerSocketChannel createServerSocketChannel(String name) throws IOException {
-        return this.createNetworkChannel(name, SelectorProvider::openServerSocketChannel, SocketBindingManager.NamedManagedBindingRegistry::registerChannel, SocketBindingManager.UnnamedBindingRegistry::registerChannel);
+        if ((this.serverSSLContext != null) && (name != null)) {
```
rhusar: Same here.

rhusar (changes requested):
Needs fixes.

However, it's also not clear to me when there is a code path in JGroups that passes null to createSocketChannel or createServerSocketChannel as I cannot find any in any code base. Can you please give me some pointers?

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
pferraro: Correct.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
pferraro: True. Fixed.

pferraro:
> Needs fixes.

@rhusar The only requested change was non-essential - did you intend to request anything else?

rhusar:
> > Needs fixes.
> 
> @rhusar The only requested change was non-essential - did you intend to request anything else?

I am still not clear about the question in – https://github.com/wildfly/wildfly/pull/20168#pullrequestreview-4538419320 – because otherwise we can just require the name to be not null

pferraro:
> However, it's also not clear to me when there is a code path in JGroups that passes null to createSocketChannel or createServerSocketChannel as I cannot find any in any code base.

I should hope so.  When this class was written, there were a few protocols that used the SocketFactory with a null service name.

> ... because otherwise we can just require the name to be not null

Even if there are no more instance of this upstream, we cannot suddenly require this to be non-null without breaking existing external/custom protocols that might be relying on an optional service name.

I'm not against adding this requirement - but we would need a period of deprecation first.  Feel free to create a jira.  Either way, such a change is not in scope for this fix.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
rhusar: Discussed over chat, that we need a separate message for situation where service is null so that the information is still conveyed in the logs.

Also to note here, that standard jgroups never creates unnamed channels.

clustering/jgroups/extension/src/main/java/org/jboss/as/clustering/jgroups/ManagedSocketFactory.java
rhusar: Discussed over chat, that we need a separate message for situation where service is null so that the information is still conveyed in the logs.

Also to note here, that standard jgroups never creates unnamed channels.

rhusar (changes requested):

rhusar (approved):
LGTM now, all feedback and wishlist items addressed with proper i18n :-)

rhusar:
Thanks @pferraro

PR #20169, opened 2026-06-22
WFLY-22023 Upgrade wildfly-clustering to 10.0.9.Final.

pferraro:
https://redhat.atlassian.net/browse/WFLY-22023

Resolves https://redhat.atlassian.net/browse/WFLY-21920


rhusar (approved):
LGTM

You are right, I could have pushed the fix directly upstream but I was trying to get something quickly backportable to 40.x.

PR #20174, opened 2026-06-22
[WFLY-22015] Stop publishing pre-built server tars to Maven

bstansberry:
https://redhat.atlassian.net/browse/WFLY-22015

Continue to build tars we attach to the GH release, just don't install/deploy them.

Stop building tars that we don't attach to the GH release.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -374,6 +374,36 @@
                             </execution>
                         </executions>
                     </plugin>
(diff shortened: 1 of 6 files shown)


rhusar (approved):
LGTM this does achieve what it sets out to.

However, just to note, the description is slightly misleading "pre-built server tars" as this also excludes all source tar.gz from `assembly-src.xml`, not just the pre-build ones. Which is correct.

bstansberry:
Thanks @rhusar

PR #20176, opened 2026-06-23
WFLY-22024 Deprecate AJP connector in Undertow subsystem

rhusar:
- Mark ajp-listener resource as deprecated since model version 15.0.0
- Remove AJP listener and socket binding from default configuration
- Update mod_cluster default listener from AJP to HTTP

Our deprecation message is this:

The AJP connector is deprecated due to inherent security deficiencies (no encryption, no authentication, CVEs), inability to support modern protocols (WebSocket, HTTP/2), and protocol stagnation (no development since 2001). Migrate to HTTP proxying which offers all the same functionality with proper TLS and mutual authentication support.

Jira
https://redhat.atlassian.net/browse/WFLY-22024


ropalka (approved):
LGTM

rhusar:
We have discussed this a while ago in the mod_cluster team, so it's really high time we pull the plug. FYI @jfclere @jajik

jfclere (commented):
DON'T USE AJP, it was created in 2000 and never improved since then.

jajik (approved):

PR #20184, opened 2026-06-24
[WFLY-20744] Throw ISE in case ConcurrentContextService is not started

emmartins:
Issue: https://redhat.atlassian.net/browse/WFLY-20744
Description: in short we use the ConcurrentContext service to learn if the deployment is running or not, and if not we invalidate the context setup, what leads to the abort and cancelling of periodic tasks left over by the deployment.
The changes regarding Component and Web/CDI setup action were due to an unknown issue, those components were not depending on the ConcurrentContextService, and with the main change this additional issue was noticed in our testsuite.

### DIFF
--- a/concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
+++ b/concurrency/spi/src/main/java/org/jboss/as/ee/concurrent/ConcurrentContext.java
@@ -86,6 +86,14 @@ public class ConcurrentContext {
 
     /**
      *
+     * @return
+     */
+    public ServiceName getServiceName() {
+        return serviceName;
+    }
+
+    /**
+     *
      * @param serviceName
      */
     public void setServiceName(ServiceName serviceName) {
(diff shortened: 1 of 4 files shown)


emmartins:
Please refrain from merging this, I am still analysing if there are any 2nd effects of this change, specially wrt graceful shutdown

emmartins:
For the record EE11 TCK is green (besides the usual signature test fail).

bstansberry (commented):
@emmartins There is a small race here, depending on how legalistic we are being about the definition of 'task's component is not started' in the spec statement 'All tasks submitted to an executor must not run if task’s component is not started'.

https://github.com/wildfly/wildfly/pull/20184/changes#diff-5e83e19b0b5a8e86e06badcd319e610479c530d7a2d0f61d95c2bffbcce15749R117 results in the service for the component having a dep on the ConcurrentContextService. So that service will stop *before* ConcurrentContextService does, so some task might run in that window.

From a legalistic point I think this is ok because it's valid IMO to consider the ConcurrentContextService to be part of the 'task's component'. How we architect the pieces that go into the EE concept of a component is up to us. The ConcurrentContextService I'm talking about is explicitly associated with the component.

From a practical point of view I think this is ok, since at a minimum it greatly improves the situation where tasks can continue running after the app is undeployed.

Based on the above, I'll approve this and merge later today, but after giving you a bit of time to respond if you wish.

emmartins:
@bstansberry agree, IMHO this is fine from the point of view that considers the concurrent context part of the component.

bstansberry (approved):

bstansberry:
Thanks @emmartins

PR #20188, opened 2026-06-26
WFLY-22038 Coarse/FineHotRodSessionExpirationTestCase fails intermittently - use dedicated deployment descriptors for HotRod session expiration tests

rhusar:
The DEFAULT_CONFIGURATION used by jboss-all_fine.xml omits expiration settings, so the Infinispan server defaults to a 60-second reaper wake-up interval causing the expiration test to miss the reaper cycle. Use a dedicated descriptor with cache-configuration="default" referencing the server template that defines a 1-second expiration reaper interval.

https://redhat.atlassian.net/browse/WFLY-22038


rhusar:
I also need to fix the *Coarse* variant...

rhusar:
> I also need to fix the _Coarse_ variant...

Fixed now.

rhusar:
(rebased only)

pferraro (approved):

rhusar:

PR #20190, opened 2026-06-26
[WFLY-21772] Remove unnecessary domain[-ha]-profile feature group dif…

bstansberry:
…ferences between WFP and standard WF

https://redhat.atlassian.net/browse/WFLY-21772

This has no effect on what ends up being provisioned because the resources that are different between the feature-groups changed by this end up being provisioned for other reasons. So this is just codebase housekeeping to save maintainers wondering why code differs in the different feature-packs.


rhusar (approved):
LGTM; aligns the preview feature pack's profiles with the standard and legacy feature packs

rhusar:
Thanks @bstansberry

PR #20198, opened 2026-06-29
[WFLY-22042] Upgrade Undertow EE to 1.0.1.Final and 2.0.1.Final

ropalka:
https://redhat.atlassian.net/browse/WFLY-22042

### DIFF
--- a/boms/legacy-ee/pom.xml
+++ b/boms/legacy-ee/pom.xml
@@ -34,7 +34,7 @@
         <version.com.fasterxml.classmate>1.5.1</version.com.fasterxml.classmate>
         <!-- n.b. H2 version here is driven by what Hibernate ORM version tests with as the H2 dialect! -->
         <version.com.h2database>2.2.224</version.com.h2database>
-        <version.io.undertow.ee>1.0.0.Final</version.io.undertow.ee>
+        <version.io.undertow.ee>1.0.1.Final</version.io.undertow.ee>
         <version.io.undertow.jastow>2.2.9.Final</version.io.undertow.jastow>
         <version.jakarta.annotation.jakarta-annotation-api>2.1.1</version.jakarta.annotation.jakarta-annotation-api>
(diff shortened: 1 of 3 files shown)


bstansberry (changes requested):
Sorry, @ropalka -- you've stumbled into a bit of cruft that needs cleaning.

We should not be declaring version.io.undertow in the root pom.xml any more. We should be using the version that comes from WildFly Core.

We intended to remove in WF 40 the **temporary** override in full WF of the undertow-core version coming from WF Core. But shortly before the 40 Final release we needed to upgrade undertow-core and there was no time to upgrade it in WF Core and then release and integrate WF C Core with that change. So we restored the override. And then we forgot to remove it again.

Can you:

1) Assign https://redhat.atlassian.net/browse/WFLY-21834 to yourself
2) Include a commit in your PR that reverts https://github.com/wildfly/wildfly/commit/19c309c8e80e238b5b04c18bbeae3bcb62433f44. using WFLY-21834 in the commit message.
3) Change the title/description of WFLY-22042 and your commit here to make it solely about Undertow EE.

ropalka:
Will do @bstansberry - but I need to wait for https://github.com/wildfly/wildfly-core/pull/6837 being applied first and WildFly Core must be upgraded in WildFly.

bstansberry:
@ropalka Ok, let's just go with what you have here then, and I'll assign WFLY-21834 to myself and will do it when @yersan sends up the next core upgrade.

I'd rather get the undertow-ee updates in now, versus having to do them at the last minute when the core upgrade comes in.

bstansberry (approved):

yersan:
I've already prepared the Core release with Undertow 1.4.2. I'm going to take the liberty to move on with this by merging this PR and assigning [WFLY-21834](https://redhat.atlassian.net/browse/WFLY-21834) since both of you are on PTO today.

yersan:
Thanks @ropalka @bstansberry

